Issue #3097, opened 2020-03-13
[Bug] tablet meta is not updated after compaction


morningman:
**Describe the bug**
After a tablet compaction, the new rowsets will be added to the tablet meta, and old rowsets will be deleted.
But we find that sometimes the tablet meta is not updated correctly. This will cause the metadata and data to be inconsistent, making it impossible to read the data correctly.

PR #3098, opened 2020-03-13
[Bug] tablet meta is not updated correctly after compaction

morningman:
This CL try to fix a potential bug describe in ISSUE: #3097. But I'm not sure this is the root cause.

Also remove lots of verbose log, and fix a memory leak.

### DIFF
--- a/be/src/olap/rowset/alpha_rowset.cpp
+++ b/be/src/olap/rowset/alpha_rowset.cpp
@@ -58,7 +58,7 @@ OLAPStatus AlphaRowset::create_reader(std::shared_ptr<RowsetReader>* result) {
 }
 
 OLAPStatus AlphaRowset::remove() {
-    LOG(INFO) << "begin to remove files in rowset " << unique_id()
+    VLOG(3) << "begin to remove files in rowset " << unique_id()
             << ", version:" << start_version() << "-" << end_version()
             << ", tabletid:" << _rowset_meta->tablet_id();
     for (auto segment_group : _segment_groups) {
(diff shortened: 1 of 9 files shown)


be/src/olap/rowset/alpha_rowset.cpp
```diff
@@ -166,7 +166,9 @@ OLAPStatus AlphaRowset::split_range(
     RowBlockPosition step_pos;
 
     std::shared_ptr<SegmentGroup> largest_segment_group = _segment_group_with_largest_size();
-    if (largest_segment_group == nullptr) {
+    if (largest_segment_group == nullptr || largest_segment_group->current_num_rows_per_row_block() == 0) {
```
kangkaisen: Why add a check for `largest_segment_group->current_num_rows_per_row_block() == 0`?
You encounter this issue https://github.com/apache/incubator-doris/issues/2750?

be/src/olap/task/engine_publish_version_task.cpp
```diff
@@ -60,7 +60,7 @@ OLAPStatus EnginePublishVersionTask::finish() {
             OLAPStatus publish_status = OLAP_SUCCESS;
             TabletInfo tablet_info = tablet_rs.first;
             RowsetSharedPtr rowset = tablet_rs.second;
-            LOG(INFO) << "begin to publish version on tablet. "
+            VLOG(3) << "begin to publish version on tablet. "
```
kangkaisen: I think we shouldn't remove these logs.
These logs are very important for track transaction issue.

kangkaisen (changes requested):

be/src/olap/rowset/alpha_rowset.cpp
morningman: Yes, so I and a check here to avoid coredump in our product env, but leave a warning log to observe. This problem still needs to be analyzed.

be/src/olap/task/engine_publish_version_task.cpp
morningman: There are too many this kind of logs in our env. Why you need a "begin" log?

be/src/olap/rowset/alpha_rowset.cpp
kangkaisen: OK

be/src/olap/task/engine_publish_version_task.cpp
kangkaisen: Would better keep all publish version and transaction log. Our publish version and transaction performance is still poor. We need these logs to track issue.

morningman: OK, I will leave the begin and finish log

kangkaisen (approved):
+1. LGTM

Issue #3102, opened 2020-03-13
[Insert] Encounter label already exist exception


morningman:
**Describe the bug**
When executing insert stmt with label like:

```
insert into tbl with label abc select 1,2;
```

If first execution encounter some exception like `tablet xxx has few replica`, then if we execute with same label again, we will meet exception:

```
ERROR 1064 (HY000): errCode = 2, detailMessage = Label [abc] has already been used.
```

Which is unexpected.

**Why**

Because the first exception is thrown in analysis phase, and we didn't abort the transaction of this insert operation when encountering analysis exception.

PR #3103, opened 2020-03-13
[Insert] Fix bug that insert meet unexpected "label already exists" exception

morningman:
This CL will abort the transaction of an insert operation when encountering exception thrown in analysis phase.

ISSUE: #3102


kangkaisen (approved):
+1. LGTM

Issue #3105, opened 2020-03-13
[Alter] Alter job got stuck because of table is untable


morningman:
**Describe the bug**
Previously, we allowed users to submit alter jobs even if the table is unstable. But this may cause the alter job to be blocked. Because when the alter job starts, the state of the table will change to SCHEMA_CHANGE/ROLLUP, and the tablet repair logic will not process the tablet of the table in the SCHEMA_CHANGE/ROLLUP state. As a result, the Alter job and the tablet repair logic wait for each other to complete.

PR #3106, opened 2020-03-13
[Alter] Alter job got stuck because of table is untable

morningman:
This CL solve the issue #3105 

I add a new temporary table state `WAITING_STABLE`. 

When an alter job is ready to start, it checks whether the table is stable. If it is not stable,
the table state is set to `WAITING_STABLE`. In this state, the tablet repair logic will continue to 
repair the tablet until the table becomes stable.

After that, the table state will be reset to SCHEMA_CHANGE/ROLLUP and alter operations will begin.

This is just a temporary state, it does not need to be persistent, and only the master FE can see this state.

### DIFF
--- a/fe/src/main/java/org/apache/doris/alter/RollupJobV2.java
+++ b/fe/src/main/java/org/apache/doris/alter/RollupJobV2.java
@@ -162,7 +162,7 @@ public class RollupJobV2 extends AlterJobV2 {
             }
         }
         MarkedCountDownLatch<Long, Long> countDownLatch = new MarkedCountDownLatch<Long, Long>(totalReplicaNum);
-        db.readLock();
+        db.writeLock();
         try {
             OlapTable tbl = (OlapTable) db.getTable(tableId);
             if (tbl == null) {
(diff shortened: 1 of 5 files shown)


fe/src/main/java/org/apache/doris/alter/RollupJobV2.java
```diff
@@ -162,7 +162,7 @@ protected void runPendingJob() throws AlterCancelException {
             }
         }
         MarkedCountDownLatch<Long, Long> countDownLatch = new MarkedCountDownLatch<Long, Long>(totalReplicaNum);
-        db.readLock();
+        db.writeLock();
```
kangkaisen: Could we reduce the write lock scope? because the `tbl.isStable` and `CreateReplicaTask` both need to iterate every replica.

morningman: OK, I will modify this.

fe/src/main/java/org/apache/doris/alter/SchemaChangeJobV2.java
```diff
@@ -185,22 +189,13 @@ protected void runPendingJob() throws AlterCancelException {
             }
         }
         MarkedCountDownLatch<Long, Long> countDownLatch = new MarkedCountDownLatch<>(totalReplicaNum);
-        db.readLock();
+        db.writeLock();
```
kangkaisen: Should still use `db.readLock`.

kangkaisen (approved):
+1, LGTM

PR #3110, opened 2020-03-13
[UT] Fix some BE unit tests

acelyc111:
And also support graceful exit for StorageEngine to avoid hang too long
time in unit test.


be/test/olap/rowset/segment_v2/bloom_filter_index_reader_writer_test.cpp
```diff
@@ -33,14 +33,15 @@
...
 namespace segment_v2 {
 
+const std::string dname = "./ut_dir/bloom_filter_index_reader_writer_test";
+
 class BloomFilterIndexReaderWriterTest : public testing::Test {
    public:
-    BloomFilterIndexReaderWriterTest() { }
-    virtual ~BloomFilterIndexReaderWriterTest() {
+    virtual void SetUp() {
+        ASSERT_TRUE(boost::filesystem::remove_all(dname));
```
chaoyli: It's better to use Env to replace boost::filesystem::remove_all

chaoyli (changes requested):

chaoyli (approved):

Issue #3112, opened 2020-03-14
[Bug] mapping columns are not exist in the table schema


wyb:
**Describe the bug**
Mapping columns are not checked in stream load job whether they exist in the table schema or not. 
If columns in the table have default value or allow null value, the load job will write the default value or null.

table schema:
k2 int, 
k3 varchar(128), 
v bigint sum not null

stream load command:
curl --location-trusted -u root: -T file -H "label:label0" -H "columns: tmp_k2,k3,v,k22=tmp_k2+1" http://host:port/api/db/table/_stream_load

k22 is not exist in table schema, and the load job writes null for k2 column.

PR #3113, opened 2020-03-14
[Bug] mapping columns are not exist in the table schema

wyb:
Check mapping columns, return errors to users if columns are not exist in the table schema.
#3112


kangkaisen (changes requested):
LGTM. But please add a UT in `StreamLoadScanNodeTest` to cover this bug.

morningman (approved):
LGTM

kangkaisen (approved):

PR #3115, opened 2020-03-14
Support schema keyword to be compatible with the mysql syntax

caiconghui:


caiconghui:
for #3115

morningman (dismissed):
LGTM, but could you add a unit test for this syntax? You can just add it in QueryPlanTest.java, by creating a database using this new syntax.

caiconghui:
> LGTM, but could you add a unit test for this syntax? You can just add it in QueryPlanTest.java, by creating a database using this new syntax.

OK

morningman (approved):
LGTM

Issue #2910, opened 2020-02-15
[PROPOSAL] Remove unused LLVM related codes


imay:
At present, there are many LLVM related codes in code base, but these codes are not really used. During my attempt to upgrade GCC, I found that the higher version of GCC is not compatible with the LLVM 3.4.2 version currently used by Doris. I propose to delete all LLVM related code in the code. With this deletion, we can gain the following benefits

1. Remove irrelevant code to make our code more refined
2. Ability to upgrade our toolchain from GCC7 to GCC9 or higher
3. In the future we will be able to use later versions of LLVM. The APIs of the LLVM versions are not compatible, and there is no way to smoothly upgrade to a higher version. And a large amount of code that is no longer used will become a constraint on our future use of higher versions of LLVM.

Since our LLVM code is not really used, this operation will not bring other harm.

HangyuanLiu:
+1
I agree with you
Thanks,
Hangyuan Liu


------------------&nbsp;原始邮件&nbsp;------------------
发件人:&nbsp;"ZHAO Chun"<notifications@github.com&gt;;
发送时间:&nbsp;2020年2月15日(星期六) 下午4:58
收件人:&nbsp;"apache/incubator-doris"<incubator-doris@noreply.github.com&gt;;
抄送:&nbsp;"Subscribed"<subscribed@noreply.github.com&gt;;
主题:&nbsp;[apache/incubator-doris] [PROPOSAL] Remove unused LLVM related codes (#2910)




At present, there are many LLVM related codes in code base, but these codes are not really used. During my attempt to upgrade GCC, I found that the higher version of GCC is not compatible with the LLVM 3.4.2 version currently used by Doris. I propose to delete all LLVM related code in the code. With this deletion, we can gain the following benefits
  
Remove irrelevant code to make our code more refined
 
Ability to upgrade our toolchain from GCC7 to GCC9 or higher
 
In the future we will be able to use later versions of LLVM. The APIs of the LLVM versions are not compatible, and there is no way to smoothly upgrade to a higher version. And a large amount of code that is no longer used will become a constraint on our future use of higher versions of LLVM.
  
Since our LLVM code is not really used, this operation will not bring other harm.

trueeyu:
Let me handle it.

PR #3116, opened 2020-03-15
Do not build llvm thirdparty

lingbin:
LLVM related codes have already be removed in master branch,
so there is no need to build llvm tool(which need a long
time to compile it).

Currently, some old release of Doris may still need it, so
for now, we just comment it, instead of remove it.

#2910


imay (approved):
LGTM

Issue #2663, opened 2020-01-06
[Proposal] Support partition overwrite operation


morningman:
# Partition Overwrite Operation

Some businesses need data backtracking, such as revising historical data or refreshing historical data. In the current version of Doris, there are only two ways to do this. One is to use the `REPLACE` aggregation model to overwrite by re-loading the data with same keys. The second way is to delete the data and then re-load the data.

The first method can guarantee that the data is overwritten atomically, but the aggregation type is limited to `REPLACE` and can only overwrite data with the same keys. In the second method, there is a gap between deleting the data and successfully loading the new data.

Therefore, we need a way that can overwrite data atomically, and without the limitations of the aggregation model.

## Two methods

There are two methods to achieve this.

1. Temporary partition

    For partitions that need to be overwritten, users can first create a temporary partition and write data to this temporary partition. After the data is written, User can replace the original partition with this temporary partition by using an atomic replacement command.

2. Overwrite directly

    Doris's underlying data is managed by MVCC. Therefore, we can add a special version, named `overwrite version` in tablet's meta, to ensure that when data is read, all data before that version can be skipped to achieve the purpose of overwriting.

Both methods have advantages and disadvantages.

The **disadvantage** of the first method is that the operation is tedious. The user needs to perform the three steps of `create temporary partition-> load data-> replace partition`. And creating temporary partitions means creating new tablets. If there are a large number of overwrite operations, a large number of temporary tablets will be generated, which is not friendly to metadata management and garbage collection mechanisms. However, the **advantage** of this method is that users can load the data in multiple batches, and then overwrite and replace them at one time through the partition replacement operation.

The **advantage** of the second method is to conform to the user's habits of overwriting write operations. Users can set any batch of load as a overwrite load. And there is no additional overhead. The **disadvantage** is that the user needs to load all the data that needs to be overwritten in a batch of load. If the amount of data to be overwritten is large, such as 100 GB or TB level, this method will lead to a higher probability of failure in the overwrite operation, and the retry cost is high.

Both methods have their own advantages and disadvantages and applicable scenarios. So I plan to support both.

### Temporary Partition (DEPRECATED, SEE BELOW)

The temporary partition method needs to consider the following:

1. How to make the load operation aware of the existence of the temporary partition? Or how to distinguish between temporary and formal partitions?
2. After the temporary partition is created, how to ensure that the formal partition will not be modified to ensure that the schema of the two partitions is exactly the same when the partition is eventually replaced?
[...]

PR #3120, opened 2020-03-16
[Temp Partition] Support loading data into temp partitions

morningman:
Related issue: #2663, #2828.

This CL support loading data into specified temporary partitions.

```
INSERT INTO tbl TEMPORARY PARTITIONS(tp1, tp2, ..) ....;

curl .... -H "temporary_partition: tp1, tp, .. "  ....

LOAD LABEL db1.label1 (
DATA INFILE("xxxx") 
INTO TABLE `tbl2`
TEMPORARY PARTITION(tp1, tp2, ...)
...
```

NOTICE: this CL change the FE meta version to 77.

There 3 major changes in this CL

## Syntax reorganization

Reorganized the syntax related to the `specify-partitions`. Removed some redundant syntax
 definitions, and unified the syntax related to the `specify-partitions` under one syntax entry.

## Meta refactor

In order to be able to support specifying temporary partitions, 
I made some changes to the way the partition information in the table is stored.

Partition information is now organized as follows:

The following two maps are reserved in OlapTable for storing formal partitions:

    ```
    idToPartition
    nameToPartition
    ```

Use the `TempPartitions` class for storing temporary partitions.

All the partition attributes of the formal partition and the temporary partition,
such as the range, the number of replicas, and the storage medium, are all stored
in the `partitionInfo` of the OlapTable.

In `partitionInfo`, we use two maps to store the range of formal partition
and temporary partition:

    ```
    idToRange
    idToTempRange
    ```

Use separate map is because the partition ranges of the formal partition and
the temporary partition may overlap. Separate map can more easily check the partition range.

All partition attributes except the partition range are stored using the same map,
and the partition id is used as the map key.

## Method to get partition

A table may contain both formal and temporary partitions.
There are several methods to get the partition of a table.
Typically divided into two categories:

1. Get partition by id
2. Get partition by name

According to different requirements, the caller may want to obtain
a formal partition or a temporary partition. These methods are
described below in order to obtain the partition by using the correct method.

1. Get by name

This type of request usually comes from a user with partition names. Such as
`select * from tbl partition(p1);`.
This type of request has clear information to indicate whether to obtain a
formal or temporary partition.
Therefore, we need to get the partition through this method:

`getPartition(String partitionName, boolean isTemp)`

To avoid modifying too much code, we leave the `getPartition(String
partitionName)`, which is same as:

`getPartition(partitionName, false)`

2. Get by id

This type of request usually means that the previous step has obtained
certain partition ids in some way,
so we only need to get the corresponding partition through this method:

`getPartition(long partitionId)`.

This method will try to get both formal partitions and temporary partitions.

3. Get all partition instances

Depending on the requirements, the caller may want to obtain all formal
partitions,
all temporary partitions, or all partitions. Therefore we provide 3 methods,
the caller chooses according to needs.

`getPartitions()`
`getTempPartitions()`
`getAllPartitions()`


fe/src/main/java/org/apache/doris/analysis/AdminRepairTableStmt.java
```diff
@@ -62,8 +63,12 @@ public void analyze(Analyzer analyzer) throws UserException {
 
         tblRef.getName().setDb(dbName);
 
-        if (tblRef.getPartitions() != null && !tblRef.getPartitions().isEmpty()) {
-            partitions.addAll(tblRef.getPartitions());
+        PartitionNames partitionNames = tblRef.getPartitionNames();
```
kangkaisen: Could we encapsulate the following logic?
```
 PartitionNames partitionNames = tblRef.getPartitionNames();
 if (partitionNames != null) {
            if (partitionNames.isTemp()) {
                throw new AnalysisException("xxxx");
            }
```

fe/src/main/java/org/apache/doris/planner/OlapTableSink.java
```diff
@@ -293,7 +293,11 @@ private TOlapTableLocationParam createLocation(OlapTable table) throws UserExcep
         TOlapTableLocationParam locationParam = new TOlapTableLocationParam();
         // BE id -> path hash
         Multimap<Long, Long> allBePathsMap = HashMultimap.create();
-        for (Partition partition : table.getPartitions()) {
+        for (Partition partition : table.getAllPartitions()) {
+            if (!partitionIds.isEmpty() && !partitionIds.contains(partition.getId())) {
```
kangkaisen: Don't need to check again, we have checked in `init` method.

morningman: I didn't get you point? no need to check what ?

fe/src/main/java/org/apache/doris/analysis/AdminRepairTableStmt.java
morningman: I can just change it to:
```
if (tblRef.isTempPartitionSpecified()) {
        throw new AnalysisException("Do not support (cancel)repair temporary partitions");
}
if (tblRef.getPartitionNames() != null) {
        partitions.addAll(tblRef.getPartitionNames().getPartitionNames());
}
```

Does it better?

fe/src/main/java/org/apache/doris/planner/OlapTableSink.java
kangkaisen: `if (!partitionIds.isEmpty() && !partitionIds.contains(partition.getId())) {`

fe/src/main/java/org/apache/doris/analysis/AdminRepairTableStmt.java
kangkaisen: I don't think which is better.  Let's keep it unchanged.

fe/src/main/java/org/apache/doris/planner/OlapTableSink.java
morningman: Here I use this `if (!partitionIds.isEmpty() && !partitionIds.contains(partition.getId()))` to filter all unrelated partitions. I didn't see any such check logic in `init` method?

kangkaisen (approved):
+1. LGTM

PR #3123, opened 2020-03-16
[Bug] Fix bug of mem limit checking in BrokerScanNode

vagetablechicken:
Ref https://github.com/apache/incubator-doris/issues/3092

BrokerScanNode set mem tracker in https://github.com/apache/incubator-doris/blob/dc07182bd48f8cba6fafbfd491859e94a544abaf/be/src/exec/exec_node.cpp#L185
limit_exceed always returns false.
Use any_limit_exceed instread, to limit memory correctly.


morningman (approved):
LGTM

PR #3128, opened 2020-03-17
[Thrift] Fix a bug when ThriftClientImpl close() error

acelyc111:
In ThriftClientImpl close(), the under layer TTransport may throw an exception,
this pathch catch the exception to avoid crash.

issue:  https://github.com/apache/incubator-doris/issues/3126

### DIFF
--- a/be/src/util/thrift_client.cpp
+++ b/be/src/util/thrift_client.cpp
@@ -78,14 +78,14 @@ Status ThriftClientImpl::open_with_retry(int num_tries, int wait_ms) {
 
 void ThriftClientImpl::close() {
     try {
-        if (_transport.get() != NULL && _transport->isOpen()) _transport->close();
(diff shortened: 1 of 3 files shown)


acelyc111:
The modification has referenced to Impala, https://github.com/apache/impala/blob/master/be/src/rpc/thrift-client.cc

be/src/util/thrift_client.cpp
```diff
@@ -65,12 +76,21 @@ Status ThriftClientImpl::open_with_retry(int num_tries, int wait_ms) {
     return status;
 }
 
-Status ThriftClientImpl::close() {
-    if (_transport->isOpen()) {
-        _transport->close();
+void ThriftClientImpl::close() {
+    try {
+        if (_transport.get() != NULL && _transport->isOpen()) _transport->close();
```
imay: prefer nullptr.

imay (approved):
LGTM

PR #3136, opened 2020-03-18
[tablet meta] Do some refactor on TabletMeta

acelyc111:
- remove some functions' return value which always return `OLAP_SUCCESS `
- optimize some loops

### DIFF
--- a/be/src/olap/snapshot_manager.cpp
+++ b/be/src/olap/snapshot_manager.cpp
@@ -558,9 +558,9 @@ OLAPStatus SnapshotManager::_convert_beta_rowsets_to_alpha(const TabletMetaShare
     }
     if (res == OLAP_SUCCESS && modified) {
         if (is_incremental) {
-            res = new_tablet_meta->revise_inc_rs_metas(new_rowset_metas);
+            new_tablet_meta->revise_inc_rs_metas(new_rowset_metas);
         } else {
-            res = new_tablet_meta->revise_rs_metas(new_rowset_metas);
+            new_tablet_meta->revise_rs_metas(new_rowset_metas);
         }
     }
     return res;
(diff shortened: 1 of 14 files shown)


be/src/olap/task/engine_storage_migration_task.cpp
```diff
@@ -234,21 +225,13 @@ OLAPStatus EngineStorageMigrationTask::_storage_medium_migrate(
...
-        LOG(WARNING) << "fail to generate new header for store is null";
-        return OLAP_ERR_HEADER_INIT_FAILED;
-    }
-    OLAPStatus res = tablet->generate_tablet_meta_copy(new_tablet_meta);
-    if (res != OLAP_SUCCESS) {
-        LOG(WARNING) << "could not generate new tablet meta. "
-                     << "tablet:" << tablet->full_name();
-        return res;
-    }
+    DCHECK(store);
```
imay:
```suggestion
    DCHECK(store != nullptr);
```
Prefer "var != nullptr" for pointer compare. Only use this when the bool value

be/src/olap/snapshot_manager.cpp
```diff
@@ -558,9 +541,9 @@ OLAPStatus SnapshotManager::_convert_beta_rowsets_to_alpha(const TabletMetaShare
     }
     if (res == OLAP_SUCCESS && modified) {
         if (is_incremental) {
-            res = new_tablet_meta->revise_inc_rs_metas(new_rowset_metas);
+            new_tablet_meta->revise_inc_rs_metas(new_rowset_metas);
```
imay: Seems that it will be better to make argument a rvalue reference.

be/src/olap/snapshot_manager.cpp
```diff
@@ -558,9 +541,9 @@ OLAPStatus SnapshotManager::_convert_beta_rowsets_to_alpha(const TabletMetaShare
...
         } else {
-            res = new_tablet_meta->revise_rs_metas(new_rowset_metas);
+            new_tablet_meta->revise_rs_metas(new_rowset_metas);
```
imay: same with above

imay (commented):
Nice job, it makes code cleaner

imay (approved):
LGTM

PR #3137, opened 2020-03-18
fix a daily test fail

yangzhg:
fix a bug of  const union query like  `select null union select null`, this because the type of SlotDescriptor  when claus is  `select null ` is null ,this will  cause be core dump, and fe find wrong cast function
this is the reason of  test fail  stream_test_insert.test_insert_union

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/CastExpr.java
+++ b/fe/src/main/java/org/apache/doris/analysis/CastExpr.java
@@ -103,12 +103,12 @@ public class CastExpr extends Expr {
                 if (fromType.isStringType() && toType.isBoolean()) {
                     continue;
                 }
-                // Disable casting from boolean to decimal
+                // Disable casting from boolean to decimal or datetime or date
                 if (fromType.isBoolean() &&
-                        (toType == Type.DECIMAL || toType == Type.DECIMALV2)) {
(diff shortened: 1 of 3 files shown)


fe/src/main/java/org/apache/doris/analysis/CastExpr.java
```diff
@@ -105,10 +105,10 @@ public static void initBuiltins(FunctionSet functionSet) {
                 }
                 // Disable casting from boolean to decimal
```
morningman: Change the comment

fe/src/main/java/org/apache/doris/analysis/SetOperationStmt.java
```diff
@@ -649,7 +649,8 @@ public void analyze(Analyzer parent) throws AnalysisException, UserException {
             if (isAnalyzed()) {
                 return;
             }
-            if (queryStmt instanceof SelectStmt && ((SelectStmt) queryStmt).fromClause_.isEmpty()) {
+            if (operation != Operation.UNION && queryStmt instanceof SelectStmt
```
morningman: Could you add comment to explain why you change here?

morningman (approved):

Issue #3139, opened 2020-03-18
[Bug] Python throw Commands out of sync; you can't run this command now


morningman:
**Describe the bug**
When using Python Mysqldb library to execute sql on Doris. Some time it will throw
error:
```
Commands out of sync; you can't run this command now
```

**To Reproduce**

1. Execute sql like: 

```
execute("insert into table tbl1 select * from tbl2;")
```

Notice that there is a `;` at the end of the statement.

2. If the `insert` statement has some warnings, then the error will throw.

knightvv:
I'm hitting this same problem still. Repro code:

`    conn = mysql.connector.connect(
        host=MYSQL_HOST,
        user=MYSQL_USER,
        password=MYSQL_PSWD,
        port=MYSQL_PORT
    )

    # Create a database
    cursor = conn.cursor()

    # For mysql, need to create db first
    cursor.execute(f"CREATE DATABASE IF NOT EXISTS {MYSQL_DB_NAME} DEFAULT CHARACTER SET 'utf8mb4';\
                   USE {MYSQL_DB_NAME};SHOW DATABASES;",
                   multi=True)
    
    # fetch results
    res = cursor.fetchall()
    conn.commit() # <--- throws MySQLInterfaceError: Commands out of sync; you can't run this command now`

PR #3140, opened 2020-03-18
[Syntax] Remove all EmptyStmt from the end of multi-statements list

morningman:
to resolve the ISSUE: #3139 

When user execute query by some client library such as python MysqlDb, if user execute like:

     "select * from tbl1;"  (with a comma at the end of statement)
     
The sql parser will produce 2 statements: `SelectStmt` and `EmptyStmt`.
Here we discard the `EmptyStmt` to make it act like one single statement.

This is for some compatibility. Because in python MysqlDb, if the first `SelectStmt` results in
some warnings, it will try to execute a `SHOW WARNINGS` statement right after the 
SelectStmt, but before the execution of `EmptyStmt`. So there will be an exception:

     `(2014, "Commands out of sync; you can't run this command now")`

I though it is a flaw of python MysqlDb.
However, in order to maintain the consistency of user use, here we remove all EmptyStmt
at the end to prevent errors.(Leave at least one statement)

But if user execute statements like:

     `"select * from tbl1;;select 2"`
     
If first `select * from tbl1` has warnings, python MysqlDb will still throw exception.

### DIFF
--- a/fe/src/main/java/org/apache/doris/common/util/SqlParserUtils.java
+++ b/fe/src/main/java/org/apache/doris/common/util/SqlParserUtils.java
@@ -17,6 +17,7 @@
 
 package org.apache.doris.common.util;
 
+import org.apache.doris.analysis.EmptyStmt;
 import org.apache.doris.analysis.SqlParser;
 import org.apache.doris.analysis.StatementBase;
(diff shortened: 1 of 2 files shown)


imay (approved):
LGTM

PR #3143, opened 2020-03-19
Non blocking OlapTableSink

vagetablechicken:
### NodeChannel
1. _cur_batch -> _pending_batches: when _cur_batch is filled up, move it to _pending_batches.
2. add_row() just produce batches.
3. try_send_and_fetch_status() tries to consume one pending batch. If has in flight packet, skip send in this round.

So we can add one sender thread to be in charge of all node channels try_send.

### IndexChannel
1. init(), open() stay the same.
2. Use for_each_node_channel() to expose the detailed changes of NodeChannel.(It's more easy to read & modify)

### Sender thread
See func OlapTableSink::_send_batch_process()
#### Why use polling？
If we use wait/notify, it will notify when generate a new batch. We can't skip sending this batch, coz it won't notify the same batch again. So wait/notify can't avoid blocking simply.
So I choose polling. 
It's wasting to continuously try_send(), but it's difficult to set the suitable polling interval. Thus, I add std::this_thread::yield() to give up the time slice, give priority to other process/threads (if there are other process/threads waiting in the queue).


imay:
Hi @vagetablechicken 

Do you have some benchmark for this PR?

vagetablechicken:
> Hi @vagetablechicken
> 
> Do you have some benchmark for this PR?

I only have two test samples now, testing environment building is in progress in our team.

I use broker load, test 5 times in each version, and the scan node wasn't do mem limit(my test code doesn't contain a80e9bf22968232dead8886d990993eb3b702424). If scan node waits when mem limit exceeded, it must take longer.

| test sample | origin ver(avg) | nonblocking ver(avg) |
|-|-|-|
| {"ScannedRows":92095001,"TaskNumber":1,"FileNumber":8,"FileSize":2798532731} | 10min3s | 6min43s |
| {"ScannedRows":951503983,"TaskNumber":1,"FileNumber":300,"FileSize":68560716133} | 4h25min  | 2h6min |

morningman:
Compilation failed:
LOG(DEBUG) << name() << " queue_push_ns: " << _queue_push_lock_ns;

There is no `LOG(DEBUG)` macro

vagetablechicken:
> Compilation failed:
> LOG(DEBUG) << name() << " queue_push_ns: " << _queue_push_lock_ns;
> 
> There is no `LOG(DEBUG)` macro

Oops, my bad. Fixed.

be/src/exec/tablet_sink.h
```diff
@@ -47,18 +48,101 @@ class ExprContext;
...
+    }
+    friend AddBatchCounter operator+(const AddBatchCounter& lhs, const AddBatchCounter& rhs) {
+        AddBatchCounter sum = lhs;
+        sum += rhs;
+        return sum;
+    }
+};
+
+template <typename T>
+class ReusableClosure : public google::protobuf::Closure {
```
imay: I think should add ref count for this closure. Because this closure is created by channel, and the RPC call has this pointer. However it is not sure when the RPC will call this->Run(). If this channel is destructed before the RPC's callback, it will make invalid memory access. Otherwise it should wait until the RPC finish, which is unnecessary.

be/src/exec/tablet_sink.cpp
```diff
@@ -128,55 +127,127 @@ Status NodeChannel::open_wait() {
     _open_closure = nullptr;
 
     // add batch closure
-    _add_batch_closure = new RefCountClosure<PTabletWriterAddBatchResult>();
-    _add_batch_closure->ref();
+    _add_batch_closure = ReusableClosure<PTabletWriterAddBatchResult>::create();
+    _add_batch_closure->addFailedHandler([&]() {
```
imay: I think it is better to declare all your capture other than a &. And what's more, referencing a local variable will lead to invalid memory access, because when the lambda is called, the reference is not valid.

be/src/exec/tablet_sink.cpp
```diff
@@ -186,80 +257,89 @@ void NodeChannel::cancel() {
...
 
-    _batch->reset();
-    return Status::OK();
+    return _send_finished ? 0 : 1;
 }
 
-IndexChannel::~IndexChannel() {
+Status NodeChannel::none_of(std::initializer_list<bool> vars) {
+    bool none = true;
+    std::string vars_str;
```
imay: I think most of the scenarios are still normal. better to construct strings when it is abnormal

imay (changes requested):

be/src/exec/tablet_sink.cpp
vagetablechicken: The two handlers both need to capture _rpc_error(the member variable), I'm afraid I can only change [&] to [this].
There's an alternate way to explicit capture, although I don't like this style. How about this?
```
auto& rpc_error_ref=_rpc_error;
_add_batch_closure->addFailedHandler([&rpc_error_ref](){})
```

imay: I think `[this]` is OK

be/src/exec/tablet_sink.h
vagetablechicken: Actually, I want the NodeChannel to the owner of add_batch_closure.

https://github.com/apache/incubator-doris/pull/3143/files#diff-4007834d7219c2282c8ae5c454f01dbaR47-R51
In NodeChannel's dtor `_add_batch_closure->release();`, release() will do brpc::Join(). So RPC won't call Run() after channel is destructed. 
If OlapTableSink is closed normally, `_add_batch_closure->release();` has no effect. 
But as you mentioned about unnecessary waiting RPC finished , if OlapTableSink is closed with error, `_add_batch_closure->release();` in NodeChannel dtor may take some time to wait add_batch repsonse, although the add_batch response is useless under the circumstance that we cancelled the whole sink.
I've considered about brpc::StartCancel(call_id), but I'm not good enough at understanding brpc, so I wrote a comment [here](https://github.com/apache/incubator-doris/pull/3143/files#diff-4007834d7219c2282c8ae5c454f01dbaR249). And just let the channel wait for the useless RPC response.

Maybe we should do brpc::StartCancel(call_id)?

imay: Yes, you are right.
It is OK to call release before delete it. But it should be commented carefully.

vagetablechicken: I think the func name `release()` is a little hard to understand, let me think about it, and I'll add more comments on the usage of ReusableClosure.

be/src/exec/tablet_sink.cpp
```diff
@@ -128,55 +129,128 @@ Status NodeChannel::open_wait() {
...
+                if (status.ok()) {
+                    if (is_last_rpc) {
+                        for (auto& tablet : result.tablet_vec()) {
+                            TTabletCommitInfo commit_info;
+                            commit_info.tabletId = tablet.tablet_id();
+                            commit_info.backendId = _node_id;
+                            _tablet_commit_infos.emplace_back(std::move(commit_info));
+                        }
+                        _add_batches_finished = true;
+                        LOG(INFO) << name() << " last rpc has responsed";
```
imay: Is this log needed?

be/src/exec/tablet_sink.cpp
```diff
@@ -877,5 +920,24 @@ int OlapTableSink::_validate_data(RuntimeState* state, RowBatch* batch, Bitmap*
...
+                running_channels_num += ch->try_send_and_fetch_status();
+            });
+        }
+
+        if (running_channels_num == 0) {
+            LOG(INFO) << "all node channels are stopped(maybe finished/offending/cancelled), "
+                         "consumer thread exit.";
+            return;
+        }
+        std::this_thread::yield();
```
imay: If there is no other workload, will this cause a busy-loop?

be/src/exec/tablet_sink.cpp
```diff
@@ -186,80 +260,90 @@ void NodeChannel::cancel() {
...
+            }
 
-    if (eos) {
-        for (auto pid : _parent->_partition_ids) {
-            _add_batch_request.add_partition_ids(pid);
+            // eos request must be the last request
+            _add_batch_closure->end_mark();
+            _send_finished = true;
+            DCHECK(_pending_batches_num == 0);
+            LOG(INFO) << name() << " send finished, should wait the last repsonse";
```
imay: should remove this trace log

be/src/exec/tablet_sink.cpp
```diff
@@ -611,57 +633,79 @@ Status OlapTableSink::close(RuntimeState* state, Status close_status) {
...
-                    LOG(WARNING) << "close channel failed, load_id=" << print_id(_load_id)
-                        << ", txn_id=" << _txn_id;
-                }
+            for (auto index_channel : _channels) {
+                index_channel->for_each_node_channel(
+                        [&](NodeChannel* ch) { WARN_IF_ERROR(ch->mark_close(), ""); });
+            }
+
+            for (auto index_channel : _channels) {
+                index_channel->for_each_node_channel([&](NodeChannel* ch) {
```
imay: change  to explicit capture

be/src/exec/tablet_sink.cpp
```diff
@@ -128,55 +129,128 @@ Status NodeChannel::open_wait() {
...
+
+    auto row_no = _cur_batch->add_row();
     if (row_no == RowBatch::INVALID_ROW_INDEX) {
-        RETURN_IF_ERROR(_send_cur_batch());
-        row_no = _batch->add_row();
+        {
+            SCOPED_RAW_TIMER(&_queue_push_lock_ns);
+            std::lock_guard<std::mutex> l(_pending_batches_lock);
+            //To simplify the add_row logic, postpone adding batch into req until the time of sending req
+            _pending_batches.emplace(std::move(_cur_batch), _cur_add_batch_request);
```
imay: we should limit the length of pending_batches.
When sender is blocked, this queue will consume memory without limit

vagetablechicken: As I commented here https://github.com/apache/incubator-doris/pull/3143/files#diff-4007834d7219c2282c8ae5c454f01dbaR67 , the scan node can do the memory limit. I should add more comments.

be/src/exec/tablet_sink.cpp
vagetablechicken: Yes, it's the easiest to implement, but not good enough. I've consider about writing an algorithm that can choose the property sleep interval, e.g. if every channel is empty, we can sleep for a while. bla bla bla
But taking the trouble to do this may lead to small gains.

be/src/exec/tablet_sink.cpp
vagetablechicken: _plan node in PlanFragmentExec may not limit the mem, so I add mem limit check in NodeChannel::add_row, it'll block the main thread if mem limit exceeded.

be/src/exec/tablet_sink.cpp
```diff
@@ -186,80 +267,89 @@ void NodeChannel::cancel() {
...
+        std::queue<AddBatchReq> empty;
+        std::swap(_pending_batches, empty);
+        _cur_batch.reset();
+    }
 }
 
-Status NodeChannel::_wait_in_flight_packet() {
-    if (!_has_in_flight_packet) {
-        return Status::OK();
+int NodeChannel::try_send_and_fetch_status() {
```
morningman: Why not using bool as return value?

morningman (dismissed):

morningman:
Hi @vagetablechicken , please resolve the conflict.
And Hi @imay , do you have any other comment?

be/src/exec/tablet_sink.cpp
vagetablechicken: I want the consumer to have global observation,  sum() is the simplest way to do it.
And maybe we can add more status(>0) to let consumer know more infos about channels. 
But I think the preliminary design should be simple.
So current design of status value is "0-channel stopped; 1-still working".

vagetablechicken:
I wanna change yeild() to sleep(), to avoid CPU busy. I need to prove "even sleep() after each loop, non-blocking sink is faster than blocking one" through testing.
So in the first step, I want the sleep interval to be the config param of BE, and I'll give a default value.

One more thing, the add_row() may be blocked cause mem limit exceeded, I'll add the blocking time to time profile.

be/src/exec/tablet_sink.cpp
vagetablechicken: change it to sleep(interval_time)

vagetablechicken:
@imay @morningman 
Sorry about my change is a bit late.
I've changed  yeild() to sleep(). And the default interval time is 10ms.
Why 10ms, it's calculated through the avg time of add batch RPCs.

If the case is large, it get faster obviously, so I focus on little cases this time.
I tested two little cases, and got the logs about add batch time. The log is generated by this piece of code.
https://github.com/apache/incubator-doris/blob/b4014d89d45e7644fe4d098eb0ac937553006bfd/be/src/exec/tablet_sink.cpp#L689-L698

So `avg time` = `add batch time` / `num`. And if you think the network time delay is not negligible, you can add it.
In my tests, the avg times are tens of milliseconds. So for the faster sink, I set it to 10ms.

And the test result(interval=10ms) is
| test sample | origin ver(avg) | nonblocking ver(avg) |
|-|-|-|
| {"ScannedRows":28582315,"TaskNumber":1,"FileNumber":3,"FileSize":569042681} | 4min37s | 3min27s |
|{"ScannedRows":100967318,"TaskNumber":1,"FileNumber":9,"FileSize":3096583144} | 12min27s | 8min27s |

be/src/exec/tablet_sink.h
```diff
@@ -68,99 +154,126 @@ class NodeChannel {
...
     void cancel();
 
-    int64_t node_id() const { return _node_id; }
+    // return:
+    // 0: stopped, send finished(eos request has been sent), or any internal error;
+    // 1: running, haven't reach eos.
+    // only allow 1 rpc in flight
+    int try_send_and_fetch_status();
+
+    void time_report(std::unordered_map<int64_t, AddBatchCounter>& add_batch_counter_map,
```
imay, line 171: If param will be modified, prefer pointer rather than reference.

be/src/exec/tablet_sink.h
```diff
@@ -68,99 +154,126 @@ class NodeChannel {
...
+        *actual_consume_ns += _actual_consume_ns;
+    }
 
-    void set_failed() { _already_failed = true; }
-    bool already_failed() const { return _already_failed; }
+    int64_t node_id() const { return _node_id; }
     const NodeInfo* node_info() const { return _node_info; }
+    std::string print_load_info() const { return _load_info; }
+    std::string name() const {
+        return "NodeChannel[" + std::to_string(_index_id) + "-" + std::to_string(_node_id) + "]";
```
imay, line 185: Prefer strings::Substitute in gutils/strings/Substitute.h.

be/src/exec/tablet_sink.cpp
```diff
@@ -128,54 +131,136 @@ Status NodeChannel::open_wait() {
...
-    _batch.reset();
-    return status;
+
+    // waiting for finished, it may take a long time, so we could't set a timeout
+    // use log to make it easier
+    LOG(INFO) << name() << "start close_wait";
+    while (!_add_batches_finished && !_cancelled) {
+        SleepFor(MonoDelta::FromMilliseconds(1));
+    }
+    LOG(INFO) << name() << "close_wait done";
```
imay, line 241: Do we need these log?

be/src/exec/tablet_sink.h
```diff
@@ -68,99 +154,126 @@ class NodeChannel {
...
     // this should be set in init() using config
     int _rpc_timeout_ms = 60000;
     int64_t _next_packet_seq = 0;
 
-    std::unique_ptr<RowBatch> _batch;
+    // user cancel or get some errors
+    std::atomic<bool> _cancelled{false};
+
+    std::atomic<bool> _send_finished{false};
+    std::atomic<bool> _add_batches_finished{false};
```
imay, line 208: what's the difference between these two flags?
Should comment to let us know

be/src/exec/tablet_sink.cpp
vagetablechicken: We just count the total time of channel's close(). There may be some slow BE hide inside. I think this logs are good for analyzing the state at that time. It's convenient for developers.
How about set it to vlog?

imay: If You want to, you can count the time with StopWatch and log the time in one log.
I suggest to use vlog.

be/src/exec/tablet_sink.h
vagetablechicken: including gutils/strings/Substitute.h will occur redefinition error
```
In file included from /root/incubator-doris/be/src/gutil/strings/stringpiece.h:128:0,
                 from /root/incubator-doris/be/src/gutil/strings/substitute.h:9,
                 from /root/incubator-doris/be/src/exec/tablet_sink.cpp:28:
/root/incubator-doris/be/src/gutil/hash/hash.h:251:26: error: redefinition of 'struct __gnu_cxx::hash<Type*>'
 template<class T> struct hash<T*> {
                          ^~~~~~~~
In file included from /root/incubator-doris/thirdparty/installed/include/butil/containers/flat_map.h:101:0,
                 from /root/incubator-doris/be/src/service/brpc.h:48,
                 from /root/incubator-doris/be/src/util/ref_count_closure.h:24,
                 from /root/incubator-doris/be/src/exec/tablet_sink.h:36,
                 from /root/incubator-doris/be/src/exec/tablet_sink.cpp:18:
/root/incubator-doris/thirdparty/installed/include/butil/containers/hash_tables.h:262:8: note: previous definition of 'struct __gnu_cxx::hash<Type*>'
 struct hash<Type*> {
        ^~~~~~~~~~~
In file included from /root/incubator-doris/be/src/gutil/strings/stringpiece.h:128:0,
                 from /root/incubator-doris/be/src/gutil/strings/substitute.h:9,
                 from /root/incubator-doris/be/src/exec/tablet_sink.cpp:28:
/root/incubator-doris/be/src/gutil/hash/hash.h:309:8: error: redefinition of 'struct __gnu_cxx::hash<std::pair<_T1, _T2> >'
 struct hash<pair<First, Second> > {
        ^~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /root/incubator-doris/thirdparty/installed/include/butil/containers/flat_map.h:101:0,
                 from /root/incubator-doris/be/src/service/brpc.h:48,
                 from /root/incubator-doris/be/src/util/ref_count_closure.h:24,
                 from /root/incubator-doris/be/src/exec/tablet_sink.h:36,
                 from /root/incubator-doris/be/src/exec/tablet_sink.cpp:18:
/root/incubator-doris/thirdparty/installed/include/butil/containers/hash_tables.h:256:8: note: previous definition of 'struct __gnu_cxx::hash<std::pair<_T1, _T2> >'
 struct hash<std::pair<Type1, Type2> > {
```

I will initialize the name_string in init(), to avoid string building in name()

be/src/exec/tablet_sink.cpp
vagetablechicken: Good advice, I'll fix it.

imay (approved):
+1, LGTM.
If there is no other comments, I will merge it tomorrow.

Issue #3145, opened 2020-03-19
[Bug] get null pointer exception when executing show alter table statement


morningman:
This PR #3130 will clear some structures of alter job after job is done.
But these structures may be used in `show alter table` statement. So
that they can not be cleared.

PR #3130, opened 2020-03-17
[Alter]Clear expire alterJobV2

WingsGo:
Too much AlterJobsV2 may consume too much memory, which may cause FullGC. Clear some data for finished or cancelled alterJobs and remove them when expire.

PR #3146, opened 2020-03-19
[Bug] Get NPE when executing show alter table statement

morningman:
This bug is introduced by this PR: #3130.
In #3130, it will clear some structures of alter job after job is done.
But these structures may be used in show alter table statement. So
that they can not be cleared.

In this CL, I removed the `clear()` method of alter job. And add a new edit log
`REMOVE_ALTER_JOB` to notify non-master FE that the expired alter job
has been removed.

Also add a new proc node for schema change job, for viewing the unfinished task of
a schema change job.

ISSUE: #3145

### DIFF
--- a/fe/src/main/java/org/apache/doris/persist/EditLog.java
+++ b/fe/src/main/java/org/apache/doris/persist/EditLog.java
@@ -728,6 +728,20 @@ public class EditLog {
                     catalog.replaySetReplicaStatus(log);
                     break;
                 }
+                case OperationType.OP_REMOVE_ALTER_JOB_V2: {
+                    RemoveAlterJobOperationLog log = (RemoveAlterJobOperationLog) journal.getData();
+                    switch (log.getType()) {
+                        case ROLLUP:
+                            catalog.getRollupHandler().replayRemoveAlterJobV2(log);
+                            break;
+                        case SCHEMA_CHANGE:
+                            catalog.getSchemaChangeHandler().replayRemoveAlterJobV2(log);
+                            break;
+                        default:
+                            break;
+                    }
+                    break;
+                }
                 default: {
                     IOException e = new IOException();
                     LOG.error("UNKNOWN Operation Type {}", opCode, e);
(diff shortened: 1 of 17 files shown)


fe/src/main/java/org/apache/doris/persist/EditLog.java
```diff
@@ -1257,4 +1271,8 @@ public void logReplaceTempPartition(ReplacePartitionOperationLog info) {
     public void logSetReplicaStatus(SetReplicaStatusOperationLog log) {
         logEdit(OperationType.OP_SET_REPLICA_STATUS, log);
     }
+
+    public void logRemoveExpiredAlterJob(RemoveAlterJobOperationLog log) {
```
WingsGo: public void logRemoveExpireAlterJobV2(RemoveAlterJobV2OperationLog log) to avoid mislead?

fe/src/main/java/org/apache/doris/persist/RemoveAlterJobOperationLog.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.google.gson.annotations.SerializedName;
+
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+
+
+// edit log from removing alter job v2
+public class RemoveAlterJobOperationLog implements Writable {
```
WingsGo: maybe `RemoveAlterJobV2OperationLog` is better

fe/src/main/java/org/apache/doris/persist/EditLog.java
morningman: OK

fe/src/main/java/org/apache/doris/persist/RemoveAlterJobOperationLog.java
morningman: OK

chaoyli (approved):

PR #3157, opened 2020-03-20
Fix unit test bug in BetaRowset and PageCacheTest

chaoyli:
1. BlockManager has been added into StorageEngine.
    So StorageEngine should be initialized when starting BetaRowset unit test.

2. Cache should not use the same buf to store value, otherwise the address
     will be freed twice and crash.


chaoyli:
#3165

morningman (approved):
LGTM

morningman (approved):
LGTM

Issue #3155, opened 2020-03-20
Datetime value is not valid, column=1000-01-01 00:00:00


HangyuanLiu:
Can we open the datetime limit？

morningman:
Does other dbms support this？

HangyuanLiu:
Yes, mysql or clickhouse support this date literal

PR #3158, opened 2020-03-20
Open datetime min value limit

HangyuanLiu:
For #3155 
the min_value in olap/type.h of datetime is 0000-01-01 00:00:00, so we don't need restrict datetime min in tablet_sink


morningman (commented):
Have you made some test about this?
And you need to update the CREATE TABLE help doc about the range of date and datetime type

chaoyli (approved):

chaoyli:
LGTM

Issue #3163, opened 2020-03-20
delete result is invalid when there is no predicate


kangpinghuang:
**Describe the bug**
the result is invalid when there is no predicate for count(*) agaist a table with delete.

**To Reproduce**

CREATE DATABASE test_sys_bloom_filter_c_test_delete_on_bf_column_db;

[...]

PR #3164, opened 2020-03-20
fix delete predicate bug for v2

kangpinghuang:
This bug is caused that the min and max wrapper field is not inited
when there is no predicate of that column.
Issue: #3163


morningman (approved):
LGTM

PR #3166, opened 2020-03-20
[CodeStyle] Rename new_partition_aggregation_node and new_partitioned_hash_table

HappenLee:


be/src/exec/partitioned_aggregation_node_ir.cc
```diff
@@ -15,9 +15,9 @@
 // specific language governing permissions and limitations
 // under the License.
 
-#include "exec/new_partitioned_aggregation_node.h"
+#include "exec/partitioned_aggregation_node.h"
 
-#include "exec/new_partitioned_hash_table.inline.h"
+#include "exec/partitioned_hash_table.inline.h"
```
chaoyli: #include "exec/partitioned_hash_table.inline.h" after #include "exprs/new_agg_fn_evaluator.h"

be/src/exec/partitioned_hash_table.h
```diff
@@ -42,7 +42,7 @@ class RowDescriptor;
 class RuntimeState;
 class Tuple;
 class TupleRow;
-class NewPartitionedHashTable;
+class PartitionedHashTable;
```
chaoyli: class PartitionedHashTable place before class RowDescriptor

chaoyli (changes requested):

chaoyli (approved):

Issue #3168, opened 2020-03-20
[Bug] Routine load task is paused when table is altering


morningman:
**Describe the bug**
First, create a routine load job, than begin to alter the table, the job may be pause
because following error:

```
Mapping column is not in table. column: __doris_shadow_k2
```

**How to resolve**
This is because before each task is executed, the `columnExprs` of the job
is modified without being restored. This can also cause memory leaks.

PR #3169, opened 2020-03-20
[RoutineLoad] Fix bug that job will be paused when table is altering

morningman:
Also add some debug log to observe the cost time of the process of routine load task
ISSUE: #3168

### DIFF
--- a/fe/src/main/java/org/apache/doris/load/Load.java
+++ b/fe/src/main/java/org/apache/doris/load/Load.java
@@ -894,18 +894,18 @@ public class Load {
 
         // If user does not specify the file field names, generate it by using base schema of table.
         // So that the following process can be unified
-        boolean specifyFileFieldNames = columnExprs.stream().anyMatch(p -> p.isColumn());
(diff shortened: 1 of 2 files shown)


fe/src/main/java/org/apache/doris/load/Load.java
```diff
@@ -888,6 +888,10 @@ public static void initColumns(Table tbl, List<ImportColumnDesc> columnExprs,
             }
         }
 
+        // We make a copy of the columnExprs so that our subsequent changes
+        // to the columnExprs will not affect the original columnExprs.
+        List<ImportColumnDesc> copiedColumnExprs = Lists.newArrayList(columnExprs);
```
imay: seems that `copiedColumnExprs` is not used by anyone

morningman: My bad, forgot to save the code...

EmmyMiao87 (approved):
LGTM

Issue #3170, opened 2020-03-20
[Bug] memory exceed limit when group by varchar columns


wyb:
4 backends, 32G memory per backend

Errors happen when executing two SQLs concurrently.

ERROR 1064 (HY000): errCode = 2, detailMessage = Memory exceed limit. New partitioned aggregation, while getting rows in streaming. Backend: 172.18.206.190, fragment: 5943ed36d4d24746-98837f1e5d257668 Used: 26422635638, Limit: 26416034611. You can change the limit by session variable exec_mem_limit.

Group by columns type:
dim1.c_name VARCHAR(300)
dim1.c_code VARCHAR(270)
dim_date.year_month_sk BIGINT
t1.type_desc  VARCHAR(50)
t1.type  VARCHAR(2)

SQL:
```sql
SELECT dim1.c_name, dim1.c_code, dim_date.year_month_sk, dim_date.year_month_sk, t1.type_desc
	, t1.type, SUM(t1.v1), SUM(t1.v2)
FROM t1
	JOIN dim1 ON dim1.code = t1.code
	LEFT JOIN dim_date ON dim_date.date_sid = t1.d_date
GROUP BY dim1.c_name, dim1.c_code, dim_date.year_month_sk, t1.type_desc, t1.type
ORDER BY 1 ASC
LIMIT 20
```

PR #3171, opened 2020-03-20
Fix group by varchar column bug

wyb:
remove varchar column deep copy in partitioned hash table EvalRow function
#3170


imay (approved):
LGTM

PR #3172, opened 2020-03-23
Change some config

gengjun-git:
1. Change word of palo to doris in conf file.
2. Set default meta_dir to ${DORIS_HOME}/doris-meta
2. Comment out be storage_root_path, leave it to ${DORIS_HOME}/storage, as exsting in be config.h.


conf/fe.conf
```diff
@@ -38,7 +38,7 @@ JAVA_OPTS_FOR_JDK_9="-Xmx4096m -XX:SurvivorRatio=8 -XX:MaxTenuringThreshold=7 -X
 sys_log_level = INFO
 
 # store metadata, create it if it is not exist.
-meta_dir = ${DORIS_HOME}/palo-meta
```
morningman: I am afraid this may cause some problem.
Some users may not notice this configuration and use the default `palo-meta` directory directly. When the cluster is upgraded, if it is replaced with the `doris-meta` directory, some problems may occur. This directory configuration recommendation remains unchanged.

morningman (changes requested):

conf/fe.conf
imay: Hi @morningman, for this change I don't agree with you.

1. configuration files will not be updated in normal case.
2. Even if he uses the new configure file, the FE will exit with error messages like "There is no meta directory". And this can be fixed by change this option to the correct directory.

In my opinion, code must keep compatible with old versions. However, configuration default value is unnecessary.

morningman: How about change this to 

```
# meta_dir = ${DORIS_HOME}/doris-meta
```

So that it is same as default value in `be.conf`, which is also commented out.

And we should also change this in FE's `config.java`

chaoyli:
If is necessary to change binary name from palo_be to doris_be, palo_fe to doris_fe ?

conf/fe.conf
imay: It works for me.
Comment for this configuration should be recorded in this file carefully. And related documents and tutorials should be changed too.

gengjun-git: OK. I will change meta_dir and storage_root_path commented out.

gengjun-git: OK. I will make meta_dir and storage_root_path commented out, and leave the default value changed to doris.

conf/be.conf
```diff
@@ -36,15 +36,16 @@ brpc_port = 8060
...
+# /home/disk2/doris.SSD, capacity limit is 1GB, SSD;
+# /home/disk2/doris, capacity limit is disk capacity, HDD(default)
 # 
 # you also can specify the properties by setting '<property>:<value>', seperate by ','
 # property 'medium' has a higher priority than the extension of path
-# storage_root_path = /home/disk1/palo,medium:ssd,capacity:50
-storage_root_path = /home/disk1/palo;/home/disk2/palo
+#
+# Default value is ${DORIS_HOME}/storage, you should create it by hand.
+# storage_root_path = ${DORIS_HOME}/storage
```
imay: be/src/common/conf.cpp should be changed too.

gengjun-git: The default value in conf.cpp is already ${DORIS_HOME}/storage.

imay (approved):
LGTM

morningman (approved):
LGTM

morningman:
> If is necessary to change binary name from palo_be to doris_be, palo_fe to doris_fe ?

Maybe in another PR.

PR #3175, opened 2020-03-23
Remove unused KUDU codes

chaoyli:


fe/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -3889,8 +3793,7 @@ public static void getDdlStmt(Table table, List<String> createTableStmt, List<St
 
         // 1.2 other table type
         sb.append("CREATE ");
-        if (table.getType() == TableType.KUDU || table.getType() == TableType.MYSQL
-                || table.getType() == TableType.ELASTICSEARCH) {
+        if (table.getType() == TableType.MYSQL || table.getType() == TableType.ELASTICSEARCH) {
```
morningman: Add table.getType() == TableType.BROKER

morningman (approved):
LGTM

Issue #3174, opened 2020-03-23
The error rollup has been selected when there is having clause in query


EmmyMiao87:
**Describe the bug**
If the aggregation function only appear in having clause such as following query, the mv selector will select a error rollup which only considers the columns that appear in the output.
···
Rollup only has column k1 will be selected in following query.
select k1 from table group by k1 having sum(v1);
···

**To Reproduce**
```
CREATE TABLE `test_tb` (

  `k1` int(11) NOT NULL COMMENT "",

  `k2` int(11) NOT NULL COMMENT "",

  `v1` varchar(4096) NOT NULL COMMENT "",

  `v2` float NOT NULL COMMENT "",

  `v3` decimal(20, 7) NOT NULL COMMENT ""

) ENGINE=OLAP

DUPLICATE KEY(`k1`, `k2`)

COMMENT "OLAP"

DISTRIBUTED BY HASH(`k1`, `k2`) BUCKETS 10

PROPERTIES (

"storage_type" = "COLUMN",

 "replication_num" = "3",

 "in_memory" = "false"

); 
```

create materialized view
```
create materialized view k1_k2 as select k1,k2 from test_tb group by k1,k2;\
```

load data
```
LOAD LABEL test_db.label_19 (DATA INFILE("hdfs://xxxx/user/palo/test/data/sys/schema_change/a_raw") INTO TABLE `test_tb`) WITH BROKER "hdfs" ("username"="root", "password"="xxx");
```

explain query
```
explain select k1 from test_tb group by k1 having max(v1) > 10;
+-----------------------------------------------------------------------------+
| Explain String                                                              |
+-----------------------------------------------------------------------------+
| PLAN FRAGMENT 0                                                             |
|  OUTPUT EXPRS:<slot 2>                                                      |
|   PARTITION: UNPARTITIONED                                                  |
|                                                                             |
|   RESULT SINK                                                               |
|                                                                             |
|   4:EXCHANGE                                                                |
|      tuple ids: 1                                                           |
|                                                                             |
| PLAN FRAGMENT 1                                                             |
|  OUTPUT EXPRS:                                                              |
|   PARTITION: HASH_PARTITIONED: <slot 2>                                     |
|                                                                             |
|   STREAM DATA SINK                                                          |
|     EXCHANGE ID: 04                                                         |
|     UNPARTITIONED                                                           |
|                                                                             |
|   3:AGGREGATE (merge finalize)                                              |
|   |  output: max(<slot 3>)                                                  |
[...]

PR #3176, opened 2020-03-23
Solve the problem of mv selector when there is having clause in query

EmmyMiao87:
All of columns which belong to top of tupleIds in query shoud be considered in mv selecotr.
For example:
Select k1 from table group by k1 having sum(v1) >1;
The candidate index should contain k1 and v1 columns instread of only k1.
The rollup which only has k1 column should not be selected.
The issue #3174 describe in detail.

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/TupleDescriptor.java
+++ b/fe/src/main/java/org/apache/doris/analysis/TupleDescriptor.java
@@ -261,7 +261,7 @@ public class TupleDescriptor {
 
     public void getTableNameToColumnNames(Map<String, Set<String>> tupleDescToColumnNames) {
         for (SlotDescriptor slotDescriptor : slots) {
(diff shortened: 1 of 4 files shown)


fe/src/main/java/org/apache/doris/analysis/TupleDescriptor.java
```diff
@@ -256,6 +259,31 @@ public void materializeSlots() {
         for (SlotDescriptor slot: slots) slot.setIsMaterialized(true);
     }
 
+    public void getTableNameToColumnNames(Map<String, Set<String>> tupleDescToColumnNames) {
+        for (SlotDescriptor slotDescriptor : slots) {
+            if (slotDescriptor.isMaterialized() == false) {
```
morningman:
```suggestion
            if (!slotDescriptor.isMaterialized()) {
```

morningman (approved):
LGTM

PR #3186, opened 2020-03-24
Support convert  Arrow data to RowBatch asynchronously in Spark-Doris-Connector

Youngwb:
Currently, in the Spark-Doris-Connector, when Spark iteratively obtains each row of data, it needs to synchronously convert the Arrow format data into the row format required by Spark. In order to speed up the conversion process, we can add an asynchronous thread in the Connector, which is responsible for obtaining the Arrow format data from BE and converting it into the row format required by Spark calculation

In our test environment, Doris cluster used 1 fe and 7 be (32C+128G). When using Spark-Doris-Connector to query a table containing 67 columns, the original query returned 69 million rows of data took about 2.5min, but after improvement, it reduced to about 1.6min, which reduced the time by about 30%

### DIFF
--- a/extension/spark-doris-connector/src/main/java/org/apache/doris/spark/serialization/RowBatch.java
+++ b/extension/spark-doris-connector/src/main/java/org/apache/doris/spark/serialization/RowBatch.java
@@ -112,8 +112,9 @@ public class RowBatch {
             }
         } catch (Exception e) {
             logger.error("Read Doris Data failed because: ", e);
-            close();
             throw new DorisException(e.getMessage());
+        } finally {
+            close();
         }
     }
 
--- a/extension/spark-doris-connector/src/main/java/org/apache/doris/spark/serialization/RowBatch.java
+++ b/extension/spark-doris-connector/src/main/java/org/apache/doris/spark/serialization/RowBatch.java
@@ -70,7 +70,8 @@ public class RowBatch {
         }
     }
 
-    private int offsetInOneBatch = 0;
+    // offset for iterate the rowBatch
+    private int offsetInRowBatch = 0;
     private int rowCountInOneBatch = 0;
     private int readRowCount = 0;
     private List<Row> rowBatch = new ArrayList<>();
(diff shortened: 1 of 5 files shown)


wuyunfeng:
Great! @vinson0526  Can you give a review?

extension/spark-doris-connector/src/main/java/org/apache/doris/spark/serialization/RowBatch.java
```diff
@@ -70,7 +70,8 @@ public void put(Object o) {
         }
     }
 
-    private int offsetInOneBatch = 0;
+    // offset for iterate the rowBatch
+    private int offsetInRowBatch = 0;
     private int rowCountInOneBatch = 0;
     private int readRowCount = 0;
     private List<Row> rowBatch = new ArrayList<>();
```
wuyunfeng: whether we can use a fixed array which size = rowCount?

Youngwb: We could get the fixed array size by add all rows of Arrow batch through `arrowStreamReader.loadNextBatch()`, in this case, we might as well use List as we do now

extension/spark-doris-connector/src/main/java/org/apache/doris/spark/serialization/RowBatch.java
```diff
@@ -87,50 +88,39 @@ public RowBatch(TScanBatchResult nextResult, Schema schema) throws DorisExceptio
...
+                    logger.debug("One batch in arrow has no data.");
+                    continue;
+                }
+                rowCountInOneBatch = root.getRowCount();
+                // init the rowBatch
+                for (int i = 0; i < rowCountInOneBatch; ++i) {
+                    rowBatch.add(new Row(fieldVectors.size()));
+                }
+                convertArrowToRowBatch();
+                readRowCount += root.getRowCount();
```
vinson0526: function close could be removed and move its content here, since read all data from arrow here.

extension/spark-doris-connector/src/main/java/org/apache/doris/spark/serialization/RowBatch.java
```diff
@@ -70,7 +70,8 @@ public void put(Object o) {
         }
     }
 
-    private int offsetInOneBatch = 0;
+    // offset for iterate the rowBatch
+    private int offsetInRowBatch = 0;
     private int rowCountInOneBatch = 0;
```
vinson0526: rowCountInOneBatch could be a local variable？

Youngwb: rowCountInOneBatch is also used in `addValueToRow` and `convertArrowToRowBatch`, so I think this is better for now

Youngwb: rowCountInOneBatch is also used in `addValueToRow` and `convertArrowToRowBatch`, so I think this is better for now

extension/spark-doris-connector/src/main/java/org/apache/doris/spark/serialization/RowBatch.java
Youngwb: OK,Done

wuyunfeng (approved):
LGTM. @vinson0526  please give a check

vinson0526 (approved):
LGTM

morningman (approved):
LGTM

Issue #3187, opened 2020-03-24
joinHints not work when table reordered


yangzhg:
**Describe the bug**
in hash join clasue， when tables needed to be reorderd, the join hint will not work

**To Reproduce**
1.  two table in tpcds data set `catalog_sales` and `store_sales`, `catalog_sales` is larger than `store_sales`
2. run query 
```
explain select count(*) from (select distinct store_sales.ss_addr_sk  from catalog_sales inner join [shuffle] store_sales  on catalog_sales.cs_bill_cdemo_sk = store_sales.ss_addr_sk) a;
```

**Expected behavior**
expected to be hash join is `INNER JOIN (PARTITIONED)` actrual is `INNER JOIN (BROADCAST)`

**Screenshots**
```
mysql> explain select count(*) from (select distinct store_sales.ss_addr_sk  from catalog_sales inner join [shuffle] store_sales  on catalog_sales.cs_bill_cdemo_sk = store_sales.ss_addr_sk) a;
+---------------------------------------------------------------------------------+
| Explain String                                                                  |
+---------------------------------------------------------------------------------+
| PLAN FRAGMENT 0                                                                 |
|  OUTPUT EXPRS:<slot 4>                                                          |
|   PARTITION: UNPARTITIONED                                                      |
|                                                                                 |
|   RESULT SINK                                                                   |
|                                                                                 |
|   9:AGGREGATE (merge finalize)                                                  |
|   |  output: count(<slot 4>)                                                    |
|   |  group by:                                                                  |
|   |  tuple ids: 4                                                               |
|   |                                                                             |
|   8:EXCHANGE                                                                    |
|      tuple ids: 4                                                               |
|                                                                                 |
| PLAN FRAGMENT 1                                                                 |
|  OUTPUT EXPRS:                                                                  |
|   PARTITION: HASH_PARTITIONED: <slot 2>                                         |
|                                                                                 |
|   STREAM DATA SINK                                                              |
|     EXCHANGE ID: 08                                                             |
|     UNPARTITIONED                                                               |
|                                                                                 |
|   4:AGGREGATE (update serialize)                                                |
|   |  output: count(*)                                                           |
|   |  group by:                                                                  |
|   |  tuple ids: 4                                                               |
|   |                                                                             |
|   7:AGGREGATE (merge finalize)                                                  |
|   |  group by: <slot 2>                                                         |
|   |  tuple ids: 2                                                               |
|   |                                                                             |
[...]

PR #3188, opened 2020-03-24
fix join hints not work when need table reorder

yangzhg:
1 fix join hints not work when need table reorder, this bug is described in #3187
2 fix cross join node does not implement `computeStats`, this will  cause brodcast join cost always 0 #3195

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
+++ b/fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
@@ -610,15 +610,12 @@ public class SelectStmt extends QueryStmt {
         }
     }
 
-    // When a join statement with a join hint, the decorated part should be reordered as a whole,
-    // rather than individually.
     protected void reorderTable(Analyzer analyzer) throws AnalysisException {
-        List<Pair<List<TableRef>, Long>> candidates = Lists.newArrayList();
+        List<Pair<TableRef, Long>> candidates = Lists.newArrayList();
 
-        for (int i = 0; i < fromClause_.size(); ++i) {
-            List<TableRef> tableRefs = new ArrayList<>();
-            TableRef tblRef = fromClause_.get(i);
-            if (tblRef.getJoinOp() != JoinOperator.INNER_JOIN) {
+        // New pair of table ref and row count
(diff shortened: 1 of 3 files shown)


fe/src/main/java/org/apache/doris/planner/CrossJoinNode.java
```diff
@@ -59,16 +59,16 @@ public TableRef getInnerRef() {
...
+         super.computeStats(analyzer);
+         if (getChild(0).cardinality == -1 || getChild(1).cardinality == -1) {
+           cardinality = -1;
+         } else {
+           cardinality = getChild(0).cardinality * getChild(1).cardinality;
+           if (computeSelectivity() != -1) {
+             cardinality = Math.round(((double) cardinality) * computeSelectivity());
+           }
+         }
+         LOG.debug("stats CrossJoin: cardinality=" + Long.toString(cardinality));
```
morningman:
```suggestion
         LOG.debug("stats CrossJoin: cardinality= {}", Long.toString(cardinality));
```

fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
```diff
@@ -612,11 +611,14 @@ public void materializeRequiredSlots(Analyzer analyzer) throws AnalysisException
...
-        List<Pair<TableRef, Long>> candidates = Lists.newArrayList();
+        List<Pair<List<TableRef>, Long>> candidates = Lists.newArrayList();
 
-        // New pair of table ref and row count
-        for (TableRef tblRef : fromClause_) {
+        for (int i = 0; i < fromClause_.size(); ++i) {
+            List<TableRef> tableRefs = new ArrayList<>();
+            TableRef tblRef = fromClause_.get(i);
             if (tblRef.getJoinOp() != JoinOperator.INNER_JOIN) {
                 // Unsupported reorder outer join
```
wutiangan, line 623: How about cross join?

EmmyMiao87 (dismissed):
LGTM

EmmyMiao87 (approved):
LGTM

Issue #3190, opened 2020-03-25
[Optimize] Simplify the delete process


xy720:
Our current DELETE strategy reuses the LoadChecker framework. The LoadChecker poll all jobs to run them and then update their state in every 5 seconds. But actually,  the DELETE jobs do not need to wait for polling, they can run the **pushTask()** function directly to delete data, and save a lot of time.

PR #3191, opened 2020-03-25
[Optimize][Delete] Simplify the delete process to make it fast

xy720:
#3190
Our current DELETE strategy reuses the LoadChecker framework. Loadchecker runs jobs in different stages by polling them in every 5 seconds. 

There are four stages of a load job, Pending/ETL/Loading/Quorum_finish, each of them is allocated to a LoadChecker. Four example, if a load job is submitted, it will be initialized to the pending state, then wait for running by the Pending LoadChecker. After the pending job is runned, its stage will change to ETL stage, and then wait for running by the next LoadChecker(ETL). Because interval time of the LoadChecker is 5s, in worst case, a pending job need to wait for 20s during its life cycle.

In particular, the DELETE jobs do not need to wait for polling, they can run the pushTask() function directly to delete. In this commit, I add a delete handler to concurrently processing delete tasks. All delete tasks will push to BE immediately, not required to wait for LoadCheker, without waiting for 2 LoadChecker(delete job started in LOADING state), at most 10s will be save(5s per LoadCheker). The delete process now is synchronized and users get response only after the delete finished or be canceled. If a delete is running over a certain period of time, it will be cancelled with a timeout exception.

当前Doris的删除策略是走的LoadChecker路线，Delete Job 被视为一般的Load Job，而一般的Load Job需要经过一次LoadChecker的轮询才能改变自己的状态。

Load Job的状态变化顺序为PENDING->ETL->LOADING->QUORUM_FINISHED，因为LoadCheker的轮询间隔为5秒，因此最坏情况下Load Job需要等待20s时间。

但是Delete Job一初始化即为LOADING状态，甚至它根本不需要状态，直接提交Push Task后等待就可以了，因此这次提交旨在将Delete Job从LoadChecker的系统中分离出来，节省不需要浪费的时间。

最好情况下，让Delete Job直接push task，可以节省5～10s时间，在提交任务后，根据此次提交涉及的tablet数量指定等待超时时间，这样同步删除的方式使一个Delete Job被取消或者完成时用户能够立刻收到消息，最后我们只要持久化已完成的Delete Job，供用户show delete命令查找即可。


kangpinghuang:
could you add some performance optimization result description?

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
```diff
@@ -0,0 +1,684 @@
...
+            String label = "delete_" + UUID.randomUUID();
+
+            // begin txn here and generate txn id
+            transactionId = Catalog.getCurrentGlobalTransactionMgr().beginTransaction(db.getId(),
+                    Lists.newArrayList(table.getId()), label,"FE: " + FrontendOptions.getLocalHostAddress(),
+                    TransactionState.LoadJobSourceType.FRONTEND, Config.stream_load_default_timeout_second);
+
+            TransactionState state = Catalog.getCurrentGlobalTransactionMgr()
+                    .getTransactionState(transactionId);
+            if (state == null) {
```
morningman, line 192: If begin transaction failed, exception will be thrown. So no need to check here.

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
```diff
@@ -0,0 +1,684 @@
...
+            deleteInfo = new DeleteInfo(db.getId(), olapTable.getId(), tableName,
+                    partition.getId(), partitionName,
+                    -1, 0, deleteConditions);
+
+            // task in fe
+            deleteTask = new DeleteTask(transactionId, deleteInfo);
+
+            writeLock();
+            try {
+                idToDeleteTask.put(transactionId, deleteTask);
```
morningman, line 205: Use a concurrent map is enough

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
```diff
@@ -0,0 +1,684 @@
...
+                    // the inverted index contains rolling up replica
+                    Long tabletId = invertedIndex.getTabletIdByReplica(replica.getId());
+                    if (tabletId == null) {
+                        LOG.warn("could not find tablet id for replica {}, the tablet maybe dropped", replica);
+                        continue;
+                    }
+                    tabletCommitInfos.add(new TabletCommitInfo(tabletId, replica.getBackendId()));
+                }
+            }
+            globalTransactionMgr.commitTransaction(db.getId(), transactionId, tabletCommitInfos);
```
morningman, line 388: use `commitAndPublishTransaction()` instead, and then `afterCommit()` can be removed.

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
```diff
@@ -0,0 +1,684 @@
...
+            }
+        } finally {
+            db.writeUnlock();
+        }
+    }
+
+    // for delete handler, we only persist those delete already finished.
+    @Override
+    public void write(DataOutput out) throws IOException {
+        out.writeInt(dbToDeleteInfos.size());
```
morningman, line 660: Use GSON serde

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
```diff
@@ -0,0 +1,684 @@
...
+                        long signature = task.getSignature();
+                        if (task.isCancel()) {
+                            break;
+                        }
+                        if (!executor.submit(task)) {
+                            Thread.sleep(1000);
+                            continue;
+                        }
+                        // re add to the tail
+                        queue.put(task);
```
morningman, line 358: this logic is wrong.

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
```diff
@@ -0,0 +1,684 @@
...
+                DeleteInfo deleteInfo = task.getDeleteInfo();
+                long dbId = deleteInfo.getDbId();
+                if (dbToDeleteInfos.containsKey(dbId)) {
+                    dbToDeleteInfos.get(dbId).add(deleteInfo);
+                } else {
+                    List<DeleteInfo> deleteInfoList = Lists.newArrayList();
+                    deleteInfoList.add(deleteInfo);
+                    dbToDeleteInfos.put(dbId, deleteInfoList);
+                }
+                Catalog.getInstance().getEditLog().logFinishSyncDelete(deleteInfo);
```
morningman, line 421: You can not write log here, because the transaction may commit fail

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
xy720: I've deleted it

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
xy720: done

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
xy720: done

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
xy720: There is problem that using GSON read() have to create a new object, but I wish to keep the old object and only use a readField() method.

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
xy720: fixed

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
xy720: I will put it in afterVisible() in interface AbstractTxnStateChangeCallback instead.

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
```diff
@@ -0,0 +1,549 @@
...
+                    if (tabletId == null) {
+                        LOG.warn("could not find tablet id for replica {}, the tablet maybe dropped", replica);
+                        continue;
+                    }
+                    tabletCommitInfos.add(new TabletCommitInfo(tabletId, replica.getBackendId()));
+                }
+            }
+            boolean isSuccess = globalTransactionMgr.commitAndPublishTransaction(db, transactionId, tabletCommitInfos, timeout);
+            if (!isSuccess) {
+                cancelJob(job, "delete timeout when waiting transaction commit");
```
morningman, line 281: `!isSuccess` does not mean the transaction is failed, it means the transaction is committed and publish timeout (not VISIBLE yet). So you should not cancel the job, but should tell the user the delete is success but will be taking effect later.

morningman (changes requested):

fe/src/main/java/org/apache/doris/persist/EditLog.java
```diff
@@ -378,8 +379,8 @@ public static void loadJournal(Catalog catalog, JournalEntity journal) {
                     break;
                 case OperationType.OP_FINISH_SYNC_DELETE: {
                     DeleteInfo info = (DeleteInfo) journal.getData();
-                    Load load = catalog.getLoadInstance();
-                    load.replayDelete(info, catalog);
+                    DeleteHandler deleteHandler = catalog.getDeleteHandler();
```
morningman, line 382: You can not use the origin `OP_FINISH_SYNC_DELETE`, because they are different operations.

fe/src/main/java/org/apache/doris/qe/ShowExecutor.java
```diff
@@ -1053,8 +1054,8 @@ private void handleShowDelete() throws AnalysisException {
         }
         long dbId = db.getId();
 
-        Load load = catalog.getLoadInstance();
-        List<List<Comparable>> deleteInfos = load.getDeleteInfosByDb(dbId, true);
+        DeleteHandler deleteHandler = catalog.getDeleteHandler();
```
morningman, line 1057: You should also show the delete info from `Load`, or after we upgrade the Doris, the old delete info can not be seen.

fe/src/main/java/org/apache/doris/task/DeleteJob.java
```diff
@@ -0,0 +1,185 @@
...
+    private DeleteState state;
+
+    // jobId(listenerId). use in beginTransaction to callback function
+    private long id;
+    // transaction id.
+    private long signature;
+    private Set<Long> totalTablets;
+    private Set<Long> quorumTablets;
+    private Set<Long> finishedTablets;
+    Map<Long, TabletDeleteInfo> tabletDeleteInfoMap;
```
morningman, line 57: There will be concurrent issue when visiting `tabletDeleteInfoMap` from different thread: the report task thread and DeleteHandler thread. So you should use ConcurrentHashMap

fe/src/main/java/org/apache/doris/task/DeleteJob.java
```diff
@@ -0,0 +1,185 @@
...
+    public boolean addTablet(long tabletId) {
+        return totalTablets.add(tabletId);
+    }
+
+    public boolean addPushTask(PushTask pushTask) {
+        return pushTasks.add(pushTask);
+    }
+
+    public boolean addFinishedReplica(long tabletId, Replica replica) {
+        TabletDeleteInfo tDeleteInfo = tabletDeleteInfoMap.get(tabletId);
```
morningman, line 136: After changing `tabletDeleteInfoMap` to ConncurrentHashMap, you should use `putIfAbsent` to perform the atomic operation here.

morningman (changes requested):

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
```diff
@@ -0,0 +1,627 @@
...
+            List<DeleteInfo> deleteInfoList = dbToDeleteInfos.get(dbId);
+            if (deleteInfoList == null) {
+                deleteInfoList = Lists.newArrayList();
+                dbToDeleteInfos.put(dbId, deleteInfoList);
+            }
+            deleteInfoList.add(job.getDeleteInfo());
+        }
+    }
+
+    public boolean cancelJob(DeleteJob job, CancelType cancelType, String reason) {
```
morningman, line 387: This method return boolean, but you never use it.
I think return true means cancel succeed(txn failed), and return false means cancel failed(txn succeed).
And the caller should use this return value to decide whether to return user success or failure.

morningman: And if transaction is COMMITTED but not VISIBLE, you should return a transaction id to user, so that user can use that it to check transaction's state.

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
```diff
@@ -0,0 +1,627 @@
...
+            sortIndex = 2;
+        }
+        ListComparator<List<Comparable>> comparator = new ListComparator<List<Comparable>>(sortIndex);
+        Collections.sort(infos, comparator);
+        return infos;
+    }
+
+    public void replayDelete(DeleteInfo deleteInfo, Catalog catalog) {
+        Database db = catalog.getDb(deleteInfo.getDbId());
+        db.writeLock();
```
morningman, line 585: No need db lock here.

morningman (changes requested):

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
xy720: Done

fe/src/main/java/org/apache/doris/persist/EditLog.java
xy720: Now add a new operation OP_FINISH_DELETE

fe/src/main/java/org/apache/doris/qe/ShowExecutor.java
xy720: done

fe/src/main/java/org/apache/doris/task/DeleteJob.java
xy720: done

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
xy720: removed it

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
```diff
@@ -0,0 +1,549 @@
...
+            // begin txn here and generate txn id
+            transactionId = Catalog.getCurrentGlobalTransactionMgr().beginTransaction(db.getId(),
+                    Lists.newArrayList(table.getId()), label,"FE: " + FrontendOptions.getLocalHostAddress(),
+                    TransactionState.LoadJobSourceType.FRONTEND, Config.stream_load_default_timeout_second);
+
+            deleteInfo = new DeleteInfo(db.getId(), olapTable.getId(), tableName,
+                    partition.getId(), partitionName,
+                    -1, 0, deleteConditions);
+            deleteJob = new DeleteJob(transactionId, deleteInfo);
+            idToDeleteJob.put(deleteJob.getTransactionId(), deleteJob);
```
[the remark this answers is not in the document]
xy720: added a new try finally block surround outside this logic

fe/src/main/java/org/apache/doris/task/DeleteJob.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    public boolean addPushTask(PushTask pushTask) {
+        return pushTasks.add(pushTask);
+    }
+
+    public boolean addFinishedReplica(long tabletId, Replica replica) {
+        tabletDeleteInfoMap.putIfAbsent(tabletId, new TabletDeleteInfo(tabletId));
+        TabletDeleteInfo tDeleteInfo =  tabletDeleteInfoMap.get(tabletId);
+        synchronized (tDeleteInfo) {
```
morningman, line 140: No need to use `synchronized` here, I think you can just use a `ConcurrentSet` in `TabletDeleteInfo`

morningman (approved):
LGTM

PR #3196, opened 2020-03-25
Support determine isPreviousLoadFinished for some alter jobs in table level 

caiconghui:
This PR is to reduce the time cost for waiting transactions to be completed in same db by filter the running transactions in table level.

### DIFF
--- a/fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
+++ b/fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
@@ -803,21 +803,16 @@ public class GlobalTransactionMgr implements Writable {
     
     // check if there exists a load job before the endTransactionId have all finished
     public boolean isPreviousTransactionsFinished(long endTransactionId, long dbId, List<Long> tableIdList) {
-        readLock();
-        try {
-            for (Map.Entry<Long, TransactionState> entry : idToTransactionState.entrySet()) {
-                if (entry.getValue().getDbId() != dbId || !isIntersectionNotEmpty(entry.getValue().getTableIdList(),
-                        tableIdList) || !entry.getValue().isRunning()) {
-                    continue;
-                }
-                if (entry.getKey() <= endTransactionId) {
-                    LOG.debug("find a running txn with txn_id={} on db: {}, less than watermark txn_id {}",
-                            entry.getKey(), dbId, endTransactionId);
-                    return false;
-                }
+        for (Map.Entry<Long, TransactionState> entry : idToTransactionState.entrySet()) {
+            if (entry.getValue().getDbId() != dbId || !isIntersectionNotEmpty(entry.getValue().getTableIdList(),
+                    tableIdList) || !entry.getValue().isRunning()) {
+                continue;
             }
-        } finally {
-            readUnlock();
+            if (entry.getKey() <= endTransactionId) {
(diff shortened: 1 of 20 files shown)


caiconghui:
for #3197

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -821,6 +821,24 @@ public boolean isPreviousTransactionsFinished(long endTransactionId, long dbId)
...
+
+    // check if there exists a intersection between the source tableId list and target tableId list
+    // if one of them is null or empty, that means that we don't know related tables in tableList,
+    // we think the two lists may have intersection for right ordered txns
+    public boolean isIntersectionNotEmpty(List<Long> sourceTableIdList, List<Long> targetTableIdList) {
+        if (sourceTableIdList == null || sourceTableIdList.isEmpty() || targetTableIdList == null ||
+                targetTableIdList.isEmpty()) {
+            return true;
+        }
+        for (int i = 0; i < sourceTableIdList.size(); i++) {
```
WingsGo: the method is low efficiency, use `sourceTableIdList.retainAll(targetTableIdList);sourceTableIdList.isEmpty()` instread?

caiconghui: > the method is low efficiency, use `sourceTableIdList.retainAll(targetTableIdList);sourceTableIdList.isEmpty()` instread?

in most case the table list size is 1, and we should not modify the table list content

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -802,25 +802,38 @@ public void finishTransaction(long transactionId, Set<Long> errorReplicaIds) thr
...
+            for (int j = 0; j < targetTableIdList.size(); j++) {
+                if (sourceTableIdList.get(i).longValue() == targetTableIdList.get(j).longValue()) {
```
kangkaisen: Why not use equals？

fe/src/main/java/org/apache/doris/transaction/TransactionState.java
```diff
@@ -564,5 +583,13 @@ public void readFields(DataInput in) throws IOException {
             callbackId = in.readLong();
             timeoutMs = in.readLong();
         }
+
+        if (Catalog.getCurrentCatalogJournalVersion() >= FeMetaVersion.VERSION_79) {
+            tableIdList = new ArrayList<>(1);
```
morningman: Why init with size 1?

fe/src/main/java/org/apache/doris/transaction/TransactionState.java
```diff
@@ -126,6 +129,7 @@ public String toString() {
     }
 
     private long dbId;
+    private List<Long> tableIdList;
```
morningman: How about using a HashSet to speed up the find?

caiconghui: actually, in most case ,the table list size is 1, and use list instead of set ,we can reduce memory usage if there are many trasactions with little performance loss

morningman (approved):
LGTM

Issue #3200, opened 2020-03-25
[Bug] Temp partition and dynamic partition have conflicts


morningman:
**Describe the bug**
1. create a table.
2. add a temp partition to this table.
3. modify the table to enable the dynamic partition feature.
4. when trying to drop the temp partition, encounter error: 
    `can not drop partition when dynamic partition is enable`
5. when trying to disable the dynamic partition, encounter error:
    `can not modify table when there is temp partitions`

This logic is a dead lock. We should fix it.

PR #3201, opened 2020-03-25
[Bug][Refactor] Fix the conflict of temp partition and dynamic partition operations

morningman:
The bug is described in issue: #3200.

This CL solve the problem by:
1. Refactor the alter operation conflict checking logic by introducing new classes `AlterOperations` and `AlterOpType`.
2. Allow add/drop temporary partition when dynamic partition feature is enabled.
3. Allow modifying table's property when there is temporary partition in table.
4. Make the properties `dynamic_partition.enable` optional, and default is true.

### DIFF
--- a/fe/src/main/java/org/apache/doris/alter/AlterOpType.java
+++ b/fe/src/main/java/org/apache/doris/alter/AlterOpType.java
@@ -1,0 +1,62 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package org.apache.doris.alter;
+
+public enum AlterOpType {
+    // rollup
+    ADD_ROLLUP,
+    DROP_ROLLUP,
+    // schema change
+    SCHEMA_CHANGE,
+    // partition
+    ADD_PARTITION,
+    DROP_PARTITION,
+    REPLACE_PARTITION,
+    MODIFY_PARTITION,
+    // rename
+    RENAME,
+    // table property
+    MODIFY_TABLE_PROPERTY,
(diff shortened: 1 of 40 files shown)


fe/src/main/java/org/apache/doris/alter/AlterOpType.java
```diff
@@ -0,0 +1,62 @@
...
+    // table property
+    MODIFY_TABLE_PROPERTY,
+    MODIFY_TABLE_PROPERTY_SYNC, // Some operations are performed synchronously, so we distinguish them by suffix _SYNC
+    // others operation, such as add/drop backend. currently we do not care about them
+    ALTER_OTHER,
+
+    INVALID_OP;
+
+    // true means 2 operations have no conflict.
+    public static Boolean[][] CONFLICT_MATRIX;
```
kangkaisen: If true means 2 operations have no conflict. which should be `compatibilityMatrix`.

fe/src/main/java/org/apache/doris/alter/AlterOperations.java
```diff
@@ -0,0 +1,104 @@
...
+
+    public boolean hasRollupOp() {
+        return currentOps.contains(AlterOpType.ADD_ROLLUP) || currentOps.contains(AlterOpType.DROP_ROLLUP);
+    }
+
+    public boolean hasRenameOp() {
+        return currentOps.contains(AlterOpType.RENAME);
+    }
+
+    // return true if it has specified operation
```
kangkaisen: This comment could remove.

fe/src/main/java/org/apache/doris/analysis/AddPartitionClause.java
```diff
@@ -46,12 +47,11 @@ public AddPartitionClause(SingleRangePartitionDesc partitionDesc,
                               DistributionDesc distributionDesc,
                               Map<String, String> properties,
                               boolean isTempPartition) {
+        super(AlterOpType.ADD_PARTITION);
         this.partitionDesc = partitionDesc;
         this.distributionDesc = distributionDesc;
         this.properties = properties;
         this.isTempPartition = isTempPartition;
-        
-        this.needTableStable = false;
```
kangkaisen: Why remove this code？

fe/src/main/java/org/apache/doris/alter/AlterOpType.java
morningman: OK

fe/src/main/java/org/apache/doris/alter/AlterOperations.java
morningman: OK

fe/src/main/java/org/apache/doris/analysis/AddPartitionClause.java
morningman: Add it back

kangkaisen (approved):
+1. LGTM

PR #3202, opened 2020-03-25
[Plugin] Create the FE plugin dir if missing

morningman:
The FE plugin dir should be created when initializing.
Also modify the pom.xml in `fe_plugins` dir to make it able to use custom maven setting.


Seaven (approved):

kangkaisen (approved):
+1. LGTM

PR #3206, opened 2020-03-26
Rewritten subquery in having clause

### LINKED ISSUE #3205: The 'between and' in having subquery is not rewritten.
**Describe the bug**
query: 

```
select k1, sum (k2) from all_type_table group by k1 having sum(k2) > (select k1 from all_type_table where k2 between 1 and 2);
ERROR 1064 (HY000): BetweenPredicate needs to be rewritten into a CompoundPredicate.
``

TPC-DS 14 throws the same error.

**To Reproduce**
1. create table 
```
CREATE TABLE `all_type_table` (
  `k1` tinyint(4) NULL COMMENT "",
  `k2` smallint(6) NULL COMMENT "",
  `k3` int(11) NULL COMMENT "",
  `k4` bigint(20) NULL COMMENT "",
  `k5` largeint(40) NULL COMMENT "",
  `k6` float NULL COMMENT "",
  `k7` double NULL COMMENT ""
) ENGINE=OLAP
DUPLICATE KEY(`k1`)
COMMENT "OLAP"
DISTRIBUTED BY HASH(`k1`) BUCKETS 3
PROPERTIES (
"storage_type" = "COLUMN",
 "replication_num" = "1",
 "in_memory" = "false"
);
```

2. query
```
select k1, sum (k2) from all_type_table group by k1 having sum(k2) > (select k1 from all_type_table where k2 between 1 and 2);
```


**Expected behavior**

subquery in having clause should be rewritten too. 
TPC-DS 14 should be passed.

EmmyMiao87:
The subquery in having clause should be rewritten too.
If not, ExprRewriteRule will not be apply in subquery.
For example:
select k1, sum (k2) from table group by k1 having sum(k2) > (select t1 from table2 where t1 between 1 and 2);
```t1 between 1 and 2``` should be rewritten to ```t1 >=1 and t1<=2```.

Fixed #3205. TPC-DS 14 will be passed after this commit.

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
+++ b/fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
@@ -1228,17 +1228,19 @@ public class SelectStmt extends QueryStmt {
         for (TableRef ref : fromClause_) {
             ref.rewriteExprs(rewriter, analyzer);
         }
+        // Also rewrite exprs in the statements of subqueries.
+        List<Subquery> subqueryExprs = Lists.newArrayList();
         if (whereClause != null) {
             whereClause = rewriter.rewrite(whereClause, analyzer);
-            // Also rewrite exprs in the statements of subqueries.
-            List<Subquery> subqueryExprs = Lists.newArrayList();
             whereClause.collect(Subquery.class, subqueryExprs);
-            for (Subquery s : subqueryExprs) {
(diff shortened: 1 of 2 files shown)


yangzhg (commented):
LGTM

kangkaisen (approved):
+1， lGTM

PR #3209, opened 2020-03-26
Add 'IndexKeysType' field in 'Desc all table stmt'

EmmyMiao87:
After doris support aggregation materialized view on duplicate table, desc stmt of metadata is confused in sometimes.
The reason is that there is no grouping information in desc stmt of metadata.

For example:
There are two materialized view as following.
    1. create materialized view k1_k2 as select k1, k2 from table;
    2. create materialzied view deduplicated_k1_k2 as select k1, k2 from table group by k1, k2;
Before this commit, the metatdata in desc stmt is the same.

   ```
    +-----------------------+-------+----------+------+-------+---------+-------+
    | IndexName             | Field | Type     | Null | Key   | Default | Extra |
    +-----------------------+-------+----------+------+-------+---------+-------+
    | k1_k2                 | k1    | TINYINT  | Yes  | true  | N/A     |       |
    |                       | k2    | SMALLINT | Yes  | true  | N/A     |       |
    | deduplicated_k1_k2    | k1    | TINYINT  | Yes  | true  | N/A     |       |
    |                       | k2    | SMALLINT | Yes  | true  | N/A     |       |
    +-----------------------+-------+----------+------+-------+---------+-------+
   ```

So, we need to show the KeysType of materialized view in desc stmt.
Now, the desc stmt of all mvs is changed as following:

    ```
    +-----------------------+---------------+-------+----------+------+-------+---------+-------+
    | IndexName             | IndexKeysType | Field | Type     | Null | Key   | Default | Extra |
    +-----------------------+---------------+-------+----------+------+-------+---------+-------+
    | k1_k2                 | DUP_KEYS      | k1    | TINYINT  | Yes  | true  | N/A     |       |
    |                       |               | k2    | SMALLINT | Yes  | true  | N/A     |       |
    | deduplicated_k1_k2    | AGG_KEYS      | k1    | TINYINT  | Yes  | true  | N/A     |       |
    |                       |               | k2    | SMALLINT | Yes  | true  | N/A     |       |
    +-----------------------+---------------+-------+----------+------+-------+---------+-------+
    ```

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/DescribeStmt.java
+++ b/fe/src/main/java/org/apache/doris/analysis/DescribeStmt.java
@@ -20,6 +20,7 @@ package org.apache.doris.analysis;
 import org.apache.doris.catalog.Catalog;
 import org.apache.doris.catalog.Column;
 import org.apache.doris.catalog.Database;
+import org.apache.doris.catalog.MaterializedIndexMeta;
 import org.apache.doris.catalog.MysqlTable;
 import org.apache.doris.catalog.OlapTable;
 import org.apache.doris.catalog.ScalarType;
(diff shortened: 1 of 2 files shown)


imay (dismissed):
LGTM

kangkaisen:
I think only show IndexKeysType isn't enough，we need also show the materialized view sql.
@EmmyMiao87 Hi, Currently, do we have a way to get the materialized view  sql?

EmmyMiao87:
> I think only show IndexKeysType isn't enough，we need also show the materialized view sql.
> @EmmyMiao87 Hi, Currently, do we have a way to get the materialized view sql?

Actually, according to keystype, iskey, column, and Extra we could infer the sql of materialized view. But this display sql really better than desc stmt.
Until now, we do not have a way to get the materialized view sql.

morningman (approved):
LGTM

Issue #3132, opened 2020-03-17
[Alter]AlterJobs in big table will use to much memory, which may cause FullGC


WingsGo:
The old alterJob has clear logic to avoid consume too much memory, but I found that alterJobV2 don't, In our scenario, This problem may cause FullGC frequently. In the screenshot, schemaChangeJob consume more than half memory.

WingsGo:
![Snip20200317_46](https://user-images.githubusercontent.com/22497357/76835830-c2aa8a80-686a-11ea-8aa9-cf2171d8be18.png)

PR #3212, opened 2020-03-27
[Alter]Clean SchemaChangeJobV2 when schema change CANCELLED or FINISHED

WingsGo:
SchemaChangeJobV2 will use too much memory in FE, which may casuse FullGC. But these data is useless after job is done, so we need to clean it up.


WingsGo:
For #3132

fe/src/main/java/org/apache/doris/alter/SchemaChangeJobV2.java
```diff
@@ -953,4 +969,131 @@ public void readFields(DataInput in) throws IOException {
             }
         }
     }
+
+    /**
+     * write data need to persist when job finished
+     */
+    private void writeJobFinishedData(DataOutput out) throws IOException {
+        // only persist data will be used in getInfo
+        out.writeInt(indexIdMap.size());
```
kangkaisen: Write double?

fe/src/main/java/org/apache/doris/alter/SchemaChangeJobV2.java
```diff
@@ -953,4 +969,131 @@ public void readFields(DataInput in) throws IOException {
...
+            for (int i = 0; i < bfNum; i++) {
+                bfColumns.add(Text.readString(in));
+            }
+            bfFpp = in.readDouble();
+        }
+
+        watershedTxnId = in.readLong();
+
+        // index
+        if (Catalog.getCurrentCatalogJournalVersion() >= FeMetaVersion.VERSION_70) {
```
kangkaisen: Don't need this check.

fe/src/main/java/org/apache/doris/alter/SchemaChangeJobV2.java
```diff
@@ -95,6 +95,9 @@
     // shadow index id -> shadow index short key count
     private Map<Long, Short> indexShortKeyMap = Maps.newHashMap();
 
+    // identify whether the job is finished and no need to persist some data
+    private boolean cleared = false;
```
kangkaisen: Please rename. Such as `isMetaPruned `

fe/src/main/java/org/apache/doris/alter/SchemaChangeJobV2.java
```diff
@@ -159,6 +161,18 @@ public void setStorageFormat(TStorageFormat storageFormat) {
         this.storageFormat = storageFormat;
     }
 
+    /**
+     * clear some date structure in this job to save memory
+     * these data structures must not used in getInfo method
+     */
+    public void clear() {
```
kangkaisen: Please rename. Such as `pruneMeta`

fe/src/main/java/org/apache/doris/alter/SchemaChangeJobV2.java
WingsGo: why use writeDouble? indexIdMap.size() is a integer type.

fe/src/main/java/org/apache/doris/alter/SchemaChangeJobV2.java
WingsGo: done

fe/src/main/java/org/apache/doris/alter/SchemaChangeJobV2.java
WingsGo: I just use the origin code, why we don't need this check?

fe/src/main/java/org/apache/doris/alter/SchemaChangeJobV2.java
kangkaisen: Ignore me. I yesterday looked the code write `out.writeInt(indexIdMap.size());` twice. So I said write double.

fe/src/main/java/org/apache/doris/alter/SchemaChangeJobV2.java
kangkaisen: We only call `readJobFinishedData` If `Catalog.getCurrentCatalogJournalVersion() >= FeMetaVersion.VERSION_80` is true.

WingsGo: done

kangkaisen (approved):
+1， LGTM

morningman (approved):
LGTM

Issue #3213, opened 2020-03-27
[Bug] Cast expr in predicate result in wrong result


HangyuanLiu:
![image](https://user-images.githubusercontent.com/12147090/77729052-becff280-7038-11ea-9cba-b1c24533629d.png)

The result of "datetime > '2020-03-25'" and "datetime > cast ('2020-03-25' as date)" is different

PR #3214, opened 2020-03-27
Fix cast date type return wrong result

HangyuanLiu:
For #3213 
We have multiple date type, and we also need to cast between different date types.
 If not cast, it will cause problems when binarypredicate

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/DateLiteral.java
+++ b/fe/src/main/java/org/apache/doris/analysis/DateLiteral.java
@@ -319,7 +319,14 @@ public class DateLiteral extends LiteralExpr {
 
     @Override
     protected Expr uncheckedCastTo(Type targetType) throws AnalysisException {
-        if (targetType.equals(this.type)) {
+        if (targetType.isDateType()) {
+            if (targetType.equals(Type.DATE)) {
+                this.castToDate();                            
+            } else if (targetType.equals(Type.DATETIME)) {
+                this.type = Type.DATETIME;                            
+            } else {
+                throw new AnalysisException("Error date literal type : " + type);
+            }
             return this;
         } else if (targetType.isStringType()) {
             return new StringLiteral(getStringValue());
--- a/fe/src/main/java/org/apache/doris/analysis/DateLiteral.java
+++ b/fe/src/main/java/org/apache/doris/analysis/DateLiteral.java
@@ -319,7 +319,7 @@ public class DateLiteral extends LiteralExpr {
 
     @Override
     protected Expr uncheckedCastTo(Type targetType) throws AnalysisException {
-        if (targetType.isDateType()) {
+        if (targetType.equals(this.type)) {
             return this;
         } else if (targetType.isStringType()) {
             return new StringLiteral(getStringValue());
--- a/fe/src/test/java/org/apache/doris/analysis/DateLiteralTest.java
+++ b/fe/src/test/java/org/apache/doris/analysis/DateLiteralTest.java
@@ -55,4 +55,23 @@ public class DateLiteralTest {
         }
         Assert.assertFalse(hasException);
     }
+
+    @Test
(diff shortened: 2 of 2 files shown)


fe/src/main/java/org/apache/doris/analysis/DateLiteral.java
```diff
@@ -319,7 +319,7 @@ protected void toThrift(TExprNode msg) {
 
     @Override
     protected Expr uncheckedCastTo(Type targetType) throws AnalysisException {
-        if (targetType.isDateType()) {
+        if (targetType.equals(this.type)) {
```
kangkaisen: Are you sure only change this point is enough？
Do we need to change this point:
FunctionCallExpr::analyzeImpl
```
        Type[] args = fn.getArgs();
        if (args.length > 0) {
            // Implicitly cast all the children to match the function if necessary
            for (int i = 0; i < argTypes.length; ++i) {
                // For varargs, we must compare with the last type in callArgs.argTypes.
                int ix = Math.min(args.length - 1, i);
                if (!argTypes[i].matchesType(args[ix]) && !(
                        argTypes[i].isDateType() && args[ix].isDateType())) {
                    uncheckedCastChild(args[ix], i);
                    //if (argTypes[i] != args[ix]) castChild(args[ix], i);
                }
            }
        }
```

fe/src/test/java/org/apache/doris/analysis/DateLiteralTest.java
```diff
@@ -55,4 +55,23 @@ public void TwoDigitYear() {
         }
         Assert.assertFalse(hasException);
     }
+
+    @Test
+    public void uncheckedCastTo() {
```
kangkaisen: Would better also add a test to `QueryPlanTest`.

kangkaisen (dismissed):
+1. LGTM

fe/src/main/java/org/apache/doris/analysis/DateLiteral.java
imay: Why not return a new DateLiteral when targetType and this.type are different but both dateType.

imay (changes requested):

fe/src/main/java/org/apache/doris/analysis/DateLiteral.java
```diff
@@ -319,7 +319,17 @@ protected void toThrift(TExprNode msg) {
 
     @Override
     protected Expr uncheckedCastTo(Type targetType) throws AnalysisException {
-        if (targetType.equals(this.type)) {
+        if (targetType.isDateType()) {
```
kangkaisen, line 322: If isDateType is true. The type could only be DATE and DATETIME.

So one `if  else` if enough?

fe/src/main/java/org/apache/doris/analysis/DateLiteral.java
```diff
@@ -320,6 +320,13 @@ protected void toThrift(TExprNode msg) {
     @Override
     protected Expr uncheckedCastTo(Type targetType) throws AnalysisException {
         if (targetType.isDateType()) {
+            if (targetType.equals(Type.DATE)) {
+                this.castToDate();                            
+            } else if (targetType.equals(Type.DATETIME)) {
+                this.type = Type.DATETIME;
```
chaoyli, line 326: If Date not to convert DateTime, is OK?

chaoyli (changes requested):

fe/src/main/java/org/apache/doris/analysis/DateLiteral.java
HangyuanLiu: Yes, is ok。And explicit type conver(cast expr) also go to this line

chaoyli (approved):

imay (approved):
LGTM

Issue #3210, opened 2020-03-26
Exception about "Filesystem closed" in broker load，any help will be appreciated


xiaoyeblue:
1. Load task failed msg:
   'type:LOAD_RUN_FAIL; msg:Open broker reader failed, broker:TNetworkAddress(hostname=****, 
    port=8000) failed:could not open file hdfs://****/part-00*.orc, cause by: Filesystem closed'

2. Code:
    LOAD LABEL ***.****_202003241209 (
          DATA INFILE("hdfs://path/*") 
          INTO TABLE ***
          FORMAT AS ORC(***)
          SET (***)
          WITH BROKER hdfs_broker (***)
          PROPERTIES ()

3. Notes: 
    a. Task failed several hours after "Loading" status
    b. Failed file blocks are not fixed
    c. The exception disappears when switching to s3a
 
4. Any questions can be added and any help will be appreciated !

PR #3216, opened 2020-03-27
fix 'Filesystem closed' in broker load

frwrdt:
fix bug in issue #3210

### DIFF
--- a/fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
+++ b/fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
@@ -84,12 +84,15 @@ public class FileSystemManager {
     private static final String DEFAULT_DFS_CLIENT_FAILOVER_PROXY_PROVIDER =
             "org.apache.hadoop.hdfs.server.namenode.ha.ConfiguredFailoverProxyProvider";
     private static final String FS_DEFAULTFS_KEY = "fs.defaultFS";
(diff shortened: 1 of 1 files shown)


fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
```diff
@@ -383,6 +384,7 @@ public BrokerFileSystem getS3AFileSystem(String path, Map<String, String> proper
                 conf.set(FS_S3A_ACCESS_KEY, accessKey);
                 conf.set(FS_S3A_SECRET_KEY, secretKey);
                 conf.set(FS_S3A_ENDPOINT, endpoint);
+                conf.set("fs.s3a.impl.disable.cache", "true");
```
kangkaisen: Please use constant value. Add add a comment to explain why need this config.

frwrdt: If this property is not set to "true", FileSystem instance will be returned from cache which is not thread-safe and may cause 'Filesystem closed' exception when it is closed by other thread.

imay: @frwrdt 
You should add this comment in the source file to let others known not just the reviewer.

frwrdt: ok

morningman (approved):
LGTM

PR #3218, opened 2020-03-27
Fix WARNING to WARN in fe.conf sys_log_level

stalary:
When I used it, I changed it to WARING in the comments, and the log didn't work because there was no warning-level log in Java


kangkaisen (approved):
+1.

PR #3222, opened 2020-03-28
Support sharding txn_map_lock into more small map locks to make good performance for txn manage task

caiconghui:
This PR is to enhance the peformance for txn manage task, when there are so many txn in doris be, the only one txn_map_lock and additional _txn_locks may cause poor performance, and now we remove the additional _txn_locks and split the txn_map_lock into many small locks.

### DIFF
--- a/be/src/olap/txn_manager.cpp
+++ b/be/src/olap/txn_manager.cpp
@@ -70,10 +70,12 @@ using std::vector;
 
 namespace doris {
 
-TxnManager::TxnManager() {
-    for (int i = 0; i < _txn_lock_num; ++i) {
-        _txn_locks[i] = std::make_shared<RWMutex>();
-    }
+TxnManager::TxnManager(int32_t txn_map_shard_size)
+        : _txn_map_shard_size(txn_map_shard_size) {
+    DCHECK_GT(_txn_map_shard_size, 0);
+    _txn_map_locks = new RWMutex[_txn_map_shard_size];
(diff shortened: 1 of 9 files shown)


caiconghui:
for #3223

be/src/olap/txn_manager.cpp
```diff
@@ -253,11 +260,11 @@ OLAPStatus TxnManager::publish_txn(OlapMeta* meta, TPartitionId partition_id, TT
     pair<int64_t, int64_t> key(partition_id, transaction_id);
     TabletInfo tablet_info(tablet_id, schema_hash, tablet_uid);
     RowsetSharedPtr rowset_ptr = nullptr;
-    WriteLock wrlock(_get_txn_lock(transaction_id));
```
morningman: I think you may not be able to remove this lock. 
`WriteLock wrlock(_get_txn_lock(transaction_id));`

The meaning of this lock is different from that of the `_txn_map_lock`. This lock prevents a single transaction from being modified at the same time. The `_txn_map_lock` prevents the map structure from being modified at the same time.

If this lock is removed here. Then there may be two threads publishing the txn at the same time, which will modify their contents(such as rowset meta) at the same time.

morningman (changes requested):

chaoyli (approved):

morningman (approved):
LGTM

Issue #3227, opened 2020-03-30
output results may  incorrect  when using intersect and except statements


yangzhg:
**Describe the bug**
The output tuple of intersect and except is one tuple. When the number of input tuples of the intersect and except nodes is greater than one, the intersect and except nodes do not properly merge the input data.

**To Reproduce**
using tpcds query 14  or query like
```
            SELECT iss.i_brand_id brand_id ,
                  iss.i_class_id class_id ,
                  iss.i_category_id category_id
            FROM store_sales ,item iss ,date_dim d1
            WHERE ss_item_sk = iss.i_item_sk
                  AND ss_sold_date_sk = d1.d_date_sk
                  AND d1.d_year
                  BETWEEN 1998
                  AND 1998 + 2

            INTERSECT

            SELECT ics.i_brand_id ,
                  ics.i_class_id ,
                  ics.i_category_id
            FROM catalog_sales ,item ics ,date_dim d2
            WHERE cs_item_sk = ics.i_item_sk
                  AND cs_sold_date_sk = d2.d_date_sk
                  AND d2.d_year
                  BETWEEN 1998
                  AND 1998 + 2
```
you will see the result is incorrect， but the reult count is correct

PR #3228, opened 2020-03-30
Fix output results may  incorrect  when using intersect and except statements

yangzhg:
fix the bug desctibed in #3227 
add class SetOperationNode and move common funtions from intersect and except to it

### DIFF
--- a/be/src/exec/except_node.cpp
+++ b/be/src/exec/except_node.cpp
@@ -141,24 +141,17 @@ Status ExceptNode::get_next(RuntimeState* state, RowBatch* out_batch, bool* eos)
     uint8_t* tuple_buf;
     RETURN_IF_ERROR(
             out_batch->resize_and_allocate_tuple_buffer(state, &tuple_buf_size, &tuple_buf));
-    uint32_t previous_hash = -1;
-    TupleRow* previous_row = nullptr;
+    memset(tuple_buf, 0, tuple_buf_size);
     while (_hash_tbl_iterator.has_next()) {
         VLOG_ROW << "find row: "
                  << get_row_output_string(_hash_tbl_iterator.get_row(), child(0)->row_desc())
                  << " matched: " << _hash_tbl_iterator.matched();
         if (!_hash_tbl_iterator.matched()) {
-            if (previous_hash != _hash_tbl_iterator.get_hash() ||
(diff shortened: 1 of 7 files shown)


be/src/exec/except_node.cpp
```diff
@@ -179,17 +137,21 @@ Status ExceptNode::get_next(RuntimeState* state, RowBatch* out_batch, bool* eos)
...
+            out_batch->resize_and_allocate_tuple_buffer(state, &tuple_buf_size, &tuple_buf));
     uint32_t previous_hash = -1;
     TupleRow* previous_row = nullptr;
     while (_hash_tbl_iterator.has_next()) {
+        VLOG_ROW << "find row: "
+                 << get_row_output_string(_hash_tbl_iterator.get_row(), child(0)->row_desc())
+                 << " matched: " << _hash_tbl_iterator.matched();
         if (!_hash_tbl_iterator.matched()) {
             if (previous_hash != _hash_tbl_iterator.get_hash() ||
                 !equals(previous_row, _hash_tbl_iterator.get_row())) {
```
wutiangan: why does hashtable has the same value？When you create a hash table, will duplicate values are inserted into the hash table？

be/src/exec/except_node.cpp
```diff
@@ -112,6 +63,8 @@ Status ExceptNode::open(RuntimeState* state) {
         RETURN_IF_LIMIT_EXCEEDED(state, " Except, while constructing the hash table.");
         // build hash table and remvoe duplicate items
```
morningman:
```suggestion
        // build hash table and remove duplicate items
```

be/src/exec/except_node.cpp
morningman: As I know, this hash table will store all rows with same hash key.

wutiangan: But this is a set operation（It is different from join）. The set operation has deduplicated, so the elements of the hash table do not need to have duplicatited value。

yangzhg: duplicated values will be iterate when get next,and deduplicate can reduce the cost of probe,you can find detail in hash_table.h,hash table in doris is not same with the common sense hashtable

morningman (approved):
LGTM

PR #3232, opened 2020-03-30
Rewrite count distinct bitmap and hll in order by and having

kangkaisen:
For https://github.com/apache/incubator-doris/issues/3231

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/QueryStmt.java
+++ b/fe/src/main/java/org/apache/doris/analysis/QueryStmt.java
@@ -18,6 +18,7 @@
 package org.apache.doris.analysis;
 
 import org.apache.doris.catalog.Database;
+import org.apache.doris.catalog.FunctionSet;
 import org.apache.doris.common.AnalysisException;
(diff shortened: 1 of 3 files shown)


fe/src/main/java/org/apache/doris/analysis/FunctionCallExpr.java
```diff
@@ -262,6 +262,11 @@ public boolean isCountDistinctBitmapOrHLL() {
             return false;
         }
 
+        // if expr isn't analyzed, the type is INVALID
+        if (!getChild(0).isAnalyzed) {
+            getChild(0).analyzeNoThrow(analyzer);
```
imay, line 267: I'm not sure if it is OK to analyze it here.
Is it possible to call this function after analyzed?

kangkaisen: I also considered this problem，But I think wrong before.

Done.

imay (approved):
LGTM

Issue #3234, opened 2020-03-31
HDFS BROKER LOAD username configuration is not effective


ShowMeYourHammar:
**Describe the bug**
While using hdfs broker to load hdfs files with username  configuration, hdfs broker always uses system environment variable HADOOP_USER_NAME as hadoop user to read hdfs files. Otherwise if HADOOP_USER_NAME is not set, it will use the linux user launchs the hdfs broker service.

And the load script is like this.

LOAD LABEL db.xx (
  DATA INFILE(
    "xx"
  ) INTO TABLE `xx`
  COLUMNS TERMINATED BY "\\x01"
) WITH BROKER hdfs_broker (
  "hadoop.security.authentication" = "simple",
  "username" = "hadoop_user_name",
  "password" = "hadoop_password",
  "dfs.nameservices" = "nameservices",
  "dfs.ha.namenodes.nameservices" = "nn1,nn2",
  "dfs.namenode.rpc-address.nameservices.nn1" = "host1:port2",
  "dfs.namenode.rpc-address.nameservices.nn2" = "host2:port2"
) PROPERTIES (
  "exec_mem_limit" = "4294967296"
);

PR #3237, opened 2020-04-01
fix the bug that 'username' in broker load is invalid 

frwrdt:
fix bug in issue #3234

### DIFF
--- a/fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
+++ b/fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
@@ -285,9 +285,6 @@ public class FileSystemManager {
                                     e.getMessage());
                         }
                     }
-                } else if (!authentication.equals(AUTHENTICATION_SIMPLE)) {
-                    throw  new BrokerException(TBrokerOperationStatusCode.INVALID_ARGUMENT,
-                            "invalid authentication.");
                 }
                 if (!Strings.isNullOrEmpty(dfsNameServices)) {
                     // ha hdfs arguments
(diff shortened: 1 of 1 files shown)


fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
```diff
@@ -287,9 +285,6 @@ public BrokerFileSystem getDistributedFileSystem(String path, Map<String, String
                                     e.getMessage());
                         }
                     }
-                } else {
-                    throw  new BrokerException(TBrokerOperationStatusCode.INVALID_ARGUMENT,
-                            "invalid authentication.");
```
HangyuanLiu, line 292: We need to determine whether the authentication method exists

fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
```diff
@@ -338,7 +333,20 @@ public BrokerFileSystem getDistributedFileSystem(String path, Map<String, String
                 }
 
                 conf.set(FS_HDFS_IMPL_DISABLE_CACHE, "true");
-                FileSystem dfsFileSystem = FileSystem.get(pathUri.getUri(), conf);
+                FileSystem dfsFileSystem = null;
```
HangyuanLiu, line 336: Is this method compatible with  versions of hdfs?

HangyuanLiu:
Please change to a understandable title

fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
frwrdt: The authentication has been checked before, so this is  redundant here.

fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
frwrdt: The version 2.7.3 of package _hadoop-common_ including this method is the same as package _hadoop-hdfs_.

HangyuanLiu (approved):

PR #3242, opened 2020-04-01
Fix bug of memory limit when group by varchar columns.

chaoyli:
select date_format(k10, '%Y%m%d') as myk10 from baseall group by myk10;
The date_format function in query above will be stored in MemPool during
the query execution. If the query handles millions of rows, it will
consume much memory. Should clear the MemPool at interval.


chaoyli:
#3170

kangkaisen (approved):
+1. LGTM

imay (approved):
LGTM

PR #3246, opened 2020-04-02
[metrics] Add some metrics for container size in BE

acelyc111:
We can observe the workload of BE, and also it's a way to check
whether there is any problem in BE, like some container increase
too large and lead to OOM.

This patch add the following metrics:
```
Name                                   Description
rowset_count_generated_and_in_use      The total count of rowset id generated and in use since BE last start
unused_rowsets_count                   The total count of unused rowset waiting to be GC
broker_count                           The total count of brokers in management
data_stream_receiver_count             The total count of data stream receivers in management
fragment_endpoint_count                The total count of fragment endpoints of data stream in management, should always equal to data_stream_receiver_count
active_scan_context_count              The total count of active scan contexts
plan_fragment_count                    The total count of plan fragments in executing
load_channel_count                     The total count of load channels in management
result_buffer_block_count              The total count of result buffer blocks for queries, each block has a limited queue size (default 1024)
result_block_queue_count               The total count of queues for fragments, each queue has a limited size (default 20, by config::max_memory_sink_batch_count)
routine_load_task_count                The total count of routine load tasks in executing
small_file_cache_count                 The total count of cached small files' digest info
stream_load_pipe_count                 The total count of stream load pipes, each pipe has a limited buffer size (default 1M)
tablet_writer_count                    The total count of tablet writers
brpc_endpoint_stub_count               The total count of brpc endpoints
```


be/src/runtime/etl_job_mgr.cpp
```diff
@@ -62,6 +63,21 @@ const std::string ERROR_FILE_PREFIX = "error_log";
 
 EtlJobMgr::EtlJobMgr(ExecEnv* exec_env) :
         _exec_env(exec_env), _success_jobs(5000), _failed_jobs(5000) {
+    REGISTER_PRIVATE_VARIABLE_METRIC(running_etl_job_count);
```
morningman, line 66: `EtlJobMgr` is deprecated, it will be removed in next version.
So you don't need to add metrics for it.

be/src/runtime/result_buffer_mgr.cpp
```diff
@@ -33,6 +34,11 @@ namespace doris {
 
 ResultBufferMgr::ResultBufferMgr()
     : _is_stop(false) {
+    REGISTER_PRIVATE_VARIABLE_METRIC(result_buffer_block_count);
+    DorisMetrics::metrics()->register_hook("result_buffer_block_count", [&]() {
+        boost::lock_guard<boost::mutex> l(_lock);
```
morningman, line 39:
```suggestion
        std::lock_guard<boost::mutex> l(_lock);
```

chaoyli:
I have a question about dynamic collect the metric.
If people can remember so many metrics and use http server to get it?
Why not add it into DorisMetric instead?

acelyc111:
> I have a question about dynamic collect the metric.
> If people can remember so many metrics and use http server to get it?
> Why not add it into DorisMetric instead?

Yes, I register all of these metrics to DorisMetric, via a new added macro in be/src/util/doris_metrics.h., and user and access '/metrics' to obtain all metrics.

morningman:
LGTM 
@acelyc111, Hi could you please resolve the conflicts?

acelyc111:
> LGTM
> @acelyc111, Hi could you please resolve the conflicts?

Resolved.

kangkaisen:
@acelyc111 Hi, Would better add a doc to list all new metrics and  describe the metrics meaning. Thanks.

chaoyli:
> > I have a question about dynamic collect the metric.
> > If people can remember so many metrics and use http server to get it?
> > Why not add it into DorisMetric instead?
> 
> Yes, I register all of these metrics to DorisMetric, via a new added macro in be/src/util/doris_metrics.h., and user can visit '/metrics' to obtain all metrics.

I mean you can write the variable in doris_metrics.h, it's can be a unified entrance.
The followers can easy know metrics in doris through reading doris_metris.h.

acelyc111:
> @acelyc111 Hi, Would better add a doc to list all new metrics and describe the metrics meaning. Thanks.

OK! And I want to know that is there a doc describing all the exsited metrics? If not, I think we could provide a way, for example CLI, to dump all metrics, with their description which has been well written when the related metric is defined/registered, then we only need to maintain the code, without an independent doc.

acelyc111:
> > > I have a question about dynamic collect the metric.
> > > If people can remember so many metrics and use http server to get it?
> > > Why not add it into DorisMetric instead?
> > 
> > 
> > Yes, I register all of these metrics to DorisMetric, via a new added macro in be/src/util/doris_metrics.h., and user can visit '/metrics' to obtain all metrics.
> 
> I mean you can write the variable in doris_metrics.h, it's can be a unified entrance.
> The followers can easy know metrics in doris through reading doris_metris.h.

Good advice, done.

morningman:
> OK! And I want to know that is there a doc describing all the exsited metrics? If not, I think we could provide a way, for example CLI, to dump all metrics, with their description which has been well written when the related metric is defined/registered, then we only need to maintain the code, without an independent doc.

It's OK to not add doc in this PR. Currently we do not have a doc for these metrics. We can add this doc in another PR later.

acelyc111:
> > OK! And I want to know that is there a doc describing all the exsited metrics? If not, I think we could provide a way, for example CLI, to dump all metrics, with their description which has been well written when the related metric is defined/registered, then we only need to maintain the code, without an independent doc.
> 
> It's OK to not add doc in this PR. Currently we do not have a doc for these metrics. We can add this doc in another PR later.
> 
> But it's better to list the new metrics you added in this PR in your commit log and make a brief description of them.

Done

morningman:
hi @acelyc111 Please resolve the conflicts

acelyc111:
> hi @acelyc111 Please resolve the conflicts

Done

morningman (approved):
LGTM

PR #3251, opened 2020-04-02
Fix Rewrite count distinct bitmap and hll in order by bug

kangkaisen:
For https://github.com/apache/incubator-doris/issues/3231


imay (approved):
LGTM

PR #3260, opened 2020-04-03
[Storage] open data dirs parallelly

vagetablechicken:
Ref https://github.com/apache/incubator-doris/issues/3247

### DIFF
--- a/be/src/olap/storage_engine.cpp
+++ b/be/src/olap/storage_engine.cpp
@@ -178,7 +178,7 @@ OLAPStatus StorageEngine::_open() {
 }
 
 OLAPStatus StorageEngine::_init_store_map() {
-    std::vector < std::pair<DataDir*> tmp_stores;
+    std::vector<DataDir*> tmp_stores;
     std::vector<std::thread> threads;
     std::atomic<bool> init_error{false};
     for (auto& path : _options.store_paths) {
@@ -189,7 +189,7 @@ OLAPStatus StorageEngine::_init_store_map() {
             auto st = store->init();
             if (!st.ok()) {
                 init_error = true;
-                LOG(WARNING) << "Store load failed, path=" << store->path();
+                LOG(WARNING) << "Store load failed, status="<< st.to_string() << ", path=" << store->path();
             }
         });
     }
--- a/be/src/olap/storage_engine.cpp
+++ b/be/src/olap/storage_engine.cpp
@@ -184,7 +184,6 @@ OLAPStatus StorageEngine::_init_store_map() {
     for (auto& path : _options.store_paths) {
         DataDir* store = new DataDir(path.path, path.capacity_bytes, path.storage_medium,
                                      _tablet_manager.get(), _txn_manager.get());
-        Status* st = new Status();
         tmp_stores.emplace_back(store);
         threads.emplace_back([store, &init_error]() {
             auto st = store->init();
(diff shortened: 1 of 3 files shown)


be/src/olap/storage_engine.cpp
```diff
@@ -148,16 +148,33 @@ void StorageEngine::load_data_dirs(const std::vector<DataDir*>& data_dirs) {
...
-            return OLAP_ERR_INVALID_ROOT_PATH;
+        tmp_vec.emplace_back(std::make_pair(store,std::async([store](){ return store->init();})));
+    }
+
+    try {
+        for (auto& pair : tmp_vec) {
+            DataDir* store = pair.first;
+            auto st = pair.second.get();
+            if (!st.ok()) {
+                throw std::runtime_error("Store load failed, path=" + store->path());
```
imay, line 164: 1. Try to avoid using exceptions
2. Should join all other threads before delete store, otherwise exceptions are prone to occur

imay (changes requested):

be/src/olap/storage_engine.cpp
vagetablechicken: Yes, I forgot it's not safe to delete stores in catch block. 
I think using exception in the phase of storage initialization should be no problem. But we should wait all threads finished, so catching exceptions can't make the code clean.
So I will modify the code like
```
std::vector<status> sts;
std::vector<std::thread> threads;
for dir in data dirs:
    threads.emplace_back([&sts[i],dir](){ sts[i}=dir->init() });
for_each(threads, [](t){ t.join();})
for st in sts:
    if(st.ok()) { ... }
    else { init_error=true; break; }
if(init_error) { 
    // clean up
    return error;
}
```

imay: Looks good for me. Should log detail error message in lambda.

be/src/olap/storage_engine.cpp
```diff
@@ -185,6 +177,39 @@ OLAPStatus StorageEngine::_open() {
...
+    std::atomic<bool> init_error{false};
+    for (auto& path : _options.store_paths) {
+        DataDir* store = new DataDir(path.path, path.capacity_bytes, path.storage_medium,
+                                     _tablet_manager.get(), _txn_manager.get());
+        tmp_stores.emplace_back(store);
+        threads.emplace_back([store, &init_error]() {
+            auto st = store->init();
+            if (!st.ok()) {
+                init_error = true;
+                LOG(WARNING) << "Store load failed, path=" << store->path();
```
imay, line 192: should print st.to_string() too.

be/src/olap/storage_engine.cpp
```diff
@@ -185,6 +177,39 @@ OLAPStatus StorageEngine::_open() {
     return OLAP_SUCCESS;
 }
 
+OLAPStatus StorageEngine::_init_store_map() {
+    std::vector < std::pair<DataDir*> tmp_stores;
```
imay, line 181: ???
```suggestion
    std::vector<DataDir*> tmp_stores;
```

vagetablechicken: cherry-pick mistake :sweat:
When can we have a look at CI output? It will be enormously helpful.

imay (approved):
LGTM

PR #3264, opened 2020-04-04
[config] Support to modify configs when BE is running without restarting

acelyc111:
Ref https://github.com/apache/incubator-doris/issues/3268
In the past, when we want to modify some BE configs, we have to modify be.conf and then restart BE.
This patch provides a way to modify configs in the type of 'threshold', 'interval', 'enable flag'
when BE is running without restarting it.
You can update a single config once by BE's http API: `be_host:be_http_port/api/update_config?config_name=new_value`

### DIFF
--- a/be/src/common/configbase.cpp
+++ b/be/src/common/configbase.cpp
@@ -28,13 +28,147 @@
 #include "common/config.h"
 #undef __IN_CONFIGBASE_CPP__
 
+#include "common/status.h"
+#include "gutil/strings/substitute.h"
+
 namespace doris {
 namespace config {
 
-std::list<Register::Field>* Register::_s_fieldlist = nullptr;
-std::map<std::string, std::string>* confmap = nullptr;
+std::map<std::string, Register::Field>* Register::_s_fieldlist = nullptr;
+std::map<std::string, std::string>* full_conf_map = nullptr;
 
 Properties props;
+
+// trim string
+std::string& trim(std::string& s) {
+    // rtrim
+    s.erase(std::find_if(s.rbegin(), s.rend(), std::not1(std::ptr_fun<int, int>(std::isspace)))
+                    .base(),
+            s.end());
+    // ltrim
+    s.erase(s.begin(),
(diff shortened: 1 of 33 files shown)


imay:
@acelyc111 This is a wonderful job, could you please to create an issue to track this feature and relate this PR to it?

be/src/http/action/update_config_action.cpp
```diff
@@ -0,0 +1,76 @@
...
+    LOG(INFO) << req->debug_string();
+
+    bool success = true;
+    std::string msg;
+    if (req->params()->size() != 1) {
+        success = false;
+        msg = "Now only support to set a single config once, via 'config_name=new_value'.";
+    } else {
+        DCHECK(req->params()->size() == 1);
+        for (const auto& it : *req->params()) {
```
imay, line 51: If number of params is 1, using the for statement can be misleading.

be/src/http/action/update_config_action.cpp
```diff
@@ -0,0 +1,76 @@
...
+            Status s = config::set_config(it.first, it.second);
+            success = s.ok();
+            LOG(WARNING) << "set_config " << it.first << "=" << it.second
```
imay, line 54: when `s.ok()` is true, print INFO log; when `s.ok()` is false, print waring log

be/src/common/configbase.cpp
```diff
@@ -28,91 +28,19 @@
...
 
+#include "common/status.h"
+#include "gutil/strings/substitute.h"
+
 namespace doris {
 namespace config {
 
-std::list<Register::Field>* Register::_s_fieldlist = nullptr;
-std::map<std::string, std::string>* confmap = nullptr;
+std::map<std::string, Register::Field>* Register::_s_fieldlist = nullptr;
```
imay, line 37:
```suggestion
std::map<std::string, Register::Field>* Register::_s_field_map = nullptr;
```

acelyc111:
> @acelyc111 This is a wonderful job, could you please to create an issue to track this feature and relate this PR to it?

OK, https://github.com/apache/incubator-doris/issues/3268

imay (approved):
LGTM

Issue #3266, opened 2020-04-06
[Bug][Plugin] Exception thrown when install/uninstall plugins


morningman:
**Describe the bug**
![3171586144867_ pic_hd](https://user-images.githubusercontent.com/2899462/78546618-49320680-7830-11ea-8f0c-75e08cac5dd6.jpg)

PR #3267, opened 2020-04-06
[Bug] Fix some bugs of install/uninstall plugins

morningman:
1. Avoid losing plugin if plugin failed to load when replaying
    When in replay process, the plugin should always be added to the plugin manager,
    even if that plugin failed to be loaded.

2. `show plugin` statement should show all plugins, not only the successfully installed plugins.

3. plugin's name should be unique globally and case insensitive.

4. Avoid creating new instances of plugins when doing metadata checkpoint.

5. Add a __builtin_ prefix for builtin plugins.

ISSUE: #3266

### DIFF
--- a/fe/src/main/java/org/apache/doris/plugin/PluginMgr.java
+++ b/fe/src/main/java/org/apache/doris/plugin/PluginMgr.java
@@ -20,6 +20,7 @@ package org.apache.doris.plugin;
 import org.apache.doris.analysis.InstallPluginStmt;
 import org.apache.doris.catalog.Catalog;
 import org.apache.doris.common.Config;
+import org.apache.doris.common.DdlException;
 import org.apache.doris.common.UserException;
 import org.apache.doris.common.io.Writable;
 import org.apache.doris.plugin.PluginInfo.PluginType;
@@ -29,6 +30,7 @@ import org.apache.doris.qe.AuditLogBuilder;
 import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -41,17 +43,23 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 
 public class PluginMgr implements Writable {
     private final static Logger LOG = LogManager.getLogger(PluginMgr.class);
 
+    public final static String BUILTIN_PLUGIN_PREFIX = "__builtin_";
+
     private final Map<String, PluginLoader>[] plugins;
+    // all dynamic plugins should have unique names,
+    private final Set<String> dynamicPluginNames;
 
     public PluginMgr() {
-        plugins = new Map[PluginType.MAX_PLUGIN_SIZE];
-        for (int i = 0; i < PluginType.MAX_PLUGIN_SIZE; i++) {
-            plugins[i] = Maps.newConcurrentMap();
+        plugins = new Map[PluginType.MAX_PLUGIN_TYPE_SIZE];
+        for (int i = 0; i < PluginType.MAX_PLUGIN_TYPE_SIZE; i++) {
+            plugins[i] = Maps.newTreeMap(String.CASE_INSENSITIVE_ORDER);
         }
+        dynamicPluginNames = Sets.newTreeSet(String.CASE_INSENSITIVE_ORDER);
     }
 
     // create the plugin dir if missing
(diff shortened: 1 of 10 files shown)


kangpinghuang (approved):

fe/src/main/java/org/apache/doris/plugin/PluginMgr.java
```diff
@@ -41,17 +43,23 @@
...
 
     public PluginMgr() {
-        plugins = new Map[PluginType.MAX_PLUGIN_SIZE];
-        for (int i = 0; i < PluginType.MAX_PLUGIN_SIZE; i++) {
-            plugins[i] = Maps.newConcurrentMap();
+        plugins = new Map[PluginType.MAX_PLUGIN_TYPE_SIZE];
+        for (int i = 0; i < PluginType.MAX_PLUGIN_TYPE_SIZE; i++) {
+            plugins[i] = Maps.newTreeMap(String.CASE_INSENSITIVE_ORDER);
         }
+        dynamicPluginNames = Sets.newTreeSet(String.CASE_INSENSITIVE_ORDER);
```
Seaven, line 62: Maybe you need ConcurrentSkipListSet? I guess

fe/src/main/java/org/apache/doris/plugin/PluginMgr.java
```diff
@@ -86,25 +112,26 @@ public PluginInfo installPlugin(InstallPluginStmt stmt) throws IOException, User
 
         try {
             PluginInfo info = pluginLoader.getPluginInfo();
-
-            if (plugins[info.getTypeId()].containsKey(info.getName())) {
+            
+            if (checkDynamicPluginNameExist(info.getName())) {
```
Seaven, line 118: Using two containers does not avoid concurrency problems

fe/src/main/java/org/apache/doris/plugin/PluginMgr.java
```diff
@@ -86,25 +112,26 @@ public PluginInfo installPlugin(InstallPluginStmt stmt) throws IOException, User
...
+            
+            plugins[info.getTypeId()].put(info.getName(), pluginLoader);
+            
             Catalog.getCurrentCatalog().getEditLog().logInstallPlugin(info);
-            LOG.info("install plugin = " + info.getName());
+            LOG.info("install plugin {}", info.getName());
             return info;
         } catch (IOException | UserException e) {
-            pluginLoader.setStatus(PluginStatus.ERROR);
+            pluginLoader.uninstall();
```
Seaven, line 135: Can't uninstall in there, will cause uninstall real plugin

fe/src/main/java/org/apache/doris/plugin/PluginMgr.java
```diff
@@ -113,30 +140,47 @@ public PluginInfo installPlugin(InstallPluginStmt stmt) throws IOException, User
...
+                loader.uninstall();
+                plugins[i].remove(name);
+                loader.setStatus(PluginStatus.UNINSTALLED);
+                removeDynamicPluginName(name);
+
+                // do not get plugin info by calling loader.getPluginInfo(). That method will try to
+                // reload the plugin properties from source if this plugin is not installed successfully.
+                // Here we only need the plugin's name for persisting.
+                // TODO(cmy): This is a bad design to couple the persist info with PluginInfo, but for
+                // the compatibility, I till use this method.
```
Seaven, line 176: It is a good idea not to use PluginInfo for persist

fe/src/main/java/org/apache/doris/plugin/PluginMgr.java
morningman: No, Concurrent collection only solves the problem of concurrency conflicts, but what I need here is a `serializable isolation`.

fe/src/main/java/org/apache/doris/plugin/PluginMgr.java
morningman: I think it works. 
I first using `checkDynamicPluginNameExist` to check if name already exist.
And after installing, I use `addDynamicPluginNameIfAbsent()` to `check-and-add` name to the `dynamicPluginNames` set. `check-and-add` is an atomic operation.

And `plugins[]` array is just for saving the plugin, I do not use it to check name existence.

fe/src/main/java/org/apache/doris/plugin/PluginMgr.java
morningman: Yes, but I don't want to change the meta persistence method in this pull request. Maybe changed later.

fe/src/main/java/org/apache/doris/plugin/PluginMgr.java
Seaven: Ignore, installPath will not exist if it is not installed

Seaven (approved):

chaoyli (approved):

PR #3276, opened 2020-04-08
[refactor] A small refactor on class DataDir

acelyc111:
main refactor points are:
- Use a single get_absolute_tablet_path function instead of 3
  independent functions
- Remove meaningless return value of register_tablet and deregister_tablet
- Some typo and format

### DIFF
--- a/be/src/olap/data_dir.cpp
+++ b/be/src/olap/data_dir.cpp
@@ -399,7 +399,7 @@ void DataDir::register_tablet(Tablet* tablet) {
     TabletInfo tablet_info(tablet->tablet_id(), tablet->schema_hash(), tablet->tablet_uid());
 
     std::lock_guard<std::mutex> l(_mutex);
-    _tablet_set.emplace(tablet_info);
+    _tablet_set.emplace(std::move(tablet_info));
 }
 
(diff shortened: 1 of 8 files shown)


be/src/olap/data_dir.cpp
```diff
@@ -395,69 +395,47 @@ OLAPStatus DataDir::get_shard(uint64_t* shard) {
...
     _tablet_set.clear();
 }
 
 std::string DataDir::get_absolute_shard_path(const std::string& shard_string) {
     return _path + DATA_PREFIX + "/" + shard_string;
 }
 
-std::string DataDir::get_absolute_tablet_path(TabletMeta* tablet_meta, bool with_schema_hash) {
+template <typename T>
+std::string DataDir::get_absolute_tablet_path(const T& tablet_meta, bool with_schema_hash) {
```
imay, line 424: I think it will be better to have a function with argument shard_id, tablet_id and schema_hash. Then there won't be the following functions which do the same thing.
And the string `operator +` performance is bad, prefer strings::Substitute.

imay (approved):
LGTM

imay:
The UT is successful.
image
I will merge this PR

Issue #3275, opened 2020-04-07
[Bug] str_to_date error


morningman:
**Describe the bug**

```
mysql> select str_to_date('11/09/2011', '%m/%d/%Y');
+---------------------------------------+
| str_to_date('11/09/2011', '%m/%d/%Y') |
+---------------------------------------+
| 2011-11-09 00:00:00                   |
+---------------------------------------+
```

Expected:
```
mysql> select str_to_date('11/09/2011', '%m/%d/%Y');
+---------------------------------------+
| str_to_date('11/09/2011', '%m/%d/%Y') |
+---------------------------------------+
| 2011-11-09                            |
+---------------------------------------+
```

PR #3279, opened 2020-04-08
[BUG] Fix cast result expr bug

HangyuanLiu:
For #3275 

When the result type is a date type, the result expr type should not be cast
Because in the fe function, the specific type of the date type is determined by the actual type of the return value, not by the function return value type

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/StatementBase.java
+++ b/fe/src/main/java/org/apache/doris/analysis/StatementBase.java
@@ -124,6 +124,9 @@ public abstract class StatementBase implements ParseNode {
         Preconditions.checkNotNull(resultExprs);
         Preconditions.checkState(resultExprs.size() == types.size());
         for (int i = 0; i < types.size(); ++i) {
+            if (resultExprs.get(i).getType().isDateType() && types.get(i).isDateType()) {
+                continue;                               
+            }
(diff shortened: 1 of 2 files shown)


kangkaisen (dismissed):
+1. LGTM

fe/src/main/java/org/apache/doris/analysis/StatementBase.java
```diff
@@ -124,6 +124,9 @@ public void castResultExprs(List<Type> types) throws AnalysisException {
         Preconditions.checkNotNull(resultExprs);
         Preconditions.checkState(resultExprs.size() == types.size());
         for (int i = 0; i < types.size(); ++i) {
+            if (resultExprs.get(i).getType().isDateType() && types.get(i).isDateType()) {
```
morningman, line 127: Please add a detailed comment here to explain the problem. Specific examples can be used.

morningman (changes requested):

fe/src/main/java/org/apache/doris/analysis/StatementBase.java
HangyuanLiu: OK

morningman (approved):
LGTM

PR #3281, opened 2020-04-08
fix openssl download url is not avaliable

yangzhg:
the OPENSSL_DOWNLOAD is not avaliable now, it moved to  https://www.openssl.org/source/old/1.0.2/openssl-1.0.2k.tar.gz


imay (approved):
LGTM

PR #3284, opened 2020-04-09
[Storage] Use getmntent_r() for thread-safe

### LINKED ISSUE #3282: Opening data dirs parallelly is not thread-safe
**Describe the bug**
162b1c5d8bfd8afb7b01951ea5c5a79d1ffe1679 makes the data dirs init parallelly, but I found it's not thread-safe.
When be is starting, it may get this error:
```
[data_dir.cpp:249] fail to find file system, path=/xxx
```
I didn't notice that `DataDir::_init_file_system()` may have race conditions.
`getmntent()` value is `MT-Unsafe race:mntentbuf locale`
so mount_entry may be the wrong value `NULL` , `is_find` will be false.
https://github.com/apache/incubator-doris/blob/614a76beeac73821c78903c46e7a703b7956796b/be/src/olap/data_dir.cpp#L228-L251

We have two solutions:
1. use `getmntent_r()` instread, `getmntent_r()` value is MT-Safe locale.
1. split `_init_meta()`(open rocksdb), just `_init_meta()`  in parallel. Keep other parts of data dir init  in sequential order.

vagetablechicken:
ref https://github.com/apache/incubator-doris/issues/3282

### DIFF
--- a/be/src/olap/data_dir.cpp
+++ b/be/src/olap/data_dir.cpp
@@ -225,7 +225,9 @@ Status DataDir::_init_file_system() {
 
     bool is_find = false;
     struct mntent* mount_entry = NULL;
-    while ((mount_entry = getmntent(mount_tablet)) != NULL) {
+    struct mntent ent;
+    char buf[1024];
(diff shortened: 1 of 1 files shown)


imay (approved):
LGTM

PR #3288, opened 2020-04-09
Fix bug when use ZoneMap/BloomFiter on column with REPLACE/REPLACE_IF_NOT_NULL

chaoyli:
Now, column with REPLACE/REPLACE_IF_NOT_NULL can be filtered by ZoneMap/BloomFilter
when the rowset is base(version starts with zero). Always we think is an optimization.
But when some case, it will occurs bug.

```
create table test(
  k1 int,
  v1 int replace,
  v2 int sum
);
```
If I have two records on different two versions
```
1 2 2 on version [0-10]
1 3 1 on version 11
```
If I perform a query 
```
select * from test where k1 = 1 and v1 = 3;
```
The result will be 1 3 1, this is not right because of the first record is filtered.
The right answer is 1 3 3, the v2 should be summed.
Remove this optimization is necessity to make the result is right.

### DIFF
--- a/be/src/olap/rowset/beta_rowset_writer.cpp
+++ b/be/src/olap/rowset/beta_rowset_writer.cpp
@@ -192,7 +192,6 @@ OLAPStatus BetaRowsetWriter::_create_segment_writer() {
 
     DCHECK(wblock != nullptr);
     segment_v2::SegmentWriterOptions writer_options;
-    writer_options.whether_to_filter_value = _context.version.first == 0;
     _segment_writer.reset(new segment_v2::SegmentWriter(
             wblock.get(), _num_segment, _context.tablet_schema, writer_options));
     _wblocks.push_back(std::move(wblock));
--- a/be/src/olap/rowset/segment_reader.cpp
+++ b/be/src/olap/rowset/segment_reader.cpp
@@ -491,10 +491,7 @@ OLAPStatus SegmentReader::_pick_row_groups(uint32_t first_block, uint32_t last_b
 
     for (auto& i : _conditions->columns()) {
         FieldAggregationMethod aggregation = _get_aggregation_by_index(i.first);
-        bool is_continue = (aggregation == OLAP_FIELD_AGGREGATION_NONE
-                            || ((aggregation == OLAP_FIELD_AGGREGATION_REPLACE ||
-                                 aggregation == OLAP_FIELD_AGGREGATION_REPLACE_IF_NOT_NULL)
-                                && _segment_group->version().first == 0));
+        bool is_continue = (aggregation == OLAP_FIELD_AGGREGATION_NONE);
         if (!is_continue) {
             continue;
         }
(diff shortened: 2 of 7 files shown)


chaoyli:
#3289

imay:
If this optimization is removed, some query will suffer a performance degradation.
When the scanned value column type is all REPLACE, it seems that this optimization can be applied.
Can you leave this optimization for this case?

chaoyli:
> If this optimization is removed, some query will suffer a performance degradation.
> When the scanned value column type is all REPLACE, it seems that this optimization can be applied.
> Can you leave this optimization for this case?

I think it's a very corner case to select all columns on REPLACE and will waste much resource.
1. Only base rowset(version starts with zero) can use it, all other rowsets have to save ZoneMap/BloomFilter for REPLACE columns, which is non-sense.
2. REPLACE is value column, the ZoneMap is not very useful because of sorted column is key column.

imay:
> > If this optimization is removed, some query will suffer a performance degradation.
> > When the scanned value column type is all REPLACE, it seems that this optimization can be applied.
> > Can you leave this optimization for this case?
> 
> I think it's a very corner case to select all columns on REPLACE and will waste much resource.
> 
> 1. Only base rowset(version starts with zero) can use it, all other rowsets have to save ZoneMap/BloomFilter for REPLACE columns, which is non-sense.
> 2. REPLACE is value column, the ZoneMap is not very useful because of sorted column is key column.

Beside ZoneMap/BloomFilter, input filter will be applied before rowset streams are merged, this will cause this problem too. And in most cases, base rowset has the most rows, and filtering data before merging will reduce merge operation dramatically. It will result in a good performance.
So I think ZoneMap and BloomFilter is not useful for REPLACE column, however filtering before merging worths it.

chaoyli:
> > > If this optimization is removed, some query will suffer a performance degradation.
> > > When the scanned value column type is all REPLACE, it seems that this optimization can be applied.
> > > Can you leave this optimization for this case?
> > 
> > 
> > I think it's a very corner case to select all columns on REPLACE and will waste much resource.
> > 
> > 1. Only base rowset(version starts with zero) can use it, all other rowsets have to save ZoneMap/BloomFilter for REPLACE columns, which is non-sense.
> > 2. REPLACE is value column, the ZoneMap is not very useful because of sorted column is key column.
> 
> Beside ZoneMap/BloomFilter, input filter will be applied before rowset streams are merged, this will cause this problem too. And in most cases, base rowset has the most rows, and filtering data before merging will reduce merge operation dramatically. It will result in a good performance.
> So I think ZoneMap and BloomFilter is not useful for REPLACE column, however filtering before merging worths it.

Year, filter not use it now, predicate only in Key column can be used to filtered. REPLACE column can be used ZoneMap and BloomFilter to filter base rowset.
For the above reasons, I think can remove it.

imay:
> Year, filter not use it now, predicate only in Key column can be used to filtered. REPLACE column can be used ZoneMap and BloomFilter to filter base rowset.
> For the above reasons, I think can remove it.

Thanks for your reply, I see. After this CL, performance won't be influenced.

imay (approved):
LGTM

Issue #3292, opened 2020-04-10
[Transaction] FE will not cancel some stream load txn when the stream load cannot be finished


WingsGo:
In current StreamLoad design, FE will recive load information from BE by RPC call, but I found that if the coordinator BE down when doing stream load fragment, FE will not receive any rollback txn call and the txns on the coordinator BE will never finished and the state  is always PREPARE,  it is because FE has no idea that the txn will aborted finally cause the fragment execute failed, and be cann't notify FE that case. So, if user set a long timeout stream load in parameters, there are a lots txns in GlobalTxnMgr, and they occupied the txn queue sizes and memory usage in FE, it may block the stream load.So, I define a callback function to cancel all txn which coordiator BE is down to solve this problem

PR #3293, opened 2020-04-10
[Transaction]Cancel all txns whose coordinate be is down.

WingsGo:
This CL solve problem:
- FE cann't aware Coordinate BE down and cancel the txns because the
txns cann't finish.
- Do some codestyle refactor

### DIFF
--- a/fe/src/main/java/org/apache/doris/system/HeartbeatMgr.java
+++ b/fe/src/main/java/org/apache/doris/system/HeartbeatMgr.java
@@ -57,7 +57,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicReference;
 
-/*
+/**
  * Heartbeat manager run as a daemon at a fix interval.
  * For now, it will send heartbeat to all Frontends, Backends and Brokers
  */
@@ -68,7 +68,7 @@ public class HeartbeatMgr extends MasterDaemon {
     private SystemInfoService nodeMgr;
     private HeartbeatFlags heartbeatFlags;
 
-    private static volatile AtomicReference<TMasterInfo> masterInfo = new AtomicReference<TMasterInfo>();
+    private static volatile AtomicReference<TMasterInfo> masterInfo = new AtomicReference<>();
 
     public HeartbeatMgr(SystemInfoService nodeMgr) {
         super("heartbeat mgr", FeConstants.heartbeat_interval_second * 1000);
@@ -87,7 +87,7 @@ public class HeartbeatMgr extends MasterDaemon {
         masterInfo.set(tMasterInfo);
     }
 
-    /*
+    /**
      * At each round:
      * 1. send heartbeat to all nodes
      * 2. collect the heartbeat response from all nodes, and handle them
@@ -176,6 +176,7 @@ public class HeartbeatMgr extends MasterDaemon {
                     if (hbResponse.getStatus() != HbStatus.OK) {
                         // invalid all connections cached in ClientPool
                         ClientPool.backendPool.clearPool(new TNetworkAddress(be.getHost(), be.getBePort()));
+                        Catalog.getCurrentCatalog().getGlobalTransactionMgr().abortTxnWhenCoordinateBeDown(be.getHost(), 100);
                     }
                     return isChanged;
                 }
@@ -293,7 +294,7 @@ public class HeartbeatMgr extends MasterDaemon {
                  */
                 JSONObject root = new JSONObject(result);
                 String status = root.getString("status");
-                if (!status.equals("OK")) {
+                if (!"OK".equals(status)) {
                     return new FrontendHbResponse(fe.getNodeName(), root.getString("msg"));
                 } else {
                     long replayedJournalId = root.getLong(BootstrapFinishAction.REPLAYED_JOURNAL_ID);
--- a/fe/src/main/java/org/apache/doris/system/HeartbeatMgr.java
+++ b/fe/src/main/java/org/apache/doris/system/HeartbeatMgr.java
@@ -176,7 +176,9 @@ public class HeartbeatMgr extends MasterDaemon {
                     if (hbResponse.getStatus() != HbStatus.OK) {
                         // invalid all connections cached in ClientPool
                         ClientPool.backendPool.clearPool(new TNetworkAddress(be.getHost(), be.getBePort()));
-                        Catalog.getCurrentCatalog().getGlobalTransactionMgr().abortTxnWhenCoordinateBeDown(be.getHost(), 100);
+                        if (!isReplay) {
+                            Catalog.getCurrentCatalog().getGlobalTransactionMgr().abortTxnWhenCoordinateBeDown(be.getHost(), 100);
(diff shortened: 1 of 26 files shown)


WingsGo:
For #3292

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -818,24 +817,26 @@ public boolean isPreviousTransactionsFinished(long endTransactionId, long dbId,
...
+     * we think the two lists may have intersection for right ordered txns
+     */
     public boolean isIntersectionNotEmpty(List<Long> sourceTableIdList, List<Long> targetTableIdList) {
         if (CollectionUtils.isEmpty(sourceTableIdList) || CollectionUtils.isEmpty(targetTableIdList)) {
             return true;
         }
-        for (int i = 0; i < sourceTableIdList.size(); i++) {
-            for (int j = 0; j < targetTableIdList.size(); j++) {
-                if (sourceTableIdList.get(i).equals(targetTableIdList.get(j))) {
+        for (Long aLong : sourceTableIdList) {
```
kangkaisen, line 829: Could  rename `aLong` and `value ` to `src` and `target`

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -1389,4 +1389,33 @@ public TransactionState getTransactionStateByCallbackId(long callbackId) {
         }
         return null;
     }
+
+    public List<Long> getTransactionIdByCoordinateBe(String coordinate, int limit) {
```
kangkaisen, line 1393: Would better rename `coordinate` to `coordinateHost`

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
WingsGo: done

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
WingsGo: done

kangkaisen (dismissed):
+1. LGTM

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -1148,8 +1149,7 @@ private void updateDBRunningTxnNum(TransactionStatus preStatus, TransactionState
                 && (curTxnState.getTransactionStatus() == TransactionStatus.PREPARE
                 || curTxnState.getTransactionStatus() == TransactionStatus.COMMITTED)) {
             ++txnNum;
-        } else if (preStatus != null
-                && (preStatus == TransactionStatus.PREPARE
+        } else if ((preStatus == TransactionStatus.PREPARE
```
morningman, line 1152: Why removing the `preStatus != null` check?

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -1389,4 +1389,33 @@ public TransactionState getTransactionStateByCallbackId(long callbackId) {
         }
         return null;
     }
+
+    public List<Long> getTransactionIdByCoordinateBe(String coordinateHost, int limit) {
+        ArrayList<Long> txnIds = new ArrayList<>();
+        readLock();
+        try {
+            idToTransactionState.values().stream()
+                    .filter(t -> (t.getCoordinator().contains(coordinateHost) && (!t.getTransactionStatus().isFinalStatus())))
```
morningman, line 1398: In some clusters, FE and BE are deployed together.
So only by checking `t.getCoordinator().contains(coordinateHost)` will also abort those transactions 
which the coordinator is FE, not the BE, which is not expected.

The current value of the `coordinator` is an informal value, it is just for displaying the job information. If we want to use this `coordinator` field to do some logic control, we should make it formal, for example, by creating a new class "TxnCoordinator" to save the coordinator info.

Currently, the `coordinator` is either "FE: 192.168.0.1" or "BE: 192.168.0.1". So the new class "TxnCoordinator" may looks like:

```
class TxnCoordinator {
    private String source;   // "FE" or "BE"
    private String ip;    // the ip
}
```

And in `getTransactionIdByCoordinateBe`, we should get all transactions with the specified IP and source is BE.

morningman (changes requested):

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
WingsGo: because `preStatus == TransactionStatus.PREPARE` contains the `preStatus != null` check

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
WingsGo: OK, I will change it.

WingsGo: @morningman Done

fe/src/main/java/org/apache/doris/system/HeartbeatMgr.java
```diff
@@ -176,6 +176,7 @@ private boolean handleHbResponse(HeartbeatResponse response, boolean isReplay) {
                     if (hbResponse.getStatus() != HbStatus.OK) {
                         // invalid all connections cached in ClientPool
                         ClientPool.backendPool.clearPool(new TNetworkAddress(be.getHost(), be.getBePort()));
+                        Catalog.getCurrentCatalog().getGlobalTransactionMgr().abortTxnWhenCoordinateBeDown(be.getHost(), 100);
```
morningman, line 179: Why here is a limit 100?

WingsGo: because i afraid too many txn cancel will influence the heatbeat function

morningman: I see. 
Another 2 things:
1. Plz rebase the code, the FE meta version 82 has been taken.
2. in `handleHbResponse()` method, if the parameter `isReplay` is true, means it is a meta log replay thread, so we don't need to process the abort in a replay thread(This is not like `clearPool()`, which should also be done in replay thread).

WingsGo: thanks, I see~, and all things is done.

morningman (approved):
LGTM

morningman (changes requested):
Compile failed.
In `DeleteHandler.java`, line 170, there is a `TxnCoordinator` with String Type, you have to change it.

WingsGo:
> Compile failed.
> In `DeleteHandler.java`, line 170, there is a `TxnCoordinator` with String Type, you have to change it.

done

morningman (approved):
LGTM

PR #3294, opened 2020-04-10
Use read lock when iterate tablet_map in TabletManager::start_trash_sweep

liutang123:
For #3299
In our cluster, there are many stream load jobs.
We found that these jobs slow down periodically.
By add logs, we found that `TabletManager::start_trash_sweep` will take write lock for a long time.
In this PR, we split `start_trash_sweep` in two phases:
First, iterate the tablet map with read lock.
Second, remove the empty tablet in map with write lock.

### DIFF
--- a/be/src/olap/tablet_manager.cpp
+++ b/be/src/olap/tablet_manager.cpp
@@ -960,28 +960,36 @@ OLAPStatus TabletManager::start_trash_sweep() {
     {
         std::vector<int64_t> tablets_to_clean;
         for (int32 i = 0; i < _tablet_map_lock_shard_size; i++) {
-            WriteLock rlock(&_tablet_map_lock_array[i]);
             tablet_map_t& tablet_map = _tablet_map_array[i];
-            for (auto& item : tablet_map) {
-                // try to clean empty item
-                if (item.second.table_arr.empty()) {
-                    // try to get schema change lock if could get schema change lock, then nobody
-                    // own the lock could remove the item
-                    // it will core if schema change thread may hold the lock and this thread will deconstruct lock
-                    if (item.second.schema_change_lock.trylock() == OLAP_SUCCESS) {
-                        item.second.schema_change_lock.unlock();
+            {
+                ReadLock r_lock(&_tablet_map_lock_array[i]);
+                for (auto& item : tablet_map) {
+                    // try to clean empty item
+                    if (item.second.table_arr.empty()) {
                         tablets_to_clean.push_back(item.first);
                     }
-                }
-                for (TabletSharedPtr tablet : item.second.table_arr) {
-                    tablet->delete_expired_inc_rowsets();
+                    for (TabletSharedPtr tablet : item.second.table_arr) {
+                        tablet->delete_expired_inc_rowsets();
+                    }
                 }
             }
-            // clean empty tablet id item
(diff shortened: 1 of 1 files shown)


kangkaisen:
@liutang123 Please add a issue for this PR.

kangkaisen (dismissed):
+1. LGTM

imay:
> By add logs, we found that `TabletManager::start_trash_sweep` will take write lock for a long time.

How long? And after this PR applied, how long will it be?

be/src/olap/tablet_manager.cpp
```diff
@@ -960,28 +960,36 @@ OLAPStatus TabletManager::start_trash_sweep() {
...
+                ReadLock r_lock(&_tablet_map_lock_array[i]);
+                for (auto& item : tablet_map) {
+                    // try to clean empty item
+                    if (item.second.table_arr.empty()) {
                         tablets_to_clean.push_back(item.first);
                     }
-                }
-                for (TabletSharedPtr tablet : item.second.table_arr) {
-                    tablet->delete_expired_inc_rowsets();
+                    for (TabletSharedPtr tablet : item.second.table_arr) {
```
chaoyli, line 971: You can put tablet->delete_expired_inc_rowsets() between the ReadLock and WriteLock.
This is bottleneck which consumes much resource.

chaoyli (changes requested):

be/src/olap/tablet_manager.cpp
liutang123: Can you give the example code? 
I am not sure whether it is safe to iterate a map without read lock.

liutang123:
> > By add logs, we found that `TabletManager::start_trash_sweep` will take write lock for a long time.
> 
> How long? And after this PR applied, how long will it be?

We set `tablet_map_shard_size` to 16, this function will holds each map lock about more than 1 second.

be/src/olap/tablet_manager.cpp
morningman: Currently, there should be only one tablet in `tablet_arr`. 
Only in the early Doris version, `tablet_arr` will appear multiple tablets.

And in `tablet->delete_expired_inc_rowsets();`, it will hold the `meta lock`, So need to be protected by `_tablet_map`.

So you can just by doing:

1. do the `tablets_to_clean.push_back(item.first);` in `ReadLock`
2. Unlock the `ReadLock`, and iterate the `tablets_to_clean` to do the `tablet->delete_expired_inc_rowsets();`
3. Do the following `tablet_map.erase(tablet_id_to_clean);` in `WriteLock`

liutang123: > 2. Unlock the `ReadLock`, and iterate the `tablets_to_clean` to do the `tablet->delete_expired_inc_rowsets();`
We should invoke `delete_expired_inc_rowsets` in all tablets but not only in tablets_to_clean. Isn't it?

morningman (approved):
LGTM

kangkaisen (approved):
+1. LGTM

liutang123:
@chaoyli Hi, do you have time to see this PR again.
This PR has been online in our cluster for two days.

chaoyli (approved):

Issue #3297, opened 2020-04-10
Rewrite some count expr case


HangyuanLiu:
Query statements such as “select min(user_id) from test having count(1) > 0“ may be generated by some BI tools. But having(count(1)) > 0 doesn't make sense. It's a constant of true. But it add a count aggregation. So we can replace it by rewriting the rule.

PR #3298, opened 2020-04-10
Add count rewrite rule

HangyuanLiu:
For #3297


chaoyli:
If emtpy table, count(1) = 0, which will results?

HangyuanLiu:
I made a mistake in my thinking，I temporarily shut down the PR

PR #3310, opened 2020-04-13
[Bug] Use equals() method to judge whether "type" are equal

### LINKED ISSUE #3309: [Bug] errCode = 2, detailMessage = errCode = 2, detailMessage = date literal [2020-03-06] is invalid
**Describe the bug**
Error thrown when executing load statement.

**To Reproduce**

```
CREATE TABLE `app_profile` (
  `event_date` date NOT NULL COMMENT "",
  `app_name` varchar(64) NOT NULL COMMENT "",
  `package_name` varchar(64) NOT NULL COMMENT "",
  `age` varchar(32) NOT NULL COMMENT "",
  `gender` varchar(32) NOT NULL COMMENT "",
  `level` varchar(64) NOT NULL COMMENT "",
  `city` varchar(64) NOT NULL COMMENT "",
  `model` varchar(64) NOT NULL COMMENT "",
  `brand` varchar(64) NOT NULL COMMENT "",
  `hours` varchar(16) NOT NULL COMMENT "",
  `use_num` int(11) SUM NOT NULL COMMENT "",
  `use_time` double SUM NOT NULL COMMENT "",
  `start_times` bigint(20) SUM NOT NULL COMMENT ""
) ENGINE=OLAP
AGGREGATE KEY(`event_date`, `app_name`, `package_name`, `age`, `gender`, `level`, `city`, `model`, `brand`, `hours`)
COMMENT "OLAP"
PARTITION BY RANGE(`event_date`)
(PARTITION p_20200301 VALUES [('2020-02-27'), ('2020-03-02')),
PARTITION p_20200306 VALUES [('2020-03-02'), ('2020-03-07')))
DISTRIBUTED BY HASH(`event_date`, `app_name`, `package_name`, `age`, `gender`, `level`, `city`, `model`, `brand`, `hours`) BUCKETS 32
PROPERTIES (
 "replication_num" = "1"
);

load label db1.app_profile_20200306 ( DATA INFILE('filexxx') INTO TABLE app_profile partition (p_20200306) COLUMNS TERMINATED BY '\t' (app_name,package_name,age,gender,level,city,model,brand,hours,use_num,use_time,start_times) SET ( event_date = default_value('2020-03-06') ) ) PROPERTIES ( 'max_filter_ratio'='0.0001' );
```

Sometimes it will throw:

```
errCode = 2, detailMessage = errCode = 2, detailMessage = date literal [2020-03-06] is invalid
```

**Why**

This is because here:

https://github.com/apache/incubator-doris/blob/be090f5929c3ed866af9a8789b2008ae4dc7e921/fe/src/main/java/org/apache/doris/analysis/DateLiteral.java#L194

the `if` clause return false, even if the `type` is truly `DATE`, but `equals()` will return true.

I really do not know why, it should be return true by using `==`, because we never create the 
new instance of `Type.DATE`, but to reference the static member in Type.

morningman:
I don't why, but I found that sometimes when I use "==" to judge the equality of type,
it return false, even if the types are exactly same.

ISSUE: #3309 

This CL only changes `==` to `equals()` to solve the problem, but the reason is still unknown.

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/Analyzer.java
+++ b/fe/src/main/java/org/apache/doris/analysis/Analyzer.java
@@ -1412,7 +1412,7 @@ public class Analyzer {
             // TODO(zc)
             compatibleType = Type.getCmpType(compatibleType, exprs.get(i).getType());
         }
-        if (compatibleType == Type.VARCHAR) {
+        if (compatibleType.equals(Type.VARCHAR)) {
             if (exprs.get(0).getType().isDateType()) {
                 compatibleType = Type.DATETIME;
             }
--- a/fe/src/main/java/org/apache/doris/analysis/CastExpr.java
+++ b/fe/src/main/java/org/apache/doris/analysis/CastExpr.java
@@ -105,8 +105,8 @@ public class CastExpr extends Expr {
                 }
                 // Disable casting from boolean to decimal or datetime or date
                 if (fromType.isBoolean() &&
-                        (toType == Type.DECIMAL || toType == Type.DECIMALV2 ||
-                                toType == Type.DATETIME || toType == Type.DATE)) {
+                        (toType.equals(Type.DECIMAL) || toType.equals(Type.DECIMALV2) ||
+                                toType.equals(Type.DATETIME) || toType.equals(Type.DATE))) {
                     continue;
                 }
                 // Disable no-op casts
--- a/fe/src/main/java/org/apache/doris/analysis/ColumnDef.java
+++ b/fe/src/main/java/org/apache/doris/analysis/ColumnDef.java
@@ -205,7 +205,7 @@ public class ColumnDef {
                 break;
             case FLOAT:
                 FloatLiteral floatLiteral = new FloatLiteral(defaultValue);
-                if (floatLiteral.getType() == Type.DOUBLE) {
+                if (floatLiteral.getType().equals(Type.DOUBLE)) {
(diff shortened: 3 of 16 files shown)


imay:
@morningman 
I think we should replace all `==` compare with `equals`.  `==` returns true only when the two sides refer the same object. I think all the type compare means to test if they are the same type, not the same object.

morningman:
> @morningman
> I think we should replace all `==` compare with `equals`. `==` returns true only when the two sides refer the same object. I think all the type compare means to test if they are the same type, not the same object.

I agree. I am just puzzled why they are not the same object. I think they should be....

imay:
> I agree. I am just puzzled why they are not the same object. I think they should be....

Without knowing why, we can still make Doris work in the right way.
So we can move it to another issue to investigate later on.

morningman:
> > I agree. I am just puzzled why they are not the same object. I think they should be....
> 
> Without knowing why, we can still make Doris work in the right way.
> So we can move it to another issue to investigate later on.

OK, I will try to find out all `==` and change them to `equals()`

kangpinghuang:
I think we should add some unit tests.

morningman:
> I think we should add some unit tests.

OK, I added a case described in #3309

kangkaisen (approved):
+1. LGTM

kangpinghuang (approved):
+1,LGTM

PR #3313, opened 2020-04-14
Fix Restore job block at COMMITTING phase

### LINKED ISSUE #3312: [Restore] Restore job block at COMMITTING phase
**Describe the bug**
restore job would block at committing phase

**To Reproduce**
There is a small probability of occurrence in backup job

**Screenshots**

image

Youngwb:
Due to some previous bug, the segment_group_id of some rowsets does not start from 0. To be compatible with the previous problem, Restore sets the segment_group_id according to the original rowsets

FIx #3312

### DIFF
--- a/be/src/olap/rowset/alpha_rowset_writer.cpp
+++ b/be/src/olap/rowset/alpha_rowset_writer.cpp
@@ -102,6 +102,7 @@ OLAPStatus AlphaRowsetWriter::add_rowset(RowsetSharedPtr rowset) {
         RETURN_NOT_OK(segment_group->link_segments_to_path(_rowset_writer_context.rowset_path_prefix,
                                                            _rowset_writer_context.rowset_id));
         _cur_segment_group->set_empty(segment_group->empty());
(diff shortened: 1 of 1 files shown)


morningman:
Could you point out what is `previous bug`?

Youngwb:
sorry, I don't known the exact bug，and I just got this information from @chaoyli

morningman:
@chaoyli Could you please review this PR?

chaoyli:
If not start with zero, it will cause another problem in future. I think you should handle this error manually.

Youngwb:
`another problem in future` means such as schema change operation may meet some problem？ `handle this error manually` means I need modify the rowset segment_group_id manually before backup it to the HDFS, so the process of restore will not meet this problem ?

chaoyli:
> `another problem in future` means such as schema change operation may meet some problem？ `handle this error manually` means I need modify the rowset segment_group_id manually before backup it to the HDFS, so the process of restore will not meet this problem ? @chaoyli

Yes, If you change it the code, it still a bug because the storage engine think the segment group id starts with zero. So you should manually modify it better.

PR #3316, opened 2020-04-14
[rowset id] Reduce memory of UniqueRowsetIdGenerator

vagetablechicken:
Ref https://github.com/apache/incubator-doris/issues/3307
save 2/3 of memory


morningman (approved):
LGTM

Issue #3319, opened 2020-04-15
Support Java Verision HyperLogLog


wangbo:
**Description**
For Spark Load Process,we need a java version hll to calculate approximate distinct value.
The main request for java's implementation is try to keep consistent with Be C++ version

**Hll Write**
The ```Hll Write``` can be consistent with C++ version.
For java treats unsigned and signed numbers in the same way when execute addition,substraction, multiplication and shift. 
The above operator is enough for ```Hll Write```.
For ```mod operator```, we can use BigInteger to execute ```mod```

**data structure**
The main data structure for hll is Integer and not touch unsigned long, so this can be consistent with C++

**Hll.estimateCardinality**
Need a further research to judge whether C++ 's float-point calculation can be totally consistent with Java's float-point calculation.
Though they all follow ```IEEE 754``` rule, but for C++, OS architecture and compile option may affect the real calculation.
```estimateCardinality``` is not necessary for Spark Load Process, so this is not a urgent problem for now.

924060929:
@wangbo I already implement hll/bitmap scala version add do many unit test in our company, see data/spark-on-doris/common/src/main/scala/org/apache/doris/spark/sql/function/

PR #3320, opened 2020-04-15
(#3319) support java version hyperloglog

wangbo:
#3319
mainly used for Spark Load process to calculate approximate deduplication value and then serialize to parquet file.
Try to keep the same calculation semantic with be's C++ version

#3433

### DIFF
--- a/fe/src/test/java/org/apache/doris/load/loadv2/HllTest.java
+++ b/fe/src/test/java/org/apache/doris/load/loadv2/HllTest.java
@@ -99,7 +99,6 @@ public class HllTest {
         Assert.assertTrue(fullHll.getType() == HLL_DATA_FULL);
         // the result 32748 is consistent with C++ 's implementation
         Assert.assertTrue(fullHll.estimateCardinality() == 32748);
-        // 2% error rate
         Assert.assertTrue(fullHll.estimateCardinality() > Short.MAX_VALUE * (1 - 0.02) &&
                 fullHll.estimateCardinality() < Short.MAX_VALUE * (1 + 0.02));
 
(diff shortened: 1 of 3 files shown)


chaoyli:
You should describe you job in commit log

morningman:
Is it possible to consider using some test data, comparing the hll data obtained from the C++ version with the java version to verify whether the two versions of the algorithm are consistent? And this verification can be put into unit tests.

wangbo:
>Is it possible to consider using some test data, comparing the hll data obtained from the C++ version with the java version to verify whether the two versions of the algorithm are consistent? And this verification can be put into unit tests.
@morningman Current UT has contains the case that the approximation from``` 0 to Short.MAX ```  is consistent with be. I'll try to find more special case

wangbo:
1 use little endian when serialize hll
2 add some ut which keep same logic which be hll ut and value check
@morningman plz review

fe/src/test/java/org/apache/doris/load/loadv2/HllTest.java
```diff
@@ -0,0 +1,268 @@
...
+        DataInputStream fullHllInputStream = new DataInputStream(new ByteArrayInputStream(fullHllOutputStream.toByteArray()));
+        Hll deserializedFullHll = new Hll();
+        deserializedFullHll.deserialize(fullHllInputStream);
+        Assert.assertTrue(deserializedFullHll.getType() == HLL_DATA_FULL);
+        Assert.assertTrue(deserializedFullHll.estimateCardinality() == fullHll.estimateCardinality());
+
+    }
+
+    // keep logic same with C++ version
+    // add additional compare logic with C++ version's estimateValue
```
morningman, line 117: better to add comment in `be/test/olap/hll_test.cpp` to notice that the following test case should be consistent with java version in `fe/src/test/java/org/apache/doris/load/loadv2/HllTest.java`

wangbo: 👌

morningman (approved):
LGTM

PR #3324, opened 2020-04-15
Enable to read and write boolean type data

### LINKED ISSUE #3323: Enable to read and write boolean type data
Is your feature request related to a problem? Please describe.

Now, Doris has boolean type, but there are still some error or bug for reading or writing boolean data, showing create table which include boolean type.

Refer to mysql, we use 0 to indicate false, and 1 to indicate true,
when string is inserted, the value is 0
when number is inserted the value is 0 when the number is 0, and the other value is 1 for other condition
when boolean values is inserted, the value is 0 for false value, and 1 for true value

caiconghui:
This PR is to  enable to read and write boolean type data for segment v1

### DIFF
--- a/be/src/olap/rowset/column_reader.cpp
+++ b/be/src/olap/rowset/column_reader.cpp
@@ -547,6 +547,7 @@ ColumnReader* ColumnReader::create(uint32_t column_id,
     }
 
     switch (column.type()) {
+    case OLAP_FIELD_TYPE_BOOL:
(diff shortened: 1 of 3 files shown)


caiconghui:
for #3323

morningman:
What about segment v2?

caiconghui:
> What about segment v2?

The segment v2 works fine now, boolean type data write and read is suppoted by RlePageBuilder

chaoyli (approved):

PR #3326, opened 2020-04-15
[LSAN] Fix some memory leak detected by LSAN

### LINKED ISSUE #3328: Use LSAN to fix some memory leaks
**Describe the bug**
In our production environment, we found BE physical memory consumed continuously and lead to OOM several days later.
image

We doubt there are some memory leak in BE, we could use [LSAN](http://doris.apache.org/documentation/cn/developer-guide/debug-tool.html#lsan) to detect memory leak of unit tests at first.

acelyc111:


morningman:
Could you make an ISSUE to describe the bug you try to fix?

be/src/exec/es/es_predicate.cpp
```diff
@@ -178,7 +179,7 @@ EsPredicate::EsPredicate(ExprContext* context,
 
 EsPredicate::~EsPredicate() {
     for(int i=0; i < _disjuncts.size(); i++) {
-        delete _disjuncts[i];
+        SAFE_DELETE(_disjuncts[i]);
```
imay, line 182: It is safe to `delete nullptr`, and `_disjuncts` is cleared later on, so keep using delete is OK

be/src/exec/es/es_query_builder.cpp
```diff
@@ -169,20 +170,16 @@ BooleanQueryBuilder::BooleanQueryBuilder() {
 }
 BooleanQueryBuilder::~BooleanQueryBuilder() {
     for (auto clause : _must_clauses) {
-        delete clause;
-        clause = nullptr;
+        SAFE_DELETE(clause);
```
imay, line 173: just delete is OK

acelyc111:
> Could you make an ISSUE to describe the bug you try to fix?

Added it, https://github.com/apache/incubator-doris/issues/3328

morningman:
Does this CL solve the memory leak problem as you describe in your issue? #3328

acelyc111:
> Does this CL solve the memory leak problem as you describe in your issue? #3328

I'll report the result after some days‘ observation, may the next Monday.

morningman (approved):
LGTM

PR #3336, opened 2020-04-16
fix cooldown timestamp bug

kangpinghuang:
when add a parition with storage_cooldown_time property like this:
alter table tablexxx ADD PARTITION p20200421 VALUES LESS THAN("1588262400") ("storage_medium" = "SSD", "storage_cooldown_time" = "2020-05-01 00:00:00");
and show partitions from tablexxx;
the CooldownTime is wrong: 2610-02-17 10:16:40, and what is more, the storage migration is based on the wrong timestamp.
The reason is that the result of DateLiteral.getLongValue is not timestamp.

### DIFF
--- a/fe/src/main/java/org/apache/doris/common/util/PropertyAnalyzer.java
+++ b/fe/src/main/java/org/apache/doris/common/util/PropertyAnalyzer.java
@@ -110,7 +110,7 @@ public class PropertyAnalyzer {
             } else if (!hasCooldown && key.equalsIgnoreCase(PROPERTIES_STORAGE_COLDOWN_TIME)) {
                 hasCooldown = true;
                 DateLiteral dateLiteral = new DateLiteral(value, Type.DATETIME);
-                coolDownTimeStamp = dateLiteral.getLongValue();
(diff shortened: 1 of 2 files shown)


yiguolei:
LGTM

fe/src/main/java/org/apache/doris/common/util/PropertyAnalyzer.java
```diff
@@ -110,7 +110,7 @@ public static DataProperty analyzeDataProperty(Map<String, String> properties, D
             } else if (!hasCooldown && key.equalsIgnoreCase(PROPERTIES_STORAGE_COLDOWN_TIME)) {
                 hasCooldown = true;
                 DateLiteral dateLiteral = new DateLiteral(value, Type.DATETIME);
-                coolDownTimeStamp = dateLiteral.getLongValue();
+                coolDownTimeStamp = dateLiteral.unixTimestamp(TimeUtils.getTimeZone());
```
kangkaisen, line 113: Please add a UT for this bug

kangkaisen (approved):
+1, LGTM

PR #3338, opened 2020-04-16
[ISSUE #3190]Make some correct in delete

### LINKED ISSUE #3190: [Optimize] Simplify the delete process

xy720:
#3190
1. Correct the directory of DeleteJob.java
2. Fix some logic fault in DeleteHandlerTest.java
3. Add timeout value in log and exception


kangkaisen (approved):
+1，LGTM

PR #3339, opened 2020-04-17
[tablet] A small refactor on class Tablet

acelyc111:
There is no functional changes in this patch.
Key refactor points are:
- Remove meaningless return value of functions in class Tablet, and
  also some related functions in other classes
- Allow RowsetGraph::capture_consistent_versions to pass a nullptr
  to the output parameter
- Use CHECK instead of LOG(FATAL) to simplify code

### DIFF
--- a/be/src/olap/rowset_graph.cpp
+++ b/be/src/olap/rowset_graph.cpp
@@ -159,11 +159,6 @@ OLAPStatus RowsetGraph::capture_consistent_versions(const Version& spec_version,
         return OLAP_ERR_INPUT_PARAMETER_ERROR;
     }
 
-    if (version_path == nullptr) {
-        LOG(WARNING) << "param version_path is nullptr.";
-        return OLAP_ERR_INPUT_PARAMETER_ERROR;
-    }
-
     // bfs_queue's element is vertex_index.
     std::queue<int64_t> bfs_queue;
     // predecessor[i] means the predecessor of vertex_index 'i'.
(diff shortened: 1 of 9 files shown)


be/src/olap/tablet.cpp
```diff
@@ -135,16 +130,12 @@ OLAPStatus Tablet::set_tablet_state(TabletState state) {
...
 // should save tablet meta to remote meta store
 // if it's a primary replica
-OLAPStatus Tablet::save_meta() {
-    OLAPStatus res = _tablet_meta->save_meta(_data_dir);
-    if (res != OLAP_SUCCESS) {
-       LOG(FATAL) << "fail to save tablet_meta. res=" << res << ", root=" << _data_dir->path();
-    }
+void Tablet::save_meta() {
+    auto res = _tablet_meta->save_meta(_data_dir);
+    CHECK_EQ(res, OLAP_SUCCESS) << "fail to save tablet_meta. res=" << res << ", root=" << _data_dir->path();
```
chaoyli, line 135: CHECK_EQ takes no effect on release version.

be/src/olap/rowset_graph.cpp
```diff
@@ -237,25 +232,26 @@ OLAPStatus RowsetGraph::capture_consistent_versions(const Version& spec_version,
...
-        int64_t tmp_start_vertex_value = _version_graph[reversed_path[path_id]].value;
-        int64_t tmp_end_vertex_value = _version_graph[reversed_path[path_id - 1]].value;
-
-        // tmp_start_vertex_value mustn't be equal to tmp_end_vertex_value
-        if (tmp_start_vertex_value <= tmp_end_vertex_value) {
-            version_path->emplace_back(tmp_start_vertex_value, tmp_end_vertex_value - 1);
-        } else {
-            version_path->emplace_back(tmp_end_vertex_value, tmp_start_vertex_value - 1);
-        }
+    if (version_path != nullptr) {
```
chaoyli, line 235: I think should not coupling outside VersionPath empty logic in this function.

be/src/olap/rowset_graph.cpp
```diff
@@ -159,11 +159,6 @@ OLAPStatus RowsetGraph::capture_consistent_versions(const Version& spec_version,
         return OLAP_ERR_INPUT_PARAMETER_ERROR;
     }
```
chaoyli, line 161: You can CHECK(VERSION_PATH != nullptr) to replace above logic.

chaoyli (changes requested):

be/src/olap/rowset_graph.cpp
acelyc111: I found `RowsetGraph::capture_consistent_versions` is now only called by `Tablet::capture_consistent_versions`, the latter is now called by two purposes. First, check version integrity, second, check version integrity and then capture versions when check success. For the first purpose, it's not needed to construct and fill a output parameter. Of course the function name `capture_consistent_versions ` maybe not accuracy in this case.

morningman (changes requested):
Compile failed.

be/src/olap/tablet.cpp
chaoyli: Yes

acelyc111:
> Compile failed.

Fixed.

be/src/olap/tablet.cpp
acelyc111: `DCHECK_EQ` has no effect, but here is `CHECK_EQ ` which has effect in release mode.

chaoyli (approved):

be/src/olap/rowset_graph.cpp
acelyc111: Either `version_path`  is nullptr or is valid for this function. If it's nullptr, just validate rowset graph; if it's not nullptr, validate rowset graph and return current versions.

morningman (approved):
LGTM

morningman (approved):
LGTM

PR #3341, opened 2020-04-17
[Doris On ES]Add simple explain for EsTable

wuyunfeng:
related issue: https://github.com/apache/incubator-doris/issues/3306
Note: this PR just remove the `es_scan_node_test.cpp` which is useless

For the moment, just add a simple explain syntax for EsTable without translating the native predicates to ES queryDSL which is better to finished with moving the `predicate translating`  from  `Doris BE` to `Doris FE`,  the whole work is still WIP.

```
 explain select * from test;
```
partial output:

image

```
 explain select * from test where k2 > 2;
```
partial output:

image


morningman (dismissed):
LGTM

kangkaisen:
For https://github.com/apache/incubator-doris/issues/3327

kangkaisen (approved):
+1, LGTM

PR #3348, opened 2020-04-17
[Bug] Partition key's type has been changed after executing queries

### LINKED ISSUE #3347: [Bug] Partition key's type has been changed after executing queries
**Describe the bug**
After executing queries, and than execute `show partitions from tbl`, you may find that the
partition key's type is DATE, but value shows like DATETIME:

```
type[DATE], value: [2020-04-05 00:00:00]
```

But it should be:

```
type[DATE], value: [2020-04-05]
```

morningman:
Expr's `uncheckedCastTo()` method should return a new instance of casted expr.
The origin expr should remain unchanged.

Fix #3347


morningman:
> @morningman Hi, Please add a UT.

Added

kangkaisen (approved):
+1, LGTM

PR #3351, opened 2020-04-19
[Doris On ES] Pushdown some  castexpr predicate to ES

wuyunfeng:
Process castexpr, such as: k (float) > 2.0, k(int) > 3.2, Doris On Es should ignore this doris native cast transformation for every row's col value, we push down this `cast semantic` to Elasticsearch.  

I believe in this `predicate` situation, would decrease the mount of data for transmission。

k1 is float:

````
k1 >= 5
````

push-down filter:

```
{"range":{"k1":{"gte":"5.000000"}}}
```
k2 is int :

```
k2 > 3.2
```

push-down filter:

```
{"range":{"k2":{"gte":"3.2"}}}
```

### DIFF
--- a/be/src/exec/es/es_predicate.cpp
+++ b/be/src/exec/es/es_predicate.cpp
@@ -235,20 +235,14 @@ Status EsPredicate::build_disjuncts_list(const Expr* conjunct) {
         if (TExprNodeType::SLOT_REF == conjunct->get_child(0)->node_type()
             || TExprNodeType::CAST_EXPR == conjunct->get_child(0)->node_type()) {
             expr = conjunct->get_child(1);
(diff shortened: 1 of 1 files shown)


wuyunfeng:
for https://github.com/apache/incubator-doris/issues/3306

kangkaisen:
@wuyunfeng Hi, please add a UT to cover this case in `EsPredicateTest`.

wuyunfeng:
@kangkaisen For now, it is difficult to construct the `ExprContext`, the previous `EsPredicateTest ` is meaningfulless

kangkaisen (approved):
+1， LGTM

imay (approved):
LGTM

Issue #3353, opened 2020-04-19
[Bug] Colocate Join Bug Fix


wangbo:
**Colocate Group sync error status between fe**
This is just a clerical error, pr is more clear.

**Colocate Group  miss remove table operation in edit log**
In current code ,no place calls ```Editlog.logColocateRemoveTable```
which means a colocate group may appear again after replay edit log.
The solution is to calling ```Editlog.logColocateRemoveTable```  when a table is finally deleted in ```CatalogRecycleBin```

PR #3354, opened 2020-04-19
(#3353) Colocate Join Bug Fix

wangbo:
(#3353)

1 Fix sync error colocate group status between fe
2 Fix losing call of ```EditLog.logColocateRemoveTable ```


wangbo:
This pr now is just for review.
It can't be merge after I sync this pr to out production env and the error status disappear.

kangkaisen (approved):
+1. LGTM

PR #3357, opened 2020-04-20
disable some unsupported syntax

yangzhg:
#3373
disable some syntax when subquery is not in binary predicate in case when clause

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
+++ b/fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
@@ -1323,7 +1323,11 @@ public class SelectStmt extends QueryStmt {
     @Override
     public void rewriteExprs(ExprRewriter rewriter) throws AnalysisException {
         Preconditions.checkState(isAnalyzed());
+        int changeNum = rewriter.getNumChanges();
         rewriteSelectList(rewriter);
+        if (changeNum < rewriter.getNumChanges()) {
+            sqlString_ = null;
+        }
         for (TableRef ref : fromClause_) {
             ref.rewriteExprs(rewriter, analyzer);
         }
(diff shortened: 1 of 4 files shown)


morningman:
Hi @yangzhg , you need to create an issue to describe your problem, better to give an example about the query you try to disable.

yangzhg:
> Hi @yangzhg , you need to create an issue to describe your problem, better to give an example about the query you try to disable.

done

fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
```diff
@@ -1362,9 +1362,24 @@ private void rewriteSelectList(ExprRewriter rewriter) throws AnalysisException {
...
+            CaseExpr caseExpr = (CaseExpr) item.getExpr();
+
+            int childIdx = 0;
+            if (caseExpr.hasCaseExpr()
+                    && caseExpr.getChild(childIdx++).contains(Predicates.instanceOf(Subquery.class))) {
+                throw new AnalysisException("Only support subquery in binary predicate in case statement.");
+            }
+            while (childIdx + 2 <= caseExpr.getChildren().size()) {
+                Expr child = caseExpr.getChild(childIdx++);
+                // case
```
morningman, line 1374: `caseExpr.getChild(childIdx++)` here you got the `when` expr

fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
```diff
@@ -1362,9 +1362,24 @@ private void rewriteSelectList(ExprRewriter rewriter) throws AnalysisException {
...
+                if (!(child instanceof BinaryPredicate) && child.contains(Predicates.instanceOf(Subquery.class))) {
+                    throw new AnalysisException("Only support subquery in binary predicate in case statement.");
+                }
+                // when
```
morningman, line 1378: Here is to skip `then` expr, not `when`

kangkaisen:
@yangzhg Please add a UT

EmmyMiao87 (approved):
LGTM

Issue #3358, opened 2020-04-20
[Bug][Dynamic Partition]Dynamic partition properties is not consistent in FE


WingsGo:
**Describe the bug**
I found that when we create a table without dynamic partition properties, and use alter clause to add dynamic partition properties to the table, we can only see the properties in master FE, this is because in `modifyTableDynamicPartition` we use the create table logic to check properties, but it may remove the `key-value` from properties, this will cause when we replay, the properties is empty, so we cann't see dynamic partition properties in Follower FE

https://github.com/apache/incubator-doris/blob/c69bf9ac449336e69e6859c40aca8c1e340c38ba/fe/src/main/java/org/apache/doris/catalog/Catalog.java#L5134-L5148

PR #3359, opened 2020-04-20
[Bug][Dynamic Partition]Fix Bug that dynamic partition properties is not consistent

WingsGo:
This CL mainly fix the bug that when add dynamic partition properties in already exist table, the properties will only exist in Master FE.


WingsGo:
For #3358

morningman (approved):
LGTM

Issue #2299, opened 2019-11-26
Alter replication number for table


imay:
**Is your feature request related to a problem? Please describe.**

Now if I want to change the replication number of a table, I can only alter partition one by one. This is a nightmare for me.

**Describe the solution you'd like**

What I want is that I can set the entire table's replication number in one command like 

```
alter table myTable set property ("replication_num" = "3")
```

PR #3360, opened 2020-04-20
Alter replication number for table in one command

xy720:
#2299 #3389 
This cl add new command to set replication number of table in one time.

```
alter table test_tbl set ("replication_num" = "3");
```

It changes replication num of a table that have only one partition.

and

```
alter table test_tbl set ("default.replication_num" = "3");
```

It changes default replication num of the specified table.

### DIFF
--- a/fe/src/main/java/org/apache/doris/catalog/Catalog.java
+++ b/fe/src/main/java/org/apache/doris/catalog/Catalog.java
@@ -5163,29 +5163,34 @@ public class Catalog {
     // The caller need to hold the db write lock
     public void modifyTableReplicationNum(Database db, OlapTable table, Map<String, String> properties) throws DdlException {
         Preconditions.checkArgument(db.isWriteLockHeldByCurrentThread());
-        short newReplicationNum = (short) -1;
-        try {
-            newReplicationNum = Short.valueOf(properties.get(PropertyAnalyzer.PROPERTIES_REPLICATION_NUM));
-            if (newReplicationNum <= 0) {
-                throw new AnalysisException("Replication num should larger than 0. (suggested 3)");
-            }
-        } catch (Exception e) {
-            throw new DdlException(e.getMessage());
+        String defaultReplicationNumName = "default."+ PropertyAnalyzer.PROPERTIES_REPLICATION_NUM;
+        PartitionInfo partitionInfo = table.getPartitionInfo();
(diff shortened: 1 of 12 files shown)


fe/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -5177,8 +5177,31 @@ public void replayModifyTableProperty(short opCode, ModifyTablePropertyOperation
     }
 
     // The caller need to hold the db write lock
-    public void modifyTableReplicationNum(Database db, OlapTable table, Map<String, String> properties) {
+    public void modifyTableReplicationNum(Database db, OlapTable table, Map<String, String> properties) throws DdlException {
         Preconditions.checkArgument(db.isWriteLockHeldByCurrentThread());
+        short newReplicationNum = (short) -1;
+        try {
+            newReplicationNum = Short.valueOf(properties.get(PropertyAnalyzer.PROPERTIES_REPLICATION_NUM));
```
morningman, line 5184: This should be done in `analyze()` of `ModifyTablePropertyClause.java`.
And you can call `PropertyAnalyzer.analyzeReplicationNum()`.

morningman (changes requested):
`alter table test_tbl set ("default.replication_num" = "3");` is an already exist statement, which is used to set the "default" replication num of a table, but not set the real replication number of each partition. So you'd better not changing its behavior.

fe/src/main/java/org/apache/doris/catalog/Catalog.java
xy720: ok.

fe/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -5176,8 +5176,50 @@ public void replayModifyTableProperty(short opCode, ModifyTablePropertyOperation
         }
     }
 
+    /**
+     * Set replication number for a table that only have one partition.
```
morningman, line 5180:
```suggestion
     * Set replication number for unpartitioned table
```

morningman (changes requested):
Please update the document of alter table

morningman (commented):
Add unit test for this

morningman (dismissed):
LGTM

morningman (approved):
LGTM

PR #3363, opened 2020-04-21
[Bug]Use OlapTableSink::close to replace OlapTableSink::finalize method to avoid OOM

### LINKED ISSUE #3362: [Bug]OlapTableSink.finalize may use a another method name to avoid GC thread execute?
I found that in OlapTableSink.java, there is a function named `finalize`. In Java, it will be called when GC destory the object, but I found that it will cause OOM. Because if we not define `finalize` method, the object will be recycle imediatelly, but if we define it, the object will insert into  `java.lang.ref.Finalizer.ReferenceQueue`, if we have large object in the queue, the fe will down cause by OOM.

https://github.com/apache/incubator-doris/blob/c6ac60bab9563a77b052a2766b0de859528aee41/fe/src/main/java/org/apache/doris/planner/OlapTableSink.java#L121-L137

Snip20200421_14

WingsGo:
This CL mainly solve the problem that when recycle `OlapTableSink`
object, GC thread will not do it immediately because the class override
the `finalize` method, and it will cause OOM.

### DIFF
--- a/fe/src/main/java/org/apache/doris/planner/OlapTableSink.java
+++ b/fe/src/main/java/org/apache/doris/planner/OlapTableSink.java
@@ -118,7 +118,7 @@ public class OlapTableSink extends DataSink {
     }
 
     // must called after tupleDescriptor is computed
-    public void close() throws UserException {
+    public void completeq() throws UserException {
         TOlapTableSink tSink = tDataSink.getOlap_table_sink();
 
         tSink.setTable_id(dstTable.getId());
--- a/fe/src/main/java/org/apache/doris/planner/OlapTableSink.java
+++ b/fe/src/main/java/org/apache/doris/planner/OlapTableSink.java
@@ -118,7 +118,7 @@ public class OlapTableSink extends DataSink {
     }
 
     // must called after tupleDescriptor is computed
-    public void completeq() throws UserException {
+    public void complete() throws UserException {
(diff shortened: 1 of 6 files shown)


WingsGo:
For #3362

kangkaisen:
@WingsGo Hi, would better choose a name could describe the `finalize` method work. current work is not `close`.

WingsGo:
@kangkaisen what about `terminate` or `finish`?

morningman:
> @kangkaisen what about `terminate` or `finish`?

How about `complete()`? Which means the end of "handle" this object.

But I am curious why I didn't encounter this kind of OOM in our environment?

And I found that `InsertStmt` also has a `finalize()` method, can you modify it too?

fe/src/main/java/org/apache/doris/planner/OlapTableSink.java
```diff
@@ -118,7 +118,7 @@ public void updateLoadId(TUniqueId newLoadId) {
     }
 
     // must called after tupleDescriptor is computed
-    public void finalize() throws UserException {
+    public void completeq() throws UserException {
```
kangkaisen, line 121:
```suggestion
    public void complete() throws UserException {
```

kangkaisen (approved):
+1, LGTM

PR #3366, opened 2020-04-21
Modify test to be compative with cov tool

jmk1011:
C ++ R syntax is not compatible with coverage tools, so modify the syntax for test case


morningman (approved):
LGTM

Issue #3364, opened 2020-04-21
[Bug] Authorization missing when auditloader plugin redirect stream load


wyb:
**Describe the bug**
Auditloader plugin load fe auditlog into doris failed.
 
fe.log
2020-04-21 11:52:07,544 INFO 36 [DorisStreamLoader.loadBatch():119] AuditLoader plugin load with label: audit_20200321_115207, response code: 200, msg: OK, content: {    "TxnId": -1,    "Label"
: "audit_20200321_115207",    "Status": "Fail",    "Message": "no valid Basic authorization",    "NumberTotalRows": 0,    "NumberLoadedRows": 0,    "NumberFilteredRows": 0,    "NumberUnselected
Rows": 0,    "LoadBytes": 0,    "LoadTimeMs": 0}

PR #3367, opened 2020-04-21
[Bug] Fix authorization missing when auditloader plugin redirect stream load bug

wyb:
HttpURLConnection can automatically redirect stream load to BE, but there is no authorization information in http request headers after redirect.
Maybe HttpURLConnection remove authorization info when do followredirect.
The solution is set the followRedirect property to false on the connection object and do the redirect request manually.
#3364


morningman (approved):
LGTM

PR #3369, opened 2020-04-21
Support txn management in db level isolation and use ArrayDeque to improve txn task performance

### LINKED ISSUE #3368: [Proposal] Making doris stream load more robust with higher concurrent performance
The main purpose of this proposal is to make doris stream load more robust with higher concurrent peformance.
Now, in our production environment, we mainly use stream load to load business data into doris.

And we found the following problem:
  1. the transaction processing in GlobalTransactionMgr performance is poor, because the transaction lock is global without any isolation, for some operation we need to iterate the whole map.
  2. Fe thrift server use Executors.newFixedThreadPool and Executors.newCachedThreadPool to construct thread pool, which may cause oom or too many thread to be created and then crash
  3. the rpc cache client num is not limited in be and when rpc client timeout, the request task is still blocked in fe, the client encounter rpc_timeout and retry,  it may cause disaster for fe sometimes.
  4. the tablet writer open cost too much time, because when user doesn't specify the partition, the all tablet writer would be opened.

Implementation
So the following work we need to implement:
  1. Support db level isolation for transaction processing and use ArrayDeque to stored the finished transaction
  2. Remove all code that used newFixedThreadPool or newCachedThreadPool, construct the thread pool explicitly, support thread pool metric monitor 
  3. use client pool to manage rpc client, set the core and max cache client num, use try lock instead of lock to support server task be can be canceled in time due to timeout
  4. open the tablet writer when needed not open all when partitions of table are not specified.

caiconghui:
This PR is the first step to make doris stream load more robust with higher concurrent performance(#3368)，the main work is to support txn management in db level isolation，and use ArrayDeque to stored final status txns


fe/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
```diff
@@ -0,0 +1,556 @@
...
+                }
+
+            }
+        } finally {
+            writeUnlock();
+        }
+    }
+
+    public int getTransactionNum() {
+        return idToRunningTransactionState.size() + finalStatusTransactionStateDeque.size();
```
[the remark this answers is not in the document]
caiconghui: I find it is only invoked by checkpoint thread, so it is ok for not thread safe, maybe there need some mement to indicate it

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -500,8 +485,15 @@ public boolean commitAndPublishTransaction(Database db, long transactionId,
         } finally {
             db.writeUnlock();
         }
-        
-        TransactionState transactionState = idToTransactionState.get(transactionId);
+        DatabaseTransactionMgr dbTransactionMgr = getDatabaseTransactioMgr(db.getId());
+        TransactionState transactionState = null;
+        dbTransactionMgr.readLock();
```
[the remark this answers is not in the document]
caiconghui: it is only exposed to GlobalTransactionMgr, we still use GlobalTransactionMgr to do some txn management work in global view, DatabaeTransactionMgr cannot be an independent existence without GlobalTransactionMgr. it is normal for  GlobalTransactionMgr to invoke some DatabaeTransactionMgr's methods and ensure they are 'atomatic' by lock, while DatabaseTransactionMgr only need to provide some basic methods about txn. If we not expose dbTransactionMgr's lock to GlobalTransactionMgr, the work for DatabaeTransactionMgr is too heavy.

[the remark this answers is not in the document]
morningman: In fact, you will find that all the logic in `GlobalTransactionMgr` is now executed in `DbTransactionMgr`. `GlobalTransactionMgr` has only become a class for managing `DbTransactionMgr`.

The usage patterns of most methods are:

```
DbTransactionMgr dbMgr = getDbTransactionMgr(dbId);
dbMgr.lock()
try {
    dbMgr.doSomeTxnThing();
} finally {
    dbMgr.unlock();
}
```

Therefore, I think it is more appropriate to move all the main implementation logic into `DbTransactionMgr`. In this way, the responsibilities of `GlobalTransactionMgr` and `DbTransactionMgr` will be clearer. `GlobalTransactionMgr` only serves as the entry class for transaction operations and is responsible for managing `DbTransactionMgr`. And `DbTransactionMgr` is the actual transaction operation class.

morningman (changes requested):

morningman:
@caiconghui I have some comments left. And could you please resolve the conflicts?

fe/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -2639,6 +2641,7 @@ public void dropDb(DropDbStmt stmt) throws DdlException {
             fullNameToDb.remove(db.getFullName());
             final Cluster cluster = nameToCluster.get(db.getClusterName());
             cluster.removeDb(dbName, db.getId());
+            globalTransactionMgr.removeDatabaseTransactionMgr(db.getId());
```
morningman, line 2644: Actually, the operation `DropDb` is just put database to CatalogRecycleBin, not actually drop it. And the database can be recovered by `Recover` operation.
So here you should not `removeDatabaseTransactionMgr()`, Instead, it should be called in `CatalogRecycleBin.eraseDatabase()`

fe/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -2686,6 +2689,7 @@ public void replayDropDb(String dbName) throws DdlException {
             idToDb.remove(db.getId());
             final Cluster cluster = nameToCluster.get(db.getClusterName());
             cluster.removeDb(dbName, db.getId());
+            globalTransactionMgr.removeDatabaseTransactionMgr(db.getId());
```
morningman, line 2692: Same to `dropDb`, this should be called in `CatalogRecycleBin.replayEraseDatabase()`

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -123,6 +70,22 @@ public TxnStateCallbackFactory getCallbackFactory() {
         return callbackFactory;
     }
 
+    public DatabaseTransactionMgr getDatabaseTransactioMgr(long dbId) throws AnalysisException {
```
morningman, line 73:
```suggestion
    public DatabaseTransactionMgr getDatabaseTransactionMgr(long dbId) throws AnalysisException {
```

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -123,6 +70,22 @@ public TxnStateCallbackFactory getCallbackFactory() {
...
+        DatabaseTransactionMgr dbTransactionMgr = dbIdToDatabaseTransactionMgrs.get(dbId);
+        if (dbTransactionMgr == null) {
+            throw new AnalysisException("databaseTransactionMgr[" + dbId + "] does not exist");
```
morningman, line 76: AnalysisException is not suitable here.
But it can be modified next time

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -156,114 +119,20 @@ public long beginTransaction(long dbId, List<Long> tableIdList, String label, TU
...
-            }
-            replayDeleteTransactionState(state);
-            editLog.logDeleteTransactionState(state);
-        } finally {
-            writeUnlock();
+            DatabaseTransactionMgr dbTransactionMgr = getDatabaseTransactioMgr(dbId);
+            return dbTransactionMgr.getLabelState(label);
+        } catch (AnalysisException e) {
+            LOG.warn("Get transaction status by label " + label + " failed", e);
+            return null;
```
morningman, line 133:
```suggestion
            return TransactionStatus.UNKNOWN;
```

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -851,493 +243,114 @@ public boolean isIntersectionNotEmpty(List<Long> sourceTableIdList, List<Long> t
...
-                    // remove the txn which labels are expired
-                    expiredTxns.add(transactionState.getTransactionId());
-                } else if (transactionState.isTimeout(currentMillis)) {
-                    // txn is running but timeout, abort it.
-                    timeoutTxns.add(transactionState.getTransactionId());
+        for (DatabaseTransactionMgr dbTransactionMgr : dbIdToDatabaseTransactionMgrs.values()) {
+            dbTransactionMgr.removeExpiredTxns();
+            List<Long> timeoutTxns = dbTransactionMgr.getTimeoutTxns(currentMillis);
+            // abort timeout txns
+            for (Long txnId : timeoutTxns) {
```
morningman, line 250: This logic(Remove the timeout txn) can also be put into `DatabaseTransactionMgr`.
`dbTransactionMgr. removeExpiredAndTimeoutTxns();`

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -851,493 +243,114 @@ public boolean isIntersectionNotEmpty(List<Long> sourceTableIdList, List<Long> t
...
 
-    public TransactionState getTransactionState(long transactionId) {
-        readLock();
-        try {
-            return idToTransactionState.get(transactionId);
-        } finally {
-            readUnlock();
-        }
+    public TransactionState getTransactionState(long dbId, long transactionId) {
+        DatabaseTransactionMgr dbTransactionMgr = dbIdToDatabaseTransactionMgrs.get(dbId);
```
morningman, line 264: No throw exception? `dbTransactionMgr` could be `null`

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -851,493 +243,114 @@ public boolean isIntersectionNotEmpty(List<Long> sourceTableIdList, List<Long> t
...
-            idToTransactionState.put(transactionState.getTransactionId(), transactionState);
-            updateTxnLabels(transactionState);
-            updateDbRunningTxnNum(preTxnState == null ? null : preTxnState.getTransactionStatus(),
-                                  transactionState);
-        } finally {
-            writeUnlock();
+            DatabaseTransactionMgr dbTransactionMgr = getDatabaseTransactioMgr(transactionState.getDbId());
+            dbTransactionMgr.replayUpsertTransactionState(transactionState);
+        } catch (AnalysisException e) {
+            LOG.warn("replay upsert transaction failed", e);
```
morningman, line 279: add transaction's id in log, for easy debugging.

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -851,493 +243,114 @@ public boolean isIntersectionNotEmpty(List<Long> sourceTableIdList, List<Long> t
...
-                // send clone task with committed version hash X, and than Master changed, the new Master FE
-                // received the clone task report with version hash X, which not equals to it own committed
-                // version hash, than the clone task is failed.
-                partition.setNextVersionHash(Util.generateVersionHash() /* next version hash */,
-                                             partitionCommitInfo.getVersionHash() /* committed version hash*/);
-            }
+            DatabaseTransactionMgr dbTransactionMgr = getDatabaseTransactioMgr(transactionState.getDbId());
+            dbTransactionMgr.deleteTransaction(transactionState);
+        } catch (AnalysisException e) {
+            LOG.warn("replay delete transaction failed", e);
```
morningman, line 289: Add txn id in log.

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -851,493 +243,114 @@ public boolean isIntersectionNotEmpty(List<Long> sourceTableIdList, List<Long> t
...
-            for (long dbId : dbIds) {
-                List<Comparable> info = new ArrayList<Comparable>();
-                info.add(dbId);
-                Database db = Catalog.getInstance().getDb(dbId);
-                if (db == null) {
-                    continue;
-                }
-                info.add(db.getFullName());
-                infos.add(info);
+        List<Long> dbIds = Lists.newArrayList();
```
morningman, line 295:
```suggestion
        List<Long> dbIds = Lists.newArrayList(dbIdToDatabaseTransactionMgrs.keySet());
```

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -851,493 +243,114 @@ public boolean isIntersectionNotEmpty(List<Long> sourceTableIdList, List<Long> t
...
-    
+
+    /**
+     * It is a non thread safe method, only invoked by checkpoint thread without any lock or image dump thread with db lock
+     */
     public int getTransactionNum() {
-        return this.idToTransactionState.size();
+        int txnNum = 0;
+        for (DatabaseTransactionMgr dbTransactionMgr : dbIdToDatabaseTransactionMgrs.values()) {
+            txnNum = txnNum + dbTransactionMgr.getTransactionNum();
```
morningman, line 351:
```suggestion
            txnNum += dbTransactionMgr.getTransactionNum();
```

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -1346,81 +359,71 @@ public TransactionIdGenerator getTransactionIDGenerator() {
...
+    public List<Pair<Long, Long>> getTransactionIdByCoordinateBe(String coordinateHost, int limit) {
+        ArrayList<Pair<Long, Long>> txnInfos = new ArrayList<>();
+        for (DatabaseTransactionMgr databaseTransactionMgr : dbIdToDatabaseTransactionMgrs.values()) {
+            txnInfos.addAll(databaseTransactionMgr.getTransactionIdByCoordinateBe(coordinateHost, limit));
+            if (txnInfos.size() > limit) {
+                break;
+            }
         }
-        return txnIds;
+        return txnInfos.size() > limit ? new ArrayList<>(txnInfos.subList(0, limit)) : txnInfos;
```
morningman, line 414:
```suggestion
        return txnInfos.size() > limit ? txnInfos.subList(0, limit) : txnInfos;
```

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -17,69 +17,33 @@
 
 package org.apache.doris.transaction;
 
+import org.apache.commons.lang3.tuple.Pair;
```
morningman, line 20: You can use `org.apache.doris.common.Pair`

morningman (changes requested):

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
caiconghui: @morningman  subList is a view of origin list, so is it a good way to directly return sublist?

morningman: I think it's OK, cause no one will modify that list.

morningman (approved):
LGTM

PR #3374, opened 2020-04-22
[Bug] `isQuery` field should be reset at the beginning of query execution

### LINKED ISSUE #3380: [Bug] White list appear some resolved IPs which is not belong to this user.
**Describe the bug**

View the page in web browser:
`fe:http_port/system?path=//auth/%27default_cluster:user%27@%27%25%27`

you may see some IPs in `resolved IPs` entry. But those IPs is not belong to the user.

morningman:
If not reset, all queries comes from same session will have save `isQuery` field value.
This bug will cause all entries in `fe.audit.log` has same `IsQuery=true`.

This CL also fix another bug:
The resolved IPs of domain of a user should not appear in other user's white list. Fix #3380

### DIFF
--- a/docs/documentation/cn/administrator-guide/segment-v2-usage.md
+++ b/docs/documentation/cn/administrator-guide/segment-v2-usage.md
@@ -62,7 +62,7 @@ V2 格式的表可以支持以下新的特性：
     
     ```
     ## 创建 V2 格式的 Rollup
-    
+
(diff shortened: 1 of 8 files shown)


chaoyli (dismissed):

kangkaisen:
This PR fix three issue :sweat:

kangkaisen (approved):
+1. LGTM

PR #3378, opened 2020-04-22
Forbidden correlated having clause

### LINKED ISSUE #3377: The incorrect result of having clause with subquery
**Describe the bug**
1. SELECT k1 FROM test GROUP BY k1 HAVING EXISTS(SELECT k1 FROM baseall GROUP BY k1 HAVING SUM(test.k1) = k1);
Expect : empty when having predicate is not satisfied.
See error: return the result of select stmt without having predicate

2. select k1, count(*) cnt from test group by k1 having k1 in (select k1 from baseall order by k1 limit 2) order by k1 limit 5 offset 3;
Expect: empty
See error: return the result of select stmt without offset.

3. create view tt as select k1, count(*) cnt from test group by k1 having k1 in (select k1 from baseall order by k1 limit 2) order by k1;
Expect: success
ERROR 1064 (HY000): errCode = 2, detailMessage = failed to init view stmt

EmmyMiao87:
Fixed #3377

The correlated slot ref should be bound by the agg tuple of outer query.
However, the correlated having clause should not be analyzed correctly so the result is incorrect.
For example: 
```
SELECT k1 FROM test GROUP BY k1 HAVING EXISTS(SELECT k1 FROM baseall GROUP BY k1 HAVING SUM(test.k1) = k1);
```
The correlated predicate is not executed.

The limit offset should be rewritten also when there is subquery in having clause.
For example: 
```
select k1, count(*) cnt from test group by k1 having k1 in (select k1 from baseall order by k1 limit 2) order by k1 limit 5 offset 3;
```
The new stmt should has a limit element with offset.

The view which has subquery in having clause should not be created.
The reason is that the toSql function of query stmt is incorrect when there is subquery in having clause.
So the view def sql is incorrect and could not pass the sql parser.
Now, the view def sql use the original sql directly to fix this error.

Change-Id: If24206a8dee94013225fe7fdd30ab1fb034de34f

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/QueryStmt.java
+++ b/fe/src/main/java/org/apache/doris/analysis/QueryStmt.java
@@ -437,6 +437,8 @@ public abstract class QueryStmt extends StatementBase {
      */
     public abstract void collectTableRefs(List<TableRef> tblRefs);
 
+    abstract List<TupleId> collectTupleIds();
+
     public ArrayList<OrderByElement> getOrderByElements() {
         return orderByElements;
     }
--- a/fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
+++ b/fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
@@ -495,6 +495,20 @@ public class SelectStmt extends QueryStmt {
         return result;
     }
 
+    @Override
+    public List<TupleId> collectTupleIds() {
+        List<TupleId> result = Lists.newArrayList();
+        resultExprs.stream().forEach(expr -> expr.getIds(result, null));
+        result.addAll(getTableRefIds());
+        if (whereClause != null) {
+            whereClause.getIds(result, null);
+        }
+        if (havingClauseAfterAnaylzed != null) {
+            havingClauseAfterAnaylzed.getIds(result, null);
+        }
+        return result;
+    }
+
     private void whereClauseRewrite() {
         Expr deDuplicatedWhere = deduplicateOrs(whereClause);
         if (deDuplicatedWhere != null) {
@@ -1077,6 +1091,18 @@ public class SelectStmt extends QueryStmt {
             LOG.debug("desctbl: " + analyzer.getDescTbl().debugString());
             LOG.debug("resultexprs: " + Expr.debugString(resultExprs));
         }
+
+        if (havingClauseAfterAnaylzed != null) {
+            // forbidden correlated subquery in having clause
+            List<Subquery> subqueryInHaving = Lists.newArrayList();
+            havingClauseAfterAnaylzed.collect(Subquery.class, subqueryInHaving);
+            for (Subquery subquery : subqueryInHaving) {
+                if (subquery.isCorrelatedPredicate(getTableRefIds())) {
+                    throw new AnalysisException("The correlated having clause is not supported");
+                }
+            }
+        }
+
         /*
          * All of columns of result and having clause are replaced by new slot ref which is bound by top tuple of agg info.
          * For example:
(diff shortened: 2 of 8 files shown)


fe/src/main/java/org/apache/doris/analysis/BaseViewStmt.java
```diff
@@ -109,33 +112,37 @@ protected void createColumnAndViewDefs(Analyzer analyzer) throws AnalysisExcepti
...
+        for (int i = 0; i < colNames.size(); ++i) {
+            if (i != 0) {
+                sb.append(", ");
+            }
+            String colRef = viewDefStmt.getColLabels().get(i);
+            if (!colRef.startsWith("`")) {
+                colRef = "`" + colRef + "`";
+            }
+            String colAlias = colNames.get(i);
+            sb.append(String.format("`%s`.%s AS `%s`", tableName.getTbl(), colRef, colAlias));
```
xy720, line 141: It is work for function call?
For example, if user input:
"create view test_view (h1, h2) as select function(a, b), col2 from testTbl"
will become
"select function(a, b) as h1, col2 as h2 from (select function(a, b), col2 from testTbl)"

fe/src/main/java/org/apache/doris/qe/ConnectProcessor.java
```diff
@@ -225,7 +226,13 @@ private void handleQuery() {
         SqlScanner input = new SqlScanner(new StringReader(originStmt), ctx.getSessionVariable().getSqlMode());
         SqlParser parser = new SqlParser(input);
         try {
-            return SqlParserUtils.getMultiStmts(parser);
+            List<StatementBase> result = SqlParserUtils.getMultiStmts(parser);
+            String[] originStmtList = originStmt.split(";");
```
morningman, line 230: There maybe `;` inside the query statement

morningman (changes requested):

fe/src/main/java/org/apache/doris/analysis/LimitElement.java
```diff
@@ -42,9 +42,9 @@ public LimitElement(long limit) {
         offset = 0;
     }
 
-    public LimitElement(long offset, long limit) {
-        this.offset = offset;
+    public LimitElement(long limit, long offset) {
```
kangkaisen, line 45: Why do we change the order?

fe/src/main/java/org/apache/doris/analysis/SelectList.java
```diff
@@ -98,6 +98,24 @@ public void rewriteExprs(ExprRewriter rewriter, Analyzer analyzer)
...
+
+    public String toOriginSql() {
+        // Select list
+        StringBuilder strBuilder = new StringBuilder();
+        strBuilder.append("SELECT ");
+        if (isDistinct) {
+            strBuilder.append("DISTINCT ");
+        }
+        for (int i = 0; i < items.size(); ++i) {
+            // strBuilder.append(selectList.getItems().get(i).toSql());
```
kangkaisen, line 110: Remove.

kangkaisen:
`The reason is that the toSql function of query stmt is incorrect when there is subquery in having clause.`

@EmmyMiao87  Hi, why can't we make the toSql function correct ?

kangkaisen:
> `The reason is that the toSql function of query stmt is incorrect when there is subquery in having clause.`
> 
> @EmmyMiao87 Hi, why can't we make the toSql function correct ?

OK, I see.

fe/src/main/java/org/apache/doris/analysis/BaseViewStmt.java
EmmyMiao87: This change has been removed.

fe/src/main/java/org/apache/doris/analysis/LimitElement.java
EmmyMiao87: changed

morningman (approved):
LGTM

PR #3381, opened 2020-04-22
Fix expect may produce incorrect values

### LINKED ISSUE #3379: except operater output may incorrect
except operater  may produce incorrect value；
like
```

mysql> select k1 from baseall except select 2;
+------+
| k1   |
+------+
|   14 |
|    9 |
|    7 |
|    2 |
|   12 |
|   11 |
|    5 |
|   15 |
|    1 |
|    6 |
|    8 |
|   13 |
|    3 |
|    4 |
|   10 |
+------+
15 rows in set (0.20 sec)
```

the value 2 sometimes not filtered

yangzhg:
Fixes #3379 
in except node build the init hash table， the build expr and the probe expr context are the same， this casuse， hash buffer the expr in the same memory， this will cause the row has same hash value will be equal in RawValue::eq

### DIFF
--- a/be/src/exec/except_node.cpp
+++ b/be/src/exec/except_node.cpp
@@ -46,6 +46,7 @@ Status ExceptNode::open(RuntimeState* state) {
         _hash_tbl_iterator = _hash_tbl->begin();
         return Status::OK();
     }
+    bool eos = false;
 
     for (int i = 1; i < _children.size(); ++i) {
         // rebuid hash table, for first time will rebuild with the no duplicated _hash_tbl,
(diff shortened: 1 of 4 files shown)


morningman (approved):
LGTM

kangkaisen:
@yangzhg `open` method seems not change. which line code fix this issue. 

**When you submit a issue or PR, would better describe the root cause for the issue and the core idea for your solution, thanks**

yangzhg:
> describe
done

kangkaisen:
The bug cause by 
```
    _hash_tbl.reset(new HashTable(_child_expr_lists[0], _child_expr_lists[0], _build_tuple_size,
                                  true, _find_nulls, id(), mem_tracker(), 1024));
```
should be
```
    _hash_tbl.reset(new HashTable(_child_expr_lists[0], _child_expr_lists[1], _build_tuple_size,
                                  true, _find_nulls, id(), mem_tracker(), 1024));
```

kangkaisen (approved):
+1,LGTM

Issue #3365, opened 2020-04-21
[Proposal] Make running profile clearer and more intuitive to improve usability


HappenLee:
# Motivation
The main purpose of this proposal is to make Doris more user-friendly and easy to use. Currently, Doris will print the running profile information of fragment in the log after the execution of fragment. This information is an important reference for Doris performance tuning and debugging. However, many of the information recorded in the current running profile is invalid, wrong or even misleading. We need to sort out the information of each counter in the current record, remove the invalid counter and add the missing but important counter. Write a document which guides user to monitor Doris and troubleshoot problems by running profile and add a reasonable counter to solve problems when appropriate.

# Implementation

So we need to do the following

* We need to eliminate invalid code and invalid monitoring items to simplify the current running profile information

* Fix which wrong and misleading monitoring items

* Add the important monitoring items but missing now

* Sort out the contents of each monitoring item and provide best practice or troubleshooting tutorials as documents for user

* Provide a document which guide user how to add monitoring items when it is necessary.

kangkaisen:
@HappenLee Good job! look forward to your PR.

PR #3383, opened 2020-04-23
Make running profile clearer and more intuitive to improve usability (#3365)

HappenLee:
This CL mainly made the following modifications:
1. Delete Invalid method in Running Profile Class.
2. Move Memlimit Counter from blockmgr to frgment and add PeakMemUsage Counter
3. Fix the bug of buffer pool memlimit counter
4. Call compute_time_in_profile() before pretty_print() to show the _local_time_percent without child running profile
5. Add TransferThread ThreadToken count in AveThreadToken Counter

ISSUE: #3365

### DIFF
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -1038,6 +1038,7 @@ Status OlapScanNode::normalize_binary_predicate(SlotDescriptor* slot, ColumnValu
 
 void OlapScanNode::transfer_thread(RuntimeState* state) {
     // scanner open pushdown to scanThread
+    state->resource_pool()->acquire_thread_token();
     Status status = Status::OK();
     for (auto scanner : _olap_scanners) {
         status = Expr::clone_if_not_exists(_conjunct_ctxs, state, scanner->conjunct_ctxs());
@@ -1171,6 +1172,7 @@ void OlapScanNode::transfer_thread(RuntimeState* state) {
         }
     }
 
+    state->resource_pool()->release_thread_token(true);
     VLOG(1) << "TransferThread finish.";
     boost::unique_lock<boost::mutex> l(_row_batches_lock);
     _transfer_done = true;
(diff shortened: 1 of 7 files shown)


be/src/exec/olap_scan_node.cpp
```diff
@@ -1038,6 +1038,7 @@ Status OlapScanNode::normalize_binary_predicate(SlotDescriptor* slot, ColumnValu
 
 void OlapScanNode::transfer_thread(RuntimeState* state) {
     // scanner open pushdown to scanThread
+    state->resource_pool()->acquire_thread_token();
```
imay, line 1041: Can you tell the side effect of this operation? Will this thread hang if there is not enough thread resource?

be/src/runtime/runtime_state.cpp
```diff
@@ -226,8 +226,12 @@ Status RuntimeState::init_mem_trackers(const TUniqueId& query_id) {
 
     // _query_mem_tracker = MemTracker::get_query_mem_tracker(
     //         query_id, bytes_limit, _exec_env->process_mem_tracker());
+
+    auto mem_tracker_counter = ADD_COUNTER(&_profile, "MemoryLimit", TUnit::BYTES);
+    mem_tracker_counter->set(bytes_limit);
+
     _query_mem_tracker.reset(
-            new MemTracker(bytes_limit, runtime_profile()->name(), _exec_env->process_mem_tracker()));
+            new MemTracker(&_profile, bytes_limit, runtime_profile()->name(), _exec_env->process_mem_tracker()));
```
imay, line 234: If you add profile for this memory tracker, why not `_instance_mem_tracker`?

be/src/exec/olap_scan_node.cpp
HappenLee: Add this operation just to Make AverageThreadTokens more accurate
   - AverageThreadTokens: 1.00  

Do not warry thread hang on this operation。
acquire_thread_token() just do a lightweight add operation without doing the really thread assignment 

inline void ThreadResourceMgr::ResourcePool::acquire_thread_token() {
    __sync_fetch_and_add(&_num_threads, 1);
}

@imay

be/src/runtime/runtime_state.cpp
HappenLee: yes，I check code。the ```_instance_mem_tracker``` is more appropriate

morningman (approved):
LGTM

PR #3391, opened 2020-04-24
[Load] Add more info in SHOW LOAD result

### LINKED ISSUE #3390: [Load] Not easy to track the progress of broker load job

In the operation and maintenance process, we will find that some broker load jobs will be stuck for some reason. This is usually because a sub-task of the job is stuck on a BE, but the information displayed through `show load` cannot directly see which BE has a problem.

So I suggest adding more information to show load to facilitate viewing task running status and bug debugging.

morningman:
Fix #3390
This CL add more info in `JobDetails` column of `SHOW LOAD` result for Broker Load Job.

For example:
```
{
	"Unfinished backends": {
		"9c3441027ff948a0-8287923329a2b6a7": [10002]
	},
        "All backends": {
		"9c3441027ff948a0-8287923329a2b6a7": [10002, 10004, 10006]
	},
	"ScannedRows": 2390016,
	"TaskNumber": 1,
	"FileNumber": 1,
	"FileSize": 1073741824
}
```

2 newly added keys:

`Unfinished backends` indicates the BE which task on them are not finished.
`All backends` indicates the BE which this job has tasks on it.

One more thing, I pass the Backend Id along with the heartbeat msg from FE to BE, so that BE can
know the Id of themselves.

### DIFF
--- a/fe/src/main/java/org/apache/doris/load/loadv2/LoadJob.java
+++ b/fe/src/main/java/org/apache/doris/load/loadv2/LoadJob.java
@@ -33,6 +33,7 @@ import org.apache.doris.common.MetaNotFoundException;
 import org.apache.doris.common.UserException;
 import org.apache.doris.common.io.Text;
 import org.apache.doris.common.io.Writable;
+import org.apache.doris.common.util.DebugUtil;
 import org.apache.doris.common.util.LogBuilder;
 import org.apache.doris.common.util.LogKey;
 import org.apache.doris.common.util.TimeUtils;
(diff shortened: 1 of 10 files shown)


fe/src/main/java/org/apache/doris/load/loadv2/LoadJob.java
```diff
@@ -138,30 +139,41 @@
...
         }
 
         public synchronized void removeLoad(TUniqueId loadId) {
             counterTbl.rowMap().remove(loadId);
+            unfinishedBackendIds.remove(loadId);
+            allBackendIds.remove(loadId);
         }
 
-        public synchronized void updateLoad(TUniqueId loadId, TUniqueId fragmentId, long rows) {
+        public synchronized void updateLoadProgress(long beId, TUniqueId loadId, TUniqueId fragmentId, long rows,
```
wutiangan, line 169: Maybe change “beId” to ”backendId“ is better，at before you use unfinished“BackendIds”  and all”BackendIds“

fe/src/main/java/org/apache/doris/load/loadv2/LoadJob.java
```diff
@@ -138,30 +139,41 @@
...
+                boolean isDone) {
             if (counterTbl.contains(loadId, fragmentId)) {
                 counterTbl.put(loadId, fragmentId, rows);
             }
-        }
-
-        public synchronized void clearAllLoads() {
-            counterTbl.clear();
+            if (isDone && unfinishedBackendIds.containsKey(loadId)) {
+                unfinishedBackendIds.get(loadId).remove(beId);
```
wutiangan, line 175: can unfinishedBackendIds.get(loadId) be empty？

morningman: It won't, there is at least one BE for a load job

fe/src/main/java/org/apache/doris/load/loadv2/LoadJob.java
morningman: changed

EmmyMiao87 (approved):
LGTM

Issue #3395, opened 2020-04-25
Evaluate Case When Expr in Fe when possible


wangbo:
In the query case ```partition column compare with a case when Expr```, evaluate Case when Expr in Fe can void unnecessary partition scan.

See a example as below

SQL
```
select event_day,sum(pv) from table2  where 
event_day= case when date_format(now(),'%H%i') = 0 then 1 else 2 end group by event_day
```

BEFORE
```
 0:OlapScanNode                                                 |
|      TABLE: table2                                               |
|      PREAGGREGATION: ON                                          |
|      PREDICATES: `event_day` = CASE WHEN FALSE THEN 1 ELSE 2 END |
|      partitions=1/3                                              |
|      rollup: table2                                              |
|      buckets=10/10                                               |
|      cardinality=14                                              |
|      avgRowSize=723.5714                                         |
|      numNodes=1                                                  |
|      tuple ids: 0      
```

AFTER
```
   0:OlapScanNode                        |
|      TABLE: table2                      |
|      PREAGGREGATION: ON                 |
|      PREDICATES: `event_day` = 2.0      |
|      partitions=1/3                     |
|      rollup: table2                     |
|      buckets=10/10                      |
|      cardinality=14                     |
|      avgRowSize=723.5714                |
|      numNodes=1                         |
|      tuple ids: 0        
```

PR #3396, opened 2020-04-25
(#3395)calculate 'case when expr' when possible

wangbo:
(#3395)
calculate 'case when expr' when possible


morningman (commented):
Could you add a ut for this case? In QueryPlanTest.java

wangbo:
@morningman  ut added

fe/src/main/java/org/apache/doris/rewrite/FoldConstantsRule.java
```diff
@@ -48,6 +50,12 @@
 
     @Override
     public Expr apply(Expr expr, Analyzer analyzer) throws AnalysisException {
+        // evaluate case when expr
+        if (expr instanceof CaseExpr && expr.getChild(0) instanceof BoolLiteral) {
```
morningman, line 54: This is not right, the CaseExpr has multi child
1. optional case expr
2. optional list<when expr, then expr>
3. optional else expr.

The following SQL will return wrong result after applying your rewrite rule:
```
select case when false then 2 when true then 3 else 0 end;
```
Expect: 3, return 1.

morningman (changes requested):

wangbo:
major change 
1 support `case when and case xxx when`
2 support multi `case when`
3 add ut

Todo
1 fix clerical error in CaseExpr
2 make CaseExpr format better
I will fix the todo in another pr,because I'm not sure the fix cost

fe/src/test/java/org/apache/doris/planner/QueryPlanTest.java
```diff
@@ -504,4 +504,106 @@ public void testDateTypeEquality() throws Exception {
...
+
+    @Test
+    public void testConvertCaseWhenToConstant() throws Exception {
+        // basic test
+        String caseWhenSql = "select "
+                + "case when date_format(now(),'%H%i')  < 123 then 1 else 0 end as col "
+                + "from test.test1 "
+                + "where time = case when date_format(now(),'%H%i')  < 123 then date_format(date_sub(now(),2),'%Y%m%d') else date_format(date_sub(now(),1),'%Y%m%d') end";
+        SelectStmt selectStmt = getAnalyzedAndRewritedStmt(caseWhenSql);
+        Assert.assertTrue(!selectStmt.toSql().contains("CASE WHEN") && !selectStmt.toSql().contains("case when"));
```
kangkaisen, line 524: you could use `org.apache.commons.lang3.StringUtils.containsIgnoreCase`

fe/src/test/java/org/apache/doris/planner/QueryPlanTest.java
```diff
@@ -504,4 +504,106 @@ public void testDateTypeEquality() throws Exception {
         Catalog.getCurrentCatalog().getLoadManager().createLoadJobV1FromStmt(loadStmt, EtlJobType.HADOOP,
                 System.currentTimeMillis());
     }
+
+    private SelectStmt getAnalyzedAndRewritedStmt(String sql) throws Exception {
```
kangkaisen, line 508: Don't need to add this method. you could use `getSQLPlanOrErrorMsg` method.

fe/src/test/java/org/apache/doris/planner/QueryPlanTest.java
```diff
@@ -504,4 +504,106 @@ public void testDateTypeEquality() throws Exception {
...
+        SelectStmt selectStmt =
+                (SelectStmt) UtFrameUtils.parseAndAnalyzeStmt(sql, connectContext);
+        selectStmt = (SelectStmt) UtFrameUtils.rewriteStmt(selectStmt, connectContext);
+        selectStmt = (SelectStmt) UtFrameUtils.reAnalyze(selectStmt, connectContext);
+        return selectStmt;
+    }
+
+    @Test
+    public void testConvertCaseWhenToConstant() throws Exception {
```
kangkaisen, line 517: Please move this test to `ConstantExpressTest`

fe/src/main/java/org/apache/doris/analysis/CaseExpr.java
```diff
@@ -251,4 +257,72 @@ public void analyzeImpl(Analyzer analyzer) throws AnalysisException {
...
+            size--;
+        } else {
+            caseExpr = new BoolLiteral(true);
+        }
+
+        if (expr.hasElseExpr) {
+            size--;
+        }
+
+        //pre return when the `when expr` can't be converted to constants
```
kangkaisen, line 290:
```suggestion
        // Early return when the `when expr` can't be converted to constants
```

fe/src/main/java/org/apache/doris/analysis/CaseExpr.java
```diff
@@ -251,4 +257,72 @@ public void analyzeImpl(Analyzer analyzer) throws AnalysisException {
...
+
+    // this method just compare literal value and not completely consistent with be,for two cases
+    // 1 not deal float
+    // 2 just compare literal value with same type. for a example sql 'select case when 123 then '1' else '2' end as col'
+    //      for be will return '1', because be only regard 0 as false
+    //      but for current LiteralExpr.compareLiteral, `123`' won't be regard as true
+    //  the case which two values has different type left to be
+    public static Expr computeCaseExpr(CaseExpr expr) {
+        LiteralExpr caseExpr;
+        int startIndex = 0;
```
kangkaisen, line 269: would better rename `startIndex ` and `size `.

The `size` compute logic seems not consistent with BE ？

```
        int loop_start = has_case_expr() ? 1 : 0; \
        int loop_end = (has_else_expr()) ? num_children - 1 : num_children; \
```

kangkaisen (changes requested):

fe/src/test/java/org/apache/doris/planner/QueryPlanTest.java
wangbo: 👌

fe/src/test/java/org/apache/doris/planner/QueryPlanTest.java
wangbo: 👌

fe/src/main/java/org/apache/doris/analysis/CaseExpr.java
wangbo: 👌

fe/src/main/java/org/apache/doris/analysis/CaseExpr.java
wangbo: yes, be logic is right.I'll fix it

fe/src/test/java/org/apache/doris/planner/QueryPlanTest.java
wangbo: ```getSQLPlanOrErrorMsg ```  is unable to meet demand.What we want is to check whether the plan changes as our wish,so we need a  tools not only retwrite the plan but alse contains the customized check logic.It's obviously that plain sql is directly,but not a common general method,because the print format may change.
Maybe list a todo here?
Anyway,I think the result of ```getSQLPlanOrErrorMsg ```  can't meet demand

fe/src/test/java/org/apache/doris/planner/QueryPlanTest.java
wangbo: ```case when expr``` rewrite more complicate than ```convert  a function to constants```,it will change the query plan structure and dependency real doris table, so I think put UT in QueryPlanTest is better

fe/src/test/java/org/apache/doris/planner/QueryPlanTest.java
kangkaisen: I don't think so.
Directly check final FE sql plan is more accurate:
1. FE send BE is the sql plan
2. Your work do indeed is changing partial plan to constant.

> What we want is to check whether the plan changes as our wish

So you should use `getSQLPlanOrErrorMsg `

> I think the result of getSQLPlanOrErrorMsg can't meet demand

Why couldn't meet you demand? I think which could.

fe/src/test/java/org/apache/doris/utframe/UtFrameUtils.java
```diff
@@ -96,6 +100,34 @@ public static StatementBase parseAndAnalyzeStmt(String originStmt, ConnectContex
         return statementBase;
     }
 
+    public static StatementBase rewriteStmt(StatementBase stmt, ConnectContext ctx) throws AnalysisException {
```
kangkaisen, line 103: These two methods could also be removed?

kangkaisen (dismissed):
+1， LGTM

kangkaisen (dismissed):
+1, LGTM

wangbo:
resolve conflict

morningman (approved):
LGTM

PR #3399, opened 2020-04-27
Remove redunant call function _wait_in_flight_packet()

WingsGo:
The function `_wait_in_flight_packet` has been called in `_send_cur_batch`. No need to call twice.


morningman (approved):
LGTM

PR #3405, opened 2020-04-27
 [Profile] Make running profile clearer and more intuitive to improve usability

HappenLee:
This CL mainly made the following modifications:
    1. Delete Invalid MemoryUsed Counter and Add PeakMemUsage in each exec node and datastreamsender
    2. Add intent in child execnode profile，make it is easily to know the relationship between execnode
    3. Del _is_result_order we not support any more in olap_scan_node.h and olap_scan_node.cpp
    4. Add scan_disk method to olap_scanner to fix the counter _num_disks_accessed_counter
    5. Now we do not use buffer pool to read and write disk, so comment out readio counter and writeio counter code
ISSUE #3365

### DIFF
--- a/be/src/exec/exec_node.cpp
+++ b/be/src/exec/exec_node.cpp
@@ -172,15 +172,13 @@ Status ExecNode::prepare(RuntimeState* state) {
     DCHECK(_runtime_profile.get() != NULL);
     _rows_returned_counter =
         ADD_COUNTER(_runtime_profile, "RowsReturned", TUnit::UNIT);
-    _memory_used_counter =
-        ADD_COUNTER(_runtime_profile, "MemoryUsed", TUnit::BYTES);
     _rows_returned_rate = runtime_profile()->add_derived_counter(
                               ROW_THROUGHPUT_COUNTER, TUnit::UNIT_PER_SECOND,
                               boost::bind<int64_t>(&RuntimeProfile::units_per_second,
                                                    _rows_returned_counter,
                                                    runtime_profile()->total_time_counter()),
                               "");
-    _mem_tracker.reset(new MemTracker(-1, _runtime_profile->name(), state->instance_mem_tracker()));
+    _mem_tracker.reset(new MemTracker(_runtime_profile.get(), -1, _runtime_profile->name(), state->instance_mem_tracker()));
     _expr_mem_tracker.reset(new MemTracker(-1, "Exprs", _mem_tracker.get()));
     _expr_mem_pool.reset(new MemPool(_expr_mem_tracker.get()));
     // TODO chenhao
@@ -339,7 +337,7 @@ Status ExecNode::create_tree_helper(
     }
 
     if (!node->_children.empty()) {
-        node->runtime_profile()->add_child(node->_children[0]->runtime_profile(), false, NULL);
+        node->runtime_profile()->add_child(node->_children[0]->runtime_profile(), true, NULL);
     }
 
     return Status::OK();
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -694,10 +694,7 @@ Status OlapScanNode::start_scan_thread(RuntimeState* state) {
                 state, this, _olap_scan_node.is_preaggregation, _need_agg_finalize, *scan_range, scanner_ranges);
             _scanner_pool->add(scanner);
             _olap_scanners.push_back(scanner);
-
-            if (disk_set.find(scanner->scan_disk()) == disk_set.end()) {
-                disk_set.insert(scanner->scan_disk());
-            }
+	    disk_set.insert(scanner->scan_disk());
         }
     }
     COUNTER_SET(_num_disks_accessed_counter, static_cast<int64_t>(disk_set.size()));
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -72,16 +72,6 @@ Status OlapScanNode::init(const TPlanNode& tnode, RuntimeState* state) {
     RETURN_IF_ERROR(ExecNode::init(tnode, state));
     _direct_conjunct_size = _conjunct_ctxs.size();
 
-    if (tnode.olap_scan_node.__isset.sort_column) {
-        _is_result_order = true;
-    } else {
-        _is_result_order = false;
-    }
-
-    // Before, we support scan data ordered, but is not used in production
-    // Now, we drop this functional
-    DCHECK(!_is_result_order) << "ordered result don't support any more";
-
     return Status::OK();
 }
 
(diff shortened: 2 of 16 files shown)


be/src/exec/exec_node.cpp
```diff
@@ -339,7 +337,7 @@ Status ExecNode::create_tree_helper(
     }
 
     if (!node->_children.empty()) {
-        node->runtime_profile()->add_child(node->_children[0]->runtime_profile(), false, NULL);
```
morningman, line 342: It seem that the `false` here is to distinguish child[0] and other children.
Are you sure this is more readable? You can put the result screenshots here to explain.

be/src/exec/olap_scan_node.cpp
```diff
@@ -702,8 +694,13 @@ Status OlapScanNode::start_scan_thread(RuntimeState* state) {
                 state, this, _olap_scan_node.is_preaggregation, _need_agg_finalize, *scan_range, scanner_ranges);
             _scanner_pool->add(scanner);
             _olap_scanners.push_back(scanner);
+
+            if (disk_set.find(scanner->scan_disk()) == disk_set.end()) {
+                disk_set.insert(scanner->scan_disk());
```
morningman, line 699: `disk_set` is a set, so no need to `find and insert`, just `insert` is OK.

be/src/exec/olap_scan_node.cpp
```diff
@@ -702,8 +694,13 @@ Status OlapScanNode::start_scan_thread(RuntimeState* state) {
...
+            }
         }
     }
+    COUNTER_SET(_num_disks_accessed_counter, static_cast<int64_t>(disk_set.size()));
```
imay, line 703: What is this metric used for? If this metric is not useful, just remove it

be/src/exec/exec_node.cpp
HappenLee: after change the ```true```, we can eaily know olap_scan_node is child of aggregation_node. 
image

be/src/exec/olap_scan_node.cpp
HappenLee: yes，i will change the code。

be/src/exec/olap_scan_node.cpp
HappenLee: The metric help we now how many disk accessed in olap scan node scan data

imay: What can it be used for when you know how many disks accessed?

HappenLee: We can know whether the tablet of a query is allocated on different disks to know whether is IO bottleneck in disk

be/src/exec/exec_node.cpp
```diff
@@ -172,15 +172,13 @@ Status ExecNode::prepare(RuntimeState* state) {
     DCHECK(_runtime_profile.get() != NULL);
     _rows_returned_counter =
         ADD_COUNTER(_runtime_profile, "RowsReturned", TUnit::UNIT);
-    _memory_used_counter =
-        ADD_COUNTER(_runtime_profile, "MemoryUsed", TUnit::BYTES);
```
gaodayue, line 176: You'd better remove this field if it's not used. Btw, `_memory_used_counter` seems to be used by `HashJoinNode::close`, is it really ok to remove it?

be/src/runtime/bufferpool/buffer_pool_counters.h
```diff
@@ -35,19 +35,19 @@ struct BufferPoolClientCounters {
...
   /// Total bytes read from disk.
   RuntimeProfile::Counter* bytes_read;
 
   /// Amount of time spent waiting for writes to disk to complete.
-  RuntimeProfile::Counter* write_wait_time;
+  //RuntimeProfile::Counter* write_wait_time;
 
   /// Total number of write I/O operations issued.
-  RuntimeProfile::Counter* write_io_ops;
+  //RuntimeProfile::Counter* write_io_ops;
```
chaoyli, line 50: not used profile can be removed

be/src/runtime/bufferpool/buffer_pool.cc
```diff
@@ -392,11 +392,11 @@ BufferPool::Client::Client(BufferPool* pool, //TmpFileMgr::FileGroup* file_group
       ADD_COUNTER(child_profile, "CumulativeAllocations", TUnit::UNIT);
   counters_.cumulative_bytes_alloced =
       ADD_COUNTER(child_profile, "CumulativeAllocationBytes", TUnit::BYTES);
-  counters_.read_wait_time = ADD_TIMER(child_profile, "ReadIoWaitTime");
-  counters_.read_io_ops = ADD_COUNTER(child_profile, "ReadIoOps", TUnit::UNIT);
+//  counters_.read_wait_time = ADD_TIMER(child_profile, "ReadIoWaitTime");
+//  counters_.read_io_ops = ADD_COUNTER(child_profile, "ReadIoOps", TUnit::UNIT);
   counters_.bytes_read = ADD_COUNTER(child_profile, "ReadIoBytes", TUnit::BYTES);
```
chaoyli, line 397: not used profile can be removed.

chaoyli (changes requested):

be/src/exec/exec_node.cpp
HappenLee: Hi, as we can see. this counter ```_memory_used_counter``` is only used for count mem used for ```_tuple_pool``` and ```_hash_tbl```. 
But the ```_tuple_pool``` and ```_hash_tbl``` was already track the mem used by ```MemTracker``` and counted mem used in ```PeakMemoryUsage``` counter.  We do really need this counter ```_memory_used_counter```?

image


So, I think keep the ```_memory_used_counter ```  named ```MemoryUsed``` is confusing. 

I have two different suggestions

* Delete all code use ```_memory_used_counter``` in the future, all mem used is tracked by ```MemTracker```, we show peakmemusage it by the counter in ```MemTracker```.
* Keep the counter ```_memory_used_counter```, but change the name ```MemoryUsed``` to ```MemoryUsedByHashBTAndTuplePool```.

Which suggestion do you think is more suitable？@gaodayue

be/src/runtime/bufferpool/buffer_pool_counters.h
HappenLee: ok

chaoyli (dismissed):

be/src/exec/exec_node.cpp
kangkaisen: >Delete all code use _memory_used_counter in the future, all mem used is tracked by MemTracker, we show peakmemusage it by the counter in MemTracker.

I prefer this.

morningman (approved):
LGTM

PR #3417, opened 2020-04-28
[metrics] Make DorisMetrics to be a real singleton

### LINKED ISSUE #3398: Make DorisMetrics class in BE to be a real singleton
Current implemantion of class DorisMetrics is not a real singleton, but we use it as a singleton. Some usage may cause problem, especially in unit tests, like concurrently calls of `initialize()` may cause memory problems; metrics will be duplicate registered in `initialize()`, etc. I found these problems when run unit tests in ASAN mode.

Make DorisMetrics class to be a real singleton before run unit tests in ASAN mode.

acelyc111:
Ref https://github.com/apache/incubator-doris/issues/3398
NOTE: There are still some member variables are static, all of them are global metrics, like `static IntCounter fragment_requests_total;`, leave them there to avoid too many changes.


chaoyli:
You can also change the static variable.

acelyc111:
> You can also change the static variable.

Done

chaoyli (approved):

PR #3422, opened 2020-04-29
[Bug] Fix bug that load data to wrong temp partitions

### LINKED ISSUE #3421: [TempPartition] Load data to wrong partition
**Describe the bug**
When loading data without specifying partition, the data may be loaded to some temp
partitions, which is not expected.

**To Reproduce**
1. create a table;
2. create a temp partition;
3. load data without specifying partition;
4. data may be found in temp partition;

**Expected behavior**
All data should be in formal partitions;

morningman:
When loading data without specifying partition, the data should only be loaded to
formal partitions, not including temp partitions;

Fix: #3421

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/InsertStmt.java
+++ b/fe/src/main/java/org/apache/doris/analysis/InsertStmt.java
@@ -88,7 +88,8 @@ public class InsertStmt extends DdlStmt {
 
     private final TableName tblName;
     private final PartitionNames targetPartitionNames;
-    // parsed from targetPartitionNames. empty means no partition specified
+    // parsed from targetPartitionNames.
+    // if targetPartitionNames is not set, add all partitions' id of the table into it
     private List<Long> targetPartitionIds = Lists.newArrayList();
     private final List<String> targetColumnNames;
     private QueryStmt queryStmt;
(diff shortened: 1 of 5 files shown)


fe/src/main/java/org/apache/doris/planner/StreamLoadPlanner.java
```diff
@@ -185,6 +185,10 @@ public TExecPlanFragmentParams plan(TUniqueId loadId) throws UserException {
                 }
                 partitionIds.add(part.getId());
             }
+        } else {
+            for (Partition partition : destTable.getPartitions()) {
```
kangkaisen, line 189: Add a comment to explain why we return all partitions.

morningman: OK

fe/src/main/java/org/apache/doris/analysis/InsertStmt.java
```diff
@@ -88,7 +88,8 @@
 
     private final TableName tblName;
     private final PartitionNames targetPartitionNames;
-    // parsed from targetPartitionNames. empty means no partition specified
+    // parsed from targetPartitionNames.
+    // if targetPartitionNames is not set, add all partitions' id of the table into it
```
EmmyMiao87, line 92: add all partition's id except temp partition

EmmyMiao87 (approved):
LGTM

Issue #3321, opened 2020-04-15
[Delete] Support delete on non-key column in duplicate-key table


morningman:
**Is your feature request related to a problem? Please describe.**

In a table with Duplicate Key storage model, the key columns are only columns for sorting.

Currently, Doris only support executing `DELETE FROM` on key columns.

we should support executing `DELETE FROM` operation on both key and non-key columns.

yangzhg:
assin to me

PR #3424, opened 2020-04-29
allow delete duplicated non-key column using delete from 

yangzhg:
#3321 

allow delete duplicated non-key column using `delete from` clasue
```
mysql> show create table table1;
+--------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table  | Create Table                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
+--------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| table1 | CREATE TABLE `table1` (
  `siteid` int(11) NULL DEFAULT "10" COMMENT "",
  `citycode` smallint(6) NULL COMMENT "",
  `username` varchar(32) NULL DEFAULT "" COMMENT "",
  `pv` bigint(20) NULL DEFAULT "0" COMMENT ""
) ENGINE=OLAP
DUPLICATE KEY(`siteid`, `citycode`, `username`)
COMMENT "OLAP"
DISTRIBUTED BY HASH(`siteid`) BUCKETS 10
PROPERTIES (
"storage_type" = "COLUMN",
 "replication_num" = "1",
 "in_memory" = "false"
); |
+--------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
mysql> select * from table1;
+--------+----------+----------+------+
| siteid | citycode | username | pv   |
+--------+----------+----------+------+
|      3 |        2 | tom      |    2 |
|      3 |        2 | tom      |    2 |
|      5 |        3 | helen    |    3 |
|      2 |        1 | grace    |    2 |
|      2 |        1 | grace    |    2 |
|      2 |        1 | grace    |    2 |
|      4 |        3 | bush     |    3 |
+--------+----------+----------+------+
7 rows in set (0.05 sec)
mysql> delete from table1 where pv =3;
Query OK, 0 rows affected (0.03 sec)
{'label':'delete_6a5c9479-5581-46b7-9ebc-c4f770639ca9', 'status':'VISIBLE', 'txnId':'13388'}

mysql> select * from table1;
+--------+----------+----------+------+
| siteid | citycode | username | pv   |
+--------+----------+----------+------+
|      2 |        1 | grace    |    2 |
|      2 |        1 | grace    |    2 |
|      2 |        1 | grace    |    2 |
|      3 |        2 | tom      |    2 |
|      3 |        2 | tom      |    2 |
+--------+----------+----------+------+
5 rows in set (0.06 sec)
```


be/src/olap/olap_cond.cpp
```diff
@@ -681,17 +681,17 @@ int Conditions::delete_pruning_filter(const std::vector<KeyRange>& zone_maps) co
     for (auto& cond_it : _columns) {
         /*
          * this is base on the assumption that the delete condition
-         * is only about key field, not about value field.
+         * is only about key field, not about value field except the storage model is duplicate.
         */
-        if (cond_it.second->is_key() && cond_it.first > zone_maps.size()) {
+        if ((cond_it.second->is_key() || _schema->keys_type() == KeysType::DUP_KEYS) && cond_it.first > zone_maps.size()) {
```
morningman, line 686: `cond_it.second->is_key() || _schema->keys_type() == KeysType::DUP_KEYS`

This condition shows many times. How about extract it to a method?

be/test/olap/delete_handler_test.cpp
```diff
@@ -265,6 +358,17 @@ TEST_F(TestDeleteConditionHandler, StoreCondNonexistentColumn) {
 
     failed_res = _delete_condition_handler.generate_delete_predicate(tablet->tablet_schema(), conditions, &del_pred);;
     ASSERT_EQ(OLAP_ERR_DELETE_INVALID_CONDITION, failed_res);
+
+    // 'v'是value列 但是在duplicate 模型汇总
```
morningman, line 362: English

morningman:
We made a bug after supporting creating materialized view on Duplicate Key table.
And could you please fix it too? That is:

After creating a table with column, eg, A, B, C, D in Duplicate mode, we can process Delete
operation on all 4 columns.
But if we create a materialized view, eg, A, SUM(B), we can't not delete on column B any more.

This check should be done in `DeleteHandler.checkDeleteV2()`, but it is missing.

chaoyli (changes requested):

be/src/olap/rowset/segment_group.cpp
```diff
@@ -697,6 +697,13 @@ const TabletSchema& SegmentGroup::get_tablet_schema() {
     return *_schema;
 }
 
+int SegmentGroup::get_num_zone_map_columns() {
```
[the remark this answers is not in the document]
yangzhg: this function only get the column num of maybe has ZoneMap, in agg, uniq, replace table  itt returns the key columns count, it certainly sorted and had ZoneMap, in duplicate table it return the all column count, the key columns has sorted and has  ZoneMap, the value column in new code also has ZoneMap, but not sorted, so it is not  `Most of the time, you will have to filter the delete predicate row by row.`,  only  when delete by duplicted values, in otherwise if we sorted the duplicate table's value column, it will be the same with key column essentially, so that the value column is unmeaningful

yangzhg:
> We made a bug after supporting creating materialized view on Duplicate Key table.
> And could you please fix it too? That is:
> 
> After creating a table with column, eg, A, B, C, D in Duplicate mode, we can process Delete
> operation on all 4 columns.
> But if we create a materialized view, eg, A, SUM(B), we can't not delete on column B any more.
> 
> This check should be done in `DeleteHandler.checkDeleteV2()`, but it is missing.

done

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
```diff
@@ -509,8 +510,8 @@ private void checkDeleteV2(OlapTable table, Partition partition, List<Predicate>
                 if (column == null) {
                     ErrorReport.reportDdlException(ErrorCode.ERR_BAD_FIELD_ERROR, columnName, indexName);
                 }
-
-                if (table.getKeysType() == KeysType.DUP_KEYS && !column.isKey()) {
+                MaterializedIndexMeta indexMeta = table.getIndexIdToMeta().get(index.getId());
+                if (indexMeta.getKeysType() != KeysType.DUP_KEYS && !column.isKey()) {
```
morningman, line 514: if index is DUPLICATED, but column type is float or double, we should also forbid the delete operation.

fe/src/main/java/org/apache/doris/load/Load.java
```diff
@@ -3123,7 +3123,7 @@ private void checkDeleteV2(OlapTable table, Partition partition, List<Predicate>
             }
 
             Column column = nameToColumn.get(columnName);
-            if (!column.isKey()) {
+            if (!column.isKey() && table.getKeysType() != KeysType.DUP_KEYS) {
```
morningman, line 3126: No need to modify here, this method is deprecated.

morningman (changes requested):

be/src/olap/rowset/column_data.cpp
```diff
@@ -509,7 +510,10 @@ int ColumnData::delete_pruning_filter() {
         return DEL_NOT_SATISFIED;
     }
 
-    if (!_segment_group->has_zone_maps()) {
+    int num_zone_maps = _schema.keys_type() == KeysType::DUP_KEYS ? _schema.num_columns() : _schema.num_key_columns();
```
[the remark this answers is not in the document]
yangzhg: float and double is disabled in fe

be/src/olap/delete_handler.cpp
```diff
@@ -105,10 +105,10 @@ OLAPStatus DeleteConditionHandler::check_condition_valid(
     // 检查指定的列是不是key，是不是float或doulbe类型
     const TabletColumn& column = schema.column(field_index);
 
-    if (!column.is_key()
+    if ((!column.is_key() && schema.keys_type() != KeysType::DUP_KEYS)
```
kangpinghuang, line 108: just check column.agg_type = AGG_NONE？

be/src/olap/olap_cond.cpp
```diff
@@ -637,7 +637,7 @@ bool Conditions::delete_conditions_eval(const RowCursor& row) const {
     }
     
     for (auto& each_cond : _columns) {
-        if (each_cond.second->is_key() && !each_cond.second->eval(row)) {
+        if (_cond_column_is_key_or_duplicate(each_cond.second) && !each_cond.second->eval(row)) {
```
kangpinghuang, line 640: I think check agg type is more clear?

be/src/olap/rowset/alpha_rowset.cpp
```diff
@@ -300,7 +300,7 @@ OLAPStatus AlphaRowset::init() {
 
         if (segment_group_meta.zone_maps_size() != 0) {
             size_t zone_maps_size = segment_group_meta.zone_maps_size();
-            size_t num_key_columns = _schema->num_key_columns();
+            size_t num_key_columns = _schema->keys_type() == KeysType::DUP_KEYS ? _schema->num_columns() : _schema->num_key_columns();
```
kangpinghuang, line 303: I think we should change the num_key_columns name.
num_key_columns  will be ambiguous.

be/src/olap/wrapper_field.h
```diff
@@ -81,7 +81,7 @@ class WrapperField {
     char* ptr() const { return _field_buf + 1; }
     size_t size() const { return _rep->size(); }
     size_t field_size() const { return _rep->field_size(); }
-    bool is_null() const { return *reinterpret_cast<bool*>(_field_buf); }
+    bool is_null() const { return _field_buf == nullptr || *reinterpret_cast<bool*>(_field_buf); }
```
kangpinghuang, line 84: why to add this condition?

fe/src/main/java/org/apache/doris/load/DeleteHandler.java
```diff
@@ -478,9 +479,11 @@ private void checkDeleteV2(OlapTable table, Partition partition, List<Predicate>
             }
 
             Column column = nameToColumn.get(columnName);
-            if (!column.isKey()) {
+            if (!column.isKey() && table.getKeysType() != KeysType.DUP_KEYS
```
kangpinghuang, line 482: should add some comment to explain why float and double type delete predicate can not be supported.

kangpinghuang (changes requested):

be/src/olap/rowset/alpha_rowset.cpp
yangzhg: it better to submit anthor pr to fix this

be/src/olap/olap_cond.cpp
yangzhg: agg type has two may be null or AGG_NONE, it maybe  more complicated

chaoyli (approved):

fe/src/main/java/org/apache/doris/alter/MaterializedViewHandler.java
```diff
@@ -549,7 +548,10 @@ private RollupJobV2 createMaterializedViewJob(String mvName, String baseIndexNam
                     }
                     keyStorageLayoutBytes += baseColumn.getType().getStorageLayoutBytes();
                     Column rollupColumn = new Column(baseColumn);
-                    if ((i + 1) <= FeConstants.shortkey_max_column_count
+                    if(changeStorageFormat) {
+                        rollupColumn.setIsKey(baseColumn.isKey());
+                        rollupColumn.setAggregationType(baseColumn.getAggregationType(), true);
```
morningman, line 553: What is about?

be/src/olap/rowset/segment_group.cpp
```diff
@@ -249,13 +249,13 @@ OLAPStatus SegmentGroup::add_zone_maps_for_linked_schema_change(
...
-    // 1. rollup tablet num_key_columns() will less than base tablet zone_map_fields.size().
+    // 1. rollup tablet get_num_zone_map_columns() will less than base tablet zone_map_fields.size().
     //    For LinkedSchemaChange, the rollup tablet keys order is the same as base tablet
-    // 2. adding column to existed table, num_key_columns() will larger than
+    // 2. adding column to existed table, get_num_zone_map_columns() will larger than
     //    zone_map_fields.size()
 
     int num_new_keys = 0;
-    for (size_t i = 0; i < _schema->num_key_columns(); ++i) {
+    for (size_t i = 0; i < get_num_zone_map_columns(); ++i) {
```
morningman, line 258: All types of column has zone map now?

fe/src/main/java/org/apache/doris/alter/MaterializedViewHandler.java
yangzhg: when create a segmentv2 rollup on duplicate table, the previous code did not handle this specific sutiation, it will create a all key rollup, all value columns will convert to key column

be/src/olap/rowset/segment_group.cpp
yangzhg: float and double can generate zone map， bitmap cannot used in duplicate table, so it OK

morningman (approved):
LGTM

Issue #3420, opened 2020-04-29
[Website] Images load failed in new website


wyb:
**Describe the bug**
Images load failed in new website.
Images can be found in old docs/resources/images dir and are gone in new docs dir.

hffariel:
Caused by relative path, should be ok now.

Issue #3419, opened 2020-04-29
[Website] Some problem about the new website.


morningman:
The new website is great. But there are some problems:

1. I can't find the `Version` and `Language` select button on Mobile version of the website.
2. The document under `Apache社区(Apache Community)` should not be related to the Doris version. So I suggest placing it alone in the top navigation bar.
3. `Apache Commnity` is typo, should be: `Apache Community`.
4. the icon on webbrowser tag is wrong:
![image](https://user-images.githubusercontent.com/2899462/80558182-036df580-8a0c-11ea-8da8-4a8961362c8f.png)

5. There is no `Makefile` under `docs/`, so need to modify the `build.sh` to remove the `make doc`

6. Why there are 2 `readme.md` under `docs/`

hffariel:
1. Sidebar menu is ok now, but still has an issue related with [vuepress nav bar dropdown issue](https://github.com/vuejs/vuepress/pull/2339)(which could only be fixed by the next vuepress release).
2. Wait to be done
3. Spelling issue has been corrected.
4. Icon has been added to asf-site

rinkydevi:
i can contribute in this issue,can please help me to proceed

morningman:
> i can contribute in this issue,can please help me to proceed

Hi @rinkydevi , thanks for your reply.
However, this issue is very old and we currently have plans to rework the new site, so if you are interested, you can contact me on WeChat(morningman-cmy) to discuss some details.

rinkydevi:
@morningman  thank u for your response,
but i don't currently use WeChat, so if u could provide me another medium that would be wonderful!

morningman:
> rinkydevi

You also contact us via email: dev@doris.apache.org
See how to subscribe: http://doris.incubator.apache.org/community/subscribe-mail-list.html#_1-send-subscription-mail

rinkydevi:
@morningman Thank u so much sir, 
i've sent the mail, and hoping to work and discuss issues under your guidance 😊

rinkydevi:
@morningman thank u so much for this sir,
but when i sent the mail it is showing some kind of error,i will attach the screenshot below
![000](https://user-images.githubusercontent.com/82359874/152666804-c624adac-6487-4bad-8f66-d77524039809.jpg)

morningman:
Sorry I have no idea about this error...
Have you try it again?

And be sure that you have subscribed the dev@doris mail list, or you may not be able to receive our reply.

morningman:
> 

[...]

PR #3427, opened 2020-04-29
[Bugfix] document issues

hffariel:
## fix navbar not showing on mobile clients(#3419)

* problem causing
    NavLink component is used by both NavBar and SideBar, but versions.json request only executed in NavBar once, so NavLinks not showing in SideBar.

* fixing
    move the versions.json request to layouts/Layout.vue(which will be the customized entry of the site), then render the whole page again after requesting.

## fix image relative path problem(#3420)

* problem causing
    Image paths in markdown files are like '/images/xxx.png', but the site finally serving with a base url like /master/ /branch-0.12/, so images couldn't be found.

* fixing
    Wrap the image paths with $withBase (which will auto prefix the path with config.base) . And this action will be done by adding a sed command during travis building.


kangkaisen (approved):
+1. LGTM

PR #3428, opened 2020-04-29
[BugFix] document issues

hffariel:
## fix navbar not showing on mobile clients(#3419) 

* problem causing 
    NavLink component is used by both NavBar and SideBar, but versions.json request only executed in NavBar once, so NavLinks not showing in SideBar.
* fixing
    move the versions.json request to layouts/Layout.vue(which will be the customized entry of the site), then render the whole page again after requesting.

## fix image relative path problem(#3420)

* problem causing 
    Image paths in markdown files are like '/images/xxx.png', but the site finally serving with a base url like `/master/` `/branch-0.12/`, so images couldn't be found.
* fixing
    Wrap the image paths with `$withBase` (which will auto prefix the path with `config.base`) . And this action will be done by adding a `sed` command during travis building.


kangkaisen:
@hffariel Hi, This PR is reduplicated with https://github.com/apache/incubator-doris/pull/3427 ?

You could refer to https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53 to edit your commit message.

Because most of Doris committers don't familiar with vue.js. So, when you submit a PR, would better describe what the root cause of the issue, how do you fix the issue,  and add more comment for the code. Thanks!

kangkaisen (approved):
+1, LGTM

PR #3431, opened 2020-04-29
[Memory Engine] Make Tablet extensible

### LINKED ISSUE #3384: [Memory Engine] Make Tablet extensible
To add support new storage engine, we need to make an extensible tablet interface, so olap::StorageEngine can support and manage new tablet types.
To start, we can create a class BaseTablet and make Tablet and new MemTablet inherit this base class.

decster:
Adding a new storage engine, we need to make an extensible tablet interface, so olap/StorageEngine can support and manage new tablet types.

To start, this commit creates a class BaseTablet and make Tablet and new MemTablet inherit this base class, some common fields & methods are moved to BaseTablet class, which fields and methods belong to base/old class is not finalized yet, it will change as the project evolves.

Fix #3384


be/src/olap/base_tablet.h
```diff
@@ -0,0 +1,133 @@
...
+#define DORIS_BE_SRC_OLAP_BASE_TABLET_H
+
+#include <memory>
+#include "olap/olap_define.h"
+#include "olap/tablet_meta.h"
+#include "olap/utils.h"
+
+namespace doris {
+
+class BaseTablet : public std::enable_shared_from_this<BaseTablet> {
```
morningman, line 28: Could you add comment for this class?

morningman (commented):
Hi, does it better to put the `memory/` dir under the `olap/` dir?

be/src/olap/base_tablet.h
decster: Sure

kangkaisen (approved):
+1, LGTM

Issue #2598, opened 2019-12-27
Support append_trailing_char_if_absent function


HangyuanLiu:
**Is your feature request related to a problem? Please describe.**
appendTrailingCharIfAbsent(s, c)
If the 's' string is non-empty and does not contain the 'c' character at the end, it appends the 'c' character to the end.
#2591

PR #3439, opened 2020-04-29
Support append_trailing_char_if_absent  function

Youngwb:
support append_trailing_char_if_absent function 
#2598


fe/src/main/java/org/apache/doris/analysis/FunctionCallExpr.java
```diff
@@ -604,6 +604,23 @@ public void analyzeImpl(Analyzer analyzer) throws AnalysisException {
             return;
         }
 
+        if (fn.getFunctionName().getFunction().equals("append_trailing_char_if_absent")) {
```
kangkaisen, line 607:
```suggestion
        if (fnName.getFunction().equalsIgnoreCase("append_trailing_char_if_absent")) {
```

be/src/exprs/string_functions.cpp
```diff
@@ -212,6 +212,28 @@ StringVal StringFunctions::rpad(
...
+    }
+    if (str.len == 0) {
+        return trailing_char;
+    }
+    if (str.ptr[str.len-1] == trailing_char.ptr[0]) {
+        return str;
+    }
+
+    StringVal result(context, str.len + trailing_char.len);
+    if (UNLIKELY(result.is_null)) {
```
kangkaisen, line 229: Why new result could null?

Youngwb: StringVal will alloc memory by context's FreePool, that may be null. I use this check according to other string_functions. But this check `is_null` may be useless, It can be removed

docs/documentation/cn/sql-reference/sql-functions/string-functions/append_trailing_char_if_absent.md
```diff
@@ -0,0 +1,46 @@
...
+under the License.
+-->
+
+# append_trailing_char_if_absent
+## description
+### Syntax
+
+`VARCHAR append_trailing_char_if_absent(VARCHAR str, VARCHAR trailing_char)`
+
+如果's'字符串非空并且末尾不包含'c'字符，则将'c'字符附加到末尾
```
imay, line 24: Should explain when input trailing_char has multiple chars, how will this function act.

gensrc/script/doris_builtins_functions.py
```diff
@@ -531,6 +531,8 @@
     [['rpad'], 'VARCHAR', ['VARCHAR', 'INT', 'VARCHAR'],
             '_ZN5doris15StringFunctions4rpadEPN9doris_udf'
             '15FunctionContextERKNS1_9StringValERKNS1_6IntValES6_'],
+    [['append_trailing_char_if_absent'], 'VARCHAR', ['VARCHAR', 'VARCHAR'],
```
imay, line 534: Could you tell me in what scenario this function is used?

imay (changes requested):

docs/documentation/cn/sql-reference/sql-functions/string-functions/append_trailing_char_if_absent.md
Youngwb: Done

docs/documentation/cn/sql-reference/sql-functions/string-functions/append_trailing_char_if_absent.md
```diff
@@ -0,0 +1,46 @@
+<!--
```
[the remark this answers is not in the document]
Youngwb: Done

gensrc/script/doris_builtins_functions.py
Youngwb: may be such like this : 
```
MySQL [test]> select append_trailing_char_if_absent('http://google.com','/');
+----------------------------------------------------------+
| append_trailing_char_if_absent('http://google.com', '/') |
+----------------------------------------------------------+
| http://google.com/                                       |
+----------------------------------------------------------+
1 row in set (0.01 sec)

MySQL [test]> select append_trailing_char_if_absent('http://google.com/','/');
+-----------------------------------------------------------+
| append_trailing_char_if_absent('http://google.com/', '/') |
+-----------------------------------------------------------+
| http://google.com/                                        |
+-----------------------------------------------------------+
```

be/src/exprs/string_functions.cpp
kangkaisen: In fact, FreePool allocate never return null.
So `result.is_null` check is unnecessary.

docs/en/sql-reference/sql-functions/string-functions/append_trailing_char_if_absent.md
```diff
@@ -0,0 +1,58 @@
...
+-->
+
+# append_trailing_char_if_absent
+## description
+### Syntax
+
+`VARCHAR append_trailing_char_if_absent(VARCHAR str, VARCHAR trailing_char)`
+
+If the s string is non-empty and does not contain the c character at the end, it appends the c character to the end.
+Trailing_char contains only one character, and it will report an error if contains more than one character
```
imay, line 34: Why not return null when the second parameter length is not 1.
You can't check its length if it is not const, such as a table column.

Youngwb: Because it's function name contains `trailing_char`, and this is also consistent with the clickhose function. I did a cast in fe `if (((StringLiteral) arg1).getValue().length() != 1) `, It will throw ClassCastException when SlotRef cast to StringLiteral. I can also add type checking if necessary

imay: In the case `append_trailing_char_if_absent(str, col1)` in which col1 is a column of a table. You can not know if the length of col1 is 1 or not.
Also there is no need to keep the same with CK.
It is convention to return null when the argument is invalid in Doris.

So what I suggest is to return null when the input argument is not valid. And there is no need to check its input in FE.

Youngwb: OK, Done

be/src/exprs/string_functions.cpp
imay: I agree with kangkaisen, this check can be removed

fe/src/main/java/org/apache/doris/analysis/FunctionCallExpr.java
```diff
@@ -605,6 +605,19 @@ public void analyzeImpl(Analyzer analyzer) throws AnalysisException {
             return;
         }
 
+        if (fnName.getFunction().equalsIgnoreCase("append_trailing_char_if_absent")) {
```
imay, line 608: I think this check is unnecessary, can be removed.
If you keep it, this will make this function more complicate.

be/test/exprs/string_functions_test.cpp
```diff
@@ -252,6 +264,26 @@ TEST_F(StringFunctionsTest, null_or_empty) {
...
+            StringVal("123c"), StringVal("c")));
+
+    ASSERT_EQ(StringVal("c"), StringFunctions::append_trailing_char_if_absent(ctx,
+            StringVal(""), StringVal("c")));
+
+    ASSERT_EQ(StringVal::null(), StringFunctions::append_trailing_char_if_absent(ctx,
+            StringVal::null(), StringVal("c")));
+
+    ASSERT_EQ(StringVal::null(), StringFunctions::append_trailing_char_if_absent(ctx,
+            StringVal("a"), StringVal::null()));
```
imay, line 284: Should add test that second argument has multiple characters.

be/src/exprs/string_functions.cpp
```diff
@@ -212,6 +212,28 @@ StringVal StringFunctions::rpad(
...
+
+StringVal StringFunctions::append_trailing_char_if_absent(doris_udf::FunctionContext* context,
+        const doris_udf::StringVal& str, const doris_udf::StringVal& trailing_char) {
+    if (str.is_null || trailing_char.is_null || trailing_char.len != 1) {
+        return StringVal::null();
+    }
+    if (str.len == 0) {
+        return trailing_char;
+    }
+    if (str.ptr[str.len-1] == trailing_char.ptr[0]) {
```
imay, line 224:
```suggestion
    if (str.ptr[str.len - 1] == trailing_char.ptr[0]) {
```

be/src/exprs/string_functions.cpp
```diff
@@ -212,6 +212,28 @@ StringVal StringFunctions::rpad(
...
+        return str;
+    }
+
+    StringVal result(context, str.len + trailing_char.len);
```
imay, line 228:
```suggestion
    StringVal result(context, str.len + 1);
```

fe/src/main/java/org/apache/doris/analysis/FunctionCallExpr.java
Youngwb: Yes，I understand there will make this function more complicate, but  I need to avoid two parameters be Non-String type，such as `append_trailing_char_if_absent(1,2)`, and I don't known how to handle this case if i don't check here. Maybe there is a better way to handle this case if you known.

imay: For this case you mentioned, Doris can cast input argument to function's definition type. So you don't handle this in this place.

Youngwb: ok，Done

be/test/exprs/string_functions_test.cpp
Youngwb: ok，Done

be/src/exprs/string_functions.cpp
Youngwb: ok,Done

docs/en/sql-reference/sql-functions/string-functions/append_trailing_char_if_absent.md
```diff
@@ -0,0 +1,58 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# append_trailing_char_if_absent
+## description
+### Syntax
```
imay, line 29: split these lines with empty line

imay (commented):
UT failed, Could you please rebase master?

imay (approved):
LGTM

kangkaisen (approved):
+1, LGTM

morningman:
UT passed. LGTM

PR #3451, opened 2020-04-30
[Fix] Fix bug that rowset meta is deleted after compaction

### LINKED ISSUE #3450: [Bug] Rowset meta is missing after compaction
**Describe the bug**
The phenomenon is like this:

Suppose we have [0-1] [2-2] [3-3] 3 versions.
A cumulative compaction starts to compact [2-2] version(a single rowset with many segments).
When the compaction is completed, the query will throw error

`failed to initialize storage reader. tablet=11070.1076216277.e647634f8b70a458-f889546ad683c8af, res=-702`

In other words, the version is missing.

**To Reproduce**
Load a relative large amount of data in a single load job, and generate a rowset with many segments. So that it will trigger a single rowset compaction, and after compaction, the query will
be failed.

Fortunately, the data has not been lost and can be recovered by restarting BE.

VERSION：0.12.8

morningman:
After compaction, the tablet rowset meta will be modified by
adding to new output rowsets and deleting the old input rowsets.
The output version may equals to the input version.

So we should delete the "input" version from _rs_version_map
before adding the "output" version to _rs_version_map. Otherwise,
the new "output" version will be lost in _rs_version_map.

Fix #3450

### DIFF
--- a/be/src/olap/tablet.cpp
+++ b/be/src/olap/tablet.cpp
@@ -234,17 +234,22 @@ OLAPStatus Tablet::add_rowset(RowsetSharedPtr rowset, bool need_persist) {
 
 void Tablet::modify_rowsets(const vector<RowsetSharedPtr>& to_add,
                             const vector<RowsetSharedPtr>& to_delete) {
+    // the compaction process allow to compact the single version, eg: version[4-4].
+    // this kind of "single version compaction" has same "input version" and "output version".
+    // which means "to_add->version()" equals to "to_delete->version()".
+    // So we should delete the "to_delete" before adding the "to_add",
+    // otherwise, the "to_add" will be deleted from _rs_version_map, eventually.
+    vector<RowsetMetaSharedPtr> rs_metas_to_delete;
+    for (auto& rs : to_delete) {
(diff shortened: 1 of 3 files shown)


kangkaisen (dismissed):
+1. LGTM

kangkaisen:
@morningman Hi, How do we avoid this bug happen again？do we need add a compaction test framework to cover all compaction test cases?

morningman:
> @morningman Hi, How do we avoid this bug happen again？do we need add a compaction test framework to cover all compaction test cases?

Agreed, I will try to find a way to implement it later.
By the way, could you help me to approve this PR again? I just fix a ut bug.

kangkaisen:
> > @morningman Hi, How do we avoid this bug happen again？do we need add a compaction test framework to cover all compaction test cases?
> 
> Agreed, I will try to find a way to implement it later.
> By the way, could you help me to approve this PR again? I just fix a ut bug.

OK

kangkaisen (dismissed):
+1, LGTM

chaoyli (approved):

Issue #3452, opened 2020-05-02
Support simple transitivity on join predicate pushdown


kangkaisen:
For query:
```
select t1.id, t1.value, t2.value
from t1
left join t2 on t1.id = t2.id
where t1.id > 1
```
We infer new  predicate `t2.id > 1` and pushdown it to t2 Scannode.

chaoyli:
```
select t1.id, t1.value, t2.value
from t1
right join t2 on t1.id = t2.id
where t2.id > 1
```
I think this query is also can be pushed down?

kangkaisen:
@chaoyli Of course. we could support right join and full out join. But inner join and left join are more common, and this PR has been tested in MeiTuan prod env two years. So I think we could merge this PR firstly, and then support  right join and full out join later.

PR #3453, opened 2020-05-02
Support simple transitivity on join predicate pushdown

kangkaisen:
For #3452 

Current implement is very simply and conservative, because our query planner is error-prone.

After we implement the new query planner, we could do this work by `Predicate Equivalence Class` and `PredicatePushDown` rule like presto.

### DIFF
--- a/fe/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
+++ b/fe/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
@@ -72,6 +72,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
+import static org.apache.doris.analysis.Predicate.canPushDownPredicate;
+
 /**
  * Constructs a non-executable single-node plan from an analyzed parse tree.
  * The single-node plan does not contain data exchanges or data-reduction optimizations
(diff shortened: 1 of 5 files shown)


fe/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
```diff
@@ -1372,6 +1414,24 @@ private PlanNode createScanNode(Analyzer analyzer, TableRef tblRef, SelectStmt s
         return scanNode;
     }
 
+    private Expr rewritePredicate(Analyzer analyzer, Expr oldPredicate, Expr leftChild) {
```
morningman, line 1417: Add comment for this method?

fe/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
```diff
@@ -72,6 +72,8 @@
 import java.util.Map;
 import java.util.UUID;
 
+import static org.apache.doris.analysis.Predicate.canPushDownPredicate;
```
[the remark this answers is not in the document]
kangkaisen: OK

fe/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
kangkaisen: OK

morningman (approved):

xinghuayu007:
@kangkaisen ，Mysql has implemented an optimizer with Multiple Equivalence Predicate (MEP). For example, a
=b and b= c and c = 123  ===>>  =(a,b,c,123), which implement transitivity of predicates. Does this algorithm solve your problem? Add my WeChat: xinghuayu520 to have a disscussion.

PR #3456, opened 2020-05-03
UT bug fix for commit (#3336) 

wangbo:
There is a UT bug for (#3336) .
SSD cooldowntime shouldn't be fix time in UT;


morningman (approved):
LGTM

Issue #3460, opened 2020-05-04
[Help] Not easy to balance the data when adding new disk to BE


morningman:
**Is your feature request related to a problem? Please describe.**
When adding new disks to BE, the data can not do automatic balance among disks within the BE.

**Describe the solution you'd like**
Give a way to balance the data among disks within the BE

**Describe alternatives you've considered**
Utilize the decommission operation. The decommission operation will move all data to other
node. And then the decommission BE is empty, which is like a newly added node.
Then the balance logic can try to balance the data back to that BE, and this will help 
redistributing data evenly to all disks of that BE.

The problem is after decommission, the BE will be dropped automatically, so we should find a
way to NOT drop it.

kangkaisen:
@morningman Hi, Intuitively,I think utilize the decommission operation to solve the balance issue is strange.

morningman:
> @morningman Hi, Intuitively,I think utilize the decommission operation to solve the balance issue is strange.

Yes, So I add a TODO in the comment of the PR #3461. This just a temporary solution.
And by the way, decommission without dropping the BE is also good for some scenarios.

For example, I want to repair a machine offline, but I don't want (or can't) perform an ADD or DROP backend operation, which will changing the topology information of the cluster.

kangkaisen:
> > @morningman Hi, Intuitively,I think utilize the decommission operation to solve the balance issue is strange.
> 
> Yes, So I add a TODO in the comment of the PR #3461. This just a temporary solution.
> And by the way, decommission without dropping the BE is also good for some scenarios.
> 
> For example, I want to repair a machine offline, but I don't want (or can't) perform an ADD or DROP backend operation, which will changing the topology information of the cluster.

OK

PR #3461, opened 2020-05-04
[Decommission] Support NOT dropping BE after decommission

morningman:
Add a new config `drop_backend_after_decommission` in FE. if this config
is false, the BE will not be dropped after finishing decommission operation.

This new config is try to solve the problem described in ISSUE: #3460 .

TODO:
This method will generate a lot of data migration, so it is only a temporary solution.
After that, we should try to solve the problem of data balancing within the BE.

This CL also add the documents of FE and BE configuration.
These documents are incomplete and can be added later.

### DIFF
--- a/docs/en/administrator-guide/config/fe_config.md
+++ b/docs/en/administrator-guide/config/fe_config.md
@@ -1,7 +1,7 @@
 ---
 {
     "title": "FE Configuration",
-    "language": "zh-CN"
+    "language": "en"
 }
 ---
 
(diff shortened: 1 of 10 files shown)


docs/en/administrator-guide/config/fe_config.md
```diff
@@ -1,7 +1,7 @@
 ---
 {
-    "title": "Configuration",
-    "language": "en"
+    "title": "FE Configuration",
+    "language": "zh-CN"
```
kangkaisen, line 4: should be en ?

morningman: Fixed

kangkaisen (approved):
+1, LGTM

PR #3462, opened 2020-05-04
[Memory Engine] Add hash index implementation

decster:
This commit adds a hash index implementation for memory engine, it maps uint64_t hashcode(key) to uint32_t rowid(value). This hash index is different from the classical hash table in many ways:

1. To save memory usage, it does store value (only rowid), user need to follow the rowid and get the actual rowkey and check rowkey equality.
2. Only support insertion, no delete support
3. One writer/multi-reader concurrency support

It's specially designed for the memory storage engine.


kangkaisen:
@decster Hi, I think we need to support SSE2 operation rollback to normal scalar operation，because Doris BE maybe run on ARM architecture machine.

be/src/olap/memory/hash_index.cpp
```diff
@@ -0,0 +1,161 @@
...
+        auto tags = _mm_load_si128(chunk.tagVector());
+        auto eqs = _mm_cmpeq_epi8(tags, tests);
+        uint32_t mask = _mm_movemask_epi8(eqs) & 0xfff;
+        while (mask != 0) {
+            uint32_t i = __builtin_ctz(mask);
+            mask &= (mask -1);
+            entries.emplace_back(chunk.values[i]);
+        }
+        if (sz == HashChunk::CAPACITY) {
+            uint64_t step =  tag*2+1; // 1;
```
morningman, line 102:
```suggestion
            uint64_t step =  tag * 2 + 1; // 1;
```
And what is the comment `1` for?

be/src/olap/memory/hash_index.cpp
```diff
@@ -0,0 +1,161 @@
...
+    uint64_t tag = key_hash & 0xff;
+    if (tag == 0) {
+        tag = 1;
+    }
+    uint64_t pos = (key_hash >> 8) & _chunk_mask;
+    uint64_t orig_pos = pos;
+    while (true) {
+        HashChunk& chunk = _chunks[pos];
+        if (chunk.size == HashChunk::CAPACITY) {
+            uint64_t step =  tag*2+1; // 1;
```
morningman, line 139:
```suggestion
            uint64_t step =  tag * 2 + 1; // 1;
```

be/src/olap/memory/hash_index.cpp
```diff
@@ -0,0 +1,161 @@
...
+    uint64_t tpos = slot & 0xf;
+    HashChunk& chunk = _chunks[pos];
+    uint64_t tag = key_hash & 0xff;
+    if (tag == 0) {
+        tag = 1;
+    }
+    chunk.tags[tpos] = tag;
+    chunk.values[tpos] = value;
+    if (tpos == chunk.size) {
+        chunk.size.store(tpos+1, std::memory_order_release);
```
morningman, line 124: Why here we use `store()` method, but in `HashIndex::add`, it simply use `chunk.size++`?

be/src/olap/memory/hash_index.cpp
```diff
@@ -0,0 +1,161 @@
...
+    auto tests = _mm_set1_epi8(static_cast<uint8_t>(tag));
+    while (true) {
+        HashChunk& chunk = _chunks[pos];
+        uint32_t sz = chunk.size.load(std::memory_order_acquire);
+        auto tags = _mm_load_si128(chunk.tagVector());
+        auto eqs = _mm_cmpeq_epi8(tags, tests);
+        uint32_t mask = _mm_movemask_epi8(eqs) & 0xfff;
+        while (mask != 0) {
+            uint32_t i = __builtin_ctz(mask);
+            mask &= (mask -1);
```
morningman, line 98:
```suggestion
            mask &= (mask - 1);
```

be/src/olap/memory/hash_index.h
```diff
@@ -0,0 +1,91 @@
...
+    size_t size() const { return _size; }
+
+    // Return max number of elements this hash index can hold
+    // If size >= max_size, this hash index needs to expand and rebuild
+    size_t max_size() const { return _max_size; }
+
+    // Find by key hash, put all candidate values into entries,
+    // and return a entry position, so later user can use this position to
+    // add a value with the same key hash directly into this hash index.
+    uint64_t find(uint64_t key_hash, std::vector<uint32_t> &entries) const;
```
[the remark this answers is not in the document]
decster: Sure

be/src/olap/memory/hash_index.cpp
decster: fixed, remove load/save entirely

be/src/olap/memory/hash_index.cpp
```diff
@@ -0,0 +1,175 @@
...
+#endif
+
+#include <algorithm>
+#include <vector>
+
+#include "gutil/stringprintf.h"
+
+namespace doris {
+
+using TagVector = __m128i;
```
morningman, line 32: if `__SSE2__ ` is undefined, the `__m128i` is also undefined.

be/src/olap/memory/hash_index.h
```diff
@@ -0,0 +1,92 @@
...
+//
+// Note: This hash index is thread-safe, but it only support
+// single-writer/multi-reader style concurrency. When hash index reaches it's
+// capacity, a writer need to expand&rebuild, this is not thread-safe, so the
+// writer need to do a copy-on-write, and make it's owner to link to the new
+// reference, while readers still reading the old reference.
+//
+// Note: for more info, please refer to:
+// https://engineering.fb.com/developer-tools/f14/
+class HashIndex : public RefCountedThreadSafe<HashIndex> {
```
chaoyli, line 50: I think you should add some comment to elaborate the hash algorithm.
From the principle and the specific addd/set/find function. SIMD also can be introduced with some description.

chaoyli (changes requested):

be/src/olap/memory/hash_index.cpp
```diff
@@ -0,0 +1,171 @@
...
+            return (pos << 4) | sz;
+        }
+    }
+}
+
+void HashIndex::set(uint64_t slot, uint64_t key_hash, uint32_t value) {
+    uint64_t pos = slot >> 4;
+    uint64_t tpos = slot & 0xf;
+    HashChunk& chunk = _chunks[pos];
+    uint64_t tag = key_hash & 0xff;
```
kangkaisen, line 128:
```suggestion
    uint64_t tag = std::max(1, key_hash & 0xff);
```

decster: fixed

be/src/olap/memory/hash_index.h
decster: Added more comments

chaoyli (approved):

morningman:
LGTM

PR #3475, opened 2020-05-05
[Bug] Fix bug of predicate pushdown logic

### LINKED ISSUE #3474: [Bug] Query throw unexpected exception: null
**Describe the bug**
Query return error: `Unexpected exception: null`

**To Reproduce**
```
CREATE TABLE test.`pushdown_test` (
  `k1` tinyint(4) NULL COMMENT "",
  `k2` smallint(6) NULL COMMENT "",
  `k3` int(11) NULL COMMENT "",
  `k4` bigint(20) NULL COMMENT "",
  `k5` decimal(9, 3) NULL COMMENT "",
  `k6` char(5) NULL COMMENT "",
  `k10` date NULL COMMENT "",
  `k11` datetime NULL COMMENT "",
  `k7` varchar(20) NULL COMMENT "",
  `k8` double MAX NULL COMMENT "",
  `k9` float SUM NULL COMMENT ""
) ENGINE=OLAP
AGGREGATE KEY(`k1`, `k2`, `k3`, `k4`, `k5`, `k6`, `k10`, `k11`, `k7`)
COMMENT "OLAP"
PARTITION BY RANGE(`k1`)
(PARTITION p1 VALUES [("-128"), ("-64")),
PARTITION p2 VALUES [("-64"), ("0")),
PARTITION p3 VALUES [("0"), ("64")))
DISTRIBUTED BY HASH(`k1`) BUCKETS 5
PROPERTIES (
"replication_num" = "1",
"in_memory" = "false",
"storage_format" = "DEFAULT"
);

SELECT *
FROM test.test2
WHERE 0 < (
    SELECT MAX(k9)
    FROM test.test2);
```

**Expected behavior**
Execute normally.

This bug is introduced by PR: #3467 3453

morningman:
When there is subquery in where clause, the query will be rewritten to join operation.
And some auxiliary binary predicates will be generated. These binary predicates
will not go through the ExprRewriteRule, so they are not normalized as
"column to the left and constant to the right" format.

We need to take this case into account so that the `canPushDownPredicate()` judgement
will not throw exception.

Fix: #3474

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/Predicate.java
+++ b/fe/src/main/java/org/apache/doris/analysis/Predicate.java
@@ -17,11 +17,12 @@
 
 package org.apache.doris.analysis;
 
-import com.google.common.base.Preconditions;
 import org.apache.doris.catalog.Type;
 import org.apache.doris.common.AnalysisException;
 import org.apache.doris.common.Pair;
 import org.apache.doris.common.Reference;
+
+import com.google.common.base.Preconditions;
 
 public abstract class Predicate extends Expr {
     protected boolean isEqJoinConjunct;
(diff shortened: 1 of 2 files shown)


kangkaisen (approved):
+1, LGTM

PR #3476, opened 2020-05-05
[Bug] Fix bug of Partition prune of constant in predicate

wutiangan:
**1. phenomenon：**

The following two statements are the same, but a query has results and the other query has no results

mysql> select * from (select '积极' as kk1, sum(k2) from table_range where k1 = '2013-01-01' group by kk1)tt where kk1 = '积极';        
+--------+-----------+
| kk1    | sum(`k2`) |
+--------+-----------+
| 积极 |         1 |
+--------+-----------+
1 row in set (0.01 sec)

mysql> select * from (select '积极' as kk1, sum(k2) from table_range where k1 = '2013-01-01' group by kk1)tt where kk1 in ('积极');     
Empty set (0.01 sec)

**2. reason：**

In partition prune, constant in predicate（‘积极’ in ‘积极’） is mistakenly considered to meet partition prune conditions, and mistakenly regarded as partition prune column. Then in partition prune , no partition is considered to meet the requirements, so it is planned to be 0 partition in query planning

@morningman

### DIFF
--- a/fe/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
+++ b/fe/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
@@ -928,6 +928,14 @@ public class SingleNodePlanner {
                 if (!inPredicate.isLiteralChildren() || inPredicate.isNotIn()) {
                     continue;
                 }
+                if ((inPredicate.getChild(0) instanceof LiteralExpr)
+                        || (inPredicate.getChild(0) instanceof CastExpr
+                                && inPredicate.getChild(0).getChild(0) instanceof LiteralExpr)) {
+                    // If child(0) of the in predicate is a constant expression,
+                    // then other children of in predicate should not be used as a condition for partition prune.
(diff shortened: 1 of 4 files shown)


morningman:
Hi @wutiangan Better to create an issue the describe the bug you met. And link the issue to this PR.

fe/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
```diff
@@ -928,6 +928,14 @@ private PartitionColumnFilter createPartitionFilter(SlotDescriptor desc, List<Ex
                 if (!inPredicate.isLiteralChildren() || inPredicate.isNotIn()) {
                     continue;
                 }
+                if ((inPredicate.getChild(0) instanceof LiteralExpr)
```
morningman, line 931: you can use `(inPredicate.getChild(0).unwrapExpr(false) instanceof LiteralExpr)`

morningman (commented):
And please add a ut in `QueryPlanTest.java`

morningman (approved):
LGTM

PR #3478, opened 2020-05-06
replace std::tr1::unordered_map with std::unordered_map

sduzh:


be/src/runtime/descriptors.h
```diff
@@ -18,12 +18,12 @@
 #ifndef DORIS_BE_RUNTIME_DESCRIPTORS_H
 #define DORIS_BE_RUNTIME_DESCRIPTORS_H
 
+#include <ostream>
+#include <unordered_map>
 #include <vector>
-#include <tr1/unordered_map>
 #include <vector>
```
imay, line 24: there are two vector

be/src/runtime/descriptors.h
```diff
@@ -18,12 +18,12 @@
...
-#include <boost/scoped_ptr.hpp>
-#include <ostream>
 
+#include <boost/scoped_ptr.hpp>
```
imay, line 26: this can be replaced with std::unique_ptr too

be/src/runtime/descriptors.h
sduzh: updated

be/src/runtime/descriptors.h
sduzh: It's unused, I removed it.

imay (approved):
+1 LGTM

kangkaisen (approved):
+1, LGTM

chaoyli:
It's better to add some description about this pull request.

PR #3481, opened 2020-05-06
[Bug] Fix bug that push down logic error on semi join

### LINKED ISSUE #3480: [Bug] Push down predicate logic return wrong plan for semi join
**Describe the bug**
Push down predicate logic return wrong plan for semi join

**To Reproduce**
```
select * from
join1 left semi join join2
on join1.id = join2.id and join2.id > 1;
```

the `join2.id > 1` should not be pushed down to table `join1`.

**Expected behavior**
`join2.id` should only push down to table `join2`.

morningman:
For SQL like:
```
select * from
join1 left semi join join2
on join1.id = join2.id and join2.id > 1;
```

the predicate `join2.id > 1` can not be pushed down to table join1.

Fix #3480

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/Analyzer.java
+++ b/fe/src/main/java/org/apache/doris/analysis/Analyzer.java
@@ -956,8 +956,9 @@ public class Analyzer {
 
 
     /**
-     * Return all registered conjuncts that are fully bound by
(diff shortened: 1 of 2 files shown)


morningman:
LGTM

kangkaisen (approved):
+1, LGTM

PR #3485, opened 2020-05-06
Enhance AssertNumRowsNode

### LINKED ISSUE #3483: report error when case when clause  has empty condition
if the subquery of case when clause in when condition and it returns empty set,  then the whole query return empty set,
this is because doris rewrite subquery in case when to cross join, if a subqery return empty the cross join return empty. 
e.g. 
```
select case when (select k1 sum from t1 where  k1 <0) > t2. then t2.k1 else 0 end a from t2;
```
this query will rewrite to 
```
select case when tt.k1 > t2. then t2.k1 else 0 end a from t2, (select k1 sum from t1 where  k1 <0 ) as tt;
```
if `select k1 sum from t1 where  k1 <0` is an empty set, then the query is empty, but in mysql this returns `else 0` branch as the resulet.
in doris we can not handle subquery very well, so shoud report an error when subquery is empty
step 1:
enhance assert row num node to support more asset conditions
step 2:
apply assert rule to subquery in case when clause

yangzhg:
fixes #3483
step1:  Enhance AssertNumRowsNode to support equal, less than, greate than,... assert conditions


be/src/exec/assert_num_rows_node.cpp
```diff
@@ -56,12 +62,46 @@ Status AssertNumRowsNode::get_next(RuntimeState* state, RowBatch* output_batch,
...
+        assert_res = _num_rows_returned >= _desired_num_rows;
+        break;
+    default:
+        break;
+    }
+
+    if (!assert_res) {
+        auto to_string_lamba = [](TAssertion::type assertion) {
+            std::map<int, const char*>::const_iterator it =
+                    _TAssertion_VALUES_TO_NAMES.find(assertion);
```
kangkaisen, line 92: Where is `_TAssertion_VALUES_TO_NAMES ` ?

be/src/exec/assert_num_rows_node.cpp
```diff
@@ -17,19 +17,25 @@
...
-    ObjectPool* pool, const TPlanNode& tnode, const DescriptorTbl& descs) :
-        ExecNode(pool, tnode, descs),
-        _desired_num_rows(tnode.assert_num_rows_node.desired_num_rows),
-        _subquery_string(tnode.assert_num_rows_node.subquery_string) {
+AssertNumRowsNode::AssertNumRowsNode(ObjectPool* pool, const TPlanNode& tnode,
+                                     const DescriptorTbl& descs)
+        : ExecNode(pool, tnode, descs),
+          _desired_num_rows(tnode.assert_num_rows_node.desired_num_rows),
+          _subquery_string(tnode.assert_num_rows_node.subquery_string) {
+    if (tnode.assert_num_rows_node.__isset.assertion) {
```
kangkaisen, line 33: Code format.

be/src/exec/assert_num_rows_node.cpp
yangzhg: `_TAssertion_VALUES_TO_NAMES` is generated by thrift, every `enum` has a variable like  `_xxxx_VALUES_TO_NAMES`

EmmyMiao87 (dismissed):
LGTM

EmmyMiao87 (approved):

kangkaisen (approved):
+1. LGTM

PR #3497, opened 2020-05-07
[BugFix] Fix the bug let fe can show each fragment execute time percent

HappenLee:
like this：
          (Active: 14.133ms, non-child: 93.20%)

ISSUE: #3365


morningman (approved):
LGTM

PR #3511, opened 2020-05-07
[Bug] Fix constant In Predicate result error

### LINKED ISSUE #3510: [Bug] Constant InPredicate result error
**Describe the bug**
`select 1 not in (2, NULL, 1);`

expected: `0`
Actual: `NULL`

morningman:
`select 1 not in (2, NULL, 1);` should return `0`

Fix: #3510

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/InPredicate.java
+++ b/fe/src/main/java/org/apache/doris/analysis/InPredicate.java
@@ -54,6 +54,8 @@ public class InPredicate extends Predicate {
     private static final String IN = "in";
     private static final String NOT_IN = "not_in";
 
+    private static final NullLiteral NULL_LITERAL = new NullLiteral();
+
     public static void initBuiltins(FunctionSet functionSet) {
         for (Type t: Type.getSupportedTypes()) {
             if (t.isNull()) continue;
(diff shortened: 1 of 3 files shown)


fe/src/main/java/org/apache/doris/analysis/InPredicate.java
```diff
@@ -54,6 +54,8 @@
     private static final String IN = "in";
     private static final String NOT_IN = "not_in";
 
+    private static final NullLiteral NULL_LITERAL = new NullLiteral();
```
kangkaisen, line 57: Why use static variable?  I think local variable is enough.

fe/src/test/java/org/apache/doris/planner/QueryPlanTest.java
```diff
@@ -725,4 +725,58 @@ public void testJoinPredicateTransitivityWithSubqueryInWhereClause() throws Exce
         Assert.assertTrue(explainString.contains("CROSS JOIN"));
         Assert.assertTrue(!explainString.contains("PREDICATES"));
     }
+
+    @Test
+    public void testConstantInPredicate() throws Exception {
```
kangkaisen, line 730: please move the test to `ConstantExpressTest`

morningman: OK

fe/src/main/java/org/apache/doris/analysis/InPredicate.java
morningman: This object does not need to have one for every `InPredicate` instance.

kangkaisen (approved):
+1, LGTM

Issue #3479, opened 2020-05-06
query failed when some fields not exist against DOE


blackfox1983:
**Describe the bug**
query failed when i run sql against doris on es. The prompt error is "inner hits node is not an array"

e.g.  
```
CREATE EXTERNAL TABLE `case_replay_for_milimin` (
  `k1` varchar(20) NULL COMMENT "",
  `k2` varchar(20) NULL COMMENT "",
  `k3` varchar(20) NULL COMMENT "", 
) ENGINE=ELASTICSEARCH
COMMENT "ELASTICSEARCH"
PROPERTIES (
"hosts" = "...",
...
"enable_docvalue_scan" = "true"
)
```
This Table has rows like below :
```
> select  k1 from table;
+------+
| k1   |
+------+
| NULL |
| kkk1 |
| NULL |
+------+
```

but, actually it returns a error:
```
> select  k1 from table;
ERROR 1064 (HY000): inner hits node is not an array
```

**To Reproduce**
1. put values into index
```
PUT myindex/_doc/1
{
  "k1": "kkk1"
}
PUT myindex/_doc/2
{
  "k2": "kkk2"
}
PUT myindex/_doc/3
{
  "k3": "kkk3"
}
```
2、fetch fields contain k1,k2,k3. e.g. select  k1,k2,k3 from table;

**Expected behavior**
ignore this situation and return ok

wuyunfeng:
maybe you should emphasize  have no value for one batch_size

blackfox1983:
pr: https://github.com/apache/incubator-doris/pull/3513

PR #3513, opened 2020-05-07
 [Doris On ES]fix bug of query failed in doc_value_mode when fields have none value

BabySid:
#3479 

Here I try to explain the cause of the problem and how to fix it.

**The Cause of The problem**
Take the case in issue(#3479 ) as an example:
The general results are as follows:
```
GET table/_doc/_search
{"query":{"match_all":{}},"stored_fields":"_none_","docvalue_fields":["k1"],"sort":["_doc"],"size":100}

{
  "took": 6,
  "timed_out": false,
  "_shards": {
    ……
  },
  "hits": {
    "total": 3,
    "max_score": null,
    "hits": [
      {
        "_index": "table",
        "_score": null,
        "sort": [
          0
        ]
      },
      {
        "_index": "table",
        "_score": null,
        "fields": {
          "k1": [
            "kkk1"
          ]
        },
        "sort": [
          0
        ]
      },
      {
        "_index": "table",
        "_score": null,
        "sort": [
          0
        ]
      }
    ]
  }
}
```

But in Doris on ES，Be fetched data parallelly on all shards, and use `filter_path` to reduce the network cost. The process will be as follows:
```
GET table/_doc/_search?preference=_shards:1&filter_path=_scroll_id,hits.hits._source,hits.total,_id,hits.hits._source.fields,hits.hits.fields
{"query":{"match_all":{}},"stored_fields":"_none_","docvalue_fields":["k1"],"sort":["_doc"],"size":100}

{
  "hits": {
    "total": 0
  }
}

GET table/_doc/_search?preference=_shards:2&filter_path=_scroll_id,hits.hits._source,hits.total,_id,hits.hits._source.fields,hits.hits.fields
{"query":{"match_all":{}},"stored_fields":"_none_","docvalue_fields":["k1"],"sort":["_doc"],"size":100}
{
  "hits": {
    "total": 1
  }
}

GET table/_doc/_search?preference=_shards:3&filter_path=_scroll_id,hits.hits._source,hits.total,_id,hits.hits._source.fields,hits.hits.fields
{"query":{"match_all":{}},"stored_fields":"_none_","docvalue_fields":["k1"],"sort":["_doc"],"size":100}
{
  "hits": {
    "total": 1,
    "hits": [
      {
        "fields": {
          "k1": [
            "kkk1"
          ]
        }
      }
    ]
  }
}
```
*Scan-Worker On BE which processed result of shard2  will failed.* 

**The reasons are as follows:**
1. "filter_path" causes the hits.hits object not exist.  
2. In the current implementation, if there are some data rows（total > 0）, the hits.hits. object must be an array

**How To Fix it**

Two Method:
1. modify "filter_path" to contain the hits.  
Pros: Fixed Code is very simple
Cons: More network cost
2. Deal with the case where fields are missing in a batch. 
Pros: No loss of performance
Cons: Code is more complex 

Performance first, I use Method2.

**Design**
1. Add a variable "_doc_value_mode" into Class "EsScrollParser" to =indicate whether the data processed by this parser is doc_value_mode or not.
2. "_doc_value_mode" is passed from ESScollReader <- ESScanner <- ScrollQueryBuilder::build() that determines whether DSL is enable doc_value_mode
3. When hits.hits of response from ES is empty and total > 0. We know there are data lines, but the corresponding fields do not exist. EsScrollParser will use "_doc_value_mode"  and _total to construct _total lines which fields are assigned with 'NULL'

### DIFF
--- a/be/src/exec/es/es_scroll_parser.cpp
+++ b/be/src/exec/es/es_scroll_parser.cpp
@@ -198,12 +198,12 @@ static Status get_float_value(const rapidjson::Value &col, PrimitiveType type, v
     return Status::OK();
 }
 
-ScrollParser::ScrollParser(bool use_doc_value) :
+ScrollParser::ScrollParser(bool doc_value_mode) :
     _scroll_id(""),
     _total(0),
     _size(0),
     _line_index(0),
-    _use_doc_value(use_doc_value) {
+    _doc_value_mode(doc_value_mode) {
 }
 
 ScrollParser::~ScrollParser() {
@@ -289,7 +289,7 @@ Status ScrollParser::fill_tuple(const TupleDescriptor* tuple_desc,
 
     if (_size <= 0 || _line_index >= _size) {
         // _source is fetched from E.S.
-        if (!_use_doc_value) {
+        if (!_doc_value_mode) {
             return Status::OK();
         }
         
--- a/be/src/exec/es/es_scroll_parser.cpp
+++ b/be/src/exec/es/es_scroll_parser.cpp
@@ -255,7 +255,7 @@ Status ScrollParser::parse(const std::string& scroll_result, bool exactly_once) 
         // put index/_doc/1 {"k2":"123"}
         // put index/_doc/2 {"k3":"123}
         // then we use sql `select k1 from table`
-        // what E.S. return is like this: {hits: {total:2}
+        // what ES return is like this: {hits: {total:2}
         return Status::OK();
     }
     const rapidjson::Value &inner_hits_node = outer_hits_node[FIELD_INNER_HITS];
@@ -288,12 +288,12 @@ Status ScrollParser::fill_tuple(const TupleDescriptor* tuple_desc,
     *line_eof = true;
 
     if (_size <= 0 || _line_index >= _size) {
-        // _source is fetched from E.S.
+        // _source is fetched from ES
         if (!_doc_value_mode) {
             return Status::OK();
         }
         
-        // _fields(doc_value) is fetched from E.S.
+        // _fields(doc_value) is fetched from ES
         if (_total <= 0 || _line_index >= _total) {
             return Status::OK();
         }
--- a/be/src/exec/es/es_scroll_parser.cpp
+++ b/be/src/exec/es/es_scroll_parser.cpp
@@ -198,11 +198,12 @@ static Status get_float_value(const rapidjson::Value &col, PrimitiveType type, v
     return Status::OK();
 }
 
-ScrollParser::ScrollParser() :
+ScrollParser::ScrollParser(bool use_doc_value) :
     _scroll_id(""),
     _total(0),
     _size(0),
-    _line_index(0) {
+    _line_index(0),
+    _use_doc_value(use_doc_value) {
 }
 
 ScrollParser::~ScrollParser() {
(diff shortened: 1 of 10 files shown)


morningman:
@wuyunfeng Please help to review this PR.

wuyunfeng:
Can you  prefix [Doris On ES] on this PR title?

be/src/exec/es/es_scroll_parser.cpp
```diff
@@ -275,7 +286,32 @@ int ScrollParser::get_total() {
 Status ScrollParser::fill_tuple(const TupleDescriptor* tuple_desc, 
             Tuple* tuple, MemPool* tuple_pool, bool* line_eof, const std::map<std::string, std::string>& docvalue_context) {
     *line_eof = true;
+
     if (_size <= 0 || _line_index >= _size) {
```
wuyunfeng, line 290: maybe _size should be `total` ? currently, the _size = inner_hits.size()

blackfox1983: i don't think so.
This problem is mainly triggered when _size = 0 and _total > 0。if use `_total`，parser() will return OK() and `_size` remain 0, the problem would not exist.

blackfox1983: > 
> 
> maybe _size should be `total` ? currently, the _size = inner_hits.size()

i don't think so.
This problem is mainly triggered when _size = 0 and _total > 0。if use `_total`，parser() will return OK() and `_size` remain 0, the problem would not exist.

wuyunfeng: Maybe the concept `size` should keep consistent with before

be/src/exec/es/es_scroll_parser.cpp
```diff
@@ -275,7 +286,32 @@ int ScrollParser::get_total() {
...
+        // _source is fetched from E.S.
+        if (!_use_doc_value) {
+            return Status::OK();
+        }
+        
+        // _fields(doc_value) is fetched from E.S.
```
wuyunfeng, line 296:
```suggestion
        // _fields(doc_value) is fetched from ES
```

be/src/exec/es/es_scroll_parser.cpp
```diff
@@ -275,7 +286,32 @@ int ScrollParser::get_total() {
...
+        if (_total <= 0 || _line_index >= _total) {
+            return Status::OK();
+        }
+       
+       
+        // here is operations for `enable_doc_value_scan`.
```
wuyunfeng, line 302:
```suggestion
        // here is operations for `use_doc_value`.
```

be/src/exec/es/es_scroll_parser.h
```diff
@@ -50,5 +50,7 @@ class ScrollParser {
 
     rapidjson::Document _document_node;
     rapidjson::Value _inner_hits_node;
+
+    bool _use_doc_value;
```
wuyunfeng, line 54: maybe use doc_value_mode is more suitable？
in future, we can use different parser for _source or doc_value mode

wuyunfeng (dismissed):
LGTM, and I left some minor comment.

@blackfox1983  Can you give some example to show How this PR works on your issue or your PR comment?

@imay

wuyunfeng (approved):
LGTM Thanks @blackfox1983 

@imay  Can you spare some time to review this PR?

imay (approved):
LGTM

Issue #3402, opened 2020-04-27
Can we give user a error msg when they create a olap table with medium which is not exists in cluster?


WingsGo:
**Is your feature request related to a problem? Please describe.**
As shown in the document below, if user create a olap table with ssd, I think we should give them a error tip that the cluster has no avaliable ssd storage, instread of store the data in hdd disk.

Note that this parameter is just a "best effort" setting when storage_medium is specified. Even if no SSD storage media is set in the cluster, no error is reported and it is automatically stored in the available data directory

morningman:
Do you mean to reject the operation, or just return a tip to user but the creation still succeed?

WingsGo:
I mean reject the operation and give the user error msg and solutions

morningman:
> I mean reject the operation and give the user error msg and solutions

I agreed. I've met this situation that user create table with SSD, but there is no SSD in cluster, this will cause lots of unnecessary migration tasks.

WingsGo:
> > I mean reject the operation and give the user error msg and solutions
> 
> I agreed. I've met this situation that user create table with SSD, but there is no SSD in cluster, this will cause lots of unnecessary migration tasks.

Yes, we meet the same situation, I will do this work.

WingsGo:
@morningman I found that CheckDataSizeQuota will only check by logic but not check by real disk usage.I have some questions:
1. One solution is get real disk space avaliable group by storage medium, and check is there are avaliable space for specify storage medium  when create or alter table/partition, But the solution has a problem that it may cannot constrain the existing partition, and I don't know if the already exist partition exceed the limit cause by data loading, how doris deal with it or just ignore the situation, because stream load also doesn't limited by the db quota in fact.
2. should we refacor the checkDataSizeQuota logic to limit the quota by real usage?And if the real disk space is not insufficient, how doris deal with it?

PR #3519, opened 2020-05-08
Check backend disk has available capactiy by storage medium before create table

WingsGo:
Currently we choose BE random without check disk is available, the create table will failed until create tablet task is sent to BE and BE will check is there has avaliable capcity to create tablet.So Check
backend disk available by storage medium will reduce unnecessary RPC call.


WingsGo:
For #3402

fe/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -4894,8 +4894,7 @@ public void renameTable(Database db, OlapTable table, TableRenameClause tableRen
 
         // check if rollup has same name
         if (table.getType() == TableType.OLAP) {
```
morningman, line 4896: We can just remove the `if (table.getType() == TableType.OLAP)`. The argument is already `OlapTable`

fe/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -4894,8 +4894,7 @@ public void renameTable(Database db, OlapTable table, TableRenameClause tableRen
...
-            OlapTable olapTable = (OlapTable) table;
-            for (String idxName: olapTable.getIndexNameToId().keySet()) {
+            for (String idxName: table.getIndexNameToId().keySet()) {
```
morningman, line 4897:
```suggestion
            for (String idxName : table.getIndexNameToId().keySet()) {
```

fe/src/main/java/org/apache/doris/system/Backend.java
```diff
@@ -346,6 +347,23 @@ public double getMaxDiskUsedPct() {
         return maxPct;
     }
 
+    public boolean checkDiskExceedLimitByStorageMedium(TStorageMedium storageMedium) {
```
morningman, line 350: Add comment to explain the return value of this method.
the name `checkDiskExceedLimitByStorageMedium` is a little bit confused.

fe/src/main/java/org/apache/doris/system/Backend.java
```diff
@@ -346,6 +347,23 @@ public double getMaxDiskUsedPct() {
...
+        // no available disk with storage medium
+        if (getDiskNumByStorageMedium(storageMedium) <= 0) {
+            return true;
+        }
+        ImmutableMap<String, DiskInfo> disks = disksRef.get();
+        boolean exceedLimit = false;
+        for (DiskInfo diskInfo : disks.values()) {
```
morningman, line 357: Is this logic correct? If all disks are offline or is not with specified storage medium, then this method will return false?

fe/src/main/java/org/apache/doris/system/SystemInfoService.java
```diff
@@ -443,7 +390,7 @@ public void releaseBackends(String clusterName, boolean isReplay) {
         }
 
         List<List<Backend>> hostList = Lists.newArrayList(hostBackendsMapInCluster.values());
-        Collections.sort(hostList, hostBackendsListComparator);
+        hostList.sort(hostBackendsListComparator);
```
morningman, line 393: Actually, the method `calculateDecommissionBackends()` is deprecated.
So could you please add `@Derepcated` to this method?

fe/src/main/java/org/apache/doris/system/SystemInfoService.java
```diff
@@ -729,8 +639,18 @@ public void releaseBackends(String clusterName, boolean isReplay) {
...
     // use synchronized to run serially
     public synchronized List<Long> seqChooseBackendIds(int backendNum, boolean needAlive, boolean isCreate,
-            String clusterName) {
-        long lastBackendId = -1L;
+                                                       String clusterName) {
+        return seqChooseBackendIds(backendNum, needAlive, isCreate, clusterName, getClusterBackends(clusterName));
+    }
+    public synchronized List<Long> seqChooseBackendIdsByStorageMedium(int backendNum, boolean needAlive, boolean isCreate,
+                                                                      String clusterName, TStorageMedium storageMedium) {
+        final List<Backend> backends = getClusterBackends(clusterName).stream().filter(v -> !v.checkDiskExceedLimitByStorageMedium(storageMedium)).collect(Collectors.toList());
```
morningman, line 647: This is not user friendly. For example, if all backends not meet the `checkDiskExceedLimitByStorageMedium()`, the user will only get the msg: "Not enough backends", but he will not know why there is no enough backend.
So could you please find a way to return the detail msg to user about the reason of failure?

morningman (changes requested):
By the way, the code refactor can be done in another PR. Or the change list is too long that will be hard for the reviewer to review the core logic of your change.

And could you add some docs for this new logic? Can be in `CREATE TABLE.md`

fe/src/main/java/org/apache/doris/system/SystemInfoService.java
WingsGo: actually, I change the detail msg to `throw new DdlException("Failed to find enough host in all backends with storage medium is " + storageMedium + ". need: " + replicationNum`

fe/src/main/java/org/apache/doris/system/SystemInfoService.java
WingsGo: done

fe/src/main/java/org/apache/doris/system/Backend.java
WingsGo: yes, you are right, I fixed it, and should we change the logic in be, just remove the judgement `_available_storage_medium_type_count == 1`?

```
for (auto& it : _store_map) {
            if (it.second->is_used()) {
                if (_available_storage_medium_type_count == 1
                    || it.second->storage_medium() == storage_medium) {
                    stores.push_back(it.second);
                }
            }
        }
```

fe/src/main/java/org/apache/doris/system/Backend.java
WingsGo: done

WingsGo:
> By the way, the code refactor can be done in another PR. Or the change list is too long that will be hard for the reviewer to review the core logic of your change.
> 
> And could you add some docs for this new logic? Can be in `CREATE TABLE.md`

Ok, I will remove the code refactor next time, it's because Catalog.java contains the core logic and I didn't remove them.By the way, doc is updated.

WingsGo:
Hi, @morningman I have remove the code refactor to other PR, this PR is ready for review~

morningman (commented):
I think we need to add a FE config like `enable_strict_storage_medium_check`, and default is false, and
can also be modified at runtime.
Because I am afraid if I upgrade to this version, it will block some existing workload which is trying to create SSD table without SSD storage medium.

After adding the config, don't forget to update the config document too.

WingsGo:
> I think we need to add a FE config like `enable_strict_storage_medium_check`, and default is false, and
> can also be modified at runtime.
> Because I am afraid if I upgrade to this version, it will block some existing workload which is trying to create SSD table without SSD storage medium.
> 
> After adding the config, don't forget to update the config document too.

OK, done, and In order to keep the origin behavior, I remove the check in BE, just check it in FE.

fe/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -1082,5 +1082,11 @@
      */
     @ConfField(mutable = true, masterOnly = true)
     public static boolean drop_backend_after_decommission = true;
+
+    /**
+     *
+     */
```
morningman, line 1088: missing coment?

WingsGo: done

morningman (approved):
LGTM

PR #3523, opened 2020-05-08
Refactor some redunant code && Replace some UT by UtFrameUtils

WingsGo:
This CL have no logic changed, just do some code refactor and use new UtFrameWork to replace some old UT.

### DIFF
--- a/fe/src/main/java/org/apache/doris/backup/RestoreJob.java
+++ b/fe/src/main/java/org/apache/doris/backup/RestoreJob.java
@@ -172,11 +172,7 @@ public class RestoreJob extends AbstractJob {
     }
 
     public synchronized boolean finishTabletSnapshotTask(SnapshotTask task, TFinishTaskRequest request) {
-        Preconditions.checkState(task.getJobId() == jobId);
-        Preconditions.checkState(task.getDbId() == dbId);
-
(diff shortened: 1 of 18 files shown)


imay (commented):
LGTM, 
I think @morningman should review this PR

fe/src/main/java/org/apache/doris/backup/RestoreJob.java
```diff
@@ -806,7 +739,54 @@ private void checkAndPrepareMeta() {
         // No log here, PENDING state restore job will redo this method
         LOG.info("finished to prepare meta and send snapshot tasks, num: {}. {}",
                  batchTask.getTaskNum(), this);
-        return;
+    }
+
+    private boolean backupReplicasNotEqual(PartitionInfo localPartInfo, Partition localPartition, Table localTbl,
```
morningman, line 744: I think the part of `generate file mapping` can be a separated method not including in this method.
Or the name of this method can be confused.

WingsGo: can we change the method name to `genFileMappingWhenBackupReplicasEqual`?

morningman: Sorry for review delay. This name looks good to me.

morningman (approved):
LGTM

PR #3524, opened 2020-05-08
[ut] disable backgrounp scan context gc to speed up unit test

acelyc111:
Each test case in ExternalScanContextMgrTest may cost 1 minitue
which is too long, we'd better disable backgrounp scan context
gc to speed up unit test.


imay (approved):
LGTM

Issue #3491, opened 2020-05-06
Compare string type and numerical type will cause full table scan


WingsGo:
**Describe the bug**
The sql like `SELECT * FROM tbl WHERE olap_date='20200501'` will not only scan partiiton `p20200501` but scan the full table, it is because `olap_date` is int type and `'20200501'` is string literal, in following logic, olap_date and string literal will convert to double and compare, which lead to a full table scan.

https://github.com/apache/incubator-doris/blob/d64704599d338de290972c1324903cf03d7bab5c/fe/src/main/java/org/apache/doris/analysis/BinaryPredicate.java#L280-L310

Solutions:
I think we should throw a DdlException when compare string literal to a numerical type to avoid this situation, Is there any solutions better?

kangkaisen:
@WingsGo Hi, I think throw a DdlException is not acceptable.
Most of query engine support int compare string. But maybe don't choose double as the common type. You could investigate Presto, Hive Or Spark. Thanks.

imay:
First of all, throwing an exception is not acceptable. Because it will make many already exist queries fail.
I think it is viable approach to handle this case in a specified case.

WingsGo:
@imay @kangkaisen can we check the binary compare type, if one type is string and the other type is numerical such as int, bigint...., we just return the numercal type in `getCmpType()` function?

WingsGo:
There is another case cause by the problem.for example: if we convert string and bigint to double, it may cause query result not match , because 
```
Double d1 = new Double(2882303761517473127L);
Double d2 = new Double("2882303761517473267");
System.out.println(d1.equals(d2));    // true
```

```
MySQL [(none)]> select 2882303761517473127 = "2882303761517473127";
+-----------------------------------------------+
| 2.8823037615174733E18 = 2.8823037615174733E18 |
+-----------------------------------------------+
|                                             1 |
+-----------------------------------------------+
1 row in set (0.07 sec)

MySQL [(none)]> select 2882303761517473127 = "2882303761517473267";
+-----------------------------------------------+
| 2.8823037615174733E18 = 2.8823037615174733E18 |
+-----------------------------------------------+
|                                             1 |
+-----------------------------------------------+
1 row in set (0.07 sec)
```

imay:
> @imay @kangkaisen can we check the binary compare type, if one type is string and the other type is numerical such as int, bigint...., we just return the numercal type in `getCmpType()` function?

[...]

PR #3527, opened 2020-05-08
[SQL]When compare StringLiteral and Numerical, cast them to Numerical type

WingsGo:
When compare StringLiteral and Numerical type, Doris will convert them both to Double Type, which will lead some unexpected behavior. For example: `select 2882303761517473127 = "2882303761517473267";` will be true.

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/BinaryPredicate.java
+++ b/fe/src/main/java/org/apache/doris/analysis/BinaryPredicate.java
@@ -101,8 +101,9 @@ public class BinaryPredicate extends Predicate implements Writable {
                     return LE;
                 case EQ_FOR_NULL:
                     return this;
+                default:
+                    return null;
             }
-            return null;
         }
 
         public Operator converse() {
--- a/fe/src/main/java/org/apache/doris/analysis/BinaryPredicate.java
+++ b/fe/src/main/java/org/apache/doris/analysis/BinaryPredicate.java
@@ -263,15 +263,9 @@ public class BinaryPredicate extends Predicate implements Writable {
 
     private boolean canCompareDate(PrimitiveType t1, PrimitiveType t2) {
         if (t1.isDateType()) {
-            if (t2.isDateType() || t2.isStringType()) {
-                return true;
-            }
-            return false;
+            return t2.isDateType() || t2.isStringType();
         } else if (t2.isDateType()) {
-            if (t1.isStringType()) {
-                return true;
-            }
-            return false;
+            return t1.isStringType();
         } else {
             return false;
         }
@@ -304,6 +298,10 @@ public class BinaryPredicate extends Predicate implements Writable {
         if ((t1 == PrimitiveType.BIGINT || t1 == PrimitiveType.LARGEINT)
                 && (t2 == PrimitiveType.BIGINT || t2 == PrimitiveType.LARGEINT)) {
             return Type.LARGEINT;
+        }
+
+        if ((t1.isNumericType() && t2.isStringType()) || (t1.isStringType() && t2.isNumericType())) {
+            return t1.isNumericType() ? Type.fromPrimitiveType(t1) : Type.fromPrimitiveType(t2);
         }
 
         return Type.DOUBLE;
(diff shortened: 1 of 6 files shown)


WingsGo:
For #3491

imay:
@WingsGo 
You should create a separated issue to track this problem. And you should explain in what way you want to change fix it.
Because this CL will influence the behavior of Doris, we must agree on the way of changing before writing code.

WingsGo:
> @WingsGo
> You should create a separated issue to track this problem. And you should explain in what way you want to change fix it.
> Because this CL will influence the behavior of Doris, we must agree on the way of changing before writing code.

OK, I create another issue #3528  to track this problem

morningman (commented):
Add unit test for this CL.

WingsGo:
> Add unit test for this CL.

Done

WingsGo:
> Add unit test for this CL.

By the way. I'm confused with our CI enviorment, what's the relationship between UT and NewDorisTest, And how they do the UT, can we unify the CI enviornment and how the CI scripts works?So that we can more easier to find out what's the problem of our PR?

WingsGo:
Hi, @morningman @imay , this PR is ready for review

fe/src/main/java/org/apache/doris/analysis/BinaryPredicate.java
```diff
@@ -263,15 +263,9 @@ public void vectorizedAnalyze(Analyzer analyzer) {
...
-                return true;
-            }
-            return false;
+            return t2.isDateType() || t2.isStringType();
         } else if (t2.isDateType()) {
-            if (t1.isStringType()) {
-                return true;
-            }
-            return false;
+            return t1.isStringType();
```
wutiangan, line 268: why not modify to  “return t1.isDateType() || t1.isStringType()”

WingsGo: I will fix it.

fe/src/main/java/org/apache/doris/analysis/BinaryPredicate.java
```diff
@@ -306,6 +300,10 @@ private Type getCmpType() {
             return Type.LARGEINT;
         }
 
+        if ((t1.isNumericType() && t2.isStringType()) || (t1.isStringType() && t2.isNumericType())) {
+            return t1.isNumericType() ? Type.fromPrimitiveType(t1) : Type.fromPrimitiveType(t2);
+        }
```
wutiangan, line 305: what about if "LargeIntType"(16byte) compare "DoubelType"(8byte) ? 
what about if "DecimalValuetType"(16byte) compare "DoubelType"(8byte)  ? 
what about if "LargeIntType" compare "DecimalType" ? 
what about if "DecimalType" compare "DecimalType" ?

WingsGo: all the above will both cast to a middle type `DoubleType` and to compare, I just handle the situation that compare `Numerical` and  `StringType`

wutiangan: At least equivalent comparison (decimaltype to decimaltype) does not need to be converted to double type, You can fix it by the way.

WingsGo: if they are both DecimalType, the function will return DecimalType, it's already checked in the function.

wutiangan: Can you explain in detail if both input parameters are decimal type.， why this funciton return decimal type？

WingsGo: you can get it in the following code
```
if ((t1 == PrimitiveType.BIGINT || t1 == PrimitiveType.DECIMAL)
                && (t2 == PrimitiveType.BIGINT || t2 == PrimitiveType.DECIMAL)) {
        return Type.DECIMAL;
}
```

WingsGo:
@morningman Can we just specify the case in partition column?because partition column only support Integer, date, datetime, so it could avoid full table scan at least?

Issue #3530, opened 2020-05-08
Group by query will core in Debug mode BE


kangkaisen:
**Describe the bug**
Run any group by query in Debug mode BE
```
F0508 18:39:32.956269 32493 reservation_tracker.cc:286] Check failed: tracker->mem_tracker_ == nullptr || !tracker->mem_tracker_->has_limit()

    @          0x3c0fe81  google::DumpStackTraceAndExit()
    @          0x3c06e5d  google::LogMessage::Fail()
    @          0x3c08ce4  google::LogMessage::SendToLog()
    @          0x3c06984  google::LogMessage::Flush()
    @          0x3c096e9  google::LogMessageFatal::~LogMessageFatal()
    @          0x1d82380  doris::ReservationTracker::TransferReservationTo()
    @          0x1d59c9c  doris::BufferPool::ClientHandle::TransferReservationFrom()
    @          0x1ef11c6  doris::InitialReservations::Claim()
    @          0x2cfe826  doris::ExecNode::claim_buffer_reservation()
    @          0x30a2f4b  doris::PartitionedAggregationNode::open()
    @          0x1e059fc  doris::PlanFragmentExecutor::open_internal()
    @          0x1e07e05  doris::PlanFragmentExecutor::open()
    @          0x1c4a903  doris::FragmentExecState::execute()
    @          0x1c5068d  doris::FragmentMgr::exec_actual()
    @          0x1c6a6a1  boost::detail::function::void_function_obj_invoker0<>::invoke()
    @          0x1b959de  doris::PriorityThreadPool::work_thread()
    @          0x1b75c69  boost::detail::thread_data<>::run()
    @          0x3bb91dd  thread_proxy
    @     0x7fb3a2f5de65  start_thread
    @     0x7fb3a236288d  __clone
```

This DCHECK is not reasonable. Because when query the limit for mem_tracker is `2147483648`.

We set memory limit in:
```
    // NOTE: this MemTracker only for olap
    _mem_tracker.reset(
            new MemTracker(bytes_limit, "fragment mem-limit", _exec_env->process_mem_tracker()));
    _runtime_state->set_fragment_mem_tracker(_mem_tracker.get());
```

So, we should remove this DCHECK.

PR #3531, opened 2020-05-08
Remove unreasonable DCHECK for mem_tracker

kangkaisen:
For #3530

### DIFF
--- a/be/src/runtime/bufferpool/reservation_tracker.cc
+++ b/be/src/runtime/bufferpool/reservation_tracker.cc
@@ -281,14 +281,11 @@ bool ReservationTracker::TransferReservationTo(ReservationTracker* other, int64_
   // so this is all atomic.
   for (ReservationTracker* tracker : other_path_to_common) {
     tracker->UpdateReservation(bytes);
-    // We don't handle MemTrackers with limit in this function - this should always
-    // succeed.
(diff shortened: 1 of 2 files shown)


be/src/runtime/bufferpool/reservation_tracker.cc
```diff
@@ -281,14 +281,11 @@ bool ReservationTracker::TransferReservationTo(ReservationTracker* other, int64_
   // so this is all atomic.
   for (ReservationTracker* tracker : other_path_to_common) {
     tracker->UpdateReservation(bytes);
-    // We don't handle MemTrackers with limit in this function - this should always
-    // succeed.
-    DCHECK(tracker->mem_tracker_ == nullptr || !tracker->mem_tracker_->has_limit());
```
morningman, line 286: I also found this bug when running in LSAN mode.
But the comment here is strange: `We don't handle MemTrackers with limit in this function`.
And if it has limit, the following `DCHECK(success)` should be removed too?
And what if it return false?

kangkaisen: We don't need to remove DCHECK(success).

I have removed `DCHECK(!_all_trackers[i]->has_limit());` in consume_local. So tracker->TryConsumeFromMemTracker will success.

morningman: OK

morningman (approved):
LGTM

Issue #3538, opened 2020-05-09
admin set frontend config will return an unexpected exception


WingsGo:
when execute Ddl `admin set frontend config` without parameters, doris will throw a Unexpected exception, like following

```
ERROR 1064 (HY000): Unexpected exception: null
```
The error log is:

```
2020-05-09 15:46:45,247 WARN 33036 [StmtExecutor.handleDdlStmt():877] DDL statement(admin set frontend config) process failed.
java.lang.IllegalStateException: null
        at com.google.common.base.Preconditions.checkState(Preconditions.java:134) ~[guava-15.0.jar:?]
        at org.apache.doris.catalog.Catalog.setConfig(Catalog.java:6330) ~[palo-fe.jar:?]
        at org.apache.doris.qe.DdlExecutor.execute(DdlExecutor.java:207) ~[palo-fe.jar:?]
        at org.apache.doris.qe.StmtExecutor.handleDdlStmt(StmtExecutor.java:868) [palo-fe.jar:?]
        at org.apache.doris.qe.StmtExecutor.execute(StmtExecutor.java:280) [palo-fe.jar:?]
        at org.apache.doris.qe.ConnectProcessor.handleQuery(ConnectProcessor.java:182) [palo-fe.jar:?]
        at org.apache.doris.qe.ConnectProcessor.dispatch(ConnectProcessor.java:314) [palo-fe.jar:?]
        at org.apache.doris.qe.ConnectProcessor.processOnce(ConnectProcessor.java:476) [palo-fe.jar:?]
        at org.apache.doris.qe.ConnectProcessor.loop(ConnectProcessor.java:486) [palo-fe.jar:?]
        at org.apache.doris.qe.ConnectScheduler$LoopHandler.run(ConnectScheduler.java:178) [palo-fe.jar:?]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) [?:?]
        at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
        at java.lang.Thread.run(Thread.java:834) [?:?]

```

PR #3539, opened 2020-05-09
Throw DdlException when use admin set frontend config

WingsGo:
When execute Ddl `admin set frontend config`. User will get msg `ERROR 1064 (HY000): Unexpected exception: null`, it is confusing, we should give user a more friendly tip info.


WingsGo:
For #3538

fe/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -6385,7 +6385,10 @@ public void replayDropFunction(FunctionSearchDesc functionSearchDesc) {
 
     public void setConfig(AdminSetConfigStmt stmt) throws DdlException {
         Map<String, String> configs = stmt.getConfigs();
-        Preconditions.checkState(configs.size() == 1);
+        // configs can not be null
+        if (configs.size() != 1) {
```
morningman, line 6389: This should be done in `AdminSetConfigStmt`'s `analyze` method.

WingsGo: done

kangkaisen (dismissed):
+1， LGTM

morningman (dismissed):
LGTM

yangzhg (changes requested):
`fe/src/test/java/org/apache/doris/analysis/AminSetConfigStmtTest` no file extension;

WingsGo:
> `fe/src/test/java/org/apache/doris/analysis/AminSetConfigStmtTest` no file extension;

done

imay (approved):

Issue #3541, opened 2020-05-09
Throw a UserException when miss plugin's md5 file


WingsGo:
When I install audit plugin, doris throw an unexpected exception, I found the error log in following, It looks like I miss some file, I finally found in code that doris will search a md5 file to check plugin file, but the information is not mentioned in docs, So I think we should give user a solution to avoid waster time to found out what's happening.

```
java.io.FileNotFoundException: http://xxx/doris-audit/auditloader.zip.md5
        at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1909) ~[?:?]
        at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1509) ~[?:?]
        at org.apache.doris.plugin.PluginZip.openUrlInputStream(PluginZip.java:140) ~[palo-fe.jar:?]
        at org.apache.doris.plugin.PluginZip.downloadRemoteZip(PluginZip.java:122) ~[palo-fe.jar:?]
        at org.apache.doris.plugin.PluginZip.downloadZip(PluginZip.java:100) ~[palo-fe.jar:?]
        at org.apache.doris.plugin.PluginZip.extract(PluginZip.java:71) ~[palo-fe.jar:?]
        at org.apache.doris.plugin.DynamicPluginLoader.getPluginInfo(DynamicPluginLoader.java:84) ~[palo-fe.jar:?]
        at org.apache.doris.plugin.PluginMgr.installPlugin(PluginMgr.java:116) ~[palo-fe.jar:?]
        at org.apache.doris.catalog.Catalog.installPlugin(Catalog.java:6452) ~[palo-fe.jar:?]
        at org.apache.doris.qe.DdlExecutor.execute(DdlExecutor.java:213) ~[palo-fe.jar:?]
        at org.apache.doris.qe.StmtExecutor.handleDdlStmt(StmtExecutor.java:868) [palo-fe.jar:?]
        at org.apache.doris.qe.StmtExecutor.execute(StmtExecutor.java:280) [palo-fe.jar:?]
        at org.apache.doris.qe.ConnectProcessor.proxyExecute(ConnectProcessor.java:431) [palo-fe.jar:?]
        at org.apache.doris.service.FrontendServiceImpl.forward(FrontendServiceImpl.java:576) [palo-fe.jar:?]
[...]

PR #3542, opened 2020-05-09
Throw a UserException when miss plugin's md5 file

WingsGo:


WingsGo:
For #3541

morningman (commented):
Actually, we do not require an m5sum when installing plugin from local dir, which is different from http. So, How about just ignore the md5sum if it does not exist?

imay:
> Actually, we do not require an m5sum when installing plugin from local dir, which is different from http. So, How about just ignore the md5sum if it does not exist?

I think we can keep it to avoid install wrong plugin which is not user's intent.

imay (approved):
LGTM

PR #3558, opened 2020-05-12
report error when subquery in case-when returns empty set

### LINKED ISSUE #3483: report error when case when clause  has empty condition

yangzhg:
fix #3483 step2, apply assert rule to subquery in case-when clause

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/CaseExpr.java
+++ b/fe/src/main/java/org/apache/doris/analysis/CaseExpr.java
@@ -166,6 +166,9 @@ public class CaseExpr extends Expr {
         // Go through when/then exprs and determine compatible types.
         for (int i = loopStart; i < loopEnd; i += 2) {
             Expr whenExpr = children.get(i);
+            if (!whenExpr.getType().isScalarType()) {
+                throw new AnalysisException("subquery in case-when must return scala type");
+            }
             if (hasCaseExpr) {
                 // Determine maximum compatible type of the case expr,
                 // and all when exprs seen so far. We will add casts to them at the very end.
@@ -186,6 +189,9 @@ public class CaseExpr extends Expr {
             // Determine maximum compatible type of the then exprs seen so far.
             // We will add casts to them at the very end.
             Expr thenExpr = children.get(i + 1);
+            if (!thenExpr.getType().isScalarType()) {
+                throw new AnalysisException("subquery in case-when must return scala type");
+            }
             returnType = analyzer.getCompatibleType(returnType, lastCompatibleThenExpr, thenExpr);
             lastCompatibleThenExpr = thenExpr;
         }
--- a/fe/src/main/java/org/apache/doris/analysis/CaseExpr.java
+++ b/fe/src/main/java/org/apache/doris/analysis/CaseExpr.java
@@ -24,6 +24,7 @@ import org.apache.doris.thrift.TExprNode;
 import org.apache.doris.thrift.TExprNodeType;
 
 import com.google.common.base.Preconditions;
+import com.google.common.base.Predicates;
 import com.google.common.collect.Lists;
 
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


fe/src/main/java/org/apache/doris/analysis/CaseExpr.java
```diff
@@ -186,6 +189,9 @@ public void analyzeImpl(Analyzer analyzer) throws AnalysisException {
             // Determine maximum compatible type of the then exprs seen so far.
             // We will add casts to them at the very end.
             Expr thenExpr = children.get(i + 1);
+            if (!thenExpr.getType().isScalarType()) {
+                throw new AnalysisException("subquery in case-when must return scala type");
+            }
```
wutiangan, line 194: why do you only thow exception where expr is ‘when expr’ and ‘then expr’？
Whether to throw an exception when ‘case expr’ and ‘else expr’ is not scarlar type？

fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
```diff
@@ -1391,9 +1391,8 @@ private void rewriteSelectList(ExprRewriter rewriter) throws AnalysisException {
             CaseExpr caseExpr = (CaseExpr) item.getExpr();
 
             int childIdx = 0;
-            if (caseExpr.hasCaseExpr()
-                    && caseExpr.getChild(childIdx++).contains(Predicates.instanceOf(Subquery.class))) {
-                throw new AnalysisException("Only support subquery in binary predicate in case statement.");
+            if (caseExpr.hasCaseExpr()) {
+                childIdx++;
             }
             while (childIdx + 2 <= caseExpr.getChildren().size()) {
```
wutiangan, line 1397: why whenExpr must be BinaryPredicate? can we support query “select case (select sex from table limit 1) when ''1” then 'male' else 'female'？

wutiangan, line 1397: maybe in 1397 line code , "<=" need to modify "<".
for example: case when expr0 then expr1 else expr2.   expr2 will be consider 'when expr' in In the following code

fe/src/main/java/org/apache/doris/analysis/CaseExpr.java
yangzhg: if `when exp`or `case expr` has subquery and subquer returns non-scalar type, the error will be thrown by `analyzer.getCompatibleType` and the message is  weird, too difficult to understand, so i throw a message here

fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
yangzhg: this is the previous code

yangzhg: after the  loop  the `childIdx` is 2  `childIdx + 2` is 4 not <= `caseExpr.getChildren().size()` (3) so expr2 will not be consider 'when expr' in In the following code

EmmyMiao87 (dismissed):
LGTM

EmmyMiao87 (approved):
LGTM

Issue #3492, opened 2020-05-06
[bug] count distinct(i, j) calc error


wutiangan:
- How to reproduce

mysql> select * from t;
+------+------+------+
| i    | j    | k    |
+------+------+------+
|    1 |    1 |    1 |
|    1 |    2 |    1 |
|    2 |    1 |    1 |
|    2 |    2 |    2 |
+------+------+------+
4 rows in set (0.00 sec)

mysql> select count(distinct i), count(distinct j), count(distinct i,j) from t;
+---------------------+---------------------+--------------------------+
| count(DISTINCT `i`) | count(DISTINCT `j`) | count(DISTINCT `i`, `j`) |
+---------------------+---------------------+--------------------------+
|                   2 |                   2 |                        2 |
+---------------------+---------------------+--------------------------+
1 row in set (0.01 sec)

**count(distinct i,j)  Should return 4 instead of 2**

PR #3561, opened 2020-05-12
[Bug] generate exceptions to avoid mulitDistinctAggregation produce wrong result #3492

wutiangan:
when a query (#3492) contain “2 DistinctAggregation with one column” and “1 DistinctAggregation with two columns”,  it will produce wrong result.

This pull request is not to solve this problem really, but to generate exceptions to avoid getting wrong results. 

This problem needs a real repair in future.

### DIFF
--- a/fe/src/test/java/org/apache/doris/analysis/AggregateTest.java
+++ b/fe/src/test/java/org/apache/doris/analysis/AggregateTest.java
@@ -1,0 +1,102 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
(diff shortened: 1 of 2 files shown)


fe/src/test/java/org/apache/doris/analysis/AggregateTest.java
```diff
@@ -0,0 +1,102 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
morningman, line 1: I think we can just move this test to `QueryPlanTest.java`

morningman (changes requested):
Please rebase the code to make UT happy

fe/src/test/java/org/apache/doris/analysis/AggregateTest.java
wutiangan: I think this is an exception in the analysis phase, not execution plan phase. 
This file is specially used for unit tests related to aggregate function analysis

morningman (approved):
LGTM

PR #3571, opened 2020-05-12
Support bitmap_intersect

### LINKED ISSUE #3552: Support Bitmap Intersect
# Support Bitmap Intersect

Support aggregate function Bitmap Intersect, it is mainly used to take intersection of grouped data.

# bitmap_intersect

Calculates the intersection of bitmap columns and returns a bitmap object.

```
bitmap_intersect(expr)
```

**Parameters**

The `expr` column type must be bitmap.

**Return value**

bitmap object

**Example**

table schema

```
create table bitmap_intersect_test (
    tag varchar(20),
    user_id bitmap bitmap_union
) 
AGGREGATE KEY(tag)
DISTRIBUTED BY HASH(tag) BUCKETS 3;
``` 

Query which users satisfy the three tags a, b, and c at the same time.

```
select bitmap_to_string(bitmap_intersect(user_id)) from 
(
    select bitmap_union(user_id) user_id from bitmap_intersect_test 
    where tag in ('a', 'b', 'c')
    group by tag
) a
```

# Design

## Semantic analysis

The child type of bitmap_intersect must be bitmap.

```
class FunctionCallExpr {

    void analyze() {
      if(fnName.equals("bitmap_intersect")) {
          ...
          if(!fn.getChild(0).isBitmapType()) {
              throw new AnalysisException("the child type of " + fnName + " must be bitmap")
          }
          ...
      }
    }
}
```

## Function implement

The function of each stage of `` `bitmap_intersect``` is declared in` `` function set```.

**Function definition**

```
FunctionName: bitmap_intersect,
InputType: bitmap,
OutputType: bitmap,
IntermediateType: varchar
```

**init**

Directly reuse the current bitmap init function

```              "_ZN5doris15BitmapFunctions11bitmap_initEPN9doris_udf15FunctionContextEPNS1_9StringValE"
```

**update**
**merge**

Perform intersection calculation on the bitmap grouped on the current node

```
void BitmapFunctions::bitmap_intersect(FunctionContext* ctx, const StringVal& src, StringVal* dst) {
    if (src.is_null) {
        return;
    }
    auto dst_bitmap = reinterpret_cast<BitmapValue*>(dst->ptr);
    // zero size means the src input is a agg object
    if (src.len == 0) {
        (*dst_bitmap) &= *reinterpret_cast<BitmapValue*>(src.ptr);
    } else {
        (*dst_bitmap) &= BitmapValue((char*) src.ptr);
    }
}
```

**serialize**
**finalize**

Directly reuse the current bitmap serialization function

```             "_ZN5doris15BitmapFunctions16bitmap_serializeEPN9doris_udf15FunctionContextERKNS1_9StringValE",
```

**Query plan**

```

mysql> explain select bitmap_intersect(user_id) from (select bitmap_union(user_id) user_id from  bitmap_intersect_test   where tag in ('a', 'b', 'c') group by tag ) a;
+----------------------------------------------------------------------------------------+
| Explain String                                                                         |
+----------------------------------------------------------------------------------------+
| PLAN FRAGMENT 0                                                                        |
|  OUTPUT EXPRS:<slot 8>                                                                 |
|   PARTITION: UNPARTITIONED                                                             |
|                                                                                        |
|   RESULT SINK                                                                          |
|                                                                                        |
|   6:AGGREGATE (merge finalize)                                                         |
|   |  output: bitmap_intersect(<slot 7>)                                                    |
|   |  group by:                                                                         |
|   |  tuple ids: 5                                                                      |
|   |                                                                                    |
|   5:EXCHANGE                                                                           |
|      tuple ids: 4                                                                      |
|                                                                                        |
| PLAN FRAGMENT 1                                                                        |
|  OUTPUT EXPRS:                                                                         |
|   PARTITION: HASH_PARTITIONED: <slot 2>                                                |
|                                                                                        |
|   STREAM DATA SINK                                                                     |
|     EXCHANGE ID: 05                                                                    |
|     UNPARTITIONED                                                                      |
|                                                                                        |
|   2:AGGREGATE (update serialize)                                                       |
|   |  output: bitmap_intersect(<slot 5>)                                                    |
|   |  group by:                                                                         |
|   |  tuple ids: 4                                                                      |
|   |                                                                                    |
|   4:AGGREGATE (merge finalize)                                                         |
|   |  output: bitmap_union(<slot 3>)                                                    |
|   |  group by: <slot 2>                                                                |
|   |  tuple ids: 2                                                                      |
|   |                                                                                    |
|   3:EXCHANGE                                                                           |
|      tuple ids: 1                                                                      |
|                                                                                        |
| PLAN FRAGMENT 2                                                                        |
|  OUTPUT EXPRS:                                                                         |
|   PARTITION: RANDOM                                                                    |
|                                                                                        |
|   STREAM DATA SINK                                                                     |
|     EXCHANGE ID: 03                                                                    |
|     HASH_PARTITIONED: <slot 2>                                                         |
|                                                                                        |
|   1:AGGREGATE (update serialize)                                                       |
|   |  STREAMING                                                                         |
|   |  output: bitmap_union(`user_id`)                                                   |
|   |  group by: `tag`                                                                   |
|   |  tuple ids: 1                                                                      |
|   |                                                                                    |
|   0:OlapScanNode                                                                       |
|      TABLE:  bitmap_intersect_test                                                  |
|      PREAGGREGATION: ON                                                                |
|      PREDICATES: `tag` IN ('a', 'b', 'c') |
|      partitions=1/1                                                                    |
|      rollup: bitmap_intersect_test                                                  |
|      tabletRatio=100/100                                                               |                                                          |
|      numNodes=6                                                                        |
|      tuple ids: 0                                                                      |
+----------------------------------------------------------------------------------------+
```

EmmyMiao87:
Support aggregate function Bitmap Intersect, it is mainly used to take intersection of grouped data.
The function 'bitmap_intersect(expr)' calculates the intersection of bitmap columns and returns a bitmap object.
The defination is following:
FunctionName: bitmap_intersect,
InputType: bitmap,
OutputType: bitmap

The scenario is as follows:
Query which users satisfy the three tags a, b, and c at the same time.

```
select bitmap_to_string(bitmap_intersect(user_id)) from
(
    select bitmap_union(user_id) user_id from bitmap_intersect_test
    where tag in ('a', 'b', 'c')
    group by tag
) a
```
Closed #3552.

Change-Id: I0d50c9bb6827f7319c4d854f621412343a3ece24

### DIFF
--- a/be/src/exprs/bitmap_function.cpp
+++ b/be/src/exprs/bitmap_function.cpp
@@ -302,8 +302,8 @@ void BitmapFunctions::bitmap_union(FunctionContext* ctx, const StringVal& src, S
     }
 }
 
-// this is the read init function for bitmap_intersect
-void BitmapFunctions::bitmap_intersect_init_real(FunctionContext* ctx, StringVal* dst) {
+// the dst value could be null
+void BitmapFunctions::nullable_bitmap_init(FunctionContext* ctx, StringVal* dst) {
     dst->is_null = true;
 }
 
@@ -369,14 +369,13 @@ StringVal BitmapFunctions::bitmap_hash(doris_udf::FunctionContext* ctx, const do
 
 StringVal BitmapFunctions::bitmap_serialize(FunctionContext* ctx, const StringVal& src) {
     if (src.is_null) {
-        BitmapValue src_bitmap;
-        return serialize(ctx, &src_bitmap);
-    } else {
-        auto src_bitmap = reinterpret_cast<BitmapValue*>(src.ptr);
-        StringVal result = serialize(ctx, src_bitmap);
-        delete src_bitmap;
-        return result;
+        return src;
     }
+
+    auto src_bitmap = reinterpret_cast<BitmapValue*>(src.ptr);
+    StringVal result = serialize(ctx, src_bitmap);
+    delete src_bitmap;
+    return result;
 }
 
 // This is a init function for intersect_count not for bitmap_intersect.
--- a/be/src/exprs/bitmap_function.cpp
+++ b/be/src/exprs/bitmap_function.cpp
@@ -302,7 +302,8 @@ void BitmapFunctions::bitmap_union(FunctionContext* ctx, const StringVal& src, S
     }
 }
 
-void BitmapFunctions::bitmap_intersect_init(FunctionContext* ctx, StringVal* dst) {
+// this is the read init function for bitmap_intersect
+void BitmapFunctions::bitmap_intersect_init_real(FunctionContext* ctx, StringVal* dst) {
     dst->is_null = true;
 }
 
@@ -378,8 +379,9 @@ StringVal BitmapFunctions::bitmap_serialize(FunctionContext* ctx, const StringVa
     }
 }
 
+// This is a init function for intersect_count not for bitmap_intersect.
 template<typename T, typename ValType>
-void BitmapFunctions::intersect_count_init(FunctionContext* ctx, StringVal* dst) {
+void BitmapFunctions::bitmap_intersect_init(FunctionContext* ctx, StringVal* dst) {
     dst->is_null = false;
     dst->len = sizeof(BitmapIntersect<T>);
     auto intersect = new BitmapIntersect<T>();
@@ -539,25 +541,26 @@ template void BitmapFunctions::bitmap_update_int<IntVal>(
 template void BitmapFunctions::bitmap_update_int<BigIntVal>(
         FunctionContext* ctx, const BigIntVal& src, StringVal* dst);
 
-template void BitmapFunctions::intersect_count_init<int8_t, TinyIntVal>(
+// this is init function for intersect_count not for bitmap_intersect
+template void BitmapFunctions::bitmap_intersect_init<int8_t, TinyIntVal>(
     FunctionContext* ctx, StringVal* dst);
-template void BitmapFunctions::intersect_count_init<int16_t, SmallIntVal>(
+template void BitmapFunctions::bitmap_intersect_init<int16_t, SmallIntVal>(
     FunctionContext* ctx, StringVal* dst);
-template void BitmapFunctions::intersect_count_init<int32_t, IntVal>(
+template void BitmapFunctions::bitmap_intersect_init<int32_t, IntVal>(
     FunctionContext* ctx, StringVal* dst);
-template void BitmapFunctions::intersect_count_init<int64_t, BigIntVal>(
+template void BitmapFunctions::bitmap_intersect_init<int64_t, BigIntVal>(
     FunctionContext* ctx, StringVal* dst);
-template void BitmapFunctions::intersect_count_init<__int128, LargeIntVal>(
+template void BitmapFunctions::bitmap_intersect_init<__int128, LargeIntVal>(
     FunctionContext* ctx, StringVal* dst);
-template void BitmapFunctions::intersect_count_init<float, FloatVal>(
+template void BitmapFunctions::bitmap_intersect_init<float, FloatVal>(
     FunctionContext* ctx, StringVal* dst);
-template void BitmapFunctions::intersect_count_init<double, DoubleVal>(
+template void BitmapFunctions::bitmap_intersect_init<double, DoubleVal>(
     FunctionContext* ctx, StringVal* dst);
-template void BitmapFunctions::intersect_count_init<DateTimeValue, DateTimeVal>(
+template void BitmapFunctions::bitmap_intersect_init<DateTimeValue, DateTimeVal>(
     FunctionContext* ctx, StringVal* dst);
-template void BitmapFunctions::intersect_count_init<DecimalV2Value, DecimalV2Val>(
+template void BitmapFunctions::bitmap_intersect_init<DecimalV2Value, DecimalV2Val>(
     FunctionContext* ctx, StringVal* dst);
-template void BitmapFunctions::intersect_count_init<StringValue, StringVal>(
+template void BitmapFunctions::bitmap_intersect_init<StringValue, StringVal>(
     FunctionContext* ctx, StringVal* dst);
 
 
--- a/be/src/exprs/bitmap_function.cpp
+++ b/be/src/exprs/bitmap_function.cpp
@@ -302,6 +302,30 @@ void BitmapFunctions::bitmap_union(FunctionContext* ctx, const StringVal& src, S
     }
 }
 
+void BitmapFunctions::bitmap_intersect_init(FunctionContext* ctx, StringVal* dst) {
+    dst->is_null = true;
+}
+
+void BitmapFunctions::bitmap_intersect(FunctionContext* ctx, const StringVal& src, StringVal* dst) {
+    if (src.is_null) {
+        return;
+    }
+    // if dst is null, the src input is the first value
+    if (dst->is_null) {
+        dst->is_null = false;
+        dst->len = sizeof(BitmapValue);
+        dst->ptr = (uint8_t*)new BitmapValue((char*) src.ptr);
+        return;
+    }
+    auto dst_bitmap = reinterpret_cast<BitmapValue*>(dst->ptr);
+    // zero size means the src input is a agg object
+    if (src.len == 0) {
+        (*dst_bitmap) &= *reinterpret_cast<BitmapValue*>(src.ptr);
+    } else {
+        (*dst_bitmap) &= BitmapValue((char*) src.ptr);
+    }
+}
+
 BigIntVal BitmapFunctions::bitmap_count(FunctionContext* ctx, const StringVal& src) {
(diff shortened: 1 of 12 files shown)


EmmyMiao87:
I will add sql-reference later.

be/src/exprs/bitmap_function.cpp
```diff
@@ -343,14 +367,19 @@ StringVal BitmapFunctions::bitmap_hash(doris_udf::FunctionContext* ctx, const do
 }
 
 StringVal BitmapFunctions::bitmap_serialize(FunctionContext* ctx, const StringVal& src) {
-    auto src_bitmap = reinterpret_cast<BitmapValue*>(src.ptr);
-    StringVal result = serialize(ctx, src_bitmap);
-    delete src_bitmap;
-    return result;
+    if (src.is_null) {
+        BitmapValue src_bitmap;
+        return serialize(ctx, &src_bitmap);
```
kangpinghuang, line 372: I think is_null is diffrent from empty bitmap.
why here return an empty bitmap?

be/src/exprs/bitmap_function.cpp
```diff
@@ -343,14 +367,19 @@ StringVal BitmapFunctions::bitmap_hash(doris_udf::FunctionContext* ctx, const do
...
+    } else {
+        auto src_bitmap = reinterpret_cast<BitmapValue*>(src.ptr);
+        StringVal result = serialize(ctx, src_bitmap);
+        delete src_bitmap;
```
kangpinghuang, line 376: why delete the src_bitmap here? who will manage the bitmap lifetime?

be/src/exprs/bitmap_function.cpp
```diff
@@ -302,6 +302,30 @@ void BitmapFunctions::bitmap_union(FunctionContext* ctx, const StringVal& src, S
...
+    // if dst is null, the src input is the first value
+    if (dst->is_null) {
+        dst->is_null = false;
+        dst->len = sizeof(BitmapValue);
+        dst->ptr = (uint8_t*)new BitmapValue((char*) src.ptr);
+        return;
+    }
+    auto dst_bitmap = reinterpret_cast<BitmapValue*>(dst->ptr);
+    // zero size means the src input is a agg object
+    if (src.len == 0) {
```
kangpinghuang, line 322: It seems that when src.len == 0, the src.ptr point to the bitmap value pointer, but when src.len != 0, the ptr points to the serialized bitmap contents, why?

kangpinghuang (changes requested):

be/src/exprs/bitmap_function.cpp
morningman: I think `null` means "no data". And for "no data", return an empty set is reasonable.

be/src/util/bitmap_value.h
```diff
@@ -1235,7 +1235,7 @@ class BitmapValue {
     enum BitmapDataType {
         EMPTY = 0,
         SINGLE = 1, // single element
-        BITMAP = 2 // more than one elements
+        BITMAP = 2// more than one elements
```
morningman, line 1238:
```suggestion
        BITMAP = 2 // more than one elements
```

be/src/exprs/bitmap_function.cpp
wutiangan: I also think there is no need to serialize null bitmap.
Set StringVal.is_null = true is better.
for in later merge function, we at first check if the StringVal is null.

EmmyMiao87: This serialize function is  multiplexed by finalize function in other bitmap function such as bitmap_union.

be/src/exprs/bitmap_function.cpp
EmmyMiao87: The src_bitmap was initialized in merge function. So it need to delete src_bitmap when UDA has been finished.

fe/src/main/java/org/apache/doris/catalog/FunctionSet.java
```diff
@@ -554,29 +555,29 @@ public boolean isNonNullResultWithNullParamFunctions(String funcName) {
     private static final Map<Type, String> BITMAP_INTERSECT_INIT_SYMBOL =
             ImmutableMap.<Type, String>builder()
                     .put(Type.TINYINT,
-                            "_ZN5doris15BitmapFunctions21bitmap_intersect_initIaN9doris_udf10TinyIntValEEEvPNS2_15FunctionContextEPNS2_9StringValE")
+                            "_ZN5doris15BitmapFunctions20intersect_count_initIaN9doris_udf10TinyIntValEEEvPNS2_15FunctionContextEPNS2_9StringValE")
```
imay, line 558: You can't change this name directly.
Because it will cause query fail when grayscale upgrade

imay (changes requested):

fe/src/main/java/org/apache/doris/catalog/FunctionSet.java
EmmyMiao87: The grayscale upgrade itself will also cause the query to fail，so I think the impact is not significant.

be/src/exprs/bitmap_function.cpp
```diff
@@ -302,6 +302,31 @@ void BitmapFunctions::bitmap_union(FunctionContext* ctx, const StringVal& src, S
...
+void BitmapFunctions::bitmap_intersect_init_real(FunctionContext* ctx, StringVal* dst) {
+    dst->is_null = true;
+}
+
+void BitmapFunctions::bitmap_intersect(FunctionContext* ctx, const StringVal& src, StringVal* dst) {
+    if (src.is_null) {
+        return;
+    }
+    // if dst is null, the src input is the first value
+    if (dst->is_null) {
```
kangkaisen, line 315: Would better add a `UNLIKELY` macros.

be/src/exprs/bitmap_function.cpp
```diff
@@ -302,6 +302,31 @@ void BitmapFunctions::bitmap_union(FunctionContext* ctx, const StringVal& src, S
     }
 }
 
+// this is the read init function for bitmap_intersect
+void BitmapFunctions::bitmap_intersect_init_real(FunctionContext* ctx, StringVal* dst) {
+    dst->is_null = true;
```
imay, line 307: why initial result bitmap as null? it seems that it will return empty bitmap when result bitmap is null

kangkaisen:
@EmmyMiao87  Hi.
当前对于 bitmap_intersect null的处理是不合理的。 

按照SQL标准，sum(null), min(null), max(null)结果为 null. count(null)为0.

所以 对于 bitmap_intersect，bitmap_union 返回bitmap的函数，输入全是null时，应该返回null,
对于 bitmap_intersect_count, bitmap_union_count 等返回count值的函数， 输入全是null时，应该返回0.

在之前最开始的时候，我认为所有hll,bitmap的函数都不需要返回null，对于输入是null，我们返回empty bitmap 或者empty hll 即可。（it's my fault...）

所以目前bitmap_init, bitmap_serialize, bitmap_finalize的实现都是not nullable的。 
所以你这个PR，你应该实现新的函数 nullable_bitmap_int, nullable_bitmap_serialize。nullable_bitmap_serialize 中src是null, 应该返回null，而不是空的empty.

be/src/exprs/bitmap_function.cpp
```diff
@@ -302,6 +302,31 @@ void BitmapFunctions::bitmap_union(FunctionContext* ctx, const StringVal& src, S
     }
 }
 
+// this is the read init function for bitmap_intersect
+void BitmapFunctions::bitmap_intersect_init_real(FunctionContext* ctx, StringVal* dst) {
```
kangkaisen, line 306: 这个init函数是通用的，将来不止bitmap_intersect 会用到。
当前 bitmap_init 的实现是not nullable的，所以你这个函数合理的命名应该是nullable_bitmap_init， 或者bitmap_nullable_init.

be/src/exprs/bitmap_function.cpp
EmmyMiao87: The initial result bitmap must be null. Otherwise, the intersection between dst and src will be empty.

docs/en/sql-reference/sql-functions/bitmap-functions/bitmap_intersect.md
```diff
@@ -0,0 +1,54 @@
+---
+{
+    "title": "bitmap_intersect",
+    "language": "en"
```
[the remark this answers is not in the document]
EmmyMiao87: Added

imay (approved):
LGTM

kangkaisen (approved):
+1, LGTM

kangpinghuang:
LGTM, +1

kangpinghuang (approved):
LGTM

Issue #3568, opened 2020-05-12
Do not link to dist.apache.org from public pages


sebbASF:
The dist.apache.org host is not intended for public downloads.
It is only for use by developers when staging releases.

As such, the page
http://doris.apache.org/master/en/community/verify-apache-release.html
must not point to dist.apache.org.

Downloaders must be directed to the ASF mirror system

imay:
Hi @sebbASF 
This issue has been fixed, please check

sebbASF:
The reference to dist.a.o has been removed, thanks.

However, I have just noticed that the verification page suggests using RAT to verify the source licenses. The RAT tool is intended for use by developers before a release is published. It is not appropriate to suggest that downloaders should repeat that check. Please remove that section.

Issue #3567, opened 2020-05-12
Download page issues


sebbASF:
The page http://doris.apache.org/master/en/downloads/downloads.html has some issues:

There must be a link to the KEYS file at https://downloads.apache.org/incubator/doris/KEYS

The phrase:
"compile and install Doirs."
should be
"compile and install Doris."

sebbASF:
Also the source download links must use the ASF mirror system

imay:
Thanks @sebbASF for your advice, we will fix it.

imay:
Hi @sebbASF 
This issue has been fixed, please check

sebbASF:
Looks OK now, thanks

PR #3574, opened 2020-05-13
[Fix]modify download links & remove some links' suffix `_EN`(branch-0.12)

hffariel:
modify download links & remove document links' suffix `_EN`(#3567 #3568)


imay (approved):
LGTM

Issue #3555, opened 2020-05-11
SelectStmt.tosql() is the same as the SQL returned by different SQL


marising:
**Describe the bug**
```
select distinct ss_sold_date_sk as sk from store_sales order by sk limit 100;
select distinct ss_item_sk as sk from store_sales order by sk limit 100;
```
Invoke FE SelectStmt.toSql()，return the same result：
```
SELECT DISTINCT <slot 2> AS `sk` FROM `default_cluster:tpcds`.`store_sales` ORDER BY `sk` ASC LIMIT 100
```

**Expected behavior**
```
SELECT DISTINCT ss_sold_date_sk  AS `sk` FROM `default_cluster:tpcds`.`store_sales` ORDER BY `sk` ASC LIMIT 100

SELECT DISTINCT ss_item_sk  AS `sk` FROM `default_cluster:tpcds`.`store_sales` ORDER BY `sk` ASC LIMIT 100
```
I try to fix the bug.

PR #3577, opened 2020-05-13
[Fix]SlotRef.tosql() is the same as the SQL returned by different sql

marising:
- Describe the bug

select distinct ss_sold_date_sk as sk from store_sales order by sk limit 100;
select distinct ss_item_sk as sk from store_sales order by sk limit 100;
Invoke FE SelectStmt.toSql()，return the same result：
SELECT DISTINCT <slot 2> AS `sk` FROM `default_cluster:tpcds`.`store_sales` ORDER BY `sk` ASC LIMIT 100

Because label in SlotRef corresponding to distinct function is null, SQL is output in the form of slot + number. When label is null, desc.getsourceexprs() is obtained to get the column name

- Modified effect

SQL1：SELECT DISTINCT <slot 2>  `ss_sold_date_sk` AS `sk` FROM `default_cluster:tpcds`.`store_sales` ORDER BY `sk` ASC LIMIT 100

SQL2：SELECT DISTINCT <slot 2>  `ss_item_sk` AS `sk` FROM `default_cluster:tpcds`.`store_sales` ORDER BY `sk` ASC LIMIT 100

#3555

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/SlotRef.java
+++ b/fe/src/main/java/org/apache/doris/analysis/SlotRef.java
@@ -163,9 +163,12 @@ public class SlotRef extends Expr {
         } else if (label != null) {
             return label + sb.toString();
         } else if (desc.getSourceExprs() != null) {
+            if (desc.getId().asInt() != 1) {
+                sb.append("<slot " + Integer.toString(desc.getId().asInt()) + ">");
+            }
             for (Expr expr : desc.getSourceExprs()) {
-                sb.append(expr.toSql());
                 sb.append(" ");
+                sb.append(expr.toSql());
             }
             return sb.toString();
         } else {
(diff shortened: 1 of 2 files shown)


fe/src/main/java/org/apache/doris/analysis/SlotRef.java
```diff
@@ -162,6 +162,12 @@ public String toSqlImpl() {
             return tblName.toSql() + "." + label + sb.toString();
         } else if (label != null) {
             return label + sb.toString();
+        } else if (desc.getSourceExprs() != null) {
+            for (Expr expr : desc.getSourceExprs()) {
+                sb.append(expr.toSql());
+                sb.append(" ");
+            }
+            return sb.toString();
```
morningman, line 170: How about return both `slot id` and `expr`. I'm afraid that `slot id` may be used for some tracing debug.

And please fix all related unit tests after you change here.

morningman (changes requested):

fe/src/main/java/org/apache/doris/analysis/SlotRef.java
marising: How about slot ID and expr? I need MD5 of SQL as the key of cache instead of slot ID. I'm trying to fix the other unit tests

morningman (approved):
LGTM

Issue #3576, opened 2020-05-13
[Doris on ES] value in like not push down correcttly


blackfox1983:
**Describe the bug**
there is a table with rows as follows:
```
select * from table;
+------+------+------+------------+------------+
| k1   | k2   | k3   | UpdateTime | ArriveTime |
+------+------+------+------------+------------+
| NULL | NULL | kkk3 |  123456789 |       NULL |
| kkk1 | NULL | NULL |  123456789 |       NULL |
| NULL | kkk2 | NULL |  123456789 |       NULL |
+------+------+------+------------+------------+
```

use sql 
```
> select * from case_replay_for_milimin where k1 like '%kk1';
Empty set (0.01 sec)
```

It should return as follows:
```
+------+------+------+------------+------------+
| k1   | k2   | k3   | UpdateTime | ArriveTime |
+------+------+------+------------+------------+
| kkk1 | NULL | NULL |  123456789 |       NULL |
+------+------+------+------------+------------+
```

**Version: master**

blackfox1983:
pr : https://github.com/apache/incubator-doris/pull/3579

PR #3579, opened 2020-05-13
[Doris On ES]fix bug of value in like not translate correcttly

BabySid:
case is desc in #3576 

**Why this case happened**
In current implement, translation into dsl only if it is not the first charactor.
Thus, when sql is write like '%abc', translation would not run.

**How fixed**
- Now, translation will trigger with charactor '?' or '*'
- if it is the first charactor, translate directly
- else,  check the preceding char is escaped or not to determin translation or not

### DIFF
--- a/be/src/exec/es/es_query_builder.cpp
+++ b/be/src/exec/es/es_query_builder.cpp
@@ -120,10 +120,10 @@ WildCardQueryBuilder::WildCardQueryBuilder(const ExtLikePredicate& like_predicat
     //      abc\\%123 ===> abc\\%123
     // NOTE. user must input sql like 'abc\\_123' or 'abc\\%ykz'
     for (int i = 0; i< _like_value.size(); i++) {
-        if (_like_value[i] == '_' || _like_value[i]== '%') {
+        if (_like_value[i] == '_' || _like_value[i] == '%') {
                 if (i == 0) {
                     _like_value[i] = (_like_value[i] == '_') ? '?' : '*';
-                } else if (_like_value[i-1] != '\\' ) {
+                } else if (_like_value[i - 1] != '\\' ) {
                     _like_value[i] = (_like_value[i] == '_') ? '?' : '*';
                 }
         }
--- a/be/src/exec/es/es_query_builder.cpp
+++ b/be/src/exec/es/es_query_builder.cpp
@@ -123,7 +123,7 @@ WildCardQueryBuilder::WildCardQueryBuilder(const ExtLikePredicate& like_predicat
         if (_like_value[i] == '_' || _like_value[i]== '%') {
                 if (i == 0) {
                     _like_value[i] = (_like_value[i] == '_') ? '?' : '*';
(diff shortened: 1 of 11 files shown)


be/src/exec/es/es_query_builder.cpp
```diff
@@ -113,12 +113,17 @@ WildCardQueryBuilder::WildCardQueryBuilder(const ExtLikePredicate& like_predicat
...
     //      abc\\_123 ===> abc\\_123
     //      abc\\%123 ===> abc\\%123
     // NOTE. user must input sql like 'abc\\_123' or 'abc\\%ykz'
     for (int i = 0; i< _like_value.size(); i++) {
-        if ((_like_value[i] == '_' || _like_value[i]== '%') && i > 0) {
-                if (_like_value[i-1] != '\\' ) {
+        if (_like_value[i] == '_' || _like_value[i]== '%') {
+                if (i == 0) {
+                    _like_value[i] = (_like_value[i] == '_') ? '?' : '*';
+                }else if (_like_value[i-1] != '\\' ) {
```
wuyunfeng, line 126:
```suggestion
                } else if (_like_value[i-1] != '\\' ) {
```

wuyunfeng (dismissed):
LGTM

blackfox1983:
@imay pls have a look?

be/src/exec/es/es_query_builder.cpp
```diff
@@ -113,12 +113,17 @@ WildCardQueryBuilder::WildCardQueryBuilder(const ExtLikePredicate& like_predicat
...
+    //      %abc123  ===> *abc123
+    //      _abc123  ===> ?abc123
+    //      \\_abc1  ===> \\_abc1
     //      abc\\_123 ===> abc\\_123
     //      abc\\%123 ===> abc\\%123
     // NOTE. user must input sql like 'abc\\_123' or 'abc\\%ykz'
     for (int i = 0; i< _like_value.size(); i++) {
-        if ((_like_value[i] == '_' || _like_value[i]== '%') && i > 0) {
-                if (_like_value[i-1] != '\\' ) {
+        if (_like_value[i] == '_' || _like_value[i]== '%') {
```
imay, line 123:
```suggestion
        if (_like_value[i] == '_' || _like_value[i] == '%') {
```

be/src/exec/es/es_query_builder.cpp
```diff
@@ -113,12 +113,17 @@ WildCardQueryBuilder::WildCardQueryBuilder(const ExtLikePredicate& like_predicat
...
+                if (i == 0) {
+                    _like_value[i] = (_like_value[i] == '_') ? '?' : '*';
+                } else if (_like_value[i-1] != '\\' ) {
```
imay, line 126:
```suggestion
                } else if (_like_value[i - 1] != '\\' ) {
```

imay (approved):
LGTM

blackfox1983:
reopened https://github.com/apache/incubator-doris/pull/3602

PR #3587, opened 2020-05-14
[Bug] Fix bug that ConcurrentModificationException thrown 

### LINKED ISSUE #3588: [Bug] ConcurrentModificationException thrown when truncate table
**Describe the bug**
ConcurrentModificationException thrown when truncate table 

**To Reproduce**
1. create table with partition
2. add some temp partitiions
3. truncate table

morningman:
When truncate the table, a ConcurrentModificationException may thrown when there
are temp partitions in this table.

Fix: #3588

### DIFF
--- a/fe/src/main/java/org/apache/doris/catalog/Catalog.java
+++ b/fe/src/main/java/org/apache/doris/catalog/Catalog.java
@@ -4514,7 +4514,7 @@ public class Catalog {
                     for (Partition partition : olapTable.getAllPartitions()) {
                         long partitionId = partition.getId();
                         DataProperty dataProperty = partitionInfo.getDataProperty(partition.getId());
-                        Preconditions.checkNotNull(dataProperty);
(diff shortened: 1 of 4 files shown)


fe/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -4514,7 +4514,7 @@ public long getNextId() {
                     for (Partition partition : olapTable.getAllPartitions()) {
                         long partitionId = partition.getId();
                         DataProperty dataProperty = partitionInfo.getDataProperty(partition.getId());
-                        Preconditions.checkNotNull(dataProperty);
+                        Preconditions.checkNotNull(dataProperty, partition.getName() + ", pId:" + partitionId + ", db: " + dbId + ", tbl: " + tableId);
```
wutiangan, line 4517: The second parameter is better to be a complete sentence.

morningman: This is not a log, just for printing some debug info. So I think it is not necessary to be a sentence.

EmmyMiao87 (approved):

Issue #3600, opened 2020-05-15
large bitmap serialization failed because of overflow of size


kangpinghuang:
**Describe the bug**
when use bitmap to do insertsect, if the bitmap is large than the MAX_INT, the allocation of memory from free pool will failed, which will lead to serialization failed because of core dump.

the following is the bitmap size I come to is 3G+.

kangpinghuang:
One of the reason is that 3G+ is large than MAX_INT, but FreePool->allocate accept int as argment, overflow will lead to allocation and serialation bug.

imay:
Hi, @kangpinghuang 
Can you explain the scenario that cause this problem?
I think it will cause many problems if the body is so large. And if we are going to support it, we should review all the related places.

PR #3601, opened 2020-05-15
fix large string val allocation failure, #3600

kangpinghuang:
Large bitmap will need use StringVal to allocate large memory, which is large than MAX_INT.
The overflow will cause serialization failure of bitmap.

### DIFF
--- a/be/src/exprs/bitmap_function.cpp
+++ b/be/src/exprs/bitmap_function.cpp
@@ -458,7 +458,8 @@ StringVal BitmapFunctions::bitmap_from_string(FunctionContext* ctx, const String
     }
 
     std::vector<uint64_t> bits;
-    if (!SplitStringAndParse({(const char*)input.ptr, input.len}, ",", &safe_strtou64, &bits)) {
+    // TODO: I think StringPiece's len should also be uint64_t
+    if (!SplitStringAndParse({(const char*)input.ptr, (int)input.len}, ",", &safe_strtou64, &bits)) {
         return StringVal::null();
     }
 
--- a/be/src/exprs/anyval_util.h
+++ b/be/src/exprs/anyval_util.h
@@ -273,7 +273,7 @@ public:
         if (type.type == FunctionContext::TYPE_VARCHAR
               || type.type == FunctionContext::TYPE_CHAR) {
             DCHECK(type.len >= 0); 
-            val->len = std::min(val->len, type.len);
+            val->len = std::min(val->len, (uint64_t)type.len);
         }   
     } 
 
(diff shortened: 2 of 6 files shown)


EmmyMiao87:
Please add a unit test about large string val

be/src/exprs/bitmap_function.cpp
```diff
@@ -458,7 +458,8 @@ StringVal BitmapFunctions::bitmap_from_string(FunctionContext* ctx, const String
     }
 
     std::vector<uint64_t> bits;
-    if (!SplitStringAndParse({(const char*)input.ptr, input.len}, ",", &safe_strtou64, &bits)) {
+    // TODO: I think StringPiece's len should also be uint64_t
+    if (!SplitStringAndParse({(const char*)input.ptr, (int)input.len}, ",", &safe_strtou64, &bits)) {
```
morningman, line 462: Why not change it to int64 too?

kangkaisen:
> Large bitmap will need use StringVal to allocate large memory, which is large than MAX_INT.
> The overflow will cause serialization failure of bitmap.

@kangpinghuang Hi. you mean the bitmap size is larger than 2G? are you sure which is reasonable？

kangpinghuang:
> than

yes, There is a user who use the bitmap to do the count distinct. In his situation, the dimension is about 30billion, the bitmap I see is about 3G.

kangpinghuang:
> > Large bitmap will need use StringVal to allocate large memory, which is large than MAX_INT.
> > The overflow will cause serialization failure of bitmap.
> 
> @kangpinghuang Hi. you mean the bitmap size is larger than 2G? are you sure which is reasonable？

yes, There is a user who use the bitmap to do the count distinct. In his situation, the dimension is about 30billion, the bitmap I see is about 3G.

be/src/exprs/bitmap_function.cpp
kangpinghuang: because it relate to another struct StringPiece, which is defined in gutil, and to much code reference it. I think whether to change StringPiece need to be discussed and if true, can be done in another pr.

kangkaisen:
> > > Large bitmap will need use StringVal to allocate large memory, which is large than MAX_INT.
> > > The overflow will cause serialization failure of bitmap.
> > 
> > 
> > @kangpinghuang Hi. you mean the bitmap size is larger than 2G? are you sure which is reasonable？
> 
> yes, There is a user who use the bitmap to do the count distinct. In his situation, the dimension is about 30billion, the bitmap I see is about 3G.

I See, Thanks. Have you tested this PR in prod env a long time?

I think simply change the StringVal size from int to uint64_t maybe have a lot of issues:

Such as, How do we handle network transfer？Mempool could handle 3G huge memory allocate well?  the StringVal and Slice exchange could work well?

EmmyMiao87:
This pr has some error. The correct pr is here #3724

PR #3602, opened 2020-05-15
[Doris On ES]fix bug of like not translate correcttly

BabySid:
case is desc in #3576

Why this case happened
In current implement, translation into dsl only if it is not the first charactor.
Thus, when sql is write like '%abc', translation would not run.

How fixed

Now, translation will trigger with charactor '?' or '*'
if it is the first charactor, translate directly
else, check the preceding char is escaped or not to determin translation or not


imay (approved):
LGTM

Issue #3464, opened 2020-05-05
[Bug]Query failed when fact table has no data in join case


wangbo:
**Describe the bug**
In the join case ```fact_table_a left join dim_table_b```, when ```dim_table_b```'s size exceeds 2GB and the ```fact_table_a``` has no data,the query will failed and prints error stack as below:
```
Memory exceed limit. Hash join, while constructing the hash table. Backend: 123.456.789, fragment: 4fbdd81809244f3d-909e26102118fe92 Used: 2147529904, Limit: 2147483648. You can change the limit by session variable exec_mem_limit.
```

**Why the bug happened**
1. broadcast cost calculation error.
    Current broadcast cost calculation logic is 
```broadcastCost = rhsDataSize * leftChildFragment.getNumNodes();```
    But when ```the leftChildFragment``` has no data which means 0 * rhsDataSize.
    So broadcastCost is 0, then broadcast join happens.
    when the right table(dim_table_b) has big size, the query must fail.
    The correct result here should be ```broadcastCost = rhsDataSize *  1```
2.  ```choose host for exchange fragment``` has better implementation
     In current implementation,```exchange fragment``` will be assigned to the most left child's node and keeps the same parallelism.
     But when the most left child has no data, there will be only one backend for it.
     So send all right table's data to one exchange node to build hash table will cause memory exceeds when the right tables' size is big.

**Solution**
1. When the ```leftChildFragment``` 's has no data and returned num nodes 0, the returned num nodes should be changed to 1. Because for a scan node has no data,there will be as least a random selected node for it.
2. Find backends for ```exchange fragment```  by ```cost``` instead of the ```left most child fragment```
    ```exchange fragment``` should be kept in the same node with child fragment which has more data to reduce data transfer across network or has more instance num to improve the parallelism.
    In the current case, ```dim_table_b``` has more data and more instance num, so the``` exchange fragment``` should be kept in the same node with ```dim_table_b```'s fragment.
    
**Temporary solution**
Change SQL from ```fact_table_a left join dim_table_b``` to ```dim_table_b right join fact_table_a```

I will commit a pr for this later

kangkaisen:
>When the leftChildFragment 's has no data and returned num nodes 0, the returned num nodes should be changed to 1. Because for a scan node has no data,there will be as least a random selected node for it.

The broadcast cost calculation is right. You should fix the `getNumInstances` method in `OlapScanNode`. the getNumInstances result at least to 1.

PR #3604, opened 2020-05-16
(#3464) fix Query failed when fact table has no data in join case

wangbo:
#3464

major work
1.  Correct the value of ```numNodes``` and ```cardinality``` when ```OlapTableScan``` computeStats so that the ``` broadcast cost``` and ```paritition join cost ``` can be calculated correctly.
2. Find a input fragment with higher parallelism for shuffle fragment to assign backend

### DIFF
--- a/fe/src/main/java/org/apache/doris/qe/Coordinator.java
+++ b/fe/src/main/java/org/apache/doris/qe/Coordinator.java
@@ -914,17 +914,29 @@ public class Coordinator {
                 // there is no leftmost scan; we assign the same hosts as those of our
                 // leftmost input fragment (so that a partitioned aggregation
                 // fragment runs on the hosts that provide the input data)
-                PlanFragmentId inputFragmentIdx = fragments.get(i).getChild(0).getFragmentId();
+
+                // find a input fragment which has a higher parallelism
+                int inputFragmentIndex = 0;
+                int maxParallelism = 0;
+                for (int j = 0; j < fragment.getChildren().size(); j++) {
+                    int currentChildFragmentParallelism = fragmentExecParamsMap.get(fragment.getChild(j).getFragmentId()).instanceExecParams.size();
+                    if (currentChildFragmentParallelism > maxParallelism) {
+                        maxParallelism = currentChildFragmentParallelism;
+                        inputFragmentIndex = j;
+                    }
+                }
+
+                PlanFragmentId inputFragmentId = fragment.getChild(inputFragmentIndex).getFragmentId();
                 // AddAll() soft copy()
                 int exchangeInstances = -1;
                 if (ConnectContext.get() != null && ConnectContext.get().getSessionVariable() != null) {
                     exchangeInstances = ConnectContext.get().getSessionVariable().getExchangeInstanceParallel();
                 }
-                if (exchangeInstances > 0 && fragmentExecParamsMap.get(inputFragmentIdx).instanceExecParams.size() > exchangeInstances) {
+                if (exchangeInstances > 0 && fragmentExecParamsMap.get(inputFragmentId).instanceExecParams.size() > exchangeInstances) {
                     // random select some instance
                     // get distinct host,  when parallel_fragment_exec_instance_num > 1, single host may execute severval instances
                     Set<TNetworkAddress> hostSet = Sets.newHashSet();
-                    for (FInstanceExecParam execParams: fragmentExecParamsMap.get(inputFragmentIdx).instanceExecParams) {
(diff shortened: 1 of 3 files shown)


morningman (dismissed):
LGTM

fe/src/main/java/org/apache/doris/qe/Coordinator.java
```diff
@@ -914,17 +914,29 @@ private void computeFragmentHosts() throws Exception {
                 // there is no leftmost scan; we assign the same hosts as those of our
                 // leftmost input fragment (so that a partitioned aggregation
                 // fragment runs on the hosts that provide the input data)
```
kangkaisen, line 916: Please also update the comment.

fe/src/main/java/org/apache/doris/qe/Coordinator.java
```diff
@@ -914,17 +914,29 @@ private void computeFragmentHosts() throws Exception {
...
-                PlanFragmentId inputFragmentIdx = fragments.get(i).getChild(0).getFragmentId();
+
+                // find a input fragment which has a higher parallelism
+                int inputFragmentIndex = 0;
+                int maxParallelism = 0;
```
kangkaisen, line 920: I think `parallelism` rename  to  instance_number is better.

wangbo: 👌

fe/src/main/java/org/apache/doris/qe/Coordinator.java
wangbo: 👌

wangbo:
After I run this pr in out production env,I find a new problem.

A dim table whose size is 1.1 GB with single replica, and 1.1G is less than perNodeMemLimit(2GB)，so when the fact table is empty the broadcast join is choosen.
But when the hash table actually builds, the memory usesage still execeeds which means HASH_TBL_SPACE_OVERHEAD(=1.1) is not so precise.

This pr can fix  the is problem when user use ```shuffle hint``` because shuffle Node runs more nodes.

So I think there may two ways to solve the problem thoroughly;
1  make HASH_TBL_SPACE_OVERHEAD a configurable  param,but it seems not a common solution, because I'm not sure what factors can affect the real hash table size.
2 maybe we can disable broadcast when dim table exceeds 100M

morningman:
> 1 make HASH_TBL_SPACE_OVERHEAD a configurable param,but it seems not a common solution, because I'm not sure what factors can affect the real hash table size.
> 2 maybe we can disable broadcast when dim table exceeds 100M

These 2 solutions looks seem, that both require manual parameter adjustment. This is no different from directly adding hints manually.

wangbo:
@morningman 
The key point is that it's currently impossible to accurately estimate the size of the memory footprint at runtime according to table size.Otherwise, we can choose the right join mode or the correct num backends.
At present, it seems that it can only be adjusted manually. User decide the join mode.

morningman:
> @morningman
> The key point is that it's currently impossible to accurately estimate the size of the memory footprint at runtime according to table size.Otherwise, we can choose the right join mode or the correct num backends.
> At present, it seems that it can only be adjusted manually. User decide the join mode.

So I think just use the `shuffle hint` is enough? No need to add new config `HASH_TBL_SPACE_OVERHEAD` or disallow broadcast large than 100MB?

wangbo:
> > @morningman
> > The key point is that it's currently impossible to accurately estimate the size of the memory footprint at runtime according to table size.Otherwise, we can choose the right join mode or the correct num backends.
> > At present, it seems that it can only be adjusted manually. User decide the join mode.
> 
> So I think just use the `shuffle hint` is enough? No need to add new config `HASH_TBL_SPACE_OVERHEAD` or disallow broadcast large than 100MB?

I think we can set ```partition join``` as default  when we're unable to estimate the cost accurately and broadcast join cost equals partition join cost and without user hint.
From
```
                && broadcastCost <= partitionCost))) {
            doBroadcast = true;
        } else {
            doBroadcast = false;
        }
```
to
```
                && broadcastCost < partitionCost))) {
            doBroadcast = true;
        } else {
            doBroadcast = false;
        }
```

morningman:
> I think we can set `partition join` as default when we're unable to estimate the cost accurately and broadcast join cost equals partition join cost and without user hint.

This LGTM.

wangbo:
set partition join as default when partition join cost equals broadcast join

morningman (approved):
LGTM

kangkaisen (approved):
+1， LGTM

PR #3605, opened 2020-05-16
[Bug] Fix bug that descriptor table is not reset before planning next routine load task

### LINKED ISSUE #3603: routine load current tasks become less after running while
**Describe the bug**
Create routine load job with desired_concurrent_number =5 , running for a while , and there is only 1 -2 running tasks. 

**To Reproduce**
1. Create routine load job with desired_concurrent_number =5 ,
2.  go to web console and see routine load 

**Expected behavior**
 there are 5 running tasks

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**
 - Version 0.11.33.1

**Additional context**
we checked the JRF

[fe.jfr.tar.gz](https://github.com/apache/incubator-doris/files/4636825/fe.jfr.tar.gz)

sounds like DescriptorTable cost most of time and DescriptorTable be increased by  the method plan in (
fe/src/main/java/org/apache/doris/planner/StreamLoadPlanner.java) every time
`
    // create the plan. the plan's query id and load id are same, using the parameter 'loadId'
    public TExecPlanFragmentParams plan(TUniqueId loadId) throws UserException {
        // construct tuple descriptor, used for scanNode and dataSink
        TupleDescriptor tupleDesc = descTable.createTupleDescriptor("DstTableTuple");
        boolean negative = streamLoadTask.getNegative();
        // here we should be full schema to fill the descriptor table
        for (Column col : destTable.getFullSchema()) {
            SlotDescriptor slotDesc = descTable.addSlotDescriptor(tupleDesc);
            slotDesc.setIsMaterialized(true);
            slotDesc.setColumn(col);
            slotDesc.setIsNullable(col.isAllowNull());
            if (negative && !col.isKey() && col.getAggregationType() != AggregateType.SUM) {
                throw new DdlException("Column is not SUM AggreateType. column:" + col.getName());
            }
        }
`

morningman:
Before planning for next routine load task, the analyzer and descriptor table
in it should be reset. Otherwise, a lot of historical objects will
accumulate inside, causing memory leaks.

Fix: #3603

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/DescriptorTable.java
+++ b/fe/src/main/java/org/apache/doris/analysis/DescriptorTable.java
@@ -41,17 +41,15 @@ import java.util.List;
 public class DescriptorTable {
     private final static Logger LOG = LogManager.getLogger(DescriptorTable.class);
 
-    private final HashMap<TupleId, TupleDescriptor> tupleDescs;
+    private final HashMap<TupleId, TupleDescriptor> tupleDescs = new HashMap<TupleId, TupleDescriptor>();
     // List of referenced tables with no associated TupleDescriptor to ship to the BE.
     // For example, the output table of an insert query.
-    private final List<Table>                       referencedTables;
+    private final List<Table> referencedTables = new ArrayList<Table>();;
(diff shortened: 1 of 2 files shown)


EmmyMiao87 (approved):
LGTM

PR #3625, opened 2020-05-18
Support more syntax in case when clause

yangzhg:
support support more syntax in case-when clause with subquey.
suport query like ` case when k1 > subquery1 and k2 < subquey2 then ... else ... ` or  `case when subquey in null then ...`

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/CaseExpr.java
+++ b/fe/src/main/java/org/apache/doris/analysis/CaseExpr.java
@@ -191,7 +191,7 @@ public class CaseExpr extends Expr {
                 throw new AnalysisException("Subquery in case-when must return scala type");
             }
             if (whenExpr.contains(Predicates.instanceOf(Subquery.class))
-                    && !((hasCaseExpr() && whenExpr instanceof Subquery || whenExpr instanceof BinaryPredicate))) {
+                    && !((hasCaseExpr() && whenExpr instanceof Subquery || !checkSubquery(whenExpr)))) {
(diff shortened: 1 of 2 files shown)


morningman:
Please add an ISSUE related to this PR

EmmyMiao87 (approved):

EmmyMiao87:
Please add issue and change the title of PR

fe/src/main/java/org/apache/doris/analysis/CaseExpr.java
```diff
@@ -191,7 +191,7 @@ public void analyzeImpl(Analyzer analyzer) throws AnalysisException {
                 throw new AnalysisException("Subquery in case-when must return scala type");
             }
             if (whenExpr.contains(Predicates.instanceOf(Subquery.class))
-                    && !((hasCaseExpr() && whenExpr instanceof Subquery || whenExpr instanceof BinaryPredicate))) {
+                    && !((hasCaseExpr() && whenExpr instanceof Subquery || !checkSubquery(whenExpr)))) {
                 throw new AnalysisException("Only support subquery in binary predicate in case statement.");
```
wutiangan, line 195: you need to modify exception message for you delete 'BinaryPredicate'

yangzhg: Instead of deleting BinaryPredicate, we still only support BinaryPredicate, but just adjust the check of BinaryPredicate from the first layer to support check of children of other expr

PR #3629, opened 2020-05-18
different cmake build directories for different build types

sduzh:
add `CMAKE_BUILD_TYPE` as the suffix of build directory.


morningman (approved):
LGTM

PR #3632, opened 2020-05-19
Add ignore config to determine whether to continue to start be when load tablet from header failed.

### LINKED ISSUE #3633: [Config] Stop BE start process when loading tablet header failed
At present, BE will read all the header information of the tablet when it starts. When an error occurs in reading, BE only prints a warning log, and then ignores the tablet to continue reading other tablets. This resulted in a large number of tablets not being loaded after some errors occurred, but BE still started to provide services.

This is a very dangerous operation. When this kind of error occurs, if we don't check the log, the operation and maintenance personnel can't find that a large number of tablets are not loaded.

The error may be caused by code bugs, such as incompatible meta information, or the tablet data file itself has been damaged. At present, these errors require human intervention to deal with, the program can not automatically recover.

So we need a mechanism to find this problem as soon as possible and access to solve it.

So I suggest that we can simply add a config to indicate whether we can ignore the header loading error. If true, BE will ignore all loading errors and start normally. If it is false, if there is a load failure, BE will terminate the startup process.

yangzhg:
Add  config `ignore_load_tablet_failure` to determine whether to continue to start be when load tablet from header failed.

Fix #3633

### DIFF
--- a/be/src/olap/data_dir.cpp
+++ b/be/src/olap/data_dir.cpp
@@ -710,9 +710,7 @@ OLAPStatus DataDir::load() {
     };
     OLAPStatus load_tablet_status = TabletMetaManager::traverse_headers(_meta, load_tablet_func);
     if (failed_tablet_ids.size() != 0 && !config::ignore_load_tablet_failure) {
-        std::stringstream ss;
-        std::copy(failed_tablet_ids.begin(), failed_tablet_ids.end(), std::ostream_iterator<int64_t>(ss, ","));
-        LOG(FATAL) << "load tablets from header failed, failed tablets are: " << ss.str();
+        LOG(FATAL) << "load tablets from header failed, failed tablets size: " << failed_tablet_ids.size(;
     }
     if (load_tablet_status != OLAP_SUCCESS) {
         LOG(WARNING) << "there is failure when loading tablet headers, path:" << _path;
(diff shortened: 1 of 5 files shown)


be/src/olap/data_dir.cpp
```diff
@@ -694,19 +694,26 @@ OLAPStatus DataDir::load() {
...
+            failed_tablet_ids.insert(tablet_id);
         } else {
             tablet_ids.insert(tablet_id);
         }
         return true;
     };
     OLAPStatus load_tablet_status = TabletMetaManager::traverse_headers(_meta, load_tablet_func);
+    if (failed_tablet_ids.size() != 0 && !config::ignore_load_tablet_failure) {
+        std::stringstream ss;
+        std::copy(failed_tablet_ids.begin(), failed_tablet_ids.end(), std::ostream_iterator<int64_t>(ss, ","));
```
morningman, line 714: There may be thousands of tablets. This time the log is very large.
In addition, the problematic tablet has actually been printed in the log, so there is no need to print the tablet id. Just print size is enough

morningman (changes requested):

morningman (approved):
LGTM

docs/zh-CN/administrator-guide/config/be_config.md
```diff
@@ -396,3 +396,8 @@ under the License.
 ### `webserver_port`
 
 ### `write_buffer_size`
+
+### ignore_load_tablet_failure
+* 类型：布尔
+* 描述：用来决定在有tablet 加在失败的情况下是否忽略错误，继续启动be
```
wutiangan, line 402: 加在--加载

PR #3638, opened 2020-05-20
Support utf-8 encoding in instr, locate, locate_pos, lpad, rpad

yangzhg:
Support utf-8 encoding for string function `instr`, `locate`, `locate_pos`, `lpad`, `rpad`
and add unit test for them

### DIFF
--- a/be/src/exprs/string_functions.cpp
+++ b/be/src/exprs/string_functions.cpp
@@ -196,28 +196,56 @@ StringVal StringFunctions::lpad(
     if (str.is_null || len.is_null || pad.is_null || len.val < 0) {
         return StringVal::null();
     }
+
+    size_t str_char_size = 0;
+    size_t pad_char_size = 0;
+    size_t byte_pos = 0;
+    std::vector<size_t> str_index;
+    std::vector<size_t> pad_index;
+    for (size_t i = 0, char_size = 0; i < str.len; i += char_size) {
+        char_size = get_utf8_byte_length((unsigned)(str.ptr)[i]);
+        str_index.push_back(byte_pos);
+        byte_pos += char_size;
+        ++str_char_size;
+    }
+    byte_pos = 0;
+    for (size_t i = 0, char_size = 0; i < pad.len; i += char_size) {
+        char_size = get_utf8_byte_length((unsigned)(pad.ptr)[i]);
+        pad_index.push_back(byte_pos);
+        byte_pos += char_size;
+        ++pad_char_size;
+    }
+    
     // Corner cases: Shrink the original string, or leave it alone.
     // TODO: Hive seems to go into an infinite loop if pad.len == 0,
     // so we should pay attention to Hive's future solution to be compatible.
-    if (len.val <= str.len || pad.len == 0) {
-        return StringVal(str.ptr, len.val);
+    if (len.val <= str_char_size || pad.len == 0) {
+        if (len.val >= str_index.size()) {
+            return StringVal::null();
+        }
+        return StringVal(str.ptr, str_index.at(len.val));
     }
 
     // TODO pengyubing
     // StringVal result = StringVal::create_temp_string_val(context, len.val);
-    StringVal result(context, len.val);
+    int32_t pad_byte_len = 0;
+    int32_t pad_times = (len.val - str_char_size) / pad_char_size;
+    int32_t pad_remainder = (len.val - str_char_size) % pad_char_size;
+    pad_byte_len = pad_times * pad.len;
+    pad_byte_len += pad_index.at(pad_remainder);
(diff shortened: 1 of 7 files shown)


morningman:
Good work.
Could please add docs for these functions?

be/src/exprs/string_functions.cpp
```diff
@@ -196,28 +196,56 @@ StringVal StringFunctions::lpad(
...
     }
 
     // TODO pengyubing
     // StringVal result = StringVal::create_temp_string_val(context, len.val);
-    StringVal result(context, len.val);
+    int32_t pad_byte_len = 0;
+    int32_t pad_times = (len.val - str_char_size) / pad_char_size;
+    int32_t pad_remainder = (len.val - str_char_size) % pad_char_size;
+    pad_byte_len = pad_times * pad.len;
+    pad_byte_len += pad_index.at(pad_remainder);
```
kangkaisen, line 235: Should use operator[]

be/src/exprs/string_functions.cpp
```diff
@@ -196,28 +196,56 @@ StringVal StringFunctions::lpad(
     if (str.is_null || len.is_null || pad.is_null || len.val < 0) {
         return StringVal::null();
     }
+
+    size_t str_char_size = 0;
+    size_t pad_char_size = 0;
+    size_t byte_pos = 0;
+    std::vector<size_t> str_index;
+    std::vector<size_t> pad_index;
+    for (size_t i = 0, char_size = 0; i < str.len; i += char_size) {
```
kangkaisen, line 205: Reduplicative code
```
    for (size_t i = 0, char_size = 0; i < str.len; i += char_size) {
        char_size = get_utf8_byte_length((unsigned)(str.ptr)[i]);
        str_index.push_back(byte_pos);
        byte_pos += char_size;
        ++str_char_size;
    }
```

be/src/exprs/string_functions.cpp
yangzhg: operator[]  isn't bound-checked  if the requested position is out of range , it will hide the problem and make it difficault to find problem when when change the code

kangkaisen: I know operator[] doesn't have bound-checked.
But:
1. Here all code is under your control. couldn't we make sure there is no out of range?

2.If you use at, you should try catch the out of range exception, otherwise, I think no meaning to use at. because operator[] performance is better then at.

yangzhg: 1 the code is now under my control, but if some modify the code in the future and has index out of range problem, it will not report any error, it will be a risk and make it difficult to solve it. 
2. this code is not the critical point of performence,  and just an if statement   will not cause too much impact on performance.  A wrong access to unexcepted memory is more harmful than this performance loss

kangkaisen (dismissed):
+1, LGTM

be/src/exprs/string_functions.cpp
```diff
@@ -49,6 +49,17 @@ size_t get_utf8_byte_length(unsigned char byte) {
     }
     return char_size;
 }
+size_t get_char_len(const StringVal& str, std::vector<size_t>* str_index) {
+    size_t char_len = 0;
+    size_t byte_pos = 0;
+    for (size_t i = 0, char_size = 0; i < str.len; i += char_size) {
```
kangpinghuang, line 55:
```suggestion
    for (size_t i = 0; i < str.len; i += char_size) {
```

be/src/exprs/string_functions.cpp
```diff
@@ -49,6 +49,17 @@ size_t get_utf8_byte_length(unsigned char byte) {
...
+        char_size = get_utf8_byte_length((unsigned)(str.ptr)[i]);
+        str_index->push_back(byte_pos);
```
kangpinghuang, line 57: just use i to replace byte_pos and remove byte_pos

be/src/exprs/string_functions.cpp
```diff
@@ -49,6 +49,17 @@ size_t get_utf8_byte_length(unsigned char byte) {
     }
     return char_size;
 }
+size_t get_char_len(const StringVal& str, std::vector<size_t>* str_index) {
+    size_t char_len = 0;
+    size_t byte_pos = 0;
```
kangpinghuang, line 54: can remove this variable

kangpinghuang (changes requested):

be/src/exprs/string_functions.cpp
yangzhg: char_size = 0 cannot removed

kangpinghuang (approved):
LGTM

PR #3641, opened 2020-05-20
[Bug] Ignore loading DELETE status tablet error when restarting BE

### LINKED ISSUE #3639: [MetaTool] Support batch delete meta using meta_tool
Currently, `meta_tool` only support delete meta for a single tablet. If there are thousands of
tablets need to be deleted, this tool run very slow because it will open the Rocksdb for each 
tablet.

### LINKED ISSUE #3640: [Bug] BE restart failed because of reading tablet in DELETE status
**Describe the bug**
Previously, we introduced a new BE parameter(#3632 ). When a tablet fails to load, BE startup fails.

We found that sometimes when the BE starts, it reads the tablet in the DELETE state and returns an error, which causes the BE to fail to start.

The tablet in the DELETE state may have been placed in the DELETE state just before the BE restarted, but has not been deleted from Rocksdb yet. These tablets will be automatically deleted in the subsequent GC process. Therefore, we need to ignore this error and continue to execute the BE startup process.

morningman:
Fix: #3640 

Also add a `batch delete meta` feature for `meta tool`
Fix #3639

### DIFF
--- a/be/src/olap/data_dir.cpp
+++ b/be/src/olap/data_dir.cpp
@@ -717,16 +717,12 @@ OLAPStatus DataDir::load() {
     };
     OLAPStatus load_tablet_status = TabletMetaManager::traverse_headers(_meta, load_tablet_func);
     if (failed_tablet_ids.size() != 0) {
+        LOG(WARNING) << "load tablets from header failed"
+            << ", loaded tablet: " << tablet_ids.size()
+            << ", error tablet: " << failed_tablet_ids.size()
+            << ", path: " << _path;
         if (!config::ignore_load_tablet_failure) {
-            LOG(FATAL) << "load tablets from header failed"
-                << ", loaded tablet: " << tablet_ids.size()
-                << ", error tablet: " << failed_tablet_ids.size()
-                << ", path: " << _path;
-        } else {
-            LOG(WARNING) << "load tablets from header failed"
-                << ", loaded tablet: " << tablet_ids.size()
-                << ", error tablet: " << failed_tablet_ids.size()
-                << ", path: " << _path;
+            LOG(FATAL) << "load tablets encounter failure. stop BE process. path: " << _path;
         }
     }
     if (load_tablet_status != OLAP_SUCCESS) {
--- a/be/src/olap/data_dir.cpp
+++ b/be/src/olap/data_dir.cpp
@@ -709,8 +709,12 @@ OLAPStatus DataDir::load() {
         return true;
     };
     OLAPStatus load_tablet_status = TabletMetaManager::traverse_headers(_meta, load_tablet_func);
-    if (failed_tablet_ids.size() != 0 && !config::ignore_load_tablet_failure) {
-        LOG(FATAL) << "load tablets from header failed, failed tablets size: " << failed_tablet_ids.size();
+    if (failed_tablet_ids.size() != 0) {
+        if (!config::ignore_load_tablet_failure) {
+            LOG(FATAL) << "load tablets from header failed, failed tablets size: " << failed_tablet_ids.size() << ", path: " << _path;
+        } else {
+            LOG(WARNING) << "load tablets from header failed, failed tablets size: " << failed_tablet_ids.size() << ", path: " << _path;
+        }
     }
     if (load_tablet_status != OLAP_SUCCESS) {
         LOG(WARNING) << "there is failure when loading tablet headers, path:" << _path;
--- a/be/src/olap/data_dir.cpp
+++ b/be/src/olap/data_dir.cpp
@@ -699,7 +699,14 @@ OLAPStatus DataDir::load() {
                                                 const std::string& value) -> bool {
         OLAPStatus status = _tablet_manager->load_tablet_from_meta(this, tablet_id, schema_hash,
                                                                    value, false, false);
-        if (status != OLAP_SUCCESS) {
+        if (status != OLAP_SUCCESS && status != OLAP_ERR_TABLE_ALREADY_DELETED_ERROR) {
+            // load_tablet_from_meta() may return OLAP_ERR_TABLE_ALREADY_DELETED_ERROR
+            // which means the tablet status is DELETED
+            // This may happen when the tablet was just deleted before the BE restarted,
+            // but it has not been cleared from rocksdb. At this time, restarting the BE
+            // will read the tablet in the DELETE state from rocksdb. These tablets have been
+            // added to the garbage collection queue and will be automatically deleted afterwards.
+            // Therefore, we believe that this situation is not a failure.
             LOG(WARNING) << "load tablet from header failed. status:" << status
                          << ", tablet=" << tablet_id << "." << schema_hash;
             failed_tablet_ids.insert(tablet_id);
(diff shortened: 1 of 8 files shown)


be/src/olap/data_dir.cpp
```diff
@@ -709,13 +716,29 @@ OLAPStatus DataDir::load() {
         return true;
     };
     OLAPStatus load_tablet_status = TabletMetaManager::traverse_headers(_meta, load_tablet_func);
-    if (failed_tablet_ids.size() != 0 && !config::ignore_load_tablet_failure) {
-        LOG(FATAL) << "load tablets from header failed, failed tablets size: " << failed_tablet_ids.size();
+    if (failed_tablet_ids.size() != 0) {
+        if (!config::ignore_load_tablet_failure) {
```
yangzhg, line 720: how about log warning first, and then
```
if (!config::ignore_load_tablet_failure) {
    LOG(FATAL) << "be will shutdown due to ignore_load_tablet_failure is false"
}
```

be/src/tools/meta_tool.cpp
```diff
@@ -122,13 +126,131 @@ void delete_meta(DataDir* data_dir) {
...
+    //      /data2/palo.HDD 100010 23049230234
+    std::ifstream infile(tablet_file);
+    std::string line;
+    int err_num = 0;
+    int delete_num = 0;
+    int total_num = 0;
+    std::unordered_map<std::string, std::unique_ptr<DataDir>> dir_map;
+    while (std::getline(infile, line)) {
+        total_num++;
+        vector<string> v = strings::Split(line, " ");
```
yangzhg, line 172: I think its better to use ',' as split

morningman: ok

be/src/olap/data_dir.cpp
morningman: ok

be/src/olap/data_dir.cpp
```diff
@@ -709,13 +716,25 @@ OLAPStatus DataDir::load() {
...
     }
     if (load_tablet_status != OLAP_SUCCESS) {
-        LOG(WARNING) << "there is failure when loading tablet headers, path:" << _path;
+        LOG(WARNING) << "there is failure when loading tablet headers"
+                << ", loaded tablet: " << tablet_ids.size()
+                << ", error tablet: " << failed_tablet_ids.size()
+                << ", path: " << _path;
     } else {
-        LOG(INFO) << "load rowset from meta finished, data dir: " << _path;
+        LOG(INFO) << "load rowset from meta finished"
```
kangpinghuang, line 734:
```suggestion
        LOG(INFO) << "load tablet meta finished"
```

be/src/tools/meta_tool.cpp
```diff
@@ -122,13 +126,131 @@ void delete_meta(DataDir* data_dir) {
     std::cout << "delete meta successfully" << std::endl;
 }
 
+Status init_data_dir(const std::string dir, DataDir** ret) {
```
kangpinghuang, line 129:
```suggestion
Status init_data_dir(const std::string& dir, DataDir** ret) {
```

be/src/tools/meta_tool.cpp
```diff
@@ -122,13 +126,131 @@ void delete_meta(DataDir* data_dir) {
...
+    st = (*ret)->init();
+    if (!st.ok()) {
+        std::cout << "data_dir load failed" << std::endl;
+        return Status::InternalError("data_dir load failed");
+    }
+
+    return Status::OK();
+}
+
+void batch_delete_meta(const std::string tablet_file) {
```
kangpinghuang, line 158:
```suggestion
void batch_delete_meta(const std::string& tablet_file) {
```

be/src/tools/meta_tool.cpp
```diff
@@ -122,13 +126,131 @@ void delete_meta(DataDir* data_dir) {
...
+    // each line in tablet file indicate a tablet to delete, format is:
+    //      data_dir,tablet_id,schema_hash
+    // eg:
+    //      /data1/palo.HDD,100010,11212389324
+    //      /data2/palo.HDD,100010,23049230234
+    std::ifstream infile(tablet_file);
+    std::string line;
```
kangpinghuang, line 165:
```suggestion
    std::string line = "";
```

be/src/tools/meta_tool.cpp
kangpinghuang: and why here use DataDir**?
I think DataDir* or std::unique_ptr<DataDir>* is ok

be/src/tools/meta_tool.cpp
```diff
@@ -122,13 +126,131 @@ void delete_meta(DataDir* data_dir) {
...
+        if (!safe_strto64(v[2].c_str(), &schema_hash)) {
+            std::cout << "invalid schema hash: " << line << std::endl;
+            err_num++;
+            continue;
+        }
+
+        OLAPStatus s = TabletMetaManager::remove(data_dir, tablet_id, schema_hash);
+        if (s != OLAP_SUCCESS) {
+            std::cout << "delete tablet meta failed for tablet_id:"
+                << FLAGS_tablet_id << ", schema_hash:" << FLAGS_schema_hash
```
kangpinghuang, line 224:
```suggestion
                << tablet_id << ", schema_hash:" << schema_hash
```

docs/zh-CN/administrator-guide/operation/tablet-meta-tool.md
```diff
@@ -93,14 +93,39 @@ api：
 
 ### 删除 header
 
-为了实现从某个 be 的某个盘中删除某个 tablet 的功能。
+为了实现从某个 be 的某个盘中删除某个 tablet 的功能。可以单独删除一个 tablet，或者批量删除一组 tablet。
```
kangpinghuang, line 96:
```suggestion
为了实现从某个 be 的某个盘中删除某个 tablet 的元数据功能。可以单独删除一个 tablet元数据，或者批量删除一组 tablet元数据。
```

docs/zh-CN/administrator-guide/operation/tablet-meta-tool.md
```diff
@@ -93,14 +93,39 @@ api：
...
 
-命令:
+删除单个tablet：
```
kangpinghuang, line 98:
```suggestion
删除单个tablet元数据：
```

docs/zh-CN/administrator-guide/operation/tablet-meta-tool.md
```diff
@@ -93,14 +93,39 @@ api：
...
 
 ```
 ./lib/meta_tool --operation=delete_meta --root_path=/path/to/root_path --tablet_id=xxx --schema_hash=xxx
 ```
 
+删除一组tablet：
```
kangpinghuang, line 104:
```suggestion
删除一组tablet元数据：
```

docs/en/administrator-guide/operation/tablet-meta-tool.md
```diff
@@ -93,14 +93,39 @@ Order:
...
+Single delete:
 
 ```
 ./lib/meta_tool --operation=delete_meta --root_path=/path/to/root_path --tablet_id=xxx --schema_hash=xxx`
 ```
 
+Batch delete:
+
+```
+./lib/meta_tool --operation=batch_delete_meta --tablet_file=/path/to/tablet_file.txt
```
kangpinghuang, line 107: will you need to support delete tablet data meanwhile?

kangpinghuang (changes requested):

be/src/olap/data_dir.cpp
morningman: OK

be/src/tools/meta_tool.cpp
morningman: I will change it to unique_ptr

be/src/tools/meta_tool.cpp
morningman: OK

be/src/tools/meta_tool.cpp
morningman: OK

be/src/tools/meta_tool.cpp
morningman: OK

docs/en/administrator-guide/operation/tablet-meta-tool.md
morningman: Maybe at another PR

kangpinghuang (approved):
LGTM

PR #3642, opened 2020-05-20
Fix some unit test failed

yangzhg:
Fix some unittest failed due to glog， this may be we change the ut build dir，and the log path is not exist in new build dir， so we change the log from file to stdout


imay:
Hi @yangzhg 
Can you explain why glog will cause unit test fail in commit message?

imay (approved):
LGTM

PR #3647, opened 2020-05-21
[Enhancement] Add detail msg to show the reason of publish failure.

### LINKED ISSUE #3646: [Enhancement][Txn] Add more info to find way publish failed.
Sometimes user may see a lot of txn blocked in COMMITTED state, and not be published for
a long time. It  usually duo to some txns failed to publish on some certain tablets, and then block
all other txns behind.

When we meet this issue, it hard to see which txn is the root cause and which tablet is failed to be
published.

morningman:
Add 2 new columns `PublishTime` and `ErrMsg` to show publish version time and  errors happen during the transaction process. Can be seen by executing: 

`SHOW PROC "/transactions/dbId/";`
or
`SHOW TRANSACTION WHERE ID=xx;`

Currently is only record error happen in publish phase, which can help us to find out which txn
is blocked.

Fix #3646

### DIFF
--- a/fe/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
+++ b/fe/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
@@ -233,6 +233,7 @@ public class DatabaseTransactionMgr {
         info.add(txnState.getSourceType().name());
         info.add(TimeUtils.longToTimeString(txnState.getPrepareTime()));
         info.add(TimeUtils.longToTimeString(txnState.getCommitTime()));
+        info.add(TimeUtils.longToTimeString(txnState.getPublishVersionTime()));
         info.add(TimeUtils.longToTimeString(txnState.getFinishTime()));
         info.add(txnState.getReason());
         info.add(String.valueOf(txnState.getErrorReplicas().size()));
--- a/fe/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
+++ b/fe/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
@@ -17,13 +17,6 @@
 
 package org.apache.doris.transaction;
 
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Joiner;
-import com.google.common.base.Preconditions;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
-import org.apache.commons.collections.CollectionUtils;
 import org.apache.doris.catalog.Catalog;
 import org.apache.doris.catalog.Database;
 import org.apache.doris.catalog.MaterializedIndex;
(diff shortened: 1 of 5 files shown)


xy720:
LGTM

fe/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
```diff
@@ -721,8 +728,12 @@ public void finishTransaction(long transactionId, Set<Long> errorReplicaIds) thr
                             }
 
                             if (healthReplicaNum < quorumReplicaNum) {
-                                LOG.info("publish version failed for transaction {} on tablet {},  with only {} replicas less than quorum {}",
+                                LOG.info("publish version failed for transaction {} on tablet {}, with only {} replicas less than quorum {}",
                                         transactionState, tablet, healthReplicaNum, quorumReplicaNum);
+                                String errMsg = String.format("publish on tablet %d failed. succeed replica num %d less than quorum %d."
+                                        + " table: %d, partition: %d, publish version: %d",
+                                        tablet.getId(), healthReplicaNum, quorumReplicaNum, tableId, partitionId, partition.getVisibleVersion() + 1);
+                                transactionState.setErrorMsg(errMsg);
```
xy720, line 736: What is the different between version and lastSuccessVersion? Why only update lastSuccessVersion but not version when we found replica's lastFailedVersion is larger than 0?

fe/src/main/java/org/apache/doris/transaction/TransactionState.java
```diff
@@ -187,7 +187,7 @@ public String toString() {
     // this state need not to be serialized
     private Map<Long, PublishVersionTask> publishVersionTasks;
     private boolean hasSendTask;
-    private long publishVersionTime;
+    private long publishVersionTime = -1;
```
xy720, line 190: If publishVersionTime will always be updated after finishTransaction failed, is it means that this publish job may keep runnning in doris forever?

morningman: Yes, publish task MUST success.

fe/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
morningman: versions before `version` is  continuous. `lastSuccessVersion` is the last success version of this replica.  `lastFailedVersion` is the last failed version of this replica. eg:
versio: 10
lastSuccessVersion:12
lastFailedVersion:11

which means version 0-10 is continuous and success. and version 11 failed to publish. but 
version 12 success.

EmmyMiao87 (approved):

WingsGo:
Hi, @morningman ,I found some case that pulish version may failed, the log is:

```
2020-05-13 15:51:50,784 DEBUG 31 [GlobalTransactionMgr.finishTransaction():700] transactionId 368227 partition commitInfo version 1072 is not equal with partition visible version 743 plus one, need wait
```
Do you have any idea why the partition commitInfo version is not equal with 743 plus one?and how can we handler the situation more graceful instead wait too long time?

morningman:
> Do you have any idea why the partition commitInfo version is not equal with 743 plus one?and how can we handler the situation more graceful instead wait too long time?

This usually because the txn 744 blocks. The publish task is executed serially by version number.
At present, we can only try to solve the txn 744 stuck problem, this problem cannot be skipped.

PR #3651, opened 2020-05-21
Support create materialized view with bitmap or hll

### LINKED ISSUE #3344: [Proposal] Materialized View 2.0

HangyuanLiu:
For #3344

### DIFF
--- a/fe/src/main/java/org/apache/doris/alter/MaterializedViewHandler.java
+++ b/fe/src/main/java/org/apache/doris/alter/MaterializedViewHandler.java
@@ -459,10 +459,8 @@ public class MaterializedViewHandler extends AlterHandler {
                                                + "duplicate table");
             }
             Column newMVColumn = new Column(baseColumn);
-
             newMVColumn.setIsKey(mvColumnItem.isKey());
             newMVColumn.setAggregationType(mvAggregationType, mvColumnItem.isAggregationTypeImplicit());
-            newMVColumn.setDefineExpr(mvColumnItem.getDefineExpr());
             if (mvColumnItem.getDefineExpr() != null) {
                 if (mvAggregationType.equals(AggregateType.BITMAP_UNION)) {
                     newMVColumn.setType(Type.BITMAP);
@@ -475,6 +473,7 @@ public class MaterializedViewHandler extends AlterHandler {
                 }
                 newMVColumn.setIsKey(false);
                 newMVColumn.setIsAllowNull(false);
+                newMVColumn.setDefineExpr(mvColumnItem.getDefineExpr());
             }
             newMVColumns.add(newMVColumn);
         }
--- a/fe/src/main/java/org/apache/doris/alter/MaterializedViewHandler.java
+++ b/fe/src/main/java/org/apache/doris/alter/MaterializedViewHandler.java
@@ -41,6 +41,7 @@ import org.apache.doris.catalog.Table;
 import org.apache.doris.catalog.Tablet;
 import org.apache.doris.catalog.TabletInvertedIndex;
 import org.apache.doris.catalog.TabletMeta;
+import org.apache.doris.catalog.Type;
 import org.apache.doris.common.AnalysisException;
 import org.apache.doris.common.Config;
 import org.apache.doris.common.DdlException;
@@ -88,6 +89,7 @@ import java.util.concurrent.ConcurrentHashMap;
 public class MaterializedViewHandler extends AlterHandler {
     private static final Logger LOG = LogManager.getLogger(MaterializedViewHandler.class);
     public static final String NEW_STORAGE_FORMAT_INDEX_NAME_PREFIX = "__v2_";
+    public static final String MATERIALIZED_VIEW_NAME_PRFIX = "__doris_materialized_view_";
 
     public MaterializedViewHandler() {
         super("materialized view");
@@ -457,8 +459,23 @@ public class MaterializedViewHandler extends AlterHandler {
                                                + "duplicate table");
             }
             Column newMVColumn = new Column(baseColumn);
+
             newMVColumn.setIsKey(mvColumnItem.isKey());
             newMVColumn.setAggregationType(mvAggregationType, mvColumnItem.isAggregationTypeImplicit());
+            newMVColumn.setDefineExpr(mvColumnItem.getDefineExpr());
+            if (mvColumnItem.getDefineExpr() != null) {
+                if (mvAggregationType.equals(AggregateType.BITMAP_UNION)) {
+                    newMVColumn.setType(Type.BITMAP);
+                    newMVColumn.setName(MATERIALIZED_VIEW_NAME_PRFIX + "bitmap_" + baseColumn.getName());
+                } else if (mvAggregationType.equals(AggregateType.HLL_UNION)){
+                    newMVColumn.setType(Type.HLL);
+                    newMVColumn.setName(MATERIALIZED_VIEW_NAME_PRFIX + "hll_" + baseColumn.getName());
+                } else {
+                    throw new DdlException("The define expr of column is only support bitmap_union or hll_union");
+                }
+                newMVColumn.setIsKey(false);
+                newMVColumn.setIsAllowNull(false);
+            }
             newMVColumns.add(newMVColumn);
         }
         return newMVColumns;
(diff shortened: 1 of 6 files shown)


HangyuanLiu:
Add UT later

fe/src/main/java/org/apache/doris/catalog/Column.java
```diff
@@ -310,6 +333,50 @@ public static String removeNamePrefix(String colName) {
...
+
+    public void setDefineExpr(Expr expr) {
+        defineExpr = expr;
+        if (expr != null) {
+            serializeDefineExpr(expr);
+        }
+    }
+
+    public void serializeDefineExpr(Expr defineExpr) {
+        defineExprString = "COLUMNS (" + name + "=" + defineExpr.toSql() + ")";
```
morningman, line 348: Actually, `toSql()` is really not a safe way for serialization. It may be change, or even can not restore the original statement, in case the expr may be much complicate and being rewritten somehow.

I was thinking maybe a safer way is to save the entire `create view` statement directly. We shall discuss this later.

fe/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
```diff
@@ -167,15 +167,25 @@ private void analyzeSelectClause() throws AnalysisException {
...
+                        && !functionName.equalsIgnoreCase("max")
+                        && !functionName.equalsIgnoreCase("bitmap_union")
+                        && !functionName.equalsIgnoreCase("hll_union")) {
                     throw new AnalysisException("The materialized view only support the sum, min and max aggregate "
                                                         + "function. Error function: " + functionCallExpr.toSqlImpl());
                 }
                 Preconditions.checkState(functionCallExpr.getChildren().size() == 1);
                 Expr functionChild0 = functionCallExpr.getChild(0);
+
+                if (functionName.equalsIgnoreCase("bitmap_union") || functionName.equalsIgnoreCase("hll_union")) {
```
morningman, line 183: I think we need an extensible framework to handle the expression part of the materialized view more elegantly, rather than simply judging a few functions.
This can prepare for more complex expressions later.

morningman (changes requested):
And by the way, when the bitmap and hll types are not yet supported on the BE side, should these new operations be prohibited first?

fe/src/main/java/org/apache/doris/alter/MaterializedViewHandler.java
```diff
@@ -457,8 +462,23 @@ private RollupJobV2 createMaterializedViewJob(String mvName, String baseIndexNam
                                                + "duplicate table");
             }
             Column newMVColumn = new Column(baseColumn);
+
             newMVColumn.setIsKey(mvColumnItem.isKey());
             newMVColumn.setAggregationType(mvAggregationType, mvColumnItem.isAggregationTypeImplicit());
+            newMVColumn.setDefineExpr(mvColumnItem.getDefineExpr());
```
EmmyMiao87, line 468: Firstly, check the define expr is forbidden or not. Secondly, set define expr in column

fe/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
```diff
@@ -167,15 +167,25 @@ private void analyzeSelectClause() throws AnalysisException {
             } else if (selectListItem.getExpr() instanceof FunctionCallExpr) {
                 FunctionCallExpr functionCallExpr = (FunctionCallExpr) selectListItem.getExpr();
                 String functionName = functionCallExpr.getFnName().getFunction();
+                Expr defineExpr = null;
                 // TODO(ml): support REPLACE, REPLACE_IF_NOT_NULL only for aggregate table, HLL_UNION, BITMAP_UNION
                 if (!functionName.equalsIgnoreCase("sum")
                         && !functionName.equalsIgnoreCase("min")
-                        && !functionName.equalsIgnoreCase("max")) {
+                        && !functionName.equalsIgnoreCase("max")
+                        && !functionName.equalsIgnoreCase("bitmap_union")
```
EmmyMiao87, line 175: The bitmap_union and hll_union should still be banned.

fe/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
```diff
@@ -187,6 +197,14 @@ else if (functionChild0 instanceof CastExpr
                     throw new AnalysisException("The children of aggregate function only support one original column. "
                                                         + "Error function: " + functionCallExpr.toSqlImpl());
                 }
+
+                if (functionName.equalsIgnoreCase("bitmap_union")) {
+                    if (!slotRef.type.isIntegerType()) {
```
EmmyMiao87, line 202: The child type of bitmap_union must be bitmap rather then integer.
Also the child type of to_bitmap function has been checked in analyzer.

fe/src/main/java/org/apache/doris/catalog/Column.java
```diff
@@ -66,6 +78,10 @@
     private String comment;
     @SerializedName(value = "stats")
     private ColumnStats stats;     // cardinality and selectivity etc.
+    @SerializedName(value = "defineExprString")
+    private String defineExprString = "";
+
```
EmmyMiao87, line 83: Remove this empty line.

fe/src/main/java/org/apache/doris/catalog/Column.java
EmmyMiao87: You need to restore the original statement rather then use the toSql function

fe/src/main/java/org/apache/doris/catalog/Column.java
```diff
@@ -416,6 +483,7 @@ public void write(DataOutput out) throws IOException {
         stats.write(out);
 
         Text.writeString(out, comment);
+        Text.writeString(out, defineExprString);
```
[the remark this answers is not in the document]
EmmyMiao87: The table object also use the writable interface The fullSchema of table also needs to serialized without GSON,

fe/src/main/java/org/apache/doris/common/FeMetaVersion.java
```diff
@@ -179,6 +179,8 @@
     public static final int VERSION_83 = 83;
     // add storage format in schema change job
     public static final int VERSION_84 = 84;
+    // support define expr in materialized view
```
EmmyMiao87, line 182: The bitmap_union mv should be forbidden now. So the meta version could not be changed.

EmmyMiao87 (changes requested):

EmmyMiao87:
The commit msg is too short.

fe/src/main/java/org/apache/doris/alter/MaterializedViewHandler.java
```diff
@@ -457,8 +462,23 @@ private RollupJobV2 createMaterializedViewJob(String mvName, String baseIndexNam
...
+                if (mvAggregationType.equals(BITMAP_UNION)) {
+                    newMVColumn.setType(Type.BITMAP);
+                    newMVColumn.setName(MATERIALIZED_VIEW_NAME_PRFIX + "bitmap_" + baseColumn.getName());
+                } else if (mvAggregationType.equals(HLL_UNION)){
+                    newMVColumn.setType(Type.HLL);
+                    newMVColumn.setName(MATERIALIZED_VIEW_NAME_PRFIX + "hll_" + baseColumn.getName());
+                } else {
+                    throw new DdlException("The define expr of column is only support bitmap_union or hll_union");
+                }
+                newMVColumn.setIsKey(false);
```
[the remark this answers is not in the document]
HangyuanLiu: If the aggregate function is computed for the original column, it should be not key whether  the original column is a key column or a value column

fe/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
HangyuanLiu: I agree with you. But schema change and  expression compute  logical is non-universal. so we can only support few function now.

EmmyMiao87 (approved):

morningman (approved):
LGTM

Issue #3652, opened 2020-05-21
HLL_HASH function  return type should be HLL


HangyuanLiu:
Currently, the return value of the HLL function is varchar. This result in problems in analyze stage. For example, HLL_RAW_AGG(hll_hash(hll_type_field)) return analyze error: "HLL_UNION_AGG, HLL_RAW_AGG and HLL_CARDINALITY's params must be HLL column".

PR #3656, opened 2020-05-21
Modify hll_hash function return type to HLL

HangyuanLiu:
1、Modify hll_hash function return type to HLL
2、Make HLL_RAW_AGG is alias of HLL_UNION
For #3652


imay (approved):
LGTM

imay (changes requested):
Fail to compile BE, you should make this script work when there is HLL type

imay (approved):
LGTM

PR #3661, opened 2020-05-21
[Optimize] Using sorted schema change processing to merge the data

### LINKED ISSUE #3660: [Alter] The rollup index's row number is same as base table.
**Describe the bug**
After creating a rollup with less key column, we found that the row number of this rollup
is just same as the base table.

**To Reproduce**
1. create a table with column k1, k2, k3, v1(sum)
2. load some data.
3. create rollup with column k1, v1(sum)

after rollup finished, the num rows of rollup is same as base table's.

**Expected behavior**
Because the number of key columns in rollup is less than base table, the data
should be more aggregate in rollup, so the row num of the rollup should be less than
base table.

morningman:
When reducing the key columns of an aggregate table by doing schema change
or rollup, we should trigger a re-sort schema change process to re-sort
and merge the data into new schema. So that the data in new schema will
be more aggregated.

Fix: #3660

### DIFF
--- a/be/src/olap/schema_change.cpp
+++ b/be/src/olap/schema_change.cpp
@@ -1859,6 +1859,29 @@ OLAPStatus SchemaChangeHandler::_parse_request(TabletSharedPtr base_tablet,
     // 若Key列的引用序列出现乱序，则需要重排序
     int num_default_value = 0;
 
+    // A, B, C are keys(sort keys), D is value
+    // The following cases are not changing the order, no need to resort:
+    // (sort keys keep in same order)
+    //      old keys:    A   B   C   D
+    //      new keys:    A   X   B   C   D
+    //
+    //      old keys:    A   B   C   D
+    //      new keys:    X   A   B   C   D
+    //
+    //      old keys:    A   B   C   D
+    //      new keys:    A   B   C
+    //
+    //      old keys:    A   B   C   D
+    //      new keys:    A   B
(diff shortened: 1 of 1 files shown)


be/src/olap/schema_change.cpp
```diff
@@ -1859,6 +1859,29 @@ OLAPStatus SchemaChangeHandler::_parse_request(TabletSharedPtr base_tablet,
...
+    //      new keys:    A   X   B   C   D
+    //
+    //      old keys:    A   B   C   D
+    //      new keys:    X   A   B   C   D
+    //
+    //      old keys:    A   B   C   D
+    //      new keys:    A   B   C
+    //
+    //      old keys:    A   B   C   D
+    //      new keys:    A   B
```
chaoyli, line 1875: After this pull request.
old keys: A B C D
new keys: A B C
will be resorted.

chaoyli (changes requested):

be/src/olap/schema_change.cpp
morningman: Sorry that this PR is not ready, I will consider more situation.

EmmyMiao87: This case also need be re-agg, so the sc_sorting must be marked as true.

EmmyMiao87:
#4587

PR #3666, opened 2020-05-22
Del some useless code and impove the performance of sort

HappenLee:
1. Delete the code of Sort Node we do not use now.
2. Optimize the quick sort by find_the_median and try to reduce recursion level of quick sort.

I have tested 50 million pieces of data sort. ordered, disordered and reversed order 
there is improved by about 10%


chaoyli:
It's better to post the performance result.

HappenLee:
@chaoyli 
OK！This is performance result.

### ENV

   CPU : Intel(R) Xeon(R) CPU E5-2620 v2 @ 2.10GHz 6cores * 2 
   RAM : 96GB

### Data
Random numbers from 1 to 50 million

### Test Result

Order of arrangement | origin |  find_the_median | Percentage increase
:--:|:--:|:--:|:--:
Order | 41s985ms | 39s389ms | 6.18%
Reverse | 36s854ms | 34s376ms | 6.67%
Random Shuffle |36s520ms | 33s888ms | 7.2%

Now Doris's optimizer is not powerful enough. I test the scenario where the order by column  cardinality is small and the repetition rate is high. If we can Intelligently choose  three way quick sort, there will be five to ten times better performance. What a pity.

chaoyli (approved):

PR #3668, opened 2020-05-22
[Memory Engine] Add TabletType to PartitionInfo and TabletMeta

### LINKED ISSUE #3442: [Memory Engine] Add MemTablet type to TabletMeta
To add MemTablet type to meta, there are 2 ways:

1. Create new meta kind, like BaseTablet and MemTablet
2. Use original meta, add new fields.

Unlike original Tablet, MemTablet doesn't have components like rowsets and may not have del_predicates, alter tasks etc. Those are vectors/ptrs so it won't use much memory(or pb space) if they are empty. Also it's worth to mention now MemTablet can handle multiple schema(schema change) inside MemTablet, it's won't be necessary to create a new tablet after schema change, but we can also keep the old behavior, depending on which way is easier.

To make things simple, we can first add a field, e.g. `bool is_memory [default=false]` to mark this tablet is a MemTablet, and latter add new fields if needed.

One potential issue is there already exists a term `in_memory` in TabletSchema when creating tablet, which means "keep all pages in page caches at best", causes confusion. 

Another option is to add an `enum StorageEngine`, with types of MEMORY and DISK.

decster:
This CL add TabletType to TabletMeta and PartitionInfo, it also adds a create table property "tablet_type" : "disk/memory" for user to specify tablet type. This is a temporary entry point for testing only, it may change in the future, so this field does not persist to FE PartitionInfo(need to upgrade meta version).

Resolves #3442

### DIFF
--- a/be/src/olap/base_tablet.h
+++ b/be/src/olap/base_tablet.h
@@ -43,6 +43,7 @@ public:
     // Property encapsulated in TabletMeta
     inline const TabletMetaSharedPtr tablet_meta();
 
+    inline bool is_memory() const;
     inline TabletUid tablet_uid() const;
     inline int64_t table_id() const;
     // Returns a string can be used to uniquely identify a tablet.
@@ -82,6 +83,10 @@ inline string BaseTablet::tablet_path() const {
 
 inline const TabletMetaSharedPtr BaseTablet::tablet_meta() {
     return _tablet_meta;
+}
+
+inline bool BaseTablet::is_memory() const {
+    return _tablet_meta->tablet_type() == TabletTypePB::TABLET_TYPE_MEMORY;
 }
 
 inline TabletUid BaseTablet::tablet_uid() const {
--- a/be/src/olap/tablet_meta.cpp
+++ b/be/src/olap/tablet_meta.cpp
@@ -58,19 +58,6 @@ OLAPStatus AlterTabletTask::set_alter_state(AlterTabletState alter_state) {
     return OLAP_SUCCESS;
 }
 
-OLAPStatus TabletMeta::create(int64_t table_id, int64_t partition_id,
-                              int64_t tablet_id, int32_t schema_hash,
-                              uint64_t shard_id, const TTabletSchema& tablet_schema,
-                              uint32_t next_unique_id,
-                              const unordered_map<uint32_t, uint32_t>& col_ordinal_to_unique_id,
-                              TabletMetaSharedPtr* tablet_meta, TabletUid& tablet_uid) {
-    tablet_meta->reset(new TabletMeta(table_id, partition_id,
-                                      tablet_id, schema_hash,
-                                      shard_id, tablet_schema,
-                                      next_unique_id, col_ordinal_to_unique_id, tablet_uid));
-    return OLAP_SUCCESS;
-}
-
 OLAPStatus TabletMeta::create(const TCreateTabletReq& request, const TabletUid& tablet_uid,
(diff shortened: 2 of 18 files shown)


be/src/olap/base_tablet.h
```diff
@@ -84,6 +85,10 @@ inline const TabletMetaSharedPtr BaseTablet::tablet_meta() {
     return _tablet_meta;
 }
 
+inline bool BaseTablet::is_memory() const {
+    return _tablet_meta->tablet_type() == TabletTypePB::TABLET_TYPE_MEMORY;
```
chaoyli, line 89: It may be not set default.

be/src/olap/tablet_meta.cpp
```diff
@@ -101,6 +89,8 @@ TabletMeta::TabletMeta(int64_t table_id, int64_t partition_id,
     tablet_meta_pb.set_cumulative_layer_point(-1);
     tablet_meta_pb.set_tablet_state(PB_RUNNING);
     *(tablet_meta_pb.mutable_tablet_uid()) = tablet_uid.to_proto();
+    tablet_meta_pb.set_tablet_type(tabletType == TTabletType::TABLET_TYPE_MEMORY ?
```
chaoyli, line 92: Normal upgrade procedure is upgrading BE process proceeding of upgrading FE.
When upgrading BE, tabletType may be not set.

fe/src/main/java/org/apache/doris/alter/SchemaChangeJobV2.java
```diff
@@ -240,7 +240,8 @@ protected void runPendingJob() throws AlterCancelException {
                                     Partition.PARTITION_INIT_VERSION, Partition.PARTITION_INIT_VERSION_HASH,
                                     tbl.getKeysType(), TStorageType.COLUMN, storageMedium,
                                     shadowSchema, bfColumns, bfFpp, countDownLatch, indexes,
-                                    tbl.isInMemory());
+                                    tbl.isInMemory(),
+                                    tbl.getPartitionInfo().getTabletType(partitionId));
                             createReplicaTask.setBaseTablet(partitionIndexTabletMap.get(partitionId, shadowIdxId).get(shadowTabletId), originSchemaHash);
```
chaoyli, line 245: You should add constraint about ALTER TABLE.
If I have a disk table.
```
ALTER TABLE test SET("tablet_type"="memory");
```
will be encounter error in BE processing.

chaoyli (changes requested):

fe/src/main/java/org/apache/doris/common/util/PropertyAnalyzer.java
```diff
@@ -233,6 +236,23 @@ public static TStorageType analyzeStorageType(Map<String, String> properties) th
         return tStorageType;
     }
 
+    public static TTabletType analyzeTabletType(Map<String, String> properties) throws AnalysisException {
+        // default is COLUMN
```
morningman, line 240: modify the comment

decster: fixed

be/src/olap/tablet_meta.cpp
decster: fixed

be/src/olap/base_tablet.h
decster: fixed

chaoyli (approved):

PR #3670, opened 2020-05-23
Allow user to set thrift_client_timeout_ms config for thrift server

caiconghui:
The value for thrift_client_timeout_ms should set to be larger than zero to prevent
 some hang up problems in java.net.SocketInputStream.socketRead0


caiconghui:
for #3671

kangkaisen:
@caiconghui Hi, Please add this config to fe config doc. Thanks.

chaoyli (approved):

PR #3674, opened 2020-05-23
[Bug] Fix bug that runningTxnNum does not equal to the real running txn num

### LINKED ISSUE #3673: [Bug] running txn number does not equals to the real running txn number.
**Describe the bug**
The result of `SHOW PROC "/transactions/dbId"` shows there are, for example, 22 running
transactions. But the result of `SHOW PROC "/transactions/dbId/running"` is empty.

And because of this, if the running txn number is higher than config `max_running_txn_num_per_db`, user can not submit load job.

morningman:
This is because the logic for modifying the number of txn running is wrong.

Because we did not persist the previous status(preStatus) of a transaction.
Therefore, when replaying the metadata log, we cannot decide whether to modify
the `runningTxnNum` value based on `preStatus`. This info is lost.

Fix #3673


kangkaisen (approved):
+1, LGTM

PR #3677, opened 2020-05-25
Support materialized view load and insert

### LINKED ISSUE #3344: [Proposal] Materialized View 2.0

HangyuanLiu:
This commit mainly supports load bitmap_union, hll_union, and count materialized views.

The main changes are as follows:
1、insert stmt support load extend column
2、load stmt support load extend column

Issue : #3344


fe/src/main/java/org/apache/doris/load/Load.java
```diff
@@ -1013,6 +1013,13 @@ public static void initColumns(Table tbl, List<ImportColumnDesc> columnExprs,
                 slotDescByName.put(realColName, slotDesc);
             }
         }
+        // The extension column of the materialized view is added to the expression evaluation of load
```
EmmyMiao87, line 1016: Please add some example

fe/src/main/java/org/apache/doris/load/Load.java
```diff
@@ -1013,6 +1013,13 @@ public static void initColumns(Table tbl, List<ImportColumnDesc> columnExprs,
...
+        for (Column column : tbl.getFullSchema()) {
+            if (column.getDefineExpr() != null) {
```
EmmyMiao87, line 1018: Are you sure that this is correct` column(a, tmp_c) set(b=f(c), c=b) `？

EmmyMiao87 (changes requested):

fe/src/main/java/org/apache/doris/analysis/InsertStmt.java
```diff
@@ -453,6 +453,26 @@ private void analyzeSubquery(Analyzer analyzer) throws UserException {
...
+            Column column = targetTable.getFullSchema().get(mvColumnIdx);
+            if (column.isNameWithPrefix(CreateMaterializedViewStmt.MATERIALIZED_VIEW_NAME_PRFIX)) {
+                List<Expr> slots = new ArrayList<>();
+                column.getDefineExpr().collect(SlotRef.class, slots);
+                //We only support one children of define expr in materialized view column
+                Preconditions.checkArgument(slots.size() == 1);
+                String origName = ((SlotRef) slots.get(0)).getColumnName();
+                for (int originColumnIdx = 0; originColumnIdx < targetColumns.size(); originColumnIdx++) {
+                    if (targetColumns.get(originColumnIdx).nameEquals(origName, false)) {
+                        origColIdx2MvColIdx.put(mvColumnIdx, originColumnIdx);
```
EmmyMiao87, line 468: mvColumnIdxToOrigColumnIdx?

fe/src/main/java/org/apache/doris/analysis/InsertStmt.java
```diff
@@ -519,17 +556,23 @@ private void analyzeSubquery(Analyzer analyzer) throws UserException {
         }
 
         // expand baseTblResultExprs and colLabels in QueryStmt
-        if (!origColIdxsForShadowCols.isEmpty()) {
+        if (!origColIdxsForShadowCols.isEmpty() || !origColIdx2MvColIdx.isEmpty()) {
             if (queryStmt.getResultExprs().size() != queryStmt.getBaseTblResultExprs().size()) {
                 for (Integer idx : origColIdxsForShadowCols) {
                     queryStmt.getBaseTblResultExprs().add(queryStmt.getBaseTblResultExprs().get(idx));
                 }
+                for (Integer idx : origColIdx2MvColIdx.keySet()) {
```
EmmyMiao87, line 564: The key of origColIdx2MvColIdx is mvColIdx. The mvColumn has been supplied at the end of result exprs in query stmt. So I think it is not match.

EmmyMiao87 (changes requested):

EmmyMiao87:
Is it possible to unify the columns of schema change and materialized view into one logical process in insert stmt?

fe/src/main/java/org/apache/doris/analysis/InsertStmt.java
```diff
@@ -453,6 +453,26 @@ private void analyzeSubquery(Analyzer analyzer) throws UserException {
             }
         }
 
+        //Get the correspondence of this column to the original column through defineExpr
```
EmmyMiao87, line 456: Please add example of mv column and schema change column

fe/src/main/java/org/apache/doris/load/Load.java
HangyuanLiu: > Are you sure that this is correct`column(a, tmp_c) set(b=f(c), c=b)`？

Fix done

morningman (dismissed):
LGTM

fe/src/main/java/org/apache/doris/analysis/InsertStmt.java
```diff
@@ -521,22 +530,22 @@ private void analyzeSubquery(Analyzer analyzer) throws UserException {
...
                     ExprSubstitutionMap smap = new ExprSubstitutionMap();
-                    smap.getLhs().add(slots.get(0));
-                    smap.getRhs().add(queryStmt.getResultExprs().get(value));
+                    smap.getLhs().add(mvColumn.getRefColumn());
+                    smap.getRhs().add(queryStmt.getResultExprs().get(origColIdx));
 
-                    queryStmt.getResultExprs().add(Expr.substituteList(Lists.newArrayList(expr), smap, analyzer, false).get(0));
+                    Expr e = Expr.substituteList(Lists.newArrayList(mvColumn.getDefineExpr()), smap, analyzer, false).get(0);
+                    queryStmt.getResultExprs().add(e);
+                    queryStmt.getBaseTblResultExprs().add(e);
```
EmmyMiao87, line 548: What's different between `resultExpr` and `baseTblResultExpr`

EmmyMiao87 (changes requested):

EmmyMiao87 (approved):
LGTM

PR #3681, opened 2020-05-25
[AuditPlugin] Modify load label of audit plugin to avoid load confliction

### LINKED ISSUE #3680: [AuditPlugin] Conflict load label from different FE
**Describe the bug**
the audit plugin is running on different FE, but currently the format of load label is all same, like：

`audit_20200425_033418`

This will cause label conflict, and you will see warn log in some be.INFO:

```
W0525 15:34:18.306124 54334 stream_load_executor.cpp:147] begin transaction failed, errmsg=errCode = 2, detailMessage = Label [audit_20200425_033418] has already been used.id=4c42099eea72f17b-faf98f349594f896, job_id=-1, txn_id=-1, label=audit_20200425_033418
```

morningman:
Fix #3680 

Change the load label of audit plugin as:

`audit_yyyyMMdd_HHmmss_feIdentity`.

The `feIdentity` is got from the FE which run this plugin, currently just use FE's IP.

### DIFF
--- a/fe/src/main/java/org/apache/doris/plugin/PluginContext.java
+++ b/fe/src/main/java/org/apache/doris/plugin/PluginContext.java
@@ -17,6 +17,7 @@
 
 package org.apache.doris.plugin;
 
+import org.apache.doris.common.Config;
 import org.apache.doris.service.FrontendOptions;
 
 public class PluginContext {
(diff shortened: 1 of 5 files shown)


fe/src/main/java/org/apache/doris/plugin/PluginContext.java
```diff
@@ -17,10 +17,19 @@
...
 
+import org.apache.doris.service.FrontendOptions;
+
 public class PluginContext {
 
     // the dir path where the plugin's files saved
     private String pluginPath;
+    // the identity of the FE which load this plugin. Some plugins need to know which FE is running with it.
+    // currently, we just use FE ip
+    private String feIdentity;
```
kangkaisen, line 28: How to handle one machine，multi Doris FE instances. Do we need to add FE port to feIdentity?

morningman: Currently we do not support multi FE with same IP...
But add a port is not a big deal. I will add it.

kangkaisen (approved):
+1， LGTM

PR #3690, opened 2020-05-26
[compaction] Update cumulative point calculate algorithm

acelyc111:
Ref https://github.com/apache/incubator-doris/issues/3687
Current cumulative point calculate algorithm may skip singleton version
rowset when the rowset has only one segment and with NONOVERLAPPING flag.
When a tablet is new created and cumulate many singleton version rowsets,
cumulative point will be calculated as the max version + 1, and then
cumulative compaction couldn't pick any rowsets and compaction failed, and
will lead the next base compaction on this tablet with all rowsets, which
can also cause memory consume problem, suppose there are thousands of rowsets.

All singleton version rowsets must be newly wrote by delta writer and hasn't
do any compaction, we should place cumulative point before any of these rowsets.

### DIFF
--- a/be/src/olap/tablet.cpp
+++ b/be/src/olap/tablet.cpp
@@ -68,7 +68,7 @@ Tablet::Tablet(TabletMetaSharedPtr tablet_meta, DataDir* data_dir) :
         _last_base_compaction_failure_millis(0),
         _last_cumu_compaction_success_millis(0),
         _last_base_compaction_success_millis(0),
-        _cumulative_point(tablet_meta->cumulative_layer_point()) {
+        _cumulative_point(kInvalidCumulativePoint) {
     _gen_tablet_path();
     _rs_graph.construct_rowset_graph(_tablet_meta->all_rs_metas());
 }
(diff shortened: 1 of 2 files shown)


be/src/olap/tablet.cpp
```diff
@@ -68,7 +68,7 @@ Tablet::Tablet(TabletMetaSharedPtr tablet_meta, DataDir* data_dir) :
         _last_base_compaction_failure_millis(0),
         _last_cumu_compaction_success_millis(0),
         _last_base_compaction_success_millis(0),
-        _cumulative_point(kInvalidCumulativePoint) {
+        _cumulative_point(tablet_meta->cumulative_layer_point()) {
```
chaoyli, line 71: cumulative point in tablet meta has non-sense. 
If will cause another persistence for clone/compaction.
So it's removed from tablet meta when refactoring BE.
You can see it because of compatibility for the old code path.

chaoyli (changes requested):

be/src/olap/tablet.cpp
acelyc111: Reverted these changes.

chaoyli (approved):

PR #3691, opened 2020-05-26
[refactor] Simple refactor on class Reader

acelyc111:
This is a simple refactor patch on class Reader without any functional changes.
Main refactor points:
- Remove some useless return value
- Use range loop
- Use empty() instead of size() for some STL containers size judgement
- Use in-class initialization instead of initialize in constructor function
- Some other small refactor

### DIFF
--- a/be/src/olap/reader.cpp
+++ b/be/src/olap/reader.cpp
@@ -43,13 +43,12 @@ class CollectIterator {
 public:
     ~CollectIterator();
 
-    // Hold reader point to get reader params,
-    // set reverse to true if need read in reverse order.
-    OLAPStatus init(Reader* reader);
+    // Hold reader point to get reader params
+    void init(Reader* reader);
 
     OLAPStatus add_child(RowsetReaderSharedPtr rs_reader);
 
-    // Get top row of the heap, NULL if reach end.
+    // Get top row of the heap, nullptr if reach end.
     const RowCursor* current_row(bool* delete_flag) const {
         if (_cur_child != nullptr) {
             return _cur_child->current_row(delete_flag);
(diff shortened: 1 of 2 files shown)


chaoyli (dismissed):

chaoyli:
LGTM

be/src/olap/reader.cpp
```diff
@@ -532,33 +521,31 @@ OLAPStatus Reader::_capture_rs_readers(const ReaderParams& read_params) {
...
+    reader_context.reader_type = read_params.reader_type;
+    reader_context.tablet_schema = &_tablet->tablet_schema();
+    reader_context.need_ordered_result = need_ordered_result;
+    reader_context.return_columns = &_return_columns;
+    reader_context.seek_columns = &_seek_columns;
+    reader_context.load_bf_columns = &_load_bf_columns;
+    reader_context.conditions = &_conditions;
+    reader_context.predicates = &_col_predicates;
+    reader_context.lower_bound_keys = &_keys_param.start_keys;
+    reader_context.is_lower_keys_included = &is_lower_keys_included;
```
morningman, line 534: Is it safe to refer a local variable `is_lower_keys_included`?
Will the local `is_lower_keys_included` be deconstructed after this method?

acelyc111: Oh, it's not safe, I've fixed it now.

morningman (approved):
This CL LGTM.

But I have reservations about changing some functions that originally returned `Status` to `Void`.
Although these functions currently have only one return value. But in terms of the properties of the function itself, some functions may return errors. Such as various `init` functions.

So I suggest that if there are other similar places in the future, you can consider modifying them as appropriate according to the meaning of the function.

acelyc111:
> This CL LGTM.
> 
> But I have reservations about changing some functions that originally returned `Status` to `Void`.
> Although these functions currently have only one return value. But in terms of the properties of the function itself, some functions may return errors. Such as various `init` functions.
> 
> So I suggest that if there are other similar places in the future, you can consider modifying them as appropriate according to the meaning of the function.

OK. Agree with you that we should consider some functions may return various values, but I think if it currently always return a certain value, i.e. success status, we would better to use `void` to replace it. non-void return value may increase code and logic complexity, and make developers confused.

PR #3694, opened 2020-05-26
[Enhancement] Improve the performance of query with IN predicate 

### LINKED ISSUE #3693: [Enhancement] Improve the performance of query with IN predicate
**Describe the bug**
In the test, we found that a simple single-table aggregate query, even in the case of hit rollup, still can not achieve a high QPS.

Table: `id INT, name varchar(32), v1 INT SUM`

Query: `SELECT SUM(v1) FROM tbl WHERE id IN (1,2,3,4,5,6, ... (more than 100))`;

The phenomenon is that the CPU util is 100%, and the delay of a single query is significantly increased.

**Why**
I found that Doris use BE config `doris_max_scan_key_num` to limit both the number of scan keys and the number of conditions of a single column that can be pushed down to the storage engine.

In above example, when the elements of IN predicate is larger than `doris_max_scan_key_num`,
no conditions will be pushed to storage engine for column `id`, So that none of index of column `id` will be used. This causes the storage engine to read all the rows and need to filter at the query layer, thus making CPU util high.

**How to solve**
The BE config `doris_max_scan_key_num` should only be used to limit the number of scan keys. And we need another config to limit the number of conditions of a single column that can be pushed down to the storage engine.

**Solution**
I add a new BE config `max_pushdown_conditions_per_column`, default is 1024.

The test case is as follows:

1. Env: 3 BE and 1 FE, Intel(R) Xeon(R) Gold 5117 CPU @ 2.00GHz, 56 core
2. Table:
```
+-----------------------------+--------------+------+-------+------------+-------+
| Field                       | Type         | Null | Key   | Default    | Extra |
+-----------------------------+--------------+------+-------+------------+-------+
| date_partition              | DATE         | No   | true  | 1900-01-01 |       |
| user_name                   | VARCHAR(256) | No   | true  |            |       |
| id                          | INT          | No   | true  | N/A        |       |
...
...                    
| commit                      | INT          | No   | false | 0          | SUM   |
+-----------------------------+--------------+------+-------+------------+-------+
25 rows in set (0.01 sec)

about 3674904 rows
```
3. SQL:
```
SELECT SUM(`commit`)  FROM tbl WHERE date_partition >= '2019-11-11' AND date_partition <= '2020-05-11' AND id IN (16730, 74818, 2575, 63428, 15571, 74345, 55241, 52375, 18464, 36687, 22306, 49033, 37124, 14697, 75679, 72319, 50219, 74723, 50778, 37782, 22152, 3573, 1567, 65673, 4366, 52784, 46973, 32716, 37056, 75585, 18418, 66659, 72058, 5423, 78679, 10656, 65591, 64949, 26560, 2709, 60128, 64406, 52200, 41163, 60266, 21974, 79801, 32669, 25025, 6374, 56828, 80599, 13163, 30660, 64337, 2134, 75150, 2985, 66069, 19964, 6054, 24464, 36888, 22353, 37502, 42207, 6411, 53184, 45586, 6169, 39933, 40313, 69984, 52314, 34314, 29295, 43091, 51925, 39434, 839, 27399, 25053, 12930, 39808, 68532, 77795, 33113, 34423, 28728, 38296, 53803, 55817, 63150, 3177, 53253, 27514, 24087, 1097, 21326, 33732, 63689, 52950, 1160, 68816, 76697, 55888, 26789, 42823, 63517, 67566, 38109, 41345, 39837, 78576, 32666, 37680, 78646, 55641, 5221, 80058, 65142, 40914, 72088, 55113, 1184, 45254, 54959, 37503, 1881, 66861, 29902, 2017, 34485, 74905, 29711, 3691, 2508, 66662, 23710, 58297, 42831, 80003, 38781, 13246, 21610, 4402, 52361, 47242, 73721, 17362, 51488, 239, 61924, 38887, 29551, 11765, 15387);
```
4. Test tool: Apache jmeter, 50 threads.

5. Result:

Origin Doris version:

251.5/s Avg:   198ms Min:    68ms Max:   805ms, CPU util: 92%

New Doris version:

604.5/s Avg:    82ms Min:    30ms Max:   369ms, CPU util: 65%

New Doris version with bitmap index on column `id`:

734.9/s Avg:    67ms Min:    18ms Max:   331ms, CPU util: 50%

image

The improve is significant.

I aslo add 2 new session variable `max_scan_key_num` and `doris_max_scan_key_num` which can overwrite the 2 configs in BE config. Because there is no CBO in Doris now, we need provide a method to let user change these 2 config to meet different query scenarios.

morningman:
Fix: #3693

This CL mainly changes:
1. Add a new BE config `max_pushdown_conditions_per_column` to limit the number of conditions of a single column that can be pushed down to storage engine.

2. Add 2 new session variables `max_scan_key_num` and `doris_max_scan_key_num` which can set in session level and overwrite the config value in BE.

### DIFF
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -72,27 +72,23 @@ Status OlapScanNode::init(const TPlanNode& tnode, RuntimeState* state) {
     RETURN_IF_ERROR(ExecNode::init(tnode, state));
     _direct_conjunct_size = _conjunct_ctxs.size();
 
+    const TQueryOptions& query_options = state->query_options();
+    if (query_options.__isset.max_scan_key_num) {
+        _max_scan_key_num = query_options.max_scan_key_num;
+    } else {
+        _max_scan_key_num = config::doris_max_scan_key_num;
+    }
+
+    if (query_options.__isset.max_pushdown_in_pred_element_num) {
+        _max_pushdown_in_pred_element_num = query_options.max_pushdown_in_pred_element_num;
+    } else {
+        _max_pushdown_in_pred_element_num = config::max_pushdown_in_pred_element_num;
+    }
+
     return Status::OK();
 }
 
 void OlapScanNode::_init_counter(RuntimeState* state) {
-#if 0
-    ADD_TIMER(profile, "GetTabletTime");
-    ADD_TIMER(profile, "InitReaderTime");
-    ADD_TIMER(profile, "ShowHintsTime");
-    ADD_TIMER(profile, "BlockLoadTime");
-    ADD_TIMER(profile, "IndexLoadTime");
-    ADD_TIMER(profile, "VectorPredicateEvalTime");
-    ADD_TIMER(profile, "ScannerTimer");
-    ADD_TIMER(profile, "IOTimer");
-    ADD_TIMER(profile, "DecompressorTimer");
-    ADD_TIMER(profile, "RLETimer");
-
-    ADD_COUNTER(profile, "RawRowsRead", TUnit::UNIT);
-    ADD_COUNTER(profile, "IndexStreamCacheMiss", TUnit::UNIT);
-    ADD_COUNTER(profile, "IndexStreamCacheHit", TUnit::UNIT);
-    ADD_COUNTER(profile, "BlockLoadCount", TUnit::UNIT);
-#endif
     ADD_TIMER(_runtime_profile, "ShowHintsTime");
 
     _reader_init_timer = ADD_TIMER(_runtime_profile, "ReaderInitTime");
@@ -134,6 +130,8 @@ void OlapScanNode::_init_counter(RuntimeState* state) {
 
     _bitmap_index_filter_counter = ADD_COUNTER(_runtime_profile, "BitmapIndexFilterCount", TUnit::UNIT);
     _bitmap_index_filter_timer = ADD_TIMER(_runtime_profile, "BitmapIndexFilterTimer");
+
+    _num_scanners = ADD_COUNTER(_runtime_profile, "NumScanners", TUnit::UNIT);
 }
 
 Status OlapScanNode::prepare(RuntimeState* state) {
@@ -545,7 +543,7 @@ Status OlapScanNode::build_scan_key() {
             break;
         }
 
-        ExtendScanKeyVisitor visitor(_scan_keys);
+        ExtendScanKeyVisitor visitor(_scan_keys, _max_scan_key_num);
         RETURN_IF_ERROR(boost::apply_visitor(visitor, column_range_iter->second));
     }
 
@@ -694,10 +692,11 @@ Status OlapScanNode::start_scan_thread(RuntimeState* state) {
                 state, this, _olap_scan_node.is_preaggregation, _need_agg_finalize, *scan_range, scanner_ranges);
             _scanner_pool->add(scanner);
             _olap_scanners.push_back(scanner);
-	    disk_set.insert(scanner->scan_disk());
+	        disk_set.insert(scanner->scan_disk());
         }
     }
     COUNTER_SET(_num_disks_accessed_counter, static_cast<int64_t>(disk_set.size()));
+    COUNTER_SET(_num_scanners, static_cast<int64_t>(_olap_scanners.size()));
 
     // init progress
     std::stringstream ss;
@@ -715,7 +714,7 @@ Status OlapScanNode::start_scan_thread(RuntimeState* state) {
 template<class T>
 Status OlapScanNode::normalize_predicate(ColumnValueRange<T>& range, SlotDescriptor* slot) {
     // 1. Normalize InPredicate, add to ColumnValueRange
-    RETURN_IF_ERROR(normalize_in_predicate(slot, &range));
+    RETURN_IF_ERROR(normalize_in_and_eq_predicate(slot, &range));
 
     // 2. Normalize BinaryPredicate , add to ColumnValueRange
     RETURN_IF_ERROR(normalize_binary_predicate(slot, &range));
@@ -737,7 +736,7 @@ static bool ignore_cast(SlotDescriptor* slot, Expr* expr) {
 }
 
 template<class T>
-Status OlapScanNode::normalize_in_predicate(SlotDescriptor* slot, ColumnValueRange<T>* range) {
+Status OlapScanNode::normalize_in_and_eq_predicate(SlotDescriptor* slot, ColumnValueRange<T>* range) {
     for (int conj_idx = 0; conj_idx < _conjunct_ctxs.size(); ++conj_idx) {
         // 1. Normalize in conjuncts like 'where col in (v1, v2, v3)'
         if (TExprOpcode::FILTER_IN == _conjunct_ctxs[conj_idx]->root()->op()) {
@@ -756,9 +755,6 @@ Status OlapScanNode::normalize_in_predicate(SlotDescriptor* slot, ColumnValueRan
                 }
             }
 
-            if (pred->is_not_in()) {
-                continue;
-            }
             std::vector<SlotId> slot_ids;
 
             if (1 == pred->get_child(0)->get_slot_ids(&slot_ids)) {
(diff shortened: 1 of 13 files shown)


be/src/exec/olap_scan_node.cpp
```diff
@@ -884,25 +910,54 @@ Status OlapScanNode::normalize_in_predicate(SlotDescriptor* slot, ColumnValueRan
...
+            } // end for each binary predicate child
+        } // end of handling eq binary predicate
+
+        if (range->get_fixed_value_size() > 0) {
+            // this columns already meet some eq predicates(IN or Binary),
+            // There is no need to continue to iterate.
+            // TODO(cmy): In fact, this part of the judgment should be completed in
+            // the FE query planning stage. For the following predicate conditions,
+            // it should be possible to eliminate at the FE side.
+            //      WHERE A = 1 and A in (2,3,4)
```
kangkaisen, line 940: I think we should do this work directly in FE, Not add a todo here. In FE we could directly get a empty set for this query.

be/src/exec/olap_scan_node.cpp
```diff
@@ -715,7 +714,7 @@ Status OlapScanNode::start_scan_thread(RuntimeState* state) {
 template<class T>
 Status OlapScanNode::normalize_predicate(ColumnValueRange<T>& range, SlotDescriptor* slot) {
     // 1. Normalize InPredicate, add to ColumnValueRange
-    RETURN_IF_ERROR(normalize_in_predicate(slot, &range));
+    RETURN_IF_ERROR(normalize_in_and_eq_predicate(slot, &range));
 
     // 2. Normalize BinaryPredicate , add to ColumnValueRange
     RETURN_IF_ERROR(normalize_binary_predicate(slot, &range));
```
kangkaisen, line 720: eq binary is also binary. So I think we would better also change the `normalize_binary_predicate` name.

yangzhg (commented):
LGTM

be/src/exec/olap_scan_node.cpp
morningman: I will try to solve this in next PR.

be/src/exec/olap_scan_node.cpp
morningman: OK

be/src/exec/olap_scan_node.cpp
```diff
@@ -736,100 +735,114 @@ static bool ignore_cast(SlotDescriptor* slot, Expr* expr) {
...
             if (Expr::type_without_cast(pred->get_child(0)) != TExprNodeType::SLOT_REF) {
+                // not a slot ref(column)
+                continue;
+            }
+
+            std::vector<SlotId> slot_ids;
+            if (pred->get_child(0)->get_slot_ids(&slot_ids) != 1) {
+                // not a single column predicate
+                continue;
+            }
```
wutiangan, line 763: 754 line can guarantee it's a single coulumn predicate?

be/src/exec/olap_scan_node.cpp
```diff
@@ -838,39 +851,52 @@ Status OlapScanNode::normalize_in_predicate(SlotDescriptor* slot, ColumnValueRan
...
+                }
 
-                    switch (slot->type().type) {
+                // begin to push condition value into ColumnValueRange
+                // clear the ColumnValueRange before adding new fixed values.
+                // because for AND compound predicates, it can overwrite previous conditions
+                switch (slot->type().type) {
                     case TYPE_TINYINT: {
                         int32_t v = *reinterpret_cast<int8_t*>(value);
+                        range->clear();
```
wutiangan, line 891: Whether ”range->clear()“ can be preceded by the switch statement

be/src/exec/olap_scan_node.cpp
kangkaisen: OK

WingsGo:
Hi, @morningman I have some question about olap_scan_node, could you plz help me figure it out?
1. what's the relationship between scanner and tablet, one tablet can be scanned by many scanner?they are n:1 or 1:1?
2. what's the relationship between TPaloScanRange and ColumnValueRange, is ColumnValueRange to deal with predicate case and TPaloScanRange can indicate that the range a scanner need to scan from StorageEngine's prefix index?
3. What's the usage of function `extend_scan_key` and the config variable `doris_scanner_row_num`?
I would appreciate your reply, thanks~

morningman:
> Hi, @morningman I have some question about olap_scan_node, could you plz help me figure it out?
> 
> 1. what's the relationship between scanner and tablet, one tablet can be scanned by many scanner?they are n:1 or 1:1?
> 2. what's the relationship between TPaloScanRange and ColumnValueRange, is ColumnValueRange to deal with predicate case and TPaloScanRange can indicate that the range a scanner need to scan from StorageEngine's prefix index?
> 3. What's the usage of function `extend_scan_key` and the config variable `doris_scanner_row_num`?
>    I would appreciate your reply, thanks~

1. scanner and tablet is `n:m` relation
2. I am not familiar with `TPaloScanRange`, need a further look...
3. `extend_scan_key` is used to extend the scan keys.  `scan keys` are used to determine the number of scanners.

 For example, `WHERE a in (1,2,3) and b > 5` and `a`, `b` are key columns of the table.

First, for `a in (1,2,3)`, the `scan keys` become: `(a=1),(a=2),(a=3)`
Second, for `b > 5`, the `scan keys` extends to `(a=1, b>5), (a=2, b>5), (a=3, b>5)`

4. `doris_scanner_row_num` is used for priority scheduling of scanner. Doris has a scanner thread pool that is shared by all scanners. Therefore, a scheduling strategy is needed to prevent a scanner from occupying threads for a long time. `doris_scanner_row_num` can control the maximum number of rows a scanner can read in one scheduling. The details of the algorithm that are not specific need to be understood by reading the code.

@chaoyli may understand this better. If I said something wrong, he can help correct it.

be/src/exec/olap_scan_node.cpp
morningman: OK

be/src/exec/olap_scan_node.cpp
morningman: I am not sure about that, looks like line `754` can guarantee.
But just leave this judgement, because no matter what, we still have to call `get_slot_ids()` to get the slot id and check if it equals to the `slot->id()` later.

WingsGo:
> > Hi, @morningman I have some question about olap_scan_node, could you plz help me figure it out?
> > 
> > 1. what's the relationship between scanner and tablet, one tablet can be scanned by many scanner?they are n:1 or 1:1?
> > 2. what's the relationship between TPaloScanRange and ColumnValueRange, is ColumnValueRange to deal with predicate case and TPaloScanRange can indicate that the range a scanner need to scan from StorageEngine's prefix index?
> > 3. What's the usage of function `extend_scan_key` and the config variable `doris_scanner_row_num`?
> >    I would appreciate your reply, thanks~
> 
> 1. scanner and tablet is `n:m` relation
> 2. I am not familiar with `TPaloScanRange`, need a further look...
> 3. `extend_scan_key` is used to extend the scan keys.  `scan keys` are used to determine the number of scanners.
> 
> For example, `WHERE a in (1,2,3) and b > 5` and `a`, `b` are key columns of the table.
> 
> First, for `a in (1,2,3)`, the `scan keys` become: `(a=1),(a=2),(a=3)`
> Second, for `b > 5`, the `scan keys` extends to `(a=1, b>5), (a=2, b>5), (a=3, b>5)`
> 
> 1. `doris_scanner_row_num` is used for priority scheduling of scanner. Doris has a scanner thread pool that is shared by all scanners. Therefore, a scheduling strategy is needed to prevent a scanner from occupying threads for a long time. `doris_scanner_row_num` can control the maximum number of rows a scanner can read in one scheduling. The details of the algorithm that are not specific need to be understood by reading the code.
> 
> @chaoyli may understand this better. If I said something wrong, he can help correct it.

Thanks for your reply, so if the `scan keys` extends to `(a=1, b>5), (a=2, b>5), (a=3, b>5)` , doris will split into 3 scanners to scan RowBatch which satisified the predicate? Do I understand correctly?

morningman:
> Thanks for your reply, so if the `scan keys` extends to `(a=1, b>5), (a=2, b>5), (a=3, b>5)` , doris will split into 3 scanners to scan RowBatch which satisified the predicate? Do I understand correctly?

Not exactly, `extend_scan_key()` only extend all possible scan keys. the number of scanners is got from `get_hints()`

EmmyMiao87 (dismissed):
LGTM

EmmyMiao87 (approved):

PR #3699, opened 2020-05-27
[Bug] Duplicate Key table core when predicate on metric column

### LINKED ISSUE #3700: [Bug] Duplicate Key table core when predicate on metric column.
```
CREATE TABLE `query_detail` (
  `query_id` varchar(100) NULL COMMENT "",
  `start_time` datetime NULL COMMENT "",
  `end_time` datetime NULL COMMENT "",
  `latency` int(11) NULL COMMENT "unit is milliseconds",
  `state` varchar(20) NULL COMMENT "RUNNING/FINISHED/FAILED",
  `sql` varchar(1024) NULL COMMENT ""
)
DUPLICATE KEY(`query_id`)

SELECT COUNT(*) FROM query_detail WHERE start_time >= '2020-05-27 14:52:16' AND start_time < '2020-05-27 14:52:31';
```
The above query will core because of ZoneMap only in query_id.
Use start_time to match ZoneMap cause this core.

chaoyli:
```
CREATE TABLE `query_detail` (
  `query_id` varchar(100) NULL COMMENT "",
  `start_time` datetime NULL COMMENT "",
  `end_time` datetime NULL COMMENT "",
  `latency` int(11) NULL COMMENT "unit is milliseconds",
  `state` varchar(20) NULL COMMENT "RUNNING/FINISHED/FAILED",
  `sql` varchar(1024) NULL COMMENT ""
)
DUPLICATE KEY(`query_id`)

SELECT COUNT(*) FROM query_detail WHERE start_time >= '2020-05-27 14:52:16' AND start_time < '2020-05-27 14:52:31';
```
The above query will core because of ZoneMap only in query_id.
Use start_time to match ZoneMap cause this core.

### DIFF
--- a/be/src/olap/olap_cond.cpp
+++ b/be/src/olap/olap_cond.cpp
@@ -649,16 +649,14 @@ bool Conditions::delete_conditions_eval(const RowCursor& row) const {
 }
 
 bool Conditions::rowset_pruning_filter(const std::vector<KeyRange>& zone_maps) const {
-    //通过所有列上的删除条件对version进行过滤
+    // ZoneMap will store min/max of rowset.
+    // The function is to filter rowset using ZoneMaps
+    // and query predicates.
     for (auto& cond_it : _columns) {
-        if (_cond_column_is_key_or_duplicate(cond_it.second) && cond_it.first > zone_maps.size()) {
-            LOG(WARNING) << "where condition not equal zone maps size. "
-                         << "cond_id=" << cond_it.first
-                         << ", zone_map_size=" << zone_maps.size();
-            return false;
-        }
-        if (_cond_column_is_key_or_duplicate(cond_it.second) && !cond_it.second->eval(zone_maps.at(cond_it.first))) {
-            return true;
+        if (_cond_column_is_key_or_duplicate(cond_it.second)) {
+            if (cond_it.first < zone_maps.size() && !cond_it.second->eval(zone_maps.at(cond_it.first))) {
+                return true;
+            }
         }
     }
     return false;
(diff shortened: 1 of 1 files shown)


be/src/olap/olap_cond.cpp
```diff
@@ -651,14 +651,10 @@ bool Conditions::delete_conditions_eval(const RowCursor& row) const {
...
-        if (_cond_column_is_key_or_duplicate(cond_it.second) && cond_it.first > zone_maps.size()) {
-            LOG(WARNING) << "where condition not equal zone maps size. "
-                         << "cond_id=" << cond_it.first
-                         << ", zone_map_size=" << zone_maps.size();
-            return false;
-        }
-        if (_cond_column_is_key_or_duplicate(cond_it.second) && !cond_it.second->eval(zone_maps.at(cond_it.first))) {
-            return true;
+        if (_cond_column_is_key_or_duplicate(cond_it.second)) {
+            if (cond_it.first < zone_maps.size() && !cond_it.second->eval(zone_maps.at(cond_it.first))) {
```
yangzhg, line 655: it cores just because of `zone_maps.size() == cond_it.first`  and `zone_maps.at(cond_it.first)` returns null ?

be/src/olap/olap_cond.cpp
```diff
@@ -651,14 +651,10 @@ bool Conditions::delete_conditions_eval(const RowCursor& row) const {
 bool Conditions::rowset_pruning_filter(const std::vector<KeyRange>& zone_maps) const {
     //通过所有列上的删除条件对version进行过滤
```
EmmyMiao87, line 652: Could you please transform Cninese to English?

be/src/olap/olap_cond.cpp
chaoyli: Yes, because zone_maps has only one item. zone_maps.at(1) will core dump.

morningman (approved):
LGTM

PR #3702, opened 2020-05-27
FixBug:  Load data Is error in json load

worker24h:
The data  is always 1 by json load
https://github.com/apache/incubator-doris/issues/3697

### DIFF
--- a/be/src/exprs/json_functions.h
+++ b/be/src/exprs/json_functions.h
@@ -72,8 +72,9 @@ public:
 
     /**
      * The `document` parameter must be has parsed.
+     * return Value Is Array object
      */
-    static rapidjson::Value* get_json_object_from_parsed_json(
+    static rapidjson::Value* get_json_array_from_parsed_json(
             const std::string& path_string,
             rapidjson::Value* document,
             rapidjson::Document::AllocatorType& mem_allocator);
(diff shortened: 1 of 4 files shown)


be/src/exprs/json_functions.h
```diff
@@ -72,6 +72,7 @@ class JsonFunctions {
 
     /**
      * The `document` parameter must be has parsed.
+     * return Value Is Array object
```
morningman, line 75: better to change its name to `get_json_array_from_parsed_json()` to make it more explict.

And in `src//exec/json_scanner.cpp` line 270, you can this method, but judge the return value as either array or non-array? the `else` should not happen?

morningman (changes requested):

be/src/exprs/json_functions.h
worker24h: The method MUST return ArrayObject.
 `else if (!root->IsArray())` is true, then I warp a ArrayObject  in `root` outside.
`else` case  MUST be ArrayObject

worker24h: The method name changed to `get_json_objects_from_parsed_json` ??

morningman (approved):
LGTM

Issue #3696, opened 2020-05-27
[Proposal] Change some default config in be to make doris more friendly to new user and make brpc_num_threads configurable to get better performance for qps


caiconghui:
Now, there are many rpcs between be and fe, and some default config in be may cause poor performance for query or stream load. In our production environment, we find that when we execute stream load in doris, the txn can be easily publish timeout and query can be easily rpc timeout, http request to be can be easily no response. 

And we read the code for server in be, and we find the num of thrift server threads is too small to handle so many rpc which include exec_plan_fragment and submit_tasks at the same time,  and http server use reactor model to process request, when the stream load performance is poor (server thread would be blocked in rpc or  other blocked function) and the num of threads in http server is small, it will cause http server can not handle http request in time.     

Finally, we do some experiments after change the some config in be, and we find that query and stream load both get good performance and http server can response in time.

we use thread pool in java client to make 10000 stream load requests to fe with 5 be instance. Every request has one row data, and we have 500 tables in on db, every table will be inserted 20 rows. the total time cost for java client is  35398ms, and some stream load results are as following:
{
    "TxnId": 276003,
    "Label": "86292a00-511b-47de-82e2-6dfa9c2d219d",
    "Status": "Success",
    "Message": "OK",
    "NumberTotalRows": 1,
    "NumberLoadedRows": 1,
    "NumberFilteredRows": 0,
    "NumberUnselectedRows": 0,
    "LoadBytes": 12,
    "LoadTimeMs": 31,
    "BeginTxnMs": 0,
    "ReadBufferMs": 0,
    "WriteBufferMs": 12,
    "StreamLoadPutMs": 0,
    "CommitAndPublishMs": 15
}
{
    "TxnId": 276006,
    "Label": "7d6023a4-5fb5-4551-9327-4c758f231851",
    "Status": "Success",
    "Message": "OK",
    "NumberTotalRows": 1,
    "NumberLoadedRows": 1,
    "NumberFilteredRows": 0,
    "NumberUnselectedRows": 0,
    "LoadBytes": 12,
    "LoadTimeMs": 31,
    "BeginTxnMs": 0,
    "ReadBufferMs": 0,
    "WriteBufferMs": 12,
    "StreamLoadPutMs": 0,
    "CommitAndPublishMs": 15
}
{
    "TxnId": 275256,
    "Label": "3b29cd47-2aa8-4255-9d37-df6148570e93",
    "Status": "Success",
    "Message": "OK",
[...]

PR #3703, opened 2020-05-27
[Load] Add more metric to trace the time cost in stream load and make brpc_num_threads configurable

caiconghui:
This PR mainly do three things:
1. Make brpc_num_threads configurable to get better performance for qps
2. Change some default config in be to make doris get better performance both in stream load and query
3. Add more metric to trace the time cost in stream load

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -30,6 +30,9 @@ namespace config {
     // port for brpc
     CONF_Int32(brpc_port, "8060");
 
+    //the num of bthreads for brpc, the default value is set to -1, which means the num of bthreads is #cpu-cores
+    CONF_Int32(brpc_num_threads, "-1")
+
     // Declare a selection strategy for those servers have many ips.
     // Note that there should at most one ip match this list.
     // this is a list in semicolon-delimited format, in CIDR notation, e.g. 10.10.10.0/24
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -78,7 +78,7 @@ namespace config {
     // the count of thread to high priority batch load
     CONF_Int32(push_worker_count_high_priority, "3");
     // the count of thread to publish version
-    CONF_Int32(publish_version_worker_count, "2");
+    CONF_Int32(publish_version_worker_count, "8");
     // the count of thread to clear transaction task
     CONF_Int32(clear_transaction_task_worker_count, "1");
     // the count of thread to delete
@@ -150,7 +150,7 @@ namespace config {
     // the maximum number of bytes to display on the debug webserver's log page
     CONF_Int64(web_log_bytes, "1048576");
     // number of threads available to serve backend execution requests
-    CONF_Int32(be_service_threads, "64");
+    CONF_Int32(be_service_threads, "512");
     // key=value pair of default query options for Doris, separated by ','
     CONF_String(default_query_options, "");
 
@@ -284,7 +284,7 @@ namespace config {
     // Port to start debug webserver on
     CONF_Int32(webserver_port, "8040");
     // Number of webserver workers
-    CONF_Int32(webserver_num_workers, "5");
+    CONF_Int32(webserver_num_workers, "48");
     // Period to update rate counters and sampling counters in ms.
     CONF_mInt32(periodic_counter_update_period_ms, "500");
 
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -150,7 +150,7 @@ namespace config {
     // the maximum number of bytes to display on the debug webserver's log page
     CONF_Int64(web_log_bytes, "1048576");
     // number of threads available to serve backend execution requests
-    CONF_Int32(be_service_threads, "512");
+    CONF_Int32(be_service_threads, "64");
     // key=value pair of default query options for Doris, separated by ','
     CONF_String(default_query_options, "");
 
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -30,7 +30,7 @@ namespace config {
     // port for brpc
     CONF_Int32(brpc_port, "8060");
 
-    //the num of bthreads for brpc, the default value is set to -1, which means the num of bthreads is #cpu-cores
+    // the number of bthreads for brpc, the default value is set to -1, which means the number of bthreads is #cpu-cores
     CONF_Int32(brpc_num_threads, "-1")
 
     // Declare a selection strategy for those servers have many ips.
--- a/be/src/runtime/stream_load/stream_load_context.cpp
+++ b/be/src/runtime/stream_load/stream_load_context.cpp
@@ -70,6 +70,17 @@ std::string StreamLoadContext::to_json() const {
     writer.Int64(receive_bytes);
     writer.Key("LoadTimeMs");
     writer.Int64(load_cost_nanos / 1000000);
+    writer.Key("BeginTxnTimeMs");
+    writer.Int64(begin_txn_cost_nanos / 1000000);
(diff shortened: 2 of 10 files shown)


caiconghui:
ref #3696

be/src/runtime/stream_load/stream_load_context.cpp
```diff
@@ -70,6 +70,17 @@ std::string StreamLoadContext::to_json() const {
...
+    writer.Key("BeginTxnTimeMs");
+    writer.Int64(begin_txn_cost_nanos / 1000000);
+    writer.Key("StreamLoadPutTimeMs");
+    writer.Int64(stream_load_put_cost_nanos / 1000000);
+    writer.Key("ReadDataTimeMs");
+    writer.Int64(read_data_cost_nanos / 1000000);
+    writer.Key("WriteDataTimeMs");
+    writer.Int(write_data_cost_nanos / 1000000);
+    writer.Key("CommitAndPublishTimeMs");
+    writer.Int64(commit_and_publish_txn_cost_nanos / 1000000);
```
morningman, line 82: modify the document of stream load to explain this new fields

docs/en/administrator-guide/config/be_config.md
```diff
@@ -77,6 +77,12 @@ Sometimes the query fails and an error message of `The server is overcrowded` wi
 
 Since this is a brpc configuration, users can also modify this parameter directly during operation. Modify by visiting `http://be_host:brpc_port/flags`.
 
+### brpc_num_threads
+
+This configuration is mainly used to modify the num of bthreads for brpc. The default value is set to -1, which means the num of bthreads is #cpu-cores.
+
+User can set this configuration to a larger value to get better QPS performance.
```
morningman, line 84: Better give an suggestion value or an example. eg: "In our env, we have xx BE with xx core, set this to xx, it can reach xx QPS"

morningman (changes requested):

be/src/runtime/stream_load/stream_load_context.cpp
caiconghui: done

docs/en/administrator-guide/config/be_config.md
caiconghui: Actually, I don't get the real QPS for brpc, I only test stream load operation in our enviroment with  brpc_num_threads set to be 500,  and the relationship between QPS and num of bthreads, I think https://github.com/apache/incubator-brpc/blob/master/docs/cn/benchmark.md is more convincing

be/src/common/config.h
```diff
@@ -30,6 +30,9 @@ namespace config {
     // port for brpc
     CONF_Int32(brpc_port, "8060");
 
+    //the num of bthreads for brpc, the default value is set to -1, which means the num of bthreads is #cpu-cores
```
wutiangan, line 33:
```suggestion
    //the number of bthreads for brpc, the default value is set to -1, which means the number of bthreads is #cpu-cores
```Keep consistent with other variables

Comment (caiconghui) [reply be/src/common/config.h]: fix

File: be/src/common/config.h [hunk 4]
```diff
@@ -137,7 +140,7 @@ namespace config {
     // the maximum number of bytes to display on the debug webserver's log page
     CONF_Int64(web_log_bytes, "1048576");
     // number of threads available to serve backend execution requests
-    CONF_Int32(be_service_threads, "64");
+    CONF_Int32(be_service_threads, "512");
```
imay, line 143: I think 512 is some big, why not keep it 64?
For most scenario, 64 is enough.

be/src/common/config.h
```diff
@@ -271,7 +274,7 @@ namespace config {
     // Port to start debug webserver on
     CONF_Int32(webserver_port, "8040");
     // Number of webserver workers
-    CONF_Int32(webserver_num_workers, "5");
+    CONF_Int32(webserver_num_workers, "48");
```
imay, line 277: 48 is some big, how about 12?

be/src/common/config.h
caiconghui: actually for 64, the query and stream load performance is poor, as mentioned in issue #3696, query and submit agent task share the same thread pool in thrift server. I have done some test, 512 can apparently improve the performance both for stream load and query(can not be rpc timeout), query would not by blocked by stream load.in this setting, the publish timeout almost disapper.

be/src/common/config.h
caiconghui: if http server only for normal  light http request, I think 12, or even 5 is ok. but stream load is a heavy http request， once user use stream load to insert data, more threads can reduce the risk of no reponse from http server. the time cost for switch between threads i think we can ignore when compared to time cost in stream load.   48 threads don't cost too much memory resources. Above all. The advantages outweigh the disadvantages.

be/src/common/config.h
caiconghui: sorry, I find this config make no sense when thrift server use threaded type , so I revert the modification

caiconghui:
As a conclusion, the http server is based on libevent , which use reactor model to process http request, some http requests are processed by one thread.  it assumes that every process function in libevent thread is non blocked, but there are some blocked operations in our current implementaion, such as rpc calls to fe and wait for write data operation finish. once a stream load is blocked, it will affect the following stream load, which cannot be processed in time. So This PR manily add some mertics to trace the Potential bottlenecks when stream load get bad performance. and change the default value of http server thread num to reduce the num of http requeset for per libevent server thread for one time.

kangkaisen:
@caiconghui Hi, If you only change the webserver_num_workers from 5 to 48, how many times stream load performance will improve?

caiconghui:
> @caiconghui Hi, If you only change the webserver_num_workers from 5 to 48, how many times stream load performance will improve?
@kangkaisen 
if every stream load perform well, this does not make a apparent improvement. But there always are some long tail task in our online environment, once a task is executed slowly, it will block all the http request on the same thread . So change the thread from 5 to 48 is to reduce the extra time cost for waiting slow task to complete, because we use rpc call and furture get function in reactor model thread.(The real improvement for us is that http request for metric never timeout now when we change the thread num from 5 to 42 in our online environment)
It is hard for us to write non blocking code in libevent callback function now.
Similar problems can see following:
https://stackoverflow.com/questions/42689219/does-calling-a-blocking-function-inside-libev-event-callback-function-blocks-who

kangkaisen:
> > @caiconghui Hi, If you only change the webserver_num_workers from 5 to 48, how many times stream load performance will improve?
> > @kangkaisen
> > if every stream load perform well, this does not make a apparent improvement. But there always are some long tail task in our online environment, once a task is executed slowly, it will block all the http request on the same thread . So change the thread from 5 to 48 can reduce the extra time cost for waiting slow task to complete, because we use rpc call and furture get function in reactor model thread.(The real improvement for us is that http request for metric never timeout now when we change the thread num from 5 to 42 in our online environment)
> > It is hard for us to write non blocking code in libevent callback function now.
> > Similar problems can see following:
> > https://stackoverflow.com/questions/42689219/does-calling-a-blocking-function-inside-libev-event-callback-function-blocks-who

@caiconghui OK, I see. Thanks.

kangkaisen (dismissed):
+1, LGTM

morningman (dismissed):
LGTM

morningman:
hi @caiconghui Please resolve the conflict

caiconghui:
> hi @caiconghui Please resolve the conflict

@morningman Done

chaoyli (approved):

Issue #3698, opened 2020-05-27
[Proposal]Add last tablet report time in show backends


WingsGo:
Recently, I found a cluster's BE's tablets is not balance, and the balance error msg is `path hash is not set`, I finally found that some BE cannot report it's tablets to BE because the report_all_tablet's logic is not good enough, and the problem is solved.But I think we should add a column in `show backends` to display the last report tablets time for BE, so that we can quickly found is there some BE encounter some problems.

morningman:
LGTM, maybe we can add a new column to show some info together in JSON format.
So that if we want to add more info later, there is no need to add more columns in result of `show backends`

WingsGo:
> LGTM, maybe we can add a new column to show some info together in JSON format.
> So that if we want to add more info later, there is no need to add more columns in result of `show backends`

It sounds good, but a JSON format string in the column is not easily to read than a column, we may need to display the JSON format string in a pretty way.

PR #3713, opened 2020-05-28
Use BackendStatus to show BE's infomation in `show backends;`

WingsGo:
The infomation is displayed in JSON format.For example:
` {"lastTabletReportTime":"2020-05-28 15:29:01"} `

### DIFF
--- a/fe/src/main/java/org/apache/doris/master/ReportHandler.java
+++ b/fe/src/main/java/org/apache/doris/master/ReportHandler.java
@@ -37,12 +37,14 @@ import org.apache.doris.common.Config;
 import org.apache.doris.common.MetaNotFoundException;
 import org.apache.doris.common.Pair;
 import org.apache.doris.common.util.Daemon;
+import org.apache.doris.common.util.TimeUtils;
 import org.apache.doris.metric.GaugeMetric;
 import org.apache.doris.metric.Metric.MetricUnit;
 import org.apache.doris.metric.MetricRepo;
 import org.apache.doris.persist.BackendTabletsInfo;
 import org.apache.doris.persist.ReplicaPersistInfo;
 import org.apache.doris.system.Backend;
+import org.apache.doris.system.Backend.BackendStatus;
 import org.apache.doris.system.SystemInfoService;
 import org.apache.doris.task.AgentBatchTask;
 import org.apache.doris.task.AgentTask;
@@ -246,6 +248,13 @@ public class ReportHandler extends Daemon {
         long start = System.currentTimeMillis();
         LOG.info("backend[{}] reports {} tablet(s). report version: {}",
                  backendId, backendTablets.size(), backendReportVersion);
+
+        final SystemInfoService currentSystemInfo = Catalog.getCurrentSystemInfo();
+        Backend reportBackend = currentSystemInfo.getBackend(backendId);
+        if (reportBackend != null) {
+            BackendStatus backendStatus = reportBackend.getBackendStatus();
+            backendStatus.lastTabletReportTime = TimeUtils.longToTimeString(start);
+        }
 
         // storage medium map
         HashMap<Long, TStorageMedium> storageMediumMap = Catalog.getInstance().getPartitionIdToStorageMediumMap();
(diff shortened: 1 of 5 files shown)


WingsGo:
For #3698

fe/src/main/java/org/apache/doris/system/Backend.java
```diff
@@ -94,6 +94,9 @@
     // this field is set by tablet report, and just for metric monitor, no need to persist.
     private AtomicLong tabletMaxCompactionScore = new AtomicLong(0);
 
+    // additional backendStatus information for BE, display in JSON format
+    private BackendStatus backendStatus;
```
morningman, line 98:
```suggestion
    private BackendStatus backendStatus = new BackendStatus();
```
And remove all constructors below.

fe/src/main/java/org/apache/doris/master/ReportHandler.java
```diff
@@ -247,6 +249,13 @@ private static void tabletReport(long backendId, Map<Long, TTablet> backendTable
         LOG.info("backend[{}] reports {} tablet(s). report version: {}",
                  backendId, backendTablets.size(), backendReportVersion);
 
+        final SystemInfoService currentSystemInfo = Catalog.getCurrentSystemInfo();
+        Backend reportBackend = currentSystemInfo.getBackend(backendId);
+        if (reportBackend != null) {
+            BackendStatus backendStatus = reportBackend.getBackendStatus();
+            backendStatus.lastTabletReportTime = TimeUtils.longToTimeString(start);
```
morningman, line 256: Better to create a method called `backend.updateStatus()`  to set these infos. 
Maybe we will add more info later.

morningman (changes requested):
Please update the doc `SHOW BACKENDS.md`, both in English and Chinese.

fe/src/main/java/org/apache/doris/master/ReportHandler.java
WingsGo: if we need add more info, we can just add a new Field in `BackendStatus` and set the field's value, and new info will be shown in `show backends`, why we need to set the value and update the value?I think we can just set the value, and the new value will be shown when we use `show backends`?

WingsGo:
> Please update the doc `SHOW BACKENDS.md`, both in English and Chinese.

Done

fe/src/main/java/org/apache/doris/master/ReportHandler.java
morningman: It ok, not a big deal.
Oh, I suggest to set this value right after the report in done, in case some other error happens during the report process.

morningman (changes requested):

fe/src/main/java/org/apache/doris/master/ReportHandler.java
WingsGo: The status mainly reflect BE's status, if some error happends, it should be FE's problem?Our mainly purpose is to monitoring BE's status, so maybe we can leave it to make sure that BE's status is OK?

WingsGo:
The status mainly reflect BE's status, if some error happends, it should be FE's problem?Our mainly purpose is to monitoring BE's status, so maybe we can leave it to make sure that BE's status is OK?

morningman:
> The status mainly reflect BE's status, if some error happends, it should be FE's problem?Our mainly purpose is to monitoring BE's status, so maybe we can leave it to make sure that BE's status is OK?

I remember this PR is to solve some problems that BE cannot report successfully because of the version out of data.

If there are other problems in the reporting process, it is equivalent to this BE not reporting successfully. SO I think it should also be able to be reflected in this status field. We should focus on the final result, not the success or failure of a certain part of the entire reporting process.

WingsGo:
> > The status mainly reflect BE's status, if some error happends, it should be FE's problem?Our mainly purpose is to monitoring BE's status, so maybe we can leave it to make sure that BE's status is OK?
> 
> I remember this PR is to solve some problems that BE cannot report successfully because of the version out of data.
> 
> If there are other problems in the reporting process, it is equivalent to this BE not reporting successfully. SO I think it should also be able to be reflected in this status field. We should focus on the final result, not the success or failure of a certain part of the entire reporting process.
> 
> So actually, `lastTabletReportTime` means `lastSuccessTabletReportTime`

Ok, I change the logic and rename the filed name~

morningman (approved):
LGTM

kangkaisen (approved):
+1, LGTM

Issue #3433, opened 2020-04-29
[Spark load] Doris support Spark load


wyb:
For many users who want to load data into doris for the first time, they have large amount of data, about 10G+，it is hard to support to load so large data into doris at one time using Broker load or Stream load. To resolve this problem, We proposal a new solution to load data by using spark cluster. 

Spark clusters are used to preprocess data (bitmap global dict build, partition, sort, aggregation) in spark load, which can improve Doris load performance of large data volume and save the computing resources of Doris. 

Spark load is mainly used for the initial migration from other systems or loading large amounts of data into Doris.

```
                 +
                 | 0. User create spark load job
            +----v----+
            |   FE    |---------------------------------+
            +----+----+                                 |
                 | 3. FE send push tasks                |
                 | 5. FE publish version                |
    +------------+------------+                         |
    |            |            |                         |
+---v---+    +---v---+    +---v---+                     |
|  BE   |    |  BE   |    |  BE   |                     |1. FE submit Spark ETL job
+---^---+    +---^---+    +---^---+                     |
    |4. BE push with broker   |                         |
+---+---+    +---+---+    +---+---+                     |
|Broker |    |Broker |    |Broker |                     |
+---^---+    +---^---+    +---^---+                     |
    |            |            |                         |
+---+------------+------------+---+ 2.ETL +-------------v---------------+
|               HDFS              +------->       Spark cluster         |
|                                 <-------+                             |
+---------------------------------+       +-----------------------------+

```

wyb:
design doc
[Proposal] support spark load #2855
[Proposal] Support Spark Convert Doris Segment #2887
Spark load etl interface #3010

imay:
1. can refer this issue in related PR and issue.
2. I will create a project "Spark Load" to track this feature.
3. You can create an issue for each part of this project.

wangbo:
**Count Distinct Module**
#3319 Support Java Verision HyperLogLog(REVIEWING)
#3061 Doris Support Using Hive Table to Build Global Dict(TESTING)
#3088 Support Java version 64 bits Integers for BITMAP type(MERGED)

wyb:
Add spark etl job config #3712

wyb:
Add resource manager #3418

wyb:
[Spark load] fe create job #3715

wyb:
[Spark load] Fe submit spark etl job #3716

wyb:
[Spark load] Fe process etl and loading state job #3717

[...]

PR #3717, opened 2020-05-28
[Spark load][Fe 6/6] Fe process etl and loading state job

wyb:
1. Fe checks the status of etl job regularly 
1.1 If status is RUNNING, update etl job progress
1.2 If status is CANCELLED, cancel load job
1.3 If status is FINISHED, get the etl output file paths, update job state to LOADING and log job update info

2. Fe sends PushTask to Be and commits transaction after all push tasks execute successfully

#3433

### DIFF
--- a/fe/src/main/java/org/apache/doris/load/loadv2/SparkLoadJob.java
+++ b/fe/src/main/java/org/apache/doris/load/loadv2/SparkLoadJob.java
@@ -17,29 +17,70 @@
 
 package org.apache.doris.load.loadv2;
 
+import static org.apache.doris.common.DataQualityException.QUALITY_FAIL_MSG;
+
 import org.apache.doris.analysis.BrokerDesc;
+import org.apache.doris.analysis.CastExpr;
+import org.apache.doris.analysis.DescriptorTable;
 import org.apache.doris.analysis.ResourceDesc;
+import org.apache.doris.analysis.Expr;
+import org.apache.doris.analysis.SlotDescriptor;
+import org.apache.doris.analysis.SlotRef;
+import org.apache.doris.analysis.TupleDescriptor;
 import org.apache.doris.catalog.Catalog;
+import org.apache.doris.catalog.Column;
+import org.apache.doris.catalog.Database;
 import org.apache.doris.catalog.Resource;
+import org.apache.doris.catalog.FsBroker;
+import org.apache.doris.catalog.MaterializedIndex;
+import org.apache.doris.catalog.MaterializedIndex.IndexExtState;
+import org.apache.doris.catalog.OlapTable;
+import org.apache.doris.catalog.Partition;
+import org.apache.doris.catalog.PrimitiveType;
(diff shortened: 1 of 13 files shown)


fe/src/main/java/org/apache/doris/load/loadv2/SparkLoadJob.java
```diff
@@ -0,0 +1,700 @@
...
+                destSlotDesc.setIsMaterialized(true);
+                destSlotDesc.setColumn(column);
+                if (column.isAllowNull()) {
+                    destSlotDesc.setIsNullable(true);
+                } else {
+                    destSlotDesc.setIsNullable(false);
+                }
+            }
+            // Push broker scan node
+            PushBrokerScanNode scanNode = new PushBrokerScanNode(destTupleDesc);
```
imay, line 156: Why is this ScanNode needed?
Seems that this scanNode will be destruct after this function.

fe/src/main/java/org/apache/doris/load/loadv2/SparkLoadJob.java
```diff
@@ -0,0 +1,700 @@
...
+    private Map<String, Pair<String, Long>> tabletMetaToFileInfo = Maps.newHashMap();
+
+    // --- members below not persist ---
+    // temporary use
+    // one SparkLoadJob has only one table to load
+    // hivedb.table for global dict
+    private String hiveTableName = "";
+    private ResourceDesc resourceDesc;
+    // for spark standalone
+    private SparkAppHandle sparkAppHandle;
```
imay, line 123: When is this assigned?

fe/src/main/java/org/apache/doris/load/loadv2/SparkLoadJob.java
```diff
@@ -0,0 +1,700 @@
...
+        // log loading state
+        logUpdateStateInfo();
+
+        // create push tasks
+        prepareLoadingInfos();
+        submitPushTasks();
+    }
+
+    private void updateToLoadingState(EtlStatus etlStatus, Map<String, Long> filePathToSize) throws LoadException {
+        writeLock();
```
imay, line 354: why not get lock in the outer public function? In most scenario, there is only one writer.
If you put this lock in a private function, there is a case that multiple threads will do proceeEtlFinish concurrently.

fe/src/main/java/org/apache/doris/load/loadv2/SparkLoadJob.java
wyb: setJobProperties in SparkLoadJob.
This is for temporary use. I am investigating load from hive table, and i will update it recently.

morningman (dismissed):
LGTM

morningman (approved):
LGTM

morningman:
The compile error of broker has been fix in #3913 .
So I just approve this PR.

PR #3724, opened 2020-05-29
Fix large string val allocation failure

### LINKED ISSUE #3600: large bitmap serialization failed because of overflow of size

EmmyMiao87:
Large bitmap will need use StringVal to allocate large memory, which is large than MAX_INT.
The overflow will cause serialization failure of bitmap.

Fixed #3600

Change-Id: I720b9ea4646188a8e1402630601928b5f16fedb2

### DIFF
--- a/be/src/exprs/bitmap_function.cpp
+++ b/be/src/exprs/bitmap_function.cpp
@@ -489,7 +489,8 @@ StringVal BitmapFunctions::bitmap_from_string(FunctionContext* ctx, const String
 
     std::vector<uint64_t> bits;
     // The contructor of `stringpiece` only support int type.
-    if (!SplitStringAndParse({(const char*)input.ptr, (int)input.len}, ",", &safe_strtou64, &bits)) {
+    if ((input.len > INT32_MAX) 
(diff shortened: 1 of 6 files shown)


be/src/exprs/bitmap_function.cpp
```diff
@@ -488,7 +488,8 @@ StringVal BitmapFunctions::bitmap_from_string(FunctionContext* ctx, const String
     }
 
     std::vector<uint64_t> bits;
-    if (!SplitStringAndParse({(const char*)input.ptr, input.len}, ",", &safe_strtou64, &bits)) {
+    // TODO(hkp): I think StringPiece's len should also be uint64_t
```
morningman, line 491: Could you explain why not changing this to int64? You can just add comment here

morningman (approved):
LGTM

Issue #3726, opened 2020-05-29
[Spark Load] Rollup Tree Builder


wangbo:
In Spark Load,we want to aggregate doris table's rollup table.
First,we need a data structure to describe the ```derivative relationship``` (such as ```rollup A``` can be calculated from ```rollup B```)  between doris tables's rollup so that we can using the minimal cost to calculate all rollup tables.

Second, we need a builder to build the ```derivative relationship``` data structure.

PR #3727, opened 2020-05-29
(#3726) [Spark Load] Rollup Tree Builder

wangbo:
see #3726

1 A tree data structure to describe doris table's rollup
2 A builder to build the data structure

ref:3433


fe/src/main/java/org/apache/doris/load/loadv2/dpp/MinimumCoverageRollupTreeBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+                             List<String> valueColumns, int id, boolean[] flags) {
+        if (root.children != null) {
+            for (RollupTreeNode child : root.children) {
+                insertIndex(child, indexMeta, keyColumns, valueColumns, id, flags);
+                if (flags[id]) {
+                    return;
+                }
+            }
+        }
+        if (flags[id]) {
```
wyb, line 108: seems to be not needed

fe/src/main/java/org/apache/doris/load/loadv2/dpp/MinimumCoverageRollupTreeBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+        return root;
+    }
+
+    // DFS traverse to build the rollup tree
+    private void insertIndex(RollupTreeNode root, EtlJobConfig.EtlIndex indexMeta,
+                             List<String> keyColumns,
+                             List<String> valueColumns, int id, boolean[] flags) {
+        if (root.children != null) {
+            for (RollupTreeNode child : root.children) {
```
wyb, line 101: It's better to check from behind.

base: c1,c2,c3,c4,c5
rollup index1: c1,c2,c3,c4
rollup index2: c1,c2,c5
rollup index3: c1,c2

It is better that rollup index3 is the child of rollup index2 than rollup index1.

fe/src/main/java/org/apache/doris/load/loadv2/dpp/MinimumCoverageRollupTreeBuilder.java
wangbo: Recursive happends when root node has children, so it is needed

fe/src/main/java/org/apache/doris/load/loadv2/dpp/MinimumCoverageRollupTreeBuilder.java
wangbo: For current Dpp implementation which every rollup need a shuffle, it's better to check from back.
If we change Dpp logic later which one level rollup shuffle once, we need to from head, because  we need to control the level the tree to reduce shuffle times.

morningman (approved):
LGTM

PR #3731, opened 2020-05-29
[Bug][FsBroker] NPE throw when username is empty

### LINKED ISSUE #3730: [Bug][Broker] Null pointer exception when using broker
**Describe the bug**
When using broker with an empty username, NPE is thrown

`IllegalArgumentException: Null user`

**To Reproduce**
create a broker load with broker property:

```
("username" = "")
```

**Expected behavior**
Not throw exception

morningman:
When using Broker with an empty username, a NPE is thrown, which is
not expected.

Fix: #3730


kangkaisen:
empty user_name is valid?

morningman:
> empty user_name is valid?

Not valid, but is OK.
This is just a self self-protection.

kangkaisen (approved):
+1, LGTM

Issue #3733, opened 2020-05-30
[Memory Engine] Add MemTabletScan


decster:
Add simple read(scan) support for MemTablet.

Part of #3466

PR #3734, opened 2020-05-30
[Memory Engine] Add MemTabletScan

decster:
This CL add read support(simple projection only full scan) for MemTablet.
It also add a unit test for MemTablet insert-update-scan loop.

Resloves #3733


be/src/olap/memory/mem_tablet_scan.h
```diff
@@ -0,0 +1,82 @@
...
+
+    uint64_t limit() const { return _limit; }
+
+    const vector<std::string> columns() const { return _columns; }
+
+private:
+    friend class MemTablet;
+
+    uint64_t _version = UINT64_MAX;
+    uint64_t _limit = UINT64_MAX;
```
chaoyli, line 43: What's the use of _limit?

be/src/olap/memory/mem_tablet.h
```diff
@@ -79,6 +81,8 @@ class MemTablet : public BaseTablet {
 
     std::mutex _write_lock;
 
+    std::atomic<uint64_t> _latest_version;
```
chaoyli, line 84: use max_version may be better.

chaoyli (changes requested):

be/src/olap/base_tablet.cpp
```diff
@@ -44,11 +43,13 @@ OLAPStatus BaseTablet::set_tablet_state(TabletState state) {
 }
 
 void BaseTablet::_gen_tablet_path() {
-    std::string path = _data_dir->path() + DATA_PREFIX;
-    path = path_util::join_path_segments(path, std::to_string(_tablet_meta->shard_id()));
-    path = path_util::join_path_segments(path, std::to_string(_tablet_meta->tablet_id()));
-    path = path_util::join_path_segments(path, std::to_string(_tablet_meta->schema_hash()));
-    _tablet_path = path;
+    if (_data_dir != nullptr) {
```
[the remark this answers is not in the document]
decster: In unittest

be/src/olap/memory/mem_tablet_scan.h
decster: For example: select * from tablet limit 10;

be/src/olap/memory/mem_tablet.h
decster: OK

chaoyli (approved):

PR #3736, opened 2020-06-01
[HttpServer] capture convert exception

vagetablechicken:
ref https://github.com/apache/incubator-doris/issues/3725

If parameter `str` is an empty string, it will throw exception too. Maybe we can add an ut for parsing parameters in http server.


morningman:
compile failed

vagetablechicken:
> compile failed

fixed

imay (approved):
LGTM

Issue #3735, opened 2020-06-01
[Bug]If the key column of the table is a double / float type, it will crash when checksum


marising:
**Describe the bug**
If the key column of the table is double/float type，it will crash when checksum.

1. be/src/olap/task/engine_checksum_task.cpp
return_columns do not contain column of double/float type
```
    // ignore float and double type considering to precision lose
    for (size_t i = 0; i < tablet->tablet_schema().num_columns(); ++i) {
        FieldType type = tablet->tablet_schema().column(i).type();
        if (type == OLAP_FIELD_TYPE_FLOAT || type == OLAP_FIELD_TYPE_DOUBLE) {
            continue;
        }

        reader_params.return_columns.push_back(i);
    }
```
2. be/src/olap/row.h Use the all columns
use num_key_columns from Schema，num_key_columns include the column of double/float type
```
template<typename LhsRowType, typename RhsRowType>
int compare_row(const LhsRowType& lhs, const RhsRowType& rhs) {
    for (uint32_t cid = 0; cid < lhs.schema()->num_key_columns(); ++cid) {
        auto res = lhs.schema()->column(cid)->compare_cell(lhs.cell(cid), rhs.cell(cid));
        if (res != 0) {
            return res;
        }
    }
    return 0;
}

// Used to compare row with input scan key. Scan key only contains key columns,
// row contains all key columns, which is superset of key columns.
// So we should compare the common prefix columns of lhs and rhs.
//
// NOTE: if you are not sure if you can use it, please don't use this function.
template<typename LhsRowType, typename RhsRowType>
int compare_row_key(const LhsRowType& lhs, const RhsRowType& rhs) {
    auto cmp_cids = std::min(lhs.schema()->num_column_ids(), rhs.schema()->num_column_ids());
    for (uint32_t cid = 0; cid < cmp_cids; ++cid) {
        auto res = lhs.schema()->column(cid)->compare_cell(lhs.cell(cid), rhs.cell(cid));
        if (res != 0) {
            return res;
        }
    }
    return 0;
}
```
3. Core file

[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `/export/home/lihaibo/deploy_jd/be/lib/palo_be'.
Program terminated with signal 11, Segmentation fault.
[...]

PR #3738, opened 2020-06-01
[Bug]fix the crash of checksum task #3735

marising:
1. the table include key column of double/float type
2. when run checksum task, will use all of key columns to compare
3. schema.column(idx) of double/float type is NULL

#3735

### DIFF
--- a/be/src/olap/row.h
+++ b/be/src/olap/row.h
@@ -59,10 +59,6 @@ bool equal_row(const std::vector<uint32_t>& ids,
 template<typename LhsRowType, typename RhsRowType>
 int compare_row(const LhsRowType& lhs, const RhsRowType& rhs) {
     for (uint32_t cid = 0; cid < lhs.schema()->num_key_columns(); ++cid) {
-        //because the num_column_ids include the column of double/float type
-        if (lhs.schema()->column(cid) == NULL) {
-            continue;
-        }
         auto res = lhs.schema()->column(cid)->compare_cell(lhs.cell(cid), rhs.cell(cid));
         if (res != 0) {
             return res;
@@ -80,10 +76,6 @@ template<typename LhsRowType, typename RhsRowType>
 int compare_row_key(const LhsRowType& lhs, const RhsRowType& rhs) {
     auto cmp_cids = std::min(lhs.schema()->num_column_ids(), rhs.schema()->num_column_ids());
     for (uint32_t cid = 0; cid < cmp_cids; ++cid) {
-        //because the num_column_ids include the column of double/float type
-        if (lhs.schema()->column(cid) == NULL) {
-            continue;
-        }
         auto res = lhs.schema()->column(cid)->compare_cell(lhs.cell(cid), rhs.cell(cid));
         if (res != 0) {
             return res;
@@ -194,7 +186,12 @@ void agg_finalize_row(const std::vector<uint32_t>& ids, RowType* row, MemPool* m
 
 template<typename RowType>
 uint32_t hash_row(const RowType& row, uint32_t seed) {
+    FieldType type;
     for (uint32_t cid : row.schema()->column_ids()) {
+        type = row.schema()->column(cid)->type();
+        if (type == OLAP_FIELD_TYPE_FLOAT || type == OLAP_FIELD_TYPE_DOUBLE) {
+            continue;
+        }
         seed = row.schema()->column(cid)->hash_code(row.cell(cid), seed);
     }
     return seed;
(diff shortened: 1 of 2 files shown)


morningman:
This CL looks good to me.
But I had a question in your issue: #3735, that why key columns include float or double.
We don't support using float or double column as key column.

marising:
It's very confusing for me，doris does not support the key column as double or float，but I've met twice. 

1. I create table and alter table add column
MySQL [checksum]> alter table catalog_sales modify column `cs_sales_price` double KEY null after `cs_item_sk`;
ERROR 1064 (HY000): Float or double can not used as a key, use decimal instead.

2. The table of customer
alter table add column xxx double key null after xxx，but no error message and execute success!

MySQL [search]> desc  xxx_sku_AB;
+--------------+-------------+------+-------+---------+-----------+
| Field        | Type        | Null | Key   | Default | Extra     |
+--------------+-------------+------+-------+---------+-----------+
...
| price        | DOUBLE      | Yes  | true  | N/A     |           |
....
+--------------+-------------+------+-------+---------+-----------

imay:
> 2\. The table of customer
>     alter table add column xxx double key null after xxx，but no error message and execute success!

@marising 
Should create an issue to fix it.

be/src/olap/row.h
```diff
@@ -59,6 +59,10 @@ bool equal_row(const std::vector<uint32_t>& ids,
 template<typename LhsRowType, typename RhsRowType>
 int compare_row(const LhsRowType& lhs, const RhsRowType& rhs) {
     for (uint32_t cid = 0; cid < lhs.schema()->num_key_columns(); ++cid) {
+        //because the num_column_ids include the column of double/float type
```
imay, line 62: This function is used in many senarios. And changing here to fix the problem is not a good idea.
Why not to make column in the schema is not NULL.

imay (changes requested):

morningman:
> 2\. The table of customer
>     alter table add column xxx double key null after xxx，but no error message and execute success!

I guess the table `customer` is a DUPLICATE KEY table, right? So that the `key` in `alter table` stmt is not exactly a `key` column, but just a sort column.

I think its ok to allow a float type column to be a sort column in DUPLICATE KEY table.

@imay , is that right?

imay:
@morningman I'm not sure about it either. However I think it is OK to make double/float column as a sort key.

marising:
CREATE TABLE `xxx_sku_AB` (
......
  `price` double NULL COMMENT "",
......
) ENGINE=OLAP
...
AGGREGATE KEY(..., `price`, ...)
...

be/src/olap/row.h
marising: Ignore the column of double/float type in engine_checksum_task.cpp,  but num_key_columns  include the double/float column index. 

In other ways, I can only add related functions such as reader.init, compare function for sort and aggregation function for checksum. The code volume is a little large.

```    
// ignore float and double type considering to precision lose
    for (size_t i = 0; i < tablet->tablet_schema().num_columns(); ++i) {
        FieldType type = tablet->tablet_schema().column(i).type();
        if (type == OLAP_FIELD_TYPE_FLOAT || type == OLAP_FIELD_TYPE_DOUBLE) {
            continue;
        }

        reader_params.return_columns.push_back(i);
    }
```

marising:
> > 2. The table of customer
> > alter table add column xxx double key null after xxx，but no error message and execute success!
> 
> @marising
> Should create an issue to fix it.

I can't reproduce this problem in the online cluster of customer, and the time for adding columns is a little long audit.log be cleaned up, I will pay attention to this problem

be/src/olap/row.h
marising: 1. remove the code of ignore the column of float/double type's in engine_checksum.cpp
2. ignore the cell of float/double type when compute row's hash code
3. the value of checksum is independent of the sorting of data rows

be/src/olap/row.h
```diff
@@ -186,7 +186,12 @@ void agg_finalize_row(const std::vector<uint32_t>& ids, RowType* row, MemPool* m
 
 template<typename RowType>
 uint32_t hash_row(const RowType& row, uint32_t seed) {
+    FieldType type;
     for (uint32_t cid : row.schema()->column_ids()) {
+        type = row.schema()->column(cid)->type();
```
imay, line 191: Should add comment for this `if` logical to let others know why.

be/src/olap/task/engine_checksum_task.cpp
```diff
@@ -118,8 +113,10 @@ OLAPStatus EngineChecksumTask::_compute_checksum() {
             OLAP_LOG_WARNING("fail to read in reader. [res=%d]", res);
             return res;
         }
-
-        row_checksum = hash_row(row, row_checksum);
+        one_checksum = 0;
```
morningman, line 116: `one_checksum = 0;` this line is meaningless.

imay (dismissed):
LGTM

morningman:
hi @marising , if you finished this PR, please let me know, thank u.

marising:
@morningman Ok，I'll tell you, if our internal review passes.

morningman (approved):
LGTM

PR #3739, opened 2020-06-01
Support materialized view schema change

HangyuanLiu:
Support materialized view schema change history data
1、bitmap
2、hll
3、count


be/src/olap/schema_change.cpp
```diff
@@ -266,6 +260,92 @@ bool RowBlockChanger::change_row_block(
...
+                                case OLAP_FIELD_TYPE_UNSIGNED_BIGINT:
+                                    origin_value = *(uint64_t *) src;
+                                    break;
+                                default:
+                                    LOG(WARNING) << "the column type which was altered from was unsupported."
+                                                 << " from_type="
+                                                 << ref_block->tablet_schema().column(ref_column).type();
+                                    return false;
+                            }
+                            if (origin_value < 0) {
```
imay, line 314: origin_value is unsigned, this will always be false

be/src/olap/schema_change.cpp
```diff
@@ -266,6 +260,92 @@ bool RowBlockChanger::change_row_block(
...
+                    // 不需要的row，每次处理到这个row时就跳过
+                    if (need_filter_data && is_data_left_vec[row_index] == 0) {
+                        continue;
+                    }
+
+                    // 指定新的要写入的row index（不同于读的row_index）
+                    mutable_block->get_row(new_row_index++, &write_helper);
+                    ref_block->get_row(row_index, &read_helper);
+
+                    if (_schema_mapping[i].materialized_function == "to_bitmap") {
```
imay, line 277: better do this if and switch out of for loop

imay (changes requested):

be/src/olap/schema_change.cpp
```diff
@@ -1425,15 +1514,17 @@ OLAPStatus SchemaChangeHandler::schema_version_convert(
 
     // a. 解析Alter请求，转换成内部的表示形式
     // 不使用DELETE_DATA命令指定的删除条件
-    RowBlockChanger rb_changer(new_tablet->tablet_schema(), base_tablet);
+    RowBlockChanger rb_changer(new_tablet->tablet_schema());
     bool sc_sorting = false;
     bool sc_directly = false;
 
+    const std::map<std::string, AlterMaterializedViewParam> materialized_function_map;
```
kangkaisen, line 1521: unorder_map is enough？

be/src/olap/schema_change.cpp
```diff
@@ -266,6 +260,92 @@ bool RowBlockChanger::change_row_block(
         int32_t ref_column = _schema_mapping[i].ref_column;
 
         if (_schema_mapping[i].ref_column >= 0) {
+            if (!_schema_mapping[i].materialized_function.empty()) {
```
kangkaisen, line 263: Would better abstract the materialized_view logic to a method. Add add more comment.

fe/src/main/java/org/apache/doris/alter/RollupJobV2.java
```diff
@@ -352,14 +353,21 @@ protected void runWaitingTxnJob() throws AlterCancelException {
                     long rollupTabletId = rollupTablet.getId();
                     long baseTabletId = tabletIdMap.get(rollupTabletId);
 
+                    Map<String, Expr> defileExprs = Maps.newHashMap();
```
EmmyMiao87, line 356: defileExpr? or defineExpr?

fe/src/main/java/org/apache/doris/task/AlterReplicaTask.java
```diff
@@ -38,10 +46,22 @@
     private long jobId;
     private AlterJobV2.JobType jobType;
 
+    private Map<String, Expr> defileExprs;
```
EmmyMiao87, line 49: defineExpr

fe/src/main/java/org/apache/doris/task/AlterReplicaTask.java
```diff
@@ -93,6 +115,15 @@ public TAlterTabletReqV2 toThrift() {
         TAlterTabletReqV2 req = new TAlterTabletReqV2(baseTabletId, signature, baseSchemaHash, newSchemaHash);
         req.setAlter_version(version);
         req.setAlter_version_hash(versionHash);
+        if (defileExprs != null) {
+            for (Map.Entry<String, Expr> expr : defileExprs.entrySet()) {
```
EmmyMiao87, line 119: `for(Map.Entry<String, Expr> entry: defineExprs.entrySet())`

fe/src/main/java/org/apache/doris/task/AlterReplicaTask.java
```diff
@@ -93,6 +115,15 @@ public TAlterTabletReqV2 toThrift() {
...
+                List<SlotRef> slots = Lists.newArrayList();
+                expr.getValue().collect(SlotRef.class, slots);
+                Preconditions.checkState(slots.size() == 1);
```
EmmyMiao87, line 122: The expr must be contain only one slot after analyzing.

fe/src/main/java/org/apache/doris/task/AlterReplicaTask.java
```diff
@@ -93,6 +115,15 @@ public TAlterTabletReqV2 toThrift() {
         TAlterTabletReqV2 req = new TAlterTabletReqV2(baseTabletId, signature, baseSchemaHash, newSchemaHash);
         req.setAlter_version(version);
         req.setAlter_version_hash(versionHash);
+        if (defileExprs != null) {
```
EmmyMiao87, line 118: Please add some comments and example of bitmap or hll

be/src/olap/schema_change.cpp
```diff
@@ -266,6 +260,92 @@ bool RowBlockChanger::change_row_block(
         int32_t ref_column = _schema_mapping[i].ref_column;
```
EmmyMiao87, line 260: Maybe you could refactor here. The main step is 
```
for (columns) {
    for(row_index) {
        1. bitmap
        2. hll
        3. convert type
        ....
    }
}
```

be/src/olap/schema_change.cpp
EmmyMiao87: +1. Abstract the conversion of bitmap and hll into functions.

be/src/olap/schema_change.cpp
```diff
@@ -266,6 +260,92 @@ bool RowBlockChanger::change_row_block(
...
+                    }
+
+                    // 指定新的要写入的row index（不同于读的row_index）
+                    mutable_block->get_row(new_row_index++, &write_helper);
+                    ref_block->get_row(row_index, &read_helper);
+
+                    if (_schema_mapping[i].materialized_function == "to_bitmap") {
+                        write_helper.set_not_null(i);
+                        BitmapValue bitmap;
+                        if (!read_helper.is_null(ref_column)) {
```
EmmyMiao87, line 280: `if (read_helper.is_null(ref_column)) write_helper.set_null(i)`

be/src/olap/schema_change.cpp
```diff
@@ -266,6 +260,92 @@ bool RowBlockChanger::change_row_block(
...
+                                return false;
+                            }
+                            bitmap.add(origin_value);
+                        }
+                        char *buf = reinterpret_cast<char *>(mem_pool->allocate(bitmap.getSizeInBytes()));
+                        Slice dst(buf, bitmap.getSizeInBytes());
+                        bitmap.write(dst.data);
+                        write_helper.set_field_content(i, reinterpret_cast<char *>(&dst), mem_pool);
+                    } else if (_schema_mapping[i].materialized_function == "hll_hash") {
+                        write_helper.set_not_null(i);
```
EmmyMiao87, line 326: if (read_helper.is_null(ref_column)) write_helper.set_null(i)

be/src/olap/schema_change.cpp
```diff
@@ -1810,10 +1902,26 @@ OLAPStatus SchemaChangeHandler::_parse_request(TabletSharedPtr base_tablet,
             continue;
         }
 
+        if (materialized_function_map.find(column_name) != materialized_function_map.end()) {
```
EmmyMiao87, line 1905: Maybe the column mapping should be init by the columnPB which comes from the thrift of defineExpr in column.
The AlterReplicaTask does not need the TAlterMaterializedViewParam

be/src/olap/schema_change.cpp
```diff
@@ -1383,6 +1463,15 @@ OLAPStatus SchemaChangeHandler::_do_process_alter_tablet_v2(const TAlterTabletRe
         sc_params.new_tablet = new_tablet;
         sc_params.ref_rowset_readers = rs_readers;
         sc_params.delete_handler = delete_handler;
+        if (request.__isset.materialized_view_params) {
+            for (auto item : request.materialized_view_params) {
+                AlterMaterializedViewParam mvParams;
```
EmmyMiao87, line 1468: Encapsulate the logic here. There will be many new rules for converting expressions into strings

be/src/olap/schema_change.cpp
HangyuanLiu: the result of to_bitmap or hll_hash always not null

be/src/olap/schema_change.cpp
HangyuanLiu: I think the materialized view is a calculation of an expression. Logic of expr should not be written to the storage layer. Different tablets are irrelevant at the storage level.
So we can refactor this part of the code until the uniform expression evaluates the logic.

gensrc/thrift/AgentService.thrift
```diff
@@ -104,6 +104,13 @@ struct TAlterTabletReqV2 {
     // version of data which this alter task should transform
     5: optional Types.TVersion alter_version
     6: optional Types.TVersionHash alter_version_hash // Deprecated
+    7: optional list<TAlterMaterializedViewParam> materialized_view_params
+}
+
+struct TAlterMaterializedViewParam {
+    1: required string column_name
+    2: required string origin_column_name
+    3: optional string mv_expr
```
morningman, line 113: Why not using TExpr?

be/src/olap/schema_change.cpp
```diff
@@ -194,6 +188,105 @@ ColumnMapping* RowBlockChanger::get_mutable_column_mapping(size_t column_index)
...
+               int field_idx, int ref_field_idx, MemPool* mem_pool) {
+    write_helper->set_not_null(ref_field_idx);
+    BitmapValue bitmap;
+    if (!read_helper->is_null(ref_field_idx)) {
+        uint64_t origin_value;
+        char *src = read_helper->cell_ptr(ref_field_idx);
+        switch (ref_column.type()) {
+            case OLAP_FIELD_TYPE_TINYINT:
+                if (*(int8_t *) src < 0) {
+                    LOG(WARNING) << "The input: " << *(int8_t *) src
```
morningman, line 202: How about return Status with this error message?
And do not print log here, it may lead to lots of log, and this error info should return to user.

be/src/olap/schema_change.cpp
```diff
@@ -194,6 +188,105 @@ ColumnMapping* RowBlockChanger::get_mutable_column_mapping(size_t column_index)
...
+    buf.resize(hll.serialize((uint8_t *) buf.c_str()));
+    Slice dst(buf);
+    write_helper->set_field_content(field_idx, reinterpret_cast<char *>(&dst), mem_pool);
+    return true;
+}
+
+bool count(RowCursor* read_helper, RowCursor* write_helper, const TabletColumn& ref_column,
+           int field_idx, int ref_field_idx, MemPool* mem_pool) {
+    write_helper->set_not_null(field_idx);
+    int64_t count = read_helper->is_null(field_idx) ? 0 : 1;
```
morningman, line 285:
```suggestion
    int8_t count = read_helper->is_null(field_idx) ? 0 : 1;
```

gensrc/thrift/AgentService.thrift
HangyuanLiu: The materialized view is a calculation of an expression.But we construct materialized view in olap/schema_change.cpp. we can't use expr and exec in olap layer now.So I used some hard coding to deal with this problem.May be we can refactor this part of the code until the uniform expression evaluates the logic.

be/src/olap/schema_change.cpp
HangyuanLiu: I think the type of count should be int64_t ,because the result type of count is BIGINT。So  we shoul keep set_field_content is 8 byte.

be/src/olap/schema_change.cpp
HangyuanLiu: This is a good suggestion，but it involves a lot of modification points, I will add this feature optimization in the next PR

be/src/olap/schema_change.cpp
morningman: > I think the type of count should be int64_t ,because the result type of count is BIGINT。So we should keep set_field_content is 8 byte.

Oh, I misunderstood. OK.

be/src/olap/schema_change.cpp
EmmyMiao87: > How about return Status with this error message?
> And do not print log here, it may lead to lots of log, and this error info should return to user.

+1

be/src/olap/schema_change.cpp
```diff
@@ -194,6 +188,105 @@ ColumnMapping* RowBlockChanger::get_mutable_column_mapping(size_t column_index)
...
+    bitmap.write(dst.data);
+    write_helper->set_field_content(field_idx, reinterpret_cast<char *>(&dst), mem_pool);
+    return true;
+}
+
+bool hll_hash(RowCursor* read_helper, RowCursor* write_helper, const TabletColumn& ref_column,
+              int field_idx, int ref_field_idx, MemPool* mem_pool) {
+    write_helper->set_not_null(field_idx);
+    HyperLogLog hll;
+    if (!read_helper->is_null(ref_field_idx)) {
```
EmmyMiao87, line 263: If the read_field is null, the hll could be null also. Same as bitmap.

be/src/olap/schema_change.cpp
```diff
@@ -194,6 +188,105 @@ ColumnMapping* RowBlockChanger::get_mutable_column_mapping(size_t column_index)
...
+    }
+    std::string buf;
+    buf.resize(hll.max_serialized_size());
+    buf.resize(hll.serialize((uint8_t *) buf.c_str()));
+    Slice dst(buf);
+    write_helper->set_field_content(field_idx, reinterpret_cast<char *>(&dst), mem_pool);
+    return true;
+}
+
+bool count(RowCursor* read_helper, RowCursor* write_helper, const TabletColumn& ref_column,
```
EmmyMiao87, line 282: How do you distinguish count(k1) and count(*)

be/src/olap/schema_change.cpp
EmmyMiao87: The value of count can only be positive, why don't you apply `uint`?

be/src/olap/schema_change.cpp
```diff
@@ -266,6 +359,33 @@ bool RowBlockChanger::change_row_block(
...
 
         if (_schema_mapping[i].ref_column >= 0) {
+            if (!_schema_mapping[i].materialized_function.empty()) {
+                bool (*_do_materialized_transform) (RowCursor*, RowCursor*, const TabletColumn&, int, int, MemPool* );
+                if (_schema_mapping[i].materialized_function == "to_bitmap") {
+                    _do_materialized_transform = to_bitmap;
+                } else if (_schema_mapping[i].materialized_function == "hll_hash") {
+                    _do_materialized_transform = hll_hash;
+                } else if (_schema_mapping[i].materialized_function == "count") {
+                    _do_materialized_transform = count;
```
EmmyMiao87, line 369: How about count(*)?

EmmyMiao87 (changes requested):

PR #3750, opened 2020-06-02
[Bug] Fix a bug that tablet may be modified to BETA_ROWSET after cloned

acelyc111:
TabletMeta's _preferred_rowset_type is not initialized after object constructing and
may be a random value, and this field is not updated when create ALPHA_ROWSET tablet,
and it will not be serialized into pb in this case. So if cloning an ALPHA_ROWSET
tablet from another BE, this new created local tablet's _preferred_rowset_type field
may be random as BETA_ROWSET and can not be overwrote after cloned, then new input
rows will be wrote as BETA_ROWSET format which is not we expect.
This patch fix this bug by giving _preferred_rowset_type a default value and updating
this field when create any type of tablet, and add an unit test and related overwrite
equal operator functions.


be/src/olap/tablet_schema.cpp
```diff
@@ -337,14 +337,12 @@ void TabletColumn::to_schema_pb(ColumnPB* column) {
...
-    : _num_columns(0),
-      _num_key_columns(0),
-      _num_null_columns(0),
-      _num_short_key_columns(0) { }
-
 void TabletSchema::init_from_pb(const TabletSchemaPB& schema) {
     _keys_type = schema.keys_type();
+    _num_columns = 0;
+    _num_key_columns = 0;
+    _num_null_columns = 0;
```
morningman, line 344: I am not sure, but did you miss the `_num_short_key_columns` here?

acelyc111: @morningman `_num_short_key_columns` will be assigned in line 358 directly, not like the above variables which will be increased in the following loop, so I didn't reset it here.

be/src/olap/tablet_schema.h
```diff
@@ -107,22 +113,30 @@ class TabletSchema {
...
+    size_t _num_rows_per_row_block = 0;
+    CompressKind _compress_kind = COMPRESS_NONE;
+    size_t _next_column_unique_id = 0;
 
     bool _has_bf_fpp = false;
     double _bf_fpp = 0;
     bool _is_in_memory = false;
 };
 
+bool operator==(const TabletSchema& a, const TabletSchema& b);
```
imay, line 137: Is this function used?

acelyc111: TabletMeta has a TabletSchema member, overload TabletSchema's == operator is aim to overload TabletMeta's == operator.
Now TabletMeta's == operator is used for convenient unit test.

imay (approved):
LGTM

PR #3753, opened 2020-06-03
[FIX] fix binaryPredicte's equals function ignore op

### LINKED ISSUE #3756: BinaryPredict's equals function compares ignore the op
**Describe the bug**
BinaryPredict's equals function compares ignore the op, the  opcode is not init in this class, it will always be INVALID_OPCODE, so if a BinaryPredict has same child  it will be qual, like `a<1` and `a>=1`
```
    @Override
    public boolean equals(Object obj) {
        if (!super.equals(obj)) {
            return false;
        }
        return ((BinaryPredicate) obj).opcode == this.opcode;
    }
```

yangzhg:
BinaryPredicte's equals funtion compare by opccode ,
but the opcode is never inited, 
so it will return true if this child is same,  for example `a>1` and `a<1` are equal
Fixes #3756

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/BinaryPredicate.java
+++ b/fe/src/main/java/org/apache/doris/analysis/BinaryPredicate.java
@@ -226,7 +226,7 @@ public class BinaryPredicate extends Predicate implements Writable {
         if (!super.equals(obj)) {
             return false;
         }
-        return this.op.equals(((BinaryPredicate) obj).getOp());
+        return ((BinaryPredicate) obj).opcode == this.opcode;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


morningman (dismissed):
LGTM

fe/src/main/java/org/apache/doris/analysis/BinaryPredicate.java
```diff
@@ -225,7 +226,7 @@ public boolean equals(Object obj) {
         if (!super.equals(obj)) {
             return false;
         }
-        return ((BinaryPredicate) obj).opcode == this.opcode;
+        return ((BinaryPredicate) obj).opcode == this.opcode && op.equals(((BinaryPredicate) obj).getOp());
```
kangkaisen, line 229: I think only compare op is enough.

morningman: It only for safe, don't know what the `opcode` for.
If it is useless, we can remove it in next PR.

EmmyMiao87 (approved):
LGTM

PR #3755, opened 2020-06-03
[Plugin] Add timeout of connection when downloading the plugins from URL

### LINKED ISSUE #3754: [Plugin] install plugin block when downloading from URL
**Describe the bug**
Install plugin process blocks when downloading from URL.

**To Reproduce**
install a plugin from an invalid URL which has some network problem.

**Expected behavior**
Throw a timeout exception

morningman:
Fix #3754

If no timeout is set, the download process may be blocked forever.

### DIFF
--- a/fe/src/main/java/org/apache/doris/plugin/PluginZip.java
+++ b/fe/src/main/java/org/apache/doris/plugin/PluginZip.java
@@ -116,13 +116,13 @@ class PluginZip {
         cleanPathList.add(zip);
 
         // download zip
-        try (InputStream in = Util.getInputStreamFromUrl(source, null, HTTP_TIMEOUT_MS, HTTP_TIMEOUT_MS)) {
+        try (InputStream in = getInputStreamFromUrl(source)) {
             Files.copy(in, zip, StandardCopyOption.REPLACE_EXISTING);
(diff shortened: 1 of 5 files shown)


yangzhg (dismissed):
LGTM

fe/src/main/java/org/apache/doris/plugin/PluginZip.java
```diff
@@ -113,13 +113,13 @@ Path downloadRemoteZip(Path targetPath) throws IOException, UserException {
         cleanPathList.add(zip);
 
         // download zip
-        try (InputStream in = openUrlInputStream(source)) {
+        try (InputStream in = Util.getInputStreamFromUrl(source, null, 5000, 5000)) {
```
kangkaisen, line 116: Should be static const value.
If the timeout isn't a config, I think set a larger value maybe better when the plugin resource is big and network is poor.

morningman: I change it to a config

kangkaisen (dismissed):
+1, LGTM

EmmyMiao87 (approved):

PR #3761, opened 2020-06-03
[StorageEngine] Make StorageEngine::open return more detailed info

acelyc111:
StorageEngine::open just return a very vague status info when failed,
we have to check logs to find out the root reason, and it's not
convenient to check logs if we run unit tests in CI dockers.
It would be better to return more detailed failure info to point out
the root reason, for example, it may return error status with message
"file descriptors limit is too small".

### DIFF
--- a/be/src/olap/data_dir.cpp
+++ b/be/src/olap/data_dir.cpp
@@ -261,14 +261,15 @@ Status DataDir::_init_meta() {
     // init meta
     _meta = new (std::nothrow) OlapMeta(_path);
     if (_meta == nullptr) {
-        LOG(WARNING) << "new olap meta failed";
-        return Status::MemoryAllocFailed("new olap meta failed");
+        RETURN_NOT_OK_STATUS_WITH_WARN(
+            Status::MemoryAllocFailed("allocate memory for OlapMeta failed"),
+            "new OlapMeta failed");
     }
     OLAPStatus res = _meta->init();
     if (res != OLAP_SUCCESS) {
-        LOG(WARNING) << "init meta failed";
-        return Status::InternalError(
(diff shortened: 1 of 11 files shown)


be/src/olap/data_dir.cpp
```diff
@@ -267,12 +281,13 @@ Status DataDir::_init_meta() {
     _meta = new (std::nothrow) OlapMeta(_path);
     if (_meta == nullptr) {
         LOG(WARNING) << "new olap meta failed";
-        return Status::InternalError("new olap meta failed");
+        return Status::MemoryAllocFailed("new olap meta failed");
```
morningman, line 284: Why not change it to `RETURN_NOT_OK_STATUS_WITH_WARN`?

acelyc111: Updated.

be/src/olap/storage_engine.cpp
```diff
@@ -170,22 +167,28 @@ OLAPStatus StorageEngine::_open() {
...
     std::vector<std::thread> threads;
     std::atomic<bool> init_error{false};
+    SpinLock error_msg_lock;
+    std::string error_msg;
     for (auto& path : _options.store_paths) {
         DataDir* store = new DataDir(path.path, path.capacity_bytes, path.storage_medium,
                                      _tablet_manager.get(), _txn_manager.get());
         tmp_stores.emplace_back(store);
-        threads.emplace_back([store, &init_error]() {
+        threads.emplace_back([&]() {
```
imay, line 183: better to capture explictly.

imay (approved):
LGTM

morningman (approved):
LGTM

PR #3766, opened 2020-06-03
[Bug] Clear Txn when load been cancelled

### LINKED ISSUE #3767: Txn is not cleared when load been cancelled.
You can reproduce the bug like this.
```
CREATE TABLE `table1` (
  `siteid` tinyint(4) NULL DEFAULT "10" COMMENT "",
  `citycode` smallint(6) NULL COMMENT "",
  `username` varchar(32) NULL DEFAULT "" COMMENT "",
  `pv` bigint(20) SUM NULL DEFAULT "0" COMMENT ""
) ENGINE=OLAP
AGGREGATE KEY(`siteid`, `citycode`, `username`)
COMMENT "OLAP"
DISTRIBUTED BY HASH(`siteid`) BUCKETS 1
PROPERTIES (
"replication_num" = "1",
"in_memory" = "false",
"storage_format" = "DEFAULT"
);
```
Data is like this
```
1111	1	'jim'	2
2222	1	'grace'	2
3333	2	'tom'	2
4444	3	'bush'	3
5555	3	'helen'	3
```

chaoyli:
If you a load task encoutering error, it will be cancelled.
At this time, FE will clear the Txn according to the DbName.
In FE, DbName should be added by cluter name.
If missing cluster name, it will encounter NullPointer.
As a result, the Txn will still exists until timeout.

### DIFF
--- a/fe/src/main/java/org/apache/doris/service/FrontendServiceImpl.java
+++ b/fe/src/main/java/org/apache/doris/service/FrontendServiceImpl.java
@@ -808,7 +808,8 @@ public class FrontendServiceImpl implements FrontendService.Iface {
             checkPasswordAndPrivs(cluster, request.getUser(), request.getPasswd(), request.getDb(),
                     request.getTbl(), request.getUser_ip(), PrivPredicate.LOAD);
         }
-        Database db = Catalog.getInstance().getDb(request.getDb());
+        String dbName = ClusterNamespace.getFullName(cluster, request.getDb());
+        Database db = Catalog.getInstance().getDb(dbName);
         if (db == null) {
             throw new MetaNotFoundException("db " + request.getDb() + " does not exist");
         }


fe/src/main/java/org/apache/doris/service/FrontendServiceImpl.java
```diff
@@ -808,7 +808,8 @@ private void loadTxnRollbackImpl(TLoadTxnRollbackRequest request) throws UserExc
             checkPasswordAndPrivs(cluster, request.getUser(), request.getPasswd(), request.getDb(),
                     request.getTbl(), request.getUser_ip(), PrivPredicate.LOAD);
         }
-        Database db = Catalog.getInstance().getDb(request.getDb());
+        String dbName = cluster + ":" + request.getDb();
```
morningman, line 811: Use `ClusterNamespace.getFullName()` for safety

morningman (changes requested):
Please create an issue to describe the bug you met

fe/src/main/java/org/apache/doris/service/FrontendServiceImpl.java
chaoyli: OK, I fixed it

chaoyli:
> Please create an issue to describe the bug you met

I have already linked issues.

morningman (approved):
LGTM

PR #3770, opened 2020-06-04
Add some log to make it easier to find out bug

yangzhg:
Added some logs to record to which be a query was sent.
Increasing the efficiency of tracing the problem

### DIFF
--- a/fe/src/main/java/org/apache/doris/qe/Coordinator.java
+++ b/fe/src/main/java/org/apache/doris/qe/Coordinator.java
@@ -409,12 +409,15 @@ public class Coordinator {
                     toBrpcHost(execBeAddr),
                     queryOptions.query_timeout * 1000);
 
+            LOG.info("dispatch query job: {} to {}", queryId, topParams.instanceExecParams.get(0).host);
+
             // set the broker address for OUTFILE sink
             ResultSink resultSink = (ResultSink) topParams.fragment.getSink();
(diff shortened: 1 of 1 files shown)


fe/src/main/java/org/apache/doris/qe/Coordinator.java
```diff
@@ -409,12 +409,15 @@ public void exec() throws Exception {
                     toBrpcHost(execBeAddr),
                     queryOptions.query_timeout * 1000);
 
+            LOG.info("dispatch query job: {} to {}", queryId, topParams.instanceExecParams.get(0).host);
```
morningman, line 412: print query id like this: `DebugUtil.printId(queryId)`

fe/src/main/java/org/apache/doris/qe/Coordinator.java
```diff
@@ -425,6 +428,7 @@ public void exec() throws Exception {
             List<Long> relatedBackendIds = Lists.newArrayList(addressToBackendID.values());
             Catalog.getCurrentCatalog().getLoadManager().initJobProgress(jobId, queryId, instanceIds,
                     relatedBackendIds);
+            LOG.info("dispatch load job: {} to {}", queryId, addressToBackendID.keySet());
```
morningman, line 431: print query id like this: `DebugUtil.printId(queryId)`

morningman (approved):
LGTM

PR #3772, opened 2020-06-04
Fix the error selectedIndexId when keysType of table is UNIQUE

### LINKED ISSUE #3771: Error selectedIndexId when KeysType of table is UNIQUE
**Describe the bug**
When user query the UNIQUE table, the new mv selector could not select correct indexId.
The warning log like 

```
2020-06-04 14:45:58,361 WARN 101 [OlapScanNode.updateScanRangeInfoByNewMVSelector():232] Using the old scan range info instead of the new one. The key type of table is aggregated., The new selected index id 0, pre aggregation tag true, reason null. The old selected index
id 213392 pre aggregation tag false reason No AggregateInfo
```

**To Reproduce**
1. Create a unique table
```
 create table unique_table(k1 int, k2 int) UNIQUE KEY(k1) distributed by hash(k1) properties("replication_num" = "1");
```
2. insert value
```
insert into unique_table values(1,1);
```
3. query
```
select * from unique_table;
```
4. see fe.log

```
2020-06-04 15:18:01,420 WARN 131 [OlapScanNode.updateScanRangeInfoByNewMVSelector():232] Using the old scan range info instead of the new one. The key type of table is aggregated., The new selected index id 0, pre aggregation tag true, reason null. The old selecte        d index id 10007 pre aggregation tag false reason No AggregateInfo
```

**Expected behavior**
No warning log, the new selector choose the correct index id which is same as the rollup selector.

EmmyMiao87:
The unique table also should be compensated candidate index.
The reason is the same as the agg table type.

Fixed #3771.
Change-Id: Ic04b0360a0b178cb0b6ee635e56f48852092ec09

### DIFF
--- a/fe/src/main/java/org/apache/doris/planner/MaterializedViewSelector.java
+++ b/fe/src/main/java/org/apache/doris/planner/MaterializedViewSelector.java
@@ -131,7 +131,8 @@ public class MaterializedViewSelector {
         // Step5: columns required to compute output expr are available in the view output
         checkOutputColumns(columnNamesInQueryOutput.get(tableName), candidateIndexIdToMeta);
         // Step6: if table type is aggregate and the candidateIndexIdToSchema is empty,
-        if (table.getKeysType() == KeysType.AGG_KEYS && candidateIndexIdToMeta.size() == 0) {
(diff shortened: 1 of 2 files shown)


morningman (approved):
LGTM

PR #3773, opened 2020-06-04
fix some flaw in format-code tutorial

xy720:
Fix some problems found when using MAC to refer to format-code tutorial.
This commit adds some details to the document in formate-code tutorial.


morningman (approved):
LGTM

PR #3784, opened 2020-06-05
[Code Refactor] Remove Catalog.getInstance() method

### LINKED ISSUE #3778: [Catalog] Should not use `getInstance()` of Catalog anymore
Catalog is currently a singleton class.

In the Master FE, this singleton class actually has two instances:  One is the Catalog used for formal services(Called `Formal Catalog`), and the other is the catalog used for metadata image generation(Called `Checkpoint Catalog`).

Therefore, we provide two ways to obtain catalog instance:
1. `getInstance()` is used to get Formal Catalog.
2. `getCurrentCatalog()` is more intelligent, it will determine which catalog is actually returned by determining whether the current thread is a Checkpoint thread. If current thread is checkpoint thread, it will return Checkpoint Catalog. Otherwise, it will return Formal Catalog.

I decided to delete the `getInstance()` method and use `getCurrentCatalog()` uniformly.
Because the `getInstance()` method may introduce many errors.

For example, in the checkpoint process, all operations should be directed against the Checkpoint Catalog. If the `getInstance()` method is called incorrectly, it is equivalent to operating the Formal Catalog in the wrong operation, which will cause metadata errors.

morningman:
Use Catalog.getCurrentCatalog() instead, to avoid potential meta operation error.

Fix: #3778

This CL does not change any logic.


kangkaisen (approved):
+1. LGTM

PR #3786, opened 2020-06-05
[Bug] Fix bug that should not use "!=" to judge the equivalence of Type

### LINKED ISSUE #3785: [Bug][Load] Now() function is only support for DATE/DATETIME column
**Describe the bug**
When using broker like this:

```
LOAD LABEL xxx
...
(k1, k2, k3)
SET
(k4 = now())
...
```

Even if `k4`'s type is `DATETIME`, this load process also report error

`Now() function is only support for DATE/DATETIME column` 

**To Reproduce**
1. create a table with datetime column
2. restart the FE twice to make sure the catalog is loaded from a meta image file.
3. send the load statement.

**Expected behavior**
Load success.

morningman:
`org.apache.doris.catalog.Type` is not an enum, so should not judge the
equivalence of Type using "==" or "!="

Fix #3785

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/CompoundPredicate.java
+++ b/fe/src/main/java/org/apache/doris/analysis/CompoundPredicate.java
@@ -17,18 +17,19 @@
 
 package org.apache.doris.analysis;
 
-import java.util.List;
-import java.util.Objects;
-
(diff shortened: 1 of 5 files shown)


kangkaisen (approved):
+1， LGTM

Issue #3792, opened 2020-06-06
Non-nullable Largeint  column max and min result is wrong


kangkaisen:
1
```
CREATE TABLE IF NOT EXISTS `test_basic` (
  `id_int` int(11) NOT NULL,
  `id_tinyint` tinyint NOT NULL,
  `id_smallint` smallint NOT NULL,
  `id_bigint` bigint NOT NULL,
  `id_largeint` largeint NOT NULL,
  `id_float` float NOT NULL,
  `id_double` double NOT NULL,
  `id_char` char(10) NOT NULL,
  `id_varchar` varchar(100) NOT NULL
) ENGINE=OLAP
DUPLICATE KEY(`id_int`)
DISTRIBUTED BY HASH(`id_int`) BUCKETS 10
PROPERTIES (
 "replication_num" = "1"
);

```

2 
```
insert into test_basic (id_int, id_tinyint, id_smallint, id_bigint, id_largeint, id_float, id_double, id_char, id_varchar) values
    (1, 10, 100, 1000, 10000, 100000.1, 1000000.1, 'kks_char', 'kks_varchar'),
    (2, 20, 200, 2000, 20000, 200000.1, 2000000.1, 'kks_char2', 'kks_varchar2'),
    (3, 30, 300, 3000, 30000, 300000.1, 3000000.1, 'kks_char3', 'kks_varchar3'),
    (4, 40, 400, 4000, 40000, 400000.1, 4000000.1, 'kks_char4', 'kks_varchar4');
```

3. min id_largeint should be 10000, not 0.

PR #3793, opened 2020-06-06
Fix largeint max min bug

kangkaisen:
For #3792

### DIFF
--- a/be/src/exprs/expr_value.h
+++ b/be/src/exprs/expr_value.h
@@ -47,7 +47,7 @@ struct ExprValue {
     DecimalValue decimal_val;
     DecimalV2Value decimalv2_val;
 
-    ExprValue() : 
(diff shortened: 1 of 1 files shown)


morningman (approved):
LGTM

be/src/exprs/expr_value.h
```diff
@@ -177,7 +177,7 @@ struct ExprValue {
             return &bigint_val;
 
         case TYPE_LARGEINT:
-            large_int_val = 0;
+            large_int_val = std::numeric_limits<int128_t>::min();
```
chaoyli, line 180: I have a question.
Does std::numeric_limits supports int128_t?

kangkaisen: Yes. It does.

PR #3798, opened 2020-06-08
Fix error message not match of restore job when job is time out

### LINKED ISSUE #3797: Restore job timeout will be reported as user canceled
**Describe the bug**
if a restore job is timeout  the timeout status will be reset to use canceled be cancel function 
image

yangzhg:
For the current code if a restore job is time out it will be reported as  user canceled. This error message is very misleading
Fixes #3797


kangkaisen (approved):
+1, LGTM

PR #3813, opened 2020-06-09
[Doris On ES] Skip function_call expr when process predicate

wuyunfeng:
https://github.com/apache/incubator-doris/issues/3801
Do not push-down ` function_call` such as `split_xxx` when process predicate, Doris BE is responsible for processing these `predicate`

All rows in table:

```
+------+------+------+------------+------------+
| k1   | k2   | k3   | UpdateTime | ArriveTime |
+------+------+------+------------+------------+
| NULL | NULL | kkk1 |  123456789 |       NULL |
| kkk1 | NULL | NULL |  123456789 |       NULL |
| NULL | kkk2 | NULL |  123456789 |       NULL |
+------+------+------+------------+------------+
```

SQL 1:

```
mysql> select * from (select split_part(k1, "1", 1) as kk from case_replay_for_milimin) t where t.kk is not null;
+------+
| kk   |
+------+
| kkk  |
+------+
1 row in set (0.02 sec)
```

SQL 2:

```
mysql> select * from (select split_part(k1, "1", 1) as kk from case_replay_for_milimin) t where t.kk > 'a';      
+------+
| kk   |
+------+
| kkk  |
+------+
```

SQL 3:

```
mysql> select * from (select split_part(k1, "1", 1) as kk from case_replay_for_milimin) t where t.kk > '2';
+------+
| kk   |
+------+
| kkk  |
+------+
1 row in set (0.03 sec)
```

SQL 4:

```
mysql> select * from (select split_part(k1, "1", 1) as kk from case_replay_for_milimin) t where t.kk > 2;
Empty set (0.02 sec)

```

Some problems：

I do not know whether SQL 4 meets expectations when Doris filters the data.

### DIFF
--- a/be/src/exec/es/es_predicate.cpp
+++ b/be/src/exec/es/es_predicate.cpp
@@ -54,8 +54,7 @@ using namespace std;
             const Expr* expr_without_cast = Expr::expr_without_cast(expr); \
             if (expr_without_cast->node_type() != TExprNodeType::SLOT_REF) { \
                 return Status::InternalError("build disjuncts failed: child is not slot ref"); \
-            }
-        } \
+            } \
     } while (false)
 
 std::string ExtLiteral::value_to_string() {
--- a/be/src/exec/es/es_predicate.cpp
+++ b/be/src/exec/es/es_predicate.cpp
@@ -49,6 +49,15 @@ namespace doris {
 
 using namespace std;
 
+#define RETURN_ERROR_IF_EXPR_IS_NOT_SLOTREF(expr) \
+    do { \
+            const Expr* expr_without_cast = Expr::expr_without_cast(expr); \
+            if (expr_without_cast->node_type() != TExprNodeType::SLOT_REF) { \
+                return Status::InternalError("build disjuncts failed: child is not slot ref"); \
+            }
+        } \
+    } while (false)
+
 std::string ExtLiteral::value_to_string() {
     std::stringstream ss;
     switch (_type) {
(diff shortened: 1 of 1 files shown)


wutiangan (dismissed):

wutiangan (approved):

morningman (approved):
LGTM

be/src/exec/es/es_predicate.cpp
```diff
@@ -49,6 +49,14 @@ namespace doris {
 
 using namespace std;
 
+#define RETURN_ERROR_IF_EXPR_IS_NOT_SLOTREF(expr) \
+    do { \
+            const Expr* expr_without_cast = Expr::expr_without_cast(expr); \
+            if (expr_without_cast->node_type() != TExprNodeType::SLOT_REF) { \
+                return Status::InternalError("build disjuncts failed: child is not slot ref"); \
```
blackfox1983, line 56: what is slot ref? add some example otherwise  it will be very confusing once this error occurs.
 e.g. slot ref(xx, xxx)

be/src/exec/es/es_predicate.cpp
```diff
@@ -235,13 +243,16 @@ Status EsPredicate::build_disjuncts_list(const Expr* conjunct) {
         if (TExprNodeType::SLOT_REF == conjunct->get_child(0)->node_type()
             || TExprNodeType::CAST_EXPR == conjunct->get_child(0)->node_type()) {
             expr = conjunct->get_child(1);
+            // process such as sub-query: select * from (select split_part(k, "_", 1) as new_field from case_replay_for_milimin) t where t.new_field > 1;
```
blackfox1983, line 246: e.... remove my name from table .... I don't want to be famous this way:(

BabySid (commented):
add more comments

PR #3815, opened 2020-06-10
Optimize the logic for getting TabletMeta from TabletInvertedIndex to reduce frequency of getting read lock 

caiconghui:
This PR is to optimize the logic for getting tabletMeta from TabletInvertedIndex to reduce frequence of getting read lock

### DIFF
--- a/fe/src/main/java/org/apache/doris/catalog/TabletInvertedIndex.java
+++ b/fe/src/main/java/org/apache/doris/catalog/TabletInvertedIndex.java
@@ -40,6 +40,7 @@ import com.google.common.collect.Table;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -271,7 +272,7 @@ public class TabletInvertedIndex {
     }
 
     public List<TabletMeta> getTabletMetaList(List<Long> tabletIdList) {
-        List<TabletMeta> tabletMetaList = Lists.newArrayList();
+        List<TabletMeta> tabletMetaList = new ArrayList<>(tabletIdList.size());
         readLock();
         try {
             for (Long tabletId : tabletIdList) {
--- a/fe/src/main/java/org/apache/doris/catalog/TabletInvertedIndex.java
+++ b/fe/src/main/java/org/apache/doris/catalog/TabletInvertedIndex.java
@@ -45,7 +45,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
-import java.util.concurrent.locks.StampedLock;
 import java.util.stream.Collectors;
 
 /*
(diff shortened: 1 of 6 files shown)


caiconghui:
for #3816

fe/src/main/java/org/apache/doris/master/ReportHandler.java
```diff
@@ -388,21 +389,26 @@ private static void sync(Map<Long, TTablet> backendTablets, ListMultimap<Long, L
                 List<Long> tabletIds = tabletSyncMap.get(dbId);
                 LOG.info("before sync tablets in db[{}]. report num: {}. backend[{}]",
                          dbId, tabletIds.size(), backendId);
-
-                for (Long tabletId : tabletIds) {
-                    long tableId = invertedIndex.getTableId(tabletId);
+                List<TabletMeta> tabletMetaList = invertedIndex.getTabletMetaList(tabletIds);
```
morningman, line 392: Have you tested its performance? There may be millions of tablet in `tabletIds`. I am not sure how long it will take to process that much of tablets.

caiconghui: cost for getting a list of size for 1000000 from map  is about 40 ~ 50ms, here we only preocess tablets in one be and which need to sync, so i think  1000000 for testing is large enough

morningman: OK

fe/src/main/java/org/apache/doris/catalog/TabletInvertedIndex.java
```diff
@@ -298,57 +262,28 @@ public Long getTabletIdByReplica(long replicaId) {
...
-            return NOT_EXIST_VALUE;
+            return tabletMetaMap.get(tabletId);
         } finally {
             readUnlock();
         }
     }
 
-    public int getEffectiveSchemaHash(long tabletId) {
-        // always get old schema hash(as effective one)
+    public List<TabletMeta> getTabletMetaList(List<Long> tabletIdList) {
```
morningman, line 274: It is very error-prone to result a list will null in it.
How about using `getOrDefault()` method, and return an `NonExistTabletMeta` is not found?
And `NonExistTabletMeta` can be a `final static` member with all id as `NON_EXIST_VALUE` in it.

In this way, we don't have to check null each time we visit the returned list.

fe/src/main/java/org/apache/doris/master/ReportHandler.java
```diff
@@ -388,21 +389,26 @@ private static void sync(Map<Long, TTablet> backendTablets, ListMultimap<Long, L
...
 
-                    long partitionId = invertedIndex.getPartitionId(tabletId);
+                    long partitionId = tabletMeta.getPartitionId();
                     Partition partition = olapTable.getPartition(partitionId);
                     if (partition == null) {
                         continue;
                     }
 
-                    long indexId = invertedIndex.getIndexId(tabletId);
+                    long indexId = tabletMeta.getPartitionId();
```
morningman, line 411:
```suggestion
                    long indexId = tabletMeta.getIndexId();
```

Please double check again for similar errors.

morningman (changes requested):

fe/src/main/java/org/apache/doris/catalog/TabletInvertedIndex.java
caiconghui: ok

morningman (approved):
LGTM

PR #3820, opened 2020-06-10
[Spill To Disk] Analytic_Eval_Node Support Spill Disk and Del Some Unless Code

HappenLee:
1. Add enable spilling in query option, upport spill disk in Analytic_Eval_Node, FE can open enable spilling by

         set enable_spilling = true;

Now, Sort Node and Analytic_Eval_Node can spill to disk.
2. Delete merge merge_sorter code we do not use now.
3. Replace buffered_tuple_stream by buffered_tuple_stream2 in Analytic_Eval_Node and support spill to disk. Delete the useless code of buffered_block_mgr and buffered_tuple_stream.
4. Add DataStreamRecvr Profile. Move the counter belong to DataStreamRecvr from fragment to DataStreamRecvr Profile to make clear of Running Profile.


chaoyli (changes requested):

be/src/runtime/buffered_block_mgr2.cc
```diff
@@ -210,8 +210,7 @@ BufferedBlockMgr2::BufferedBlockMgr2(RuntimeState* state, TmpFileMgr* tmp_file_m
     _max_block_size(block_size),
     // Keep two writes in flight per scratch disk so the disks can stay busy.
     _block_write_threshold(tmp_file_mgr->num_active_tmp_devices() * 2),
-    // _disable_spill(state->query_ctx().disable_spilling),
-    _disable_spill(false),
+    _disable_spill(state->disable_spill()),
```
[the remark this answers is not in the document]
HappenLee: ok，i will change the bool var

chaoyli (approved):

gensrc/thrift/PaloInternalService.thrift
```diff
@@ -136,6 +136,8 @@ struct TQueryOptions {
   // see BE config `max_pushdown_conditions_per_column` for details
   // if set, this will overwrite the BE config.
   30: optional i32 max_pushdown_conditions_per_column
+  // whether enable spilling to disk
+  31: optional bool enable_spilling = false;
```
kangkaisen, line 140: Please add this variable to doc.

be/src/runtime/data_stream_recvr.cc
```diff
@@ -423,7 +428,7 @@ void DataStreamRecvr::close() {
     _mgr = NULL;
     _merger.reset();
     _mem_tracker->close();
-    _mem_tracker->unregister_from_parent();
```
kangkaisen, line 426: Why remove this code?

be/src/runtime/data_stream_recvr.cc
```diff
@@ -360,9 +360,14 @@ DataStreamRecvr::DataStreamRecvr(
             _row_desc(row_desc),
             _is_merging(is_merging),
             _num_buffered_bytes(0),
-            _profile(profile),
             _sub_plan_query_statistics_recvr(sub_plan_query_statistics_recvr) {
-    _mem_tracker.reset(new MemTracker(-1, "DataStreamRecvr", parent_tracker));
+    _profile.reset(new RuntimeProfile(nullptr, "DataStreamRecvr"));
+    profile->add_child(_profile.get(), true, nullptr);
+
+    // TODO: Now the parent tracker may cause problem when we need spill to disk, so we
```
kangkaisen, line 367: What's concrete problem will happen?

be/src/exec/exchange_node.cpp
```diff
@@ -110,7 +110,7 @@ Status ExchangeNode::close(RuntimeState* state) {
     if (_stream_recvr != NULL) {
         _stream_recvr->close();
     }
-    _stream_recvr.reset();
+    // _stream_recvr.reset();
```
kangkaisen, line 113: directly remove it, not comment.

be/src/exec/analytic_eval_node.cpp
```diff
@@ -195,9 +194,19 @@ Status AnalyticEvalNode::open(RuntimeState* state) {
...
     //RETURN_IF_ERROR(QueryMaintenance(state));
     RETURN_IF_ERROR(child(0)->open(state));
-    // RETURN_IF_ERROR(state->block_mgr()->RegisterClient(2, mem_tracker(), state, &client_));
-    _input_stream.reset(new BufferedTupleStream(state, child(0)->row_desc(), state->block_mgr()));
-    RETURN_IF_ERROR(_input_stream->init(runtime_profile()));
+    RETURN_IF_ERROR(state->block_mgr2()->register_client(2, mem_tracker(), state, &client_));
+    _input_stream.reset(new BufferedTupleStream2(state, child(0)->row_desc(), state->block_mgr2(), client_, false, true));
+    RETURN_IF_ERROR(_input_stream->init(id(), runtime_profile(), true));
+
+    bool got_read_buffer;
```
kangkaisen, line 201: would better give a better name.

gensrc/thrift/PaloInternalService.thrift
HappenLee: The variable already in the doc。
image

I will change some detail of the doc.

be/src/exec/exchange_node.cpp
HappenLee: ok, I will remove it

be/src/exec/analytic_eval_node.cpp
HappenLee: @kangkaisen Hello, I could not think of a better variable name. Could you please give me a suggestion. Thank you.

be/src/runtime/data_stream_recvr.cc
HappenLee: To be honest, the problem is a little tricky. When the data to be spilled to disk which be read from the underlying ```EXCHANE NODE```. The process is asynchronous. There are two different threads involved

1. ```DataStreamRecv``` will continuously obtain the batch through RPC call and write it to the local queue. In this process, the memory consumption of the batch will be recorded in ```MemTracker```.

2. When we need to spill to disk, ```BlockMgr``` will try to use all the memory to speed up the query. It's going to keep calling ```try_ consume``` to get memory from ```MemTracker```.

3. Once ```try_ consume``` failed. The ```SortNode``` will try to drop the disk. At this time, the memory usage has approached the threshold. (at this time, the memory usage does not exceed the ```MemTracker``` limit)

4. If thread switching occurs at this time, switch from the ```SortNode``` to ```DataStreamRecv```. Because the ```SortNode``` is performing like sort and other CPU time consuming operations, it does not retrieve the data from the ```Exchange Node``` in time. ```DataStreamRecv``` writes data to the queue and records memory consumption. (memory usage exceeds ```MemTracker``` limit at this time)

5. When the thread switches back to ```SortNode```. The check found that ```MemTracker``` exceeded the memory limit, and the query failed.



Although the queue in ```DataStreamRecv``` has memory limitation, it only records the size before Protobuffer deserialization. ```MemTracker``` records the size after deserialization. Because Protobuffer compresses the data, there is a gap of three times in my test scenario.


At present, My idea is to use this part of memory independently and use a running profile to record the memory usage by ```DataStreamRecv```, so as to avoid this problem temporarily. How can we better solve the problem later? I want to have a further discussion in the community

gensrc/thrift/PaloInternalService.thrift
kangkaisen: OK

be/src/runtime/data_stream_recvr.cc
kangkaisen: The very detailed explanation, Thank you!

be/src/exec/analytic_eval_node.cpp
kangkaisen: is_got_buffer or has_got_buffer?

HappenLee: ok

morningman (approved):
LGTM

Issue #3836, opened 2020-06-11
[BUG] Document site got pushed into asf-site even if a error occurred during travis building 


hffariel:
Document site got pushed into asf-site even if a error occurred during travis building.

## Caused by

Commands in `after_script`  section will always be executed even when the build is failed.

## Solve the issue

Move the commands under `after_script` to `after_success` section.

PR #3838, opened 2020-06-11
stop travis building when an error occurred [master] 

hffariel:
stop travis building when an error occurred #3836


morningman (approved):
LGTM

PR #3844, opened 2020-06-11
[UDF] StringVal::operator==(): avoid pass NULL to memcmp()

vagetablechicken:
Ref https://github.com/apache/incubator-doris/issues/3843

If we exec "StringVal(len=0, ptr="") == StringVal(len=0,ptr=NULL)", it will pass NULL ptr to memcmp(). It should be avoided.

### DIFF
--- a/be/src/udf/udf.h
+++ b/be/src/udf/udf.h
@@ -644,7 +644,7 @@ struct StringVal : public AnyVal {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


chaoyli (approved):

chaoyli:
You can add some commit info to describe the solution.

vagetablechicken:
> You can add some commit info to describe the solution.

Added.

PR #3846, opened 2020-06-11
fix for be server crash which throwing syntax error  when parse json …

sdgshawn:
fix for be server crash which throwing syntax error  when parse json from kafka message


be/src/exprs/json_functions.cpp
```diff
@@ -291,7 +291,7 @@ rapidjson::Value* JsonFunctions::get_json_array_from_parsed_json (
     }
 
     rapidjson::Value* root = match_value(parsed_paths, document, mem_allocator, true);
-    if (root == document) {// not found
+    if (root==nullptr || root == document) {// not found
```
imay, line 294:
```suggestion
    if (root == nullptr || root == document) {// not found
```

imay (approved):
LGTM

PR #3848, opened 2020-06-12
[Bug] fix uninitialized member vars

vagetablechicken:
Ref https://github.com/apache/incubator-doris/issues/3821

This fix is based on UBSAN unit test. So if we create & use class obj in a different way, may have `runtime error: load of value XX, which is not a valid value for type 'YYY'` warnings again.

Unit test should build in DEBUG or XXSAN mode(at lease DEBUG). RELEASE mode will add `-DNDEBUG`, turn off dchecks/asserts/debug.


kangkaisen (approved):
+1, LGTM

PR #3850, opened 2020-06-12
[FE] Prohibit pointing helper to itself when starting FE

morningman:
When starting FE with `start_fe.sh --helper xxx` command, do not allow to
point helper to FE itself. Because this is meaningless and may cause some
confusing problems.


EmmyMiao87 (approved):
LGTM

PR #3851, opened 2020-06-12
[Enhancement] Replace N/A use NULL

### LINKED ISSUE #3852: [Enhancement] Replace N/A use NULL
Before, Doris use N/A to indicate not available value.
It's better to use NULL to indicate it in Database.

chaoyli:
Before, Doris use N/A to indicate not available value.
It's better to use \N to indicate it in Database.
\N will be encoded to NULL before send to client.

### DIFF
--- a/fe/src/main/java/org/apache/doris/alter/RollupJob.java
+++ b/fe/src/main/java/org/apache/doris/alter/RollupJob.java
@@ -33,6 +33,7 @@ import org.apache.doris.catalog.Tablet;
 import org.apache.doris.catalog.TabletInvertedIndex;
 import org.apache.doris.catalog.TabletMeta;
 import org.apache.doris.common.Config;
+import org.apache.doris.common.FeConstants;
 import org.apache.doris.common.FeMetaVersion;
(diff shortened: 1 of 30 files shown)


morningman:
Actually, in MySQL protocal, the `null` is represented as `(byte) (251 & 0xff)`.

https://github.com/apache/incubator-doris/blob/88a5429165c65690eb443c017ed0b3bd7d786330/fe/src/main/java/org/apache/doris/mysql/MysqlSerializer.java#L64-L66

https://github.com/apache/incubator-doris/blob/88a5429165c65690eb443c017ed0b3bd7d786330/be/src/util/mysql_row_buffer.cpp#L270-L281

So If we want to make it compatible with real `null`, better to follow the protocol.

I have a suggestion. Currently you are modifying the result set related to `Show`. The current ShowResultSet uses `String`. We can encapsulate a `NullableString`. In this way, when sending ShowResultSet, we can judge and return true `null`.

It is not recommended to use `null` directly, because that will cause some results forwarded from the follower FE to not be able to be serialized to thrift.

morningman (approved):
LGTM

PR #3860, opened 2020-06-14
[Compaction] Resolve the `-230` error by saving latest read version of a tablet

### LINKED ISSUE #3859: [Proposal] Resolve error "fail to init reader.res=-230"
**Describe the bug**

The specific meaning of this error can be found in #3270. And in PR #3271, we try to alleviate this problem by introducing a new BE config `cumulative_compaction_skip_window_seconds`.

However, this config does not completely solve this problem, because in the following two cases, we can not accurately set the config value.

1. Query with Join

In the query with Join operation, the query engine will first perform a Build operation on the right table, and wait until the right table is built before `Open` the left table for data reading. The operation of obtaining the rowset reader according to the version only starts when it is `Open`. If the build phase is very time-consuming, it will cause the `Open` on the left to be executed long after the query starts, and the version that should be read at this time may has been merged. The time of the build phase cannot be estimated, so it is difficult to set a reasonable value for the config `cumulative_compaction_skip_window_seconds`.

2. Transaction publish delay

Each BE merges its own tablet separately. Only the published version will be merged. The `Publish` here only represents the publish of BE's own tablet, not the corresponding successful publish of the load transaction. This causes the following problems, for example:

    1. Assume that the version corresponding to a load transaction is 10.
    2. In this load transaction, there are three replicas of a tablet, one of which is published successfully at 10:00, and the other two replicas are published successfully at 10:10. So the finish time of the entire load transaction is 10:10. In other words, users can query the 10 version of the data from 10:10. And because there is already one replica published at 10:00. That replica may have received the 11 version of the publish task at 10:02. Then at 10:05, a compaction is performed to form the [0-11] version. Then when 10:10, the user uses the 10 version to query, that replica will report an error "-230".

The above two situations can be alleviated by increasing the `cumulative_compaction_skip_window_seconds` configuration, but
    
    1. The larger the config may cause the version backlog, and the compaction is not timely.
    2. This is just an empirical value and cannot completely solve this problem.

**Resolution**

The essence of the problem is that before we read the specified version, the specified version has been merged. We can control the timing of compaction by recording the current largest read version in the tablet.

In the tablet, the current latest read version is recorded in field `latest_read_version`. This version is set every time the tablet is read, and only the largest version currently read is recorded.

When compaction selects a version for merging, only data whose version is less than or equal to `latest_read_version` is selected. This will ensure that if a version(or a version larger than it), has not been read, the version will not be compacted.

Of course, under this strategy, if there has been no read request, the compaction will not be triggered. So we need to set a threshold to prevent the data from never being merged. We can use `cumulative_compaction_skip_window_seconds` to configure this threshold. That is, when the threshold time is exceeded, `latest_read_version` will be ignored and the version merge will start.

This strategy may have problems in the following scenarios: the query frequency is low, such as once every 10 minutes, but the load frequency is high, such as 10 times per minute. In this case, if the `cumulative_compaction_skip_window_seconds` setting is large, it may accumulate large number of versions, eg, 100 versions. If the setting is small, the `-230` error may still be generated. However, in terms of actual operation and maintenance experience, setting `cumulative_compaction_skip_window_seconds` to 5 minutes can basically solve most problems.

morningman:
Fix #3859

This CL mainly changes:
1. add a new field `_latest_read_version` for tablet on BE to save the largest version which has
been read by query.
2. Modify the cumulative compaction logic. When select rowset to merge, consider both
`cumulative_compaction_skip_window_seconds` and `_latest_read_version`.
3. Add some docs for config
4. Remove unused config `cumulative_compaction_budgeted_bytes`

### DIFF
--- a/be/src/olap/tablet.cpp
+++ b/be/src/olap/tablet.cpp
@@ -60,7 +60,8 @@ Tablet::Tablet(TabletMetaSharedPtr tablet_meta, DataDir* data_dir) :
         _last_base_compaction_failure_millis(0),
         _last_cumu_compaction_success_millis(0),
         _last_base_compaction_success_millis(0),
-        _cumulative_point(kInvalidCumulativePoint) {
+        _cumulative_point(kInvalidCumulativePoint),
+        _latest_read_version(0) {
     _rs_graph.construct_rowset_graph(_tablet_meta->all_rs_metas());
 }
 
@@ -852,12 +853,11 @@ TabletInfo Tablet::get_tablet_info() const {
     return TabletInfo(tablet_id(), schema_hash(), tablet_uid());
 }
 
-void Tablet::pick_candicate_rowsets_to_cumulative_compaction(int64_t skip_window_sec,
-                                                             std::vector<RowsetSharedPtr>* candidate_rowsets) {
-    int64_t now = UnixSeconds();
+void Tablet::pick_candicate_rowsets_to_cumulative_compaction(
+        std::vector<RowsetSharedPtr>* candidate_rowsets) {
     ReadLock rdlock(&_meta_lock);
     for (auto& it : _rs_version_map) {
-        if (it.first.first >= _cumulative_point && (it.second->creation_time() + skip_window_sec < now)) {
+        if (it.first.first >= _cumulative_point) {
             candidate_rowsets->push_back(it.second);
         }
     }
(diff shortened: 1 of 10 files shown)


be/src/olap/tablet.cpp
```diff
@@ -852,12 +853,11 @@ TabletInfo Tablet::get_tablet_info() const {
...
 
-void Tablet::pick_candicate_rowsets_to_cumulative_compaction(int64_t skip_window_sec,
-                                                             std::vector<RowsetSharedPtr>* candidate_rowsets) {
-    int64_t now = UnixSeconds();
+void Tablet::pick_candicate_rowsets_to_cumulative_compaction(
+        std::vector<RowsetSharedPtr>* candidate_rowsets) {
     ReadLock rdlock(&_meta_lock);
     for (auto& it : _rs_version_map) {
-        if (it.first.first >= _cumulative_point && (it.second->creation_time() + skip_window_sec < now)) {
+        if (it.first.first >= _cumulative_point) {
```
morningman, line 860: For reviewer: Remove the `skip_window_sec` check here, it will be checked later in cumulative compaction process.

xunfeng1980:
why close ?

Issue #3832, opened 2020-06-11
[Bug]BE core due to invalid bitmap aggragate column


WingsGo:
**Describe the bug**
If we create a table with column type is bitmap and aggregate type is replace.BE will core during load data.
```
Thu Jun 11 15:23:42 CST 2020: *** Aborted at 1591860222 (unix time) try "date -d @1591860222" if you are using GNU date ***
Thu Jun 11 15:23:42 CST 2020: PC: @           0xe974e1 doris::Field::consume()
Thu Jun 11 15:23:42 CST 2020: *** SIGSEGV (@0x18) received by PID 28285 (TID 0x7fb89f22b700) from PID 24; stack trace: ***
Thu Jun 11 15:23:42 CST 2020: @     0x7fb9222a9250 (unknown)
Thu Jun 11 15:23:42 CST 2020: @           0xe974e1 doris::Field::consume()
Thu Jun 11 15:23:42 CST 2020: @           0xec990f doris::MemTable::insert()
Thu Jun 11 15:23:42 CST 2020: @          0x155ddd6 doris::DeltaWriter::write()
Thu Jun 11 15:23:42 CST 2020: @           0xfedd76 doris::TabletsChannel::add_batch()
Thu Jun 11 15:23:42 CST 2020: @           0xfea5c6 doris::LoadChannel::add_batch()
Thu Jun 11 15:23:42 CST 2020: @           0xfe5bc1 doris::LoadChannelMgr::add_batch()
Thu Jun 11 15:23:42 CST 2020: @          0x106a5b5 _ZZN5doris20PInternalServiceImplIN4palo16PInternalServiceEE23tablet_writer_add_batchEPN6google8protobuf13RpcControllerEPKNS_28PTabletWriterAddBatchRequestEPNS_27PTabletWriterAddBatchResultEPNS5_7ClosureEENKUlvE_clEv
Thu Jun 11 15:23:42 CST 2020: @           0xf80105 doris::PriorityThreadPool::work_thread()
Thu Jun 11 15:23:42 CST 2020: @          0x1a2096d thread_proxy
Thu Jun 11 15:23:42 CST 2020: @     0x7fb92205fdc5 start_thread
Thu Jun 11 15:23:42 CST 2020: @     0x7fb92236b73d __clone

```

**To Reproduce**
1. create a table
```
CREATE TABLE `xm_stat_addcart_goods_user_minute` (
  `stat_date` date NOT NULL DEFAULT "1970-01-01" ,
[...]

PR #3866, opened 2020-06-15
[Enhancement]Add Field check to avoid create invalid field

WingsGo:
This CL includes:
1. add field check to avoid FE's invalid request cause BE coredump
2. use clang-format to format related code


WingsGo:
For #3832

WingsGo:
In order to make review easier, I revert some formatted code, after logic is right, I will add it later.

kangkaisen:
@WingsGo Hi, it seems code format code still exist in this PR.

WingsGo:
> @WingsGo Hi, it seems code format code still exist in this PR.

Yes, It's because the code file includes some core function,so I have to revert it one by one, but it's much more less than before~0v0

Issue #3868, opened 2020-06-15
BE crash when query contains HLL column


Youngwb:
**Describe the bug**
BE will core when user query table which contains HLL data and it's write by old version(maybe 0.10),  This bug happens occasionally, not always

**Screenshots**
![image](https://user-images.githubusercontent.com/9495145/84635030-3a995700-af25-11ea-9b12-5b7e82d445b6.png)

**Additional context**
According to the core file, the reason is that when HyperLogLog initialized with Slice, the Slice's data pointed to an invalid address, causing be to crash.  And the address 0x2c87f2031 is used to Identifies whether the cell is NULL

PR #1908, opened 2019-09-27
Make HLL be able to handle invalid data

imay:
In this change list
1. validate HLL column when loading data, if data is invalid, this row
will be filtered.
2. seems as empty HLL when serializing invalid type of HLL data, with
this change, all ingested data will be valid.
3. seems as empty HLL when deserializing nullptr or invalid type of HLL data.
With this change, dirty data can be handled normally.
4. rename function empty_hll to hll_empty.
5. disable memtable_flush_execute_test because this will fails
sometimes. When tearing down, some thread is not joined, and they will
visit destroyed resource, which is invalid.

PR #3869, opened 2020-06-15
Fix BE crash when query contains HLL column

Youngwb:
According #3868,  BE will core because of invalid address, This bug happens occasionally because most of the time HyperLogLog's data are nullptr, which does not cause crashes, but sometimes this value is invalid address, which causes crashes during deserialize.
   Whether slice data is nullptr or invalid address is determined by the RowBlock, however the RowBlock did not use `memset` when it was initialized, so invalid data may exist.

### DIFF
--- a/be/src/olap/aggregate_func.h
+++ b/be/src/olap/aggregate_func.h
@@ -449,15 +449,22 @@ struct AggregateFuncTraits<OLAP_FIELD_AGGREGATION_REPLACE_IF_NOT_NULL, OLAP_FIEL
 template <>
 struct AggregateFuncTraits<OLAP_FIELD_AGGREGATION_HLL_UNION, OLAP_FIELD_TYPE_HLL> {
     static void init(RowCursorCell* dst, const char* src, bool src_null, MemPool* mem_pool, ObjectPool* agg_pool) {
-        DCHECK_EQ(src_null, false);
         dst->set_not_null();
 
-        auto* src_slice = reinterpret_cast<const Slice*>(src);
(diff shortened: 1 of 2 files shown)


chaoyli (approved):

be/src/olap/aggregate_func.h
```diff
@@ -449,15 +449,22 @@ struct AggregateFuncTraits<OLAP_FIELD_AGGREGATION_REPLACE_IF_NOT_NULL, OLAP_FIEL
...
-        DCHECK_EQ(src_null, false);
         dst->set_not_null();
 
         auto* src_slice = reinterpret_cast<const Slice*>(src);
         auto* dst_slice = reinterpret_cast<Slice*>(dst->mutable_cell_ptr());
 
+        // Because of history bug, we ingest some NULL HLL data in storage,
+        // which slice data is nullptr or invalid address,
+        // we must handle this case to avoid process crash.
+        HyperLogLog* hll = nullptr;
```
kangkaisen, line 460: I don't agree simply handle null here. If we do this way, we must handle null forever, which is terrible.

At least，We should find a way to convert the invalid null HLL to empty and Remove null handle in future version.

Youngwb: I think this pr is a fix for #1908, which handles the scenario where HLL data is nullptr but does not handle dirty data. And according to it's comments,  it suggested to remove these check for null HLL column later.

kangkaisen: If the data has dirty, I think we should repair it. 
You could reload the table data or develop a tool to repair dirty data. 
I think we shouldn't make code to be compatible with dirty data. which means we forever couldn't remove the compatible code.

PR #3872, opened 2020-06-15
[Bug] Fix bug that BE crash when doing Insert Operation

### LINKED ISSUE #3871: [Bug] BE crash when doing insert stmt
**Describe the bug**
BE crash with coredump like:

```
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x0000000001045667 in release (bytes=8192, this=0x633b7d40) at /var/doris/palo-core/be/src/runtime/mem_tracker.h:221
#2  doris::RowBatch::clear (this=0x8fc6bce0) at /var/doris/palo-core/be/src/runtime/row_batch.cpp:294
#3  0x00000000010459dd in clear (this=0x8fc6bce0) at /var/doris/palo-core/be/src/common/object_pool.h:54
#4  doris::RowBatch::~RowBatch (this=0x8fc6bce0, __in_chrg=<optimized out>) at /var/doris/palo-core/be/src/runtime/row_batch.cpp:301
#5  0x0000000001045a01 in doris::RowBatch::~RowBatch (this=0x8fc6bce0, __in_chrg=<optimized out>) at /var/doris/palo-core/be/src/runtime/row_batch.cpp:302
#6  0x00000000014a0299 in operator() (this=<optimized out>, __ptr=<optimized out>) at /var/doris/doris-toolchain/gcc730/include/c++/7.3.0/bits/unique_ptr.h:78
#7  ~unique_ptr (this=0x7070ec98, __in_chrg=<optimized out>) at /var/doris/doris-toolchain/gcc730/include/c++/7.3.0/bits/unique_ptr.h:268
#8  doris::stream_load::NodeChannel::~NodeChannel (this=0x7070ec00, __in_chrg=<optimized out>) at /var/doris/palo-core/be/src/exec/tablet_sink.cpp:41
#9  0x00000000014a53f2 in ~SpecificElement (this=0x77645500, __in_chrg=<optimized out>) at /var/doris/palo-core/be/src/common/object_pool.h:75
#10 doris::ObjectPool::SpecificElement<doris::stream_load::NodeChannel>::~SpecificElement (this=0x77645500, __in_chrg=<optimized out>) at /var/doris/palo-core/be/src/common/object_pool.h:76
#11 0x0000000001052efd in clear (this=0x6a213220) at /var/doris/palo-core/be/src/common/object_pool.h:54
#12 ~ObjectPool (this=0x6a213220, __in_chrg=<optimized out>) at /var/doris/palo-core/be/src/common/object_pool.h:37
#13 std::_Sp_counted_ptr<doris::ObjectPool*, (__gnu_cxx::_Lock_policy)2>::_M_dispose (this=<optimized out>) at /var/doris/doris-toolchain/gcc730/include/c++/7.3.0/bits/shared_ptr_base.h:376
#14 0x000000000104b591 in _M_release (this=0x6a212fa0) at /var/doris/doris-toolchain/gcc730/include/c++/7.3.0/bits/shared_ptr_base.h:154
#15 ~__shared_count (this=0x891de310, __in_chrg=<optimized out>) at /var/doris/doris-toolchain/gcc730/include/c++/7.3.0/bits/shared_ptr_base.h:684
#16 ~__shared_ptr (this=0x891de308, __in_chrg=<optimized out>) at /var/doris/doris-toolchain/gcc730/include/c++/7.3.0/bits/shared_ptr_base.h:1123
#17 ~shared_ptr (this=0x891de308, __in_chrg=<optimized out>) at /var/doris/doris-toolchain/gcc730/include/c++/7.3.0/bits/shared_ptr.h:93
#18 doris::RuntimeState::~RuntimeState (this=0x891de300, __in_chrg=<optimized out>) at /var/doris/palo-core/be/src/runtime/runtime_state.cpp:128
#19 0x000000000103dcc6 in checked_delete<doris::RuntimeState> (x=0x891de300) at /var/doris/thirdparty/installed/include/boost/core/checked_delete.hpp:34
#20 ~scoped_ptr (this=0x88082ce0, __in_chrg=<optimized out>) at /var/doris/thirdparty/installed/include/boost/smart_ptr/scoped_ptr.hpp:89
#21 doris::PlanFragmentExecutor::~PlanFragmentExecutor (this=0x88082b70, __in_chrg=<optimized out>) at /var/doris/palo-core/be/src/runtime/plan_fragment_executor.cpp:62
#22 0x0000000000fd39ed in doris::FragmentExecState::~FragmentExecState (this=0x88082b00, __in_chrg=<optimized out>) at /var/doris/palo-core/be/src/runtime/fragment_mgr.cpp:175
```

**Debug**

In some abnormal situations, performing Insert to load will cause BE to crash. The reasons for troubleshooting are as follows:

* FE side:

https://github.com/apache/incubator-doris/blob/2211cb0ee0fcd23d4fd2445494aba6cf1a020987/fe/src/main/java/org/apache/doris/qe/Coordinator.java#L475-L489

During the execution of `execState.execRemoteFragmentAsync()`, if an RPC error occurs, if the corresponding BE is down, an exception will be thrown directly instead of returning the error via `Future<PExecPlanFragmentResult>`. This time, the Coordinator will not proceed with the subsequent `Cancel` operation.

After an exception is thrown, the thread stack returns to `handleInsertStmt()` and directly returns the user error message. Insert failed. So far, FE has no further processing.

* BE side

BE receives the execution plan of Insert and calls `_sink->open` in `PlanFragmentExecutor::open_internal()` to open `TabletSink`.

https://github.com/apache/incubator-doris/blob/2211cb0ee0fcd23d4fd2445494aba6cf1a020987/be/src/runtime/plan_fragment_executor.cpp#L272-L292

The Open method of `TabletSink` will open all related NodeChannels via RPC, but because some BEs problem, some of NodeChannels fail to open. An error message appears in the BE log:

```
tablet open failed, load_id=4e2f4d6076cc4692-b8f26cc924b6be0d, txn_id144942945, node=10.xx.xx.xx:8060, errmsg=failed to open tablet writer
```

However, because the majority of NodeChannels were successfully opened, TabletSink Open operations returned success.

Next, in `PlanFragmentExecutor::open_internal()` will call `get_next_internal()` to start fetching data, because Insert already failed at this time, so it returns failure here. **And there are bugs in the later `update_status(status)` method**:

https://github.com/apache/incubator-doris/blob/2211cb0ee0fcd23d4fd2445494aba6cf1a020987/be/src/runtime/plan_fragment_executor.cpp#L485-L503

Line 493: `if (_status.ok())` should be `if (!_status.ok())`. This error caused the `_status` variable not to be updated. This will cause the NodeChannel to be closed instead of being canceled when the TabletSink is finally closed.

The normal NodeChannel close operation will send the last batch of RowBatch it holds and destroy the RowBatch object. And because some NodeChannels were not opened normally, they will not be closed normally. **The RowBatch held by these NodeChannels will not be destroyed**.

After the execution of the entire plan is completed, the PlanFragmentExecutor's destruction process is entered. The destructor call chain is as follows:

```
PlanFragmentExecutor
|--- RuntimeState
     |--- RuntimeProfile
     |--- ObjectPool
          |--- NodeChannel
               |--- RowBatch
                    |--- MemTracker->release()
                         |--- profile->_consumption->add(-bytes)
```

Note that the RuntimeProfile will be first destructed in RuntimeState, which will cause the `_consumption` object that has been destructed to be called when destructing the RowBatch in the NodeChannel, which will eventually cause BE crash.

**The whole process has the following problems**:

1. The bug in `update_status(status)` caused the NodeChannel to not be canceled correctly, which caused problems in the final destruction. (If Cancel of NodeChannel is called in advance, RowBatch will be destructed in advance).
2. When the FE Coordinator executes `execRemoteFragmentAsync()`, if it finds an RPC error, it should return a Future with an error code, continue the process afterwards, and actively call Cancel().
3. The `_status` in RuntimeState has no lock protection, which may cause some potential problems.

morningman:
Fix: #3871

Mainly change:
1. Fix the bug in `update_status(status)` of `PlanFragmentExecutor`.
2. When the FE Coordinator executes `execRemoteFragmentAsync()`, if it finds an RPC error, return a Future with an error code instead of exception.
3. Protect the `_status` in RuntimeState with lock
4. Move the `_runtime_profile` of RuntimeState before the `_obj_pool`, so that the profile will be
deconstructed after the object pool.
5. Remove the unused `ObjectPool` param in RuntimeProfile constructor. If I don't remove it,
RuntimeProfile will depends on the `_obj_pool` in RuntimeProfile.

### DIFF
--- a/be/src/runtime/plan_fragment_executor.cpp
+++ b/be/src/runtime/plan_fragment_executor.cpp
@@ -323,7 +323,12 @@ Status PlanFragmentExecutor::open_internal() {
     {
         SCOPED_TIMER(profile()->total_time_counter());
         collect_query_statistics();
-        Status status = _sink->close(runtime_state(), _status);
+        Status status;
+        {
+            boost::lock_guard<boost::mutex> l(_status_lock);
+            status = _status;
+        }
+        status = _sink->close(runtime_state(), status);
         RETURN_IF_ERROR(status);
     }
 
@@ -489,8 +494,7 @@ void PlanFragmentExecutor::update_status(const Status& status) {
 
     {
         boost::lock_guard<boost::mutex> l(_status_lock);
-
-        if (_status.ok()) {
+        if (!_status.ok()) {
             if (status.is_mem_limit_exceeded()) {
                 _runtime_state->set_mem_limit_exceeded(status.get_error_msg());
             }
@@ -547,7 +551,12 @@ void PlanFragmentExecutor::close() {
 
         if (_sink.get() != NULL) {
             if (_prepared) {
-                _sink->close(runtime_state(), _status);
+                Status status;
+                {
+                    boost::lock_guard<boost::mutex> l(_status_lock);
+                    status = _status;
+                }
+                _sink->close(runtime_state(), status);
             } else {
                 _sink->close(runtime_state(), Status::InternalError("prepare failed"));
             }
--- a/fe/src/main/java/org/apache/doris/qe/Coordinator.java
+++ b/fe/src/main/java/org/apache/doris/qe/Coordinator.java
@@ -52,6 +52,7 @@ import org.apache.doris.planner.SetOperationNode;
 import org.apache.doris.planner.UnionNode;
 import org.apache.doris.proto.PExecPlanFragmentResult;
 import org.apache.doris.proto.PPlanFragmentCancelReason;
+import org.apache.doris.proto.PStatus;
 import org.apache.doris.qe.QueryStatisticsItem.FragmentInstanceInfo;
 import org.apache.doris.rpc.BackendServiceProxy;
 import org.apache.doris.rpc.RpcException;
@@ -1469,8 +1470,42 @@ public class Coordinator {
             try {
                 return BackendServiceProxy.getInstance().execPlanFragmentAsync(brpcAddress, rpcParams);
             } catch (RpcException e) {
-                SimpleScheduler.addToBlacklist(backend.getId());
-                throw e;
+                // DO NOT throw exception here, return a complete future with error code,
+                // so that the following logic will cancel the fragment.
+                Future<PExecPlanFragmentResult> future = new Future<PExecPlanFragmentResult>() {
+                    @Override
+                    public boolean cancel(boolean mayInterruptIfRunning) {
+                        return false;
+                    }
+
+                    @Override
+                    public boolean isCancelled() {
+                        return false;
+                    }
+
+                    @Override
+                    public boolean isDone() {
+                        return true;
+                    }
+
+                    @Override
+                    public PExecPlanFragmentResult get() throws InterruptedException, ExecutionException {
+                        PExecPlanFragmentResult result = new PExecPlanFragmentResult();
+                        PStatus pStatus = new PStatus();
+                        pStatus.error_msgs.add(e.getMessage());
+                        // use THRIFT_RPC_ERROR so that this BE will be added to the blacklist later.
+                        pStatus.status_code = TStatusCode.THRIFT_RPC_ERROR.getValue();
+                        result.status = pStatus;
+                        return result;
+                    }
+
+                    @Override
+                    public PExecPlanFragmentResult get(long timeout, TimeUnit unit)
+                            throws InterruptedException, ExecutionException, TimeoutException {
+                        return get();
+                    }
+                };
+                return future;
             }
         }
 
--- a/be/src/exec/exec_node.cpp
+++ b/be/src/exec/exec_node.cpp
@@ -552,7 +552,7 @@ void ExecNode::try_do_aggregate_serde_improve() {
 void ExecNode::init_runtime_profile(const std::string& name) {
     std::stringstream ss;
     ss << name << " (id=" << _id << ")";
-    _runtime_profile.reset(new RuntimeProfile(_pool, ss.str()));
+    _runtime_profile.reset(new RuntimeProfile(ss.str()));
     _runtime_profile->set_metadata(_id);
 }
 
--- a/be/src/exec/tablet_sink.cpp
+++ b/be/src/exec/tablet_sink.cpp
@@ -457,7 +457,7 @@ Status OlapTableSink::prepare(RuntimeState* state) {
     _num_senders = state->num_per_fragment_instances();
 
     // profile must add to state's object pool
-    _profile = state->obj_pool()->add(new RuntimeProfile(_pool, "OlapTableSink"));
+    _profile = state->obj_pool()->add(new RuntimeProfile("OlapTableSink"));
     _mem_tracker = _pool->add(new MemTracker(-1, "OlapTableSink", state->instance_mem_tracker()));
 
     SCOPED_TIMER(_profile->total_time_counter());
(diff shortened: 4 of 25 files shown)


fe/src/main/java/org/apache/doris/qe/Coordinator.java
```diff
@@ -1469,8 +1470,41 @@ public boolean isBackendStateHealthy() {
...
                 return BackendServiceProxy.getInstance().execPlanFragmentAsync(brpcAddress, rpcParams);
             } catch (RpcException e) {
-                SimpleScheduler.addToBlacklist(backend.getId());
-                throw e;
+                // Do not throw exception here.
+                // return a completed future, the caller will check the result and do the cancel process.
+                PExecPlanFragmentResult result = new PExecPlanFragmentResult();
+                PStatus pStatus = new PStatus();
+                pStatus.error_msgs.add(e.getMessage());
+                // use THRIFT_RPC_ERROR so that this BE will be added to the black list later.
```
wutiangan, line 1478:
```suggestion
                // use THRIFT_RPC_ERROR so that this BE will be added to the black list later.
```
// use THRIFT_RPC_ERROR so that this BE will be added to the blacklist later.

wutiangan, line 1478: black list  --> blacklist

kangkaisen:
> When the FE Coordinator executes execRemoteFragmentAsync(), if it finds an RPC error, return a Future with an error code instead of exception.

@morningman Hi, where is the code?

morningman:
> > When the FE Coordinator executes execRemoteFragmentAsync(), if it finds an RPC error, return a Future with an error code instead of exception.
> 
> @morningman Hi, where is the code?

Oh, sorry, I think i miss some part of the code...

fe/src/main/java/org/apache/doris/qe/Coordinator.java
morningman: done

morningman:
> > > When the FE Coordinator executes execRemoteFragmentAsync(), if it finds an RPC error, return a Future with an error code instead of exception.
> > 
> > 
> > @morningman Hi, where is the code?
> 
> Oh, sorry, I think i miss some part of the code...

Done

morningman:
hi @kangkaisen Could you help to review again?

kangkaisen:
OK

be/src/runtime/plan_fragment_executor.cpp
```diff
@@ -323,7 +323,12 @@ Status PlanFragmentExecutor::open_internal() {
     {
         SCOPED_TIMER(profile()->total_time_counter());
         collect_query_statistics();
-        Status status = _sink->close(runtime_state(), _status);
+        Status status;
+        {
+            boost::lock_guard<boost::mutex> l(_status_lock);
```
kangkaisen, line 328: If you have time. I think we would better use `std::lock_guard` and `std:: mutex ` instead of `boost::lock_guard` and `boost::mutex`.

kangkaisen (dismissed):
LGTM

be/src/runtime/plan_fragment_executor.cpp
morningman: This PR compile failed, I will fix it and change this lock too~

EmmyMiao87 (approved):
LGTM

PR #3873, opened 2020-06-15
Support materialized view schema change [#3739]

HangyuanLiu:
-  Building the materialized view function for schema_change here based on defineExpr.
-  This is a trick because the current storage layer does not support expression evaluation. 
-  count distinct materialized view will set mv_expr with to_bitmap or hll_hash.
-  count materialized view will set mv_expr with count.
-  Support to regenerate historical data when a new materialized view is created in BE。
     1. Support to_bitmap function
     2. Support hll_hash function
     3. Support count(field) function
For #3344


EmmyMiao87:
Please provide detailed commit message

HangyuanLiu:
> Please provide detailed commit message

Done

EmmyMiao87 (approved):
LGTM

PR #3881, opened 2020-06-16
[BUG]Fix broker read buffer size from input stream

### LINKED ISSUE #3879: [BUG]Broker cannot read the full length of buffer size
**Describe the bug**
Our current broker cannot read the full length of buffer from input stream when the buffer size is set larger than 128k.

This bug will cause the data size returned by pread request to be less than 128K all the time.

**To Reproduce**
Set the pread request's length larger than 128 * 1024 (128k).

xy720:
Fix #3879 
This commit fixs a bug that broker cannot read the full length of buffer size, when the buffer size is set larger than 128k.

This bug will cause the data size returned by pread request to be less than 128K all the time.

### DIFF
--- a/fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
+++ b/fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
@@ -550,19 +550,20 @@ public class FileSystemManager {
                             currentStreamOffset, offset);
                 }
             }
-            byte[] buf;
+            ByteBuffer buf;
             if (length > readBufferSize) {
-                buf = new byte[readBufferSize];
+                buf = ByteBuffer.allocate(readBufferSize);
             } else {
-                buf = new byte[(int) length];
+                buf = ByteBuffer.allocate((int) length);
             }
             try {
-                int readLength = fsDataInputStream.read(buf);
+                int readLength = readByteBufferFully(fsDataInputStream, buf);
                 if (readLength < 0) {
                     throw new BrokerException(TBrokerOperationStatusCode.END_OF_FILE,
                             "end of file reached");
                 }
-                return ByteBuffer.wrap(buf, 0, readLength);
+                logger.info("read length:" + length + ", readBufferSize:" + readBufferSize + ", return length:" + readLength);
+                return buf;
             } catch (IOException e) {
                 logger.error("errors while read data from stream", e);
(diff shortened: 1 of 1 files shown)


fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
```diff
@@ -550,19 +550,20 @@ public ByteBuffer pread(TBrokerFD fd, long offset, long length) {
...
             }
             try {
-                int readLength = fsDataInputStream.read(buf);
+                int readLength = readByteBufferFully(fsDataInputStream, buf);
                 if (readLength < 0) {
                     throw new BrokerException(TBrokerOperationStatusCode.END_OF_FILE,
                             "end of file reached");
                 }
-                return ByteBuffer.wrap(buf, 0, readLength);
+                logger.info("read length:" + length + ", readBufferSize:" + readBufferSize + ", return length:" + readLength);
```
yiguolei, line 565: Maybe too many logs here.

yiguolei (changes requested):
some comments

fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
xy720: Turn log level to debug.

fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
```diff
@@ -550,19 +550,20 @@ public ByteBuffer pread(TBrokerFD fd, long offset, long length) {
...
             }
             try {
-                int readLength = fsDataInputStream.read(buf);
+                int readLength = readByteBufferFully(fsDataInputStream, buf);
                 if (readLength < 0) {
                     throw new BrokerException(TBrokerOperationStatusCode.END_OF_FILE,
                             "end of file reached");
                 }
-                return ByteBuffer.wrap(buf, 0, readLength);
+                logger.debug("read buffer from input stream, buffer size:" + buf.capacity() + ", read length:" + readLength);
```
morningman, line 565: if (logger.isDebugEnable()) {
    logger.debug("read buffer from input stream, buffer size: {}, read length: {}", buf.capacity(), readLength);
}

xy720: done

morningman (approved):
LGTM

PR #3886, opened 2020-06-16
[Planner] Set MysqlScanNode's cardinality to avoid unexpected shuffle join

### LINKED ISSUE #3885: [Planner] Unexpected shuffle join slow down the query
**Describe the bug**
When execute sql with a olap table join a mysql table, SQL planner will return a SHUFFLE join
plan, which may cause a lot of network data transmission.

This is because after PR #3604, the logic of selecting join method has been changed, which is,
when broadcast cost equals to the shuffle cost, planner will prefer to choose SHUFFLE.

But for MysqlScanNode, its cardinality is not initialized and always be -1, so the cost is not actually calculated, and both broadcast and shuffle cost is 0. which results in a SHUFFLE join.

**To Reproduce**
1. create an olap table and load some data;
2. create a mysql table
3. execute a join

**Expected behavior**
Expect a broadcast join

morningman:
Fix #3885

1. Set MysqlScanNode's cardinality to 0

### DIFF
--- a/fe/src/main/java/org/apache/doris/planner/DistributedPlanner.java
+++ b/fe/src/main/java/org/apache/doris/planner/DistributedPlanner.java
@@ -294,6 +294,11 @@ public class DistributedPlanner {
         // broadcast: send the rightChildFragment's output to each node executing
         // the leftChildFragment; the cost across all nodes is proportional to the
         // total amount of data sent
+
+        // NOTICE(cmy): for now, only MysqlScanNode and OlapScanNode has Cardinality. And MysqlScanNode's Cardinality
+        // is always 0. Other ScanNode's cardinality is -1, so they won't be able the calculate the cost of
+        // join normally and result in both "broadcastCost" and "partitionCost" be 0. And this will lead
+        // to a SHUFFLE join.
         PlanNode rhsTree = rightChildFragment.getPlanRoot();
         long rhsDataSize = 0;
         long broadcastCost = 0;
--- a/fe/src/main/java/org/apache/doris/planner/DistributedPlanner.java
+++ b/fe/src/main/java/org/apache/doris/planner/DistributedPlanner.java
@@ -295,8 +295,13 @@ public class DistributedPlanner {
         // the leftChildFragment; the cost across all nodes is proportional to the
         // total amount of data sent
 
-        // NOTICE(cmy): for now, only MysqlScanNode and OlapScanNode has Cardinality. And MysqlScanNode's Cardinality
-        // is always 0. Other ScanNode's cardinality is -1, so they won't be able to calculate the cost of
+        // NOTICE(cmy):
+        // for now, only MysqlScanNode and OlapScanNode has Cardinality.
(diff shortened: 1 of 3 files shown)


kangkaisen (dismissed):
+1, LGTM

fe/src/main/java/org/apache/doris/planner/DistributedPlanner.java
```diff
@@ -294,6 +294,11 @@ private PlanFragment createHashJoinFragment(HashJoinNode node, PlanFragment righ
         // broadcast: send the rightChildFragment's output to each node executing
         // the leftChildFragment; the cost across all nodes is proportional to the
         // total amount of data sent
+
+        // NOTICE(cmy): for now, only MysqlScanNode and OlapScanNode has Cardinality. And MysqlScanNode's Cardinality
+        // is always 0. Other ScanNode's cardinality is -1, so they won't be able the calculate the cost of
```
wutiangan, line 299:
```suggestion
        // is always 0. Other ScanNode's cardinality is -1, so they won't be able to calculate the cost of
```

wutiangan:
```suggestion
        // is always -1. Other ScanNode's cardinality is greater than or equal 1, so they won't be able the calculate the cost of
```

fe/src/main/java/org/apache/doris/planner/MysqlScanNode.java
```diff
@@ -149,4 +150,14 @@ public int getNumInstances() {
         return 1;
     }
 
+    @Override
+    public void computeStats(Analyzer analyzer) {
+        super.computeStats(analyzer);
+        // even current node scan has no data,at least on backend will be assigned when the fragment actually execute
```
wutiangan, line 156:
```suggestion
        // even if current node scan has no data,at least on backend will be assigned when the fragment actually execute
```

fe/src/main/java/org/apache/doris/planner/DistributedPlanner.java
```diff
@@ -294,6 +294,11 @@ private PlanFragment createHashJoinFragment(HashJoinNode node, PlanFragment righ
         // broadcast: send the rightChildFragment's output to each node executing
         // the leftChildFragment; the cost across all nodes is proportional to the
         // total amount of data sent
+
+        // NOTICE(cmy): for now, only MysqlScanNode and OlapScanNode has Cardinality. And MysqlScanNode's Cardinality
```
wuyunfeng, line 298:
```suggestion
        // NOTICE: for now, only MysqlScanNode and OlapScanNode has Cardinality. And MysqlScanNode's Cardinality
```

fe/src/main/java/org/apache/doris/planner/DistributedPlanner.java
morningman: done

wutiangan (dismissed):

EmmyMiao87 (approved):
LGTM

PR #3898, opened 2020-06-18
Support import true or false as boolean value

### LINKED ISSUE #3831: [BUG] load data with bool value will convert to all true value;
**Describe the bug**
When importing bool data into doris, the result of the import is wrong, all will be converted to 1

**To Reproduce**
the table is 
```
+-------+---------+------+-------+---------+-------+
| Field | Type    | Null | Key   | Default | Extra |
+-------+---------+------+-------+---------+-------+
| k1    | BOOLEAN | Yes  | true  | N/A     |       |
+-------+---------+------+-------+---------+-------+
```

the data file is 
```
true
false
true
false
```

execute command 
```
cat bool_data| curl --location-trusted -u root: -T - http://127.0.0.1:8030/api/test/tbl_bool/_stream_load
```
the result is 
```
+------+
| k1   |
+------+
|    1 |
|    1 |
|    1 |
|    1 |
+------+
```
**Expected behavior**
```
+------+
| k1   |
+------+
|    1 |
|    0 |
|    1 |
|    0 |
+------+
```

yangzhg:
Fixes #3831
After this PR 
insert into: `1/"1" -> 1, 0/"0"->0, true/"true"->1, false/"false" -> 0, "10"->null, "xxxx" -> null`
load: `1/true -> 1, 0/false -> 0,  other -> null`

### DIFF
--- a/be/src/exprs/cast_functions.cpp
+++ b/be/src/exprs/cast_functions.cpp
@@ -230,14 +230,16 @@ StringVal CastFunctions::cast_to_string_val(FunctionContext* ctx, const StringVa
 }
 
 BooleanVal CastFunctions::cast_to_boolean_val(FunctionContext* ctx, const StringVal& val) {
-    if (val.is_null) return BooleanVal::null(); 
-        StringParser::ParseResult result;
-        BooleanVal ret;
-        ret.val = StringParser::string_to_bool(reinterpret_cast<char*>(val.ptr), val.len, &result);
-        if (UNLIKELY(result != StringParser::PARSE_SUCCESS || std::isnan(ret.val) || std::isinf(ret.val))) { 
-            return BooleanVal::null();
-        }
-        return ret;
+    if (val.is_null) {
+        return BooleanVal::null(); 
+    }
+    StringParser::ParseResult result;
+    BooleanVal ret;
+    ret.val = StringParser::string_to_bool(reinterpret_cast<char*>(val.ptr), val.len, &result);
+    if (UNLIKELY(result != StringParser::PARSE_SUCCESS)) { 
+        return BooleanVal::null();
+    }
+    return ret;
 }
 
 #if 0
--- a/be/src/exprs/cast_functions.cpp
+++ b/be/src/exprs/cast_functions.cpp
@@ -236,7 +236,7 @@ BooleanVal CastFunctions::cast_to_boolean_val(FunctionContext* ctx, const String
         ret.val = StringParser::string_to_bool(reinterpret_cast<char*>(val.ptr), val.len, &result);
         if (UNLIKELY(result != StringParser::PARSE_SUCCESS || std::isnan(ret.val) || std::isinf(ret.val))) { 
             return BooleanVal::null();
-		}
+        }
         return ret;
 }
 
--- a/be/src/exprs/cast_functions.cpp
+++ b/be/src/exprs/cast_functions.cpp
@@ -235,9 +235,16 @@ BooleanVal CastFunctions::cast_to_boolean_val(FunctionContext* ctx, const String
     }
     StringParser::ParseResult result;
     BooleanVal ret;
-    ret.val = StringParser::string_to_bool(reinterpret_cast<char*>(val.ptr), val.len, &result);
-    if (UNLIKELY(result != StringParser::PARSE_SUCCESS)) { 
-        return BooleanVal::null();
+    IntVal int_val = cast_to_int_val(ctx, val);
+    if (!int_val.is_null && int_val.val == 0) {
+        ret.val = false;
+    } else if (!int_val.is_null && int_val.val == 1) {
+        ret.val = true;
+    } else {
+        ret.val = StringParser::string_to_bool(reinterpret_cast<char*>(val.ptr), val.len, &result);
+        if (UNLIKELY(result != StringParser::PARSE_SUCCESS)) { 
+            return BooleanVal::null();
+        }
     }
     return ret;
 }
--- a/be/src/exprs/cast_functions.cpp
+++ b/be/src/exprs/cast_functions.cpp
@@ -229,6 +229,17 @@ StringVal CastFunctions::cast_to_string_val(FunctionContext* ctx, const StringVa
   return sv;
 }
 
+BooleanVal CastFunctions::cast_to_boolean_val(FunctionContext* ctx, const StringVal& val) {
+    if (val.is_null) return BooleanVal::null(); 
+        StringParser::ParseResult result;
+        BooleanVal ret;
+        ret.val = StringParser::string_to_bool(reinterpret_cast<char*>(val.ptr), val.len, &result);
+        if (UNLIKELY(result != StringParser::PARSE_SUCCESS || std::isnan(ret.val) || std::isinf(ret.val))) { 
+            return BooleanVal::null();
+		}
+        return ret;
+}
+
 #if 0
 StringVal CastFunctions::CastToChar(FunctionContext* ctx, const StringVal& val) {
   if (val.is_null) return StringVal::null();
--- a/be/src/olap/utils.cpp
+++ b/be/src/olap/utils.cpp
@@ -51,6 +51,7 @@
 #include "olap/olap_define.h"
 #include "util/errno.h"
 #include "util/mutex.h"
+#include "util/string_parser.hpp"
 
 using std::string;
 using std::set;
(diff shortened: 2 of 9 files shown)


be/src/exprs/cast_functions.cpp
```diff
@@ -229,6 +229,17 @@ StringVal CastFunctions::cast_to_string_val(FunctionContext* ctx, const StringVa
...
 }
 
+BooleanVal CastFunctions::cast_to_boolean_val(FunctionContext* ctx, const StringVal& val) {
+    if (val.is_null) return BooleanVal::null(); 
+        StringParser::ParseResult result;
+        BooleanVal ret;
+        ret.val = StringParser::string_to_bool(reinterpret_cast<char*>(val.ptr), val.len, &result);
+        if (UNLIKELY(result != StringParser::PARSE_SUCCESS || std::isnan(ret.val) || std::isinf(ret.val))) { 
+            return BooleanVal::null();
+		}
```
chaoyli, line 239: Brackets is not aligned.

be/src/olap/utils.cpp
```diff
@@ -1264,6 +1265,15 @@ bool valid_datetime(const string &value_str) {
     }
 }
 
+bool valid_bool(const std::string& value_str) {
+    if (value_str == "0" || value_str == "1") {
+        return true;
+    }
+    StringParser::ParseResult result;
+    StringParser::string_to_bool(value_str.c_str(), value_str.length(), &result);
```
chaoyli, line 1273: I found string_to_bool will convert "Truetest" to be true, it is ok?
```
543 inline bool StringParser::string_to_bool_internal(const char* s, int len, ParseResult* result) {
544     *result = PARSE_SUCCESS;
545
546     if (len >= 4 && (s[0] == 't' || s[0] == 'T')) {
547         bool match = (s[1] == 'r' || s[1] == 'R') &&
548             (s[2] == 'u' || s[2] == 'U') &&
549             (s[3] == 'e' || s[3] == 'E');
550         if (match && LIKELY(is_all_whitespace(s + 4, len - 4))) {
551             return true;
552         }
553     } else if (len >= 5 && (s[0] == 'f' || s[0] == 'F')) {
554         bool match = (s[1] == 'a' || s[1] == 'A') &&
555             (s[2] == 'l' || s[2] == 'L') &&
556             (s[3] == 's' || s[3] == 'S') &&
557             (s[4] == 'e' || s[4] == 'E');
558         if (match && LIKELY(is_all_whitespace(s + 5, len - 5))){
559             return false;
560         }
561     }
```

chaoyli (changes requested):

be/src/olap/utils.cpp
yangzhg: "Truetest" won't be converted to `true`  because of `is_all_whitespace()`, it will return  `PARSE_FAILURE`. In additional, this is just test the string is `true` or `false` ignore case;

be/src/exprs/cast_functions.cpp
yangzhg: Please review it again @chaoyli

be/src/olap/utils.cpp
chaoyli: If "true " will be ok ? @yangzhg

be/src/exprs/cast_functions.cpp
```diff
@@ -229,6 +229,17 @@ StringVal CastFunctions::cast_to_string_val(FunctionContext* ctx, const StringVa
   return sv;
 }
 
+BooleanVal CastFunctions::cast_to_boolean_val(FunctionContext* ctx, const StringVal& val) {
+    if (val.is_null) return BooleanVal::null(); 
+        StringParser::ParseResult result;
```
wutiangan, line 234: modify the indent

be/src/olap/utils.cpp
wutiangan: in BoolLiteral.BoolLiteral(String value) funciton, string "1" is  consider to true. but in string_to_bool_internal(), string "1"  is not consider to true, so it is puzzle. maybe you can unify it.

be/src/exprs/cast_functions.cpp
```diff
@@ -229,6 +229,17 @@ StringVal CastFunctions::cast_to_string_val(FunctionContext* ctx, const StringVa
...
+        BooleanVal ret;
+        ret.val = StringParser::string_to_bool(reinterpret_cast<char*>(val.ptr), val.len, &result);
+        if (UNLIKELY(result != StringParser::PARSE_SUCCESS || std::isnan(ret.val) || std::isinf(ret.val))) {
```
wutiangan, line 237: why you add the "std::isnan(ret.val) || std::isinf(ret.val)"？if result == StringParser::PARSE_SUCCESS, can val be nan or inf ?

be/src/exprs/cast_functions.cpp
```diff
@@ -229,6 +229,17 @@ StringVal CastFunctions::cast_to_string_val(FunctionContext* ctx, const StringVa
   return sv;
 }
 
+BooleanVal CastFunctions::cast_to_boolean_val(FunctionContext* ctx, const StringVal& val) {
+    if (val.is_null) return BooleanVal::null(); 
+        StringParser::ParseResult result;
+        BooleanVal ret;
+        ret.val = StringParser::string_to_bool(reinterpret_cast<char*>(val.ptr), val.len, &result);
```
wutiangan, line 236: StringParser::string_to_bool will consider "1" to false, is it ok?

yangzhg: "1" will convert to int  not bool

caiconghui:
@yangzhg Hi, I think the result of process true false value  for stream load and insert stmt should keep   consistent. 
For now, like insert into xx values(1), result=true, insert into xx values(0) result=false, insert into xxx values("false") result=true  insert into xx values(false) result=false  insert into xx values("0") result=true, insert into xx values(true) result=true. 
This is because insert stmt can analyze the true data type for insert value, so like 1 2 3, will use int_cast_to_bool, even can analyze boolen type. but for stream load, we lost the data type information. we don't know the real data type of the value, we treat all column values as string, so I think may be stream load should be processed specially to keep consistent with insert stmt. maybe need more discussion.

yangzhg:
> @yangzhg Hi, I think the result of processing bool value for stream load and insert stmt should keep consistent.
> For now, like insert into xx values(1), result=true, insert into xx values(0) result=false, insert into xxx values("false") result=true insert into xx values(false) result=false insert into xx values("0") result=true, insert into xx values(true) result=true.
> This is because insert stmt can analyze the real data type for insert value, so like 1 2 3, will use int_cast_to_bool, even can analyze boolean type. but for stream load, we lost the data type information. we don't know the real data type of the value, we treat all column values as string, so I think may be stream load should be processed specially to keep consistent with insert stmt. maybe need more discussion.



> @yangzhg Hi, I think the result of processing bool value for stream load and insert stmt should keep consistent.
> For now, like insert into xx values(1), result=true, insert into xx values(0) result=false, insert into xxx values("false") result=true insert into xx values(false) result=false insert into xx values("0") result=true, insert into xx values(true) result=true.
> This is because insert stmt can analyze the real data type for insert value, so like 1 2 3, will use int_cast_to_bool, even can analyze boolean type. but for stream load, we lost the data type information. we don't know the real data type of the value, we treat all column values as string, so I think may be stream load should be processed specially to keep consistent with insert stmt. maybe need more discussion.

insert into xx values("0") and stream load import file 0 result itself is contradictory, in the current code insert into xx values("0") result=true, stream load 0 result=false, and now cast_to_bool Only occurs when the type of the column is bool, so it is not contradictory to the current insert

chaoyli:
I think the meaning of caiconghui is to make consistent of stream load and insert in any situation.
Otherwise, it will cause some misunderstanding.

yangzhg:
> @yangzhg Hi, I think the result of processing bool value for stream load and insert stmt should keep consistent.
> For now, like insert into xx values(1), result=true, insert into xx values(0) result=false, insert into xxx values("false") result=true insert into xx values(false) result=false insert into xx values("0") result=true, insert into xx values(true) result=true.
> This is because insert stmt can analyze the real data type for insert value, so like 1 2 3, will use int_cast_to_bool, even can analyze boolean type. but for stream load, we lost the data type information. we don't know the real data type of the value, we treat all column values as string, so I think may be stream load should be processed specially to keep consistent with insert stmt. maybe need more discussion.

I will change the insert statement  and stream load to keep consistent with mysql

chaoyli (approved):

PR #3902, opened 2020-06-18
Support max_stream_load_timeout_second config in fe

### LINKED ISSUE #3905: [Config]max_stream_load_timeout_second config is needed for preventing user's large stream load timeout setting
Now, all load type txns share same max_load_timeout_second config, which is not sutiable for stream load txn. stream load txn cost much less time than broker load. so max_stream_load_timeout_second is needed for us to preventing user's large stream load timeout setting, which may cause failed stream load cannot be aborted within a short time

caiconghui:
This configuration is specifically used to limit timeout setting for stream load. It is to prevent that failed stream load transactions cannot be canceled within a short time because of the user's large timeout setting.

### DIFF
--- a/fe/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/src/main/java/org/apache/doris/common/Config.java
@@ -474,10 +474,16 @@ public class Config extends ConfigBase {
     public static int stream_load_default_timeout_second = 600; // 600s
 
     /*
-     * Max load timeout applicable to all type of load
+     * Max load timeout applicable to all type of load except for stream load
      */
     @ConfField(mutable = true, masterOnly = true)
     public static int max_load_timeout_second = 259200; // 3days
+
+    /*
+     * Max stream load and streaming mini load timeout
+     */
+    @ConfField(mutable = true, masterOnly = true)
+    public static int max_stream_load_timeout_second = 600; // 600s
 
     /*
     * Min stream load timeout applicable to all type of load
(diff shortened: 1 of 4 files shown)


caiconghui:
#3902

chaoyli (dismissed):

fe/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -474,11 +474,17 @@
...
+     * Max load timeout applicable to all type of load except for stream load
      */
     @ConfField(mutable = true, masterOnly = true)
     public static int max_load_timeout_second = 259200; // 3days
 
+    /*
+     * Max stream load and streaming mini load timeout
+     */
+    @ConfField(mutable = true, masterOnly = true)
+    public static int max_stream_load_timeout_second = 600; // 600s
```
morningman, line 486: You need to set the default value to a large value, I suggest to be same as `max_load_timeout_second`.
Otherwise, after online, some existing load job may fail to be submitted.

caiconghui:
ref #3902

morningman (changes requested):

fe/src/main/java/org/apache/doris/common/Config.java
caiconghui: ok

morningman (approved):
LGTM

PR #3903, opened 2020-06-18
[Bug] Make compaction metrics value is right

### LINKED ISSUE #3904: Compaction not collect the number of rowsets compacted rightly
I0618 16:08:34.805496 216799 base_compaction.cpp:58] _input_rowsets.size() : 0, base_compaction_deltas_total: 0

chaoyli:
Now _input_rowsets will be cleared when calling gc_used_rowsets().
After that, the metrics is not right upon be calculated.


morningman (approved):
LGTM

PR #3909, opened 2020-06-19
[webserver] Make BE webserver more pretty

### LINKED ISSUE #3911: [Proposal] Make BE web site more pretty
Doris backends web server doesn't support to GET static file, and also doesn't support CSS style.
When we want to show more detail information on BE's web site, it's not easy to add elements. For example, as for https://github.com/apache/incubator-doris/issues/3714, we want to show memtrackers, it's hard to add elements and show them in a pretty style.

Transfer the webserver utils which Impala and Kudu using:
- Web CSS framwork [boost-table](https://github.com/wenzhixin/bootstrap-table): helpful to make website more pretty
- A thirdparty library [mustache](https://github.com/henryr/cpp-mustache): helpful to tracnslate JSON structure to html

And then do some refactor on current BE web handler.
After that, our BE web site might looks like:
image

acelyc111:
Improve BE's website by:
- Introduce a web CSS framework boost-table which is helpful to make website more pretty
- Introduce a thirdparty library mustache which is helpful to tracnslate JSON structure to html
- Do some refactor on BE's webserver to handle static files


be/src/www/doris.js
```diff
@@ -0,0 +1,148 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
imay, line 1: Why not put all web to /webroot?

be/src/www/bootstrap/js/bootstrap.min.js
```diff
@@ -0,0 +1,7 @@
+/*!
```
imay, line 1: If new thirdparty file is added, "LICENSE.txt" will be updated

imay (changes requested):

thirdparty/build-thirdparty.sh
```diff
@@ -685,6 +685,18 @@ build_cctz() {
     make -j$PARALLEL && make install
 }
 
+#mustache
+build_mustache() {
```
morningman, line 689: In `https://github.com/henryr/cpp-mustache/` it said:
`Easiest way is to drop mustache.{cc,h} in your project.`
so why not just introduce the 2 files? So that we don't need to update our compile env.

be/src/www/doris.js
acelyc111: I've moved all be's web files to /webroot/be

be/src/www/bootstrap/js/bootstrap.min.js
acelyc111: Added.

acelyc111:
@imay @morningman could you please rereview this PR, thanks!

acelyc111:
I‘ll split the pr to multiple prs, the 1st: https://github.com/apache/incubator-doris/pull/4021

Issue #3229, opened 2020-03-30
Doris version 0.11.0 dockerfile ask maven version 3.6.0，but where is it???????


yangbiao121:
![image](https://user-images.githubusercontent.com/26757248/77898799-4ecaa200-72ae-11ea-95b4-8a8f04591052.png)

imay:
@yangbiao121 
1. we have fixed mvn version to 3.6.3 in master branch.
2. It is not a good idea to build docker file yourself. we have already provided a docker image, you can use it directly.

I will close this issue.

PR #3910, opened 2020-06-19
[Bug] Unify the timezone

### LINKED ISSUE #3919: [Bug] Create table with dynamic partition encounter exception
**Describe the bug**
When create table with dynamic partition without specifying the time zone, it will throw exception:

`nknown or incorrect time zone: 'PRC'`

**To Reproduce**
Ccreate table with dynamic partition without specifying the time zone

**Expected behavior**
Use default time zone to create table with dynamic partition successfully.

morningman:
When we get default system time zone, it will return `PRC`, which is not supported by us, thus
will cause dynamic partition create failed. Fix #3919

This CL mainly changes:
1. Use a unified method to get the system default time zone
2. Now the default variable `system_time_zone` and `time_zone` is set to the default system
time zone, which is `Asia/Shanghai`.
3. Modify related unit test.
4. Support time zone `PRC`.

### DIFF
--- a/fe/src/main/java/org/apache/doris/catalog/DynamicPartitionProperty.java
+++ b/fe/src/main/java/org/apache/doris/catalog/DynamicPartitionProperty.java
@@ -48,7 +48,7 @@ public class DynamicPartitionProperty {
     private int buckets;
     private StartOfDate startOfWeek;
     private StartOfDate startOfMonth;
-    private TimeZone tz = TimeUtils.getDefaultTimeZone();
+    private TimeZone tz = TimeUtils.getSystemTimeZone();
 
     public DynamicPartitionProperty(Map<String, String> properties) {
         if (properties != null && !properties.isEmpty()) {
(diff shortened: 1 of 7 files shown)


fe/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -554,6 +554,13 @@ public static Catalog getCurrentCatalog() {
         }
     }
 
+    // NOTICE: In most cases, you should use getCurrentCatalog() to get the catalog.
+    // getServiingCatalog() is only used for Checkpoint thread and some special places
```
WingsGo, line 558: getServingCatalog

morningman: Oh, I moved the change to PR #3229
I will rebase after that PR merged.

morningman: Already for review.

WingsGo:
LGTM

EmmyMiao87 (approved):
LGTM

PR #3913, opened 2020-06-19
[Bug] Checkpoint thread is not running

morningman:
This bug is introduced by PR #3784 
In #3784, I remove the `Catalog.getInstance()`, and use `Catalog.getCurrentCatalog()` instead.

But actually, there are some place that should use the serving catalog explicitly.

Mainly changed:

1. Add a new method `getServingCatalog()` to explicitly return the real catalog instance.
2. Fix a compile bug of broker introduced by #3881


EmmyMiao87 (approved):
LGTM

Issue #3908, opened 2020-06-18
Column can not be resloved  after reorderTable


spaces-X:
1. Create two tables:

```
CREATE TABLE `s1` (
  `dt` int(11) NULL COMMENT "",
  `id` int(11) NULL COMMENT "",
  `pv` int(11) SUM NULL COMMENT ""
) ENGINE=OLAP
AGGREGATE KEY(`dt`, `id`)
COMMENT "OLAP"
DISTRIBUTED BY HASH(`id`) BUCKETS 1
PROPERTIES (
"replication_num" = "1",
"in_memory" = "false",
"storage_format" = "DEFAULT"
);
```

```
CREATE TABLE `s2` (
  `dt` int(11) NULL COMMENT "",
  `id` int(11) NULL COMMENT "",
  `pv` int(11) SUM NULL COMMENT ""
) ENGINE=OLAP
AGGREGATE KEY(`dt`, `id`)
COMMENT "OLAP"
DISTRIBUTED BY HASH(`id`) BUCKETS 1
PROPERTIES (
"replication_num" = "1",
"in_memory" = "false",
"storage_format" = "DEFAULT"
);
```
2. load some data:
Actually this step can be ignored, only to simulate normal use.
```
insert into s1 (dt, id, pv) values (20200609, 1, 10), (20200609, 2, 20);
insert into s2 (dt, id, pv) values (20200609, 1, 10), (20200609, 2, 20);
```
Query sql:
```
SELECT tt1.dt FROM s1 tt1
INNER JOIN s2  tt2 ON tt2.dt = 20200610
INNER JOIN (
   SELECT id
   FROM s2 ) tt3
ON tt3.id = tt2.id
AND tt1.pv >  1+1
INNER JOIN s1 tt4 ON tt4.pv = tt1.pv
```

ERROR 1054 (42S22): errCode = 2, detailMessage = Unknown column 'id' in 'tt2'

This error is caused by reorderTable.
We expect the order of these tables is that tt3 should be after tt2 because tt3  depends on tt2  with onclause `ON tt3.id = tt2.id`. But actually tt3 is ahead of tt2, which will cause `id` in `tt2` has not been resloved before using.

Expected:      `tt1->tt2->tt3->tt4`    At least tt2 should be ahead of tt3.
reordered:     `tt3->tt1->tt2->tt4`
After FromClause#sortTableRefForSubquery:  `tt1->tt3->tt2->tt4`

[...]

PR #3915, opened 2020-06-19
fix subquery bug when reorder table

spaces-X:
To solve this [issue](https://github.com/apache/incubator-doris/issues/3908)

Ask a question here:
What‘s the benefits for using candidates which is ordered by rowcounts rather than the original  `fromclause` ?

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
+++ b/fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
@@ -813,12 +813,6 @@ public class SelectStmt extends QueryStmt {
                 return;
             }
         }
-
-        // can not get AST only with equal join, MayBe cross join can help
-        fromClause_.clear();
-        for (Pair<TableRef, Long> candidate : candidates) {
-            fromClause_.add(candidate.first);
-        }
     }
 
     // reorder select table


imay:
Doris will construct a left-deep tree for the join tables. In the star schema, fact table usually is the biggest table, it will be make the leftmost node by ordering the table by row counts. When doing hash-join, all the right table will be loaded into memory to build a hash table. If the fact table is regarded as build table, the performance will be bad.

fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
```diff
@@ -813,12 +813,6 @@ protected void reorderTable(Analyzer analyzer) throws AnalysisException {
                 return;
             }
         }
-
-        // can not get AST only with equal join, MayBe cross join can help
```
imay, line 817: Should not remove these code

imay (changes requested):

spaces-X:
> Doris will construct a left-deep tree for the join tables. In the star schema, fact table usually is the biggest table, it will be make the leftmost node by ordering the table by row counts. When doing hash-join, all the right table will be loaded into memory to build a hash table. If the fact table is regarded as build table, the performance will be bad.

Thanks a lot.
We have encoutered some problems in reordering table. [#3908 ](https://github.com/apache/incubator-doris/issues/3908)
After reordering table and query rewrite, stmt will be reanalyzed.
In reanalyze process, column can not be resloved.
Do you have some advice to improve this case?

spaces-X:
@imay Hi, I think star schema will return before in line 813.
When we go to these code, it probably means this sql is not star schema.
So, we should not use the row counts by force, which will cause errors in snowflakes shema.
Recover the order of tables in from clause is more reasonable.

PR #3812, opened 2020-06-09
Forbidden float column in short key

EmmyMiao87:
When the user does not specify the key column, doris will automatically supplement the key column.
However, doris does not support float or double as the key column, so when adding the key column, doris should avoid setting those column as the key column.

The CreateMaterailizedView, AddRollup and CreateDuplicateTable need to forbidden float column in short key.
If the float column is directly encountered during the supplement process, the subsequent columns are all value columns.
If the first column is float or double, Doris will throw the exception.

Fixed #3811

Change-Id: Ib66d9355acefcd8f281906bcb7b4dd684319ff08

PR #3916, opened 2020-06-19
[Bug] Can not use non-key column as partition column in duplicate table

morningman:
The following statement will throw error:
```
create table test.tbl2
(k1 int, k2 int, k3 float)
duplicate key(k1)
partition by range(k2)
(partition p1 values less than("10"))
distributed by hash(k3) buckets 1
properties('replication_num' = '1'); 
```
Error: `Only key column can be partition column`

But in duplicate key table, columns can be partition or distribution column
even if they are not in duplicate keys.

This bug is introduced by #3812

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/RangePartitionDesc.java
+++ b/fe/src/main/java/org/apache/doris/analysis/RangePartitionDesc.java
@@ -150,7 +150,7 @@ public class RangePartitionDesc extends PartitionDesc {
             for (Column column : schema) {
                 if (column.getName().equalsIgnoreCase(colName)) {
                     if (!column.isKey() && column.getAggregationType() != AggregateType.NONE) {
-                        throw new DdlException("Partition column[" + colName + "] is not key column");
+                        throw new DdlException("The partition column could not be aggregated column");
                     }
 
                     if (column.getType().isFloatingPointType()) {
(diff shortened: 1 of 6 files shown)


fe/src/main/java/org/apache/doris/analysis/RangePartitionDesc.java
```diff
@@ -145,9 +149,14 @@ public PartitionInfo toPartitionInfo(List<Column> schema, Map<String, Long> part
             boolean find = false;
             for (Column column : schema) {
                 if (column.getName().equalsIgnoreCase(colName)) {
-                    if (!column.isKey()) {
+                    if (!column.isKey() && column.getAggregationType() != AggregateType.NONE) {
```
EmmyMiao87, line 152: Why does it need a second check here?

morningman: Historical issues...
It is OK, may be removed in future.

fe/src/main/java/org/apache/doris/analysis/RangePartitionDesc.java
```diff
@@ -145,9 +149,14 @@ public PartitionInfo toPartitionInfo(List<Column> schema, Map<String, Long> part
...
                         throw new DdlException("Partition column[" + colName + "] is not key column");
```
EmmyMiao87, line 153:
```suggestion
                        throw new DdlException("The partition column could not be aggregated column");
```

EmmyMiao87 (approved):
LGTM

PR #3923, opened 2020-06-21
[Spill to Disk] Aggregation Node Support Spill to disk in big query

HappenLee:
#3926 issue

This pr mainly about:

1. Clear_reservations of buffer block client when node close()
2. Replace ```buffered_tuple_stream3``` by ```buffered_tuple_stream2``` in ```PartitionAggregationNode``` to support spill to disk 
3. Fix bug of available IO buffer can be used in ```spill_sorter```.
4. Fix the bug of count of ```_unfullfilled_reserved_buffers``` in ```buffered_block_mgr2.cc```, we should only count reverved_buffer when the buffer size == IO block size.


morningman:
Hi @HappenLee , Please to create an issue to describe the why you did it and how it works.
And also in the PR's comment, you should describe the main changes of your PR.

PR #3925, opened 2020-06-22
[log] Downgrade a log in RunLengthByteReader from WARNING to VLOG

acelyc111:
There are too many logs in be.WARNING looks like:
```
W0622 17:47:52.513341 26554 run_length_byte_reader.cpp:102] fail to ReadOnlyFileStream seek.[res = -1705]
W0622 17:47:52.513417 26554 run_length_byte_reader.cpp:102] fail to ReadOnlyFileStream seek.[res = -1705]
W0622 17:47:52.513466 26554 run_length_byte_reader.cpp:102] fail to ReadOnlyFileStream seek.[res = -1705]
```
It's a normal case when a run length is EOF, so we can downgrade it from
WARNING to INFO to reduce useless log in be.WARNING

### DIFF
--- a/be/src/olap/rowset/run_length_byte_reader.cpp
+++ b/be/src/olap/rowset/run_length_byte_reader.cpp
@@ -99,7 +99,7 @@ OLAPStatus RunLengthByteReader::seek(PositionProvider* position) {
 
     res = _input->seek(position);
     if (OLAP_SUCCESS != res) {
-        OLAP_LOG_WARNING("fail to ReadOnlyFileStream seek.[res = %d]", res);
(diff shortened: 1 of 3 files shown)


morningman:
If its, how about move it to VLOG?

chaoyli:
> If its, how about move it to VLOG?
> @chaoyli Any idea?

I agree it, downgrade to INFO will still have much of log.
You can use vlog(10) to replace it.

acelyc111:
> > If its, how about move it to VLOG?
> > @chaoyli Any idea?
> 
> I agree it, downgrade to INFO will still have much of log.
> You can use vlog(10) to replace it.

Done

chaoyli (approved):

PR #3928, opened 2020-06-23
[Bug Fix]Fix the core in Data_Stream_Recvr

HappenLee:
Remove the map in DataStreamRecvr and replace by vector<pair<Closure*, MonotonicStopWatch>>.

### DIFF
--- a/be/src/runtime/data_stream_recvr.cc
+++ b/be/src/runtime/data_stream_recvr.cc
@@ -128,7 +128,7 @@ private:
     std::unordered_set<int> _sender_eos_set; // sender_id
     std::unordered_map<int, int64_t> _packet_seq_map; // be_number => packet_seq
 
-    std::deque<google::protobuf::Closure*> _pending_closures;
+    std::deque<std::pair<google::protobuf::Closure*, MonotonicStopWatch>> _pending_closures;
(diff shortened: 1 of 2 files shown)


be/src/runtime/data_stream_recvr.cc
```diff
@@ -176,14 +176,12 @@ Status DataStreamRecvr::SenderQueue::get_batch(RowBatch** next_batch) {
     *next_batch = _current_batch.get();
 
     if (!_pending_closures.empty()) {
-        auto done = _pending_closures.front();
-        done->Run();
+        auto closure_pair = _pending_closures.front();
+        closure_pair.first->Run();
```
chaoyli, line 180: Seems have not call start() of MonotonicStopWatch

chaoyli (changes requested):

be/src/runtime/data_stream_recvr.cc
HappenLee: image
@chaoyli Hi，the MonotonicStopWatch start here.

chaoyli (approved):

chaoyli:
You can describe some the solution in the commit message.

PR #3933, opened 2020-06-23
[Bug] Fix bug that a query plan is not correctly cancelled

morningman:
This bug is introduced by #3872.
It will cause some expected to be cancelled queries not being cancelled.

### DIFF
--- a/be/src/runtime/plan_fragment_executor.cpp
+++ b/be/src/runtime/plan_fragment_executor.cpp
@@ -487,21 +487,22 @@ Status PlanFragmentExecutor::get_next_internal(RowBatch** batch) {
     return Status::OK();
 }
 
-void PlanFragmentExecutor::update_status(const Status& status) {
-    if (status.ok()) {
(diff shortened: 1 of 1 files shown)


yangzhg (approved):
LGTM

be/src/runtime/plan_fragment_executor.cpp
```diff
@@ -487,21 +487,22 @@ Status PlanFragmentExecutor::get_next_internal(RowBatch** batch) {
     return Status::OK();
 }
```
chaoyli, line 489: Only rename status to new_status?

chaoyli (changes requested):

be/src/runtime/plan_fragment_executor.cpp
morningman: See changes in line 497 `if (_status.ok())`

chaoyli (approved):

Issue #3929, opened 2020-06-23
[Bug] BE crash occasionally


morningman:
**Describe the bug**
BE crash occasionally and `be.out` shows:

```
palo_be: ../nptl/pthread_mutex_lock.c:80: __pthread_mutex_lock: Assertion `mutex->__data.__owner == 0' failed.
*** Aborted at 1592890873 (unix time) try "date -d @1592890873" if you are using GNU date ***
PC: @     0x7f599a2af3f7 __GI_raise
*** SIGABRT (@0x1f4000081a6) received by PID 33190 (TID 0x7f59797b6700) from PID 33190; stack trace: ***
    @     0x7f599a2af470 (unknown)
    @     0x7f599a2af3f7 __GI_raise
    @     0x7f599a2b07d8 __GI_abort
    @     0x7f599a2a8516 __assert_fail_base
    @     0x7f599a2a85c2 __GI___assert_fail
    @     0x7f599a06658c __GI___pthread_mutex_lock
    @          0x1ba34d6 pthread_mutex_lock
    @          0x145f4ac doris::OlapScanNode::scanner_thread()
    @           0xfa8a35 doris::PriorityThreadPool::work_thread()
    @          0x1a5bbed thread_proxy
    @     0x7f599a0641c3 start_thread
    @     0x7f599a36112d __clone
``

The reason is that when trying to lock a mutex, the assertion failed at `mutex->__data.__owner == 0`. It expected `__owner == 0`, which is not.

But when I look into the core dump file, the `__owner` field of that mutex is 0.

```
#0  0x00007f599a2af3f7 in raise () from /opt/compiler/gcc-4.8.2/lib64/libc.so.6
#1  0x00007f599a2b07d8 in abort () from /opt/compiler/gcc-4.8.2/lib64/libc.so.6
#2  0x00007f599a2a8516 in __assert_fail_base () from /opt/compiler/gcc-4.8.2/lib64/libc.so.6
#3  0x00007f599a2a85c2 in __assert_fail () from /opt/compiler/gcc-4.8.2/lib64/libc.so.6
#4  0x00007f599a06658c in pthread_mutex_lock () from /opt/compiler/gcc-4.8.2/lib64/libpthread.so.0
#5  0x0000000001ba34d6 in pthread_mutex_lock_impl (mutex=0x67521610) at /home/palo/thirdparty/src/incubator-brpc-0.9.5/src/bthread/mutex.cpp:551
[...]

PR #3934, opened 2020-06-23
[Refactor] Replace some boost to std in OlapScanNode

morningman:
Replace some boost to std in OlapScanNode.

This refactor seems solve the problem describe in #3929.
Because I found that BE will crash to calling `boost::condition_variable.notify_all()`.
But after upgrade to this, BE does not crash any more.

### DIFF
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -304,7 +304,7 @@ Status OlapScanNode::close(RuntimeState* state) {
     _scan_batch_added_cv.notify_all();
 
     // join transfer thread
-    _transfer_thread.join();
+    _transfer_thread->join();
 
     // clear some row batch in queue
     for (auto row_batch : _materialized_row_batches) {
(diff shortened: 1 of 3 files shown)


be/src/exec/olap_scan_node.cpp
```diff
@@ -186,9 +185,9 @@ Status OlapScanNode::get_next(RuntimeState* state, RowBatch* row_batch, bool* eo
 
     // check if Canceled.
     if (state->is_cancelled()) {
-        boost::unique_lock<boost::mutex> l(_row_batches_lock);
+        std::unique_lock<std::mutex> l(_row_batches_lock);
```
chaoyli, line 188: You can use util/condition_variable.h.
It comes from kudu, it's use is simple than std::

chaoyli (changes requested):

be/src/exec/olap_scan_node.cpp
morningman: This PR is only try to solve the BE crash problem and I don't want to introduce any new potential problems.
I can't see any necessary to replace the `std::condition_variable` with `ConditionVariable`. If it is,
maybe we can replace all `std::condition_variable` in another PR.

chaoyli (approved):

PR #3943, opened 2020-06-24
[Bug] Fix bug that delete stmt with filter condition delete all data from table on segment v2

### LINKED ISSUE #3942: [Bug] Delete stmt with filter condition delete all data from table on segment v2
**Describe the bug**
delete stmt with filter condition delete all data from table on segment v2.

**To Reproduce**
1. set global default_rowset_type="beta";

2. create table
```   
CREATE TABLE `test_beta` (
  `month` datetime NOT NULL,
  `service_name` varchar(100) NOT NULL,
  `cluster_name` varchar(100) NOT NULL
) ENGINE=OLAP
DUPLICATE KEY(`month`, `service_name`, `cluster_name`)
COMMENT "OLAP"
PARTITION BY RANGE(`month`)
(PARTITION p202004 VALUES [('2020-04-01 00:00:00'), ('2020-05-01 00:00:00')))
DISTRIBUTED BY HASH(`service_name`) BUCKETS 1
PROPERTIES (
"replication_num" = "1",
"in_memory" = "false",
"storage_format" = "DEFAULT"
); 
```

3. insert data into table
```  
insert into test_beta values("2020-04-01 00:00:00", "shop11", "xxxxx"), ("2020-04-01 00:00:00", "shop12", "11");
```

4. delete the table
```
delete from test_beta partition p202004 where service_name="shop11" and month="2020-04-01 00:00:00";
```
5.  
```
mysql> select * from test_beta;
Empty set (0.02 sec)
```
    
**Expected behavior**
```
mysql> select * from test_beta;
+---------------------+--------------+--------------+
| month               | service_name | cluster_name |
+---------------------+--------------+--------------+
| 2020-04-01 00:00:00 | shop12       | 11           |
+---------------------+--------------+--------------+
```
1 row in set (0.01 sec)

caiconghui:
Fix #3942


chaoyli:
You can describe the solution in the commit message.

caiconghui:
> You can describe the solution in the commit message.

done

morningman:
It is strange that the case you described in issue is truly has problem. But the following case is OK:

```
mysql> create table tbl4 (k1 int, k2 int) distributed by hash(k1) buckets 1 properties("replication_num" = "1");

mysql> insert into tbl4 values(1,2),(3,4);
Query OK, 2 rows affected (0.04 sec)
{'label':'insert_20ddaec3-5d29-43a9-b065-7d3d680e9d33', 'status':'VISIBLE', 'txnId':'3014'}

mysql> select * from tbl4;
+------+------+
| k1   | k2   |
+------+------+
|    1 |    2 |
|    3 |    4 |
+------+------+
2 rows in set (0.01 sec)

mysql> delete from tbl4 where k2=2 and k1=1;
Query OK, 0 rows affected (0.04 sec)
{'label':'delete_1e19b1ba-0a42-48ee-989c-813758ea9a36', 'status':'VISIBLE', 'txnId':'3015'}

mysql> select * from tbl4;
+------+------+
| k1   | k2   |
+------+------+
|    3 |    4 |
+------+------+
1 row in set (0.01 sec)
```

morningman:
I think the root cause is we lost the relationship of the 2 delete condition in same delete stmt.
For example, the following 2 cases should has different result:

Base data:
```
k1, k2
1, 2
1, 3
```

case 1:
`delete from tbl where k1=1 and k2=2`;
case 2:
`delete from tbl where k1=1;`
`delete from tbl where k2=2;`

We treat the above 2 cases as same, which is incorrect.

So I think only change the `intersect` to `union` is not the right way to fix the bug.

morningman:
I am not familiar with this part. But following is the  pseudocode I thought:

```
All DeleteConditions:

DeleteConditions 1: k1 = 1 and k2 = 2;
DeleteConditions 2: k1 = 4
DeleteConditions 3: k1 = 5


RowRanges result;
foreach(DeleteConditions dcs : All DeleteConditions) {
    RowRanges rr;
    foreach(DeleteCondition dc : dcs) {
        rr = rr UNION (rows filtered by dc)
    }
    result = result INTERSECT rr;
}
return result;


```

caiconghui:
@morningman it is a good implementation to filter block which not need read, although  only change the intersect to union can fix the bug but not an efficient fix.

chaoyli (dismissed):

chaoyli (dismissed):

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -253,30 +253,37 @@ Status SegmentIterator::_get_row_ranges_from_conditions(RowRanges* condition_row
...
+    // final filter data with delete conditions
+    for (auto& delete_condition : _opts.delete_conditions) {
+        RowRanges delete_condition_row_ranges = RowRanges::create_single(0);
+        for (auto& delete_column_condition : delete_condition->columns()) {
+            const int32_t cid = delete_column_condition.first;
+            CondColumn* column_cond = _opts.conditions->get_column(cid);
+            RowRanges single_delete_condition_row_ranges = RowRanges::create_single(num_rows());
+            RETURN_IF_ERROR(
+                    _column_iterators[cid]->get_row_ranges_by_zone_map(
+                            column_cond,
```
morningman, line 279: Should the `column_cond` be null here?
We already filter the data by column conditions before.

caiconghui: @morningman here we use column_cond to skip zone that not need read, so we don't need to process skip zone with delete conditions

morningman: OK

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -253,30 +253,37 @@ Status SegmentIterator::_get_row_ranges_from_conditions(RowRanges* condition_row
...
-    }
+    // second filter data by zone map
     for (auto& cid : cids) {
         // get row ranges by zone map of this column,
         RowRanges column_row_ranges = RowRanges::create_single(num_rows());
-        CondColumn* column_cond = nullptr;
-        if (_opts.conditions != nullptr) {
-            column_cond = _opts.conditions->get_column(cid);
-        }
+        CondColumn* column_cond = _opts.conditions->get_column(cid);
```
morningman, line 260: `_opts.conditions` may be `nullptr` here.
You need to check it before use, like the old way.

morningman (changes requested):

be/src/olap/rowset/segment_v2/segment_iterator.cpp
caiconghui: done

chaoyli (approved):

morningman (approved):
LGTM

PR #3947, opened 2020-06-25
[Enhance] Add prepare phase for some timestamp functions

### LINKED ISSUE #3946: [Enhance] Add prepare phase for some timestamp functions
**Motivation**

We find some timestamp functions are slow and very CPU cost. Like `from_unixtime`.
And stack shows most of time are spend to `find_cctz_time_zone` and RE2 regex matches.

This is because `from_unixtime()` does not have a `prepare` phase. So for each row, it has to find the timezone
by the timezone string.

morningman:
Fix: #3946 

CL:
1. Add prepare phase for `from_unixtime()`, `date_format()` and `convert_tz()` functions, to handle the format string once for all.
2. Find the cctz timezone when init `runtime state`, so that don't need to find timezone for each rows.
3. Add constant rewrite rule for `utc_timestamp()`
4. Add doc for `to_date()`
5. Comment out the `push_handler_test`, it can not run in DEBUG mode, will be fixed later.
6. Remove `timezone_db.h/cpp` and add `timezone_utils.h/cpp`

The performance shows bellow:

11,000,000 rows

SQL1: `select count(from_unixtime(k1)) from tbl1;`
Before: 8.85s
After: 2.85s

SQL2: `select count(from_unixtime(k1, '%Y-%m-%d %H:%i:%s')) from tbl1 limit 1;`
Before: 10.73s
After: 4.85s

The date string format seems still slow, we may need a further enhancement about it.


be/src/exprs/timestamp_functions.cpp
```diff
@@ -599,7 +673,7 @@ IntVal TimestampFunctions::to_unix(
 
 DateTimeVal TimestampFunctions::utc_timestamp(FunctionContext* context) {
     DateTimeValue dtv;
-    if (!dtv.from_unixtime(context->impl()->state()->timestamp_ms() / 1000, "+00:00")) {
+    if (!dtv.from_unixtime(context->impl()->state()->timestamp_ms() / 1000, "+0:00")) {
```
wutiangan, line 676: why "+00:00" modify to "+0:00"

be/src/exprs/timestamp_functions.cpp
```diff
@@ -451,20 +451,79 @@ BigIntVal TimestampFunctions::timestamp_diff(FunctionContext* ctx, const DateTim
...
     }
+
     DateTimeValue ts_value = DateTimeValue::from_datetime_val(ts_val);
-    if (ts_value.compute_format_len((const char*)format.ptr, format.len) >= 128) {
-        return StringVal::null();
+    FormatCtx* fc = reinterpret_cast<FormatCtx*>(ctx->get_function_state(FunctionContext::FRAGMENT_LOCAL));
+    if (UNLIKELY(fc == nullptr)) {
+        // prepare phase failed, calculate at runtime
+        StringVal new_fmt = convert_format(ctx, format);
+        if (DateTimeValue::compute_format_len((const char*) new_fmt.ptr, new_fmt.len) >= 128) {
```
wutiangan, line 510: in old code, first invoke function "compute_format_len" , then invoke function "convert_format", but your call sequence is the opposite , is it matter?

be/src/runtime/datetime_value.cpp
```diff
@@ -1535,7 +1521,10 @@ bool DateTimeValue::unix_timestamp(int64_t* timestamp, const std::string& timezo
     if (!find_cctz_time_zone(timezone, ctz)) {
```
wutiangan, line 1521: can ctz put it into cache like 'data foramt' function?

be/src/exprs/timestamp_functions.cpp
morningman: Doesn't matter, the old code's order is incorrect, but it doesn't matter, because `compute_format_len` will return < 128 in most cases.

be/src/exprs/timestamp_functions.cpp
morningman: My bad, I will change it back.

be/src/runtime/datetime_value.cpp
morningman: It needs a thread safe cache.
Since there is not much calls of `find_cctz_time_zone()`, I think we can improve it later.

wutiangan (approved):

yangzhg (approved):
LGTM

PR #3948, opened 2020-06-25
Support rpc_timeout property in stream load request to cancel request in fe in time when stream load request is timeout

caiconghui:
This PR is to enable cancel stream load request in fe in time when stream load request is timeout to make stream load more robust.

### DIFF
--- a/fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
+++ b/fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
@@ -186,7 +186,7 @@ public class GlobalTransactionMgr implements Writable {
             List<TabletCommitInfo> tabletCommitInfos, long timeoutMillis,
             TxnCommitAttachment txnCommitAttachment)
             throws UserException {
-        if (!db.tryWriteLock(timeoutMillis, TimeUnit.MICROSECONDS)) {
+        if (!db.tryWriteLock(timeoutMillis, TimeUnit.MILLISECONDS)) {
             throw new UserException("get database read lock timeout, database=" + db.getFullName());
         }
         try {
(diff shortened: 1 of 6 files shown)


caiconghui:
for #3949

gensrc/thrift/FrontendService.thrift
```diff
@@ -600,6 +601,7 @@ struct TLoadTxnCommitRequest {
     9: optional list<Types.TTabletCommitInfo> commitInfos
     10: optional i64 auth_code
     11: optional TTxnCommitAttachment txnCommitAttachment
+    12: required i64 thrift_rpc_timeout_ms
```
morningman, line 604: new fields should be `optional`, for compatibility.
And it need to be checked `isset` when use it.

gensrc/thrift/FrontendService.thrift
```diff
@@ -550,6 +550,7 @@ struct TStreamLoadPutRequest {
     22: optional bool isTempPartition
     23: optional bool strip_outer_array
     24: optional string jsonpaths
+    25: required i64 thrift_rpc_timeout_ms
```
morningman, line 553: optional

fe/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -185,7 +186,9 @@ public boolean commitAndPublishTransaction(Database db, long transactionId,
             List<TabletCommitInfo> tabletCommitInfos, long timeoutMillis,
             TxnCommitAttachment txnCommitAttachment)
             throws UserException {
-        db.writeLock();
+        if (!db.tryWriteLock(timeoutMillis, TimeUnit.MILLISECONDS)) {
+            throw new UserException("get database read lock timeout, database=" + db.getFullName());
```
morningman, line 190:
```suggestion
            throw new UserException("get database write lock timeout, database=" + db.getFullName());
```

morningman (changes requested):

morningman (approved):
LGTM

PR #3951, opened 2020-06-26
[Bug]Dynamic partition check interval seconds is not right

WingsGo:
dynamic partition interval's unit is ms, should multiplied by 1000.


morningman (approved):
LGTM

PR #3961, opened 2020-06-28
[Doris On ES][Bug-fix] Sync ES metadata failure after restart or upgrade FE

wuyunfeng:
ISSUE:https://github.com/apache/incubator-doris/issues/3960
PR https://github.com/apache/incubator-doris/pull/3454  introduce the caching for EsClient,  but the initialization of the `client` was only during `editlog` replay, all this work should done also during `image` replay.

This happens when restart or upgrade FE

BTW: modify a UT failure for metric


morningman (approved):
LGTM

PR #3965, opened 2020-06-28
[Dynamic Partition]Support set replication_num in dynamic partition property

WingsGo:
This CL mainly support set replication_num property in dynamic partition table.
 If dynamic_partition.replication_num is not set, the value is the same as table's default replication_num.


morningman (approved):
LGTM

Issue #3971, opened 2020-06-29
Improve the read performance of unique table


yangzhg:
In the current code, when reading the UNIQUE table, the method of merging from the lower version to the higher version one by one is adopted, but due to the characteristics of the UNIQUE table, only the highest version needs to be read, and the remaining versions can be skipped, which can be less The call of agg_update, thereby reducing memory copy and improving read performance

PR #3974, opened 2020-06-29
Improve performance of  unique table read

yangzhg:
Implements #3971 
the test table as list:
```
mysql> desc test;
+------------+---------+------+-------+---------+---------+
| Field      | Type    | Null | Key   | Default | Extra   |
+------------+---------+------+-------+---------+---------+
| rid        | BIGINT  | No   | true  | 0       |         |
| qid        | BIGINT  | No   | true  | 0       |         |
| qidDeleted | TINYINT | No   | false | 0       | REPLACE |
| type       | TINYINT | No   | false | 0       | REPLACE |
| uid        | BIGINT  | No   | false | 0       | REPLACE |
| toUid      | BIGINT  | No   | false | 0       | REPLACE |
| status     | INT     | No   | false | 0       | REPLACE |
| createTime | INT     | No   | false | 0       | REPLACE |
| source     | INT     | No   | false | 0       | REPLACE |
| misFlag    | INT     | No   | false | 0       | REPLACE |
| anonymous  | TINYINT | No   | false | 0       | REPLACE |
| uv         | TINYINT | No   | false | 1       | REPLACE |
+------------+---------+------+-------+---------+---------+
12 rows in set (0.00 sec)

mysql> select count(*) from test;
+----------+
| count(*) |
+----------+
|  1093760 |
+----------+
1 row in set (1.00 sec)
```
There is 29 versions at present
image
I run the query `select sum(uv) from test` for 10 times,
the average ScanTime reduced from `9s277ms`  to `8s206ms`

### DIFF
--- a/be/src/olap/reader.cpp
+++ b/be/src/olap/reader.cpp
@@ -420,7 +420,7 @@ OLAPStatus Reader::_unique_key_next_row(RowCursor* row_cursor, MemPool* mem_pool
         // skip the lower version rows;
         while (nullptr != _next_key) {
             auto res = _collect_iter->next(&_next_key, &_next_delete_flag);
-             if (res != OLAP_SUCCESS) {
+            if (res != OLAP_SUCCESS) {
                 if (res != OLAP_ERR_DATA_EOF) {
                     return res;
                 }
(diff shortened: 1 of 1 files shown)


morningman (approved):
LGTM

PR #3975, opened 2020-06-29
[UT] resolve metric ut fails for Doris FE

wuyunfeng:
Two extra metric were added in Doris, but the logic of `MetricsTest.java` was outdated


morningman (approved):
LGTM

PR #3986, opened 2020-06-30
[Config]Allower use to config different thrift server model

WingsGo:
Doris only support TThreadPoolServer model in thrift server, but the
server model is not effective in some high concurrency scenario, so this
PR introduced new config to allower use to choose different server model
by their scenario.


WingsGo:
For #3987

morningman (commented):
Better to use a `string` instead of `int` type of this config.
And  TSimpleServer seems unnecessary.

WingsGo:
> Better to use a `string` instead of `int` type of this config.
> And TSimpleServer seems unnecessary.

use string is better, but Config have no check logic, we can only set default server model when use have typo error like `Threadd`,In this case, user may think they use a `Threaded` server model, but it is not what he think, use int type is to avoid situation like this.

morningman:
> > Better to use a `string` instead of `int` type of this config.
> > And TSimpleServer seems unnecessary.
> 
> use string is better and easy to understand, but Config have no check logic, we can only set default server model when use have typo error like `Threadd`,In this case, user may think they use a `Threaded` server model, but it is not what he think, use int type is to avoid situation like this.

Actually, the config SHOULD has a validation logic. We may add it later.
So I suggest to keep the config in a friendly way, and we can add the checking later.

WingsGo:
> > > Better to use a `string` instead of `int` type of this config.
> > > And TSimpleServer seems unnecessary.
> > 
> > 
> > use string is better and easy to understand, but Config have no check logic, we can only set default server model when use have typo error like `Threadd`,In this case, user may think they use a `Threaded` server model, but it is not what he think, use int type is to avoid situation like this.
> 
> Actually, the config SHOULD has a validation logic. We may add it later.
> So I suggest to keep the config in a friendly way, and we can add the checking later.

It look like the best solution, done.

morningman (approved):
LGTM

PR #3991, opened 2020-07-01
Fix the problem of mem exec, when analytic eval node need to spill to disk with a low mem limit

HappenLee:
[Bug] Fix the problem of mem exec, when analytic eval node need to spill to disk with a low mem limit. And clear_reservations
of Analytic node reservation of block manager.
[Running Profile] Add Spilled flag in Running Profile, when Analytic eval node and sort node spill to Disk.

### DIFF
--- a/be/src/exec/analytic_eval_node.cpp
+++ b/be/src/exec/analytic_eval_node.cpp
@@ -710,10 +710,9 @@ Status AnalyticEvalNode::process_child_batch(RuntimeState* state) {
     // Transfer resources to _prev_tuple_pool when enough resources have accumulated
     // and the _prev_tuple_pool has already been transfered to an output batch.
 
-    // The memory limit of _curr_tuple_pool is set by the fixed value MAX_TUPLE_POOL_SIZE
+    // The memory limit of _curr_tuple_pool is set by the fixed value 
     // The size is specified as 8MB, which is used in the extremely strict memory limit.
-    // Eg: exec_mem_limit < 100MB may cause memory exeecded limit problem. So change it to half of block size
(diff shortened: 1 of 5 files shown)


be/src/runtime/spill_sorter.cc
```diff
@@ -1109,6 +1110,7 @@ Status SpillSorter::add_batch(RowBatch* batch) {
             // The current run is full. Sort it and begin the next one.
             RETURN_IF_ERROR(sort_run());
             RETURN_IF_ERROR(_sorted_runs.back()->unpin_all_blocks());
+	    _spilled = true;
```
wutiangan, line 1113: Align spaces

be/src/exec/analytic_eval_node.cpp
```diff
@@ -710,7 +709,13 @@ Status AnalyticEvalNode::process_child_batch(RuntimeState* state) {
 
     // Transfer resources to _prev_tuple_pool when enough resources have accumulated
     // and the _prev_tuple_pool has already been transfered to an output batch.
-    if (_curr_tuple_pool->total_allocated_bytes() > MAX_TUPLE_POOL_SIZE &&
+
+    // The memory limit of _curr_tuple_pool is set by the fixed value MAX_TUPLE_POOL_SIZE
```
wutiangan, line 713: there is no MAX_TUPLE_POOL_SIZE， you deleted it before

HappenLee: ok，i will change the comment

morningman (approved):
LGTM

Issue #3930, opened 2020-06-23
[Proposal] Doris support version column for REPLACE aggregate type


Youngwb:
### BackGround
Doris currently use REPLACE to update data, but the replacement order cannot be guaranteed for the data import of the same batch. The user needs to guarantee that there is no same key column in the imported data of the same batch to guarantee the replacement order, which is very inconvenient for the user. To solve this problem, we can use a **version** column to specify the replacement order.

### Goal
The user specifies a **version column** when creating the table. Doris relies on this column to update the data of REPLACE type. The larger version column data can REPLACE the data of the smaller version column, while the data of the smaller version column cannot REPLACE the larger version column data.

### Create Table Interface
```
CREATE TABLE `test` (
`id` bigint(20) NOT NULL,
`date` date NOT NULL,
`group_id` bigint(20) NOT NULL,
`version` int MAX NOT NULL,
`keyword` varchar(128) REPLACE NOT NULL,
`clicks` bigint(20) SUM NULL DEFAULT "0" ,
`cost` bigint(20) SUM NULL DEFAULT "0" 
) ENGINE=OLAP
AGGREGATE KEY(`id`, `date`, `group_id`)
DISTRIBUTED BY HASH(`id`) BUCKETS 16
PROPERTIES (
  "replace_version_column" = "version"
);
```
When creating a table, the user simply adds the **replace_version_column** attribute in PROPERTIES to identify the version column, which requires a MAX aggregation type to ensure that only the largest version column is retained for the same key column.

### Query 
When a user's query does not contain the REPLACE column, the original logic follows. When a user's query contains REPLACE columns, BE needs to extend the Version column on which the REPLACE column depends, and compare the value column when it is aggregated. These operations can be done by extending **Reader return columns**, and in FE，the **isPreAggregation** is OFF because of the REPLACE column is value column in StorageEngine
，which means the storage engine needs to aggregate the data before returning to scan node，so we can guarantee that the same key columns will be aggregated in Reader.

### Compaction
Base and Cumulative Compaction use Reader to aggregate data, and it use all tablet columns as return columns, so similar to the query processing, we can use Reader for replace based on version columns.

### Load
With the same batch of data load, Doris uses one or more **MemTable**.  We need to ensure that the same key column in one MemTable, columns of REPLACE type are replaced with version column, while the data in different MemTable is not guaranteed in LOAD because Query and Compaction guarantee the order of replacement.

### RollUp
[...]

PR #3993, opened 2020-07-01
Keep order of read from segment consist with the write order

### LINKED ISSUE #3989: UNIQUE_KEYS table data may not stable when user load data with multiple same key lines
When importing data into UNIQUE_KEYS table , if there are rows with the same key in the data, when the amount of data is large, the rows with the same key may be distributed in different segments. The priority_queue used when reading the rowset is an unstable sort, which may occur the lastest imported data row is read first, resulting is inconsistent between the final state of the data and the imported state。
The comparator of merge only compare the row not include segment id.
image

yangzhg:
Fixes #3989 
Add segemt id to the comparator

### DIFF
--- a/be/src/olap/generic_iterators.cpp
+++ b/be/src/olap/generic_iterators.cpp
@@ -142,6 +142,8 @@ public:
 
     int is_partial_delete() const { return _block.delete_state() == DEL_PARTIAL_SATISFIED; }
 
+    uint64_t data_id() const { return _iter->data_id(); }
+
 private:
     // Load next block into _block
     Status _load_next_block();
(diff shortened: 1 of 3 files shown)


be/src/olap/generic_iterators.cpp
```diff
@@ -227,8 +229,13 @@ class MergeIterator : public RowwiseIterator {
         bool operator()(const MergeIteratorContext* lhs, const MergeIteratorContext* rhs) const {
             auto lhs_row = lhs->current_row();
             auto rhs_row = rhs->current_row();
-
-            return compare_row(lhs_row, rhs_row) > 0;
+            int cmp_res = compare_row(lhs_row, rhs_row);
+            if (cmp_res !=  0) {
+                return cmp_res > 0;
+            }
+            // if row cursors equal, compare segment id.
```
wutiangan, line 236: add some comments why  sort by row asc, and sort by segment_id desc.

chaoyli:
Do you can reproduce the problem using UNIQUE KEY?
Because I have a little double why this bug exists for so long time.

yangzhg:
@chaoyli this problem is really exists for a long time. the issue #3930 describe the load cannot keep order and this pr fix read cannot keep order

morningman (approved):
LGTM

Issue #2581, opened 2019-12-26
[Feature][Cache] Doris caches query results based on partition


marising:
1.If the table is partitioned by date and only the data of the current day is updated, the data before the current day will not be updated. When querying 30 days of data, 29 days can be queried from the cache.

For example, today is 20191226,Data is being imported in real time through StreamLoad.

**SELECT event_date，COUNT( event_id ) AS event_count FROM music_event WHERE event_date>=20191127 and event_date <= 20191226 GROUP BY event_date ORDER BY event_date；**

The row batch of event_date and event_count  of 20191127-20191225 can be obtained from the memory cache, and only the data of 20191226 can be queried from the physical table.

If the first query fails to hit the cache, the query results will be cached.

2.In another case, although the data is not queried by partition key, the data is updated by day. For example, userprofile uses UserID as partition key to query the number of users in each country。

**SELECT country，COUNT( UserID ) FROM UserProfile GROUP BY country;**

This query result can be cached.

Therefore, queries that are not updated in real time can be cached, or data that is partitioned on a daily basis can be cached on a partition by partition basis only when the most recent partition is updated.

This feature will reduce query time and improve cluster QPS.

The specific design and use details will be supplemented according to the development progress.

kangkaisen:
Great Job！Looking forward to your design and PR.

PR #3994, opened 2020-07-01
SQL result cache and partition cache

marising:
#2581

## Solutions

This cache give priority to ensuring data consistency. On this basis, it refines the cache granularity and improves the hit rate. Therefore, it has the following characteristics:

* User don't need to worry about data consistency. Cache invalidation is controlled by version. The cached data is consistent with the data queried from be
* Without additional components and costs, the cache results are stored in the memory of be, and user can adjust the cache memory size according to their needs
* Two caching strategies are implemented, sql and partition cache, which are more granular
* The cache algorithm in be is an improved LRU

### SQLCache
Sql cache stores and fetches the cache according to the **SQL signature**, **partition ID** of the query table, and the **latest version** of the partition. 

The combination of the three determines a cache dataset. If any one of them changes, such as SQL changes, query fields or conditions are not the same, or the version after data update changes, the cache will not be hit.

If multiple tables are joined, the latest partition ID and the latest version number are used. If one of the tables is updated, the partition ID or version number will be different, and the cache will not be hit.

Sql cache is more suitable for the scenario of T + 1 update. When the data is updated in the morning, the results of the first query are obtained from be and put into the cache, and the subsequent same query is obtained from the cache. Real time update data can also be used, but there may be a low hit rate. Please refer to the following partitioncache.

### PartitionCache
Query the number of users per day in the last 7 days, such as partitioning by date, data is only written to the current partition, and the data of other partitions other than that day are fixed. Under the same query SQL, query a partition that does not update The indicators are fixed. As follows, the number of users in the 7 days before the query on 2020-03-09, the data from 2020-03-03 to 2020-03-07 comes from the cache, the first query from 2020-03-08 comes from the partition, and the subsequent queries come from the cache , 2020-03-09 because of the non-stop writing that day, so from the partition.

Therefore, querying the data of N days, the latest D days of the data update, each day is only a query with a similar date range, only need to query D partitions, the other parts are all from the cache, which can effectively reduce the cluster load and reduce the query time.

```
MySQL [(none)]> SELECT eventdate,count(userid) FROM testdb.appevent WHERE eventdate>="2020-03-03" AND eventdate<="2020-03-09" GROUP BY eventdate ORDER BY eventdate;
+------------+-----------------+
| eventdate  | count(`userid`) |
+------------+-----------------+
| 2020-03-03 |              15 | //From cache
| 2020-03-04 |              20 | ...
| 2020-03-05 |              25 |
| 2020-03-06 |              30 |
| 2020-03-07 |              35 |
| 2020-03-08 |              40 | //From cache
| 2020-03-09 |              25 | //From disk
+------------+-----------------+
7 rows in set (0.02 sec)
```

## Reference
For more information, please read [partition_cache.md](https://github.com/apache/incubator-doris/compare/master...marising:partition_cache_0.3?expand=1#diff-bfa91ffdf5960b0af00c5775fa9a2db2)

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -549,6 +549,15 @@ namespace config {
 
     // Soft memory limit as a fraction of hard memory limit.
     CONF_Double(soft_mem_limit_frac, "0.9");
+    
+    // Set max cache's size of query results, the unit is M byte
+    CONF_Int32(cache_max_size, "256"); 
+
+    //Cache memory is pruened when reach cache_max_size + cache_elasticity_size
+    CONF_Int32(cache_elasticity_size, "128");
+
+    //Maximum number of cache partitions corresponding to a SQL
+    CONF_Int32(cache_max_partition_count, "1024");
 } // namespace config
 
 } // namespace doris
--- a/be/src/runtime/cache/result_cache.h
+++ b/be/src/runtime/cache/result_cache.h
@@ -1,0 +1,117 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#ifndef DORIS_BE_SRC_RUNTIME_RESULT_CACHE_H
+#define DORIS_BE_SRC_RUNTIME_RESULT_CACHE_H
+
+#include <boost/thread.hpp>
+#include <cassert>
+#include <cstdio>
+#include <cstdlib>
+#include <exception>
+#include <iostream>
+#include <list>
+#include <map>
+#include <mutex>
+#include <shared_mutex>
+#include <thread>
+
+#include "common/config.h"
+#include "runtime/cache/cache_utils.h"
+#include "runtime/cache/result_node.h"
+#include "runtime/mem_pool.h"
+#include "runtime/mem_tracker.h"
+#include "runtime/row_batch.h"
+#include "runtime/tuple_row.h"
+
+namespace doris {
+
+typedef std::unordered_map<UniqueId, ResultNode*> ResultNodeMap;
+
+// a doubly linked list class
+class ResultNodeList {
+public:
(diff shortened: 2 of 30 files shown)


be/src/common/config.h
```diff
@@ -536,6 +536,15 @@ namespace config {
     // Whether to continue to start be when load tablet from header failed.
     CONF_Bool(ignore_load_tablet_failure, "false");
 
+    // Set max cache's size of query results, the unit is M byte
+    CONF_Int32(cache_max_size, "256"); 
+
+    //Cache memory is pruened when reach cache_max_size + cache_elasticity_size
```
blackfox1983, line 542: cache memory will be shrinked?
i do not understand what does this mean....

be/src/runtime/cache/result_cache.h
```diff
@@ -0,0 +1,117 @@
...
+#include "runtime/mem_pool.h"
+#include "runtime/mem_tracker.h"
+#include "runtime/row_batch.h"
+#include "runtime/tuple_row.h"
+
+namespace doris {
+
+typedef std::unordered_map<UniqueId, ResultNode*> ResultNodeMap;
+
+// a doubly linked list class
```
blackfox1983, line 45: why not use std::list?

docs/zh-CN/administrator-guide/partition_cache.md
```diff
@@ -0,0 +1,205 @@
...
+fe.conf添加enable_sql_cache=true
+```
+vim fe/conf/fe.conf
+enable_sql_cache=true
+```
+在MySQL命令行中设置变量
+```
+MySQL [(none)]> set [global] enable_sql_cache=true;
+```
+注：globa是全局变量，不加指当前会话变量
```
wutiangan, line 123:
```suggestion
注：global是全局变量，不加指当前会话变量
```

fe/src/main/java/org/apache/doris/qe/cache/CacheAnalyzer.java
```diff
@@ -0,0 +1,449 @@
...
+     * 4、Partition key must be included in the group by clause
+     * 5、Where clause must contain only one partition key predicate
+     * CacheMode.Sql
+     * xxx FROM user_profile, updated before Config.last_version_interval_time
+     * CacheMode.Partition, partition by event_date, only the partition of today will be updated.
+     * SELECT xxx FROM app_event WHERE event_date >= 20191201 AND event_date <= 20191207 GROUP BY event_date
+     * SELECT xxx FROM app_event INNER JOIN user_Profile ON app_event.user_id = user_profile.user_id xxx
+     * SELECT xxx FROM app_event INNER JOIN user_profile ON xxx INNER JOIN site_channel ON xxx
+     */
+    public void checkCacheMode(long now) {
```
blackfox1983, line 176: Check whether a SQL hit the cache in explain?
otherwise we will check the log, it's inconvenient

be/src/common/config.h
marising: In order to avoid frequent cache cleaning and keep high hit rate, set two config item, cache_max_size  and cache_elasticity_size，such as default config value,  when reach 256M+128M，cache memory is pruned to 256M

marising:
I split the PR and submit the be part first
#4005

PR #3996, opened 2020-07-01
Add log for loading image

xy720:
When fe load image failed, more logs should be printed to help users analyze errors.

### DIFF
--- a/fe/src/main/java/org/apache/doris/catalog/Catalog.java
+++ b/fe/src/main/java/org/apache/doris/catalog/Catalog.java
@@ -1844,7 +1844,6 @@ public class Catalog {
             LOG.info("finished replay transactionState from image");
             return newChecksum;
         }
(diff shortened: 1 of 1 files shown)


fe/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -1833,8 +1841,10 @@ public long loadTransactionState(DataInputStream dis, long checksum) throws IOEx
             int size = dis.readInt();
             long newChecksum = checksum ^ size;
             globalTransactionMgr.readFields(dis);
+            LOG.info("finished replay transactionState from image");
             return newChecksum;
         }
+        LOG.info("finished replay transactionState from image");
```
morningman, line 1847: This log can  be removed.

morningman (approved):
LGTM

PR #3999, opened 2020-07-02
Fix some code & comments

vagetablechicken:
TPlanExecParams::volume_id is never used, so delete the `print_volume_ids()` function.
Fix log, and log if PlanFragmentExecutor::open() returns error.
Fix some comments


morningman (approved):
LGTM

PR #4000, opened 2020-07-02
[Planner]Apply inner join condition reorganization rule

### LINKED ISSUE #3908: Column can not be resloved  after reorderTable

wutiangan:
fix #3908

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
+++ b/fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
@@ -816,9 +816,103 @@ public class SelectStmt extends QueryStmt {
 
         // can not get AST only with equal join, MayBe cross join can help
         fromClause_.clear();
+        List<Pair<Expr, Expr>> misMatchJoinConditions = Lists.newArrayList();
+        List<TupleId> tupleIds = Lists.newArrayList();
+
         for (Pair<TableRef, Long> candidate : candidates) {
-            fromClause_.add(candidate.first);
+            TableRef tableRef = candidate.first;
+            fromClause_.add(tableRef);
+            tupleIds.add(tableRef.desc.getId());
+            applyInnerJoinConditionReorganizeRule(tableRef, tupleIds, misMatchJoinConditions);
         }
+
+        Preconditions.checkState(misMatchJoinConditions.isEmpty());
+    }
+
+    /**
+     *  After adjusting the order of tables, it is necessary to reselect the join condition of each table,
+     *  otherwise some columns will not be recognized by the current table space
+     * @param tableRef:
+     * @param tupleIds: TableSpaces visible up to now
+     * @param misMatchJoinConditions: join conditions no matching yet, the first is fatherExpr, the second is sonExpr need to replace
+     */
+    void applyInnerJoinConditionReorganizeRule(TableRef tableRef, List<TupleId> tupleIds,
+                List<Pair<Expr, Expr>> misMatchJoinConditions) {
+        List<Pair<Expr, Expr>> tempMisMatchConjuncts = Lists.newArrayList();
+
+        // Extract all join conditions from joinOnClause to misMatchJoinConditions,
+        // which cannot be bound by existing tableSpaces(tupleIds)
+        findUnboundedJoinConditions(tableRef.getOnClause(), tupleIds, misMatchJoinConditions);
+
+        for (Pair<Expr, Expr> exprPair : misMatchJoinConditions) {
+            Expr father = exprPair.first;
+            Expr expr = exprPair.second;
+            if (!expr.isBoundByTupleIds(tupleIds)) {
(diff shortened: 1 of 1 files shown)


fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
```diff
@@ -816,9 +816,103 @@ protected void reorderTable(Analyzer analyzer) throws AnalysisException {
...
+        // Extract all join conditions from joinOnClause to misMatchJoinConditions,
+        // which cannot be bound by existing tableSpaces(tupleIds)
+        findUnboundedJoinConditions(tableRef.getOnClause(), tupleIds, misMatchJoinConditions);
+
+        for (Pair<Expr, Expr> exprPair : misMatchJoinConditions) {
+            Expr father = exprPair.first;
+            Expr expr = exprPair.second;
+            if (!expr.isBoundByTupleIds(tupleIds)) {
+                tempMisMatchConjuncts.add(new Pair<Expr, Expr>(null, expr));
+                if (father.getChild(0) == expr) {
```
spaces-X, line 852: father may be null here.

fe/src/main/java/org/apache/doris/analysis/SelectStmt.java
```diff
@@ -816,9 +816,103 @@ protected void reorderTable(Analyzer analyzer) throws AnalysisException {
...
+                    // no need to reAnalysis
+                    tableRef.getOnClause().analysisDone();
+                }
+            }
+        }
+        misMatchJoinConditions.clear();
+        misMatchJoinConditions.addAll(tempMisMatchConjuncts);
+    }
+
+    boolean ifExprIsCompandAndPredicate(Expr expr) {
```
morningman, line 877:
```suggestion
    boolean ifExprIsCompoundAndPredicate(Expr expr) {
```

spaces-X:
@morningman @wutiangan 
Fix the NullPointerException of the variable `father` in this [PR ](https://github.com/apache/incubator-doris/pull/4145)

wuyunfeng (changes requested):
Suggest you change the title for this PR, You should describe what you did in one sentence for your PR. Thanks.

BTW the title for this PR is so disgusted

wuyunfeng:
@wutiangan Are you still work on this PR?

jackwener:
Please rebase the `master` and add an UT for this PR.

PR #4003, opened 2020-07-02
Fix except node child not order correctly

### LINKED ISSUE #3995: UNION and EXCEPT result is wrong
```
SELECT 449
UNION
  SELECT 670
EXCEPT
  SELECT 449;
```
Current doris result is empty. correct result should be 670.

yangzhg:
Fixes #3995 
## Why does it happen
When SetOperations encounters that the previous node needs Aggregate, the timing of add AggregationNode is wrong. You should add AggregationNode first before add other children.
## Why doesn't intersect and union have this problem
intersect and union conform to the commutation law, so it doesn't matter if the order is wrong
## Why this problem has not been tested before
In the previous test case, not cover the previous node was not AggregationNode

### DIFF
--- a/fe/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
+++ b/fe/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
@@ -1621,6 +1621,23 @@ public class SingleNodePlanner {
             default:
                 throw new AnalysisException("not supported set operations: " + operation);
         }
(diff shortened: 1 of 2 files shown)


morningman (commented):
Could you add a ut for this?

EmmyMiao87:
Approved

wutiangan (approved):

EmmyMiao87 (approved):
LGTM

Issue #4007, opened 2020-07-02
[Bug]Schema Change from int type to varchar failed because miss some logic check


WingsGo:
**Describe the bug**
Schema Change from int type to varchar failed because miss some logic check

**To Reproduce**
1. create a table with int type

```
CREATE TABLE `table2` (
  `siteid` int(11) NULL DEFAULT "10" COMMENT "",
  `k` smallint(6) NULL COMMENT "",
  `v` int NULL COMMENT ""
) ENGINE=OLAP
DUPLICATE KEY(`siteid`, `k`)
COMMENT "OLAP"
DISTRIBUTED BY HASH(`siteid`) BUCKETS 2
PROPERTIES (
"replication_num" = "1",
"in_memory" = "false",
"storage_format" = "DEFAULT"
);
```

2. do schema change `alter table table2 modify column v varchar(32);

The reason why schema change failed is because in following judgement, we forget to add supported convert type in if statement.

https://github.com/apache/incubator-doris/blob/5ade21b55d773b4160dfc0070f14bdc31a44981f/be/src/olap/schema_change.cpp#L340-L373

**Expected behavior**
schema change to varchar type success.

PR #4009, opened 2020-07-02
[Bug]Fix some schema change not work right

WingsGo:
This CL including:
1. Fix numeric type to varchar type not work right bug
2. remove convert from largeint type to double type in doc.
3. change logic check in FE to support convert decimal type to varchar type which has already been implement in BE.


WingsGo:
For #4007

chaoyli:
It's better add more discrete message in commit info like you issue.

WingsGo:
> It's better add more discrete message in commit info like you issue.

Ok, done, and I tested the schema change supported in doc, but I found that convert LargetInt to Double is not support in code implement, maybe you can confirm that why it is added to doc in case It is supported but there is some bug exists?

WingsGo:
In order to make review easiler, I will not add code format until review ended.

be/src/olap/schema_change.cpp
```diff
@@ -194,6 +194,64 @@ ColumnMapping* RowBlockChanger::get_mutable_column_mapping(size_t column_index)
...
+ConvertTypeResolver::ConvertTypeResolver() {
+    // supported type convert should annotate in doc:
+    // http://doris.incubator.apache.org/master/zh-CN/sql-reference/sql-statements/Data%20Definition/ALTER%20TABLE.html#description
+    // from varchar type
+    add_convert_type_mapping<OLAP_FIELD_TYPE_VARCHAR, OLAP_FIELD_TYPE_TINYINT>();
+    add_convert_type_mapping<OLAP_FIELD_TYPE_VARCHAR, OLAP_FIELD_TYPE_SMALLINT>();
+    add_convert_type_mapping<OLAP_FIELD_TYPE_VARCHAR, OLAP_FIELD_TYPE_INT>();
+    add_convert_type_mapping<OLAP_FIELD_TYPE_VARCHAR, OLAP_FIELD_TYPE_BIGINT>();
+    add_convert_type_mapping<OLAP_FIELD_TYPE_VARCHAR, OLAP_FIELD_TYPE_LARGEINT>();
+    add_convert_type_mapping<OLAP_FIELD_TYPE_VARCHAR, OLAP_FIELD_TYPE_FLOAT>();
```
morningman, line 231: is VARCHAR to DOUBLE and DATETIME support?

morningman (commented):
comment

be/src/olap/schema_change.cpp
WingsGo: VAHRCHAR->DOUBLE is supported. Support VARCHAR->DATETIME maybe supported in next PR.

WingsGo:
I have tested by following:

```
CREATE TABLE `test_schema_change` (
	`k1` int NULL,
	`v1` varchar NULL
) ENGINE=OLAP
DUPLICATE KEY(`k1`)
PROPERTIES (
 "replication_num" = "1"
);
alter table test_schema_change modify column v1 tinyint;
alter table test_schema_change modify column v1 varchar;
alter table test_schema_change modify column v1 smallint;
alter table test_schema_change modify column v1 varchar;
alter table test_schema_change modify column v1 int;
alter table test_schema_change modify column v1 varchar;
alter table test_schema_change modify column v1 bigint;
alter table test_schema_change modify column v1 varchar;
alter table test_schema_change modify column v1 largeint;
alter table test_schema_change modify column v1 varchar;
alter table test_schema_change modify column v1 float;
alter table test_schema_change modify column v1 varchar;
alter table test_schema_change modify column v1 double;
alter table test_schema_change modify column v1 varchar;
alter table test_schema_change modify column v1 date;
```

morningman (dismissed):
LGTM

chaoyli:
> > It's better add more discrete message in commit info like you issue.
> 
> Ok, done, and I tested the schema change supported in doc, but I found that convert LargetInt to Double is not support in code implement, maybe you can help me confirm that why it is added to doc in case It is supported but there is some bug exists?

I test the case, it's forbidden by FE to change LARGEINT to DOUBLE. Because LARGEINT is int128_t and double is 2^52 and will lose information.
So you can remove from the doc.

WingsGo:
> > > It's better add more discrete message in commit info like you issue.
> > 
> > 
> > Ok, done, and I tested the schema change supported in doc, but I found that convert LargetInt to Double is not support in code implement, maybe you can help me confirm that why it is added to doc in case It is supported but there is some bug exists?
> 
> I test the case, it's forbidden by FE to change LARGEINT to DOUBLE. Because LARGEINT is int128_t and double is 2^52 and will lose information.
> So you can remove from the doc.

Ok, I removed it, thx~

morningman:
@chaoyli  Any comments?

chaoyli (approved):

PR #4010, opened 2020-07-02
[ShowIndex] Make Show Index stmt act same as MySQL behavior

morningman:
`SHOW INDEX FROM db2.tbl1 FROM db1;` will be same as
`SHOW INDEX FROM db1.tbl1;`


yangzhg (approved):
LGTM

Issue #3976, opened 2020-06-29
Lost data when load data to partitions created by old version doris


liutang123:
In #2358, we remove an important logic of `_check_added_rowset`:
```Version version = {rowset->start_version(), rowset->end_version()};	
    RowsetSharedPtr exist_rs = get_rowset_by_version(version);	
    // if there exist a rowset with version_hash == 0, should delete it	
    if (exist_rs != nullptr && exist_rs->version_hash() == 0) {	
        vector<RowsetSharedPtr> to_add;	
        vector<RowsetSharedPtr> to_delete;	
        to_delete.push_back(exist_rs);	
        RETURN_NOT_OK(modify_rowsets(to_add, to_delete));	
    }
```
This logic has a very important role to make BE compatible with 0.10.

In our cluster, there are many partitions created by 0.10, and the tablets have two initial RSs [(0-1), 0] and [(2-2), 0].
We updated the cluster to 0.11 and then to 0.12.
Now, we load data to these partitions. BE will create a new RS. When pubish version, FE will try to let BE to set this RS to [(2-2), XXX] and add it to tablet. But, tablet has a same version RS [(2-2), 0]. So, the new RS will not be added to the tablet and the data will **lost**.

Q1: Why we remove version_hash in 0.13?
Q2: How to make doris to be compatible with previous version in this case?
Q3: I use `/api/meta/header` to see these tablet and find that like follows:
```
{
    "tablet_id": 218278919,
    "schema_hash": 39606704,
    "creation_time": 1576654229,
    ...
"rs_metas": [
        {
            "rowset_id": 0,
            "tablet_id": 218278919,
            "tablet_schema_hash": 39606704,
            ...
           "empty": true,
           "creation_time": 1576654229,
},
        {
            "rowset_id": 0,
            "tablet_id": 218278919,
            "tablet_schema_hash": 39606704,
            ...
           "empty": true,
            "creation_time": 1576654229,
         }
}
```
Is it possible to use the `creation_time` to determine whether a RS is the initial RS?

liutang123:
@chaoyli 
cc @imay @gaodayue

chaoyli:
Previously, the version_hash is used to differentiate the same load with the same version.
In old days, FE will send version to BE before load finished. Assuming I have three BE(A, B, C) nodes.
If A succeed, but B, C failed. The version belongs to load will failed. So The version will be reused by another load.
Under this circumstance, version_hash is useful to differentiate the same version in Node A.

[...]

PR #4011, opened 2020-07-02
Remove rs when version conflict when insert to a tablet

liutang123:
For #3976 
In this PR, when insert to a tablet, if the tablet has a different rs with the same version, remove it.

### DIFF
--- a/be/src/olap/tablet.cpp
+++ b/be/src/olap/tablet.cpp
@@ -190,6 +190,8 @@ OLAPStatus Tablet::add_rowset(RowsetSharedPtr rowset, bool need_persist) {
     if (_contains_rowset(rowset->rowset_id())) {
         return OLAP_SUCCESS;
     }
+    // If has same version rs and it is the max version rs of this tablet, remove it.
+    _drop_rowset_if_version_conflict(rowset);
     // Otherwise, the version shoud be not contained in any existing rowset.
     RETURN_NOT_OK(_contains_version(rowset->version()));
 
@@ -305,6 +307,10 @@ OLAPStatus Tablet::add_inc_rowset(const RowsetSharedPtr& rowset) {
     if (_contains_rowset(rowset->rowset_id())) {
         return OLAP_SUCCESS;
     }
+
+    // If has same version rs and it is the max version rs of this tablet, remove it.
+    _drop_rowset_if_version_conflict(rowset);
+
     RETURN_NOT_OK(_contains_version(rowset->version()));
 
     RETURN_NOT_OK(_tablet_meta->add_rs_meta(rowset->rowset_meta()));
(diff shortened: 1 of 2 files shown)


liutang123:
@chaoyli

be/src/olap/tablet.cpp
```diff
@@ -826,6 +832,33 @@ void Tablet::_print_missed_versions(const std::vector<Version>& missed_versions)
...
+
+    RowsetSharedPtr exist_rs = get_rowset_by_version(insert_version);
+    // if a rs is created by doris 0.10 exists, remove it.
+    if (exist_rs != nullptr) {
+        vector<RowsetSharedPtr> to_add;
+        vector<RowsetSharedPtr> to_delete;
+        to_delete.push_back(exist_rs);
+        modify_rowsets(to_add, to_delete);
+        std::stringstream ss;
+        ss << "Remove rs " << exist_rs->rowset_id().to_string();
```
chaoyli, line 853: LOG(WARNING) << "Remove rs " << exist_rs->rowset_id().to_string()
                             << " in " << full_name()
                             << " for version conflict with " << rowset->rowset_id().to_string()
There is no necessity to use std::stringstream.

be/src/olap/tablet.cpp
```diff
@@ -826,6 +832,33 @@ void Tablet::_print_missed_versions(const std::vector<Version>& missed_versions)
     LOG(WARNING) << ss.str();
 }
 
+bool Tablet::_drop_rowset_if_version_conflict(const RowsetSharedPtr& rowset) {
```
chaoyli, line 835: There has a problem in this place.
modify_rowsets() only change the TabletMeta in memory. But not change the meta in disk.
So you should perform do_tablet_meta_checkpoint to save the meta and remove the rowsetmeta in RowsetMetaManager.

chaoyli (changes requested):

chaoyli:
> @chaoyli Now, should we prohibit insertion when `OLAP_ERR_PUSH_VERSION_ALREADY_EXIST` appears?

Your function will add little cost to remove OLAP_ERR_PUSH_VERSION_ALREADY_EXIST.
The cost will reflect on save the meta.

be/src/olap/tablet.cpp
liutang123: Why don't we need to do `do_tablet_meta_checkpoint` when adding a rs, but we need to do it when we delete a rs?

chaoyli: Because before add the rowset, it has already stills in RowsetMetaManager.
It only make it visible in memory.
After that, if you want to delete the rowset, you should delete from the RowsetMetaManager and add it to TabletMeta and save it.

PR #4012, opened 2020-07-03
[Doris On ES][Refactor] refactor and enchanment ES sync meta logic

wuyunfeng:
After PR https://github.com/apache/incubator-doris/pull/3454 was merged, we should refactor and reorganize some logic for long-term sustainable iteration for Doris On ES.
To facilitate code review，I would divided into this work  to multiple PRs (some other WIP work I also need to think carefully)

This PR include:
1. introduce `SearchContext` for all state we needed 
2. divide meta-sync logic into three phase
3. modify some logic processing
4. introduce version detect logic for future using


morningman (approved):
LGTM

Issue #3966, opened 2020-06-29
[Proposal]Optimize the memory consumption of TransactionMgr


WingsGo:
In production enviorment, the data loading is 7*24 with lots of txns, but we found that the TransactionMgr costs too much memeory, which will leads OOM.If we want to improve the concurrent of stream load, we must optimize the memory usage in order to support frequent data load.There are some way such as clean up useless data after txn finished #3130, reduce the interval of clean txn, use ZGC to recycle memory in a short time....But It is not enough, I found FE will OOM accidental cause by TransactionMgr usage.
There are two way to reduce the memory usage, one is change the data struct in DatabaseTransactionMgr to save memeory, the other is scattered the Txn to other follower FE to share the memory usage, I will invetigate the feasibility of two solutions, any advice or solution is welcome.

WingsGo:
By the way, too much DatabaseTransactionMgr will cause some more serious accident.For example, if DatabaseTransactionMgr costs too much memory, checkpoint image will not work because it will check java heap memory in order to create image in memory, so when FE down cause by OOM, it will replay too much EditLog, in the worst case, it will replay more than one hour, It means that data load is not available in the meantime.

WingsGo:
I found that most of memory usage in DatabaseTransactionMgr is a HashMap struct idToFinalStatusTransactionState, following shows that 270W TransactionState object almost cost 50G memory, but some data in TransactionState is useless, so we have no need to save the data.

```
Label | Number of Objects | Used Heap Size | Retained Heap Size
-- | -- | -- | --
java.util.HashMap$NodeFirst 10 of 2,718,787 objects | 2,718,787 | 130,501,776 | 53,717,135,376
```

wangbo:
I think the ```Check Point Thread``` can run in follower.
It's not necessary for ```Master FE``` to run it.

PR #4013, opened 2020-07-03
[Bug][Memroy Leak]Fix bug TransactionState is not clear from idToFinalStatusTransactionState

WingsGo:
This CL includes:
1. Memory leak because transactionState is not removed.
2. Extacting the clear logic to method to avoid forget.


WingsGo:
Ref #3966

morningman (approved):
LGTM

PR #4014, opened 2020-07-03
Add framework of mv selector

### LINKED ISSUE #3344: [Proposal] Materialized View 2.0

EmmyMiao87:
This commit mainly supports creating bitmap_union, hll_union, and count materialized views.
* The main changes are as follows:
1. When creating a materialized view, doris judge the semantic analysis of the newly supported aggregate function.
Only bitmap_union(to_bitmap(column)), hll_union(hll_hash(column)) and count(column) are supported.

2. Match the correct materialized view when querying.
After the user sends the query, if there is a possibility of matching the materialized view, the query will be rewritten firstly.
    Such as:
    Table: k1 int, k2 int
    MV: k1 int, mv_bitmap_union_k2 bitmap mv_bitmap_union
        mv_bitmap_union = to_bitmap(k2)
    Query: select k1, count(distinct k2) from Table
    Found that there is a match between the materialized view column and the query column, the query is rewritten as:
    Rewritten query: select k1, bitmap_union_count(mv_bitmap_union_k2) from table

Then when the materialized view is matched, it can be matched to the query materialized view table.
Sometimes the rewritten query may not match any materialized view, which means that the rewriting failed. The query needs to be re-parsed and executed again.

### DIFF
--- a/fe/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
+++ b/fe/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
@@ -54,19 +54,19 @@ import java.util.Set;
  */
 public class CreateMaterializedViewStmt extends DdlStmt {
     public static final String MATERIALIZED_VIEW_NAME_PREFIX = "mv_";
-    public static final Map<String, MVColumnPattern> fnNameToPattern;
+    public static final Map<String, MVColumnPattern> FN_NAME_TO_PATTERN;
 
     static {
-        fnNameToPattern = Maps.newHashMap();
-        fnNameToPattern.put(AggregateType.SUM.name().toLowerCase(),
+        FN_NAME_TO_PATTERN = Maps.newHashMap();
+        FN_NAME_TO_PATTERN.put(AggregateType.SUM.name().toLowerCase(),
                 new MVColumnOneChildPattern(AggregateType.SUM.name().toLowerCase()));
-        fnNameToPattern.put(AggregateType.MIN.name().toLowerCase(),
+        FN_NAME_TO_PATTERN.put(AggregateType.MIN.name().toLowerCase(),
                 new MVColumnOneChildPattern(AggregateType.MIN.name().toLowerCase()));
-        fnNameToPattern.put(AggregateType.MAX.name().toLowerCase(),
+        FN_NAME_TO_PATTERN.put(AggregateType.MAX.name().toLowerCase(),
                 new MVColumnOneChildPattern(AggregateType.MAX.name().toLowerCase()));
-        fnNameToPattern.put("count", new MVColumnOneChildPattern("count"));
-        fnNameToPattern.put("bitmap_union", new MVColumnBitmapUnionPattern());
-        fnNameToPattern.put("hll_union", new MVColumnHLLUnionPattern());
+        FN_NAME_TO_PATTERN.put("count", new MVColumnOneChildPattern("count"));
+        FN_NAME_TO_PATTERN.put("bitmap_union", new MVColumnBitmapUnionPattern());
+        FN_NAME_TO_PATTERN.put("hll_union", new MVColumnHLLUnionPattern());
     }
 
     private String mvName;
@@ -187,7 +187,7 @@ public class CreateMaterializedViewStmt extends DdlStmt {
                 // Function must match pattern.
                 FunctionCallExpr functionCallExpr = (FunctionCallExpr) selectListItem.getExpr();
                 String functionName = functionCallExpr.getFnName().getFunction();
-                MVColumnPattern mvColumnPattern = fnNameToPattern.get(functionName.toLowerCase());
+                MVColumnPattern mvColumnPattern = FN_NAME_TO_PATTERN.get(functionName.toLowerCase());
                 if (mvColumnPattern == null) {
                     throw new AnalysisException(
                             "Materialized view does not support this function:" + functionCallExpr.toSqlImpl());
--- a/fe/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
+++ b/fe/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
@@ -18,6 +18,7 @@
 package org.apache.doris.analysis;
 
 import org.apache.doris.catalog.AggregateType;
+import org.apache.doris.catalog.FunctionSet;
 import org.apache.doris.catalog.KeysType;
 import org.apache.doris.catalog.PrimitiveType;
 import org.apache.doris.catalog.Type;
@@ -64,9 +65,9 @@ public class CreateMaterializedViewStmt extends DdlStmt {
                 new MVColumnOneChildPattern(AggregateType.MIN.name().toLowerCase()));
         FN_NAME_TO_PATTERN.put(AggregateType.MAX.name().toLowerCase(),
                 new MVColumnOneChildPattern(AggregateType.MAX.name().toLowerCase()));
-        FN_NAME_TO_PATTERN.put("count", new MVColumnOneChildPattern("count"));
-        FN_NAME_TO_PATTERN.put("bitmap_union", new MVColumnBitmapUnionPattern());
-        FN_NAME_TO_PATTERN.put("hll_union", new MVColumnHLLUnionPattern());
+        FN_NAME_TO_PATTERN.put(FunctionSet.COUNT, new MVColumnOneChildPattern(FunctionSet.COUNT));
+        FN_NAME_TO_PATTERN.put(FunctionSet.BITMAP_UNION, new MVColumnBitmapUnionPattern());
+        FN_NAME_TO_PATTERN.put(FunctionSet.HLL_UNION, new MVColumnHLLUnionPattern());
     }
 
     private String mvName;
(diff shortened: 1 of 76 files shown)


fe/src/main/java/org/apache/doris/analysis/MVColumnHLLUnionPatten.java
```diff
@@ -0,0 +1,55 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package org.apache.doris.analysis;
+
+public class MVColumnHLLUnionPatten  implements MVColumnPatten{
```
HangyuanLiu, line 20:
```suggestion
public class MVColumnHLLUnionPatten  implements MVColumnPattern{
```

fe/src/main/java/org/apache/doris/analysis/QueryStmt.java
```diff
@@ -113,6 +113,40 @@
...
+     * Query: select k3, bitmap_union(to_bitmap(k2)) from TableA
+     * First analyze: MV rewriter enable and this set is empty
+     *     select k3, bitmap_union(mv_bitmap_union_k2) from TableA
+     * SingleNodePlanner: could not select any index for TableA
+     *     Add table to disableMVRewriteTupleIds.
+     * `disableMVRewriteTupleIds` = {TableA}
+     * Re-executed:
+     * Second analyze: MV rewrite disable in table and use origin stmt.
+     *     select k3, bitmap_union(to_bitmap(k2)) from TableA
+     * SingleNodePlanner: base index selected
```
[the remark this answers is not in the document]
EmmyMiao87: Under normal circumstances, the tuple id of the first parsing statement and the second parsing statement should be the same.
If it involves query rewriting, I tested it, and no new tuple will be generated. So in theory, there is no problem.

fe/src/main/java/org/apache/doris/qe/StmtExecutor.java
```diff
@@ -440,58 +439,10 @@ public void analyze(TQueryOptions tQueryOptions) throws AnalysisException, UserE
...
-                } else {
-                    planner.plan(((CreateTableAsSelectStmt) parsedStmt).getInsertStmt(),
-                            analyzer, new TQueryOptions());
-                }
-                // TODO(zc):
-                // Preconditions.checkState(!analyzer.hasUnassignedConjuncts());
+                analyzeAndGenerateQueryPlan(tQueryOptions);
+            } catch (MVSelectFailedException e) {
+                resetAnalyzerAndStmt();
+                analyzeAndGenerateQueryPlan(tQueryOptions);
```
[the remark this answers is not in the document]
EmmyMiao87: Added

fe/src/main/java/org/apache/doris/rewrite/mvrewrite/CountFieldToSum.java
```diff
@@ -0,0 +1,94 @@
...
+        Column mvColumn = olapTable.getVisibleColumn(mvColumnName);
+        if (mvColumn == null) {
+            return expr;
+        }
+
+        // rewrite expr
+        return rewriteExpr(fnChild0, mvColumn, analyzer);
+    }
+
+    public Expr rewriteExpr(SlotRef queryColumnSlotRef, Column mvColumn, Analyzer analyzer) {
```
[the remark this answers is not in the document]
EmmyMiao87: Changed

fe/src/main/java/org/apache/doris/rewrite/mvrewrite/MVExprEquivalent.java
```diff
@@ -0,0 +1,45 @@
...
+// under the License.
+
+package org.apache.doris.rewrite.mvrewrite;
+
+import org.apache.doris.analysis.Expr;
+
+import com.google.common.collect.ImmutableList;
+
+/*
+Only support the once match from originExpr and newExpr
```
morningman, line 25:
```suggestion
 * Only support the once match from originExpr and newExpr
```

fe/src/main/java/org/apache/doris/rewrite/mvrewrite/SlotRefEqualRule.java
```diff
@@ -0,0 +1,39 @@
...
+    public static MVExprEqualRule INSTANCE = new SlotRefEqualRule();
+
+    @Override
+    public boolean equal(Expr queryExpr, Expr mvColumnExpr) {
+        if ((!(queryExpr instanceof SlotRef)) || (!(mvColumnExpr instanceof SlotRef))) {
+            return false;
+        }
+        SlotRef querySlotRef = (SlotRef) queryExpr;
+        SlotRef mvColumnSlotRef = (SlotRef) mvColumnExpr;
+        if (querySlotRef.getColumnName().equalsIgnoreCase(mvColumnSlotRef.getColumnName())) {
```
morningman, line 34: Is it same to check equivalence using column name?
Is it possible to get same column name from different tables?

fe/src/main/java/org/apache/doris/rewrite/mvrewrite/ToBitmapToSlotRefRule.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.doris.common.AnalysisException;
+import org.apache.doris.rewrite.ExprRewriteRule;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Lists;
+
+import java.util.List;
+
+/*
+Rewrite bitmap_union(to_bitmap(c1)) to bitmap_union(mv_bitmap_c1)
```
morningman, line 41:
```suggestion
 * Rewrite bitmap_union(to_bitmap(c1)) to bitmap_union(mv_bitmap_c1)
```

fe/src/main/java/org/apache/doris/rewrite/mvrewrite/SlotRefEqualRule.java
EmmyMiao87: The mvColumnExpr is constructed by MVSelector from the same table

fe/src/main/java/org/apache/doris/analysis/QueryStmt.java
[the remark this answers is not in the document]
morningman: OK

fe/src/main/java/org/apache/doris/rewrite/mvrewrite/SlotRefEqualRule.java
morningman: OK

fe/src/main/java/org/apache/doris/analysis/MVColumnOneChildPattern.java
```diff
@@ -0,0 +1,56 @@
...
+            return false;
+        }
+        Expr functionChild0 = functionCallExpr.getChild(0);
+        if (functionChild0 instanceof SlotRef) {
+            return true;
+        } else if (functionChild0 instanceof CastExpr && (functionChild0.getChild(0) instanceof SlotRef)) {
+            return true;
+        } else {
+            return false;
+        }
```
wutiangan, line 48: 41-47 can rewrite:
SlotRef slotRef = functionCallExpr.getChild(0).unwrapSlotRef();
return slotRef != null;

fe/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
```diff
@@ -50,7 +53,21 @@
...
+        FN_NAME_TO_PATTERN.put(AggregateType.SUM.name().toLowerCase(),
+                new MVColumnOneChildPattern(AggregateType.SUM.name().toLowerCase()));
+        FN_NAME_TO_PATTERN.put(AggregateType.MIN.name().toLowerCase(),
+                new MVColumnOneChildPattern(AggregateType.MIN.name().toLowerCase()));
+        FN_NAME_TO_PATTERN.put(AggregateType.MAX.name().toLowerCase(),
+                new MVColumnOneChildPattern(AggregateType.MAX.name().toLowerCase()));
+        FN_NAME_TO_PATTERN.put("count", new MVColumnOneChildPattern("count"));
+        FN_NAME_TO_PATTERN.put("bitmap_union", new MVColumnBitmapUnionPattern());
+        FN_NAME_TO_PATTERN.put("hll_union", new MVColumnHLLUnionPattern());
+    }
```
wutiangan, line 70: Can you add comment which pattern is suitable for AggregateModel, which pattern is suitable for DuplicatedModel? if it is a need to add a type to distinguish the  case.

fe/src/main/java/org/apache/doris/analysis/MVColumnItem.java
```diff
@@ -79,4 +94,19 @@ public Expr getDefineExpr() {
...
     }
+
+    public Column toMVColumn(OlapTable olapTable) throws DdlException {
+        Column baseColumn = olapTable.getColumn(name);
+        if (baseColumn == null) {
+            Preconditions.checkNotNull(defineExpr != null);
+            Column result = new Column(name, type, isKey, aggregationType, ColumnDef.DefaultValue.ZERO, "");
+            result.setDefineExpr(defineExpr);
+            return result;
+        } else {
```
HangyuanLiu, line 105: baseColumn may be is not null , but we should set define expr in column. such as two different materialized view aggregate function use same base filed .

HangyuanLiu (changes requested):

fe/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
EmmyMiao87: The patterns in here do not distinguish Agg model and duplicate model. These patterns are compatible.

HangyuanLiu (dismissed):
LGTM

morningman (dismissed):
LGTM

morningman (dismissed):
LGTM

morningman (approved):
LGTM

PR #4015, opened 2020-07-03
[Bug] Fix bug that BE crash when doing alter table task

morningman:
Need to check delete condition first.
This bug is introduced by #3943


yangzhg (approved):
LGTM

PR #4020, opened 2020-07-05
[Load][Json] Refactor json load logic to make it more reasonable

morningman:
This CL mainly changes:

1. Reorganized the code logic to limit the supported json format to two, and the import behavior is more consistent.
2. Modified the statistical behavior of the number of error rows when loading in json format, so that the error rows can be counted correctly.
3. See `load-json-format.md` to get details of loading json format.

### DIFF
--- a/be/src/exec/json_scanner.cpp
+++ b/be/src/exec/json_scanner.cpp
@@ -602,7 +602,6 @@ Status JsonReader::_handle_complex_json(Tuple* tuple, const std::vector<SlotDesc
     }
 }
 
-// 
 Status JsonReader::read(Tuple* tuple, const std::vector<SlotDescriptor*>& slot_descs, MemPool* tuple_pool, bool* eof) {
     if (_parsed_jsonpaths.empty()) { // input is a simple json-string
         return _handle_simple_json(tuple, slot_descs, tuple_pool, eof);
--- a/be/src/exec/json_scanner.cpp
+++ b/be/src/exec/json_scanner.cpp
@@ -143,6 +143,7 @@ Status JsonScanner::open_next_reader() {
         strip_outer_array = range.strip_outer_array;
     }
     _cur_file_reader = new JsonReader(_state, _counter, _profile, file, jsonpath, strip_outer_array);
+    RETURN_IF_ERROR(_cur_file_reader->init());
 
     return Status::OK();
 }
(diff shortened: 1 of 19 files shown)


be/src/exec/json_scanner.cpp
```diff
@@ -360,26 +408,31 @@ Status JsonReader::set_tuple_value(rapidjson::Value& objectValue, Tuple* tuple,
...
+Status JsonReader::_handle_simple_json(Tuple* tuple, const std::vector<SlotDescriptor*>& slot_descs, MemPool* tuple_pool, bool* eof) {
     do {
         bool valid = false;
         if (_next_line >= _total_lines) {//parse json and generic document
-            RETURN_IF_ERROR(parse_json_doc(eof));
+            Status st = _parse_json_doc(eof);
+            if (st.is_data_quality_error()) {
+                continue; // continue to read next
+            }
+            RETURN_IF_ERROR(st); // terminate if encounter other errors
```
worker24h, line 435: I think that this code can be deleted

be/src/exec/json_scanner.cpp
```diff
@@ -406,54 +459,60 @@ Status JsonReader::handle_simple_json(Tuple* tuple, const std::vector<SlotDescri
...
+            Status st = _parse_json_doc(eof);
+            if (st.is_data_quality_error()) {
+                continue; // continue to read next
             }
-            _total_lines = get_data_by_jsonpath(slot_descs);
-            if (_total_lines == -1) {
-                return Status::InternalError("Parse json data is failed.");
-            } else if (_total_lines == 0) {
-                *eof = true;
+            RETURN_IF_ERROR(st); // terminate if encounter other errors
```
worker24h, line 502: I think that this code can be deleted

be/src/exec/json_scanner.cpp
```diff
@@ -471,10 +530,14 @@ Status JsonReader::handle_nest_complex_json(Tuple* tuple, const std::vector<Slot
...
+Status JsonReader::_handle_flat_array_complex_json(Tuple* tuple, const std::vector<SlotDescriptor*>& slot_descs, MemPool* tuple_pool, bool* eof) {
     do {
-        if (_next_line >= _total_lines) {//parse json and generic document
-            RETURN_IF_ERROR(parse_json_doc(eof));
+        if (_next_line >= _total_lines) {
+            Status st = _parse_json_doc(eof);
+            if (st.is_data_quality_error()) {
+                continue; // continue to read next
+            }
+            RETURN_IF_ERROR(st); // terminate if encounter other errors
```
worker24h, line 540: I think that this code can be deleted

be/src/exec/json_scanner.cpp
```diff
@@ -524,21 +593,21 @@ Status JsonReader::handle_flat_array_complex_json(Tuple* tuple, const std::vecto
...
     if (_strip_outer_array) {
-        return handle_flat_array_complex_json(tuple, slot_descs, tuple_pool, eof);
+        return _handle_flat_array_complex_json(tuple, slot_descs, tuple_pool, eof);
     } else {
-        return handle_nest_complex_json(tuple, slot_descs, tuple_pool, eof);
+        return _handle_nested_complex_json(tuple, slot_descs, tuple_pool, eof);
     }
 }
 
+//
```
worker24h, line 605: Delete it

be/src/exec/json_scanner.cpp
morningman: file reader may return other error

yangzhg (approved):
LGTM

PR #4021, opened 2020-07-05
[webserver] Make BE webserver handle static files

acelyc111:
Make BE webserver handle static files, e.g. css, js, ico, then we can make
BE website more pretty.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -535,9 +535,6 @@ namespace config {
 
     // Whether to continue to start be when load tablet from header failed.
     CONF_Bool(ignore_load_tablet_failure, "false");
-
-    // Directory where BE's website files placed.
-    CONF_String(www_path, "${DORIS_HOME}/www");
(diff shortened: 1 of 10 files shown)


be/src/common/config.h
```diff
@@ -536,6 +536,8 @@ namespace config {
     // Whether to continue to start be when load tablet from header failed.
     CONF_Bool(ignore_load_tablet_failure, "false");
 
+    // Directory where BE's website files placed.
+    CONF_String(www_path, "${DORIS_HOME}/www");
```
morningman, line 540: Is it necessary to be a config? I think it can be hard coded?

acelyc111: Seems not necessary, I'll remove it.

morningman (approved):
LGTM

acelyc111:
A step for https://github.com/apache/incubator-doris/issues/3911

PR #4034, opened 2020-07-07
[Bug][Broker] Fix bug that the error exception is not caught when closing the broker writer

### LINKED ISSUE #4033: [Restore] Restore failed due to empty file
**Describe the bug**
Restore failed because of following errors shows in `SHOW RESTORE` result.

```
[19685: invalid md5 of downloaded file: /path/to/palobe/data/snapshot/20200630112902.98.1728000/19551/812794862/020000000155a0807d46caeef77bc8fed6308e240e8f9cbc_0_68.dat, expected: 469acc8e8430f1afb1b0abe5bbc81a71, get: d41d8cd98f00b204e9800998ecf8427e
```

**DEBUG**

The value `d41d8cd98f00b204e9800998ecf8427e` is the md5 of empty file. So this error means we got an empty file from repositories.

That is, we upload an  empty file when doing BACKUP.

This is because when doing BACKUP, sometime we may encounter error when closing the file writer. It will results in an empty file left on repository. But the fs broker does not return this error, so BE didn't notice it and continue processing.

morningman:
Broker should return the error status.
Fix #4033

### DIFF
--- a/fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
+++ b/fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
@@ -586,6 +586,8 @@ public class FileSystemManager {
                 fsDataInputStream.close();
             } catch (IOException e) {
                 logger.error("errors while close file input stream", e);
+                throw new BrokerException(TBrokerOperationStatusCode.TARGET_STORAGE_SERVICE_ERROR,
+                        e, "errors while close file input stream");
             } finally {
                 clientContextManager.removeInputStream(fd);
             }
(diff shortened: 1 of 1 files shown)


xy720:
The md5sum was generated before the local file upload to remote, which means the local file is a empty file.
Is it normal we have a empty local file?

fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
```diff
@@ -586,6 +586,8 @@ public void closeReader(TBrokerFD fd) {
                 fsDataInputStream.close();
             } catch (IOException e) {
                 logger.error("errors while close file input stream", e);
+                throw new BrokerException(TBrokerOperationStatusCode.TARGET_STORAGE_SERVICE_ERROR,
+                        e, "errors while close file input stream");
```
xy720, line 590: The md5sum was generated before the local file upload to remote, which means the local file is a empty file.
Is it normal we have a empty local file?

xy720:
LGTM

fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
morningman: Not because an empty file was uploaded, but because the uploaded file was not closed properly, resulting in an empty file.

yangzhg (approved):
LGTM

PR #4035, opened 2020-07-07
[Doris On ES] [UT] Add UT test for all search phase

wuyunfeng:
I forget push some UT test in this PR https://github.com/apache/incubator-doris/pull/4012, and remove `_cluster/state` resource because DOE does not rely the full ES cluster state meta.


morningman (approved):
LGTM

PR #4041, opened 2020-07-07
Fix predicate pushdown may incorrect when groupby with grouping sets

### LINKED ISSUE #4040: Predicate pushdown amy incorrect when groupby with grouping sets
with sql like 
```
SELECT
    IF(col2 IS NULL, 'ALL', col2) AS col2,
    IF(col3 IS NULL, 'ALL', col3) AS col3,
    pv
FROM
(
    SELECT
        col1,
        col2,
        col3,
        SUM(pv) AS pv
    FROM  a
    WHERE col1 = 0
        AND col4 = 1
        AND col3 = 'foo'
    GROUP BY 
    GROUPING SETS (
        (col1),
        (col1, col2),
        (col1, col3),
        (col1, col2, col3)
    )
) t
WHERE IF(col2 IS NULL, 'ALL', col2) = 'ALL'
```

the IF(col2 IS NULL, 'ALL', col2) = 'ALL'  will be push down to scan node of table a, if grouping set is given and column is not in all grouping set list, we cannot push the predicate since the column value can be null

yangzhg:
Fixes #4040 
Fix predicate pushdown amy incorrect when groupby with grouping sets

### DIFF
--- a/fe/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
+++ b/fe/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
@@ -1974,7 +1974,7 @@ public class SingleNodePlanner {
         }
 
         // Push down predicates to Windows' child until they are assigned successfully.
-        final List<Expr> pushDownPredicates = getPredicatesBoundedByGroupbysSourceExpr(predicates, analyzer);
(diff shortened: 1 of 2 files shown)


EmmyMiao87 (dismissed):
LGTM

wutiangan (dismissed):

EmmyMiao87 (approved):

Issue #4046, opened 2020-07-07
[ThreadPool Resource Leak]There are more and more scheduler threads in our production environment


caiconghui:
![image](https://user-images.githubusercontent.com/55968745/86786379-82c81700-c096-11ea-826e-b1a0e0d0ffd6.png)

caiconghui:
I get the thread snapshot of fe process, and find there are many schduler thread. 
grep "ScheduledThreadPoolExecutor.java:899" temp | wc -l
3826

caiconghui:
https://github.com/apache/incubator-doris/blob/master/fe/src/main/java/org/apache/doris/task/MasterTaskExecutor.java#L39-#L43

PR #4049, opened 2020-07-07
[Thread Resource Leak] Fix thread resource leak after checkpoint catalog destroyed

caiconghui:
This PR is mainly to fix thread resource leak, and then add some notice to use newDaemonScheduledThreadPool api in ThreadPoolManager .


caiconghui:
for #4046

fe/src/main/java/org/apache/doris/common/publish/FixedTimePublisher.java
```diff
@@ -28,7 +28,7 @@
 public class FixedTimePublisher {
```
morningman, line 28: Looks like this class is not used now.
Maybe we can mark it as @Deprecated now, or even remove it?

caiconghui: ok, I will remove it.

fe/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -458,10 +459,10 @@ public DynamicPartitionScheduler getDynamicPartitionScheduler() {
     }
 
     private static class SingletonHolder {
-        private static final Catalog INSTANCE = new Catalog();
+        private static final Catalog INSTANCE = new Catalog(false);
     }
 
-    private Catalog() {
+    private Catalog(boolean isCheckpointCatalog) {
```
WingsGo, line 465: maybe we should add some comment to this arg, such as: isCheckpointCatalog=true means we need to do some thread metrics collect.

caiconghui: ok

morningman:
Most of UT are failed. I do not know why... Have you run it in your local env?

caiconghui:
> Most of UT are failed. I do not know why... Have you run it in your local env?

my bad, I will fix it.

morningman (approved):
LGTM

PR #4052, opened 2020-07-08
 [Bug] Fix core of double delete, when RowBatch call transfer_resource_ownership

HappenLee:
Resource release should be done by dest RowBatch.  When we call method transfer_resource_ownership. if we don't clear the corresponding resources, which will cause the core problem of double delete.


wutiangan (approved):

yangzhg (approved):
LGTM

PR #4057, opened 2020-07-09
[runtime] Replace the thread pool in FragmentMgr

vagetablechicken:
Ref https://github.com/apache/incubator-doris/issues/3416
Thread pool submit func may return error. To support error handling, we need add cancel_before_execute() for it.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -324,7 +324,7 @@ namespace config {
     // Fragment thread pool
     CONF_Int32(fragment_pool_thread_num_min, "64");
     CONF_Int32(fragment_pool_thread_num_max, "512");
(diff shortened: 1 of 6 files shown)


chaoyli:
If you replace PriorityThreadPool with util/threadpool.h, I suggest you set fragment_pool_queue_size larger. Because There will be thousands of qps in dashboard reporting. You can double the default value to 2048.

vagetablechicken:
> If you replace PriorityThreadPool with util/threadpool.h, I suggest you set fragment_pool_queue_size larger. Because There will be thousands of qps in dashboard reporting. You can double the default value to 2048.

fixed.

chaoyli (approved):

Issue #1946, opened 2019-10-10
Add tablet meta checkpoint mechanism


yiguolei:
Currently, there are 3 operations during add a new rowset:
1. add rowset meta to meta store.
2. add rowset to tablet meta in memory.
3. save tablet meta to meta store.
4. remove rowset meta from meta store since it is already in tablet meta.
The operation "save tablet meta"  is a very heavy operation and it will occupy a lot of bandwidth in ECON [decouple compute and storage] mode.  
If BE could do tablet meta checkpoint periodly and step 1 and 2 is useful. And there will be only one operation to meta store. It will improve lot of performance in ECON mode.

PR #4061, opened 2020-07-09
[Bug]Support get all rowset meta info in memory from tablet meta url

caiconghui:
This PR is to fix bug that we cannot get the newest tablet meta info from tablet meta url.

### DIFF
--- a/be/src/http/action/meta_action.cpp
+++ b/be/src/http/action/meta_action.cpp
@@ -60,13 +60,11 @@ Status MetaAction::_handle_header(HttpRequest* req, std::string* json_meta) {
         LOG(WARNING) << "no tablet for tablet_id:" << tablet_id << " schema hash:" << schema_hash;
         return Status::InternalError("no tablet exist");
     }
-    OLAPStatus s =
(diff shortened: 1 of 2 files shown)


caiconghui:
for #4061

be/src/http/action/meta_action.cpp
```diff
@@ -60,13 +60,11 @@ Status MetaAction::_handle_header(HttpRequest* req, std::string* json_meta) {
         LOG(WARNING) << "no tablet for tablet_id:" << tablet_id << " schema hash:" << schema_hash;
         return Status::InternalError("no tablet exist");
     }
-    OLAPStatus s =
-            TabletMetaManager::get_json_meta(tablet->data_dir(), tablet_id, schema_hash, json_meta);
```
morningman, line 64: Why are some rowset meta not in rocksdb but only in memory?
Is this part of rowset meta not effective?

caiconghui: this problem is introduced by #1946 , tablet will do meta checkpoint periodly

chaoyli (approved):

morningman (approved):
LGTM

PR #4062, opened 2020-07-09
[webserver] Introduce mustache to simplify BE's website render

acelyc111:
cpp-mustache is a C++ implementation of a Mustache template engine
with support for RapidJSON, and in order to simplify RapidJSON object
building, we introduce class EasyJson from Apache Kudu.


acelyc111:
easy_json.{h,.cc} and easy_json-test.cpp are copied from Kudu repository, mustache.{h, cc} are copied from cpp-mustache repository.
Now BE's web site looks like:
image

morningman (commented):
Could you help to comment out the `logs/` entry? which is not implemented yet.

acelyc111:
> Could you help to comment out the `logs/` entry? which is not implemented yet.

@morningman I've updated to disable to show it on navigate bar.

morningman (approved):
LGTM

PR #4063, opened 2020-07-10
Support using B/K/KB/M/MB/G/GB/T/TB/P/PB as unit in  session variable exec_mem_limit

yangzhg:
#4064 Support using B/K/KB/M/MB/G/GB/T/TB/P/PB as unit in  session variable exec_mem_limit


morningman (approved):
LGTM

PR #4079, opened 2020-07-10
Fix be may core dump when linked schema change

### LINKED ISSUE #4078: [BUG] be may be core dump if the server upgrade  from 0.11
**Describe the bug**
If the server upgrade  from 0.11 and has a duplicate table. After upgrade we add a column by linked schema change, this will be cause a core dump

**To Reproduce**
1. in version  0.11  create a duplicate table, and load some data;
```
CREATE TABLE table1
(
    siteid INT DEFAULT '10',
    citycode SMALLINT,
    username VARCHAR(32) DEFAULT '',
    pv BIGINT DEFAULT '0'
)
duplicate KEY(siteid, citycode, username)
DISTRIBUTED BY HASH(siteid) BUCKETS 10
insert into table1 values
(1,1,"jim",2),
(2,1,"grace",2),
(3,2,"tom",2),
(4,3,"bush",3),
(5,3,"helen",3);
```
2. upgrade to 0.12 
3. add a column by linked schema change, like `ALTER TABLE table1 ADD COLUMN tc_type varchar(24) DEFAULT "all" AFTER pv;`

**Screenshots**
image

yangzhg:
fixes #4078
If the server upgrade from 0.11 and has a duplicate table. After upgrade we add a column by linked schema change, this will be cause a core dump
this is becase of the zone map in duplcate table in 0.11 does not contains the value column, so that the vector may out of boundary

### DIFF
--- a/be/src/olap/rowset/segment_group.cpp
+++ b/be/src/olap/rowset/segment_group.cpp
@@ -256,7 +256,7 @@ OLAPStatus SegmentGroup::add_zone_maps_for_linked_schema_change(
 
     int zonemap_col_num = get_num_zone_map_columns();
     CHECK(zonemap_col_num <= schema_mapping.size()) << zonemap_col_num << " vs. " << schema_mapping.size();
-    // if uprade from 0.11 the zone map not contain the value column, so zone_map_fields.size() is less
+    // if upgrade from 0.11 the zone map not contain the value column, so zone_map_fields.size() is less
     // than zonemap_col_num, this may be cause core dump at access zone_map_fields by index in the following code
     for (size_t i = 0; i < zonemap_col_num && i < zone_map_fields.size(); ++i) {
         const TabletColumn& column = _schema->column(i);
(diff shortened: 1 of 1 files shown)


be/src/olap/rowset/segment_group.cpp
```diff
@@ -256,7 +256,9 @@ OLAPStatus SegmentGroup::add_zone_maps_for_linked_schema_change(
 
     int zonemap_col_num = get_num_zone_map_columns();
     CHECK(zonemap_col_num <= schema_mapping.size()) << zonemap_col_num << " vs. " << schema_mapping.size();
-    for (size_t i = 0; i < zonemap_col_num; ++i) {
+    // if uprade from 0.11 the zone map not contain the value column, so zone_map_fields.size() is less
```
wutiangan, line 259:
```suggestion
    // if upgrade from 0.11 the zone map not contain the value column, so zone_map_fields.size() is less
```

chaoyli:
You can elaborate the commit message explicitly not like fix xxx.
It will be better to read.

morningman (approved):
LGTM

PR #4089, opened 2020-07-13
Fix some problem related with publish version task

caiconghui:
This PR is mainly do following three things:
1. Add thread name in fe log to make trace problem more easy.
2. Add agent_task_resend_wait_time_ms config to escape sending duplicate agent task to be.
3. Skip to continue to update replica version when new version is lower than replica version in fe.

### DIFF
--- a/docs/en/administrator-guide/config/fe_config.md
+++ b/docs/en/administrator-guide/config/fe_config.md
@@ -112,7 +112,13 @@ There are two ways to configure FE configuration items:
 
 ### `agent_task_resend_wait_time_ms`
 
-当代理任务的创建时间被设置的时候，此配置将决定是否重新发送代理任务， 当且仅当当前时间减去创建时间大于agent_task_task_resend_wait_time_ms时，ReportHandler可以重新发送代理任务。    
+This configuration will decide whether to resend agent task when create_time for agent_task is set, only when current_time - create_time > agent_task_resend_wait_time_ms can ReportHandler do resend agent task.     
+
+This configuration is currently mainly used to solve the problem of repeated sending of `PUBLISH_VERSION` agent tasks. The current default value of this configuration is 5000, which is an experimental value.
+ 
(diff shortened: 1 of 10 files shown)


caiconghui:
fix #4090

fe/src/main/java/org/apache/doris/catalog/Replica.java
```diff
@@ -318,12 +318,17 @@ private void updateReplicaInfo(long newVersion, long newVersionHash,
...
-        LOG.debug("before update: {}", this.toString());
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("before update: {}", this.toString());
+        }
 
         if (newVersion < this.version) {
-            // yiguolei: could not find any reason why new version less than this.version should run???
-            LOG.warn("replica {} on backend {}'s new version {} is lower than meta version {}",
-                    id, backendId, newVersion, this.version);
+            // This case means that replica meta version has been updated by ReportHandler before
```
yangzhg, line 326: why the replica meta version been updated by ReportHandler before?
this is abnormal, I think using waring is better

caiconghui: @yangzhg 
for example, now, the publish version daemon send publis verison task(2 3 4 5 6) to one be, and the be finish all publish version task, its replica version is 6 now, but fe need wait other be finish all publish version, now the replica version in fe may be 4, I think it is the common case that be replica version higher than fe replica version, when ReportHandler sync tablet. 
https://github.com/apache/incubator-doris/blob/14ac49dde557b54e9578d91f9dba7e148e5c975b/fe/src/main/java/org/apache/doris/master/ReportHandler.java#L473-L477

morningman: Good explaination. I suggest to add this directly to the comment.

morningman (dismissed):
LGTM

morningman:
Hi @caiconghui , the FE code structure has been changed, please rebase the code to solve the conflicts

chaoyli (approved):

PR #4097, opened 2020-07-14
Fix load orc decimal value wrong

### LINKED ISSUE #4096: [BUG] Result may error when ORC load negative decimal value
Result may error when ORC load negative decimal value 

When load negative decimal which has pre zero , the result is wrong.
eg -0.0014, the orc result is -14(precision ... 0)

HangyuanLiu:
FIX: #4096


morningman (approved):
LGTM

PR #4103, opened 2020-07-15
Do not choose decommissioned BE in colocate balance 

### LINKED ISSUE #4102: Decommission cannot be stopped due to Coloratebalancer's balance
When set `disable_colocate_balance ` to false and set some be to decommission, `Coloratebalancer#balanceGroup` will choose decommissioned BE to locate tablets.

liutang123:
Fix #4102


fe/src/main/java/org/apache/doris/clone/ColocateTableBalancer.java
```diff
@@ -469,8 +469,11 @@ private void balanceGroup() {
             }
 
             List<Long> allBackendIds = infoService.getClusterBackendIds(db.getClusterName(), true);
```
morningman, line 471: This line can be removed

morningman (approved):
LGTM

Issue #4085, opened 2020-07-13
[Proposal]Introduce kudu's ThreadPoolMgr && register metric to manage BE's thread


WingsGo:
Currently, FE's Thread is managed by `ThreadPoolManager`, but BE's thread manage is messy, some are `PriorityThreadPool` some a native pthread, It is inconvient to trace some problem such as thread num increase up to thounds and CPU load is very high.
     Now, we introduce kudu's ThreadPool in #2915, and prepare to replace some native pthread to ThreadPool #3416 , and I think we should also introduce kudu's ThreadPoolMgr to manage ThreadPool, and collect metric such as `user cpu time used`, `system cpu time used`, `context switches`, escepcially for plan fragment executor's thread to help us positioning if the thread is doing a long task which was unexpected or they are just waiting by potential deadlock problems.

WingsGo:
Hi, @chaoyli , I will add these metrics to help users to positioning some slow query's problem in plan fragment threads' problem, do you have any suggestions for the solution, is it OK?

Issue #4071, opened 2020-07-10
Add pull request template to doris project


yangzhg:
Add a  pull request template to doris project

PR #4106, opened 2020-07-16
[Feature][ThreadPool]Add Web Page to display thread's stats

WingsGo:
## Proposed changes
Support display thread's stats in BE's website.Ref ISSUE: #4085 

- add some methods to get thread's stats from Linux's system file in env.
- support get thread's stats by http method.
- register page handle in BE to show thread's stats to help developer position some thread relate problem.

## Types of changes

## Checklist


WingsGo:
The Page looks like

Snip20200716_8
Snip20200716_9

be/src/env/env_util.cpp
```diff
@@ -46,5 +47,59 @@ Status open_file_for_random(Env *env, const string &path, shared_ptr<RandomAcces
     return Status::OK();
 }
 
+static Status do_write_string_to_file(Env* env, const Slice& data,
+                                  const std::string& fname,
```
acelyc111, line 51: indent

be/src/http/default_path_handlers.h
```diff
@@ -20,6 +20,8 @@
 
 #include <stdio.h>
 
+#include "util/thread.h"
```
acelyc111, line 23: You can move this include to cpp file to reduce includes in header files

be/src/util/thread.h
```diff
@@ -18,27 +18,25 @@
...
 
-#include <atomic>
 #include <pthread.h>
 #include <syscall.h>
 
+#include <atomic>
+
 #include "common/status.h"
 #include "gutil/ref_counted.h"
+#include "http/web_page_handler.h"
```
acelyc111, line 28: Forward declare ‘ WebPageHandler’ and move this include to cpp

be/src/http/default_path_handlers.h
WingsGo: ok, done

be/src/env/env_util.cpp
WingsGo: done

PR #4108, opened 2020-07-16
Remove unused session variables

EmmyMiao87:
Remove unused session variables named sql_select_limit.

Change-Id: I35244cd6e490638d08f86a5c08131ff3cdba80b6

### DIFF
--- a/fe/src/main/java/org/apache/doris/qe/SessionVariable.java
+++ b/fe/src/main/java/org/apache/doris/qe/SessionVariable.java
@@ -54,7 +54,6 @@ public class SessionVariable implements Serializable, Writable {
     public static final String COLLATION_DATABASE = "collation_database";
     public static final String COLLATION_SERVER = "collation_server";
     public static final String SQL_AUTO_IS_NULL = "SQL_AUTO_IS_NULL";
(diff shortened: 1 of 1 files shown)


fe/src/main/java/org/apache/doris/qe/SessionVariable.java
```diff
@@ -556,7 +548,8 @@ public void readFields(DataInput in) throws IOException {
             queryCacheType = in.readInt();
             autoIncrementIncrement = in.readInt();
             maxAllowedPacket = in.readInt();
-            sqlSelectLimit = in.readLong();
```
wuyunfeng, line 559: Maybe you should consider BWC.

morningman: What is BWC?

morningman (commented):
This variable is used for compatibility for some low version mysql lib.
Better not to remove it.

fe/src/main/java/org/apache/doris/qe/SessionVariable.java
wuyunfeng: BackWardsCompatible

morningman:
Should not remove.

Issue #4058, opened 2020-07-09
[Spark Load] Add lock when building Global Dict


wangbo:
**Why build global dict needs a lock**
We use ```insert dict_table as select xxx from global dict``` sql to update dict table.
This sql contains a ```read,change,overwrite``` logic.
If multiple load job update dict table concurrently, the data of dict table is wrong.
So we need to guarantee write to dict is a mutually exclusive operation.

**Short-term Solution:LoadJobScheduler submit task one by one**
1 LoadJobScheduler fetch a LoadJob which wants to load data to a doris table contains bitmap from queue, names ```job_A```
2 LoadJobScheduler get ```job_A```'s doris table, names ```doris_tab_A```.
3 LoadJobScheduler traverses ```loadTaskScheduler```'s runningTasks to check whether there is already a running task which loading data to ```doris_tab_A```.If there is ,then skip ```job_A```.

**Some questions about Short-term Solution**
1 If user submit too many load stmt,LoadJobScheduler's queue may be too long. So we need to forbid submiting new load stmt sql when there is too many load job which needs build dict.
2 The advantage of this solution is that it's easy to implement.The disadvantage of this solution is that the granularity of the lock is too coarse.

Currently I prefer to use short-term solution to quickly land ```Spark Load``` in env prod.
But we still has a ultimate solution to discuss.

**Ultimate Solution: We need a distributed lock**
**Plan A: Using Zookeeper**
Actually this is the quickest solution and it's graceful enough to implement in code level.
But this will introduce additional components for doris which will make  whole architecture more complicated.

**Plan B: Implement a distributed lock in FE**
1 FE keys a lock set which contains lockname.
2 Add lock.Client calls http api to write lockname to FE's lock set.
3 Unlock.
3.1 Client calls http api to remove lockname from FE's lock set.
3.2 Client keeps a heartbeat report to FE,if the client heart beat timeout,then FE will release the lock.

**Some questions about implement a distributed lock in FE**
1 When FE leader changes, how new leader get lock set?
   For client which owns lock and find leader changes, then will try to lock again and tell new leader to own lock again.
   For new leader, it won't accept a new add lock request until client timeout to guarantee the client which owns lock can get lock again.

2 How does Client know which FE to connect?
2.1 Use Domain
2.2 Client keeps all FE's host.

3 How to deal the case that a client which owns the lock timeout and reconnected again?
First it shouldn't be owned the lock even reconnected.
Then the global dict may already be wrong, and we need to reload data again.
[...]

PR #4109, opened 2020-07-16
[Spark Load] add job granularity global dict lock

wangbo:
#4058 implement a job granularity lock for global dict
**Design Ideas**
1 Implement a lock
    Class ```LoadJobScheduler``` keeps a ```runningBitmapTableMap``` which contains all running job's related Doris table.
    If a new Spark Load Job is submitted, ```LoadJobScheduler``` will check whether the Job's related Doris tables is already in ```runningBitmapTableMap```, if it is, the new Spark Load Job will be pending.

2 The cases to get lock
        2.1 LoadJobScheduler.submit success
        2.2 SparkLoadJob in ETL then replay

3 The cases to release lock
     3.1 SparkLoadJob in pending status then failed
     3.2 SparkLoadJob in etl then status failed
     3.3 SparkLoadJob in etl then status finish
     3.4 LoadJobScheduler.submit failed
     3.5 SparkLoadJob's state convert to Loading
     3.6 transaction abort

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
@@ -583,6 +583,10 @@ public class Catalog {
         return pullLoadJobMgr;
     }
 
+    public LoadJobScheduler getLoadJobScheduler() {
+        return loadJobScheduler;
+    }
+
     public BrokerMgr getBrokerMgr() {
         return brokerMgr;
     }
--- a/fe/fe-core/src/main/java/org/apache/doris/load/BrokerFileGroup.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/BrokerFileGroup.java
@@ -32,6 +32,7 @@ import org.apache.doris.catalog.KeysType;
 import org.apache.doris.catalog.OlapTable;
 import org.apache.doris.catalog.OlapTable.OlapTableState;
 import org.apache.doris.catalog.Partition;
+import org.apache.doris.catalog.PrimitiveType;
 import org.apache.doris.catalog.Table;
 import org.apache.doris.common.AnalysisException;
 import org.apache.doris.common.DdlException;
@@ -82,6 +83,7 @@ public class BrokerFileGroup implements Writable {
     // load from table
     private long srcTableId = -1;
     private boolean isLoadFromTable = false;
+    private boolean isContainsBitmapColumns = false;
 
     // for unit test and edit log persistence
     private BrokerFileGroup() {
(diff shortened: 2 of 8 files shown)


fe/src/main/java/org/apache/doris/load/loadv2/LoadJobScheduler.java
```diff
@@ -47,6 +51,10 @@
 
     private LinkedBlockingQueue<LoadJob> needScheduleJobs = Queues.newLinkedBlockingQueue();
 
+    // Used to implement a job granularity lock for update global dict table serially
+    // The runningBitmapTables keeps the doris table id of the spark load job which in state pending or etl
+    private Map<Long, Set<Long>> runningBitmapTableMap = new HashMap<>();
```
morningman, line 56: Why calling it a "bitmap" table?

fe/src/main/java/org/apache/doris/load/loadv2/SparkLoadJob.java
```diff
@@ -742,6 +760,26 @@ private void unprotectedLogUpdateStateInfo() {
         Catalog.getCurrentCatalog().getEditLog().logUpdateLoadJob(info);
     }
 
+    // because current SparkLoadJob didn't persist job's tableId, so we need to get job's tableid from transaction mgr
+    private void fillJobTableIdSetWhenReplay() {
+        try {
+            DatabaseTransactionMgr databaseTransactionMgr =
```
morningman, line 766: How about just save the table ids in SparkLoad, so that we don't need to get it from TxnMgr? Because of this logic, the coupling with TxnMgr may be very serious. we have to handle the problem such as "db not exist"...

fe/src/main/java/org/apache/doris/load/loadv2/SparkLoadJob.java
```diff
@@ -909,4 +949,30 @@ private void initTDescriptorTable(DescriptorTable descTable) {
             tDescriptorTable = descTable.toThrift();
         }
     }
+
+    public Set<Long> getTableWithBitmapColumn() {
+        if (tableWithBitmapColumn.size() != 0) {
```
morningman, line 954: I think we can prepare this bitmap table map when analyzing the job, and then persist it.

fe/src/main/java/org/apache/doris/load/loadv2/SparkLoadJob.java
wangbo: 👌

fe/src/main/java/org/apache/doris/load/loadv2/LoadJobScheduler.java
wangbo: Maybe "runningTableWithBitmapColumns" is better ?

fe/src/main/java/org/apache/doris/load/loadv2/SparkLoadJob.java
wangbo: 👌

PR #4112, opened 2020-07-17
[Bug] Fix bug that tablet meta lock twice

### LINKED ISSUE #4111: [Bug] BE crash when doing clone
**Describe the bug**
The meta lock of a tablet is double locked when doing clone task.

This bug is introduced by #4061

morningman:
## Proposed changes

The tablet meta lock may already be hold before calling
`generate_tablet_meta_copy()`, so we need provide a unlocked
version of `generate_tablet_meta_copy()`.
This bug  is introduced by #4061 

## Types of changes

## Checklist

- [x] I have create an issue on Fix #4111

### DIFF
--- a/be/src/olap/tablet.h
+++ b/be/src/olap/tablet.h
@@ -222,6 +222,8 @@ public:
     void build_tablet_report_info(TTabletInfo* tablet_info);
 
     void generate_tablet_meta_copy(TabletMetaSharedPtr new_tablet_meta) const;
+    // caller should hold the _meta_load before calling this method
+    void generate_tablet_meta_copy_unlocked(TabletMetaSharedPtr new_tablet_meta) const;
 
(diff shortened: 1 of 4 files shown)


be/src/olap/tablet.h
```diff
@@ -222,6 +222,8 @@ class Tablet : public BaseTablet {
     void build_tablet_report_info(TTabletInfo* tablet_info);
 
     void generate_tablet_meta_copy(TabletMetaSharedPtr new_tablet_meta) const;
+    // caller should hold the _meta_load before calling this method
```
caiconghui, line 225:
```suggestion
    // caller should hold the _meta_lock before calling this method
```

morningman: done

EmmyMiao87 (approved):
LGTM

PR #4118, opened 2020-07-20
[Bug] fix use-after-poison bug in ut schema_change_test 

vagetablechicken:
Ref issue https://github.com/apache/incubator-doris/issues/4116
Using slice->data to create HyperLogLog, it will exec `HyperLogLog(Slice(const char*))`. Then `Slice(const char*)` will use `strlen(data)` to calc the size. But the slice in this ut isn't a C-string. Need to use `Slice`.

### DIFF
--- a/be/test/olap/schema_change_test.cpp
+++ b/be/test/olap/schema_change_test.cpp
@@ -873,9 +873,8 @@ TEST_F(TestColumn, ConvertCharToHLL) {
     RowCursor mv_row_cursor;
     mv_row_cursor.init(mv_tablet_schema);
     mutable_block.get_row(0, &mv_row_cursor);
-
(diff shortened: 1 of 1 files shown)


chaoyli:
You can simplify the commit message to easy understand.

chaoyli (approved):

vagetablechicken:
> You can simplify the commit message to easy understand.

How about this commit message?

chaoyli:
> > You can simplify the commit message to easy understand.
> 
> How about this commit message?

You should make another to know what the commit's addressing problem and the solution.

vagetablechicken:
> > > You can simplify the commit message to easy understand.
> > 
> > 
> > How about this commit message?
> 
> You should make another to know what the commit's addressing problem and the solution.

done

chaoyli (approved):

PR #4125, opened 2020-07-21
[Bug][Load][Json] #4124 Load json format with stream load failed

### LINKED ISSUE #4124: [Bug][Load][Json] Load json format with stream load failed
**Describe the bug**
if the json data is larger than 4096 bytes, the data will be truncated, making it impossible to obtain the json data completely

**To Reproduce**
Load a large json file with stream load

morningman:
## Proposed changes

Stream load should read all the data completely before parsing the json.
And also add a new BE config `streaming_load_max_batch_read_mb`
to limit the data size when loading json data.

Fix: #4124

## Types of changes

## Checklist

- [x] I have create an issue on #4124, and have described the bug/feature there in detail

### DIFF
--- a/be/src/exec/json_scanner.cpp
+++ b/be/src/exec/json_scanner.cpp
@@ -427,25 +427,34 @@ void JsonReader::_set_tuple_value(rapidjson::Value& objectValue, Tuple* tuple, c
 Status JsonReader::_handle_simple_json(Tuple* tuple, const std::vector<SlotDescriptor*>& slot_descs, MemPool* tuple_pool, bool* eof) {
     do {
         bool valid = false;
-        if (_next_line >= _total_lines) {//parse json and generic document
+        if (_next_line >= _total_lines) { // parse json and generic document
             Status st = _parse_json_doc(eof);
             if (st.is_data_quality_error()) {
                 continue; // continue to read next
             }
             RETURN_IF_ERROR(st); // terminate if encounter other errors
-            if (*eof) {// read all data, then return
+            if (*eof) { // read all data, then return
                 return Status::OK();
             }
             if (_json_doc.IsArray() ) {
                 _total_lines = _json_doc.Size();
+                if (_total_lines == 0) {
+                    // may be passing an empty json, such as "[]"
+                    std::stringstream str_error;
+                    str_error << "Empty json line";
+                    _state->append_error_msg_to_file(_print_json_value(_json_doc), str_error.str());
+                    _counter->num_rows_filtered++;
+                    continue;
+                }
             } else {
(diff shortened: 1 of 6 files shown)


wutiangan (dismissed):

be/src/http/action/stream_load.cpp
```diff
@@ -312,7 +312,7 @@ Status StreamLoadAction::_process_put(HttpRequest* http_req, StreamLoadContext*
     request.formatType = ctx->format;
     request.__set_loadId(ctx->id.to_thrift());
     if (ctx->use_streaming) {
-        auto pipe = std::make_shared<StreamLoadPipe>();
+        auto pipe = std::make_shared<StreamLoadPipe>(1024 * 1024, 64* 1024, ctx->body_bytes);
```
wutiangan, line 315:
```suggestion
        auto pipe = std::make_shared<StreamLoadPipe>(1024 * 1024, 64 * 1024, ctx->body_bytes);
```

be/src/exec/json_scanner.cpp
```diff
@@ -243,6 +243,7 @@ Status JsonReader::_parse_json_doc(bool* eof) {
     uint8_t* json_str = nullptr;
     size_t length = 0;
     RETURN_IF_ERROR(_file_reader->read_one_message(&json_str, &length));
+    LOG(INFO) << "cmy get json length: " << length;
```
wutiangan, line 246:
```suggestion
    LOG(INFO) << "get json length: " << length;
```

be/src/runtime/stream_load/stream_load_pipe.h
```diff
@@ -85,30 +87,14 @@ class StreamLoadPipe : public MessageBodySink, public FileReader {
...
-        auto buf = _buf_queue.front();
-        *length = buf->remaining();
-        *data = new uint8_t[*length];
-        buf->get_bytes((char*)(*data) , *length);
-
-        _buf_queue.pop_front();
-        _buffered_bytes -= buf->limit;
-        _put_cond.notify_one();
-        return Status::OK();
+        *data = new uint8_t[_total_length];
```
wutiangan, line 93: if _total_lenght == 0?

be/src/exec/json_scanner.cpp
worker24h: Please remove it，otherwise Log file will burst

morningman: Removed

be/src/runtime/stream_load/stream_load_pipe.h
morningman: I will add a check here

PR #4128, opened 2020-07-21
Add compaction point changing log

ZhangYu0123:
Add log to record the compaction point changing. When OLAP_ERR_BE_SEGMENTS_OVERLAPPING happens, it will be used to  track the bugs. Add issue #4134 link.


yangzhg (dismissed):
You'd better to create an issue to describe why add those log

yangzhg (approved):
LGTM

PR #4140, opened 2020-07-22
Add materialized view related syntax

### LINKED ISSUE #3344: [Proposal] Materialized View 2.0

HangyuanLiu:
## Proposed changes

1、Add restrict when create materialized which partition by key mush be in group by 
2、Add show alter table materialized view syntax
3、Add cancel alter materialized view syntax

For #3344 

## Types of changes

## Checklist

- [ ] I have create an issue on [Doris's issues](https://github.com/apache/incubator-doris/issues), and have described the bug/feature there in detail

## Further comments


fe/fe-core/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
```diff
@@ -224,6 +232,9 @@ public void analyzeSelectClause() throws AnalysisException {
         if (beginIndexOfAggregation == 0) {
             throw new AnalysisException("The materialized view must contain at least one key column");
         }
+        if (partitionColumns.size() > 0) {
```
morningman, line 235: There may be no group by clause at all?

Issue #4142, opened 2020-07-22
[BUG] Alter table add key column fails in agg module using LinkedSchemaChange policy


ZhangYu0123:
**Describe the bug**
When alter table add key column in agg model using LinkedSchemaChange policy, "ALTER, status_code: RUNTIME_ERROR"  will be returned. 
The error log as follows:

E0721 13:48:18.405797 34117 alpha_rowset.cpp:308] column pruning size is error.KeysType=AGG_KEYS, zone_maps_size=4, num_key_columns=5, num_columns=5
W0721 13:48:18.405853 34117 alpha_rowset_writer.cpp:227] rowset init failed when build new rowset, res=-402

**To Reproduce**
1. Create table
CREATE TABLE `table13` (
  `siteid` varchar(64) NOT NULL COMMENT "",
  `siteid2` varchar(64) NOT NULL COMMENT "",
  `siteid3` varchar(64) NOT NULL COMMENT ""
) ENGINE=OLAP
AGGREGATE KEY(`siteid`, `siteid2`, `siteid3`)
COMMENT "OLAP"
DISTRIBUTED BY HASH(`siteid`) BUCKETS 1
PROPERTIES (
"replication_num" = "1",
"in_memory" = "false",
"storage_format" = "DEFAULT"
);

2. insert data
3. alter table add key column
alter table `table13` add column b int key default '0' after `siteid2`

SHOW ALTER TABLE COLUMN;

**Expected behavior**
success

**Screenshots**
![image](https://user-images.githubusercontent.com/67053339/88165526-69070200-cc48-11ea-83b6-619d50a2333e.png)

EmmyMiao87:
Same issue: #4617

PR #4143, opened 2020-07-22
Fix alter schema add key column bug in agg model

ZhangYu0123:
## Proposed changes
Fix the bug "alter schema add key column bug in agg model when using LinkedSchemaChange policy",  the detail description #4142. 
The change is when add key column  in agg model , we set sc_directly = true to use SchemaChangeDirectly policy.

## Types of changes

## Checklist

### DIFF
--- a/be/src/olap/rowset/segment_group.cpp
+++ b/be/src/olap/rowset/segment_group.cpp
@@ -257,6 +257,13 @@ OLAPStatus SegmentGroup::add_zone_maps_for_linked_schema_change(
     CHECK(zonemap_col_num <= schema_mapping.size()) << zonemap_col_num << " vs. " << schema_mapping.size();
 
     for (size_t i = 0; i < zonemap_col_num; ++i) {
+        
+        // in duplicated table update from 0.11 to 0.12, zone map index may be missed
+        if (_schema->keys_type() == DUP_KEYS && 
+            schema_mapping[i].ref_column != -1 &&
+            schema_mapping[i].ref_column >= zone_map_fields.size()) {
+            continue;
+        }
         const TabletColumn& column = _schema->column(i);
 
         WrapperField* first = WrapperField::create(column);
@@ -265,8 +272,8 @@ OLAPStatus SegmentGroup::add_zone_maps_for_linked_schema_change(
         WrapperField* second = WrapperField::create(column);
         DCHECK(second != NULL) << "failed to allocate memory for field: " << i;
 
-        // add column also can add zone map index
-        if (schema_mapping[i].ref_column == -1 || i >= zone_map_fields.size()) {
+        // when this is no ref_column (add new column), fill default value
+        if (schema_mapping[i].ref_column == -1 || schema_mapping[i].ref_column >= zone_map_fields.size()) {
             // ref_column == -1 means this is a new column.
             // for new column, use default value to fill into column_statistics
             if (schema_mapping[i].default_value != nullptr) {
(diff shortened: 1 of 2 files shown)


be/src/olap/schema_change.cpp
```diff
@@ -2111,6 +2111,11 @@ OLAPStatus SchemaChangeHandler::_parse_request(TabletSharedPtr base_tablet,
     for (size_t i = 0; i < new_tablet->num_columns(); ++i) {
         ColumnMapping* column_mapping = rb_changer->get_mutable_column_mapping(i);
         if (column_mapping->ref_column < 0) {
+            // check agg model, add key-column operation use sc_directly to build index
+            if (new_tablet_schema.keys_type() == AGG_KEYS && new_tablet_schema.column(i).is_key()) {
+                *sc_directly = true;
+                return OLAP_SUCCESS;
+            }
```
yangzhg, line 2118: this will disable all linked schema in when add key column in agg table, try some better way

chaoyli:
Now, Add key column in prefix index will do directly schema change.
```
if (base_tablet->num_short_key_columns() != new_tablet->num_short_key_columns()) {
  // the number of short_keys changed, can't do linked schema change
  *sc_directly = true;
  return OLAP_SUCCESS;
}
```
Another add key column will not change the order of record, it will be ok to do linked schema change.
So you should check why linked schema change will failed on zone_map.
You can reference the following function which is used to handle the zone map upon linked schema change.
```
add_rowset_for_linked_schema_change
```

be/src/olap/rowset/segment_group.cpp
```diff
@@ -268,11 +265,14 @@ OLAPStatus SegmentGroup::add_zone_maps_for_linked_schema_change(
...
 
-        if (schema_mapping[i].ref_column == -1) {
+        // add column also can add zone map index
+        if (schema_mapping[i].ref_column == -1 || i >= zone_map_fields.size()) {
             // ref_column == -1 means this is a new column.
             // for new column, use default value to fill into column_statistics
-            first->copy(schema_mapping[i].default_value);
-            second->copy(schema_mapping[i].default_value);
+            if (schema_mapping[i].default_value != nullptr) {
+                first->copy(schema_mapping[i].default_value);
```
yangzhg, line 273: if the zone map set to default value， it may be filter data incorrectly

be/src/olap/rowset/segment_group.cpp
```diff
@@ -268,11 +265,14 @@ OLAPStatus SegmentGroup::add_zone_maps_for_linked_schema_change(
         WrapperField* second = WrapperField::create(column);
         DCHECK(second != NULL) << "failed to allocate memory for field: " << i;
 
-        if (schema_mapping[i].ref_column == -1) {
+        // add column also can add zone map index
```
yangzhg, line 268: better to make this comment more readable

be/src/olap/rowset/segment_group.cpp
```diff
@@ -268,11 +265,14 @@ OLAPStatus SegmentGroup::add_zone_maps_for_linked_schema_change(
...
+        if (schema_mapping[i].ref_column == -1 || i >= zone_map_fields.size()) {
             // ref_column == -1 means this is a new column.
             // for new column, use default value to fill into column_statistics
-            first->copy(schema_mapping[i].default_value);
-            second->copy(schema_mapping[i].default_value);
+            if (schema_mapping[i].default_value != nullptr) {
```
yangzhg, line 272: if you set it to default value， in duplicated table update from 0.11 to 0.12 may cause value clumn filtered unexpected

yangzhg (commented):
LGTM

yangzhg (approved):

PR #4148, opened 2020-07-22
[Bug][TabletRepair] Fix bug that too many replicas generated when decommission BE

### LINKED ISSUE #4147: [Bug] Lots of version incomplete replica created when decommission BE
**Describe the bug**
When decommission BE, some of tablets may has lots of replicas generated by clone task.

**Why**

Give examples of why:

1. Tablet X has 3 replicas on A, B, C 3 BEs.
2. C is decommission, so we choose the Backend D to relocating the new replica,
3. After relocating, Tablet X has 4 replicas: A, B, C(decommision), D(may be version incomplete)
4. D may be version incomplete because the clone task ran a long time, the new version has been published.
5. At the next time of tablet checking, Tablet X's status is still REPLICA_RELOCATING.
    If we don't choose D as dest BE to do the new relocating, it will choose new backend E
    to store the new replicas. So back and forth, the number of replicas will increase forever.

So a better solution is to select D as dest BE again to do the clone task. This may trigger an increment clone task
that can be done more fast.

morningman:
## Proposed changes

Try to select the BE with an existing replicas as the destination BE for
REPLICA_RELOCATING clone task.
Fix #4147 

Also add 2 new FE configs `max_clone_task_timeout_sec` and `min_clone_task_timeout_sec`

## Types of changes

## Checklist

- [x] I have create an issue on [Doris's issues](#4147), and have described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/master/ReportHandler.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/master/ReportHandler.java
@@ -481,7 +481,6 @@ public class ReportHandler extends Daemon {
                             // 1. PUSH finished in BE but failed or not yet report to FE
                             // 2. repair for VERSION_INCOMPLETE finished in BE, but failed or not yet report to FE
                             replica.updateVersionInfo(backendVersion, backendVersionHash, dataSize, rowCount);
-                            replica.setBad(false);
 
                             if (replica.getLastFailedVersion() < 0 && !isInitVersion) {
                                 // last failed version < 0 means this replica becomes health after sync,
--- a/be/src/olap/task/engine_clone_task.cpp
+++ b/be/src/olap/task/engine_clone_task.cpp
@@ -330,7 +330,7 @@ AgentStatus EngineCloneTask::_clone_copy(
         std::string remote_url_prefix;
         {
             // TODO(zc): if snapshot path has been returned from source, it is some strange to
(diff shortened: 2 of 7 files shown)


kangkaisen (dismissed):
+1， LGTM

morningman:
Rebase to solve the conflict.

kangkaisen (dismissed):
+1

morningman:
Sorry that this solution still has problem, I will fix it in this PR after testing in our product env.

fe/fe-core/src/main/java/org/apache/doris/master/ReportHandler.java
```diff
@@ -481,7 +481,6 @@ private static void sync(Map<Long, TTablet> backendTablets, ListMultimap<Long, L
                             // 1. PUSH finished in BE but failed or not yet report to FE
                             // 2. repair for VERSION_INCOMPLETE finished in BE, but failed or not yet report to FE
                             replica.updateVersionInfo(backendVersion, backendVersionHash, dataSize, rowCount);
-                            replica.setBad(false);
```
morningman, line 484: For reviewer:
In the original design logic, when the status of a replica is bad, it cannot become not bad again.
This is a clerical error and will cause the `admin set replica status` function to fail.

kangkaisen (approved):
+1

PR #4150, opened 2020-07-23
[BUG] fix 4149, add sessionVariable to choose broadcastjoin first when cardinality cannot be estimated

### LINKED ISSUE #4149: [BUG] choose broadcastjoin first when cardinality cannot be estimated
sql：
`select * from (select k1 from jointest group by k1)t2, jointest t1 where t1.k1 = t2.k1`

reason:
In version 0.11, the SQL choose broadcastjoin and returns it in 9s. However, in version 0.12, the SQL choose shuffle join and runs 32S.
When the cardinality cannot be estimated, it is better to choose broadcast join first. When there is a performance bottleneck, users can manually add hint to shuffle point.

wutiangan:
fix #4149

add sessionVariable "prefer_broadcast_join" to choose broadcastjoin first when cardinality cannot be estimated
the sessionVariable can set local or global. the default value is true, prefer broadcast join first where the two cost is equal.

### DIFF
--- a/docs/zh-CN/administrator-guide/variables.md
+++ b/docs/zh-CN/administrator-guide/variables.md
@@ -338,6 +338,6 @@ SET forward_to_master = concat('tr', 'u', 'e');
 
     是否将 bitmap 和 hll 类型的 count distinct 查询重写为 bitmap_union_count 和 hll_union_agg 。
 
-* `prefer_broadcast_join`
+* `prefer_join_method`
 
     在选择join的具体实现方式是broadcast join还是shuffle join时，如果broadcast join cost和shuffle join cost相等时，优先选择哪种join方式。
--- a/docs/zh-CN/administrator-guide/variables.md
+++ b/docs/zh-CN/administrator-guide/variables.md
@@ -338,3 +338,6 @@ SET forward_to_master = concat('tr', 'u', 'e');
 
     是否将 bitmap 和 hll 类型的 count distinct 查询重写为 bitmap_union_count 和 hll_union_agg 。
 
+* `prefer_broadcast_join`
+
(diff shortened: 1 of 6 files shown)


kangkaisen (dismissed):
+1

docs/zh-CN/administrator-guide/variables.md
```diff
@@ -338,3 +338,6 @@ SET forward_to_master = concat('tr', 'u', 'e');
 
     是否将 bitmap 和 hll 类型的 count distinct 查询重写为 bitmap_union_count 和 hll_union_agg 。
 
+* `prefer_broadcast_join`
```
morningman, line 341: How about name it to "prefer_join_method", and value is "broadcast" or "shuffle"

morningman (dismissed):
LGTM

fe/fe-core/src/main/java/org/apache/doris/planner/DistributedPlanner.java
```diff
@@ -286,7 +286,8 @@ private PlanFragment createScanFragment(PlanNode node) {
      * Therefore, we add a SessionVariable to help users choose a better join implementation.
      */
     private boolean isBroadcastCostSmaller(long broadcastCost, long partitionCost)  {
-        if (ConnectContext.get().getSessionVariable().isPreferBroadcastJoin()) {
+        String joinMethod = ConnectContext.get().getSessionVariable().getPreferJoinMethod();
+        if (joinMethod.equals("broadcast")) {
```
kangkaisen, line 290: `equalsIgnoreCase` should more better.

kangkaisen (approved):
+1

Issue #3926, opened 2020-06-22
[Proposal] Support spilling to disk for Doris


HappenLee:
## Motivation
At present, the use of Doris often encounters the limitation bottleneck of ```mem limit```, which leads to many queries can not be completed.

Although we can solve this problem by adjusting the ```mem_limit``` of query. But in some memory bottleneck scenarios, this is futile.

The capacity of the disk is usually about 100 times of the memory, if we can spill the data beyond the memory limit to the disk. This almost solves the above problem perfectly, but the speed of disk is much slower than that of memory, which will also lead to long execution time of query.

### It can bring us the following benefits:

1. In some memory tight scenarios, more memory is available at the expense of query execution time. This is necessary in some scenarios

2. Doris can dispose larger query without memory constraints

## Implementation

1. Now, The ```BufferedBlockMgr2``` and ```DiskIOMgr``` have already supported to spill mem data to disk. We need to use these functions to writes data to a temporary work area on disk. The default location of this work area is ```doris-scratch```,  when an operation completes, the data is removed from the disk.

2. There are 3 version of ```BufferedTupleStream``` which make us confuse. We need to unify the abstraction of this important part to do a good job for spilling to disk.

3. Successively implement the disk dropping function of the following execution nodes：
        
    * Sort
    * Aggregation
    * Analytic function
    * Join 

4. Remove redundant code, such as ```BufferTupleStream```, ```HashTable``` and so on.

5.  Some optimization of spilling to disk:
     * Size limit of temporary file
     * Limit of IO speed of spilling to disk
     * Using the IO capability of SSD
     * Compression and decompression of spilling data

PR #4165, opened 2020-07-24
[Spill To Disk][3/6] Add some function and variables in base class:BlockingJoinNode. Prepare for Derived class:PartitionHashJoinNode

HappenLee:
issue:#3926

## Proposed changes
BlockingJoinNode is the base class of Blocking Join。Now only have derived class：CrossJoinNode
We add some variables and function in BlockingJoinNode，prepare for implementation of Derived class:PartitionHashJoinNode.
The modification of the current base class does not affect the implementation logic of the now derived class.

## Types of changes

## Checklist

### DIFF
--- a/be/src/exec/blocking_join_node.cpp
+++ b/be/src/exec/blocking_join_node.cpp
@@ -36,7 +36,7 @@ BlockingJoinNode::BlockingJoinNode(const std::string& node_name,
       _node_name(node_name),
       _join_op(join_op),
       _current_left_child_row(nullptr),
-      _semi_join_staging_row(nullptr){
+      _semi_join_staging_row(nullptr) {
 }
 
 Status BlockingJoinNode::init(const TPlanNode& tnode, RuntimeState* state) {
@@ -88,8 +88,7 @@ Status BlockingJoinNode::prepare(RuntimeState* state) {
 }
 
 Status BlockingJoinNode::close(RuntimeState* state) {
-    // TODO(zhaochun): avoid double close
-    // if (is_closed()) return Status::OK();
+    if (is_closed()) return Status::OK();
     _left_batch.reset();
     _build_batch.reset();
     if (_semi_join_staging_row != nullptr) delete[] _semi_join_staging_row;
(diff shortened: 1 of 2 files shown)


be/src/exec/blocking_join_node.cpp
```diff
@@ -34,7 +34,9 @@ BlockingJoinNode::BlockingJoinNode(const std::string& node_name,
                                    const DescriptorTbl& descs)
     : ExecNode(pool, tnode, descs),
       _node_name(node_name),
-      _join_op(join_op) {
+      _join_op(join_op),
+      _current_left_child_row(nullptr),
+      _semi_join_staging_row(nullptr){
```
wutiangan, line 39:
```suggestion
      _semi_join_staging_row(nullptr) {
```

be/src/exec/blocking_join_node.cpp
```diff
@@ -74,14 +76,23 @@ Status BlockingJoinNode::prepare(RuntimeState* state) {
...
+    _build_batch.reset(new RowBatch(child(1)->row_desc(), state->batch_size(), mem_tracker()));
     return Status::OK();
 }
 
 Status BlockingJoinNode::close(RuntimeState* state) {
     // TODO(zhaochun): avoid double close
     // if (is_closed()) return Status::OK();
     _left_batch.reset();
+    _build_batch.reset();
+    if (_semi_join_staging_row != nullptr) delete[] _semi_join_staging_row;
```
wutiangan, line 95: should set _semi_join_staging_row  to zero to avoid double close?

be/src/exec/blocking_join_node.h
```diff
@@ -84,11 +88,20 @@ class BlockingJoinNode : public ExecNode {
     // This should be the same size as the left child tuple row.
     int _result_tuple_row_size;
 
+    /// Row assembled from all lhs and rhs tuples used for evaluating the non-equi-join
+    /// conjuncts for semi joins. Semi joins only return the lhs or rhs output tuples,
+    /// so this tuple is temporarily assembled for evaluating the conjuncts.
+    TupleRow* _semi_join_staging_row;
```
wutiangan, line 94: now we can't support no-equi-join,  will you add no-euqal-join semi join latter?

HappenLee: No，This comment is not very accurate. I will revise this part.

be/src/exec/blocking_join_node.cpp
HappenLee: ok，I will check this part。

PR #4172, opened 2020-07-24
Simplify logic and improve efficiency

Pslydhh:


morningman:
Could you please create as ISSUE to describe the problem you fixed?

be/src/util/blocking_priority_queue.hpp
```diff
@@ -85,34 +85,32 @@ class BlockingPriorityQueue {
...
-                    while (!_queue.empty()) {
-                        T v = _queue.top();
-                        _queue.pop();
-                        ++v;
-                        tmp_queue.push(v);
-                    }
-                    swap(_queue, tmp_queue);
-                    _upgrade_counter = 0;
+        if (!_queue.empty()) {
+            // 定期提高队列中残留的任务优先级
```
wuyunfeng, line 89: Maybe you should write the comment using the English

Issue #4130, opened 2020-07-21
[BUG] from_base64 


yuanfeng0905:
example sql:

```sql
select a,from_base64(a) 
from (
    select "aGVsbG8=" a 
    union all
    select null a
) t
```
The expect result is:

| a | from_base64(`a`) |
| - | - |
|  aGVsbG8= | hello |
|    |    |

But now the result is:

| a | from_base64(`a`) |
| - | - |
|  aGVsbG8= | hello |
|  | hello |

**Doris version: 0.11.25 (baidu-doris/incubator-doris)**

yangzhg:
It is a bug,and I will fix it, to_base64 function also has this problem, and to_base64 also has an bug may cause memory problem

PR #4183, opened 2020-07-24
Fix from/to_base64 may return incorrect value

yangzhg:
## Proposed changes

from/to_base64 may return incorrect value when the value is null #4130 
remove the duplicated base64 code
fix the base64 encoded string length is wrong， and this will cause the memory error；

## Types of changes

## Checklist

- [x] Commit messages in my PR start with the related issues ID


kangkaisen (approved):
+1

Issue #3950, opened 2020-06-25
Please support timezone as more as possible


lxxstc:
The time zone of my server is UTC. so, It will throw an exception like "Unknown or incorrect time zone: 'UTC'", when I perform some DDL. Please support the timezone as more as possible, at least including UTC.
Below is git diff between my workspace and branch master

```
index 21612bfc..6dcfe0ce 100644
--- a/fe/src/main/java/org/apache/doris/common/util/TimeUtils.java
+++ b/fe/src/main/java/org/apache/doris/common/util/TimeUtils.java
@@ -256,7 +256,7 @@ public class TimeUtils {
             Matcher matcher = TIMEZONE_OFFSET_FORMAT_REG.matcher(value);
             // it supports offset and region timezone type, "CST" use here is compatibility purposes.
             boolean match = matcher.matches();
-            if (!value.contains("/") && !value.equals("CST") && !match) {
+            if (!value.contains("/") && !value.equals("UTC") && !value.equals("CST") && !match) {
                 ErrorReport.reportDdlException(ErrorCode.ERR_UNKNOWN_TIME_ZONE, value);
             }
             if (match) {
```

lxxstc:
I have checked the output of  `timedatectl list-timezones`, only `UTC` has been lost.

lxxstc:
another way to fix this issue with the command below
```
# timedatectl set-timezone Atlantic/Reykjavik
```
replace the time zone UTC with time zone Atlantic/Reykjavik

morningman:
Could you submit a PR to add UTC for us?

PR #4187, opened 2020-07-26
[TimeZone]Get default timezone from session variable && add UTC timezone alias to UTC +00:00

WingsGo:
## Proposed changes

TimeUtils cannot identify timezone "UTC", but in some OS system, we will get default timezone value `UTC`.It will cause some DDL failed with msg "Unknown or incorrect time zone: UTC", and some UT will also failed cause by this.So This PR is to get default timezone from session vaiable && set UTC mapping to default_timezone `UTC+00:00` to avoid situation like this. Relate ISSUE: #3950 

## Types of changes

## Checklist


fe/fe-core/src/main/java/org/apache/doris/common/util/TimeUtils.java
```diff
@@ -55,7 +55,7 @@
 
     // set CST to +08:00 instead of America/Chicago
     public static final ImmutableMap<String, String> timeZoneAliasMap = ImmutableMap.of(
-            "CST", DEFAULT_TIME_ZONE, "PRC", DEFAULT_TIME_ZONE);
+            "CST", DEFAULT_TIME_ZONE, "PRC", DEFAULT_TIME_ZONE, "UTC", DEFAULT_TIME_ZONE);
```
morningman, line 58: I think `DEFAULT_TIME_ZONE` means `UTC+8` ? Not equals to `UTC`?

fe/fe-core/src/main/java/org/apache/doris/common/util/TimeUtils.java
```diff
@@ -256,7 +256,7 @@ public static String checkTimeZoneValidAndStandardize(String value) throws DdlEx
             Matcher matcher = TIMEZONE_OFFSET_FORMAT_REG.matcher(value);
             // it supports offset and region timezone type, "CST" use here is compatibility purposes.
             boolean match = matcher.matches();
-            if (!value.contains("/") && !value.equals("CST") && !match) {
+            if (!value.contains("/") && !value.equals("CST") && !value.equals("UTC") && !match) {
```
morningman, line 259: With more and more time zone added, I think it is better to refactor this judgment to make it easy to maintain.

WingsGo: The root case of the problem such as `PRC` Timezone is that TimeUtils check is only check by regex we given, maybe we should also check the timezone exists in `ZoneId.getAvailableZoneIds`?

morningman: > The root case of the problem such as `PRC` Timezone is that TimeUtils check is only check by regex we given, maybe we should also check the timezone exists in `ZoneId.getAvailableZoneIds`?

This timezone feature is introduced by @HangyuanLiu , as he explains, the time zone represented by this time zone abbreviation may be uncertain. For example, CST can be regarded as multiple time zones, so there is no support for such abbreviations, only a few cases are supported.

HangyuanLiu: Yes.  the time zone represented by this time zone abbreviation may be uncertain.
Another reason is the timezone string may transfer to be, the library of timezone in be may not support short time zone.
So may be you should test UTC can run correctly in be time function ,such as from_unixtime

WingsGo:
Hi, @morningman  @HangyuanLiu  I rename `TimeUtils.getTimeZone` to `TimeUtils.getSessionTimeZone` and use the function to get default timezone instead get default time zone from system && add alias `UTC` to `Africa/Abidjan`.We need to support set system time zone in fe.conf and add releated doc, maybe in next PR.

fe/fe-core/src/main/java/org/apache/doris/common/util/TimeUtils.java
```diff
@@ -32,6 +32,7 @@
 
 import org.apache.logging.log4j.LogManager; 
 import org.apache.logging.log4j.Logger;
+import org.apache.parquet.Strings;
```
morningman, line 35: Wrong import

WingsGo: done

WingsGo:
Hi, @morningman, I unify the timezone storage value from ZoneId.getId() in order to get a unify timezone string, and I add simple time_zone_test in BE to ensure they are supported, the purpose of the change is to support timezone format like "+08:00" or "counrty/city" or some specified timezone alias, such as "CST", "PRC" and "UTC" etc, I think it will cover most of situations, If there is some other case, we only need add alias in FE and check it worked in BE.

WingsGo:
Hi, @morningman Can you take some time to review it, Thx~~

PR #4195, opened 2020-07-27
[Bug] Divide the upload in spark repository into two steps

### LINKED ISSUE #4194: [Bug]Broker may fail to write file when fe uploads spark archive
When Fe uploads the spark archive, the broker may fail to write the file, resulting in the bad file being uploaded to the repository. 

Therefore, in order to prevent spark from reading bad files, we need to divide the upload into two steps. The first step is to upload the file, and the second step is to rename the file with MD5 value.

xy720:
## Proposed changes

Fix #4194 

## Types of changes

## Checklist


morningman (approved):
LGTM

PR #4199, opened 2020-07-28
Change type of sum, min, max function column in mv

EmmyMiao87:
## Proposed changes
If the agg function is sum, the type of mv column will be bigint.
The only exception is that if the base column is largeint, the type of mv column will be largeint.

If the agg function is min or max, the type of mv column will be same as the type of base column.
For example, the type of mv column is smallint when the agg function is min.

Change-Id: I1b541fd03342ceba093c4aabd8a29793c52781dc

## Types of changes

## Checklist

- [ ] I have create an issue on [Doris's issues](https://github.com/apache/incubator-doris/issues), and have described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
@@ -355,7 +355,13 @@ public class CreateMaterializedViewStmt extends DdlStmt {
             case "sum":
                 mvColumnName = baseColumnName;
                 mvAggregateType = AggregateType.valueOf(functionName.toUpperCase());
-                type = baseColumnRef.getType().getPrimitiveType() == PrimitiveType.LARGEINT ? Type.LARGEINT : Type.BIGINT;
+                PrimitiveType baseColumnType = baseColumnRef.getType().getPrimitiveType();
+                if (baseColumnType == PrimitiveType.TINYINT || baseColumnType == PrimitiveType.SMALLINT
+                        || baseColumnType == PrimitiveType.INT) {
+                    type = Type.BIGINT;
+                } else {
+                    type = Type.fromPrimitiveType(baseColumnRef.getType().getPrimitiveType());
+                }
                 break;
             case "min":
             case "max":
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
@@ -338,7 +338,7 @@ public class CreateMaterializedViewStmt extends DdlStmt {
         }
     }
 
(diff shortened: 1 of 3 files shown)


kangkaisen:
Please add a UT to cover this PR.

fe/fe-core/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
```diff
@@ -30,11 +34,6 @@
 import org.apache.doris.common.FeNameFormat;
 import org.apache.doris.common.UserException;
```
wutiangan, line 36: plz import the jar package according following sequence：
4=javax
3=java
2=org
1=com
0=org.apache.doris

wutiangan (changes requested):

fe/fe-core/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
kangkaisen: Hi, you could update the `checkstyle.xml`.

EmmyMiao87:
> Please add a UT to cover this PR.

Added

fe/fe-core/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
EmmyMiao87: Changed

kangkaisen:
> If the agg function is sum, the type of mv column will be bigint.

@EmmyMiao87 Hi. for sum(double), sum(decimal), the type of mv column will also be bigint?

EmmyMiao87:
> > If the agg function is sum, the type of mv column will be bigint.
> 
> @EmmyMiao87 Hi. for sum(double), sum(decimal), the type of mv column will also be bigint?

Fixed

fe/fe-core/src/main/java/org/apache/doris/analysis/CreateMaterializedViewStmt.java
```diff
@@ -353,11 +353,21 @@ public MVColumnItem buildMVColumnItem(FunctionCallExpr functionCallExpr) throws
...
         switch (functionName.toLowerCase()) {
             case "sum":
+                mvColumnName = baseColumnName;
+                mvAggregateType = AggregateType.valueOf(functionName.toUpperCase());
+                PrimitiveType baseColumnType = baseColumnRef.getType().getPrimitiveType();
+                if (baseColumnType == PrimitiveType.TINYINT || baseColumnType == PrimitiveType.SMALLINT
+                        || baseColumnType == PrimitiveType.INT) {
+                    type = Type.BIGINT;
+                } else {
+                    type = Type.fromPrimitiveType(baseColumnRef.getType().getPrimitiveType());
```
kangkaisen, line 363: Sum float result type should be double.

kangkaisen (dismissed):
+1

morningman (approved):
LGTM

PR #4217, opened 2020-07-30
[Bug][Load][Json] #4124 Load json format with stream load failed

### LINKED ISSUE #4124: [Bug][Load][Json] Load json format with stream load failed

morningman:
## Proposed changes

Stream load should read all the data completely before parsing the json.
And also add a new BE config `streaming_load_max_batch_read_mb`
to limit the data size when loading json data.

Fix the bug of loading empty json array `[]`

Add doc to explain some certain case of loading json format data.

Fix: #4124

## Types of changes

## Checklist

- [x] I have create an issue on #4124, and have described the bug/feature there in detail

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -303,11 +303,14 @@ namespace config {
     CONF_Int64(load_data_reserve_hours, "4");
     // log error log will be removed after this time
     CONF_mInt64(load_error_log_reserve_hours, "48");
-    // Deprecated, use streaming_load_max_mb instead
-    // CONF_Int64(mini_load_max_mb, "2048");
     CONF_Int32(number_tablet_writer_threads, "16");
 
+    // The maximum amount of data that can be processed by a stream load
     CONF_mInt64(streaming_load_max_mb, "10240");
+    // Some data formats, such as JSON, cannot be streamed.
+    // Therefore, it is necessary to limit the maximum number of
+    // such data when using stream load to prevent excessive memory consumption.
+    CONF_mInt64(streaming_load_max_batch_size_mb, "100");
     // the alive time of a TabletsChannel.
     // If the channel does not receive any data till this time,
     // the channel will be removed.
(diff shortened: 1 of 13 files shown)


be/src/common/config.h
```diff
@@ -303,11 +303,14 @@ namespace config {
     CONF_Int64(load_data_reserve_hours, "4");
     // log error log will be removed after this time
     CONF_mInt64(load_error_log_reserve_hours, "48");
-    // Deprecated, use streaming_load_max_mb instead
-    // CONF_Int64(mini_load_max_mb, "2048");
     CONF_Int32(number_tablet_writer_threads, "16");
 
+    // The maximum amount of data that can be processed by a stream load
```
wuyunfeng, line 308: a stream load can process 10G by default?

be/src/http/action/stream_load.cpp
```diff
@@ -234,11 +234,19 @@ Status StreamLoadAction::_on_header(HttpRequest* http_req, StreamLoadContext* ct
...
             std::stringstream ss;
             ss << "unknown data format, format=" << http_req->header(HTTP_FORMAT_KEY);
             return Status::InternalError(ss.str());
         }
+
+        size_t max_body_bytes = config::streaming_load_max_batch_size_mb * 1024 * 1024;
+        if (ctx->format == TFileFormatType::FORMAT_JSON) {
+            if (ctx->body_bytes > max_body_bytes) {
+                std::stringstream ss;
+                ss << "body exceed max size of json format: " << ctx->body_bytes << ", limit: " << max_body_bytes;
```
wuyunfeng, line 246: the size of this batch exceed the max size of json type data
```suggestion
                ss << "the size of this batch exceed the max size [" << max_body_bytes << "]  of json type data " << " data [ " << ctx->body_bytes << " ] "
```
And I suggest you should truncate the logged `body_bytes` such as just show 1024 byte

be/src/common/config.h
morningman: Actually, this default value is too big for a stream load.
But I am not going to modify this to avoid some user troubles.

be/src/http/action/stream_load.cpp
morningman: OK

yangzhg (approved):
LGTM

PR #4221, opened 2020-07-30
Add approx count distinct function

### LINKED ISSUE #4220: [FEATURE] Support approx_count_distinct function
We currently have a function called "NDV", but the name of this function is not easy to understand, and we cannot literally understand what the function does. 
So we should add a common alias "approx_count_distinct" to "NDV"

HangyuanLiu:
FIX #4220 
This PR not modify logical, only add a alias to ndv

### DIFF
--- a/docs/.vuepress/sidebar/en.js
+++ b/docs/.vuepress/sidebar/en.js
@@ -267,7 +267,7 @@ module.exports = [
               "hll_union_agg",
               "max",
               "min",
-              "ndv",
+              "approx_count_distinct",
               "percentile_approx",
               "stddev",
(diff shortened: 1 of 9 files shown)


kangkaisen:
@HangyuanLiu Hi, you should also add a doc for this function.

kangkaisen (approved):
+1

morningman (changes requested):
This PR just add alias for ndv() function.
But why you remove all docs about ndv?

morningman (approved):
function name `ndv()` is not recommended to be used any more.

Issue #4224, opened 2020-07-30
Config netty param in HttpServer


liutang123:
Now, if the length of URL is longer than 4096 bytes, netty will refuse.
The case can be reproduced by constructing a very long URL(longer than 4096bytes)
```
curl -sS -m 1800 --location-trusted -u user:pass http://FE_HOST:FE_PORT/api/db/table/_load?label=XXXX\&columns=a,b,c,d...
```
Netty will generate an invalid `HttpRequest`:
![image](https://user-images.githubusercontent.com/17537020/88929067-31cdcc00-d2ac-11ea-9046-a7aa5646889d.png)

PR #4225, opened 2020-07-30
Config netty param in HttpServer

liutang123:
for #4224
Add three http param to control the netty behavior.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/http/HttpServer.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/http/HttpServer.java
@@ -17,6 +17,17 @@
 
 package org.apache.doris.http;
 
+import io.netty.bootstrap.ServerBootstrap;
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelInitializer;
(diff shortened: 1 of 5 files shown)


kangkaisen (dismissed):
+1

kangkaisen:
@liutang123 Hi. please add these new configs to doc.

fe/fe-core/src/main/java/org/apache/doris/http/HttpServer.java
```diff
@@ -17,6 +17,17 @@
 
 package org.apache.doris.http;
 
+import io.netty.bootstrap.ServerBootstrap;
```
morningman, line 20: remain the import order.

imay: We'd better to provide some config for popular IDE

liutang123:
> @liutang123 Hi. please add these new configs to doc.

Because `maxChunkSize` is removed after netty 5.x, so I don't add it to docs and just add the other two.

imay (approved):
LGTM

Issue #4241, opened 2020-08-03
Outer join a union view with null column throw Exception


liutang123:
Bug repreduce:
```
CREATE VIEW test_view AS
SELECT NULL AS k
UNION ALL
SELECT NULL AS k
;

EXPLAIN
SELECT v1.k
FROM
test_view AS v1
LEFT OUTER JOIN 
test_view AS v2
on v1.k=v2.k
;
```
or
```
EXPLAIN
WITH test_view(k) AS
(
SELECT NULL AS k
UNION ALL
SELECT NULL AS k
)
SELECT v1.k
FROM
test_view AS v1
LEFT OUTER JOIN 
test_view AS v2
on v1.k=v2.k
;
```

The result is:
```
ERROR 1064 (HY000): errCode = 2, detailMessage = Unexpected exception: Implicit casts should never throw analysis exception.
```
In FE, the log is:
```
2020-08-03 22:57:40,304 ERROR 149 [CastExpr.<init>():64] Implicit casts should never throw analysis exception.
org.apache.doris.common.AnalysisException: errCode = 2, detailMessage = Invalid type cast of if(TupleIsNull(2), NULL, <slot 2>) from BOOLEAN to NULL_TYPE
        at org.apache.doris.analysis.CastExpr.analyze(CastExpr.java:197) ~[palo-fe.jar:?]
        at org.apache.doris.analysis.CastExpr.<init>(CastExpr.java:62) [palo-fe.jar:?]
        at org.apache.doris.analysis.Expr.uncheckedCastTo(Expr.java:1239) [palo-fe.jar:?]
        at org.apache.doris.analysis.TupleIsNullPredicate.wrapExpr(TupleIsNullPredicate.java:138) [palo-fe.jar:?]
        at org.apache.doris.analysis.TupleIsNullPredicate.wrapExprs(TupleIsNullPredicate.java:111) [palo-fe.jar:?]
        at org.apache.doris.planner.SingleNodePlanner.createInlineViewPlan(SingleNodePlanner.java:1035) [palo-fe.jar:?]
        at org.apache.doris.planner.SingleNodePlanner.createTableRefNode(SingleNodePlanner.java:1581) [palo-fe.jar:?]
        at org.apache.doris.planner.SingleNodePlanner.createJoinNode(SingleNodePlanner.java:1518) [palo-fe.jar:?]
[...]

PR #4245, opened 2020-08-04
Do not wrap NULL type expr to IF(TupleIsNull(tids), NULL, expr)

liutang123:
for #4241


imay (dismissed):
LGTM

fe/fe-core/src/main/java/org/apache/doris/analysis/CastExpr.java
```diff
@@ -61,6 +61,7 @@ public CastExpr(Type targetType, Expr e) {
             try {
                 analyze();
             } catch (AnalysisException ex) {
+                LOG.error("Implicit casts fail", ex);
```
morningman, line 64: please use `warn` level.
`error` always means some fatal error.

imay (approved):
LGTM

Issue #4254, opened 2020-08-05
[Tablet][Repair] Support using empty tablet to repair the damaged or missing tablet


morningman:
**Is your feature request related to a problem? Please describe.**

In some very special circumstances, such as code bugs, or human misoperation, etc., all replicas of some tablets may be lost. In this case, the data has been substantially lost. However, in some scenarios, the business still hopes to ensure that the query will not report errors even if there is data loss, and reduce the perception of the user layer. At this point, we can use the blank Tablet to fill the missing replica to ensure that the query can be executed normally.

**Describe the solution you'd like**
Doris already had this feature, but has some problem, and missing manual doc

PR #4255, opened 2020-08-05
[Tablet][Recovery] Support using empty tablet to repair the damaged or missing tablet

### LINKED ISSUE #4274: [Bug] FE crash when doing checkpoint

FE crash on `replayBackendtabletsInfo()`

This is because the checkpoint thread try to get a replica from TabletInvertedIndex.
But TabletInvertedIndex is not initialized in checkpoint thread.

morningman:
## Proposed changes

In some very special circumstances, such as code bugs, or human misoperation, etc., all replicas of some tablets may be lost. In this case, the data has been substantially lost. However, in some scenarios, the business still hopes to ensure that the query will not report errors even if there is data loss, and reduce the perception of the user layer. At this point, we can use the blank Tablet to fill the missing replica to ensure that the query can be executed normally.

Add a new FE config `recover_with_empty_tablet`. default is false. true means to use empty tablet to fill the missing one.

Also fix a bug in Fix #4274

## Types of changes

## Checklist

- [x] I have create an issue on #4254, and have described the bug/feature there in detail


EmmyMiao87 (dismissed):
LGTM

EmmyMiao87 (approved):
LGTM

PR #4257, opened 2020-08-05
secure singleton mode

xinghuayu007:
## Proposed changes

The class org.apache.doris.service.ExecuteEnv is not safe in multiple thread environment. This PR is to implement a secure singleton mode.

## Types of changes

## Checklist

- [] I have create an issue on #ISSUE, and have described the bug/feature there in detail

## Further comments


WingsGo:
Maybe you can reference the implement in `Catalog.SingletonHoder` to get a thread-safe singleton without lock it.

xinghuayu007:
> Maybe you can reference the implement in `Catalog.SingletonHoder` to get a thread-safe singleton instread of use double check lock to implement.

good ideal!

imay (approved):
LGTM

WingsGo:
Hi, @imay The code has no problem in logic, but I think sychornized will have poor performance than static method, why we keep the double check lock? Is there anything i didn't consider into account?thx~~

caiconghui:
> Hi, @imay The code has no problem in logic, but I think sychornized will have poor performance than static method, why we keep the double check lock? Is there anything i didn't consider into account?thx~~

@WingsGo The performance is not poor, only double check in initial operation

imay:
> Hi, @imay The code has no problem in logic, but I think sychornized will have poor performance than static method, why we keep the double check lock? Is there anything i didn't consider into account?thx~~

My opinion is the same with conghui's.

WingsGo:
@caiconghui , @imay The PR look good to me, but I still think the static method's implement is more elegant and less error-prone

morningman:
I have not tested the efficiency issue, but I think that it is better to keep the singleton implementation consistent in a system.
Both should be OK, but should choose one and only use one.

PR #4267, opened 2020-08-06
Revert "[BUG] Using attachement strategy of brpc to send packet with big size."

chaoyli:
Reverts apache/incubator-doris#4237


morningman (approved):
LGTM
This commit can cause coredump.

PR #4268, opened 2020-08-06
[Feature][Web] Add new feature to list all tablets on a particular BE

weizuo93:
## Proposed changes

A new feature has been added to acquire tablet id and schema hash of all the tablets on a particular BE node via Web page，so that more detailed information of each tablet can be obtained according to these tablet id and schema hash. In accordance with different web request, there are two ways (table and json)to show these acquired tablet id and schema hash  on Web page.

## Types of changes

## Checklist

- [] I have create an issue on #ISSUE, and have described the bug/feature there in detail

## Further comments

Not needed.

### DIFF
--- a/be/src/http/CMakeLists.txt
+++ b/be/src/http/CMakeLists.txt
@@ -37,6 +37,7 @@ add_library(Webserver STATIC
   http_client.cpp
   action/mini_load.cpp
   action/health_action.cpp
+  action/tablets_info_action.cpp
   action/checksum_action.cpp
   action/snapshot_action.cpp
   action/reload_tablet_action.cpp
--- a/be/src/http/action/tablets_info_action.cpp
+++ b/be/src/http/action/tablets_info_action.cpp
@@ -18,12 +18,12 @@
 #include "http/action/tablets_info_action.h"
 
 #include <string>
-#include <unistd.h>
 
 #include "http/http_channel.h"
 #include "http/http_request.h"
 #include "http/http_headers.h"
 #include "http/http_status.h"
+#include "service/backend_options.h"
 #include "olap/storage_engine.h"
 #include "olap/tablet_manager.h"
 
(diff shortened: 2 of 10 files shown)


morningman:
That would be a useful feature! But I have some suggestions:

1. There may be hundreds of thousands of tablet on one BE, so it better to add some query parameter like `limit` or `offset` to limit the number of tablets returned. And use a default value, like limit 1000.

2. I sorting out the RESTful API of Doris, in this PR #4261, it only describe the  RESTful API response body of FE, bug I thinks BE can also follow the rule. the response body is like:

```
    {
     	"msg": "OK",
     	"code": 0,
     	"data": {
     		your real result data....
     	},
     	"count": 0
     }
```

3. Document. Every new API need a document in both English and Chinese. You can put them under `administrator-guide/http-actions/`

weizuo93:
> That would be a useful feature! But I have some suggestions:
> 
> 1. There may be hundreds of thousands of tablet on one BE, so it better to add some query parameter like `limit` or `offset` to limit the number of tablets returned. And use a default value, like limit 1000.
> 2. I sorting out the RESTful API of Doris, in this PR #4261, it only describe the  RESTful API response body of FE, bug I thinks BE can also follow the rule. the response body is like:
> 
> ```
>     {
>      	"msg": "OK",
>      	"code": 0,
>      	"data": {
>      		your real result data....
>      	},
>      	"count": 0
>      }
> ```
> 
> 1. Document. Every new API need a document in both English and Chinese. You can put them under `administrator-guide/http-actions/`



Thank you very much for your suggestions, and I think your suggestions are very useful. I will modify my work:

(1) API Response body will follow the rule in PR #4261.

(2) I will add a document in both English and Chinese to describe the new API.

However, the purpose of our work is to view all the tablets on a BE node, it is necessary to return all the tablets. In addition, only the tablet id and schema hash are returned in the data, thus the returned data capacity is not large.

morningman (approved):
LGTM

PR #4275, opened 2020-08-06
Fix cancel query bug

gengjun-git:
ConnectContext.kill() use executor to cancel query, but executor has never been set.


morningman (approved):
LGTM

PR #4284, opened 2020-08-07
[Apache][Doris][Feature][FE]Support users with high QPS in the same query through result set cache

### LINKED ISSUE #4283: [doris][fe]Support users with high QPS in the same query through result cache
- Server Result Cache Concepts

The server result cache is a memory pool . This memory pool consists of the SQL query result cache—which stores results of SQL queries—and the SQL function result cache, which stores values returned by SQL functions.

The benefits of using the server result cache depend on the application. OLAP applications can benefit significantly from its use. Good candidates for caching are queries that access a high number of rows but return a small number, such as those in a data warehouse. 

**Understanding How the Server Result Cache Works**
When a query executes, the database searches the cache memory to determine whether the result exists in the result cache. If the result exists, then the database retrieves the result from memory instead of executing the query. If the result is not cached, then the database executes the query, returns the result as output, and stores the result in the result cache.

When users execute queries and functions repeatedly, the database retrieves rows from the cache, decreasing response time. Cached results become invalid when data in dependent database objects is modified.

songenjie:
-Support users with high QPS in the same query through result set cache

Fixes #4283


fe/fe-core/src/main/java/org/apache/doris/common/FeMetaVersion.java
```diff
@@ -190,6 +190,7 @@
     // force drop db, force drop table, force drop partition
     // make force drop operation do not recycle meta
     public static final int VERSION_89 = 89;
+    // fe result cache
```
yangzhg, line 193: you should add a new Version 91 if you change the meta

fe/fe-core/src/main/java/org/apache/doris/qe/SessionVariable.java
```diff
@@ -623,6 +653,9 @@ public void readFields(DataInput in) throws IOException {
             if (Catalog.getCurrentCatalogJournalVersion() >= FeMetaVersion.VERSION_62) {
                 exchangeInstanceParallel = in.readInt();
             }
+            if (Catalog.getCurrentCatalogJournalVersion() >= FeMetaVersion.VERSION_90) {
```
yangzhg, line 656: this is unnecessary, we use json for  serialization now, this functiuon is deprecated

fe/fe-core/src/main/java/org/apache/doris/common/FeMetaVersion.java
yangzhg, line 193: this is unnecessary

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -1218,4 +1218,24 @@
...
+     * Specify how long an entry will be expired in milliseconds, 10000 by default.
+     */
+    @ConfField(mutable = true)
+    public static long result_cache_expire_after_in_milliseconds = 10*1000;
+
+    /**
+     * Specify the overall threshold of local cache in bytes, 1G bytes by default.
+     */
+    @ConfField(mutable = true)
+    public static long result_cache_size_in_bytes = 1024 * 1024 * 1024;
```
yangzhg, line 1240: you should also limit ervey cache result entity size to avoid one large result use all caches,  we prefer to cache small size and need more time to calculating result,

yangzhg (commented):
you should also add a trigger that when a table has load or delete ops, you should refresh all cache related that table

kangkaisen:
@songenjie Hi, what's your design doc for this PR?  I think this PR is repeated with SQL mode cache in https://github.com/apache/incubator-doris/pull/4330 , what do you think it？

songenjie:
> @songenjie Hi, what's your design doc for this PR? I think this PR is repeated with SQL mode cache in #4330 , what do you think it？

@kangkaisen  

1. Sql Cache Mode 
- If this switch is turned on, the SQL query result set will be cached. If the interval between the last visit version time in all partitions of all tables in the query is greater than cache_last_version_interval_second, and the result set is less than cache_result_max_row_count, the result set will be cached, and the next same SQL will hit the cache.
2. this PR `Result Cache`
-  Solve the same connect ID, the same SQL high QPS problem, the same user, in `result_ cache_ expire_ after_ In_ Within milliseconds` time, executing the same SQL will hit the result cache

kangkaisen:
> 1. this PR `Result Cache`
> 
> * Solve the same connect ID, the same SQL high QPS problem, the same user, in `result_ cache_ expire_ after_ In_ Within milliseconds` time, executing the same SQL will hit the result cache

Hi, I think the selfsame OLAP querys in prod env is very rare. Do you test this PR in your prod env? what's the cache hit rate?

The Apache Kylin use SQL result cache Like this PR,  After Kylin run many years in MEITUAN, we could confirm the SQL result cache is meaningless for OLAP system.

The Apache Druid use segment cache like #4330, After Druid run many years in MEITUAN, we could confirm the segment cache is useful for OLAP system.

So I think If we already have segment query cache, we don't need the query result cache.

songenjie:
> Hi, I think the selfsame OLAP querys in prod env is very rare. Do you test this PR in your prod env? what's the cache hit rate?  The Apache Kylin use SQL result cache Like this PR, After Kylin run many years in MEITUAN, we could confirm the SQL result cache is meaningless for OLAP system.


3ks

the result  cache has been used in Jingdong for more than half a year, and the cache hit rate is also very high, especially in the e-commerce industry like ours, in special periods and business scenarios, its hit rate is very high

As far as I know, most of the business of meituan is near real-time, and the scope of business use and analysis is slightly different from ours. So maybe the hit rate of cache is not high

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -1218,4 +1300,30 @@
...
     public static boolean recover_with_empty_tablet = false;
-}
+
+    /**
+     * Sql_result_cache
+     * Whether or not the result cache is enabled in Fe level, it can be overwritten with connection/session
+     * level setting in Context.
+     */
+    @ConfField(mutable = true)
+    public static boolean enable_result_cache = false;
```
morningman, line 1310: Please add docs for the new config in `administrator-guide/config/fe_config.md`

fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
```diff
@@ -582,31 +592,55 @@ private void handleSetStmt() {
     }
 
     // Process a select statement.
-    private void handleQueryStmt() throws Exception {
+    private boolean handleQueryStmt() throws Exception {
         // Every time set no send flag and clean all data in buffer
         context.getMysqlChannel().reset();
         QueryStmt queryStmt = (QueryStmt) parsedStmt;
+        // Use connection ID as session identifier
+        NamedKey namedKey = new NamedKey(String.valueOf(context.getConnectionId()),
```
morningman, line 600: We had a `originStmt.originStmt`, which is the origin sql string. how about use it instead of `queryStmt.toSql()`?

fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
```diff
@@ -582,31 +592,55 @@ private void handleSetStmt() {
     }
 
     // Process a select statement.
-    private void handleQueryStmt() throws Exception {
+    private boolean handleQueryStmt() throws Exception {
```
morningman, line 595: Add comment to explain the return value.

fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
```diff
@@ -582,31 +592,55 @@ private void handleSetStmt() {
...
         }
         coord = new Coordinator(context, analyzer, planner);
 
-        QeProcessorImpl.INSTANCE.registerQuery(context.queryId(), 
+        QeProcessorImpl.INSTANCE.registerQuery(context.queryId(),
                 new QeProcessorImpl.QueryInfo(context, originStmt.originStmt, coord));
 
-        coord.exec();
+        boolean isCacheEnabled = context.getSessionVariable().isEnableResultCache();
+        LOG.debug("Session level cache is " + (isCacheEnabled ? "enabled" : false));
```
morningman, line 625:
```suggestion
        LOG.debug("Session level cache is {}", (isCacheEnabled ? "enabled" : false));
```

fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
```diff
@@ -582,31 +592,55 @@ private void handleSetStmt() {
...
     // Process a select statement.
-    private void handleQueryStmt() throws Exception {
+    private boolean handleQueryStmt() throws Exception {
         // Every time set no send flag and clean all data in buffer
         context.getMysqlChannel().reset();
         QueryStmt queryStmt = (QueryStmt) parsedStmt;
+        // Use connection ID as session identifier
+        NamedKey namedKey = new NamedKey(String.valueOf(context.getConnectionId()),
+                StringUtils.toUtf8(queryStmt.toSql()));
+        LOG.debug("Result Cache NamedKey [" + namedKey + "]");
```
morningman, line 602:
```suggestion
        LOG.debug("Result Cache NamedKey [{}]", namedKey);
```

fe/fe-core/src/main/java/org/apache/doris/cache/SimpleLocalCache.java
```diff
@@ -0,0 +1,164 @@
...
+    @Override
+    public byte[] get(NamedKey key) {
+        return decompress(cache.getIfPresent(key));
+    }
+
+    @Override
+    public void put(NamedKey key, byte[] value) {
+        final byte[] compresssize = compress(value);
+        if (compresssize.length > Config.result_cache_size_per_query_in_bytes) {
+            LOG.info(" result size more than result_cache_size_per_query_in_bytes: "
```
morningman, line 91:
```suggestion
            LOG.debug(" result size more than result_cache_size_per_query_in_bytes: "
```

fe/fe-core/src/main/java/org/apache/doris/cache/SimpleLocalCache.java
```diff
@@ -0,0 +1,164 @@
...
+                    + Config.result_cache_size_per_query_in_bytes + " so not storage in cache");
+            return;
+        }
+        cache.put(key, compress(value));
```
morningman, line 95:
```suggestion
        cache.put(key, compresssize);
```

morningman:
I didn't see any code about cache invalid? Did I miss something?

songenjie:
> I didn't see any code about cache invalid? Did I miss something?

https://github.com/apache/incubator-doris/blob/6c2ca4690139acdf917193ec4d0acda39b7a148b/fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java#L635

fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
songenjie: 3ks,done

morningman (commented):
> > I didn't see any code about cache invalid? Did I miss something?
> 
> https://github.com/apache/incubator-doris/blob/6c2ca4690139acdf917193ec4d0acda39b7a148b/fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java#L635

I mean, where is the logic for judging cache invalidation? The cache key (NamedKey) is only composed by connection id and  origin stmt? What if the data has been unchanged?

fe/fe-core/src/main/java/org/apache/doris/qe/SessionVariable.java
```diff
@@ -37,7 +37,7 @@
 
 // System variable
 public class SessionVariable implements Serializable, Writable {
-    
+
```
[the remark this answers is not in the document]
songenjie: this LOG object no  used in SessionVariable classs, could i clear it ?

songenjie:
> I mean, where is the logic for judging cache invalidation? The cache key (NamedKey) is only composed by connection id and origin stmt? What if the data has been unchanged?

https://github.com/apache/incubator-doris/blob/dd42c6b580afcbb53436dd8b531e03e5fa5d01ed/fe/fe-core/src/main/java/org/apache/doris/cache/SimpleLocalCache.java#L63

use  Caffeine.newBuilder().recordStats();

morningman:
> > I mean, where is the logic for judging cache invalidation? The cache key (NamedKey) is only composed by connection id and origin stmt? What if the data has been unchanged?
> 
> https://github.com/apache/incubator-doris/blob/dd42c6b580afcbb53436dd8b531e03e5fa5d01ed/fe/fe-core/src/main/java/org/apache/doris/cache/SimpleLocalCache.java#L63
> 
> use Caffeine.newBuilder().recordStats();

So it only depends on time? If data changed within `Config.result_cache_expire_after_in_milliseconds`, user in same context still read the old data?

songenjie:
> > > I mean, where is the logic for judging cache invalidation? The cache key (NamedKey) is only composed by connection id and origin stmt? What if the data has been unchanged?
> > 
> > 
> > https://github.com/apache/incubator-doris/blob/dd42c6b580afcbb53436dd8b531e03e5fa5d01ed/fe/fe-core/src/main/java/org/apache/doris/cache/SimpleLocalCache.java#L63
> > 
> > use Caffeine.newBuilder().recordStats();
> 
> So it only depends on time? If data changed within `Config.result_cache_expire_after_in_milliseconds`, user in same context still read the old data?

right

Issue #4290, opened 2020-08-07
[BUG]The query id cannot be repeated when the query is retried


gengjun-git:
Now, query id is generated before handleQuery, this will cause unexpected bug when query is retried. So we may generate query id in every handle function.
https://github.com/apache/incubator-doris/blob/master/fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java#L229

gengjun-git:
fixed

PR #4291, opened 2020-08-07
[Bug fix]fix query id assign bug

gengjun-git:
for #4290


chaoyli (dismissed):

morningman (changes requested):
the QueryId is not only used for `QueryStmt`. So better not delete the origin place where the uuid is generated.
You can reset the uuid when query retried.

imay (approved):
LGTM

morningman (approved):
LGTM

Issue #4051, opened 2020-07-08
[Proposal] Doris supports batch delete On UNIQUE_KEYS table


yangzhg:
# <center>Doris supports batch delete </center>

## BACKGROUND

At present, Doris supports multiple import methods such as broker load, routine load, stream load, etc. For the deletion of data, it can only be deleted through the delete statement. When the delete statement is used, a new version of the data will be generated every time the delete is executed. If frequent deletions will seriously affect query performance, and when using delete mode to delete, it is achieved by generating an empty rowset to record the deletion conditions. Each read must filter the deletion jump condition, also when there are many conditions Impact on performance. Compared with other systems, the implementation of greenplum is more like a traditional database product, and snowflake is implemented through merge syntax.

For a scenario similar to the import of cdc data, insert and delete are generally interspersed in the data data. In the face of this scenario, our current import method cannot be satisfied, even if we can separate insert and delete. Problem, but still cannot solve the problem of deletion.

### Functional level:

Enhance the import function so that it can support the following scenarios:

* ~~Simple batch import, currently supported~~
* Batch point deletion
* Import and delete mixed data import
* only support for unique_keys

### Ease of use:

Minimize the modification of import syntax, and be compatible with the current import syntax

### Performance

Import and read performance should be basically the same as the current import method, and there should not be too much performance loss

## detailed design

The import syntax is to add a column to indicate whether the current row is imported or deleted. If there is no default behavior to insert rows, the function of this level of upgrade is only implemented on segmentV2, v1 is not considered for the time being, in the index file of the segment file `IndexRegion` A bitmap index similar to null bitmap is added to mark the rows to be deleted.

### Data structure design

A bitmap index (delete_index_page) needs to be added to the segment structure to indicate which row is marked for deletion. The PagePointerPB structure is the same as previously defined, using bitmap as the index.

```
message SegmentFooterPB {
    optional uint32 version = 1 [default = 1]; // file version
    repeated ColumnMetaPB columns = 2; // tablet schema
    optional uint32 num_rows = 3; // number of values
    optional uint64 index_footprint = 4; // total idnex footprint of all columns
    optional uint64 data_footprint = 5; // total data footprint of all columns
    optional uint64 raw_data_footprint = 6; // raw data footprint

    optional CompressionTypePB compress_type = 7 [default = LZ4F]; // default compression type for file columns
    repeated MetadataPairPB file_meta_datas = 8; // meta data of file

    // Short key index's page
[...]

PR #4310, opened 2020-08-10
Support batch delete [part 1]

yangzhg:
## Proposed changes
* Implements the grammar of the batch delete #4051 
* Process create, alter table when table has delete sign column
* Support the syntax for enabling the delete column
* Automatically filtered deleted data in the select statement.
* Automatically add delete sign when create  rollup table
TODO:
 * Optimize the reading and compaction logic on the be side, so that the data marked as deleted will be completely deleted during base compaction

## Types of changes

## Checklist

- [x] I have create an issue on #ISSUE, and have described the bug/feature there in detail


be/src/olap/field.h
```diff
@@ -261,6 +262,7 @@ class Field {
     const KeyCoder* _key_coder;
     uint16_t _index_size;
     bool _is_nullable;
+    std::string _name;
```
sduzh, line 265: Place `_name` under `_key_coder` could make a lesser size of `Field`.

be/src/olap/field.h
```diff
@@ -55,6 +55,7 @@ class Field {
     inline int32_t length() const { return _length; }
     inline size_t field_size() const { return size() + 1; }
     inline size_t index_size() const { return _index_size; }
+    inline std::string name() const { return _name; }
```
sduzh, line 58: `const std::string& name() const { return _name; }`

be/src/olap/schema.h
```diff
@@ -53,19 +53,21 @@ class Schema {
...
         _init(columns, col_ids, num_key_columns);
     }
 
     // All the columns of one table may exist in the columns param, but col_ids is only a subset.
     Schema(const std::vector<TabletColumn>& columns, const std::vector<ColumnId>& col_ids) {
         size_t num_key_columns = 0;
-        for (const auto& c: columns) {
-            if (c.is_key()) {
+        for (size_t i = 0; i < columns.size(); ++i) {
+            if (columns.at(i).is_key()) {
```
sduzh, line 64: `std::vector::at` has extra runtime overhead, compared with `operator[]`.

fe/fe-core/pom.xml
```diff
@@ -36,7 +36,7 @@ under the License.
 
     <properties>
         <doris.home>${basedir}/../../</doris.home>
-        <fe_ut_parallel>${env.FE_UT_PARALLEL}</fe_ut_parallel>
+        <fe_ut_parallel>1</fe_ut_parallel>
```
morningman, line 39: Why?

fe/fe-core/src/main/java/org/apache/doris/analysis/AlterTableStmt.java
```diff
@@ -70,6 +76,50 @@ public void analyze(Analyzer analyzer) throws UserException {
...
+        for (AlterClause alterClause : ops) {
+            if (alterClause instanceof EnableFeatureClause
+                    && ((EnableFeatureClause) alterClause).getFeature() == EnableFeatureClause.Features.BATCH_DELETE) {
+                if (table.getKeysType() != KeysType.UNIQUE_KEYS) {
+                    throw new AnalysisException("Batch delete only support in unique tables.");
+                }
+                // has rollup table
+                if (table.getVisibleIndex().size() > 1) {
+                    for (MaterializedIndex idx : table.getVisibleIndex()) {
+                        if (idx.getId() == table.getBaseIndexId()) {
```
morningman, line 90: miss to modify the base index?

fe/fe-core/src/main/java/org/apache/doris/analysis/DataDescription.java
```diff
@@ -172,6 +183,9 @@ public DataDescription(String tableName,
         this.columnMappingList = columnMappingList;
         this.whereExpr = whereExpr;
         this.srcTableName = srcTableName;
+        this.mergeType = mergeType;
+        this.deleteCondition = deleteCondition
+        ;
```
morningman, line 188: code style

fe/fe-core/src/main/java/org/apache/doris/analysis/AlterTableStmt.java
yangzhg: add  a column to rollup index it will add to base table automaticlly, if add a column here it will duplicated

fe/fe-core/pom.xml
yangzhg: This can avoid the problem that intilj will prompt a pom syntax error when the environment variable is not set

morningman (dismissed):
LGTM

morningman (approved):
Approve

PR #4324, opened 2020-08-11
[Bug] Fix bug that modification of global variable can not be persisted.

### LINKED ISSUE #4323: [Bug] Modification of global variables is not correctly persisted.
**Describe the bug**
After executing `set global default_rowset_type=beta` and then restart the FE.
the variables remain unchanged in result of `show variabels.`

**Expected behavior**
the `default_rowset_type` should be `beta` after FE restart

morningman:
## Proposed changes

When setting global variables, such as `set global default_rowset_type=beta`,
the operation is not correctly persisted.

This CL change the fe meta version to 90.

---------------

The main reason for this problem is that for the modification of global variable,
we directly use Java's reflection mechanism to modify static member variables in `GlobalVariable` class.

But in the persistence method of the `set` operation, we only persist the value stored
in the `globalSessionVariable` variable, and this variable does not contain Global Variable.

So I added a new OperationType: `OP_GLOBAL_VARIABLE_V2`,
and added a `GlobalVarPersistInfo` class to record all changes.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4323), and have described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/SetVar.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/SetVar.java
@@ -143,8 +143,8 @@ public class SetVar {
         }
 
         if (getVariable().toLowerCase().equals("exec_mem_limit")) {
-            this.value = new StringLiteral(TimeUtils.checkTimeZoneValidAndStandardize(getValue().getStringValue()));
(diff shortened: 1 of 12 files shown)


EmmyMiao87 (approved):
LGTM

PR #4327, opened 2020-08-11
[Metrics] Support tablet level metrics

acelyc111:
## Proposed changes

Sometimes we want to detect the hotspot of a cluster, for example, hot scanned tablet, hot wrote tablet, but we have no insight about tablets in the cluster.
This patch introduce tablet level metrics to help to achieve this object, now support 4 metrics on tablets: `query_scan_bytes `, `query_scan_rows `, `flush_bytes `, `flush_count `. However, one BE may holds hundreds of thousands of tablets, so I add a parameter for the metrics HTTP request, and not return tablet level metrics by default.

ref https://github.com/apache/incubator-doris/issues/3517 https://github.com/apache/incubator-doris/issues/4027
## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/be/src/http/action/metrics_action.cpp
+++ b/be/src/http/action/metrics_action.cpp
@@ -34,13 +34,14 @@ namespace doris {
 
 void MetricsAction::handle(HttpRequest* req) {
     const std::string& type = req->param("type");
+    const std::string& with_tablet = req->param("with_tablet");
     std::string str;
     if (type == "core") {
         str = _metric_registry->to_core_string();
     } else if (type == "json") {
-        str = _metric_registry->to_json();
+        str = _metric_registry->to_json(with_tablet == "true");
     } else {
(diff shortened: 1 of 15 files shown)


be/src/http/action/metrics_action.cpp
```diff
@@ -34,13 +34,14 @@ namespace doris {
 
 void MetricsAction::handle(HttpRequest* req) {
     const std::string& type = req->param("type");
+    const std::string& with_tablet = req->param("with_tablet");
```
morningman, line 37: Need to update the doc `administrator-guide/operation/monitor-metrics/be-metrics.md`

morningman:
UT compile failed

acelyc111:
> UT compile failed

Fixed, but I found some UT which are disabled ever, now will execute and fail.

morningman (approved):
LGTM

PR #4333, opened 2020-08-12
FE frontend and backend separation 

hf200012:
According to https://github.com/apache/incubator-doris/issues/4308
In order not to affect the original http server, resubmit the code


hf200012:
这个提交错了大家忽略

PR #4343, opened 2020-08-12
Forbidden aggregated partition key column on mv

### LINKED ISSUE #4342: Be core when add partition on table which has mv
**Describe the bug**
The be core when add partition on table.
This table has materialized view.
The materialized view has a partition column which is not key in mv.

**To Reproduce**
1. create table 

```
test_sys_materialized_view_2_test_add_partiton_with_mv_db.test_sys_materialized_view_2_test_add_partiton_with_mv_tb ( k0 boolean NOT NULL, k1 tinyint NOT NULL, k2 smallint NOT NULL, k3 int NOT NULL, k4 bigint NOT NULL, k5 largeint NOT NULL, k6 decimal(9, 3) NOT NULL, k7 char(5) NOT NULL, k8 date NOT NULL, k9 datetime NOT NULL, k10 varchar(20) NOT NULL, k11 double NOT NULL, k12 float NOT NULL )  PARTITION BY RANGE(k1) ( PARTITION p1 VALUES LESS THAN ("-10"), PARTITION p2 VALUES LESS THAN ("0"), PARTITION p3 VALUES LESS THAN ("10") ) DISTRIBUTED BY HASH(k4, k5) BUCKETS 10
```

2. create materialized view

```
CREATE MATERIALIZED VIEW s8e55_ys_materialized_view_2_test_add_partiton_with_mv_index AS SELECT k0, max(k9), min(k11), sum(k12), hll_union(hll_hash(k7)), bitmap_union(to_bitmap(k1)) from test_sys_materialized_view_2_test_add_partiton_with_mv_db.test_sys_materialized_view_2_test_add_partiton_with_mv_tb group by k0
```

3. add partition

```
ALTER TABLE test_sys_materialized_view_2_test_add_partiton_with_mv_db.test_sys_materialized_view_2_test_add_partiton_with_mv_tb ADD PARTITION p4 VALUES LESS THAN ("20")
```

4. add partition failed
image
image

**Expected behavior**

The partition column also must be the key in materialized view.
So  the mv could not be created.

EmmyMiao87:
## Proposed changes

The partition column of table also must be the key in materialized view.
If not, when user wants to add partition of table, the be will core.
The materialized view could not create partition correctly when partition column has been aggregated.

Fixed #4342
Change-Id: Idee42a49c61034368c018d33c522dfced673f3bd

## Types of changes

## Checklist

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/alter/MaterializedViewHandler.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/alter/MaterializedViewHandler.java
@@ -458,7 +458,14 @@ public class MaterializedViewHandler extends AlterHandler {
                 newMVColumns.add(mvColumnItem.toMVColumn(olapTable));
             }
         } else {
+            List<String> partitionOrDistributedColumnName = olapTable.getPartitionColumnNames();
+            partitionOrDistributedColumnName.addAll(olapTable.getDistributionColumnNames());
             for (MVColumnItem mvColumnItem : mvColumnItemList) {
+                if (partitionOrDistributedColumnName.contains(mvColumnItem.getBaseColumnName().toLowerCase())
+                        && mvColumnItem.getAggregationType() != null) {
+                    throw new DdlException("The partition and distributed columns " + mvColumnItem.getBaseColumnName()
+                                                   + " must be key column in mv");
+                }
                 newMVColumns.add(mvColumnItem.toMVColumn(olapTable));
             }
         }
--- a/fe/fe-core/src/main/java/org/apache/doris/alter/MaterializedViewHandler.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/alter/MaterializedViewHandler.java
@@ -458,7 +458,7 @@ public class MaterializedViewHandler extends AlterHandler {
                 newMVColumns.add(mvColumnItem.toMVColumn(olapTable));
             }
         } else {
-            List<String> partitionOrDistributedColumnName = olapTable.getPartitionColumnNames();
+            Set<String> partitionOrDistributedColumnName = olapTable.getPartitionColumnNames();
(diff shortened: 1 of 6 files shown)


fe/fe-core/src/main/java/org/apache/doris/catalog/OlapTable.java
```diff
@@ -563,10 +563,36 @@ public PartitionInfo getPartitionInfo() {
         return partitionInfo;
     }
 
+    public List<String> getPartitionColumnNames() {
+        List<String> partitionColumnNames = Lists.newArrayList();
+        if (partitionInfo instanceof SinglePartitionInfo) {
+            return partitionColumnNames;
```
wutiangan, line 569: partitionColumnNames has upperCase？

fe/fe-core/src/main/java/org/apache/doris/alter/MaterializedViewHandler.java
```diff
@@ -458,7 +458,14 @@ private RollupJobV2 createMaterializedViewJob(String mvName, String baseIndexNam
                 newMVColumns.add(mvColumnItem.toMVColumn(olapTable));
             }
         } else {
+            List<String> partitionOrDistributedColumnName = olapTable.getPartitionColumnNames();
```
morningman, line 461: A `Set` may be better

fe/fe-core/src/main/java/org/apache/doris/catalog/OlapTable.java
EmmyMiao87: It can but we need to check it ignore case

fe/fe-core/src/main/java/org/apache/doris/alter/MaterializedViewHandler.java
EmmyMiao87: Fixed

morningman (approved):
LGTM

PR #4345, opened 2020-08-12
[Bug][MemTracker] Cleanup the mem tracker's constructor to avoid wrong usage

### LINKED ISSUE #4344: [Bug]BE crash when doing LOADING phase of spark load
**Describe the bug**

```
*** Aborted at 1597223166 (unix time) try "date -d @1597223166" if you are using GNU date ***
PC: @          0x2668380 (unknown)
*** SIGSEGV (@0x0) received by PID 14074 (TID 0x7fd1f4f28700) from PID 0; stack trace: ***
    @     0x7fd26e6fe470 (unknown)
    @          0x2668380 (unknown)
    @          0x1030e99 doris::MemTracker::~MemTracker()
    @          0x164a9a2 doris::PushHandler::_convert_v2()
    @          0x164d3d7 doris::PushHandler::_do_streaming_ingestion()
    @          0x164e00b doris::PushHandler::process_streaming_ingestion()
    @          0x162e056 doris::EngineBatchLoadTask::_push()
    @          0x16304ab doris::EngineBatchLoadTask::_process()
    @          0x162c799 doris::EngineBatchLoadTask::execute()
    @           0xe3798e doris::StorageEngine::execute_task()
    @          0x146ea2f doris::TaskWorkerPool::_push_worker_thread_callback()
    @     0x7fd26e4b31c3 start_thread
    @     0x7fd26e7b012d __clone
```

This is because we use wrong way to create the mem tracker in PushHandler.

morningman:
## Proposed changes

After PR: #4135, If a mem tracker has parent, it should be created by 'CreateTracker'.
So I removed other unused constructors.

And also fix the bug described in #4344 

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4344 ), and have described the bug/feature there in detail

### DIFF
--- a/be/src/runtime/mem_tracker.h
+++ b/be/src/runtime/mem_tracker.h
@@ -579,11 +579,12 @@ class PoolMemTrackerRegistry {
   /// with the process tracker as its parent. There is no explicit per-pool byte_limit
   /// set at any particular impalad, so newly created trackers will always have a limit
   /// of -1.
-  /// TODO(cmy): this method is not used for now. the memtracker returned from here is
+  /// TODO(cmy): this function is not used for now. the memtracker returned from here is
   ///            got from a shared_ptr in `pool_to_mem_trackers_`.
-  ///            I don't know why it need to return a raw pointer.
-  ///            If this method be further used, please notice that.
-  MemTracker* GetRequestPoolMemTracker(
+  ///            This funtion is from
+  ///            https://github.com/cloudera/Impala/blob/495397101e5807c701df71ea288f4815d69c2c8a/be/src/runtime/mem-tracker.h#L497
+  ///            And in impala this function will return a raw pointer.
+  std::shared_ptr<MemTracker> GetRequestPoolMemTracker(
       const std::string& pool_name, bool create_if_not_present);
 
  private:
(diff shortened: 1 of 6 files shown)


be/src/runtime/mem_tracker.h
```diff
@@ -588,14 +579,18 @@ class PoolMemTrackerRegistry {
   /// with the process tracker as its parent. There is no explicit per-pool byte_limit
   /// set at any particular impalad, so newly created trackers will always have a limit
   /// of -1.
+  /// TODO(cmy): this method is not used for now. the memtracker returned from here is
+  ///            got from a shared_ptr in `pool_to_mem_trackers_`.
+  ///            I don't know why it need to return a raw pointer.
+  ///            If this method be further used, please notice that.
```
vagetablechicken, line 585: GetRequestPoolMemTracker is from
https://github.com/cloudera/Impala/blob/495397101e5807c701df71ea288f4815d69c2c8a/be/src/runtime/mem-tracker.h#L497
So we didn't make it shared last time, just Ieave it as it is. Maybe we could return shared ptr, I think there should be no problem.

morningman: Okay, since this function has not been used yet, I will change the return value to shared_ptr

EmmyMiao87 (approved):
LGTM

PR #4348, opened 2020-08-14
[Profile] Add 2 Segment related metrics in query profile

morningman:
## Proposed changes

NumSegmentFiltered and NumSegmentTotal.

The number of segments actually read can be obtained through these two metrics.

## Types of changes

## Checklist


kangkaisen (approved):
+1

PR #4352, opened 2020-08-14
[Doris On ES][Bug-Fix] ES queries always route at same 3 BE nodes (#4351)

### LINKED ISSUE #4351: [Doris On ES][Bug] ES queries always route at same 3 BE nodes
I create a external ES table and test some queries on ES.Then I find all of the queries route at same 3 BE nodes.
I find the code at fe/fe-core/src/main/java/org/apache/doris/planner/EsScanNode.java
227: candidateBeList.add(backendList.get(beIndex++ % numBe));
numBe is always 3 when number of be is greater than 3. So that candidateBeList always add 3 front of beckendList.
It should be corrected like:
candidateBeList.add(backendList.get(beIndex++ % backendList.size()));
Then FE can balance ES queries between BE nodes.

caoyang10:
Fix Issue:
https://github.com/apache/incubator-doris/issues/4351

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/EsScanNode.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/EsScanNode.java
@@ -204,13 +204,14 @@ public class EsScanNode extends ScanNode {
                     String.join(",", unPartitionedIndices),
                     String.join(",", partitionedIndices));
         }
-        int beIndex = random.nextInt(backendList.size());
+        int size = backendList.size();
+        int beIndex = random.nextInt(size);
         List<TScanRangeLocations> result = Lists.newArrayList();
         for (EsShardPartitions indexState : selectedIndex) {
             for (List<EsShardRouting> shardRouting : indexState.getShardRoutings().values()) {
                 // get backends
                 Set<Backend> colocatedBes = Sets.newHashSet();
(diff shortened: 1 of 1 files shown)


fe/fe-core/src/main/java/org/apache/doris/planner/EsScanNode.java
```diff
@@ -224,7 +224,7 @@ private void assignBackends() throws UserException {
                 List<Backend> candidateBeList = Lists.newArrayList();
                 if (usingRandomBackend) {
                     for (int i = 0; i < numBe; ++i) {
-                        candidateBeList.add(backendList.get(beIndex++ % numBe));
+                        candidateBeList.add(backendList.get(beIndex++ % backendList.size()));
```
wuyunfeng, line 227:
```suggestion
                        candidateBeList.add(backendList.get(beIndex++ % size));
```
obtain size = backendList.size() in other place

wuyunfeng (commented):
This Looks good to me, but I left some minor comment

caoyang10:
I've obtained size = backendList.size() in the front of loop. Is it better?
please check the commit 6e17409

wuyunfeng (approved):
It would be nice to add some UT to cover this case, but writing this is a bit complicated for you. At this time I would  +1 and  merge this PR soon. Thanks.

PR #4359, opened 2020-08-16
[MySQL Compatibility 1/4][Bug] Fix bug that set sql_mode with concat() function failed

### LINKED ISSUE #4358: [Bug] set sql_mode with concat() function failed.
**Describe the bug**

`set sql_mode = concat(@@sql_mode, "STRICT_TRANS_TABLES");`
This stmt should work, but it failed with error:

` Set statement does't support computing expr:concat(@@sql_mode, 'STRICT_TRANS_TABLES')`

morningman:
## Proposed changes

Support `set sql_mode = concat(@@sql_mode, "STRICT_TRANS_TABLES");`

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4358), and have described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/ExpressionFunctions.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/ExpressionFunctions.java
@@ -22,6 +22,8 @@ import org.apache.doris.catalog.Function;
 import org.apache.doris.catalog.ScalarType;
 import org.apache.doris.catalog.Type;
 import org.apache.doris.common.AnalysisException;
+import org.apache.doris.qe.ConnectContext;
+import org.apache.doris.qe.VariableMgr;
 import org.apache.doris.rewrite.FEFunction;
 import org.apache.doris.rewrite.FEFunctions;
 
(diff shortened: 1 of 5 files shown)


fe/fe-core/src/main/java/org/apache/doris/analysis/ExpressionFunctions.java
```diff
@@ -96,6 +98,14 @@ public Expr evalExpr(Expr constExpr) {
                     return constExpr;
                 }
             }
+        } else if (constExpr instanceof SysVariableDesc) {
+            try {
+                VariableMgr.fillValue(ConnectContext.get().getSessionVariable(), (SysVariableDesc) constExpr);
+                return ((SysVariableDesc) constExpr).getLiteralExpr();
+            } catch (AnalysisException e) {
```
EmmyMiao87, line 105: Why not just throw the exception to the user?

morningman: This method is used for constant folding rules. If it fails, it will directly return to the original expression, and other processes will handle this error.

kangkaisen (approved):
+1

fengyunhe:
BUT still failed for this (when I use SAP DataService to connect Doris ):

`SET SESSION sql_mode = (SELECT CONCAT(@@sql_mode,',ANSI_QUOTES'));`

Issue #4360, opened 2020-08-16
[Bug] select user() stmt does return the IP which user connected in


morningman:
**Describe the bug**

```
mysql> select user();
+----------+
| USER()   |
+----------+
| 'root'@% |
+----------+
```

Expected:

```
mysql> select user();
+-----------------------+
| USER()                |
+-----------------------+
| 'root'@'192.1.68.10.11' |
+-----------------------+
```

PR #4362, opened 2020-08-16
[MySQL Compatibility 2/4][Bug] Fix bug and improve compatibility with mysql protocol

### LINKED ISSUE #4361: [Bug][MySQL compatibility] Query to info schema table does not return correct results.
**Describe the bug**
```
SELECT TABLE_SCHEMA TABLE_CAT, NULL  TABLE_SCHEM,  TABLE_NAME, IF(TABLE_TYPE='BASE TABLE', 'TABLE', TABLE_TYPE) as TABLE_TYPE, TABLE_COMMENT REMARKS, NULL TYPE_CAT, NULL TYPE_SCHEM, NULL TYPE_NAME, NULL SELF_REFERENCING_COL_NAME,  NULL REF_GENERATION FROM INFORMATION_SCHEMA.TABLES  WHERE (ISNULL(database()) OR (TABLE_SCHEMA = database())) AND (TABLE_NAME LIKE '%') AND TABLE_TYPE IN ('BASE TABLE','VIEW','FOREIGN TABLE','MATERIALIZED VIEW','EXTERNAL TABLE') ORDER BY TABLE_TYPE, TABLE_SCHEMA, TABLE_NAME;
```

Above sql will return empty, which expected to return the table info in the current database.

**Why**

This is because `database()` function will return the full name of a database like `default_cluster:db1`.
But `TABLE_SCHEMA` returned from `INFORMATION_SCHEMA.TABLES` is the database name without `default_cluster`.
So the where predicate `TABLE_SCHEMA = database()` is false.

morningman:
## Proposed changes

1. select database() will only return database name, without cluster name. #4361
    
2. select user() will return the IP which user connected in. #4360

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4361 #4360 ), and have described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/InformationFunction.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/InformationFunction.java
@@ -18,6 +18,7 @@
 package org.apache.doris.analysis;
 
 import org.apache.doris.catalog.Type;
+import org.apache.doris.cluster.ClusterNamespace;
 import org.apache.doris.common.AnalysisException;
 import org.apache.doris.qe.ConnectContext;
 import org.apache.doris.thrift.TExprNode;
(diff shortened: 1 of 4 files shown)


fe/fe-core/src/main/java/org/apache/doris/analysis/InformationFunction.java
```diff
@@ -75,3 +80,5 @@ public String toSqlImpl() {
         return funcType + "()";
     }
 }
+
+
```
yangzhg, line 84: two empty lines？

morningman, line 84:
```suggestion
```

kangkaisen (dismissed):
+1

EmmyMiao87 (approved):
LGTM

PR #4371, opened 2020-08-17
[License] Organize and modify the license of the code

morningman:
## Proposed changes

1. Disable the MySQL client and LZO library by default when building the Doris.

    MySQL client library is used for MySQL external table feature.
    This feature will be replaced by the new ODBC external table soon.

    LZO library is used to compress/decompress data of some old data format of Doris,
    which is no longer used anymore.

2. Add missing license to some files.

3. For all non-Apache-License code, all are explained in NOTICE file and the corresponding license is declared.

4. Remove the js source code from webroot, it will be downloaded as thirdparty

## Types of changes

- [x] License

## Checklist


docs/en/README.md
```diff
@@ -74,3 +74,23 @@ cases:
 actionText: Quick Start →
 actionLink: /en/installing/compilation
 ---
+
```
WillemJiang, line 77: Do we need to add the License header on the markdown file?

build.sh
```diff
@@ -140,6 +124,13 @@ if [ ${CLEAN} -eq 1 -a ${BUILD_BE} -eq 0 -a ${BUILD_FE} -eq 0 -a ${BUILD_SPARK_D
     exit 1
 fi
 
+if [[ -z ${WITH_MYSQL} ]]; then
```
WillemJiang, line 127: Just a quick question how can we enable the build with MYSQL?

docs/zh-CN/installing/compilation.md
```diff
@@ -106,3 +106,22 @@ under the License.
     ```
     
     编译完成后，产出文件在 `output/` 目录中。
+
+## 特别声明
+
+自 0.13 版本开始，默认的编译产出中将取消对 [1] 和 [2] 两个第三方库的依赖。这两个第三方库为 [GNU General Public License V3](https://www.gnu.org/licenses/gpl-3.0.en.html) 协议。该协议与 [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0) 协议不兼容，因此默认不出现在 Apache 发布版本中。
+
+依赖库 [1] 会导致无法访问 MySQL 外部表。访问 MySQL 外部表的功能会在后续版本中通过 UnixODBC 实现。
```
WillemJiang, line 114: 建议前面依赖库前面 加一个 “移除”。
还有就是依赖库[1] 应该是MySQL client吧！

WillemJiang, line 114: 建议前面依赖库前面 加一个 “移除”。

docs/en/README.md
morningman: We have added a license header to all markdown files.
But this file is special and needs to be rendered as the homepage of the Doris website through `Vuepress`.
We have tried to add the license, but that will cause the rendering to fail.
Does this need to be specifically stated in the LICENSE file?

build.sh
morningman: Like this:
```
WITH_MYSQL=ON WIHT_LZO=ON sh build.sh
```

I had put it in the document of `compilation.md`

docs/zh-CN/installing/compilation.md
morningman: > 建议前面依赖库前面 加一个 “移除”。
好的

> 还有就是依赖库[1] 应该是MySQL client吧！
事实上，我们在第三方库中下载了完整的 MySQL 源码，其中包括 server 和 client 部分，但是我们只使用了其中的 client 部分。注释 [1] 中标注的 `mysql-5.7.18`，是在 `thirdparty/var.sh` 中写明的依赖名称，感觉两边一致的话，更便于用户查找。

morningman: OK

docs/en/README.md
morningman: Oh, I can put the license in the `frontmatter` section. I will add it

EmmyMiao87 (approved):
Great!

PR #4373, opened 2020-08-17
[Optimize]Optimize the disk selection strategy on BE for tablet creation

### LINKED ISSUE #4329: [Proposal] Optimize the disk selection strategy for tablet creation
When creating a tablet, it is necessary to select a disk from all disks that meet the requirements on the BE node to store the tablet. In doris, the current disk selection strategy is to randomly select a disk from all disks that meet the requirements for tablet creation.  After the cluster has been running for a long time, we found that  the distribution of the number of tablets on different disks in a BE node is severely unbalanced. 

After analysis, we think the main reason for this problem is that the random disk selection strategy used in doris does not consider the problem of load balance between different disks on a BE node.  A good disk selection strategy should take into account both the randomness of selection and disk load. In order to solve this problem, we introduced the the  algorithm of "two random choices" for disk selection when creating the tablet: 
(1) Select two disks from all disks that meet the requirements on the BE node randomly；
(2)  Choose the disk with a smaller number of  tablet from the two disks selected in (1) for tablet creation.

In the initial state, the distribution of tablets among different disks is balanced. The simulation  experiments show that the distribution of tablets on different disks in the BE node always remains relatively balanced with the creation of tablets after used the "two random choices" disk selection algorithm. (The top figure shows the range(极差) trend of tablet distribution on different disks in a BE node with the creation of tablets.  The bottom figure shows the standard deviation(标准差) trend of tablet distribution on different disks in a BE node with the creation of tablets. In these two figures, the red line indicates the random disk selection strategy used in doris for tablet creation and the blue line indicates "two random choices"  disk selection strategy we introduced for tablet creation )
figure_5

In the initial state, the distribution of tablets among different disks is unbalanced. The simulation  experiments show that the "two random choice" disk selection algorithm can also gradually balance the load of different disks with the creation of tablets. (The top figure shows the range(极差) trend of tablet distribution on different disks in a BE node with the creation of tablets.  The bottom figure shows the standard deviation(标准差) trend of tablet distribution on different disks in a BE node with the creation of tablets. In these two figures, the red line indicates the random disk selection strategy used in doris for tablet creation and the blue line indicates "two random choices"  disk selection strategy we introduced for tablet creation )
figure_6

weizuo93:
## Proposed changes

When creating a tablet, it is necessary to select a disk from all disks that meet the requirements on the BE node to store the tablet. In doris, the current disk selection strategy is to randomly select a disk from all disks that meet the requirements for tablet creation. After the cluster has been running for a long time, we found that the distribution of the number of tablets on different disks in a BE node is unbalanced. In order to solve this problem, we introduced the algorithm of "two random choices" for disk selection when creating the tablet:
(1) Select two disks from all disks that meet the requirements on the BE node randomly；
(2) Choose the disk with a smaller number of tablet from the two disks selected in (1) for tablet creation.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4329), and have described the bug/feature there in detail

### DIFF
--- a/be/src/olap/storage_engine.cpp
+++ b/be/src/olap/storage_engine.cpp
@@ -429,13 +429,13 @@ std::vector<DataDir*> StorageEngine::get_stores_for_create_tablet(
     srand(rd());
     std::random_shuffle(stores.begin(), stores.end());
     // Two random choices
-    for (int i = 0; i < stores.size();) {
+    for (int i = 0; i < stores.size(); i++) {
         int j = i + 1;
         if (j < stores.size()) {
             if (stores[i]->tablet_set().size() > stores[j]->tablet_set().size()) {
                 std::swap(stores[i], stores[j]);
-            }   
-            i = i + 2;
+            }
+            std::random_shuffle(stores.begin() + j, stores.end());
         } else {
             break;
         }   
--- a/be/src/olap/storage_engine.cpp
+++ b/be/src/olap/storage_engine.cpp
@@ -428,6 +428,18 @@ std::vector<DataDir*> StorageEngine::get_stores_for_create_tablet(
     std::random_device rd;
     srand(rd());
     std::random_shuffle(stores.begin(), stores.end());
+    // Two random choices
+    for (int i = 0; i < stores.size();) {
+        int j = i + 1;
+        if (j < stores.size()) {
+            if (stores[i]->tablet_set().size() > stores[j]->tablet_set().size()) {
+                std::swap(stores[i], stores[j]);
+            }   
+            i = i + 2;
+        } else {
+            break;
+        }   
+    } 
     return stores;
 }
 
--- a/be/src/olap/data_dir.h
+++ b/be/src/olap/data_dir.h
@@ -124,6 +124,8 @@ public:
 
     void update_user_data_size(int64_t size);
 
+    std::set<TabletInfo> tablet_set() { return _tablet_set; }
+
 private:
     std::string _cluster_id_path() const { return _path + CLUSTER_ID_PREFIX; }
     Status _init_cluster_id();


be/src/olap/storage_engine.cpp
```diff
@@ -428,6 +428,18 @@ std::vector<DataDir*> StorageEngine::get_stores_for_create_tablet(
...
     srand(rd());
     std::random_shuffle(stores.begin(), stores.end());
+    // Two random choices
+    for (int i = 0; i < stores.size(); i++) {
+        int j = i + 1;
+        if (j < stores.size()) {
+            if (stores[i]->tablet_set().size() > stores[j]->tablet_set().size()) {
+                std::swap(stores[i], stores[j]);
+            }
+            std::random_shuffle(stores.begin() + j, stores.end());
```
chaoyli, line 438: What's the purpose of the this random_shuffle?

be/src/olap/storage_engine.cpp
```diff
@@ -428,6 +428,18 @@ std::vector<DataDir*> StorageEngine::get_stores_for_create_tablet(
     std::random_device rd;
     srand(rd());
     std::random_shuffle(stores.begin(), stores.end());
+    // Two random choices
+    for (int i = 0; i < stores.size(); i++) {
+        int j = i + 1;
+        if (j < stores.size()) {
+            if (stores[i]->tablet_set().size() > stores[j]->tablet_set().size()) {
+                std::swap(stores[i], stores[j]);
+            }
```
acelyc111, line 437: @weizuo93 you can use `std::sort` directly, and `std::random_shuffle` is not needed?

be/src/olap/storage_engine.cpp
weizuo93: @chaoyli 
The purpose of the this random_shuffle is that:
       If it failed to create tablet on the first disk selected using "two random choices", it is necessary to select another disk from the remaining disks to create the tablet by using "two random choices" again. This random_shuffle is the beginning of the next "two random choices".

be/src/olap/storage_engine.cpp
weizuo93: @acelyc111 
If we use std::sort directly, all tablets will be created on the same disk which has the least amount of tablets. I think it would be better to take into account both the number of tablets on the disk and the randomness of disk selection.

acelyc111: @weizuo93 I got it.

morningman (approved):
LGTM

PR #4378, opened 2020-08-17
FIX: fix dynamic partition replicationNum error

stalary:
dynamic_partition.replication_num default is replication_num, but show create table show -1


fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -3973,43 +3974,55 @@ public static void getDdlStmt(Table table, List<String> createTableStmt, List<St
...
-                sb.append(",\n\"").append(PropertyAnalyzer.PROPERTIES_COLOCATE_WITH).append("\" = \"");
-                sb.append(colocateTable).append("\"");
+                appendProperties(sb, PropertyAnalyzer.PROPERTIES_COLOCATE_WITH, colocateTable);
             }
 
             // dynamic partition
             if (olapTable.dynamicPartitionExists()) {
-                sb.append(olapTable.getTableProperty().getDynamicPartitionProperty().toString());
+                DynamicPartitionProperty dynamicPartitionProperty = olapTable.getTableProperty().getDynamicPartitionProperty();
+                appendProperties(sb, DynamicPartitionProperty.ENABLE, dynamicPartitionProperty.getEnable());
```
morningman, line 4001: This is not a good implments. Could you move these `appendProperties` into a method of `DynamicPartitionProperty`. So that if we add more properties in future, we only need to modify one place.
You can pass the table's default replication number to that method.

stalary: okay，I will modify it later

PR #4379, opened 2020-08-18
[JsonLoad] Fix bug that row num stat is not correct when loading json

morningman:
## Proposed changes

When all fields are null, the row is invalid, it should be filtered

## Types of changes

## Checklist


yangzhg (approved):
LGTM

PR #4387, opened 2020-08-18
[SegmentV2] Change the default storage format to SegmentV2

### LINKED ISSUE #4384: [Bug][SparkLoad] Spark load will create rowset with incorrect rowset type
**Describe the bug**
1. create a table with segment v2 format
2. load data with spark load
3. the rowset with version [2-2] is with storage format SegmentV1, which is expected to be V2.

**Why**

the push handler only set the rowset writer's rowset type by the config of `default_rowset_type` of BE,
not checking the `prefer_rowset_type` of the tablet.

### LINKED ISSUE #4385: [Bug] tablet type in tablet meta is wrong
**Describe the bug**
The `tablet_type` in tablet meta should be `TABLET_TYPE_DISK`.
But it is set to `TABLET_TYPE_MEMORY`

### LINKED ISSUE #4386: [SegmentV2] Change the default storage format to SegmentV2

SInce the Segment V2 has been released for a long time, we should make it as default storage format for newly created table.

morningman:
## Proposed changes

Since the Segment V2 has been released for a long time, we should make it as default storage format for newly created table.

This CL mainly changes:
1. For all newly created tables, their default storage format is Segment V2.
2. For all already exist tablets, their storage format remain unchanged.
3. Fix  bugs described in Fix #4384 and Fix #4385

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4386), and have described the bug/feature there in detail

## Further comments

We should provide a more friendly way to check the conversion progress of Segment V2


yangzhg (commented):
Why not change  the default value of config `default_rowset_type` in `common/config.h` to `BETA`

chaoyli (dismissed):

morningman:
> Why not change the default value of config `default_rowset_type` in `common/config.h` to `BETA`

Modifying this config will cause BE to automatically convert all existing data.
In fact, this is a very bad design because it is completely uncontrollable. . .
But since we have released this feature in version 0.12, even if we make changes, we cannot change the status quo.

I will add some explain in document.

yangzhg:
> > Why not change the default value of config `default_rowset_type` in `common/config.h` to `BETA`
> 
> Modifying this config will cause BE to automatically convert all existing data.
> In fact, this is a very bad design because it is completely uncontrollable. . .
> But since we have released this feature in version 0.12, even if we make changes, we cannot change the status quo.
> 
> I will add some explain in document.

We will change it to BETA in 0.13？ 
the real default is BETA but the value in config says default is ALPHA this it too werid

morningman:
> > > Why not change the default value of config `default_rowset_type` in `common/config.h` to `BETA`
> > 
> > 
> > Modifying this config will cause BE to automatically convert all existing data.
> > In fact, this is a very bad design because it is completely uncontrollable. . .
> > But since we have released this feature in version 0.12, even if we make changes, we cannot change the status quo.
> > I will add some explain in document.
> 
> We will change it to BETA in 0.13？
> the real default is BETA but the value in config says default is ALPHA this it too werid

yes, but I can't find a good way to solve it. I think we can just leave it and explain it in document.

EmmyMiao87 (commented):
Should we remove the config of storage format in BE?

morningman:
> Should we remove the config of storage format in BE?

Can not. Because some user already use it.

EmmyMiao87 (approved):
LGTM

PR #4413, opened 2020-08-21
optimize default value of replication_num

### LINKED ISSUE #4412: Optimize the default value of `replication_num` when create table
In the current code, if we do not specify `replication_num` when creating OlapTable, the default value of 3 will be used, but for some test environments and small clusters, the number of backends is likely to be less than 3, which will cause confusion for users
Therefore, when the number of backends is less than 3 When the default replication_num should be equal to the number of backends, the default value should be 3 when the number of backends is greater than 3.

yangzhg:
## Proposed changes

In the current code, if we do not specify replication_num when creating OlapTable, the default value of 3 will be used, but for some test environments and small clusters, the number of backends is likely to be less than 3, which will cause confusion for users
Therefore, when the number of backends is less than 3 When the default replication_num should be equal to the number of backends, the default value should be 3 when the number of backends is greater than 3.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4412 ), and have described the bug/feature there in detail

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/CreateTableStmt.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/CreateTableStmt.java
@@ -73,6 +73,8 @@ public class CreateTableStmt extends DdlStmt {
     private String comment;
     private List<AlterClause> rollupAlterClauseList;
 
+    private int replicaNum = FeConstants.default_replication_num;
+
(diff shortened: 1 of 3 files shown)


morningman (commented):
I think we'd better use `show warnings` to show user the warning information.
It need more discuss.

PR #4424, opened 2020-08-21
Fix explicit broadcast join bug

wyb:
Use broadcast join when users specify explicitly `[BROADCAST]` in queries.


kangkaisen (approved):
+1

PR #4425, opened 2020-08-22
Support batch delete[part 2] 

### LINKED ISSUE #4051: [Proposal] Doris supports batch delete On UNIQUE_KEYS table

ZhangYu0123:
## Proposed changes

- In base compaction, merge and delete deleted rows  #4051.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4051 ), and have described the bug/feature there in detail

what alternatives you considered, etc...


be/src/olap/field.h
```diff
@@ -535,4 +535,4 @@ class FieldFactory {
 
 }  // namespace doris
 
-#endif // DORIS_BE_SRC_OLAP_FIELD_H
+#endif // DORIS_BE_SRC_OLAP_FIELD_H
```
yangzhg, line 538: why？

be/src/olap/row_cursor.h
```diff
@@ -109,6 +109,14 @@ class RowCursor {
         return column_schema(index)->index_size();
     }
 
+    inline bool is_delete() const {
+        auto sign_idx = _schema->delete_sign_idx();
+        if (sign_idx < 0) {
+            return false;
+        }
+        return (*(char*) (cell(sign_idx).cell_ptr())) > 0;
```
yangzhg, line 117: better user reinterpret_cast

be/src/olap/schema.h
```diff
@@ -158,4 +158,4 @@ class Schema {
     int32_t _delete_sign_idx = -1;
 };
 
-} // namespace doris
+} // namespace doris
```
yangzhg, line 161: why？

be/src/olap/schema.cpp
```diff
@@ -105,4 +105,4 @@ Schema::~Schema() {
     }
 }
 
-}
+}
```
yangzhg, line 108: why？

yangzhg (commented):
LGTM

yangzhg (dismissed):

be/src/olap/reader.cpp
```diff
@@ -1033,14 +1039,20 @@ void Reader::_init_load_bf_columns(const ReaderParams& read_params) {
 }
 
 OLAPStatus Reader::_init_delete_condition(const ReaderParams& read_params) {
+
```
yangzhg, line 1042: unnessary empty lines

be/src/olap/reader.cpp
```diff
@@ -427,14 +430,17 @@ OLAPStatus Reader::_unique_key_next_row(RowCursor* row_cursor, MemPool* mem_pool
                 }
                 break;
             }
-
+
```
yangzhg, line 433: white spaces

be/src/olap/tablet_schema.h
```diff
@@ -134,6 +134,7 @@ class TabletSchema {
     bool _has_bf_fpp = false;
     double _bf_fpp = 0;
     bool _is_in_memory = false;
+
```
yangzhg, line 137: why?

be/src/olap/reader.cpp
```diff
@@ -411,13 +414,13 @@ OLAPStatus Reader::_unique_key_next_row(RowCursor* row_cursor, MemPool* mem_pool
...
             return OLAP_SUCCESS;
         }
-
         cur_delete_flag = _next_delete_flag;
         // the verion is in reverse order, the first row is the highest version,
         // in UNIQUE_KEY highest version is the final result, there is no need to
         // merge the lower versions
         direct_copy_row(row_cursor, *_next_key);
         agg_finalize_row(_value_cids, row_cursor, mem_pool);
+
```
yangzhg, line 423: why?

yangzhg (approved):
LGTM

PR #4440, opened 2020-08-24
[refactor] Optimize threads usage mode in BE

acelyc111:
## Proposed changes

BE can not graceful exit because some threads are running in endless
loop. This patch do the following optimization:
- Use the well encapsulated Thread and ThreadPool instead of std::thread
  and std::vector<std::thread>
- Use CountDownLatch in thread's loop condition to avoid endless loop
- Introduce a new class Daemon for daemon works, like tcmalloc_gc,
  memory_maintenance and calculate_metrics
- Decouple statistics type TaskWorkerPool and StorageEngine notification
  by submit tasks to TaskWorkerPool's queue
- Reorder objects' stop and deconstruct in main(), i.e. stop network
  services at first, then internal services
- Use libevent in pthreads mode, by calling evthread_use_pthreads(),
  then EvHttpServer can exit gracefully in multi-threads
- Call brpc::Server's Stop() and ClearServices() explicitly

## Types of changes

## Checklist

## Further comments

none

### DIFF
--- a/be/src/olap/olap_server.cpp
+++ b/be/src/olap/olap_server.cpp
@@ -363,7 +363,7 @@ void StorageEngine::_path_scan_thread_callback(DataDir* data_dir) {
         LOG(INFO) << "try to perform path scan!";
         data_dir->perform_path_scan();
 
-        int32_t interval = config::path_scan_interval_second;
+        interval = config::path_scan_interval_second;
         if (interval <= 0) {
             LOG(WARNING) << "path gc thread check interval config is illegal:" << interval
                          << "will be forced set to one day";
--- a/be/src/olap/olap_server.cpp
+++ b/be/src/olap/olap_server.cpp
@@ -39,42 +39,28 @@ using std::string;
 
 namespace doris {
 
-// TODO(yingchun): should be more graceful in the future refactor.
-#define SLEEP_IN_BG_WORKER(seconds)               \
-  int64_t left_seconds = (seconds);               \
-  while (!_stop_bg_worker && left_seconds > 0) {  \
-      sleep(1);                                   \
-      --left_seconds;                             \
-  }                                               \
-  if (_stop_bg_worker) {                          \
-      break;                                      \
-  }
-
 // number of running SCHEMA-CHANGE threads
 volatile uint32_t g_schema_change_active_threads = 0;
 
 Status StorageEngine::start_bg_threads() {
-    _unused_rowset_monitor_thread =  std::thread(
-        [this] {
-            _unused_rowset_monitor_thread_callback(nullptr);
-        });
-    _unused_rowset_monitor_thread.detach();
+    RETURN_IF_ERROR(
+        Thread::create("StorageEngine", "unused_rowset_monitor_thread",
+                       [this]() { this->_unused_rowset_monitor_thread_callback(); },
+                       &_unused_rowset_monitor_thread));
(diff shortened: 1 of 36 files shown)


morningman (dismissed):
LGTM

morningman:
Hi @acelyc111 please resolve the conflict

acelyc111:
> Hi @acelyc111 please resolve the conflict

Resolved

be/src/olap/olap_server.cpp
```diff
@@ -263,27 +238,24 @@ void* StorageEngine::_garbage_sweeper_thread_callback(void* arg) {
...
     ProfilerRegisterThread();
 #endif
-    while (!_stop_bg_worker) {
+
+    int32_t interval = config::disk_stat_monitor_interval;
+    do {
         _start_disk_stat_monitor();
 
-        int32_t interval = config::disk_stat_monitor_interval;
+        interval = config::disk_stat_monitor_interval;
```
chaoyli, line 252: redundant assignment

be/src/olap/olap_server.cpp
```diff
@@ -330,108 +303,92 @@ void* StorageEngine::_cumulative_compaction_thread_callback(void* arg, DataDir*
                 _perform_cumulative_compaction(data_dir);
             }
         }
-        int32_t interval = config::cumulative_compaction_check_interval_seconds;
+
+        interval = config::cumulative_compaction_check_interval_seconds;
```
chaoyli, line 307: redundant assignment

be/src/olap/olap_server.cpp
```diff
@@ -330,108 +303,92 @@ void* StorageEngine::_cumulative_compaction_thread_callback(void* arg, DataDir*
...
-    LOG(INFO) << "try to start path scan thread!";
-
-    while (!_stop_bg_worker) {
+    int32_t interval = config::path_scan_interval_second;
+    do {
         LOG(INFO) << "try to perform path scan!";
-        ((DataDir*)arg)->perform_path_scan();
+        data_dir->perform_path_scan();
 
         int32_t interval = config::path_scan_interval_second;
```
chaoyli, line 366: redundant assignment

be/src/olap/olap_server.cpp
```diff
@@ -208,20 +186,17 @@ void* StorageEngine::_base_compaction_thread_callback(void* arg, DataDir* data_d
             }
         }
 
-        int32_t interval = config::base_compaction_check_interval_seconds;
+        interval = config::base_compaction_check_interval_seconds;
```
chaoyli, line 189: abundant assignment

chaoyli (changes requested):

be/src/olap/olap_server.cpp
acelyc111: Yes, I redundant assign it intentionally here, because `base_compaction_check_interval_seconds` is a mutable config, we can change it by http api dynamically (`/api/update_config?config_name=new_value`), it may be changed when do the next loop.

be/src/olap/olap_server.cpp
acelyc111: Same

be/src/olap/olap_server.cpp
acelyc111: Same

chaoyli (approved):

chaoyli (approved):

PR #4450, opened 2020-08-25
Fix mysql return bug

### LINKED ISSUE #4449: Golang cannot get err when query timeout
**Describe the bug**
Golang cannot get error when query timeout

**To Reproduce**
execute a long time used SQL using driver.Query(SQL) api.
```golang
rows, err := driver.Query(SQL)
if err != nil {
      //print err and return
}
for rows.Next() {
      //read rows
}
```

**Expected behavior**
err != nil but err == nil and rows.Next() return false

gengjun-git:
Fix #4449 
Send fields after first row arrived so that error packet can be send to client when exception thrown from coord.getNext(). Golang and Python can not identify error if fields packet arrived before error packet.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
@@ -640,7 +640,7 @@ public class StmtExecutor {
                 break;
             }
         }
-        if (!isSendFields) {
+        if (!isSendFields && !isOutfileQuery) {
             sendFields(queryStmt.getColLabels(), queryStmt.getResultExprs());
         }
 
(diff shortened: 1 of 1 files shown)


morningman:
Hi @gengjun-git  , you can write "Fix #4449", so that the related issue will be closed automatically when this PR is merged.

fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
```diff
@@ -622,13 +622,15 @@ private void handleQueryStmt() throws Exception {
         RowBatch batch;
         MysqlChannel channel = context.getMysqlChannel();
         boolean isOutfileQuery = queryStmt.hasOutFileClause();
-        if (!isOutfileQuery) {
-            sendFields(queryStmt.getColLabels(), queryStmt.getResultExprs());
-        }
+        boolean isSendFields = false;
```
morningman, line 625: So the reason is that the `fields` has been sent but no data is sent following?

gengjun-git: Yes, fields has been sent, but query may fail because of timeout. In this case, client will get a fields packet, and error packet, but some language mysql driver cannot recognize the error, and returns a success and no rows result.

morningman (approved):
LGTM

PR #4458, opened 2020-08-26
[Bug] Fix bug that memory copy may overflow in MemIndex::load_segment

acelyc111:
## Proposed changes

Segment index file content is not set as 0 when it is constructed in write procedure, so when load index from this file, and meet a null VARCHAR cell, the null field of this cell is 0, but the length field which is not initialized maybe a large random number, then memory copy may cause overflow.
This patch fix this bug, and also skip useless memory copy to improve a bit of performance.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix https://github.com/apache/incubator-doris/issues/4459), and have described the bug/feature there in detail

## Further comments

none

### DIFF
--- a/be/src/olap/olap_index.cpp
+++ b/be/src/olap/olap_index.cpp
@@ -208,6 +208,11 @@ OLAPStatus MemIndex::load_segment(const char* file, size_t *current_num_rows_per
                 memory_copy(mem_ptr, storage_ptr, null_byte);
 
                 // 2. copy length and content
+                bool is_null = *reinterpret_cast<bool*>(mem_ptr);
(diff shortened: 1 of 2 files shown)


morningman (approved):
LGTM

morningman:
I'm curious, why hasn't this part of the code been a problem before?
It looks like this code has been around for a long time.

acelyc111:
> I'm curious, why hasn't this part of the code been a problem before?
> It looks like this code has been around for a long time.

I'm not clear, this issue will only happened when the value is null, and nullable byte is correctly copied

PR #4464, opened 2020-08-27
[Spark load] Fix dpp and submit push task bugs

wyb:
1. fix write dpp result when dpp throw exception
2. boolean value：true, false(IgnoreCase), 0, 1
3. wrong dest column for source data check
4. support * in source file path 
5. if job state is cancelled or finished, submitPushTasks would throw all partitions have no load data exception, because tableToLoadPartitions was already cleaned up

#3433


morningman (approved):
LGTM

Issue #4469, opened 2020-08-27
[Spark Load] Redirect the spark launcher's log to a separate log file


xy720:
Currently we print all spark launcher's log to fe.log. If multiple spark load jobs are submitted at the same time, their logs will be confused and difficult to distinguish.

Therefore, I suggest that the log of each spark load should be separated and written into a single log file. These logs will be saved in the `fe/log/spark-launcher-log/` directory until the old label is cleared.

PR #4470, opened 2020-08-27
[Spark Load] Redirect the spark launcher's log to a separated log file

xy720:
## Proposed changes

#4469 

## Types of changes

## Checklist

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/load/loadv2/SparkLauncherMonitor.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/loadv2/SparkLauncherMonitor.java
@@ -70,7 +70,6 @@ public class SparkLauncherMonitor {
         private SparkLoadAppHandle handle;
         private long submitTimeoutMs;
         private boolean isStop;
-        private StringBuffer outputBuffer;
         private OutputStream outputStream;
 
         private static final String STATE = "state";
@@ -87,7 +86,6 @@ public class SparkLauncherMonitor {
             this.handle = handle;
             this.process = handle.getProcess();
             this.isStop = false;
-            this.outputBuffer = new StringBuffer();
             setSubmitTimeoutMs(DEFAULT_SUBMIT_TIMEOUT_MS);
         }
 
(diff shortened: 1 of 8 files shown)


fe/fe-core/src/main/java/org/apache/doris/load/loadv2/SparkLauncherMonitor.java
```diff
@@ -103,7 +114,7 @@ public void run() {
             try {
                 outReader = new BufferedReader(new InputStreamReader(process.getInputStream()));
                 while (!isStop && (line = outReader.readLine()) != null) {
-                    LOG.info("monitor log: " + line);
+                    outputBuffer.append(line + '\n');
```
morningman, line 117: why not just output to the `outputStream`? So that we don't need to buffer a log of content in memory?

fe/fe-core/src/main/java/org/apache/doris/load/loadv2/SparkLoadJob.java
```diff
@@ -701,6 +702,20 @@ protected long getEtlStartTimestamp() {
         return etlStartTimestamp;
     }
 
+    public SparkLoadAppHandle getHandle() {
+        return sparkLoadAppHandle;
+    }
+
+    public void clearSparkLauncherLog() {
+        String logPath = sparkLoadAppHandle.getLogPath();
```
morningman, line 710: Are you sure the `sparkLoadAppHandle` is not null here?

fe/fe-core/src/main/java/org/apache/doris/load/loadv2/SparkLauncherMonitor.java
xy720: ok. I changed it

fe/fe-core/src/main/java/org/apache/doris/load/loadv2/SparkLoadJob.java
xy720: Not sure. So I made a new check here.

morningman (approved):
LGTM

PR #4471, opened 2020-08-27
[Colocation] Fix Colocation balance endless loop bug

wyb:
1. Only one available backend.
2. All backends are checked but this round is not changed. For example, all backends are on the same host.


morningman (approved):
LGTM

PR #4475, opened 2020-08-27
[Refactor] Refactor storage medium migration task process

morningman:
## Proposed changes

This CL refactor the storage medium migration task process in BE.
I did not modify the execution logic. Just extract part of the logic
in the migration task and put it in `task_work_pool`.

In this way, the migration task is only used to process the migration
from the specified tablet to the specified data dir.

Later, we can use this task to migrate of tablets between different disks. #4476

## Types of changes

## Checklist


EmmyMiao87 (approved):
LGTM

Issue #4481, opened 2020-08-28
 Do not add exchange when table's distributioin satisfy the distribution requirements


liutang123:
issue Reason
--
Now, Doris will create two fragments for aggretion.
Sometime, **Exchange for aggregation is unnecessary.**
Think about follow cases:

1. aggregate an unpartitioned table.
create table SQL:
```
CREATE TABLE `llj_test_1` (
  `dt` int(11) NOT NULL COMMENT "",
  `dis_key` varchar(20) NOT NULL COMMENT ""
) ENGINE=OLAP
DUPLICATE KEY(`dt`, `dis_key`)
COMMENT "OLAP"
DISTRIBUTED BY HASH(`dt`, `dis_key`) BUCKETS 3
PROPERTIES (
"replication_num" = "3",
"in_memory" = "false",
"storage_format" = "DEFAULT"
);
```
query SQL:`select dt, dis_key,count(1) from  llj_test_1  group by dt, dis_key;`
2. aggregate a partitioned table.
create table SQL:
```
CREATE TABLE `llj_test` (
  `dt` int(11) NOT NULL COMMENT "",
  `dis_key` varchar(20) NOT NULL COMMENT ""
) ENGINE=OLAP
DUPLICATE KEY(`dt`, `dis_key`)
COMMENT "OLAP"
PARTITION BY RANGE(`dt`)
(PARTITION p20180822 VALUES [("19000101"), ("20181021")),
PARTITION p20181207 VALUES [("20181021"), ("20181022")))
DISTRIBUTED BY HASH(`dt`, `dis_key`) BUCKETS 3
PROPERTIES (
"replication_num" = "1",
"in_memory" = "false",
"storage_format" = "DEFAULT"
);
```
query SQL:`select dt, dis_key,count(1) from  llj_test  group by dt, dis_key;`

Suggestion
--
In `DistributedPlanner`, do not add the unnecessary `Exchange`s.
For case 1, we only need to judge that the table's distribute hash keys is a subset of the aggregate keys.
For case 2, we should jude two conditions:
- partition keys are also hash keys.
- the table's distribute hash keys is a subset of the aggregate keys.

PR #4482, opened 2020-08-28
Do not add exchange when table's distributioin satisfy the distribution requirements

liutang123:
For #4481.

## Proposed changes

In DistributedPlanner, do not add the unnecessary Exchanges.
For case 1, we only need to judge that the table's distribute hash keys is a subset of the aggregate keys.
For case 2, we should jude two conditions:
 - partition keys are also hash keys.
 - the table's distribute hash keys is a subset of the aggregate keys.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/OlapTable.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/OlapTable.java
@@ -19,6 +19,9 @@ package org.apache.doris.catalog;
 
 import org.apache.doris.alter.MaterializedViewHandler;
 import org.apache.doris.analysis.CreateTableStmt;
+import org.apache.doris.analysis.Expr;
+import org.apache.doris.analysis.SlotDescriptor;
+import org.apache.doris.analysis.SlotRef;
 import org.apache.doris.backup.Status;
 import org.apache.doris.backup.Status.ErrCode;
 import org.apache.doris.catalog.DistributionInfo.DistributionInfoType;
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/org/apache/doris/catalog/OlapTable.java
```diff
@@ -1577,4 +1580,31 @@ public TStorageFormat getStorageFormat() {
         }
         return tableProperty.getStorageFormat();
     }
+
+    public boolean satisfyHashDistribution(List<Expr> hashExprs) {
+        if (hashExprs == null) {
```
kangkaisen, line 1585: partitionExprs is null has checked in  `if (partitionExprs == null) partitionExprs = groupingExprs;`

fe/fe-core/src/main/java/org/apache/doris/catalog/OlapTable.java
```diff
@@ -1577,4 +1580,31 @@ public TStorageFormat getStorageFormat() {
         }
         return tableProperty.getStorageFormat();
     }
+
+    public boolean satisfyHashDistribution(List<Expr> hashExprs) {
```
kangkaisen, line 1584: Rename to `couldLocalGorupBy` or similar name?

fe/fe-core/src/main/java/org/apache/doris/planner/DistributedPlanner.java
```diff
@@ -863,6 +863,14 @@ private PlanFragment createMergeAggregationFragment(AggregationNode node, PlanFr
             //     childFragment.addPlanRoot(node);
             //     return childFragment;
             // }
+
+            PlanNode childPlan = childFragment.getPlanRoot();
+            if (childPlan instanceof OlapScanNode &&
+                    ((OlapScanNode)childPlan).getOlapTable().satisfyHashDistribution(partitionExprs)) {
```
kangkaisen, line 869: I think we move this check to `createAggregationFragment` 780 line is better?  Because `createMergeAggregationFragment ` should create a merge aggregate fragment.

fe/fe-core/src/main/java/org/apache/doris/catalog/OlapTable.java
liutang123: Rename to `meetAggDistributionRequirements` and change the param to `AggregateInfo`.

kangkaisen (approved):
+1

PR #4388, opened 2020-08-18
Add OLAP_ERR_DATE_QUALITY_ERR error status to display schema change failure

HangyuanLiu:
## Proposed changes

In the process of historical data transformation of materialized views, it may occur that the transformation fails due to data quality. 
Add an error status code ：` OLAP_ERR_DATE_QUALITY_ERR ` to determine if a data problem is causing the failure

#3344

## Types of changes

What types of changes does your code introduce to Doris?
_Put an `x` in the boxes that apply_

- [] Bugfix (non-breaking change which fixes an issue)
- [] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [] Documentation Update (if none of the other choices apply)
- [] Code refactor (Modify the code structure, format the code, etc...)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [] I have create an issue on (Fix #ISSUE), and have described the bug/feature there in detail
- [] Compiling and unit tests pass locally with my changes
- [] I have added tests that prove my fix is effective or that my feature works
- [] If this change need a document change, I have updated the document
- [] Any dependent changes have been merged

## Further comments

If this is a relatively large or complex change, kick off the discussion at dev@doris.apache.org by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #4486, opened 2020-08-31
[Bug] Fix bug of load error hub and schema change

morningman:
## Proposed changes

1. When WITH_MYSQL is off, load error hub does not suport MySQL load error hub,
   we should check its return value. (introduced by #4371 )

2. misjudge the return value of `change_row_block` in schema_change.cpp (Introduced by #4388 )

## Types of changes

### DIFF
--- a/be/src/olap/schema_change.cpp
+++ b/be/src/olap/schema_change.cpp
@@ -1197,10 +1197,9 @@ OLAPStatus SchemaChangeWithSorting::process(RowsetReaderSharedPtr rowset_reader,
         }
 
         uint64_t filtered_rows = 0;
-        if (!_row_block_changer.change_row_block(ref_row_block, rowset_reader->version().second,
-                                                 new_row_block, &filtered_rows)) {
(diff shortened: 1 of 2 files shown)


be/src/olap/schema_change.cpp
```diff
@@ -1197,10 +1197,9 @@ OLAPStatus SchemaChangeWithSorting::process(RowsetReaderSharedPtr rowset_reader,
         }
 
         uint64_t filtered_rows = 0;
-        if (!_row_block_changer.change_row_block(ref_row_block, rowset_reader->version().second,
-                                                 new_row_block, &filtered_rows)) {
+        if ((res = _row_block_changer.change_row_block(ref_row_block, rowset_reader->version().second,
+                                                 new_row_block, &filtered_rows)) != OLAP_SUCCESS) {
```
yangzhg, line 1201: It is better not to assign values in if statements

morningman: OK

yangzhg (approved):
+1

PR #4495, opened 2020-09-01
[Bug] function str_to_date()'s behavior on BE and FE is inconsistent

### LINKED ISSUE #4494: [Bug] function str_to_date()'s behavior on BE and FE is inconsistent
**Describe the bug**
On FE, `str_to_date()` is implemented by `joda.DateTimeFormatter`,
and on BE, it is implemented by Doris it self.

These 2 implementations are inconsistent in some cases:

1. FE can not to the constant folding for `str_to_date("2020-08-08", "%Y-%m-%d %H:%i:%s")`
    It falls to BE and return `2020-08-08`, but in mysql, it returns `2020-08-08 00:00:00`.

2. `select str_to_date("20-09-01", "%Y-%m-%d");`
    FE return `0020-09-01`, but BE return `2020-09-01`, which is same as MySQL.

3. Some BI will generate predicate like
    `DATE_ADD(DATE_ADD(str_to_date(DATE_FORMAT(NOW(),'%Y-%m-%d'),'%Y-%m-%d %H:%i:%s'),INTERVAL 1 DAY),INTERVAL -1 SECOND)`

    It should be able to do the constant folding, but failed on `str_to_date`.

**Resolution**

We should keep the  behavior on BE and FE consistency.

morningman:
## Proposed changes

Main CL:
1. Copy the code from BE to implement the `str_to_date()` function in FE. 
2. `str_to_date("2020-08-08", "%Y-%m-%d %H:%i:%s")` will return `2020-08-08 00:00:00` instead of `2020-08-08`.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4494), and have described the bug/feature there in detail


fe/fe-core/src/main/java/org/apache/doris/analysis/DateLiteral.java
```diff
@@ -632,4 +669,435 @@ public long getSecond() {
     public int hashCode() {
         return 31 * super.hashCode() + Objects.hashCode(unixTimestamp(TimeZone.getDefault()));
     }
+
+    // parset the date string value in 'value' by 'format' pattern.
+    // return the next position to parse if hasSubVal is true.
+    // throw InvalidFormatException if encounter errors.
+    // this method is exaclty same as from_date_format_str() in be/src/runtime/datetime_value.cpp
+    // change this method should also change that.
+    public int fromDateFormatStr(String format, String value, boolean hasSubVal) throws InvalidFormatException {
```
kangkaisen, line 678: Move these util methods to a single util class?

morningman: It is now only used for DateLiteral, and this method is also in `datetime_value.cpp` in BE.
So I think we can just leave it here.

kangkaisen: Ok. I think the Literal class should be simple. But which isn't a big issue.

kangkaisen (approved):
+1

PR #4510, opened 2020-09-02
Add a session variable to show or hide hidden columns

### LINKED ISSUE #4509: Add a session variable to show or hide hidden columns
Sometimes it is necessary to display the value from the hidden column to help debug, such as determining whether the result of batch deletion of imported data is correct, whether the result of compaction is correct, etc. In the current implementation, this information is completely hidden and cannot be viewed. Therefore, you need to add a session variable `show_hidden_columns` to control whether to show hidden columns and filter hidden column data.
When `show_hidden_columns` is true, functions that get the table's schema such as get_schema will not return hidden columns at the beginning, and the filter condition of `__DELETE_SIGN__=true` will be added to olap sacanNode. When `show_hidden_columns` is false, the opposite is true

yangzhg:
## Proposed changes

Sometimes we need to show hidden columns for debug.
So we need to add a session variable to show or hide hidden columns

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4509), and have described the bug/feature there in detail
- [x If this change need a document change, I have updated the document

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/common/util/Util.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/util/Util.java
@@ -480,12 +480,7 @@ public class Util {
     }
 
     public static boolean showHiddenColumns() {
-        try {
-            return ConnectContext.get().getSessionVariable().showHiddenColumns();
-        } catch (Exception e) {
-            LOG.info("get showHiddenColumns error:", e);
-            return false;
-        }
+        return ConnectContext.get() != null && ConnectContext.get().getSessionVariable().showHiddenColumns();
     }
 }
 
--- a/fe/fe-core/src/main/java/org/apache/doris/common/util/Util.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/util/Util.java
@@ -20,6 +20,7 @@ package org.apache.doris.common.util;
 import org.apache.doris.catalog.Column;
 import org.apache.doris.catalog.PrimitiveType;
 import org.apache.doris.common.AnalysisException;
+import org.apache.doris.qe.ConnectContext;
 
 import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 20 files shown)


chaoyli:
1. Sometimes is when and where to hidden the columns ?
2. Can be hidden the columns transparent to users. Because that it's only meaningful to doris, not to user.

morningman:
Hi @yangzhg , I think we need a document first to explain the logic of whether the hidden column is displayed or not.
So that it can provide guidance for future development.

yangzhg:
> Hi @yangzhg , I think we need a document first to explain the logic of whether the hidden column is displayed or not.
> So that it can provide guidance for future development.

I have update the issue about the  session variable logic, and add batch delete documents in this pr

fe/fe-core/src/main/java/org/apache/doris/common/util/Util.java
```diff
@@ -477,5 +478,14 @@ public static InputStream getInputStreamFromUrl(String urlStr, String encodedAut
         conn.setReadTimeout(readTimeoutMs);
         return conn.getInputStream();
     }
+
+    public static boolean showHiddenColumns() {
+        try {
+            return ConnectContext.get().getSessionVariable().showHiddenColumns();
```
morningman, line 484: ConnectContext.get() may return null, better check it first, and if it is null, return a default value.

fe/fe-core/src/main/java/org/apache/doris/planner/SchemaScanNode.java
```diff
@@ -91,6 +91,7 @@ protected void toThrift(TPlanNode msg) {
                 msg.schema_scan_node.setDb("SESSION");
             }
         }
+        msg.schema_scan_node.show_hidden_cloumns = ConnectContext.get().getSessionVariable().showHiddenColumns();
```
morningman, line 94: Use `Util. showHiddenColumns()` instead

morningman (approved):
LGTM

PR #4364, opened 2020-08-16
[Mysql Compatibility 3/4] Support convert() and signed/unsigned interger cast

morningman:
## Proposed changes

1. Support convert(expr, target_type) function, which is same as CastExpr
2. Support cast (expr as signed/unsigned int)
   This is just for compatibility, the signed/unsigned specification is meaningless.
3. Add more columns for table `information_schema.columns`

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have create an issue on (Fix #4363 ), and have described the bug/feature there in detail
- [x] Compiling and unit tests pass locally with my changes

PR #4511, opened 2020-09-02
[Bug] Fix bug that BE will crash when querying information_schema.columns

morningman:
## Proposed changes

This bug is introduced from #4364

## Types of changes


yangzhg (approved):
+1

PR #4514, opened 2020-09-02
[Feature] support hour time unit with dynamic parition

caoyang10:
Many tables are so large that need seperate partitions with "HOUR" time unit. But now dynamic partition doesn't support "HOUR" time unit and it was marked as "TODO". So I support the feature and it works.


morningman (approved):
LGTM

Issue #3992, opened 2020-07-01
[Feature]Support cancel load jobs in batch


xy720:
Related issue #2298
**Is your feature request related to a problem? Please describe.**
As I cancel load jobs in the same pattern, I only cancel them one by one, which is so tedious.

**Describe the solution you'd like**
Inspired by `show load where label like ""` command，I suggest we should support cancel load jobs in batch with following command:

```
cancel load where label like "load_label_pattern"
```

PR #4515, opened 2020-09-02
[Feature] Support cancel load jobs in batch

xy720:
## Proposed changes

#3992 

## Types of changes

## Checklist

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/load/Load.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/Load.java
@@ -1619,27 +1619,15 @@ public class Load {
             }
 
             // check state here
-            if (isAccurateMatch) {
-                // only the last one should be running
-                LoadJob job = matchLoadJobs.get(matchLoadJobs.size() - 1);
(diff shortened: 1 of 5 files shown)


fe/fe-core/src/main/java/org/apache/doris/load/Load.java
```diff
@@ -1570,6 +1581,105 @@ public boolean isLabelExist(String dbName, String label) throws DdlException {
...
+                }
+            }
+
+            if (matchLoadJobs.isEmpty()) {
+                throw new DdlException("Load job does not exist");
+            }
+
+            // check state here
+            if (isAccurateMatch) {
+                // only the last one should be running
```
morningman, line 1623: We should not rely on "the last one is running", better filter it by state.
So that we can unify the logic just same as batch cancel.

xy720: ok. I just unify the logic.

morningman (approved):
LGTM

PR #4517, opened 2020-09-02
optimize third party build script

### LINKED ISSUE #4516: 【Refactor】Optimize thirdparty buid script and upgrade brpc
here are a lot of lib64 and lib processing logic in the build script of thridparty, but this is still not enough in different linux distributions and needs to be unified into a directory. The latest version of BRPC has integrated support for ARM64, we should use the official support of brpc instead of our own patch, so we need to upgrade the version of brpc

yangzhg:
## Proposed changes

here are a lot of lib64 and lib processing logic in the build script of thridparty, but this is still not enough in different linux distributions and needs to be unified into a directory. The latest version of BRPC has integrated support for ARM64, we should use the official support of brpc instead of our own patch, so we need to upgrade the version of brpc

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4516), and have described the bug/feature there in detail


thirdparty/build-thirdparty.sh
```diff
@@ -68,7 +68,8 @@ else
 fi
 
 # prepare installed prefix
-mkdir -p ${TP_DIR}/installed
+mkdir -p ${TP_DIR}/installed/lib64
+ln -sf ${TP_DIR}/installed/lib64 ${TP_DIR}/installed/lib
```
chaoyli, line 72: You want to unified into lib?

yangzhg: yes

PR #4535, opened 2020-09-04
[Bug] Fix bug that table ids is not set right for hadoop load job

morningman:
## Proposed changes

We store all table ids involved in the Load Job in TransactionState.
However, for Hadoop Load job, table ids are set incorrectly. 

This caused the WAITING_TXN phase to not correctly wait for the completion
of the previous load transaction when doing the alter table,
which caused some data version loss problems.

## Types of changes


EmmyMiao87 (approved):
LGTM

morningman:
I merged it now, the release test needs it.

PR #4536, opened 2020-09-04
[Spark load][Bug] fix that cancelling a spark load in the `PENDING` phase will not succeed

### LINKED ISSUE #4518: [Spark Load][Bug] Cancelling a spark load in the pending phase will not succeed
**Describe the bug**
Cancelling a spark load in the pending phase will not succeed.

**To Reproduce**
1. Submit a spark load job
2. The job state is `PENDDING` at the beginning.
3. cancel load job.
4. Show load. And you can see the job state is `CANCELLED` with msg `user cancel`
5. Open the web interface of the yarn cluster with browser, you can see that the spark job has been submitted successfully and is running well.

**Expected behavior**
The spark job should not be submmited.

xy720:
## Proposed changes

**To Reproduce**
1. Submit a spark load job
2. The job state is `PENDDING` at the beginning.
3. cancel load job.
4. Show load. And you can see the job state is `CANCELLED` with msg `user cancel`
5. Open the web interface of the yarn cluster with browser, you can see that the spark job has been submitted successfully and is running well.

**Expected behavior**
The spark job should not be submmited.

**The reson of this Bug**
When a user wants to cancel a loadjob in the pending phase, it is possible that the `appId` in the SparkLoadJob is empty,  causing the yarn command to fail, so we need to get the `appId` through the handle.

If the `appId` still cannot be get from handle, which means the spark mission hasn't been submitted, the spark launcher process should be killed to prevent the spark mission from being submitted to yarn cluster.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4518 ), and have described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/load/loadv2/SparkEtlJobHandler.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/loadv2/SparkEtlJobHandler.java
@@ -264,8 +264,18 @@ public class SparkEtlJobHandler {
     }
 
     public void killEtlJob(SparkLoadAppHandle handle, String appId, long loadJobId, SparkResource resource) throws LoadException {
-        Preconditions.checkNotNull(appId);
         if (resource.isYarnMaster()) {
+            // When users kill the load job in PENDING phase, it is possible that the appId is empty. This is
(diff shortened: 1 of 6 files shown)


morningman (approved):
LGTM

PR #4505, opened 2020-09-01
Add log for trace broker 

ZhangYu0123:
## Proposed changes

Add tracing broker log. When fe get filestatus for distributing load task to broker, the broker maybe get empty files and not give correct error code.
Add this log to easy track which broker process filestatus operation and we can get the error log.

## Types of changes

What types of changes does your code introduce to Doris?
_Put an `x` in the boxes that apply_

- [x] Bugfix (non-breaking change which fixes an issue)
- [] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [] Documentation Update (if none of the other choices apply)
- [] Code refactor (Modify the code structure, format the code, etc...)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I have create an issue on (Fix #4506 ), and have described the bug/feature there in detail
- [x] Compiling and unit tests pass locally with my changes
- [] I have added tests that prove my fix is effective or that my feature works
- [] If this change need a document change, I have updated the document
- [] Any dependent changes have been merged

PR #4550, opened 2020-09-07
【Feature】Doris can  set default configuration of database data quota and replica quota #4540#

### LINKED ISSUE #4540: [New Feature]Set the default quota value of the database in the configuration file
Hi, All: 

In our system, we use software to automatically create the doris database. However, the default quota of the Doris database is 1 TB, That is too small. Therefore, after the program automatically creates the database, we need to manually set the quota to 10 TB.

Therefore, we hope that the default quota can be set in the configuration file. After the database is created, the quota is automatically set to the default value in the configuration file.

When Doris-FE startup, FE reads the quota value from the configuration file. When creating a database, set quota to the value in the configuration file.

I have implemented this feature. The database quota is read from the configuration file. When creating a database, set the quota value to the value read from the configuration file. I want to feed this feature back to Apache-Doris-Community.

zhaojintaozhao:
Doris can  set default configuration of database data quota and replica quota #4505#
## Proposed changes
When Doris create a  database, the database of data quota and replica quota is initialized by configuration(Config.default_db_data_quota_bytes and Config.default_db_replica_quota_bytes). In other words, I convert FeConstants.default_db_data_quota_bytes to Config.default_db_data_quota_bytes and convert FeConstants.default_db_replica_quota_size to Config.default_db_replica_quota_bytes. Config.default_db_data_quota_bytes and Config.default_db_replica_quota_bytes can be change in configuration file.

## Types of changes

- [-] New feature (non-breaking change which adds functionality)
- [-] Documentation Update (if none of the other choices apply)

## Checklist

- [-] I have create an issue on (Fix #4540 ), and have described the bug/feature there in detail
- [-] Compiling and unit tests pass locally with my changes
- [-] If this change need a document change, I have updated the document

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/Database.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/Database.java
@@ -19,8 +19,8 @@ package org.apache.doris.catalog;
 
 import org.apache.doris.catalog.Table.TableType;
 import org.apache.doris.cluster.ClusterNamespace;
+import org.apache.doris.common.Config;
 import org.apache.doris.common.DdlException;
-import org.apache.doris.common.FeConstants;
 import org.apache.doris.common.FeMetaVersion;
 import org.apache.doris.common.Pair;
 import org.apache.doris.common.UserException;
@@ -111,8 +111,8 @@ public class Database extends MetaObject implements Writable {
         this.rwLock = new ReentrantReadWriteLock(true);
         this.idToTable = new ConcurrentHashMap<>();
         this.nameToTable = new HashMap<>();
-        this.dataQuotaBytes = FeConstants.default_db_data_quota_bytes;
-        this.replicaQuotaSize = FeConstants.default_db_replica_quota_size;
+        this.dataQuotaBytes = Config.default_db_data_quota_bytes;
+        this.replicaQuotaSize = Config.default_db_replica_quota_bytes;
         this.dbState = DbState.NORMAL;
         this.attachDbName = "";
         this.clusterName = "";
@@ -483,7 +483,7 @@ public class Database extends MetaObject implements Writable {
         if (Catalog.getCurrentCatalogJournalVersion() >= FeMetaVersion.VERSION_81) {
             replicaQuotaSize = in.readLong();
         } else {
-            replicaQuotaSize = FeConstants.default_db_replica_quota_size;
+            replicaQuotaSize = Config.default_db_replica_quota_bytes;
         }
     }
 
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/Database.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/Database.java
@@ -21,6 +21,7 @@ import org.apache.doris.catalog.Table.TableType;
 import org.apache.doris.cluster.ClusterNamespace;
 import org.apache.doris.common.Config;
 import org.apache.doris.common.DdlException;
+import org.apache.doris.common.FeConstants;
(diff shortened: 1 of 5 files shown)


fe/fe-core/src/main/java/org/apache/doris/catalog/Database.java
```diff
@@ -483,7 +483,7 @@ public void readFields(DataInput in) throws IOException {
         if (Catalog.getCurrentCatalogJournalVersion() >= FeMetaVersion.VERSION_81) {
             replicaQuotaSize = in.readLong();
         } else {
-            replicaQuotaSize = FeConstants.default_db_replica_quota_size;
+            replicaQuotaSize = Config.default_db_replica_quota_bytes;
```
morningman, line 486: If change this to Config, different FE may has different configuration. So this will result in inconsistency of meta data.

This is just a default value filled in for compatibility with old data. I suggest to leave it as `FeConstants.default_db_replica_quota_size`

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -1250,4 +1250,16 @@
...
+     * Used to set default db data quota bytes.
+     */
+    @ConfField(mutable = true, masterOnly = false)
+    public static long default_db_data_quota_bytes = 1024 * 1024 * 1024 * 1024L; // 1TB
+
+    /**
+     * Used to set default db replica quota bytes.
+     */
+    @ConfField(mutable = true, masterOnly = false)
+    public static long default_db_replica_quota_bytes = 1024 * 1024 * 1024;
```
morningman, line 1264: The unit is wrong. This is not a "byte size", but a "number of" replicas.

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -1250,4 +1250,16 @@
      */
     @ConfField(mutable = true, masterOnly = true)
     public static boolean enable_batch_delete_by_default = false;
+
+    /**
+     * Used to set default db data quota bytes.
+     */
+    @ConfField(mutable = true, masterOnly = false)
+    public static long default_db_data_quota_bytes = 1024 * 1024 * 1024 * 1024L; // 1TB
```
morningman, line 1258: Both config should be `masterOnly = true`

morningman (changes requested):

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
zhaojintaozhao: Yes, that's right. I change "masterOnly" to true

fe/fe-core/src/main/java/org/apache/doris/catalog/Database.java
zhaojintaozhao: Hi morningman:
You are right, I restore the modification of default_db_replica_quota_size.

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
zhaojintaozhao: I restore the modification of default_db_replica_quota_size.

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
zhaojintaozhao: OK， I change the "masterOnly" property of default_db_data_quota_bytes to true

morningman (approved):
LGTM

PR #4553, opened 2020-09-07
[Feature] FE add default configuration of max_fiter_ratio #4541#

### LINKED ISSUE #4541: [New Feature]Set the default value of max_filter_ratio in the configuration file
Hi, All

The default value of Max_filter_ratio is 0. In this case, if the data is irregulary, we need to set max_Filter_Ratio in each load statement, that is troublesome.

Therefore, we put the default value of max_Filter_Ratio in the configuration file. The default value is no longer fixed at 0. Instead, you can set different values as required. In this way, max_Filter_Ratio is not required for most load statements in scenarios where data is irregulary. The default value is used only when max_Filter_Ratio is not set in the Load statement.

I have implemented the feature of setting the default value of max_Filter_Ratio in the configuration file. I hope I can contribute to the community.

zhaojintaozhao:
Fe add default configuration of database data quota and replica quota #4541#

Proposed changes

Add default_max_filter_ratio configuration in Config.java. It will be overridden by 'max_filter_ratio' from the load job properties.
 So, if a load job set 'max_filter_ratio' properties, the 'max_filter_ratio' will effective, otherwise the max_filter_ratio of the load job will use 'default_max_filter_ratio'. We don't set 'max_filter_ratio' properties in each load statement.

Types of changes

[] Bugfix (non-breaking change which fixes an issue)
[-] New feature (non-breaking change which adds functionality  #4541#)
[] Breaking change (fix or feature that would cause existing functionality to not work as expected)
[-] Documentation Update (if none of the other choices apply)
[] Code refactor (Modify the code structure, format the code, etc...)

Checklist

[-] I have create an issue on (Fix #4541 ), and have described the bug/feature there in detail
[-] Compiling and unit tests pass locally with my changes
[] I have added tests that prove my fix is effective or that my feature works
[-] If this change need a document change, I have updated the document
[] Any dependent changes have been merged

### DIFF
--- a/docs/en/administrator-guide/config/fe_config.md
+++ b/docs/en/administrator-guide/config/fe_config.md
@@ -692,6 +692,13 @@ If this parameter is set to true, Doris can support ODBC external table creation
 
 The function is still in the experimental stage, so the default value is false.
 
+
 ### `default_db_data_quota_bytes`
 
 Used to set default database data quota size, default is 1T.
+
+
+### 'default_max_filter_ratio'
+
+Used to set default max filter ratio of load Job. It will be overridden by 'max_filter_ratio' of the load job properties，default value is 0, value range 0-1.
+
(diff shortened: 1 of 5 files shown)


morningman (approved):
LGTM

PR #4593, opened 2020-09-12
[Bug] Fix bug of cumulative compaction and deletion of stale version

### LINKED ISSUE #4551: [Routine Load][BE]Too many VersionCount cause by wrong compaction failure time

morningman:
## Proposed changes

When selecting candicate rowsets to do the cumulative compaction,
some rowsets may not be selected because the protection time has not expired.
Therefore, we need to find the current longest continuous version path in the candicate rowsets.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4551), and have described the bug/feature there in detail

### DIFF
--- a/be/src/olap/compaction.cpp
+++ b/be/src/olap/compaction.cpp
@@ -181,8 +181,12 @@ OLAPStatus Compaction::gc_unused_rowsets() {
     return OLAP_SUCCESS;
 }
 
-// find the longest consecutive version path in "rowset", from begining.
-OLAPStatus Compaction::find_longest_consecutive_version(vector<RowsetSharedPtr>* rowsets) {
+// Find the longest consecutive version path in "rowset", from begining.
+// Two versions before and after the missing version will be saved in missing_version,
+// if missing_version is not null.
+OLAPStatus Compaction::find_longest_consecutive_version(
+        vector<RowsetSharedPtr>* rowsets,
+        vector<Version>* missing_version) {
     if (rowsets->empty()) {
         return OLAP_SUCCESS;
     }
@@ -191,11 +195,10 @@ OLAPStatus Compaction::find_longest_consecutive_version(vector<RowsetSharedPtr>*
     for (; i < rowsets->size(); ++i) {
         RowsetSharedPtr rowset = (*rowsets)[i];
         if (rowset->start_version() != prev_rowset->end_version() + 1) {
-            LOG(WARNING) << "There are missed versions among rowsets. "
-                << "prev_rowset verison=" << prev_rowset->start_version()
-                << "-" << prev_rowset->end_version()
-                << ", rowset version=" << rowset->start_version()
-                << "-" << rowset->end_version();
+            if (missing_version != nullptr) {
+                missing_version->push_back(prev_rowset->version());
+                missing_version->push_back(rowset->version());
+            }
             break;
         }
         prev_rowset = rowset;
--- a/be/src/olap/compaction.cpp
+++ b/be/src/olap/compaction.cpp
@@ -181,6 +181,27 @@ OLAPStatus Compaction::gc_unused_rowsets() {
     return OLAP_SUCCESS;
 }
 
+// find the longest consecutive version path in "rowset", from begining.
+OLAPStatus Compaction::find_longest_consecutive_version(vector<RowsetSharedPtr>* rowsets) {
+    RowsetSharedPtr prev_rowset = rowsets->front();
+    size_t i = 1;
+    for (; i < rowsets->size(); ++i) {
+        RowsetSharedPtr rowset = (*rowsets)[i];
+        if (rowset->start_version() != prev_rowset->end_version() + 1) {
+            LOG(WARNING) << "There are missed versions among rowsets. "
+                << "prev_rowset verison=" << prev_rowset->start_version()
+                << "-" << prev_rowset->end_version()
+                << ", rowset version=" << rowset->start_version()
+                << "-" << rowset->end_version();
+            break;
+        }
+        prev_rowset = rowset;
+    }
+
+    rowsets->resize(i);
+    return OLAP_SUCCESS;
+}
+
 OLAPStatus Compaction::check_version_continuity(const vector<RowsetSharedPtr>& rowsets) {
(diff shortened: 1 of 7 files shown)


ZhangYu0123:
LGTM

yangzhg (dismissed):
+1

be/src/olap/compaction.cpp
```diff
@@ -181,6 +181,30 @@ OLAPStatus Compaction::gc_unused_rowsets() {
...
+OLAPStatus Compaction::find_longest_consecutive_version(vector<RowsetSharedPtr>* rowsets) {
+    if (rowsets->empty()) {
+        return OLAP_SUCCESS;
+    }
+    RowsetSharedPtr prev_rowset = rowsets->front();
+    size_t i = 1;
+    for (; i < rowsets->size(); ++i) {
+        RowsetSharedPtr rowset = (*rowsets)[i];
+        if (rowset->start_version() != prev_rowset->end_version() + 1) {
+            LOG(WARNING) << "There are missed versions among rowsets. "
```
imay, line 194: Is it appropriate to print the warning log here?
I think this function only deals with specific logic, and the caller can decide whether to print the log. This allows this function to be used in more scenarios.

morningman: Done

be/src/olap/cumulative_compaction.cpp
```diff
@@ -83,7 +83,18 @@ OLAPStatus CumulativeCompaction::pick_rowsets_to_compact() {
         return OLAP_ERR_CUMULATIVE_NO_SUITABLE_VERSIONS;
     }
 
-    RETURN_NOT_OK(check_version_continuity(candidate_rowsets));
+    // candidate_rowsets may not be continuous. Because some rowset may not be selected
+    // because the protection time has not expired(config::cumulative_compaction_skip_window_seconds).
+    // So we need to choose the longest continuous path from it.
+    std::vector<Version> missing_versions;
+    RETURN_NOT_OK(find_longest_consecutive_version(&candidate_rowsets, &missing_versions));
+    if (missing_versions.empty()) {
```
imay, line 91:
```suggestion
    if (!missing_versions.empty()) {
```

yangzhg (approved):
+1

imay (approved):
LGTM

PR #4584, opened 2020-09-10
[UI Part 2] Add HTTP API documents of Frontend

morningman:
## Proposed changes

This PR add documents of all HTTP APIs of Frontends, list as bellow:

```
Origin:

bootstrap-action
cancel-load-action
check-decommission-action
check-storage-type-action
config-action
connection-action
get-ddl-stmt-action
get-load-info-action
get-load-state
get-log-file-action
get-small-file
ha-action
hardware-info-action
health-action
log-action
meta-action
meta-info-action
meta-replay-state-action
profile-action
query-detail-action
query-profile-action
row-count-action
session-action
set-config-action
show-data-action
show-proc-action
show-runtime-info-action
system-action
table-query-plan-action
table-row-count-action
table-schema-action

New:

logout-action
show-meta-info-action
statement-execution-action
upload-action
```

This Part 2 of feature proposed in #4308
The code implementation of the API described in the document will be submitted in Part 3

## Types of changes

- [x] Documentation Update (if none of the other choices apply)

Issue #4308, opened 2020-08-10
 [Proposal][UI][HttpServer] FE frontend and backend separation


hf200012:
## Status Quo and Motivation

The current Doris FE module has a simple http server and UI built in. The UI currently has relatively simple functions and is only used for the display of some information.

At present, Doris does not use a frontend and backend separation framework, resulting in a large amount of html and js code embedded in the FE module. When we want to modify some frontend elements, we have to update and upgrade the main code, which is expensive to maintain.

At the same time, the current http server is written by Netty, which has relatively simple functions. If more advanced functions are needed, a large amount of additional code needs to be written. There are already many very mature web frameworks in the industry, and we should use them to simplify our development.

## Implementation

`Spring` is a very mature open source framework in the industry. It is very convenient to build a complete Web application through SpringMVC and Spring Boot. We can also easily use the spring framework to complete a series of functions such as authentication, session, and cookie. At the same time, RESTful standard API can be developed quickly.

The UI code is completely stripped from the FE module. The FE module only provides standard RESTful API. This allows more flexible use of the frontend framework to develop Doris's UI interface. Here I choose to use `Vue` as the frontend framework, and I will transplant all the functions of the current WebUI to the new framework. This UI may not be the final form of the UI, but I think it is a good start, and more extensions can be made under this framework in the future.

## How to submit my code

There are many code changes this time. In order to ensure that the code can be merged without affecting the trunk code, I am going to submit the code according to the following steps.

1. Update the maven dependencies of Frontends

2. Submit the new version of the API document for the interface Review.

3. Submit the UI part, which is completely independent and will not affect the existing functions of the main code. Before the new version of RESTful API is enabled, this part of the code will not take effect.

4. Submit the new RESTFul API code. This part only adds files and does not modify existing functions. And this part of the function will not be enabled, so you can ensure that the trunk code can still run normally.

5. Submit a new version of HttpServer, based on Spring. This part of the code will replace the old http server implementation before, and connect the new version of the RESTful API and the new version of the UI. After the code is merged, all new functions are enabled.

morningman:
Nice Job!
I followed this feature for a while before, and it looks good to me.

hffariel:
Nice Job!

Just have had a quick tour of the code, and some tips here could be of help:

- Precompiled css frameworks like `sass` `stylus` `less` might be better for a theme customization?
- Replace some `px` with `rem` might be better for the future responsive development?
[...]

PR #4596, opened 2020-09-14
[UI Part 3] New implemented HTTP RESTful API of Frontend

hf200012:
Use spring mvc rest to replace the original netty http rest
Created a new package `org/apache/doris/httpv2`,
and the origin implementations under `org/apache/doris/http` remain unchanged.

This part of the code will not be used at present, so it will not affect existing functions.

API document can be found in #4584 

Proposal #4308


morningman (approved):
LGTM

PR #4597, opened 2020-09-14
[Bug] Tablet and Disk report thread not work

morningman:
## Proposed changes

The tablet and disk information reporting threads need to report to the FE periodically.
At the same time these two reporting threads will also be triggered by certain events.

The modification in PR #4440 caused these two threads to be triggered only by events,
and could not report regularly.

## Types of changes

### DIFF
--- a/be/src/agent/task_worker_pool.h
+++ b/be/src/agent/task_worker_pool.h
@@ -43,6 +43,8 @@ class ThreadPool;
 
 class TaskWorkerPool {
 public:
+    // You need to modify the content in TYPE_STRING at the same time,
+    // and pay attention to ensure that the order is consistent.
     enum TaskWorkerType {
         CREATE_TABLE,
         DROP_TABLE,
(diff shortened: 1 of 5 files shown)


be/src/agent/task_worker_pool.h
```diff
@@ -43,6 +43,8 @@ class ThreadPool;
 
 class TaskWorkerPool {
 public:
+    // You need to modify the content in TYPE_STRING at the same time,
```
acelyc111, line 46: Maybe it's not easy to keep order consistency some days later, how about use a function to get name of type?

morningman: OK

be/src/olap/storage_engine.cpp
```diff
@@ -975,11 +975,17 @@ void StorageEngine::deregister_report_listener(TaskWorkerPool* listener) {
...
+/// if submit_task is true, it will notify the listeners by submmiting a task.
+/// otherwise, it just notify the thread
+void StorageEngine::notify_listeners(bool submit_task) {
     std::lock_guard<std::mutex> l(_report_mtx);
     for (auto& listener : _report_listeners) {
-        TAgentTaskRequest task;
-        listener->submit_task(task);
+        if (submit_task) {
+            TAgentTaskRequest task;
+            listener->submit_task(task);
```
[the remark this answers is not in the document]
morningman: OK

acelyc111 (approved):

EmmyMiao87 (approved):
LGTM

PR #4601, opened 2020-09-14
[Feature] Add a http interface to acquire the tablets distribution between different disks of a particular BE

weizuo93:
## Proposed changes

For the task of rebalancing tablet among different disks on the same BE, I think it might be an effective strategy to ensure all tablets under the same partition evenly distribute on the different disks. Thus, it is necessary to obtain the distribution of tablets under the same partition between different disks on a BE. 

This patch  add a new http interface for FE to acquire the distribution of tablets under each partition between different disks of a particular BE.

## Types of changes

## Checklist

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/http/rest/TabletDiskDistributionAction.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/http/rest/TabletDiskDistributionAction.java
@@ -80,6 +80,7 @@ public class TabletDiskDistributionAction extends RestBaseAction {
         Map<Long, List<Pair<Long, Integer>>> partitionToTablet = new HashMap<Long, List<Pair<Long, Integer>>>();
         Map<Pair<Long, Integer>, Long> tabletToDataSize = new HashMap<Pair<Long, Integer>, Long>();
         Map<Pair<Long, Integer>, Long> tabletToPathHash = new HashMap<Pair<Long, Integer>, Long>();
+        Map<Long, String> partitionToTable = new HashMap<Long, String>();
         int tablet_total = 0;
         List<Long> dbIds = Catalog.getCurrentCatalog().getDbIds();
         for (long dbId : dbIds) {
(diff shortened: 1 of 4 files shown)


fe/fe-core/src/main/java/org/apache/doris/http/rest/TabletDiskDistributionAction.java
```diff
@@ -0,0 +1,192 @@
...
+        jsonObject_result.put("code", 0);
+        jsonObject_result.put("data", jsonObject_backend);
+        jsonObject_result.put("count", tablet_total);
+        String result =jsonObject_result.toString();
+
+        response.setContentType("application/json");
+        response.getContent().append(result);
+        sendResult(request, response);
+    }
+}
```
acelyc111, line 192: Add a newline

fe/fe-core/src/main/java/org/apache/doris/http/rest/TabletDiskDistributionAction.java
```diff
@@ -0,0 +1,192 @@
...
+                                        tablet_total++;
+                                    }
+                                }
+                            }
+                        }
+                        partitionToTablet.put(partitionId, tabletSchemaHash);
+                    }
+                }
+            } catch (Exception e) {
+                LOG.info(e);
```
acelyc111, line 133: When exception throwed, an error response should be returned.

fe/fe-core/src/main/java/org/apache/doris/http/rest/TabletDiskDistributionAction.java
```diff
@@ -0,0 +1,192 @@
...
+            } finally {
+                db.readUnlock();
+            }
+        }
+
+        JSONObject jsonObject_backend = new JSONObject();
```
acelyc111, line 139: Naming `jsonObjectBackend ` or `json_object_backend `

acelyc111, line 139: How about split this function to two parts? One for collecting info, the other for construct JSON?

weizuo93: OK

PR #4605, opened 2020-09-15
[Bug] Fix bug that forwarding show tablet stmt return error

morningman:
## Proposed changes

When forwarding `show tablet` stmt to Master FE. If the result contains null, FE will throw exception.
Fix it by using `\N` instead of "null" in result, to make thrift rpc happy.

## Types of changes


EmmyMiao87 (approved):
LGTM

PR #4632, opened 2020-09-18
[Feature] Add CPU and Heap profile in BE webserver

morningman:
## Proposed changes

Add CPU Profile and Heap Profile in BE webserver.
So that we can more easily diagnose system performance bottlenecks through perf tools.

## Types of changes

## Screenshots

image

image


yangzhg (approved):
+1

PR #4671, opened 2020-09-25
[Bug]External engines(e.g. ES) don't have segments, ignore those tables

caoyang10:
It will crash when there are External engine tables in doris after executing "python show_segment_status.py"
The main reason is external engine tables don't have any index and partitions. So it should be ignored.


morningman (approved):
LGTM

PR #4678, opened 2020-09-27
[Bug] Fix hard cardinality check which makes queries fail

### LINKED ISSUE #4581: [Bug] SelectNode.cardinality < 0 cause query exception "Unexpected exception: null"

ccoffline:
## Proposed changes

#4581
fix hard `Preconditions.checkState()` by setting `cardinality` to `-1` when computation is invalid.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4581), and have described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/SelectNode.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/SelectNode.java
@@ -69,7 +69,7 @@ public class SelectNode extends PlanNode {
         } else {
             this.cardinality = Math.round(cardinality * selectivity);
         }
-        LOG.info("stats Select: cardinality={}", this.cardinality);
+        LOG.debug("stats Select: cardinality={}", this.cardinality);
     }
 
     @Override
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/SelectNode.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/SelectNode.java
@@ -24,7 +24,6 @@ import org.apache.doris.analysis.Expr;
 import org.apache.doris.thrift.TExplainLevel;
 import org.apache.doris.thrift.TPlanNode;
 import org.apache.doris.thrift.TPlanNodeType;
-import com.google.common.base.Preconditions;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.LogManager;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/org/apache/doris/planner/SelectNode.java
```diff
@@ -63,13 +62,14 @@ public void init(Analyzer analyzer) throws UserException {
...
+        double selectivity = computeSelectivity();
+        if (cardinality < 0 || selectivity < 0) {
+            this.cardinality = -1;
         } else {
-            cardinality = Math.round(((double) getChild(0).cardinality) * computeSelectivity());
-            Preconditions.checkState(cardinality >= 0);
+            this.cardinality = Math.round(cardinality * selectivity);
         }
-        LOG.info("stats Select: cardinality=" + Long.toString(cardinality));
+        LOG.info("stats Select: cardinality={}", this.cardinality);
```
morningman, line 72:
```suggestion
        LOG.debug("stats Select: cardinality={}", this.cardinality);
```

morningman (approved):
LGTM

PR #4684, opened 2020-09-28
[UI Part 5] Enable HTTP Server 2 by FE config

### LINKED ISSUE #4308: [Proposal][UI][HttpServer] FE frontend and backend separation

morningman:
## Proposed changes

This is the last PR of proposal #4308

1. Add a new FE config `enable_http_server_v2` to enable new HTTP Server implementation. The default value is false.
2. Add a new FE config `http_api_extra_base_path` so that we can set base path for Frontend UI.
3. Refactor the new HTTP API response body. The return http status code is always 200, and using internal code in response body to indicate the certain error.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4308), and have described the bug/feature there in detail


yangzhg (dismissed):
+1

yangzhg (approved):
+1

PR #4719, opened 2020-10-12
[Bug] Add regular column when materialized slot is empty in tuple

### LINKED ISSUE #4716: Failed to query constant which is in inline view
**Describe the bug**

Situation: Select a constant column from inline view which includes at least one regular column.
Query failed!

```
MySQL [test]> select a from (select 'c1' a , k1 from test) a;
ERROR 1064 (HY000): errCode = 2, detailMessage = failed to build storage scanner, no materialized slot!
```

**To Reproduce**
1. create table 
```
CREATE TABLE `test` (
  `k1` int(11) NULL COMMENT "",
  `k2` bigint(20) NULL COMMENT "",
  `k3` decimal(9, 0) NULL COMMENT ""
) ENGINE=OLAP
DUPLICATE KEY(`k1`, `k2`)
COMMENT "OLAP"
DISTRIBUTED BY HASH(`k1`) BUCKETS 10
PROPERTIES (
"replication_num" = "1",
"in_memory" = "false",
"storage_format" = "DEFAULT"
);
```

2. insert values

```
insert into test values(2,2,2);
```

3. query

```
MySQL [test]> select a from (select 'c1' a , k1 from test) a;
ERROR 1064 (HY000): errCode = 2, detailMessage = failed to build storage scanner, no materialized slot!
```

**Expected behavior**

```
+------+
| a    |
+------+
| c1   |
+------+
```

If there are two lines in table, the result will be:

```
+------+
| a    |
+------+
| c1   |
| c1   |
+------+
```

EmmyMiao87:
If the materialized slot is empty in inline view, the scanner will throw exception 'no materialized slot!'.
When the outer query does not need any real columns of the inner query at all, such as the outer query only contains constant calculations,
all slots in the inner query will not be materialized.
For example:
```
select c1 from (select 'xx' c1, k1 from test) t1;
```

But this does not mean that the inner query (select 'xx' c1, k1 from test) does not need to be executed.
Because the number of rows in the outer result needs to be determined by the number of rows in the inner query.
In this case, the inner query scan node needs to add a column to ensure the correct result.
At the same time it can avoid the scan node reporting errors.

Fixed #4716

Change-Id: I86bb52efab65b658ea78094037543cb23dbff1d4

## Proposed changes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
@@ -1902,7 +1902,7 @@ public class SingleNodePlanner {
      * @param analyzer
      */
     private void materializeBaseTableRefResultForCrossJoinOrCountStar(BaseTableRef tblRef, Analyzer analyzer) {
-        if (tblRef.getDesc().getSlots().isEmpty()) {
+        if (tblRef.getDesc().getMaterializedSlots().isEmpty()) {
             Column minimuColumn = null;
             for (Column col : tblRef.getTable().getBaseSchema()) {
                 if (minimuColumn == null || col.getDataType().getSlotSize() < minimuColumn
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
@@ -1887,7 +1887,7 @@ public class SingleNodePlanner {
      */
     private void materializeTableResultForCrossJoinOrCountStar(TableRef tblRef, Analyzer analyzer) {
         if (tblRef instanceof BaseTableRef) {
-            materializeBaseTableRefResultForCrossJoinOrCountStar((BaseTableRef) tblRef, analyzer);
+            materializeSlotForEmptyMaterializedTableRef((BaseTableRef) tblRef, analyzer);
         } else if (tblRef instanceof InlineViewRef) {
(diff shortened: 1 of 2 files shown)


kangkaisen:
+1. But please add a UT for this bug.

EmmyMiao87:
> +1. But please add a UT for this bug.

This single test is a bit difficult to add, because the slot information is not displayed in the explain.

Or should I just add a single test for this function `materializeBaseTableRefResultForCrossJoinOrCountStar` ?

kangkaisen:
> > +1. But please add a UT for this bug.
> 
> This single test is a bit difficult to add, because the slot information is not displayed in the explain.
> 
> Or should I just add a single test for this function `materializeBaseTableRefResultForCrossJoinOrCountStar` ?

OK

kangkaisen (dismissed):
+1

fe/fe-core/src/main/java/org/apache/doris/planner/SingleNodePlanner.java
```diff
@@ -1902,7 +1902,7 @@ private void materializeTableResultForCrossJoinOrCountStar(TableRef tblRef, Anal
      * @param analyzer
      */
     private void materializeBaseTableRefResultForCrossJoinOrCountStar(BaseTableRef tblRef, Analyzer analyzer) {
```
morningman, line 1904: I think this method name should be changed to a more common name?
like: `materializeSlotForEmptyMaterializedTableRef()`?
And add examples in comment to explain why we call it.

morningman (approved):
LGTM

PR #4735, opened 2020-10-13
[Repair] modified to prohibit partition column to be defined to "allow null" when create table

xinyiZzz:
## Proposed changes

Currently, partition column can be defined as "allow null" when create table. 
But errors may occur when null as the value of partition columns. 
Therefore, modified to prohibit partition column to be defined to "allow null"

## Types of changes

## Checklist

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/ConnectProcessor.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/ConnectProcessor.java
@@ -475,6 +475,9 @@ public class ConnectProcessor {
         // no matter the master execute success or fail, the master must transfer the result to follower
         // and tell the follower the current journalID.
         TMasterOpResult result = new TMasterOpResult();
+        if (ctx.queryId() != null) {
+            result.setQueryId(ctx.queryId);
+        }
         result.setMaxJournalId(Catalog.getCurrentCatalog().getMaxJournalId().longValue());
         result.setPacket(getResultPacket());
         if (executor != null && executor.getProxyResultSet() != null) {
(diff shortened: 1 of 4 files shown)


imay:
Hi, @xinyiZzz 
Is there an issue describing the specific problem?
The previous version of Doris had problems when handling Null partition columns, and the current code should have been fixed.
If you want to prohibit the Partition column from being null, you need to discuss it in the Issue or Mail list.

imay:
> > Hi, @xinyiZzz
> > Is there an issue describing the specific problem?
> > The previous version of Doris had problems when handling Null partition columns, and the current code should have been fixed.
> > If you want to prohibit the Partition column from being null, you need to discuss it in the Issue or Mail list.
> 
> The null value to be a partition value is very confusing.
> Currently, the null value is stored in the partition with MIN_VALUE lower bound. But if that partition is dropped, the null value
> can not be loaded. So it is not a good way to solve the problem.
> 
> Some other database use a special partition to save null value. But I think it is very hard to maintain and hard to understand.
> Is it necessary to allow null value to be a partition value?

Yes，I agree with you about Null partition is confusing.

However some tables currently contain Null values. If the null partition column is not supported in the future, how to deal with the current existing data, I think there should be a more complete plan.

How about creating another issue to discuss about it?

morningman:
@xinyiZzz , hi , i think we can discuss it in dev@ mail list.
Could you please send a email to start with it？

xinyiZzz:
> @xinyiZzz , hi , i think we can discuss it in dev@ mail list.
> Could you please send a email to start with it？

Sorry for the late reply. I have sent an email to Mail list to discuss this, thanks @imay @morningman for your reply and attention

PR #4737, opened 2020-10-14
[Bug] fix illegal defer in Tablet::rowset_with_max_version()

### LINKED ISSUE #4736: [Bug] Tablet::rowset_with_max_version() may defer the end iter in release mode
**Describe the bug**
https://github.com/apache/incubator-doris/blob/a605b3160f27b2afac75ae9aa887f770a5ce7618/be/src/olap/tablet.cpp#L323-L333

We use `max_version` to find `iter`. And if `iter==end()`, it will return `end()->second`, cuz `DCHECK` won't effect in release mode.
It'll cause a coredump, because the `->second` is a `RowsetSharedPtr`.

**To Reproduce**
Setting an invalid `max_version` will reproduce it.
Stack:
```
#0  std::__shared_ptr<doris::Rowset, (__gnu_cxx::_Lock_policy)2>::__shared_ptr (this=this@entry=0x7ffd2498a5e0) at /usr/include/c++/7.3.0/bits/shared_ptr_base.h:1121
#1  0x0000000000c9f2e1 in shared_ptr (this=0x7ffd2498a5e0) at /usr/include/c++/7.3.0/bits/shared_ptr.h:119
#2  doris::Tablet::rowset_with_max_version (this=<optimized out>) at /root/doris-env/doris/be/src/olap/tablet.cpp:333
```
**Expected behavior**
if `iter == _rs_version_map.end()`, return nullptr.

vagetablechicken:
## Proposed changes

We use max_version to find iter. And if iter==end(), it will return end()->second, cuz DCHECK won't effect in release mode.
It'll cause a coredump, because the ->second is a RowsetSharedPtr.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4736 ), and have described the bug/feature there in detail

### DIFF
--- a/be/src/olap/tablet.cpp
+++ b/be/src/olap/tablet.cpp
@@ -327,8 +327,10 @@ const RowsetSharedPtr Tablet::rowset_with_max_version() const {
     }
 
     auto iter = _rs_version_map.find(max_version);
-    DCHECK(_rs_version_map.find(max_version) != _rs_version_map.end())
-        << "invalid version:" << max_version;
+    if (iter == _rs_version_map.end()) {
+        DCHECK(false) << "invalid version:" << max_version;
+        return nullptr;
+    }
     return iter->second;
 }
 


imay (approved):
LGTM

PR #4750, opened 2020-10-15
[Bug][Compaction] Fix bug that compaction may be blocked

morningman:
## Proposed changes

the logic of compaction producer thread may failed to produce compaction task due to
invalid order of modifying task map.

## Types of changes


yangzhg (approved):
+1

Issue #4755, opened 2020-10-17
Check date type to avoid scan all partitions


xinghuayu007:
**Describe the bug**
`select day from test where day='2020-10-32'`, table 'test' is parititioned by day. In this case, '2020-10-32' will be taken as CastExpr not LiteralExpr, and condition "day='2020-10-32'"  will not be recognized as partitionfilter. This case will scan all partitions. To avoid scall all partitions, it is better to filter invalid date value.

Doris supports datetime with micro second, but save datetime with second in BE. Doris FE use joda time library to parse a datetime of string type. If a datetime with micro second, it will transfer it into CastExpr, which will be recognized into invalid time. Therefore in the begining, a datetime with micro second will be parsed into datetime with second. It make no difference. For expample k1 <= "2020-10-10 10:10:10.00" will be transfered into k1 <= "2020-10-10 10:10:10";

**To Reproduce**
1. create table:
 CREATE TABLE `test` (
  `date` datetime NULL COMMENT "日期",
  `day` date NULL COMMENT "",
  `app_id` varchar(1000) NULL COMMENT "",
 `device_id` int(11) 0 comment ""
) ENGINE=OLAP
DUPLICATE KEY(`date`, `day`, `app_id`)
COMMENT "OLAP"
PARTITION BY RANGE(`day`)()
DISTRIBUTED BY HASH(`device_id`) BUCKETS 64
PROPERTIES (
"replication_num" = "1",
"dynamic_partition.enable" = "true",
"dynamic_partition.time_unit" = "DAY",
"dynamic_partition.time_zone" = "Asia/Shanghai",
"dynamic_partition.start" = "-30",
"dynamic_partition.end" = "3",
"dynamic_partition.prefix" = "p",
"dynamic_partition.replication_num" = "1",
"dynamic_partition.buckets" = "32",
"in_memory" = "false",
"storage_format" = "V2"
); 
2. insert some data into each partition
3. execute a query which day is invalid, it will scan all partitions
explain select day from test where '2020-10-32' = day;
![2020-10-17 20-44-01屏幕截图](https://user-images.githubusercontent.com/12771191/96337427-b9f88c00-10b9-11eb-864d-56bdfe80e295.png)
explain select day from test where 'hello' = day;
![2020-10-17 20-44-26屏幕截图](https://user-images.githubusercontent.com/12771191/96337439-cbda2f00-10b9-11eb-9a24-be0c308b9d85.png)
4. if day is valid, it will scan some partitions
[...]

PR #4756, opened 2020-10-17
[FEATURE]Check date type to avoid scan all partitions

xinghuayu007:
## Proposed changes

select day from test where day='2020-10-32', table 'test' is parititioned by day. In this case, '2020-10-32' will be taken as CastExpr not LiteralExpr, and condition "day='2020-10-32'" will not be recognized as partitionfilter. This case will scan all partitions. To avoid scall all partitions, it is better to filter invalid date value.

issue: #4755
## Types of changes

## Checklist

## Further comments


fe/fe-core/src/main/java/org/apache/doris/rewrite/ExprRewriter.java
```diff
@@ -55,6 +58,12 @@ public Expr rewrite(Expr expr, Analyzer analyzer) throws AnalysisException {
                 rewrittenExpr = applyRuleRepeatedly(rewrittenExpr, rule, analyzer);
             }
         } while (oldNumChanges != numChanges_);
+        if (expr instanceof BinaryPredicate) {
```
xy720, line 61: Why not check the type in analysis state?

fe/fe-core/src/main/java/org/apache/doris/rewrite/ExprRewriter.java
```diff
@@ -55,6 +58,12 @@ public Expr rewrite(Expr expr, Analyzer analyzer) throws AnalysisException {
...
+            Expr valueExpr = ((BinaryPredicate) expr).getBinding();
+            if(valueExpr != null && valueExpr.getType() == Type.DATETIME && valueExpr instanceof CastExpr) {
+                throw new AnalysisException("invalid date type :" + valueExpr.toSql());
```
xy720, line 64: I think throwing a exception is not the best solution of this problem, may be rewrite the expr `day = '2020-10-32'` to something like `1 = 0` in where statement is better. We should discuss it first.

xinghuayu007: > I think throwing a exception is not the best solution of this problem, may be rewrite the binary expr `day = '2020-10-32'` to something like contant expr BoolLiteral[false] in where statement is better. We should discuss it first.

ok，I will modify it;

fe/fe-core/src/main/java/org/apache/doris/rewrite/BinaryPredicatesDateRule.java
```diff
@@ -0,0 +1,46 @@
...
+ * date = "2020-10-32" => FALSE
+ */
+public class BinaryPredicatesDateRule implements ExprRewriteRule {
+    public static ExprRewriteRule INSTANCE = new BinaryPredicatesDateRule();
+
+    @Override
+    public Expr apply(Expr expr, Analyzer analyzer) throws AnalysisException {
+        if (!(expr instanceof BinaryPredicate)) return expr;
+        Expr valueExpr = expr.getChild(1);
+        if(valueExpr.getType() == Type.DATETIME && valueExpr instanceof CastExpr) {
```
xy720, line 41:
```suggestion
        if(valueExpr.getType() == Type.DATETIME && valueExpr instanceof CastExpr && valueExpr.getChild(0).getType() == Type.VARCHAR) {
```
There is something wrong with this If statement.
For example, `where day = 20201030` will rewrite to `where false`, which is not right.
You should only support covert StringLiteral now:
`where day = '20201032'` -> `where false`

xy720 (commented):
LGTM

fe/fe-core/src/main/java/org/apache/doris/rewrite/BinaryPredicatesDateRule.java
xinghuayu007: The rule of FoldConstantsRule will convert `where day = 20201030` into `where day='2020-10-30 00:00:00'`, whic is DateLiteral type.

morningman (approved):
LGTM

PR #4492, opened 2020-09-01
[New Feathure] Make Unique key semi join syntax error more clear

xinghuayu007:
## Proposed changes

**"SELECT * FROM t1 LEFT SEMI JOIN t0 ON t1.c0 = t0.c0"**,  and **"t0"** is an unique key. This select clause will appear error: "Illegal column/field reference 't0.__DORIS_DELETE_SIGN__' of semi-/anti-join", but user don't know what is the meaning of this error, how to solve this problem. In this patch, it will tell user "__DORIS_DELETE_SIGN__ is a hidden column to mark whether a row deleted when unique key existed, '__DORIS_DELETE_SIGN__ = 0' will appear in where clause, that will cause semi join syntax error, please use '[not] in/existed' as replacement."

## Types of changes

What types of changes does your code introduce to Doris?
_Put an `x` in the boxes that apply_

- [] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [] Documentation Update (if none of the other choices apply)
- [] Code refactor (Modify the code structure, format the code, etc...)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [] I have create an issue on (Fix #ISSUE), and have described the bug/feature there in detail
- [x] Compiling and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] If this change need a document change, I have updated the document
- [x] Any dependent changes have been merged

## Further comments

If this is a relatively large or complex change, kick off the discussion at dev@doris.apache.org by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #4770, opened 2020-10-20
Fix  delete_sign predicate assing to join node

yangzhg:
## Proposed changes

Fix delete_sign predicate maybe assign to join node incorrectly

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4492), and have described the bug/feature there in detail

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/OlapScanNode.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/OlapScanNode.java
@@ -23,6 +23,7 @@ import org.apache.doris.analysis.BinaryPredicate;
 import org.apache.doris.analysis.CastExpr;
 import org.apache.doris.analysis.Expr;
 import org.apache.doris.analysis.InPredicate;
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/org/apache/doris/planner/OlapScanNode.java
```diff
@@ -278,6 +280,15 @@ protected String debugString() {
     @Override
     public void init(Analyzer analyzer) throws UserException {
         super.init(analyzer);
+        if (!Util.showHiddenColumns() && olapTable.hasDeleteSign()) {
```
EmmyMiao87, line 283: Maybe you can unify this logic into one function

yangzhg: olaptable may also has other hidden columns. It is not appropriate to add such a function to each hidden column, and the writing method is not particularly complicated.

morningman (approved):
LGTM

PR #4771, opened 2020-10-20
[LoadBalance] make BeLoadRebalancer extends from base class Rebalancer

### LINKED ISSUE #4763: [Proposal] refactor the LoadBalancer to support more rebalance strategies
We want to add more strategies for load balance, actually for rebalance. And the new strategies may want to execute a concrete move of a replica from one server to another, a move can be described as {tablet, from_be, dest_be}. Thus, the rebalance class should support to delete the specified replica(in from_be).
So we need to define a new base class for rebalance, the origin rebalance strategy could work on it as well.

The new base class, named `Rebalancer`.
class methods:

- selectAlternativeTablets: use a strategy to determine which tablet to rebalance.

- createBalanceTask: try to create a clone task for rebalance.

-getSrcReplicaId: needed by the new priority "replica is src replica of rebalance" in handleRedundantReplica.The origin rebalance doesn't need to override getSrcReplicaId. The move-based strategies should override it. 

- updateLoadStatistic: the rebalancer may have cache, so it should be stateful. Add a method to update the load statistic.

The origin `LoadBalancer`, which will named as `BeLoadRebalancer`,  just need to extend from `Rebalancer` without any logic change.

vagetablechicken:
## Proposed changes
Create a base class `Rebalancer`, support to delete the specified replica(actually add a new priority in handling redundant replica).
The origin LoadBalancer, which will named as BeLoadRebalancer, just need to extend from Rebalancer without any logic change.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4763), and have described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
@@ -65,6 +65,8 @@ public abstract class Rebalancer {
         return alternativeTablets;
     }
 
+    // The return TabletSchedCtx should have the tablet id at least. {srcReplica, destBe} can be complete here or
+    // later(when createBalanceTask called).
     protected abstract List<TabletSchedCtx> selectAlternativeTabletsForCluster(
             String clusterName, ClusterLoadStatistic clusterStat, TStorageMedium medium);
 
@@ -76,10 +78,10 @@ public abstract class Rebalancer {
     }
 
     // Before createCloneReplicaAndTask, we need to complete the TabletSchedCtx.
-    // 1. If you generate {tabletId, srcReplica, destReplica} in selectAlternativeTablets(), it may be invalid at
+    // 1. If you generate {tabletId, srcReplica, destBe} in selectAlternativeTablets(), it may be invalid at
     // this point(it may have a long interval between selectAlternativeTablets & createBalanceTask).
     // You should check the moves' validation.
-    // 2. If you want to generate {srcReplica, destReplica} here, just do it.
+    // 2. If you want to generate {srcReplica, destBe} here, just do it.
     // 3. You should check the path slots of src & dest.
     protected abstract void completePlan(TabletSchedCtx tabletCtx, Map<Long, PathSlot> backendsWorkingSlots)
             throws SchedException;
--- a/fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
@@ -46,14 +46,15 @@ import java.util.Map;
 public abstract class Rebalancer {
     private static final Logger LOG = LogManager.getLogger(Rebalancer.class);
 
+    // When Rebalancer init, the statisticMap is empty so that it's no need to be an arg.
+    // Use updateLoadStatistic() to load stats only.
     protected Map<String, ClusterLoadStatistic> statisticMap;
     protected TabletInvertedIndex invertedIndex;
     protected SystemInfoService infoService;
 
-    public Rebalancer(Map<String, ClusterLoadStatistic> statisticMap) {
-        this.statisticMap = statisticMap;
-        this.infoService = Catalog.getCurrentSystemInfo();
-        this.invertedIndex = Catalog.getCurrentInvertedIndex();
+    public Rebalancer(SystemInfoService infoService, TabletInvertedIndex invertedIndex) {
+        this.infoService = infoService;
+        this.invertedIndex = invertedIndex;
     }
 
     public List<TabletSchedCtx> selectAlternativeTablets() {
--- a/fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
@@ -35,11 +35,11 @@ import java.util.Map;
  * 1. selectAlternativeTablets: selecting alternative tablets by one rebalance strategy,
  * and return them to tablet scheduler(maybe contains the concrete moves, or maybe not).
  * 2. createBalanceTask: given a tablet, try to create a clone task for this tablet.
- * 3. getCachedSrcBackendId: if the rebalance strategy wants to delete the replica of a specified be,
+ * 3. getSrcBackendId: if the rebalance strategy wants to delete the replica of a specified be,
  * override this func.
  * NOTICE:
  * It may have a long interval between selectAlternativeTablets() & createBalanceTask(). So the concrete moves may be
- * invalid when we createBalanceTask(), you should check the moves' validation.
+ * invalid when we generating a clone task in createBalanceTask(), you should check the moves' validation.
  */
 public abstract class Rebalancer {
     // When Rebalancer init, the statisticMap is usually empty. So it's no need to be an arg.
@@ -70,7 +70,7 @@ public abstract class Rebalancer {
     public abstract void createBalanceTask(TabletSchedCtx tabletCtx, Map<Long, PathSlot> backendsWorkingSlots,
                                            AgentBatchTask batchTask) throws SchedException;
 
-    public Long getCachedSrcBackendId(Long tabletId) {
+    public Long getSrcBackendId(Long tabletId) {
         return -1L;
     }
 
--- a/fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
@@ -28,6 +28,7 @@ import com.google.common.collect.Lists;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
```diff
@@ -0,0 +1,80 @@
...
+        return alternativeTablets;
+    }
+
+    protected abstract List<TabletSchedCtx> selectAlternativeTabletsForCluster(
+            String clusterName, ClusterLoadStatistic clusterStat, TStorageMedium medium);
+
+    public abstract void createBalanceTask(TabletSchedCtx tabletCtx, Map<Long, PathSlot> backendsWorkingSlots,
+                                           AgentBatchTask batchTask) throws SchedException;
+
+    public Long getCachedSrcBackendId(Long tabletId) {
```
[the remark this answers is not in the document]
vagetablechicken: I prefer that cache should be managed by `Rebalancer`, no need to make `getCachedSrcBackendId` to delete. You can do whatever you want, e.g. only clear cache when stats update.

About cancellations(scheduler canceled, task running cancelled, etc.):
Before these cancellations, adding tablet to the scheduler may get failed. And after the clone task completed, the rest of a rebalance move may get failed too. So there're so many places that need to delete cache. I think we shouldn't 'delete the cache asap'. Maybe clearing cache when stats update is better.

kangkaisen, line 73: For a  general balance interface.  method name `getCachedSrcBackendId` is not very clear.   `delete the replica of a specified be` in  `getCachedSrcBackendId` is also strange.

fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
```diff
@@ -0,0 +1,80 @@
...
+ * 1. selectAlternativeTablets: selecting alternative tablets by one rebalance strategy,
+ * and return them to tablet scheduler(maybe contains the concrete moves, or maybe not).
+ * 2. createBalanceTask: given a tablet, try to create a clone task for this tablet.
+ * 3. getCachedSrcBackendId: if the rebalance strategy wants to delete the replica of a specified be,
+ * override this func.
+ * NOTICE:
+ * It may have a long interval between selectAlternativeTablets() & createBalanceTask(). So the concrete moves may be
+ * invalid when we createBalanceTask(), you should check the moves' validation.
+ */
+public abstract class Rebalancer {
```
kangkaisen, line 44: For a balance, the core strategy is how to choose which tablet, choose src backend, choose dest backend. But I don't see the interface about how to choose dest backend.

fe/fe-core/src/main/java/org/apache/doris/clone/TabletScheduler.java
```diff
@@ -826,12 +831,25 @@ private boolean deleteReplicaNotInCluster(TabletSchedCtx tabletCtx, boolean forc
         return false;
     }
 
+    private boolean deleteSrcReplicaOfLB(TabletSchedCtx tabletCtx, boolean force) throws SchedException {
```
kangkaisen, line 834: What's the meaning of `LB` ？

fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
vagetablechicken: It does seem odd, I'll reconsider the naming

fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
vagetablechicken: Choosing dest be can be done before generating a clone task in `createBalanceTask`. The base class doesn't limit it.
For example:
The origin rebalance algo: `selectAlternativeTablets` just select a tablet, choose dest be in `createBalanceTask`
The algo which has a concrete move: `selectAlternativeTablets` will select a tablet and src be, dest be. In `createBalanceTask`, we should check the validation.

fe/fe-core/src/main/java/org/apache/doris/clone/TabletScheduler.java
vagetablechicken: LoadBalance，I think the naming should be changed too.

fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
kangkaisen: > Choosing dest be can be done before generating a clone task in createBalanceTask. The base class doesn't limit it.

how to choose the dest be is core for balance strategy, if the base class doesn't limit it, what's the meaning of balance base class.

fe/fe-core/src/main/java/org/apache/doris/clone/TabletScheduler.java
```diff
@@ -826,12 +832,25 @@ private boolean deleteReplicaNotInCluster(TabletSchedCtx tabletCtx, boolean forc
...
     }
 
+    private boolean deleteReplicaChosenByRebalancer(TabletSchedCtx tabletCtx, boolean force) throws SchedException {
+        Long id = rebalancer.getSrcReplicaId(tabletCtx.getTabletId());
+        if (id == -1L) {
+            return false;
+        }
+        Replica chosenReplica = tabletCtx.getTablet().getReplicaById(id);
+        if (chosenReplica != null) {
+            deleteReplicaInternal(tabletCtx, chosenReplica, "src replica of rebalance", force);
```
kangkaisen, line 842: I don't understand why we need to delete the src replica of rebalance?

kangkaisen:
In Apache Druid. https://github.com/apache/druid/blob/master/server/src/main/java/org/apache/druid/server/coordinator/BalancerStrategy.java

`pickSegmentToMove` will choose which tablet replica to balance, naturally, the BE for the tablet replica is the src backend.

`findNewSegmentHomeBalancer` will choose the dest backend.

I think the Apache Druid `BalancerStrategy` interface logic and naming is clear.

fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
kangkaisen:
```
 * 3. getSrcReplicaId: if the rebalance strategy wants to delete the src replica,
 * override this func.
```
Do a delete operation in a get method, I still think which is confusing.

vagetablechicken:
> How to choose the dest be is core for balance strategy, if the base class doesn't limit it, what's the meaning of balance base class.

It's based on the specific strategies. Some strategies can delay choosing the dest be until create a clone task, but some 
strategies can't . So we don't overemphasize this.
The meaning of balance base class is to support delete the src replica, which the current class can't do.

> In Apache Druid. https://github.com/apache/druid/blob/master/server/src/main/java/org/apache/druid/server/coordinator/BalancerStrategy.java
> 
> `pickSegmentToMove` will choose which tablet replica to balance, naturally, the BE for the tablet replica is the src backend.
> 
> `findNewSegmentHomeBalancer` will choose the dest backend.
> 
> I think the Apache Druid `BalancerStrategy` interface logic and naming is clear.

We want to import the kudu's rebalance strategy, it'll calculate a concrete move{partition id, src be, dest be}, so it's no need to separate into `choose one tablet and src be` & `find dest be`.

In Doris, rebalance has 2 steps:
1. select tablet, and clone a replica on dest be, so the tablet is redundant
2. delete redundant replica

**The key point** is,  **step 2** will generate another TabletSchedCtx , we can't know which replica is the replica we want to delete. So we should try to get it from rebalancer.

kangkaisen:
@vagetablechicken Hi, I see what's your mean.

I agree with your interface logic. But for the method naming, I think we would better rename it.

For `selectAlternativeTabletsForCluster` method,  the nameing or comment should let other people know we 
select {tablet id, src be, dest be} at one time in this method.

For `getSrcReplicaId`, rename it to `getDeleteReplicaId` ?

fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
morningman: In origin LoadBalancer class, both "Choose Source" and "Choose Destination" are done in `createBalanceTask`. 

@kangkaisen  may think that as a base class, these two interfaces need to be explicitly provided. But I think this is just a design pattern consideration, for the actual code implementation, there is little difference between the two. One is to implement the interfaces separately, and the other is to implement them in one interface.
The former has a clearer interface meaning, while the latter currently has relatively fewer code changes.

morningman:
I have a question about the problem described in #4763. If we implement a new LoadBalancer that allow users to specify a tablet to migrate from node A to node B. But after the migration is completed, how to ensure that this tablet will not be migrated back again by the default balance strategy?

kangkaisen:
> I have a question about the problem described in #4763. If we implement a new LoadBalancer that allow users to specify a tablet to migrate from node A to node B. But after the migration is completed, how to ensure that this tablet will not be migrated back again by the default balance strategy?

I think in one cluster, we could only enable one balance strategy?

vagetablechicken:
> > I have a question about the problem described in #4763. If we implement a new LoadBalancer that allow users to specify a tablet to migrate from node A to node B. But after the migration is completed, how to ensure that this tablet will not be migrated back again by the default balance strategy?
> 
> I think in one cluster, we could only enable one balance strategy?

Yes, only one.

fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
```diff
@@ -0,0 +1,96 @@
...
+        batchTask.addTask(tabletCtx.createCloneReplicaAndTask());
+    }
+
+    // Before createCloneReplicaAndTask, we need to complete the TabletSchedCtx.
+    // 1. If you generate {tabletId, srcReplica, destBe} in selectAlternativeTablets(), it may be invalid at
+    // this point(it may have a long interval between selectAlternativeTablets & createBalanceTask).
+    // You should check the moves' validation.
+    // 2. If you want to generate {srcReplica, destBe} here, just do it.
+    // 3. You should check the path slots of src & dest.
+    protected abstract void completePlan(TabletSchedCtx tabletCtx, Map<Long, PathSlot> backendsWorkingSlots)
```
kangkaisen, line 86: In the whole FE balance module, we don't use `Plan`. So I think we could rename it to `completeSchedCtx`?

vagetablechicken: 👌

kangkaisen (approved):
+1

morningman (approved):
LGTM

PR #4778, opened 2020-10-22
Error happened when open a view in DBeaver

### LINKED ISSUE #4777: Error happened when open a view in DBeaver
**Describe the bug**
When you are operating doris using DBeaver, an error happens when you are opening a view.
You can't see the description information for a view in DBeaver.
This operation need to call SELECT CHECK_OPTION,DEFINER,IS_UPDATABLE FROM information_schema.VIEWS
information_schema.VIEWS doesn't exist, so it will get an error.

**To Reproduce**
1. Download DBeaver on version 6.0.0
2. Create a view
3. Open the view in DBeaver
4. You can see the error and can't get the description of the view.

**Expected behavior**
I want to see the description of the view.

**Additional context**
a new table names views is needed in database information_schema, which has the same schema as views in mysql.

pengxiangyu:
## Proposed changes

Fix bug #4777：Error happened when open a view in DBeaver

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4777), and have described the bug/feature there in detail
- [] Compiling and unit tests pass locally with my changes

## Further comments

Fix bug #4777

### DIFF
--- a/be/src/exec/schema_scanner/schema_views_scanner.cpp
+++ b/be/src/exec/schema_scanner/schema_views_scanner.cpp
@@ -29,24 +29,13 @@ SchemaScanner::ColumnDesc SchemaViewsScanner::_s_tbls_columns[] = {
     { "TABLE_CATALOG", TYPE_VARCHAR, sizeof(StringValue), true},
     { "TABLE_SCHEMA", TYPE_VARCHAR, sizeof(StringValue), false},
     { "TABLE_NAME",   TYPE_VARCHAR, sizeof(StringValue), false},
-    { "TABLE_TYPE",   TYPE_VARCHAR, sizeof(StringValue), false},
-    { "ENGINE",       TYPE_VARCHAR, sizeof(StringValue), true},
-    { "VERSION",      TYPE_BIGINT, sizeof(int64_t), true},
-    { "ROW_FORMAT",   TYPE_VARCHAR, sizeof(StringValue), true},
-    { "TABLE_ROWS",   TYPE_BIGINT, sizeof(int64_t), true},
-    { "AVG_ROW_LENGTH", TYPE_BIGINT, sizeof(int64_t), true},
-    { "DATA_LENGTH",   TYPE_BIGINT, sizeof(int64_t), true},
-    { "MAX_DATA_LENGTH", TYPE_BIGINT, sizeof(int64_t), true},
-    { "INDEX_LENGTH", TYPE_BIGINT, sizeof(int64_t), true},
-    { "DATA_FREE", TYPE_BIGINT, sizeof(int64_t), true},
-    { "AUTO_INCREMENT", TYPE_BIGINT, sizeof(int64_t), true},
-    { "CREATE_TIME", TYPE_DATETIME, sizeof(DateTimeValue), true},
-    { "UPDATE_TIME", TYPE_DATETIME, sizeof(DateTimeValue), true},
-    { "CHECK_TIME", TYPE_DATETIME, sizeof(DateTimeValue), true},
-    { "TABLE_COLLATION", TYPE_VARCHAR, sizeof(StringValue), true},
-    { "CHECKSUM", TYPE_BIGINT, sizeof(int64_t), true},
-    { "CREATE_OPTIONS", TYPE_VARCHAR, sizeof(StringValue), true},
-    { "TABLE_COMMENT", TYPE_VARCHAR, sizeof(StringValue), false},
+    { "VIEW_DEFINITION",   TYPE_VARCHAR, sizeof(StringValue), true},
+    { "CHECK_OPTION",       TYPE_VARCHAR, sizeof(StringValue), true},
+    { "IS_UPDATABLE",       TYPE_VARCHAR, sizeof(StringValue), true},
+    { "DEFINER",       TYPE_VARCHAR, sizeof(StringValue), true},
+    { "SECURITY_TYPE",       TYPE_VARCHAR, sizeof(StringValue), true},
+    { "CHARACTER_SET_CLIENT",       TYPE_VARCHAR, sizeof(StringValue), true},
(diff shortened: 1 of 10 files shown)


imay:
It is best to talk about the cause of the problem and the idea of PR change.
This will allow others to better review the code

fe/fe-core/src/main/java/org/apache/doris/catalog/Table.java
```diff
@@ -95,6 +95,8 @@
     // table(view)'s comment
     protected String comment = "";
 
+    protected String ddlSql = "";
```
morningman, line 98: Add comment for this new field

be/src/exec/schema_scanner/schema_views_scanner.cpp
```diff
@@ -0,0 +1,227 @@
...
+    {
+        void *slot = tuple->get_slot(_tuple_desc->slots()[6]->tuple_offset());
+        StringValue* str_slot = reinterpret_cast<StringValue*>(slot);
+        const std::string definer = "root@%";
+        str_slot->len = definer.length();
+        str_slot->ptr = (char *)pool->allocate(str_slot->len);
+        if (NULL == str_slot->ptr) {
+            return Status::InternalError("Allocate memcpy failed.");
+        }
+        memcpy(str_slot->ptr, definer.c_str(), str_slot->len);
```
yangzhg, line 154: length always 6？

fe/fe-core/src/main/java/org/apache/doris/catalog/SchemaTable.java
```diff
@@ -281,6 +281,23 @@ public static Builder builder() {
...
+                                TableType.SCHEMA,
+                                builder()
+                                        .column("TABLE_CATALOG", ScalarType.createVarchar(512))
+                                        .column("TABLE_SCHEMA", ScalarType.createVarchar(64))
+                                        .column("TABLE_NAME", ScalarType.createVarchar(64))
+                                        .column("VIEW_DEFINITION", ScalarType.createVarchar(8096))
+                                        .column("CHECK_OPTION", ScalarType.createVarchar(8))
+                                        .column("IS_UPDATABLE", ScalarType.createVarchar(3))
+                                        .column("DEFINER", ScalarType.createVarchar(77))
+                                        .column("SECURITY_TYPE", ScalarType.createVarchar(7))
```
yangzhg, line 297: why this column length is 77, you should add a comment, this column in mysql is 288

fe/fe-core/src/main/java/org/apache/doris/catalog/SchemaTable.java
```diff
@@ -281,6 +281,23 @@ public static Builder builder() {
...
+                                        .column("CHARACTER_SET_CLIENT", ScalarType.createVarchar(32))
+                                        .column("COLLATION_CONNECTION", ScalarType.createVarchar(32))
```
yangzhg, line 299: why 32

fe/fe-core/src/main/java/org/apache/doris/service/FrontendServiceImpl.java
```diff
@@ -248,7 +248,19 @@ public TListTableStatusResult listTableStatus(TGetTablesParams params) throws TE
...
+                    tables = db.getTables();
+                } else {
+                    switch (params.getType()) {
+                        case "VIEW":
+                            tables = db.getViews();
+                            break;
+                        default:
+                            tables = db.getTables();
+                    }
+                }
```
yangzhg, line 262:
```suggestion
               List<Table> tables = db.getTables();
                if (params.isSetType() && "VIEW".equals(params.getType())) {
                    tables = db. getViews();
                }
```

be/src/exec/schema_scanner/schema_views_scanner.cpp
pengxiangyu: this is column for table views in mysql, palo has no this information, we can only set a default one

fe/fe-core/src/main/java/org/apache/doris/catalog/SchemaTable.java
pengxiangyu: Maybe you get another version of mysql. the sql I got is like this.

fe/fe-core/src/main/java/org/apache/doris/catalog/SchemaTable.java
pengxiangyu: the sql I got is like this.

fe/fe-core/src/main/java/org/apache/doris/service/FrontendServiceImpl.java
pengxiangyu: If type is view, this will call getTables() which is not required. When the count of tables in this db is too many, this will take more time.

yangzhg (dismissed):
+1

be/src/exec/schema_scanner/schema_views_scanner.cpp
```diff
@@ -0,0 +1,231 @@
...
+            return Status::InternalError("Allocate memcpy failed.");
+        }
+        memcpy(str_slot->ptr, check_option.c_str(), str_slot->len);
+    }
+    // is_updatable
+    {
+        void *slot = tuple->get_slot(_tuple_desc->slots()[5]->tuple_offset());
+        StringValue* str_slot = reinterpret_cast<StringValue*>(slot);
+        // This is from views in mysql
+        const std::string is_updatable = "YES";
```
morningman, line 138: Better set it to "No". Because we do not support DELETE or UPDATE on view.

pengxiangyu: DONE

morningman (approved):
LGTM

PR #4786, opened 2020-10-25
[Bug] Fix some date functions to make their result same as MySQL

### LINKED ISSUE #4785: [Bug] Function "dayofweek"'s result is not same as MySQL's
**Describe the bug**
Doris:
```
mysql> select dayofweek("10:00:00");
+-----------------------+
| dayofweek('10:00:00') |
+-----------------------+
|                     2 |
+-----------------------+
1 row in set (0.06 sec)
```

MySQL:
```
mysql> select dayofweek("10:00:00");
+-----------------------+
| dayofweek("10:00:00") |
+-----------------------+
|                  NULL |
+-----------------------+
1 row in set, 1 warning (0.05 sec)
```

**Why**
This is because "10:00:00" is converted to "2020-00-00", both in Doris and MySQL.
but `dayofweek("2020-00-00")` results in `NULL` in MySQL.

morningman:
## Proposed changes

dayofweek, dayofmonth, dayofyear, weekofyear, timediff

Also fix ut compilation problem

## Types of changes

## Checklist

- [] I have create an issue on (Fix #4785), and have described the bug/feature there in detail

### DIFF
--- a/be/src/exprs/timestamp_functions.cpp
+++ b/be/src/exprs/timestamp_functions.cpp
@@ -120,6 +120,7 @@ IntVal TimestampFunctions::month(
     const DateTimeValue& ts_value = DateTimeValue::from_datetime_val(ts_val);
     return IntVal(ts_value.month());
 }
+
 IntVal TimestampFunctions::day_of_week(
         FunctionContext* context, const DateTimeVal& ts_val) {
     if (ts_val.is_null) {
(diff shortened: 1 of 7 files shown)


be/src/exprs/timestamp_functions.cpp
```diff
@@ -126,7 +126,10 @@ IntVal TimestampFunctions::day_of_week(
         return IntVal::null();
     }
     const DateTimeValue& ts_value = DateTimeValue::from_datetime_val(ts_val);
-    return IntVal((ts_value.weekday() + 1 ) % 7 + 1);
+    if (ts_value.is_valid_date()) {
+        return IntVal((ts_value.weekday() + 1 ) % 7 + 1);
+    }
+    return IntVal::null();
```
kangkaisen, line 132: Should add UT to cover this case.

morningman: done

kangkaisen (approved):
+1

PR #4804, opened 2020-10-27
[TabletScheduler] Fix some bug where decommission operations cannot be completed

### LINKED ISSUE #4802: [Decommission] Decommission operation can not be completed
**Describe the bug**
When trying to decommission a batch of Backends at same time. The decommission operation
is running forever and all tablet schedule task is pending with error:

`unable to choose dest replica`

morningman:
## Proposed changes

1.
When we decommission some BEs with SSD disks,
if there are no SSD disks on the remaining BEs, it will be impossible to select a suitable destination path.
In this case, we need to ignore the storage medium property and try to select the destination path again.
Set `isSupplement` to true will ignore the storage medium property.

2.
When the BE nodes where all replicas of a tablet are located are decommission,
and this task is a VERSION_INCOMPLETE task.
This will lead to failure to select a suitable dest replica.
At this time, we should try to convert this task to a REPLICA_MISSING task, and schedule it again.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4802 ), and have described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/clone/TabletScheduler.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/clone/TabletScheduler.java
@@ -632,7 +632,25 @@ public class TabletScheduler extends MasterDaemon {
             throw new SchedException(Status.UNRECOVERABLE, "cluster does not exist");
         }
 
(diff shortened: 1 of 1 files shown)


EmmyMiao87 (approved):
LGTM

PR #4810, opened 2020-10-28
[Broker Load] Ignore empty file when file format is parquet or orc.

### LINKED ISSUE #4812: [Load] Error when reading empty Parquet or Orc files
**Describe the bug**
Doris can not read empty Parquet or Orc files when doing load. So these empty file should be skipped.

Error:
`type:LOAD_RUN_FAIL; msg:ParseError : Invalid column selected _col0`

morningman:
## Proposed changes

We can not read empty parquet or orc format file. So we should skip them
when doing broker load.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4812), and have described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/load/loadv2/BrokerLoadPendingTask.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/loadv2/BrokerLoadPendingTask.java
@@ -27,10 +27,10 @@ import org.apache.doris.load.BrokerFileGroupAggInfo.FileGroupAggKey;
 import org.apache.doris.load.FailMsg;
 import org.apache.doris.thrift.TBrokerFileStatus;
 
+import com.google.common.collect.Lists;
+
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-
-import com.google.common.collect.Lists;
 
 import java.util.List;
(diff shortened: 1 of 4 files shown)


fe/fe-core/src/main/java/org/apache/doris/load/loadv2/BrokerLoadPendingTask.java
```diff
@@ -77,18 +77,32 @@ private void getAllFileStatus() throws UserException {
...
+                        if (LOG.isDebugEnabled()) {
+                            LOG.debug(new LogBuilder(LogKey.LOAD_JOB, callback.getCallbackId())
+                                    .add("file_status", fstatus).build());
+                        }
                     }
                 }
+                fileStatusList.add(filteredFileStatuses);
                 tableTotalFileSize += groupFileSize;
-                tableTotalFileNum += fileStatuses.size();
+                tableTotalFileNum += fileNum;
```
kangkaisen, line 103: `filteredFileStatuses.size` seems equal `fileNum`, we could remove fileNum.

morningman: Removed

kangkaisen (approved):
+1

Issue #4833, opened 2020-11-02
[Bug] Unexpected exception: org.apache.doris.analysis.CastExpr cannot be cast to org.apache.doris.analysis.LiteralExpr


ccoffline:
**Describe the bug**
```
ERROR 1064 (HY000): errCode = 2, detailMessage = Unexpected exception: org.apache.doris.analysis.CastExpr cannot be cast to org.apache.doris.analysis.LiteralExpr
```

**To Reproduce**
```sql
CREATE DATABASE IF NOT EXISTS `test`; USE `test`;
CREATE TABLE `test_t` (
  `dt` bigint(20) NULL,
  `id` bigint(20) NULL,
  `name` varchar(100) NULL
) ENGINE=OLAP
DUPLICATE KEY(`dt`, `id`, `name`)
PARTITION BY RANGE(`dt`) (
PARTITION p20201001 VALUES [("19700101"), ("20201001")),
PARTITION p20201002 VALUES [("20201001"), ("20201002")),
PARTITION p20201003 VALUES [("20201002"), ("20201003")),
PARTITION p20201004 VALUES [("20201003"), ("20201004")),
PARTITION p20201005 VALUES [("20201004"), ("20201005")),
PARTITION p20201006 VALUES [("20201005"), ("20201006")),
PARTITION p20201007 VALUES [("20201006"), ("20201007")),
PARTITION p20201008 VALUES [("20201007"), ("20201008")),
PARTITION p20201009 VALUES [("20201008"), ("20201009")),
PARTITION p20201010 VALUES [("20201009"), ("20201010")),
PARTITION p20201011 VALUES [("20201010"), ("20201011")),
PARTITION p20201012 VALUES [("20201011"), ("20201012")),
PARTITION p20201013 VALUES [("20201012"), ("20201013")),
PARTITION p20201014 VALUES [("20201013"), ("20201014")),
PARTITION p20201015 VALUES [("20201014"), ("20201015")),
PARTITION p20201016 VALUES [("20201015"), ("20201016")),
PARTITION p20201017 VALUES [("20201016"), ("20201017")),
PARTITION p20201018 VALUES [("20201017"), ("20201018")),
PARTITION p20201019 VALUES [("20201018"), ("20201019")),
PARTITION p20201020 VALUES [("20201019"), ("20201020")),
PARTITION p20201021 VALUES [("20201020"), ("20201021")),
PARTITION p20201022 VALUES [("20201021"), ("20201022")),
[...]

PR #4836, opened 2020-11-03
[Bug] Fix ClassCastException (#4833)

ccoffline:
## Proposed changes

fix to #4833 

## Types of changes

## Checklist

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/PartitionKey.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/PartitionKey.java
@@ -137,11 +137,11 @@ public class PartitionKey implements Comparable<PartitionKey>, Writable {
             try {
                 LiteralExpr newKey = key1;
                 if (key1.getType() != destType) {
-                    newKey = (LiteralExpr) key1.castTo(destType);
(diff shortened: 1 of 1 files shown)


morningman:
I think if the predicate like `substr(`dt`, 7, 8) in ('09')`, we should not do the partition prunner.
So here:
https://github.com/apache/incubator-doris/blob/f40868a4805f0ba503bcca9a9c04f704b61c121b/fe/fe-core/src/main/java/org/apache/doris/planner/SingleNodePlanner.java#L958-L972

We should check if `inPredicate.getChild(0)` is not a slotRef, do not create partition filter for it.
So it won't goes to the `compareLiteral()` logic.

ccoffline:
@morningman I think cast `CastExpr` to `LiteralExpr` will definitely cause an exception.
I just don't know whether `getResultValue()` will solve that though I didn't see anything wrong in the test.
https://github.com/apache/incubator-doris/blob/b1c1ffda4a0d5ff2f0be4f9d2f25997a31537ea8/fe/fe-core/src/main/java/org/apache/doris/catalog/PartitionKey.java#L135-L150
I'm not fully understand the logic in `partition pruner`. Maybe it should be in another PR=.=

PR #4840, opened 2020-11-03
[Bug] Fix bug that failed to create view with complex select stmt

### LINKED ISSUE #4839: [Bug] Create view failed
**Describe the bug**
Create view failed with error: "failed to init view stmt"

**To Reproduce**
create a view with some complex select stmt such as union and join and group by.

**Why**
This is because the `toSql()` method of SlotRef will return like "<slot 20> col as alias", if that SlotRef is analyzed.

morningman:
## Proposed changes

Fix bug that failed to create view with complex select stmt

## Types of changes

## Checklist

- [] I have create an issue on (Fix #4839), and have described the bug/feature there in detail


EmmyMiao87 (approved):
LGTM

PR #4842, opened 2020-11-03
[Bug][SQL] Fix bug that query failed when SQL contains Union and Colocate join

### LINKED ISSUE #4841: [Bug][SQL] query failed with Union and Colocate Join
**Describe the bug**
Query failed with error like:
`failed to get tablet. tablet_id=26846, with schema_hash=398972982, reason=tablet does not exist`

SQL looks like:

`select a join b union select c join d`;

And both a/b and c/d are colocate join.

morningman:
## Proposed changes

SQL like:
`select a join b union select c join d`;

if both a/b and c/d are colocate join, the query may failed with error like:

`failed to get tablet. tablet_id=26846, with schema_hash=398972982, reason=tablet does not exist`

This is because when assigning exec host for fragment, the planner failed to distinguish which scan nodes
belong to the certain fragment. So it may choose the wrong host for scan range.

## Types of changes

## Checklist

- [] I have create an issue on (Fix #4841), and have described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/Coordinator.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/Coordinator.java
@@ -81,17 +81,17 @@ import org.apache.doris.thrift.TStatusCode;
 import org.apache.doris.thrift.TTabletCommitInfo;
 import org.apache.doris.thrift.TUniqueId;
 
+import org.apache.commons.collections.map.HashedMap;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.thrift.TException;
+
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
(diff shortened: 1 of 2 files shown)


EmmyMiao87 (approved):
LGTM

fe/fe-core/src/main/java/org/apache/doris/qe/Coordinator.java
```diff
@@ -1140,17 +1141,26 @@ private long getScanRangeLength(final TScanRange scanRange) {
 
     private void computeColocateJoinInstanceParam(PlanFragmentId fragmentId, int parallelExecInstanceNum, FragmentExecParams params) {
         Map<Integer, TNetworkAddress> bucketSeqToAddress = fragmentIdToSeqToAddressMap.get(fragmentId);
```
HappenLee, line 1143: Maybe `bucketSeqToScanRang` should be a map like `fragmentIdToSeqToAddressMap` to solve this problem is a
better understanding  way？

morningman: I will leave it when refactoring them.

PR #4869, opened 2020-11-09
[Bug] Fix some bugs of load job scheduler

### LINKED ISSUE #4868: [Bug][Load] Load task submit logic has some errors.
**Describe the bug**
1.
The pending load task thread pool and loading load task thread pool should be separated.
Because when doing the  pending load task in thread pool, the pending task will try to create a new loading task
and add it to the thread pool. If the thread pool is full, it will be blocked and wait the thread pool to be available.
But itself also occupy the thread. It is just like a dead lock.

2. 
the load task thread pool's queue size should be as long as config of `desired_max_waiting_jobs`.

3. 
When submitting a load task to thread pool, the operation should be outside the database lock, because it may take
a while to wait the thread pool to be available.

morningman:
## Proposed changes

1. The fix load meta bug logic should be removed since 0.12.
2. The load task thread pool's waiting queue should be as long as desired pending jobs num.
3. Submit the load task outside database lock to prevent holding lock for long time.
4. Use 2 thread pools for loading and pending load tasks, separately. To avoid dead lock submission.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4868), and have described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
@@ -387,7 +387,9 @@ public class Catalog {
 
     private TabletChecker tabletChecker;
 
-    private MasterTaskExecutor loadTaskScheduler;
+    // Thread pools for pending and loading task, separately
+    private MasterTaskExecutor pendingLoadTaskScheduler;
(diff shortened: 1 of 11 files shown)


EmmyMiao87:
Waiting for checking Unit test

EmmyMiao87 (approved):
LGTM

PR #4872, opened 2020-11-09
[Bug] Make 'LastStartTime' in backends list as the actual BE start time

acelyc111:
## Proposed changes

We use 'LastStartTime' in backends list to check whether there is an unexpected
restart of BE, but it will be changed as BE's first heartbeat time after FE
restarted, it would be better to set it to BE's actual start time.

## Types of changes

## Checklist

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/system/HeartbeatMgr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/system/HeartbeatMgr.java
@@ -242,9 +242,12 @@ public class HeartbeatMgr extends MasterDaemon {
                     if (tBackendInfo.isSetVersion()) {
                         version = tBackendInfo.getVersion();
                     }
-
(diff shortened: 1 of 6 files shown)


fe/fe-core/src/main/java/org/apache/doris/system/HeartbeatMgr.java
```diff
@@ -242,9 +242,12 @@ public HeartbeatResponse call() {
                     if (tBackendInfo.isSetVersion()) {
                         version = tBackendInfo.getVersion();
                     }
-
+                    long beStartTime = System.currentTimeMillis();
```
caiconghui, line 245:
```suggestion
                    long beStartTime = tBackendInfo.isSetBeStartTime() ? tBackendInfo.getBeStartTime() : System.currentTimeMillis();
```

caiconghui (approved):
LGTM

PR #4877, opened 2020-11-11
Revert "[FEATURE]Check date type to avoid scan all partitions"

xinghuayu007:
Reverts apache/incubator-doris#4756


morningman:
This commit has some problem when handling predicate like:
`k1 = "2020-10-10 10:00:00.000"`

This is a valid predicate, and FE Datetime can not support milli or micro seconds, so it will treat it as invalid date time value.

morningman (approved):
LGTM

PR #4825, opened 2020-10-30
[Feature] Running Profile OLAP_SCAN_NODE layering and enhance readability

xinyiZzz:
## Proposed changes

mainly includes:
- `OLAP_SCAN_NODE` profile layering: `OLAP_SCAN_NODE`,`OlapScanner`, and `SegmentIterator`.
- Delete meaningless statistical values. mainly in scan_node.cpp.
- Increase `RowsConditionsFiltered` statistical, split from `RowsDelFiltered`, the meaning is the number of rows filtered by various column indexes, only in segment V2.
- Modify the document based on the above, and enhance readability.

## Types of changes

- [x] Documentation Update (if none of the other choices apply)
- [x] Code refactor (Modify the code structure, format the code, etc...)

## Checklist

- [x] I have create an issue on (Fix #4821 ), and have described the bug/feature there in detail
- [x] If this change need a document change, I have updated the document

PR #4881, opened 2020-11-12
[Bug] Fix bug introduced by split RowsDelFiltered profile

xinyiZzz:
## Proposed changes

bug introduced from pr #4825, will cause `schema_change` to report an error:
```
schema_change.cpp:1271] fail to check row num! source_rows=1, merged_rows=0, filtered_rows=0, new_index_rows=0
schema_change.cpp:1921] failed to process the version. version=2-2
schema_change.cpp:1615] failed to alter tablet. base_tablet=44643.1383650721.b140317f6662c1e0-65bcbc87db8d22bc, drop new_tablet=45680.1530531459.474e41f3dd538fb6-9284085daac24f83
```

## Types of changes

### DIFF
--- a/be/src/olap/rowset/beta_rowset_reader.h
+++ b/be/src/olap/rowset/beta_rowset_reader.h
@@ -48,7 +48,8 @@ public:
 
     RowsetSharedPtr rowset() override { return std::dynamic_pointer_cast<Rowset>(_rowset); }
 
-    int64_t filtered_rows() override { return _stats->rows_del_filtered; }
+    // Return the total number of filtered rows, will be used for validation of schema change
+    int64_t filtered_rows() override { return _stats->rows_del_filtered + _stats->rows_conditions_filtered; }
 
 private:
     BetaRowsetSharedPtr _rowset;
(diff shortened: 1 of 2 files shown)


be/src/olap/rowset/beta_rowset_reader.h
```diff
@@ -48,7 +48,7 @@ class BetaRowsetReader : public RowsetReader {
 
     RowsetSharedPtr rowset() override { return std::dynamic_pointer_cast<Rowset>(_rowset); }
 
-    int64_t filtered_rows() override { return _stats->rows_del_filtered; }
+    int64_t filtered_rows() override { return _stats->rows_del_filtered + _stats->rows_conditions_filtered; }
```
morningman, line 51: Add a comment for this method to notify that this method will be used for validation of schema change task, should be modified carefully.

xinyiZzz: Related comment have been added, I will be familiar with the development process again.

morningman (approved):
LGTM

PR #4884, opened 2020-11-12
(#4883) Java Version BitmapValue deserialized failed

### LINKED ISSUE #4883: [BUG][BITMAP]Java Version BitmapValue deserialized failed when only has 32-bit bitmap
**Describe the bug**
1. add elements to ```org.apache.doris.load.loadv2.dpp.BitmapValue``` which 32-bit bitmap is enough to store
2. Serialize ```BitmapValue```, and then deserialize it will be failed.
This is because the logic of serialization and deserialization is not consistent when ```BitmapValue``` only has 32-bit bitmap

wangbo:
Proposed changes

## Checklist

- [x] I have create an issue on (Fix #4883), and have described the bug/feature there in detail


morningman (approved):
LGTM

PR #4889, opened 2020-11-12
[BUG] modify isDecommissioned be capacity calculate rule

stalary:
## Proposed changes

I use containerized deployment of BE nodes, both using the same distributed disk. When doing data migration, the current logic will lead to errors. For example, my distributed disk has 10t and has been used by other services for 9T, at this time, it is assumed that all the 9T data is used by BE nodes

## Types of changes

https://github.com/apache/incubator-doris/issues/4888


morningman (approved):
LGTM

PR #4925, opened 2020-11-19
[Refactor] Refactor DeleteHandler and Cond module

acelyc111:
## Proposed changes
This patch mainly do the following refactors:
- Use int64_t instead of int32_t for 'version' in DeleteHandler
- Move some comments from .cpp to .h file, add some new comments in .h files, and also remove some meaningless comments
- Use switch...case... instead of multiple if..else.. for DeleteConditionHandler::is_condition_value_valid
- Use range loop to simplify code
- Reduce some compare operations in Cond::del_eval
- Improve some branch predictions in Reader
- Fix and improve some unit tests

## Types of changes

## Checklist

### DIFF
--- a/be/src/olap/olap_cond.cpp
+++ b/be/src/olap/olap_cond.cpp
@@ -87,9 +87,6 @@ static CondOp parse_op_type(const string& op) {
     return OP_NULL;
 }
 
-Cond::Cond()
-        : op(OP_NULL), operand_field(nullptr), min_value_field(nullptr), max_value_field(nullptr) {}
-
 Cond::~Cond() {
     delete operand_field;
     for (auto& it : operand_set) {
@@ -109,6 +106,7 @@ OLAPStatus Cond::init(const TCondition& tcond, const TabletColumn& column) {
     }
     if (op == OP_IS) {
         // 'is null' or 'is not null'
+        DCHECK_EQ(tcond.condition_values.size(), 1);
         auto operand = tcond.condition_values.begin();
         std::unique_ptr<WrapperField> f(WrapperField::create(column, operand->length()));
         if (f == nullptr) {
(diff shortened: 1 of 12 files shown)


be/src/olap/olap_cond.cpp
```diff
@@ -382,28 +386,30 @@ int Cond::del_eval(const std::pair<WrapperField*, WrapperField*>& stat) const {
             } else if (stat.first->is_null() && !stat.second->is_null()) {
                 ret = DEL_PARTIAL_SATISFIED;
             } else {
+                DCHECK(false);
                 //不会出现min不为NULL，max为NULL
-                ret = DEL_NOT_SATISFIED;
+                ret = DEL_SATISFIED;
```
morningman, line 391: I think `DEL_NOT_SATISFIED` is more safe. Or you'd better change DCHECK to CHECK

be/src/olap/olap_cond.cpp
```diff
@@ -382,28 +386,30 @@ int Cond::del_eval(const std::pair<WrapperField*, WrapperField*>& stat) const {
...
             }
         } else {
             if (stat.first->is_null() && stat.second->is_null()) {
                 ret = DEL_NOT_SATISFIED;
             } else if (stat.first->is_null() && !stat.second->is_null()) {
                 ret = DEL_PARTIAL_SATISFIED;
             } else {
+                DCHECK(false);
```
morningman, line 399: Change to CHECK, or ret = DEL_NOT_SATISFIED

be/src/olap/reader.cpp
```diff
@@ -599,13 +599,12 @@ OLAPStatus Reader::_init_params(const ReaderParams& read_params) {
...
     if (_tablet->tablet_schema().has_sequence_col()) {
         _sequence_col_idx = _tablet->tablet_schema().sequence_col_idx();
-        if (_sequence_col_idx != -1) {
-            for (auto col : _return_columns) {
-                // query has sequence col
-                if (col == _sequence_col_idx) {
-                    _has_sequence_col = true;
-                    break;
-                }
+        DCHECK_NE(_sequence_col_idx, -1);
```
morningman, line 602: Why using DCHECK here?

acelyc111: Because `has_sequence_col ` in line 600 is `inline bool has_sequence_col() const { return  _sequence_col_idx != -1; }`, so `_sequence_col_idx` shoule always be `-1`.

morningman (dismissed):
LGTM

morningman (dismissed):
LGTM

morningman (approved):
LGTM

Issue #4926, opened 2020-11-19
UNION ALL query return wrong type, NULL value but get ''


rongyousu:
the detail is that:
sql:
with a as 
(
  select 2 as id, '1234' as user_id
   union all
    select 1 as id, NULL as user_id 
),
b as 
(  
  select 4 as id, '543' as user_id
  union all
   select 3 as id, NULL as user_id
)
select id,user_id from a
union all
select id ,user_id from b

the result is :
![image](https://user-images.githubusercontent.com/7709573/99634610-bfac0d80-2a7b-11eb-915c-7cb9ef125068.png)

my value is "NULL "  ,but  use with lable ,the result is issue, now is ''

HappenLee:
Hi，It's look like is FE query plan's problem. I will try to fix the problem

PR #4932, opened 2020-11-20
[Bug] Fix the bug of NULL do not show in CTE statement.

HappenLee:
All Column create in inlineView will set `allowNull = false`, which will cause `NULL` data in CTE be process will be ignore.
So we should set column in inlineView allowNull to make sure correct of query.

issue:#4926

## Proposed changes

## Types of changes

## Checklist

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/InlineViewRef.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/InlineViewRef.java
@@ -266,7 +266,9 @@ public class InlineViewRef extends TableRef {
             }
 
             columnSet.add(colAlias);
-            columnList.add(new Column(colAlias, selectItemExpr.getType().getPrimitiveType()));
+            // TODO: inlineView threat all column is nullable to make sure query results are correct
+            // we should judge column whether is nullable by selectItemExpr in the future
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/org/apache/doris/analysis/InlineViewRef.java
```diff
@@ -266,7 +266,9 @@ public TupleDescriptor createTupleDescriptor(Analyzer analyzer) throws AnalysisE
             }
 
             columnSet.add(colAlias);
-            columnList.add(new Column(colAlias, selectItemExpr.getType().getPrimitiveType()));
+            // TODO: inlineView threat all column is nullable to make sure query results are correct
+            // we should judge column whether is nullable by selectItemExpr in the future
+            columnList.add(new Column(colAlias, selectItemExpr.getType().getPrimitiveType(), true));
```
EmmyMiao87, line 271: Unfortunately, we cannot directly obtain the information about whether slot desc is nullable or not from Expr.

EmmyMiao87 (approved):
LGTM

EmmyMiao87:
Please add more commit log

HappenLee:
> Please add more commit log

done

PR #4933, opened 2020-11-20
[Refactor] Short circuit return to avoid meaningless loop

acelyc111:
## Proposed changes

I found function Tablet::rowset_meta_is_useful will cost a considerable CPU load, we can short circuit return to avoid meaningless loop.

## Types of changes

## Checklist

### DIFF
--- a/be/src/olap/tablet.cpp
+++ b/be/src/olap/tablet.cpp
@@ -1214,11 +1214,10 @@ void Tablet::do_tablet_meta_checkpoint() {
 
 bool Tablet::rowset_meta_is_useful(RowsetMetaSharedPtr rowset_meta) {
     ReadLock rdlock(&_meta_lock);
-    bool find_rowset_id = false;
     bool find_version = false;
     for (auto& version_rowset : _rs_version_map) {
         if (version_rowset.second->rowset_id() == rowset_meta->rowset_id()) {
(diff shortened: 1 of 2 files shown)


be/src/olap/tablet.cpp
```diff
@@ -1214,33 +1214,32 @@ void Tablet::do_tablet_meta_checkpoint() {
 
 bool Tablet::rowset_meta_is_useful(RowsetMetaSharedPtr rowset_meta) {
```
kangkaisen, line 1215: Could you add a comment for `rowset_meta_is_useful`? what's the meaning of `useful`?

Why when `find_version` is false, the rowset is unuseful?

acelyc111: @kangkaisen Comment added.

> Why when find_version is false, the rowset is unuseful?

I think you want to ask why when find_version is **true**, the rowset is unuseful. Because when version founded, means the tablet has the data in this version (maybe compacted), we can sweep up this rowset. Only when the rowset has some new data (version is not covered, i.e. find_version is false), the rowset is useful.

kangkaisen: OK, Thanks!

kangkaisen (approved):
+1

chaoyli (approved):

Issue #4937, opened 2020-11-23
[Proposal] Support more column type in schema change


HappenLee:
**Is your feature request related to a problem? Please describe.**
Now Doris support schema change column type:
```
TINYINT/SMALLINT/INT/BIGINT is converted to TINYINT/SMALLINT/INT/BIGINT/DOUBLE.
Convert VARCHAR to TINYINT/SMALLINT/INT/BIGINT/LARGEINT/FLOAT/DOUBLE.
Convert VARCHAR to DATE (currently support six formats: "%Y-%m-%d", "%y-%m-%d", "%Y%m%d", "%y%m%d", "%Y/%m/%d, "%y/%m/%d")
```

we should support:
```
TINYINT/SMALLINT/INT/BIGINT/LARGEINT/FLOAT/DOUBLE convert to a wider range of numeric types
Convert CHAR to TINYINT/SMALLINT/INT/BIGINT/LARGEINT/FLOAT/DOUBLE.
Convert CHAR to DATE (currently support six formats: "%Y-%m-%d", "%y-%m-%d", "%Y%m%d", "%y%m%d", "%Y/%m/%d, "%y/%m/%d")
```

PR #4938, opened 2020-11-23
[Schema change] Support More column type in schema change

HappenLee:
1. Support modify column type CHAR to TINYINT/SMALLINT/INT/BIGINT/LARGEINT/FLOAT/DOUBLE/DATE
and TINYINT/SMALLINT/INT/BIGINT/LARGEINT/FLOAT/DOUBLE convert to a wider range of numeric types (#4937)

2. Use template to refactor code of types.h and schema_change.cpp to delete redundant code.

fix issue:#4937

## Proposed changes

## Types of changes

## Checklist

### DIFF
--- a/be/src/olap/types.h
+++ b/be/src/olap/types.h
@@ -507,18 +507,17 @@ static void prepare_char_before_convert(const void* src) {
 }
 
 template <typename T>
-T convert_from_varchar(const Slice* src_value, StringParser::ParseResult& parse_res , std::true_type) {
+T convert_from_varchar(const Slice* src_value, StringParser::ParseResult& parse_res, std::true_type) {
     return StringParser::string_to_int<T>(src_value->get_data(), src_value->get_size(), &parse_res);
 }
 
 template <typename T>
-T convert_from_varchar(const Slice* src_value, StringParser::ParseResult& parse_res , std::false_type) {
+T convert_from_varchar(const Slice* src_value, StringParser::ParseResult& parse_res, std::false_type) {
     return StringParser::string_to_float<T>(src_value->get_data(), src_value->get_size(), &parse_res);
 }
 
 template <typename T>
 OLAPStatus arithmetic_convert_from_varchar(void* dest, const void* src) {
-    prepare_char_before_convert(src);
     auto src_value = reinterpret_cast<const Slice*>(src);
     StringParser::ParseResult parse_res;
     //TODO: use C++17 if constexpr to replace label assignment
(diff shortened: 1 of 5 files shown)


be/src/olap/types.h
```diff
@@ -495,12 +496,32 @@ struct BaseFieldtypeTraits : public CppTypeTraits<field_type> {
...
+
 template <typename T>
-OLAPStatus convert_int_from_varchar(void* dest, const void* src) {
-    using SrcType = typename CppTypeTraits<OLAP_FIELD_TYPE_VARCHAR>::CppType;
-    auto src_value = reinterpret_cast<const SrcType*>(src);
+T convert_from_varchar(const Slice* src_value, StringParser::ParseResult& parse_res, std::true_type) {
+    return StringParser::string_to_int<T>(src_value->get_data(), src_value->get_size(), &parse_res);
+}
+
+template <typename T>
```
chaoyli, line 514: There may use traits ?
Or use if constexpr (std::is_floating_point_v<T>)) to distinguish the two cases.

be/src/olap/types.h
```diff
@@ -509,98 +530,55 @@ OLAPStatus convert_int_from_varchar(void* dest, const void* src) {
...
-    static OLAPStatus convert_from(void* dest, const void* src, const TypeInfo* src_type, MemPool* mem_pool) {
-        if (src_type->type() == OLAP_FIELD_TYPE_VARCHAR) {
-            return convert_int_from_varchar<CppType>(dest, src);
-        }
-        return OLAPStatus::OLAP_ERR_INVALID_SCHEMA;
-    }
-};
+template <FieldType fieldType>
+struct FieldTypeTraits : public ArithmeTicFieldtypeTraits<fieldType,
+        std::is_arithmetic<typename BaseFieldtypeTraits<fieldType>::CppType>::value && std::is_signed<typename BaseFieldtypeTraits<fieldType>::CppType>::value> {};
```
chaoyli, line 563: use std::is_signed for what?

be/src/olap/types.h
```diff
@@ -509,98 +530,55 @@ OLAPStatus convert_int_from_varchar(void* dest, const void* src) {
...
+    prepare_char_before_convert(src);
+    return arithmetic_convert_from_varchar<T>(dest, src);
 }
 
-template<FieldType field_type>
-struct FieldTypeTraits : public BaseFieldtypeTraits<field_type> { };
+// Using ArithmeTicFieldtypeTraits to Derived code for OLAP_FIELD_TYPE_XXXINT, OLAP_FIELD_TYPE_FLOAT,
+// OLAP_FIELD_TYPE_DOUBLE, to reduce redundant code
+template <FieldType fieldType, bool isArithmetic>
+struct ArithmeTicFieldtypeTraits : public BaseFieldtypeTraits<fieldType> {
```
chaoyli, line 541: Numeric may be a better name.

chaoyli (changes requested):

be/src/olap/types.h
HappenLee: OK，i will consider to change the name of it

be/src/olap/types.h
HappenLee: Now， ArithmeticField just to keep the small with old code.
old code only Partial Template Specialization the OLAP_FIELD_TYPE_INT but not OLAP_FIELD_TYPE_UNSIGNED_INT. There is no need to Partial Template Specialization for unsign type, so there check the CppType whether is_signed

be/src/olap/types.h
HappenLee: Hi, chaoyli.

```if constexpr``` only support in c++17 and ```std::is_floating_point_v``` support in c++14.

Now, Doris only support C++11, So I use label assignment to distinguish the float and int.

chaoyli (approved):

Issue #4935, opened 2020-11-20
Colocate table stays unstable when a large number of be restarts frequently


gengjun-git:
**Describe the bug**
When colocate table balance, if multiple be down, the tablets will be distributed to the existing machines, but if the existing machine disk is not enough to accommodate these tablets, and the balance must wait for the migration of these tablets to complete,  The strategy will fall into an endless loop.

gengjun-git:
fix in #4936

PR #4955, opened 2020-11-26
[BUG] Fix colocate balance bug when there is decommissioned be

gengjun-git:
fix bug for issue #4935, PR #4936 
We should ignore decommissioned be when select bes to balance group bucketSeq.


kangkaisen (approved):
+1

PR #4959, opened 2020-11-26
[Bug] Fix the loss of precision when Decimal calculates variance/stddev

### LINKED ISSUE #3127: STDDEV_SAMP函数计算结果与mysql偏差大
版本：doris0.10.11
描述：STDDEV_SAMP函数计算结果与mysql偏差有些大。数据参考附件，mysql里结果为：0.46870514257154544；doris里计算结果为0.5081618046536185

附件：
[data.txt](https://github.com/apache/incubator-doris/files/4338752/data.txt)

xinyiZzz:
Update on 2020-11-30:
Replaced by calculating variance based on Decimal as intermediate variable, this is a more elegant way. See below for details.
The first trick method of cast to double before can be seen: https://github.com/xinyiZzz/incubator-doris/commit/1f172482b87167cede1b6dc4ca0b2e4c880a40a1
——————————————————————————————————————————
## Proposed changes

By cast Decimal to Double, realize the variance calculation of Decimal type column.
In my local Spark2.4 and Mysql, the variance of Decimal is also calculated by cast to Double.
NOTE: There may be two parts of precision loss:
- Decimal cast to Double
- In the variance calculation process, the precision of the Double type intermediate variable is lost (the effective number of digits of Double is 16)

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #3127), and have described the bug/feature there in detail

## Further comments

Before fix the variance calculation by Decimal cast to Double, I implemented a method of directly calculating the variance on Decimal, using DecimalV2Value as the type of intermediate variable. see Commit: https://github.com/xinyiZzz/incubator-doris/commit/0288a6f76245bdf1f6eb0b54d3f5921414772066

Compared with cast to Double, the considerations and advantages of calculating variance directly based on Decimal:
- Less loss of precision.
- Faster performance.
- Consistent with other functions, such as `avg()`, `sum()`, etc., these functions use Decimal to calculate directly in Doris, and use DecimalV2Value as the type of intermediate variable.

However, there is an overflow problem in calculating variance based on Decimal. When `Variance> 9223372036854775807999999999/N` (9223372036854775807999999999 is the maximum value of DecimalV2Value, N is the number of data rows), the intermediate variable in the calculation process will overflow and return a maximum value.

I feel that theoretical business requirements will not calculate the variance greater than `9223372036854775807999999999/N`, so cast to Double means lower accuracy and worse performance, but both Spark and Mysql use cast to double methods, so I Pull this code.

#### precision comparison:
- When the first 16 bits of the data are the same, such as `12345678901234560` and `12345678901234561`, at this time, the result calculated by cast to double is 0 (loss of precision), and the correct result by direct calculation by Decimal is 0.5
- When the variance result is large, compared to the result after cast to double `42981942961.912369`, the result directly calculated by Decimal `42981942961.913988699` is more accurate

#### Performance comparison:
The test results of variance/stddev and other functions are as follows:
- Cast to Double calculation: 0.054s
- Decimal direct calculation: 0.034s

Table building method:
```
CREATE TABLE stddev_samp_test29
(event_day int, money DECIMAL(27,9) DEFAULT "0")
DISTRIBUTED BY HASH(event_day) BUCKETS 5
```

Data with 10w rows of equal length, such as `99.00066`, is generated by:
```
f = open('data.txt','w')
n = 100000
while (n):
 f.write('0,' + "%.5f"% (99 + random.random()) +'\n')
 n -= 1
```

It’s too important to complete the preliminary research, so sad...

### DIFF
--- a/be/src/exprs/aggregate_functions.cpp
+++ b/be/src/exprs/aggregate_functions.cpp
@@ -2061,6 +2061,13 @@ struct KnuthVarianceState {
     int64_t count;
 };
 
+// Use Decimal to store the intermediate results of the variance algorithm
+struct DecimalV2KnuthVarianceState {
+    DecimalV2Val mean;
+    DecimalV2Val m2;
+    int64_t count = 0;
+};
+
 // Set pop=true for population variance, false for sample variance
 static double compute_knuth_variance(const KnuthVarianceState& state, bool pop) {
     // Return zero for 1 tuple specified by
@@ -2070,6 +2077,16 @@ static double compute_knuth_variance(const KnuthVarianceState& state, bool pop) 
     return state.m2 / (state.count - 1);
 }
 
+// The algorithm is the same as above, using decimal as the intermediate variable
+static DecimalV2Value decimalv2_compute_knuth_variance(const DecimalV2KnuthVarianceState& state, bool pop) {
+    DecimalV2Value new_count = DecimalV2Value();
+    new_count.assign_from_double(state.count);
+    if (state.count == 1) return new_count;
+    DecimalV2Value new_m2 = DecimalV2Value::from_decimal_val(state.m2);
+    if (pop) return new_m2 / new_count;
+    else return new_m2 / new_count.assign_from_double(state.count - 1);
+}
+
 void AggregateFunctions::knuth_var_init(FunctionContext* ctx, StringVal* dst) {
     dst->is_null = false;
     // TODO(zc)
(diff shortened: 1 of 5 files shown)


EmmyMiao87 (dismissed):
LGTM

morningman:
I think this PR it too trick.
https://github.com/xinyiZzz/incubator-doris/commit/0288a6f76245bdf1f6eb0b54d3f5921414772066
This seems better to me.

xinyiZzz:
> I think this PR it too trick.
> [xinyiZzz@0288a6f](https://github.com/xinyiZzz/incubator-doris/commit/0288a6f76245bdf1f6eb0b54d3f5921414772066)
> This seems better to me.

Maybe I can do it in an elegant way, doing cast to double in Be's function.
This will change more places. The trick above is the easiest way I can find.
Because I am not sure if it is inconsistent with Spark, whether it will cause problems later.

morningman (approved):
LGTM

PR #4212, opened 2020-07-30
Compaction rules optimization

ZhangYu0123:
## Proposed changes

Compaction rules optimization, the detail problem description and design to see #4164.
This pr commits 2 functions:
(1) add the cumulative policy configable, and implement original policy.
(2) implement universal policy, the optimization version in #4164.

## Types of changes

What types of changes does your code introduce to Doris?
_Put an `x` in the boxes that apply_

- [] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation Update (if none of the other choices apply)
- [] Code refactor (Modify the code structure, format the code, etc...)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I have create an issue on #ISSUE, and have described the bug/feature there in detail
- [x] Compiling and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] If this change need a document change, I have updated the document
- [x] Any dependent changes have been merged

PR #4964, opened 2020-11-26
[Bug][Compaction] Fix bug that output rowset is not deleted after compaction failure.

### LINKED ISSUE #4963: [Bug][Compaction] The data files are not deleted after compaction failed.
**Describe the bug**
When the compaction fails, the new data files generated are not cleaned up,
resulting in a large number of files remaining on the disk.

The reason of compaction failure is because the number of filtered rows is wrong, so that
the validation after compaction will fail. log like this:

```
row_num does not match between cumulative input and output! input_row_num=4, merged_row_num=0, filtered_row_num=0, output_row_num=0
failed to init base compaction. res=-224, table=10209.1604932729.e643691394eb8627-841daf125e14c3a0
```

Version 0.13.11

morningman:
## Proposed changes

This CL fix 2 bugs:

1. 
When the compaction fails, we must explicitly delete the output rowset,
otherwise the GC logic cannot process these rows.

2. 
Base compaction failed if compaction process include some delete version in SegmentV2,
Because the number of filtered rows is wrong. 

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4963 ), and have described the bug/feature there in detail

### DIFF
--- a/be/src/olap/compaction.cpp
+++ b/be/src/olap/compaction.cpp
@@ -37,6 +37,20 @@ Compaction::Compaction(TabletSharedPtr tablet, const std::string& label,
 
 Compaction::~Compaction() {}
 
+OLAPStatus Compaction::compact() {
+    RETURN_NOT_OK(prepare_compact());
+    RETURN_NOT_OK(execute_compact());
+    return OLAP_SUCCESS;
+}
+
+OLAPStatus Compaction::execute_compact() {
+    OLAPStatus st = execute_compact_impl();
+    if (st != OLAP_SUCCESS) {
+        gc_output_rowset();
+    }
+    return st;
+}
+
 OLAPStatus Compaction::do_compaction(int64_t permits) {
     TRACE("start to do compaction");
     _tablet->data_dir()->disks_compaction_score_increment(permits);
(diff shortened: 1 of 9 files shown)


be/src/olap/reader.h
```diff
@@ -128,7 +128,7 @@ class Reader {
     }
 
     uint64_t filtered_rows() const {
-        return _stats.rows_del_filtered;
```
xy720, line 131: Hi, is it compatible with segment v1?
In alpha_rowset_reader.h, filtered_rows() is only calculate rows_del_filtered.

morningman: Yes, it is compatible. Because rows_conditions_filtered is only used for SegmentV2, so it should always to 0 in Segment V1.

xy720 (commented):
LGTM

yangzhg (dismissed):
+1

be/src/olap/compaction.cpp
```diff
@@ -164,17 +173,11 @@ void Compaction::modify_rowsets() {
     _tablet->save_meta();
 }
 
-OLAPStatus Compaction::gc_unused_rowsets() {
+void Compaction::gc_output_rowset() {
```
chaoyli, line 176: why not call gc_output_rowset in every succeed? It seems a new imported bug ?

chaoyli (changes requested):

be/src/olap/compaction.cpp
morningman: `gc_output_rowset()` should only be called when compaction succeed.

In #4212, we have implemented a new compaction logic, so that the input rowset will be deleted later. And we only need to delete the output rowset when failed.

chaoyli (dismissed):

yangzhg (approved):
+1

PR #4975, opened 2020-11-27
Fix locate bug when start_pos larger than str len

### LINKED ISSUE #4976: Fix locate bug when start_pos larger than str len
```
select locate('', 'abc', 10); 
```
Return 10, it's a bug

chaoyli:
```
select locate('', 'abc', 10); 
```
Return 0 not 10


kangkaisen (approved):
+1

PR #5008, opened 2020-12-02
[Optimize][Cache]Implementation of Separated Page Cache

### LINKED ISSUE #4995: [Proposal] Separate Page Cache by page type
Currently, all types of pages (including DATA_PAGE, INDEX_PAGE, DICT_PAGE, SHORT_KEY_PAGE) in segment_v2 are cached by StoragePageCache in one single memory space. In many query situations, we expect to reuse the index pages in cache, but which are probably replaced by the large amount of data pages due to the LRU algorithm. Even in some extreme cases, the high frequency of LRU replacement usage caused by data pages will bring great pressure to the system.

A separated page cache could be used to solve this problem.

Implementation:
- Change the "_cache" member of StoragePageCache class into "_data_cache" (for DATA) and "_index_page" (for INDEX, DICT, SHORT_KEY) and Add an "_index_cache_ratio" member as for the ratio of index cache.
- Add a configuration item "index_cache_ratio" to set the ratio. The range is: 0 <= ratio <= 1, 0/1 stands for only data/index pages are cached.
- Add "page_type" as an argument when reading pages, so that the StoragePageCache could use it to allocate the cache.

Advantages:
- By separating data and index pages, the hit rate of index pages could be highly improved and by tuning the ratio the hit rate of data pages will not be much decreased.
- By setting the config::index_cache_ratio (including 0 or 1), users can adapt the cache ratio to fit their query situations.

Skysheepwang:
## Proposed changes
#4995
**Implementation of Separated Page Cache**
- Add config "index_page_cache_ratio" to set the ratio of capacity of index page cache
- Change the member of StoragePageCache to maintain two type of cache
- Change the interface of StoragePageCache for selecting type of cache
- Change the usage of page cache in read_and_decompress_page in page_io.cpp
  - add page type as argument
  - check if current page type is available in StoragePageCache (cover the situation of ratio == 0 or 1)
- Add type as argument in superior call of read_and_decompress_page
- Change Unit Test

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4995), and have described the bug/feature there in detail


be/src/common/config.h
```diff
@@ -260,6 +260,8 @@ CONF_Int64(index_stream_cache_capacity, "10737418240");
 
 // Cache for storage page size
 CONF_String(storage_page_cache_limit, "20G");
+// Ratio for index page cache
```
acelyc111, line 263: I think it's better to give more description, such as what does the left size used for?

be/src/olap/page_cache.h
```diff
@@ -53,36 +54,44 @@ class StoragePageCache {
...
     // This function is thread-safe, and when two clients insert two same key
     // concurrently, this function can assure that only one page is cached.
     // The in_memory page will have higher priority.
     void insert(const CacheKey& key, const Slice& data, PageCacheHandle* handle,
-                bool in_memory = false);
+                segment_v2::PageTypePB page_type, bool in_memory = false);
+
+    // Page cache available check.
+    // When ratio is set to 0 or 1, the index or data cache will not be allocated.
+    bool is_cache_available(segment_v2::PageTypePB page_type);
```
acelyc111, line 86: Better to define it as a static finction

be/src/olap/page_cache.cpp
```diff
@@ -21,35 +21,81 @@ namespace doris {
...
-    static StoragePageCache instance(capacity);
+    static StoragePageCache instance(capacity, index_cache_ratio);
     _s_instance = &instance;
 }
 
-StoragePageCache::StoragePageCache(size_t capacity)
-        : _cache(new_lru_cache("StoragePageCache", capacity)) {}
+StoragePageCache::StoragePageCache(size_t capacity, double index_cache_ratio)
+        : _index_cache_ratio(index_cache_ratio) {
+    if (index_cache_ratio == 0) {
```
acelyc111, line 32: For double type values, better to use expression like `std::abs(index_cache_ratio) < 1e-6` to judge whether they are equaled. If you think it's too ugly, you can use percentagle int value instead.

be/src/olap/page_cache.cpp
```diff
@@ -21,35 +21,81 @@ namespace doris {
...
+        _data_page_cache = std::unique_ptr<Cache>(new_lru_cache("DataPageCache", capacity));
+    }
+    else if (index_cache_ratio == 1) {
+        _index_page_cache = std::unique_ptr<Cache>(new_lru_cache("IndexPageCache", capacity));
+    }
+    else if (index_cache_ratio > 0 && index_cache_ratio < 1) {
+        _data_page_cache = std::unique_ptr<Cache>(new_lru_cache("DataPageCache", static_cast<size_t>(capacity * (1 - index_cache_ratio))));
+        _index_page_cache = std::unique_ptr<Cache>(new_lru_cache("IndexPageCache", static_cast<size_t>(capacity * index_cache_ratio)));
+    }
```
acelyc111, line 41: What if the other values passed, how about add `CHECK(false) << "xxx"` here?

be/src/olap/page_cache.cpp
```diff
@@ -21,35 +21,81 @@ namespace doris {
...
+        }
+        *handle = PageCacheHandle(_data_page_cache.get(), lru_handle);
+        break;
+    case segment_v2::INDEX_PAGE:
+        lru_handle = _index_page_cache->lookup(key.encode());
+        if (lru_handle == nullptr) {
+            return false;
+        }
+        *handle = PageCacheHandle(_index_page_cache.get(), lru_handle);
+        break;
```
acelyc111, line 60: There are some duplicate code, how about define a function like `Cache::Handle* get_cache(segment_v2::PageTypePB page_type)` to get `lru_handle ` at first, then do the same thing as before?

be/src/olap/page_cache.cpp
```diff
@@ -21,35 +21,81 @@ namespace doris {
...
+    Cache::Handle* lru_handle = nullptr;
+    switch (page_type) {
+    case segment_v2::DATA_PAGE:
+        lru_handle = _data_page_cache->insert(key.encode(), data.data, data.size, deleter, priority);
+        *handle = PageCacheHandle(_data_page_cache.get(), lru_handle);
+        break;
+    case segment_v2::INDEX_PAGE:
+        lru_handle = _index_page_cache->insert(key.encode(), data.data, data.size, deleter, priority);
+        *handle = PageCacheHandle(_index_page_cache.get(), lru_handle);
+        break;
```
acelyc111, line 85: Same

be/src/olap/page_cache.cpp
```diff
@@ -21,35 +21,81 @@ namespace doris {
...
+    }
+}
+
+bool StoragePageCache::is_cache_available(segment_v2::PageTypePB page_type) {
+    if (page_type == segment_v2::DATA_PAGE && _index_cache_ratio == 1) {
+        return false;
+    }
+    else if (page_type == segment_v2::INDEX_PAGE && _index_cache_ratio == 0) {
+        return false;
+    }
```
acelyc111, line 97: If you do the refactor as mentioned above, you can simply judge whether the return `lru_handle ` is nullptr or not

docs/en/administrator-guide/config/be_config.md
```diff
@@ -695,6 +695,8 @@ Indicates how many tablets in this data directory failed to load. At the same ti
 
 ### `storage_page_cache_limit`
 
+### `index_page_cache_ratio`
```
acelyc111, line 698: Add descriptions.

docs/zh-CN/administrator-guide/config/be_config.md
```diff
@@ -693,6 +693,8 @@ load tablets from header failed, failed tablets size: xxx, path=xxx
 
 ### `storage_page_cache_limit`
 
+### `index_page_cache_ratio`
```
acelyc111, line 696: Add descriptions.

be/src/olap/page_cache.cpp
Skysheepwang: I think it is a good idea to use the percentage as index cache ratio because of the safety problem as you mentioned and could also avoid using the type cast. I will change the definition. Thank you very much!

be/src/olap/page_cache.cpp
acelyc111: Sorry, it should be `Cache* get_cache(segment_v2::PageTypePB page_type)` to get Cache at first

be/src/olap/page_cache.cpp
```diff
@@ -21,35 +21,70 @@ namespace doris {
...
-    auto lru_handle = _cache->lookup(key.encode());
+bool StoragePageCache::lookup(const CacheKey& key, PageCacheHandle* handle, segment_v2::PageTypePB page_type) {
+    Cache::Handle* lru_handle = nullptr;
+    lru_handle = _get_page_cache(page_type)->lookup(key.encode());
     if (lru_handle == nullptr) {
         return false;
     }
-    *handle = PageCacheHandle(_cache.get(), lru_handle);
+    *handle = PageCacheHandle(_get_page_cache(page_type), lru_handle);
     return true;
```
acelyc111, line 54: How about
```
auto cache = _get_page_cache(page_type);
auto lru_handle = cache->lookup(key.encode());
if (lru_handle == nullptr) {
    return false;
}
*handle = PageCacheHandle(cache, lru_handle);
return true;
```

be/src/olap/page_cache.cpp
```diff
@@ -21,35 +21,70 @@ namespace doris {
...
+    Cache::Handle* lru_handle = nullptr;
+    lru_handle = _get_page_cache(page_type)->insert(key.encode(), data.data, data.size, deleter, priority);
+    *handle = PageCacheHandle(_get_page_cache(page_type), lru_handle);
+}
+
+bool StoragePageCache::is_cache_available(segment_v2::PageTypePB page_type) {
+    if (_get_page_cache(page_type) == nullptr) {
+        return false;
+    }
+    return true;
```
acelyc111, line 75:
```suggestion
    return _get_page_cache(page_type) != nullptr;
```

be/src/olap/page_cache.cpp
```diff
@@ -21,35 +21,70 @@ namespace doris {
...
     CachePriority priority = CachePriority::NORMAL;
     if (in_memory) {
         priority = CachePriority::DURABLE;
     }
 
-    auto lru_handle = _cache->insert(key.encode(), data.data, data.size, deleter, priority);
-    *handle = PageCacheHandle(_cache.get(), lru_handle);
+    Cache::Handle* lru_handle = nullptr;
+    lru_handle = _get_page_cache(page_type)->insert(key.encode(), data.data, data.size, deleter, priority);
+    *handle = PageCacheHandle(_get_page_cache(page_type), lru_handle);
```
acelyc111, line 68:
```suggestion
    auto cache = _get_page_cache(page_type);
    Cache::Handle* lru_handle = cache->insert(key.encode(), data.data, data.size, deleter, priority);
    *handle = PageCacheHandle(cache, lru_handle);
```

be/src/olap/page_cache.cpp
Skysheepwang: Thank you! I think it is a really good change. I will take it and change the method to an inline member method.

be/src/olap/page_cache.cpp
```diff
@@ -21,35 +21,51 @@ namespace doris {
...
+        _data_page_cache = std::unique_ptr<Cache>(new_lru_cache("DataPageCache", capacity));
+    }
+    else if (index_cache_percentage == 100) {
+        _index_page_cache = std::unique_ptr<Cache>(new_lru_cache("IndexPageCache", capacity));
+    }
+    else if (index_cache_percentage > 0 && index_cache_percentage < 100) {
+        _data_page_cache = std::unique_ptr<Cache>(new_lru_cache("DataPageCache", capacity * (100 - index_cache_percentage) / 100));
+        _index_page_cache = std::unique_ptr<Cache>(new_lru_cache("IndexPageCache", capacity * index_cache_percentage / 100));
+    }
+    else {
```
acelyc111, line 42:
```suggestion
    } else if (index_cache_percentage == 100) {
        _index_page_cache = std::unique_ptr<Cache>(new_lru_cache("IndexPageCache", capacity));
    } else if (index_cache_percentage > 0 && index_cache_percentage < 100) {
        _data_page_cache = std::unique_ptr<Cache>(new_lru_cache("DataPageCache", capacity * (100 - index_cache_percentage) / 100));
        _index_page_cache = std::unique_ptr<Cache>(new_lru_cache("IndexPageCache", capacity * index_cache_percentage / 100));
    } else {
```

acelyc111 (approved):

acelyc111 (dismissed):
LGTM

be/src/olap/page_cache.cpp
```diff
@@ -21,35 +21,48 @@ namespace doris {
...
+        _index_page_cache = std::unique_ptr<Cache>(new_lru_cache("IndexPageCache", capacity * index_cache_percentage / 100));
+    } else {
+        CHECK(false) << "invalid index page cache percentage";
+    }
+}
 
-bool StoragePageCache::lookup(const CacheKey& key, PageCacheHandle* handle) {
-    auto lru_handle = _cache->lookup(key.encode());
+bool StoragePageCache::lookup(const CacheKey& key, PageCacheHandle* handle, segment_v2::PageTypePB page_type) {
+    auto cache = _get_page_cache(page_type);
```
morningman, line 45: the cache maybe null if the index_page_cache_percentage is 0 or 100

be/src/olap/page_cache.cpp
```diff
@@ -21,35 +21,48 @@ namespace doris {
...
     auto deleter = [](const doris::CacheKey& key, void* value) { delete[](uint8_t*) value; };
 
     CachePriority priority = CachePriority::NORMAL;
     if (in_memory) {
         priority = CachePriority::DURABLE;
     }
 
-    auto lru_handle = _cache->insert(key.encode(), data.data, data.size, deleter, priority);
-    *handle = PageCacheHandle(_cache.get(), lru_handle);
+    auto cache = _get_page_cache(page_type);
```
morningman, line 63: cache may be null

be/src/olap/rowset/segment_v2/page_io.cpp
```diff
@@ -112,7 +112,7 @@ Status PageIO::read_and_decompress_page(const PageReadOptions& opts, PageHandle*
     auto cache = StoragePageCache::instance();
     PageCacheHandle cache_handle;
     StoragePageCache::CacheKey cache_key(opts.rblock->path(), opts.page_pointer.offset);
-    if (opts.use_page_cache && cache->lookup(cache_key, &cache_handle)) {
+    if (opts.use_page_cache && cache->is_cache_available(opts.type) && cache->lookup(cache_key, &cache_handle, opts.type)) {
```
morningman, line 115: I think the check of `is_cache_available()` can be put into `lookup()` method for easy to use?
As well as the following `insert` logic

Skysheepwang: I choose to make the `is_cache_available()` check as an interface basically for two reasons:

1. If the check is failed (which means that the percentage is set to 0 or 100), the `lookup()` and `insert()` method will not be called, which could save some of the cost of parameter passing.
2. It will do minimal changes to the original `read_and_decompress_page()` method in page_io.cpp and remain its design on whether call the `lookup()` and `insert()` methods.

be/src/olap/page_cache.cpp
Skysheepwang: Actually not because the `is_cache_available()` check will be called before using the `lookup()` method.

be/src/olap/page_cache.cpp
Skysheepwang: Same as above.

be/src/olap/page_cache.cpp
morningman: Ok

morningman (dismissed):
LGTM

morningman:
hi @Skysheepwang Could you please rebase the trunk to make UT the NewDorisTest happy?

Skysheepwang:
@morningman I 've already rebased but there is still something wrong in UT (Incubator Doris), but my local UT was passed so could you please help me check the problem?

morningman:
> @morningman I 've already rebased but there is still something wrong in UT (Incubator Doris), but my local UT was passed so could you please help me check the problem?

It seems our proxy of github has some problem. I checked the UT, it works.

morningman (dismissed):
LGTM

acelyc111 (dismissed):
LGTM

Skysheepwang:
@morningman @acelyc111 Sorry I just made a mistake to push my code to this branch, which dismissed your approving review, please add again.

acelyc111 (dismissed):
LGTM

morningman (dismissed):
LGTM

morningman:
Hi @Skysheepwang, Sorry for late review. Please resolve the conflict~

acelyc111:
> Hi @Skysheepwang, Sorry for late review. Please resolve the conflict~

@morningman conflicts have been resolved.

morningman (approved):
LGTM

PR #5013, opened 2020-12-02
[Repair] Add an option whether to allow the partition column to be NULL

### LINKED ISSUE #5012: [Proposal] Add an option whether to allow the partition column to be NULL
Strict data consistency is required in similar financial scenarios, when the partition column in the data is NULL, the import needs to report an error instead of the smallest partition, In relational databases support strong verification

Add an option whether to allow the partition column to be NULL, When the option is enabled, the partition column must be NOT NULL when creating table, otherwise an error will be reported.

xinyiZzz:
## Proposed changes

Add an option `allow_partition_column_nullable`, which means whether to allow the partition column to be NULL.
When the option is enabled, the partition column must be NOT NULL when create table, otherwise an error will be reported.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #5012), and have described the bug/feature there in detail

### DIFF
--- a/docs/en/administrator-guide/variables.md
+++ b/docs/en/administrator-guide/variables.md
@@ -73,6 +73,7 @@ Variables that support both session-level and global-level setting include:
 * `batch_size`
 * `parallel_fragment_exec_instance_num`
 * `parallel_exchange_instance_num`
+* `allow_partition_column_nullable`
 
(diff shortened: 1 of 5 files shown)


morningman (approved):
LGTM

PR #5028, opened 2020-12-05
Revert "[Refactor] Refactor DeleteHandler and Cond module (#4925)"

morningman:
This reverts commit 9c9992e0aa28ee85364eebf86a6675f1073e08fb.

## Proposed changes

The PR #4925 will cause BE crash by following case:

```
CREATE TABLE `tbl1` (
  `k1` int(11) NULL COMMENT "",
  `k2` int(11) NULL COMMENT ""
) ENGINE=OLAP
DUPLICATE KEY(`k1`, `k2`)
COMMENT "OLAP"
PROPERTIES (
"replication_num" = "1",
"in_memory" = "false",
"storage_format" = "V2"
);

insert into tbl1 values(1,2);
delete from tbl1 where k1=1;
alter table tbl1 modify column k2 bigint key;
```

BE will crash when doing schema change with delete condition.

## Types of changes

### DIFF
--- a/be/src/olap/collect_iterator.cpp
+++ b/be/src/olap/collect_iterator.cpp
@@ -162,7 +162,7 @@ const RowCursor* CollectIterator::Level0Iterator::current_row() const {
     return _current_row;
 }
 
-int64_t CollectIterator::Level0Iterator::version() const {
+int32_t CollectIterator::Level0Iterator::version() const {
     return _rs_reader->version().second;
 }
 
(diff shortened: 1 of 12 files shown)


acelyc111:
Sorry about that, I'll try to fix it this weekend. Anyway, you can revert it before I fix it.

EmmyMiao87 (approved):
LGTM

PR #4863, opened 2020-11-07
[New Feature]Support udf  when loading data

xinghuayu007:
## Proposed changes

Many time, our users want to use UDFs they developed to ETL the data when loading the data into Doris.But currently, `broker load` does not support to use UDF. As UDF belongs to a database, it needs to check whether it has the `SELECT` permission of the database. This patch try to solve this problem. 

## Types of changes

What types of changes does your code introduce to Doris?
_Put an `x` in the boxes that apply_

- [] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [] Documentation Update (if none of the other choices apply)
- [] Code refactor (Modify the code structure, format the code, etc...)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [] I have create an issue on (Fix #ISSUE), and have described the bug/feature there in detail
- [x] Compiling and unit tests pass locally with my changes
- [] I have added tests that prove my fix is effective or that my feature works
- [x] If this change need a document change, I have updated the document
- [x] Any dependent changes have been merged

## Further comments

If this is a relatively large or complex change, kick off the discussion at dev@doris.apache.org by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #5029, opened 2020-12-05
[Bug] Fix bug that failed to write meta image of load job

morningman:
## Proposed changes

In #4863, we add userInfo in load job, but the userInfo must be analyzed
so that it can be written to the image.

## Types of changes


yangzhg (approved):
+1

PR #5034, opened 2020-12-07
[Bug] Add a flag to prevent repeated close operation of OlapTabletSink

morningman:
## Proposed changes

The close method of OlapTabletSink may be called twice.
In the open_internal() method of plan_fragment_executor, close is called once.
If an error occurs in this call, it will be called again in fragment_mgr.
So here we use a flag to prevent repeated close operations.

## Types of changes


yangzhg (approved):
LGTM

PR #5052, opened 2020-12-09
[Enhance] Add profile for load job

morningman:
## Proposed changes

Add viewable profile for broker load. Similar to the query profile,
the user can submit the import job by setting the session variable `is_report_success` to true,
and then view the running profile of the job on the FE web page for easy analysis and debugging.

## Types of changes


fe/fe-core/src/test/java/org/apache/doris/load/loadv2/BrokerLoadJobTest.java
```diff
@@ -17,22 +17,29 @@
 
 package org.apache.doris.load.loadv2;
 
+import mockit.Expectations;
+import mockit.Injectable;
+import mockit.Mock;
```
EmmyMiao87, line 22: Maybe it should under the java.*

EmmyMiao87 (approved):
LGTM

PR #5055, opened 2020-12-10
[enhancement]improve performance of json load

### LINKED ISSUE #5054: improve performance of json load
There are two points that affect performance in the current json load:
1 json is actually a traversal operation for searching by key. We call HasMember and operator[] twice, and need to traverse twice
2 If you use a string as the key to find the json object, strlen will be called every time, so the number of calls to strlen is the row  number * column number

yangzhg:
## Proposed changes

There are two points that affect performance in the current json load:
1 json is actually a traversal operation for searching by key. We call HasMember and operator[] twice, and need to traverse twice
2 If you use a string as the key to find the json object, strlen will be called every time, so the number of calls to strlen is the row number * column number

## Types of changes

## Checklist

- [] I have create an issue on (Fix #5054), and have described the bug/feature there in detail

## Further comments

### DIFF
--- a/be/src/exec/json_scanner.cpp
+++ b/be/src/exec/json_scanner.cpp
@@ -54,7 +54,6 @@ Status JsonScanner::open() {
 }
 
 Status JsonScanner::get_next(Tuple* tuple, MemPool* tuple_pool, bool* eof) {
-    SCOPED_TIMER(_read_timer);
     // Get one line
     while (!_scanner_eof) {
         if (_cur_file_reader == nullptr || _cur_file_eof) {
@@ -152,7 +151,8 @@ Status JsonScanner::open_next_reader() {
     if (range.__isset.num_as_string) {
         num_as_string = range.num_as_string;
     }
-    _cur_file_reader = new JsonReader(_state, _counter, _profile, file, strip_outer_array, num_as_string);
+    _cur_file_reader =
+            new JsonReader(_state, _counter, _profile, file, strip_outer_array, num_as_string);
     RETURN_IF_ERROR(_cur_file_reader->init(jsonpath, json_root));
 
     return Status::OK();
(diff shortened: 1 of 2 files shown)


be/src/exec/json_scanner.cpp
```diff
@@ -54,7 +54,6 @@ Status JsonScanner::open() {
 }
 
 Status JsonScanner::get_next(Tuple* tuple, MemPool* tuple_pool, bool* eof) {
-    SCOPED_TIMER(_read_timer);
     // Get one line
```
HappenLee, line 57: Why delete the reader timer？

be/src/exec/json_scanner.cpp
```diff
@@ -54,7 +54,6 @@ Status JsonScanner::open() {
 }
 
 Status JsonScanner::get_next(Tuple* tuple, MemPool* tuple_pool, bool* eof) {
-    SCOPED_TIMER(_read_timer);
```
HappenLee, line 57: why delete this read_timer? Maybe should keep the same with orc scanner

be/src/exec/json_scanner.cpp
```diff
@@ -273,9 +269,11 @@ void JsonReader::_close() {
 // return Status::OK() if parse succeed or reach EOF.
 Status JsonReader::_parse_json_doc(bool* eof) {
     // read a whole message, must be delete json_str by `delete[]`
+    SCOPED_TIMER(_read_timer);
```
HappenLee, line 272: Add a new timer name 'parse json timer' to track the parse json time.

HappenLee (changes requested):
should revise some code of timer

HappenLee (approved):

morningman (approved):
LGTM

PR #5067, opened 2020-12-11
[Bug] Fix bug that query multi mysql external table with union will get incomplete result

### LINKED ISSUE #5066: [Bug] Query external table with UNION get incomplete result.
**Describe the bug**
How to reproduce:

```
CREATE EXTERNAL TABLE db1.tbl1
(
k1 INT
)
ENGINE=mysql
PROPERTIES
(
"host" = "xxxx",
"port" = "8777",
"user" = "root",
"password" = "xxx",
"database" = "cmytest",
"table" = "tbl1"
);

CREATE EXTERNAL TABLE db1.tbl2
(
k1 INT
)
ENGINE=mysql
PROPERTIES
(
"host" = "xxxx",
"port" = "8777",
"user" = "root",
"password" = "xxxx",
"database" = "cmytest",
"table" = "tbl1"
);

CREATE EXTERNAL TABLE db1.tbl3
(
k1 INT
)
ENGINE=mysql
PROPERTIES
(
"host" = "xxxx",
"port" = "8777",
"user" = "root",
"password" = "xxxx",
"database" = "cmytest",
"table" = "tbl1"
);

create view v1 as select * from tbl1 union all select * from tbl2 union all select * from tbl3;
```
And query like:
```
select count(*) from v1;
```

if there are more than 1024 rows in mysql table, the query will return less than 3000.

 - Version [0.13.11]

morningman:
## Proposed changes

The `eos` flag should be reset to false after opening next child of union node.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #5066 ), and have described the bug/feature there in detail

### DIFF
--- a/be/src/exec/union_node.cpp
+++ b/be/src/exec/union_node.cpp
@@ -159,7 +159,10 @@ Status UnionNode::get_next_materialized(RuntimeState* state, RowBatch* row_batch
             _child_row_idx = 0;
             // open the current child unless it's the first child, which was already opened in
             // UnionNode::open().
-            if (_child_eos) RETURN_IF_ERROR(child(_child_idx)->open(state));
+            if (_child_eos) {
+                RETURN_IF_ERROR(child(_child_idx)->open(state));
+                _child_eos = false;
+            }
             // The first batch from each child is always fetched here.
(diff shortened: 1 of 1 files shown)


EmmyMiao87 (approved):
LGTM

PR #5077, opened 2020-12-14
[UT] cleanup storage engine creation in tablet_mgr_test etc

vagetablechicken:
## Proposed changes

Mistakenly use the string '_engine_data_path' as the path, actually the storage engine is not open, so option/path is needless. Cleaup it to avoid any doubt about the file path management.

## Types of changes

## Checklist


morningman (approved):
LGTM

PR #5085, opened 2020-12-15
[UT] fix test_env & add a sample

### LINKED ISSUE #5062: [UT] TestEnv needs fix
It's good to create a TestEnv for unit tests of be. But the current TestEnv can't work. 
And it doesn't need to be a friend of RuntimeState or any other class, because the unit tests can access private members.
https://github.com/apache/incubator-doris/blob/6fedf5881be01363d3a246e07c79ffb417574d40/be/src/runtime/runtime_state.h#L365

So I want to move it to the `be/test/` folder, and fix it.

vagetablechicken:
## Proposed changes

Easily create tests.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #5062), and have described the bug/feature there in detail


morningman (approved):
LGTM

PR #5093, opened 2020-12-16
[Bug] Fix bug that routine load may lost some data

### LINKED ISSUE #5095: [Bug] Routine load lose data
**Describe the bug**
When the cluster load is high, the routine load will lose some data of kafka.

 - Version [0.13.12]

morningman:
## Proposed changes

In the previous implementation, whether a subtask is in commit or abort state,
we will try to update the job progress, such as the consumed offset of kafka.
Under normal circumstances, the aborted transaction does not consume any data,
and all progress is 0, so even we update the progress, the progress will remain
unchanged.
However, in the case of high cluster load, the subtask may fail half of the execution on the BE side.
At this time, although the task is aborted, part of the progress is updated.
Cause the next subtask to skip these data for consumption, resulting in data loss.

Bad case:
LoadedRows = 0;
comittedOffset > 0
txn: aborted

Also modify some log level and config

## Types of changes

## Checklist

- [] I have create an issue on (Fix #5095 ), and have described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/load/routineload/KafkaProgress.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/routineload/KafkaProgress.java
@@ -128,25 +128,6 @@ public class KafkaProgress extends RoutineLoadProgress {
         }
     }
 
-    // check if the given progress has same partitions with this one
-    public boolean isPartitionMatched(KafkaProgress otherProgress) {
-        return this.partitionIdToOffset.keySet().equals(otherProgress.partitionIdToOffset.keySet());
-    }
-
-    // return the offset diff of this progress and the given progress.
-    // the partition in newProgress must be in this progress.
-    public long offsetDiff(KafkaProgress newProgress) {
-        long diff = 0;
-        for (Map.Entry<Integer, Long> entry : newProgress.partitionIdToOffset.entrySet()) {
-            diff += (entry.getValue() - getOffsetByPartition(entry.getKey()));
-        }
-        return diff;
-    }
-
-    public int getPartitionNum() {
-        return this.partitionIdToOffset.keySet().size();
-    }
-
     @Override
     public String toString() {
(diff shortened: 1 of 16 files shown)


fe/fe-core/src/main/java/org/apache/doris/load/routineload/KafkaProgress.java
```diff
@@ -38,6 +42,8 @@
  */
 // {"partitionIdToOffset": {}}
 public class KafkaProgress extends RoutineLoadProgress {
+    private static final Logger LOG = LogManager.getLogger(KafkaProgress.class);
```
yangzhg, line 45:
```suggestion
    private static final Logger LOG = LogManager.getLogger(RoutineLoadProgress.class);
```

fe/fe-core/src/main/java/org/apache/doris/load/routineload/RoutineLoadJob.java
```diff
@@ -1049,8 +1053,9 @@ private void executeTaskOnTxnStatusChanged(RoutineLoadTaskInfo routineLoadTaskIn
                                           + " maybe task was aborted by master when timeout")
                                   .build());
             }
-        } else if (checkCommitInfo(rlTaskTxnCommitAttachment, txnState.getTransactionStatus())) {
+        } else if (checkCommitInfo(rlTaskTxnCommitAttachment, txnState)) {
             // step2: update job progress
+            // only committed task need to update progress
```
EmmyMiao87, line 1058: error commit...

EmmyMiao87 (dismissed):
LGTM

hf200012 (commented):
After the patch is applied, there is no data loss problem so far.

EmmyMiao87 (dismissed):
LGTM

EmmyMiao87 (approved):
LGTM

PR #5104, opened 2020-12-17
[Enhancement]Make Cholocate table join more load balance

xinghuayu007:
## Proposed changes

When two cholocate tables make `join` operation, to make `join` operation locally, the tablet belongs to the same bucket sequence will be distributed to the same host. When chosing which host for a bucket sequence, it takes random strategy. Random strategy can not make query task load balance logically for one query. Therefore, this patch takes round-robin strategy, make buckets distributed evenly. For example, if there are 6 bucket sequences and 3 hosts, it is better to distributed 2 buket sequence for every host. 

## Types of changes

## Checklist

## Further comments


fe/fe-core/src/main/java/org/apache/doris/qe/Coordinator.java
```diff
@@ -1276,10 +1276,22 @@ private void computeScanRangeAssignmentByColocate(
     }
 
     // randomly choose a backend from the TScanRangeLocations for a certain bucket sequence.
```
kangkaisen, line 1278: Please update the old comment

fe/fe-core/src/main/java/org/apache/doris/qe/Coordinator.java
```diff
@@ -1276,10 +1276,22 @@ private void computeScanRangeAssignmentByColocate(
...
-    private void getExecHostPortForFragmentIDAndBucketSeq(TScanRangeLocations seqLocation, PlanFragmentId fragmentId, Integer bucketSeq) throws Exception {
-        int randomLocation = new Random().nextInt(seqLocation.locations.size());
+    private void getExecHostPortForFragmentIDAndBucketSeq(TScanRangeLocations seqLocation, PlanFragmentId fragmentId, Integer bucketSeq,
+                                                          HashMap<TNetworkAddress, Long> assignedBytesPerHost) throws Exception {
+        Long minAssignedBytes = Long.MAX_VALUE;
```
kangkaisen, line 1281: These codes are also used in `computeScanRangeAssignmentByScheduler`, would better abstract these codes to a method
```
            Long minAssignedBytes = Long.MAX_VALUE;
            TScanRangeLocation minLocation = null;
            for (final TScanRangeLocation location : scanRangeLocations.getLocations()) {
                Long assignedBytes = findOrInsert(assignedBytesPerHost, location.server, 0L);
                if (assignedBytes < minAssignedBytes) {
                    minAssignedBytes = assignedBytes;
                    minLocation = location;
                }
            }
```

xinghuayu007: Thanks for your code review. The code has been improved.

fe/fe-core/src/main/java/org/apache/doris/qe/SimpleScheduler.java
```diff
@@ -92,6 +92,38 @@ public static TNetworkAddress getHost(long backendId,
                         backends, locations.size()));
     }
 
+    public static TScanRangeLocation getLocation(TScanRangeLocation minLocation,
```
kangkaisen, line 95: Why add this method?  The diff between `getLocation` and `getHost` is first arg, but we could get `long backendId` from `TScanRangeLocation minLocation`?

fe/fe-core/src/main/java/org/apache/doris/qe/Coordinator.java
```diff
@@ -1275,15 +1275,40 @@ private void computeScanRangeAssignmentByColocate(
...
         Reference<Long> backendIdRef = new Reference<Long>();
-        TNetworkAddress execHostPort = SimpleScheduler.getHost(seqLocation.locations.get(randomLocation).backend_id, seqLocation.locations, this.idToBackend, backendIdRef);
+        distributeHost(seqLocation, assignedBytesPerHost, backendIdRef);
+        Backend backend = this.idToBackend.get(backendIdRef.getRef());
+        TNetworkAddress execHostPort = new TNetworkAddress(backend.getHost(), backend.getBePort());
         this.addressToBackendID.put(execHostPort, backendIdRef.getRef());
         this.fragmentIdToSeqToAddressMap.get(fragmentId).put(bucketSeq, execHostPort);
     }
 
+    public TScanRangeLocation distributeHost(TScanRangeLocations seqLocation,
```
kangkaisen, line 1289: I think we need a more concrete name. Such as `selectBackendsByRoundRobin`

fe/fe-core/src/main/java/org/apache/doris/qe/SimpleScheduler.java
xinghuayu007: Because function `computeScanRangeAssignmentByScheduler` need `minLocation` to set volume_id. Therefore it must return param `minLocation`

// Volume is optional, so we need to set the value and the is-set bit
scanRangeParams.setVolumeId(minLocation.volume_id);

fe/fe-core/src/main/java/org/apache/doris/qe/Coordinator.java
xinghuayu007: ok. I will modify it.

kangkaisen (approved):
+1

PR #5111, opened 2020-12-22
[Bug] Fix scanner threads heap-use-after-free

### LINKED ISSUE #5102: [Bug] scanner threads has the risk of 'heap-use-after-free'
**Describe the bug**
In OlapScanNode, the scanner threads is running in `state->exec_env()->thread_pool();`. So we can't control the scanner threads directly, and their destructions.
When we close `OlapScanNode`, we only wait for `_transfer_thread`.
https://github.com/apache/incubator-doris/blob/81c7c0360e61a6ad5ab4221f3a687d031f548e50/be/src/exec/olap_scan_node.cpp#L314
But we don't wait for the scanner threads, it has a risk:
**scanner threads may be running and using the member vars of OlapScanNode, when the OlapScanNode has already destroyed.**

**To Reproduce**
To illustrate the risk, we can add a sleep before scanner threads exit. It's an unlikely scenario, just for simulation.
https://github.com/apache/incubator-doris/blob/81c7c0360e61a6ad5ab4221f3a687d031f548e50/be/src/exec/olap_scan_node.cpp#L1354-L1361

For example:
```
    // NOTICE: scanner thread live some more
    SleepFor(MonoDelta::FromSeconds(5));
    _scan_batch_added_cv.notify_one(); 
```
ASAN report:
```
==64370==ERROR: AddressSanitizer: heap-use-after-free on address 0x61c0000706c8 at pc 0x000003106ab4 bp 0x7f8aff3cfcb0 sp 0x7f8aff3cfca8
#0 0x3106ab3 in doris::OlapScanNode::scanner_thread(doris::OlapScanner*) /root/doris-env/doris/be/src/exec/olap_scan_node.cpp:1345
...
0x61c0000706c8 is located 1608 bytes inside of 1912-byte region [0x61c000070080,0x61c0000707f8)
freed by thread T198 (FragmentMgrThre) here:
 #0 0x1822940 in operator delete(void*) ../../.././libsanitizer/asan/asan_new_delete.cc:124
#1 0x30f3c19 in doris::OlapScanNode::~OlapScanNode() /root/doris-env/doris/be/src/exec/olap_scan_node.cpp:69
```

**Expected behavior**
Maybe we can use `_running_thread` to be the last accessed member variable. And OlapScanNode should wait for `_running_thread==0`.

vagetablechicken:
## Proposed changes

Scanner threads may be running and using the member vars of OlapScanNode, when the OlapScanNode has already destroyed.

We can use `_running_thread` to be the last accessed member variable. And `transfer_thread` need to wait for `_running_thread==0`. After `transfer_thread` joined, `OlapScanNode::close()` can continue. 

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #5102), and have described the bug/feature there in detail

### DIFF
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -59,7 +59,6 @@ OlapScanNode::OlapScanNode(ObjectPool* pool, const TPlanNode& tnode, const Descr
           _status(Status::OK()),
           _resource_info(nullptr),
           _buffered_bytes(0),
-          _running_thread(0),
           _eval_conjuncts_fn(nullptr) {}
 
 OlapScanNode::~OlapScanNode() {}
@@ -1257,11 +1256,7 @@ void OlapScanNode::transfer_thread(RuntimeState* state) {
     _row_batch_added_cv.notify_all();
 
     std::unique_lock<std::mutex> l(_scan_batches_lock);
-    while (_running_thread != 0) {
-        l.unlock();
-        SleepFor(MonoDelta::FromMilliseconds(5));
-        l.lock();
-    }
+    _scan_thread_exit_cv.wait(l, [this] { return _running_thread == 0; });
     VLOG(1) << "Scanner threads have been exited. TransferThread exit.";
 }
 
@@ -1378,6 +1373,7 @@ void OlapScanNode::scanner_thread(OlapScanner* scanner) {
     // Do not access class members after this code.
     std::unique_lock<std::mutex> l(_scan_batches_lock);
     _running_thread--;
+    _scan_thread_exit_cv.notify_one()
 }
 
 Status OlapScanNode::add_one_batch(RowBatchInterface* row_batch) {
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -1252,8 +1252,8 @@ void OlapScanNode::transfer_thread(RuntimeState* state) {
     {
         std::unique_lock<std::mutex> l(_row_batches_lock);
         _transfer_done = true;
+        _row_batch_added_cv.notify_all();
     }
-    _row_batch_added_cv.notify_all();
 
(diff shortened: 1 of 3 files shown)


be/src/exec/olap_scan_node.cpp
```diff
@@ -1250,12 +1250,25 @@ void OlapScanNode::transfer_thread(RuntimeState* state) {
...
-    std::unique_lock<std::mutex> l(_row_batches_lock);
-    _transfer_done = true;
+    {
+        std::unique_lock<std::mutex> l(_row_batches_lock);
+        _transfer_done = true;
+    }
     _row_batch_added_cv.notify_all();
+
+    std::unique_lock<std::mutex> l(_scan_batches_lock);
+    while (_running_thread != 0) {
```
caiconghui, line 1260: Is it better to use C++11 promise and future to instead of checking _running_thread == 0? every scanner_thread just need to set its promise value when finish task with lock free, and close OlapScanNode close just need to wait all scanner_thread return future result ?

be/src/exec/olap_scan_node.cpp
```diff
@@ -1250,12 +1250,25 @@ void OlapScanNode::transfer_thread(RuntimeState* state) {
...
+        l.unlock();
+        SleepFor(MonoDelta::FromMilliseconds(5));
+        l.lock();
```
acelyc111, line 1263: Is it elegant to use sleep to do multithread sync? why not use a cv to wait here, and nofity after _running_thread--?

vagetablechicken: I don't want use `_scan_batch_added_cv`, it's hard to understand the relationship between `_scan_batch_added_cv` & `_running_thread `. I'll add a new condition variable.

acelyc111 (dismissed):
LGTM

morningman (dismissed):
LGTM

morningman (approved):
LGTM

PR #5143, opened 2020-12-24
[Optimize][Compaction] Add candidate tablets mechanism for compaction producer

### LINKED ISSUE #4988: [Proposal] Avoid traversing all tablets when selecting a tablet for compaction task
The current tablet selection strategy for compaction task is to traverse all tablets and then find a tablet with the highest score for compaction task. This mechanism is expensive to select a tablet for compaction task.

Can we hold a `heap`with specified number of candidate tablets for each disk?

 Heap elements are sorted by score and the top tablet in `heap`hold the highest score. Top element in `heap` will be selected when producer generates compaction task for this disk. After a tablet is selected for compaction, `pop`for this tablet is unnecessary but `update and sort heap` is needed. 

A new tablet will be pushed into the `heap` when the following occurs：
（1）`load` operation on the tablet is performed;
（2）`scan`operation on the tablet is performed for `n`times and `n` could be configurable.
To keep the number of tablets in the `heap` constant, the last tablet will pop from `heap`  when a new tablet is pushed into `heap`.

It seems that this mechanism can avoid traversing all tablets when selecting a tablet for compaction task. Of course, some thread-safety problems need to be addressed.

weizuo93:
## Proposed changes

The current tablet selection strategy for compaction task is to traverse all tablets and then find a tablet with the highest score for compaction task. This mechanism is expensive to select a tablet for compaction task.

This patch adds a `candidate tablets mechanism` for compaction producer which is described in ISSUE #4988 .

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #4988 ), and have described the bug/feature there in detail


be/src/exec/olap_scanner.cpp
```diff
@@ -538,6 +538,14 @@ Status OlapScanner::close(RuntimeState* state) {
     _reader.reset();
     Expr::close(_conjunct_ctxs, state);
     _is_closed = true;
+    if (config::scan_count_push_tablet_into_compaction_heap != 0 &&
+        _tablet->query_scan_count->value() % config::scan_count_push_tablet_into_compaction_heap ==
+                0) {
+        _tablet->data_dir()->push_tablet_into_compaction_heap(CompactionType::BASE_COMPACTION,
```
morningman, line 544: Have you test the performance of `push_tablet_into_compaction_heap`.
This is on the critical path of READ, and the default `scan_count_push_tablet_into_compaction_heap` is 1,
so the tablet will call this method every time it being read.

If the query concurrency is high, i doubt it may impact the performance.

morningman (commented):
I have a question about how to deal with the tablet which being sacrifice and popped out of the candidates vector?

For example:
there are 3 tablet is vector with compaction score:
[10, 9, 8]

And then a tablet with score 11 is pushed to the vector, so the tablet with score 8 will be
sacrificed. Finally the vector remains:
[11, 10, 9]

be/src/olap/data_dir.cpp
```diff
@@ -1018,4 +1018,163 @@ void DataDir::disks_compaction_score_increment(int64_t delta) {
...
+    OlapStopWatch watch;
+    if (compaction_type == CompactionType::BASE_COMPACTION) {
+        std::unique_lock<std::mutex> lock(_base_compaction_heap_mutex);
+        std::vector<TabletSharedPtr>::iterator it =
+                find(_base_compaction_heap.begin(), _base_compaction_heap.end(), tablet);
+        if (it == _base_compaction_heap.end()) {
+            _base_compaction_heap.push_back(tablet);
+        }
+        std::make_heap(_base_compaction_heap.begin(), _base_compaction_heap.end(),
+                       TabletScoreComparator(CompactionType::BASE_COMPACTION));
```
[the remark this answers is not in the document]
weizuo93: @morningman 
Because the `priority_queue` only supports pushing element through the rear of queue and popping element through the head of queue, but we need to pop an element through the `queue head` when selecting a tablet for compaction and pop an element through the `queue rear` when sacrificing a tablet. Thus `priority_queue`is not enough.

weizuo93:
> I have a question about how to deal with the tablet which being sacrifice and popped out of the candidates vector?
> 
> For example:
> there are 3 tablet is vector with compaction score:
> [10, 9, 8]
> 
> And then a tablet with score 11 is pushed to the vector, so the tablet with score 8 will be
> sacrificed. Finally the vector remains:
> [11, 10, 9]

@morningman 
The sacrificed tablet will be `pop_back` from underlying container vector because tablets in vector are sorted by score.

be/src/exec/olap_scanner.cpp
weizuo93: @morningman 
The default max size of compaction heap is 20. After test, the cost of `push_tablet_into_compaction_heap`for each tablet is `0.05 ms` averagely.

weizuo93: The default `scan_count_push_tablet_into_compaction_heap` is 1 and this value can be set dynamically.

PR #5148, opened 2020-12-25
[Feature] Implementation of histogram metric

### LINKED ISSUE #5146: [Proposal] Add Histogram Metrics Implementation
Currently, Doris supports three types of metrics, which are Atomic, LockSimple and CoreLocalCounter. All of them are not quite suitable for statistics analysis. 
Histogram metrics is already in used in many DBMS like kudu, rocksdb and so on. It measures the distribution of values in a stream of data and allow you to measure not just easy things like the min, mean, max, and standard deviation of values, but also quantiles like the median or 95th percentile.
As metrics for statistics analysis, the following interface should be supported:
- `add(uint64_t value)`
- `min()`, `max()`, `num()`, `sum()`
- `average()`, `median()`, `percentile(double p)`, `standard_deviation()`

Skysheepwang:
## Proposed changes

#5146
Add histogram metrics into `util/metrics.h`. The data structure of histogram is implemented in `util/histogram.h`, which could also be used in other situations that in need of histogram. Unit testis added as well.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #5146), and have described the bug/feature there in detail

## Further comments


be/src/util/metrics.h
```diff
@@ -159,6 +160,35 @@ class CoreLocalCounter : public Metric {
     CoreLocalValue<T> _value;
 };
 
+class HistogramMetric : public Metric {
+public:
+    HistogramMetric() { clear(); }
```
acelyc111, line 165: No need to clear() in constructor?

be/src/util/histogram.h
```diff
@@ -0,0 +1,113 @@
...
+#include <string>
+#include <vector>
+#include <map>
+#include <atomic>
+
+#include "common/logging.h"
+
+namespace doris {
+
+// Histogram data structure implementation:
```
acelyc111, line 29: These code are almost copied from rocksdb, we have to update LICENSE.txt

be/src/util/metrics.h
Skysheepwang: Yes, the `_stat.clear()` will be called in `HistogramStat` 's constructor. I will remove it.

acelyc111 (approved):
LGTM

morningman (approved):
LGTM

PR #5156, opened 2020-12-28
[Load] support ignoring eovercrowded when tablet sink

### LINKED ISSUE #5147: [Proposal] support ignoring eovercrowded when adding batch
We met lots of `EOVERCROWDED` brpc error. Unfortunately, it will make load job failed(index intolerable failure).
We could change it by `brpc_socket_max_unwritten_bytes`, but it can't be modified in runtime.
And we've tried to set large numbers, but the `EOVERCROWDED` error still exist, and load job still get failures. To be on the safe side, the config couldn't be larger.
The root cause is not the config, but we must endeavour to accomplish the load job.

So I want to add a more flexible approach.
We can add a mutable config `node_channel_ignore_eovercrowded`. If it's true, we will add the `ignore_eovercrowded` flag to  `_add_batch_closure->cntl`. Thus the `NodeChannel` won't failed on `EOVERCROWDED`. It only effects the `NodeChannel`(the load job), other rpc requests will still check if overcrowded.

vagetablechicken:
## Proposed changes

If adding the ignore_eovercrowded flag, the `PTabletWriterAddBatchRequest` won't failed on `EOVERCROWDED` to avoid load jobs failed in this error. It only effects the NodeChannel(the load job), other rpc requests will still check if overcrowded.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #5147 ), and have described the bug/feature there in detail


be/src/exec/tablet_sink.cpp
```diff
@@ -327,6 +327,9 @@ int NodeChannel::try_send_and_fetch_status() {
 
         _add_batch_closure->reset();
         _add_batch_closure->cntl.set_timeout_ms(_rpc_timeout_ms);
+        if (config::tablet_writer_ignore_eovercrowded) {
+            _add_batch_closure->cntl.ignore_eovercrowded();
```
morningman, line 331: What will happen if OVERCROWDED being ignored?

vagetablechicken: If the controller ignores evorcrowded, this request will be sent even `current_buf > brpc_socket_max_unwritten_bytes`.

kangkaisen:
@vagetablechicken LGTM. please update the doc for the new config.

vagetablechicken:
> @vagetablechicken LGTM. please update the doc for the new config.

Done. And I found opening tablet writer may be failed too. So I add `ignore` option to all the rpc requests in `tablet_sink`.

kangkaisen (dismissed):
+1

kangkaisen:
@vagetablechicken Hi, please fix the conficts.

kangkaisen (approved):

PR #5158, opened 2020-12-28
[Config] change some static config to dynamic config and delete some unused config

weizuo93:
## Proposed changes

1. Change some static config of BE to dynamic config so that we can reset some parameters without restart BE.
2. Delete some unused BE config.

## Types of changes

## Checklist

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -26,15 +26,15 @@ namespace config {
 CONF_String(custom_config_dir, "${DORIS_HOME}/conf");
 
 // cluster id
-CONF_mInt32(cluster_id, "-1");
+CONF_Int32(cluster_id, "-1");
 // port on which BackendService is exported
-CONF_mInt32(be_port, "9060");
+CONF_Int32(be_port, "9060");
 
 // port for brpc
-CONF_mInt32(brpc_port, "8060");
+CONF_Int32(brpc_port, "8060");
 
 // the number of bthreads for brpc, the default value is set to -1, which means the number of bthreads is #cpu-cores
-CONF_mInt32(brpc_num_threads, "-1")
+CONF_Int32(brpc_num_threads, "-1")
 
 // Declare a selection strategy for those servers have many ips.
 // Note that there should at most one ip match this list.
@@ -46,9 +46,9 @@ CONF_String(priority_networks, "");
 //// tcmalloc gc parameter
 ////
 // min memory for TCmalloc, when used memory is smaller than this, do not returned to OS
-CONF_mInt64(tc_use_memory_min, "10737418240");
+CONF_Int64(tc_use_memory_min, "10737418240");
 // free memory rate.[0-100]
-CONF_mInt64(tc_free_memory_rate, "20");
+CONF_Int64(tc_free_memory_rate, "20");
 
 // Bound on the total amount of bytes allocated to thread caches.
 // This bound is not strict, so it is possible for the cache to go over this bound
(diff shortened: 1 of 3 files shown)


caiconghui (dismissed):
+1

be/src/common/config.h
```diff
@@ -256,21 +222,18 @@ CONF_Int32(file_descriptor_cache_capacity, "32768");
 // modify them upon necessity
 CONF_Int32(min_file_descriptor_number, "60000");
 CONF_Int64(index_stream_cache_capacity, "10737418240");
-// CONF_Int64(max_packed_row_block_size, "20971520");
 
 // Cache for storage page size
 CONF_String(storage_page_cache_limit, "20%");
 // whether to disable page cache feature in storage
-CONF_Bool(disable_storage_page_cache, "false");
+CONF_mBool(disable_storage_page_cache, "false");
```
acelyc111, line 229: There is a case that `disable_storage_page_cache ` is true at first, then changed to false, will the memory used by cache has no chance to free?

be/src/common/config.h
```diff
@@ -424,8 +378,7 @@ CONF_Bool(enable_quadratic_probing, "false");
 CONF_String(pprof_profile_dir, "${DORIS_HOME}/log");
 
 // for partition
-// CONF_Bool(enable_partitioned_hash_join, "false")
-CONF_Bool(enable_partitioned_aggregation, "true");
+CONF_mBool(enable_partitioned_aggregation, "true");
```
acelyc111, line 381: There is a memory leak risk, `enable_partitioned_aggregation` is true at first and then changed to false, memory allocated has no chance to free.

be/src/common/config.h
```diff
@@ -472,7 +412,7 @@ CONF_String(buffer_pool_clean_pages_limit, "20G");
 CONF_mInt64(memory_maintenance_sleep_time_s, "10");
 
 // Alignment
-CONF_Int32(memory_max_alignment, "16");
+CONF_mInt32(memory_max_alignment, "16");
```
acelyc111, line 415: I think you can remove this config, and define it as a static const value, there is no reason to update it.

be/src/common/config.h
```diff
@@ -482,8 +422,8 @@ CONF_mInt64(write_buffer_size, "104857600");
 // NOTICE(cmy): set these default values very large because we don't want to
 // impact the load performance when user upgrading Doris.
 // user should set these configs properly if necessary.
-CONF_Int64(load_process_max_memory_limit_bytes, "107374182400"); // 100GB
-CONF_Int32(load_process_max_memory_limit_percent, "80");         // 80%
+CONF_mInt64(load_process_max_memory_limit_bytes, "107374182400"); // 100GB
+CONF_mInt32(load_process_max_memory_limit_percent, "80");         // 80%
```
acelyc111, line 426: These 2 configs are only changed in LoadChannelMgr's init function, which will only be called only once when server start, they are immutable.

be/src/common/config.h
weizuo93: The config `disable_storage_page_cache`is not used to close cache but to choose whether to use the cache during read.  So the memory used by cache will not be freed.

be/src/common/config.h
weizuo93: OK.

be/src/common/config.h
weizuo93: OK.

morningman (approved):
LGTM

PR #5160, opened 2020-12-28
[BUG] Follower shouldn't forward non-query statement to master repeatedly

924060929:
## Proposed changes
Follower shouldn't forward non-query statement to master repeatedly

## Types of changes

## Checklist

## Further comments
None

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/MasterOpExecutor.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/MasterOpExecutor.java
@@ -42,7 +42,9 @@ public class MasterOpExecutor {
     // the total time of thrift connectTime add readTime and writeTime
     private int thriftTimeoutMs;
 
-    public MasterOpExecutor(OriginStatement originStmt, ConnectContext ctx, RedirectStatus status) {
+    private boolean shouldNotRetry;
+
+    public MasterOpExecutor(OriginStatement originStmt, ConnectContext ctx, RedirectStatus status, boolean isQuery) {
         this.originStmt = originStmt;
         this.ctx = ctx;
         if (status.isNeedToWaitJournalSync()) {
@@ -51,6 +53,8 @@ public class MasterOpExecutor {
             this.waitTimeoutMs = 0;
         }
         this.thriftTimeoutMs = ctx.getSessionVariable().getQueryTimeoutS() * 1000;
+        // if isQuery=false, we shouldn't retry twice when catch exception because of Idempotency
+        this.shouldNotRetry = !isQuery;
     }
 
     public void execute() throws Exception {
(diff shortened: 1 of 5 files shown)


caiconghui:
plz add an issue related to this PR

924060929:
> plz add an issue related to this PR

done, see #5161

caiconghui:
> > plz add an issue related to this PR
> 
> done, see #5161
what exception will transfer operation throw, when network problem occur and it it not a timeout exception?

924060929:
> > > plz add an issue related to this PR
> > 
> > 
> > done, see #5161
> > what exception will transfer operation throw, when network problem occur and it it not a timeout exception?

TTransportException.UNKNOWN

fe/fe-core/src/main/java/org/apache/doris/qe/MasterOpExecutor.java
```diff
@@ -42,7 +42,9 @@
     // the total time of thrift connectTime add readTime and writeTime
     private int thriftTimeoutMs;
 
-    public MasterOpExecutor(OriginStatement originStmt, ConnectContext ctx, RedirectStatus status) {
+    private boolean shouldNotRetry;
+
+    public MasterOpExecutor(OriginStatement originStmt, ConnectContext ctx, RedirectStatus status, boolean isQuery) {
```
caiconghui, line 47: isQuery rename to shouldRetry? and variable 'shouldNotRetry' in MasterOpExecutor rename to "shouldRetry"? I think it is more readable. and then we just need to think when need to retry

924060929: I was discussed this name with my colleague, the constructor of MasterOpExecutor expose 'shouldRetry' could lead to misuse, because could not remember the retry related to statement type when invoke `new MasterOpExecutor()`.

Although `isQuery` doesn't look good, but it always write collect.

And `shouldNotRetry` means exactly that: we should not retry forward to master

caiconghui: If we want retry a statement, just need it is a idempotent operation, not just query, right?

924060929: yes，the condition isQuery is conservative & convenient

924060929: the principle is：we must ensure idempotent operation execute at-most-once

caiconghui: Thank You, I see

caiconghui (approved):
LGTM

PR #5163, opened 2020-12-28
[Load Parallel][2/3] Support parallel flushing memtable during load

morningman:
## Proposed changes

In the previous implementation, in an load job,
multiple memtables of the same tablet are written to disk sequentially.
In fact, multiple memtables can be written out of order in parallel,
only need to ensure that each memtable uses a dependent segment writer.

This is 2nd part of improving load performance.

#5281

## Types of changes


yangzhg (approved):
LGTM

PR #5168, opened 2020-12-29
[FE] Add log for abort transaction.

killxdcj:
## Proposed changes

Currently there is no abort transaction related log. When troubleshooting transaction failure related problems, we need to use some debug tools to find the failed transaction, so we'd better add a log when abort transaction

## Types of changes

## Checklist


morningman (approved):
LGTM

PR #5193, opened 2021-01-05
[UT] fix the bug of getting current running dir

### LINKED ISSUE #5182: [UT] plugin_zip_test gets failed occasionally
**Describe the bug**

```
[ RUN      ] PluginZipTest.http_normal
the path: /root/doris-env/doris/be/ut_build_ASAN/test/plugin
ASAN:DEADLYSIGNAL
=================================================================
==119381==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f00020b70a0 bp 0x000000000400 sp 0x7effff2fcd90 T2)
==119381==The signal is caused by a READ memory access.
==119381==Hint: address points to the zero page.
    #0 0x7f00020b709f in _IO_fread (/lib64/libc.so.6+0x6f09f)
    #1 0xa4011b in doris::HttpTestHandler::handle(doris::HttpRequest*) (/root/doris-env/doris/be/ut_build_ASAN/test/plugin/plugin_zip_test+0xa4011b)
```

**To Reproduce**
It's occasionally, but we can make a demo. The root cause is https://github.com/apache/incubator-doris/blob/44325ae850d0cda8f2675af5b6825acb246fb988/be/test/plugin/plugin_zip_test.cpp#L54

We can speculate that `fp==nullptr`, so the demo:
```
#include <cstdio>

int main(){
    FILE* fp = nullptr;
    char f[1024];
    auto size = fread(f, 1, 1024, fp);
    return 0;
}
```
asan it, will get report:
```
ASAN:DEADLYSIGNAL
=================================================================
==32155==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7fb8a3fad1ca bp 0x7fff8ee18970 sp 0x7fff8ee18530 T0)
==32155==The signal is caused by a READ memory access.
==32155==Hint: address points to the zero page.
    #0 0x7fb8a3fad1c9 in _IO_fread (/lib/x86_64-linux-gnu/libc.so.6+0x6e1c9)
    #1 0x400649 in main (/home/hw/test/test_fread/a.out+0x400649)
    #2 0x7fb8a3f5f83f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2083f)
    #3 0x400568 in _start (/home/hw/test/test_fread/a.out+0x400568)
```

**Expected behavior**
It's the fault of opening the source file, not the evHttpServer worker. 
And why did we open file failed?
It's the wrong way to get the `dir_path`.
We can build a demo to test:
```
int main(){
        char buf[1024];
        readlink("/proc/self/exe", buf, 1023);
        char* dir_path = dirname(buf);
        std::string path = std::string(dir_path);
        std::cout << path << std::endl;
        return 0;
}
```
After run it multiple times, we can find some wrong results, e.g.
```
/root/doris-env/test/get_path/a.out^O
/root/doris-env/test/get_path/a.out~Y
/root/doris-env/test/get_path/a.out
...
```

So when the path is incorrect, we can't open file in `doris::HttpTestHandler::handle`, the plugin_zip_test will failed.

I'll fix the way to get the folder containing the currently running executable.

vagetablechicken:
## Proposed changes

Fixed the logic after `readlink`, add a test_util function `GetCurrentRunningDir()`.

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #5182), and have described the bug/feature there in detail

### DIFF
--- a/be/test/plugin/plugin_loader_test.cpp
+++ b/be/test/plugin/plugin_loader_test.cpp
@@ -18,9 +18,9 @@
 #include "plugin/plugin_loader.h"
 
 #include <gtest/gtest.h>
-#include <libgen.h>
 
 #include "plugin/plugin.h"
+#include "test_util/test_util.h"
 #include "util/file_utils.h"
 
 namespace doris {
@@ -55,10 +55,8 @@ int close_plugin(void* ptr) {
 class PluginLoaderTest : public testing::Test {
 public:
     PluginLoaderTest() {
-        char buf[1024];
-        readlink("/proc/self/exe", buf, 1023);
-        char* dir_path = dirname(buf);
-        _path = std::string(dir_path);
+        _path = GetCurrentRunningDir();
+        EXPECT_FALSE(_path.empty());
     }
 
     ~PluginLoaderTest() {}
--- a/be/test/plugin/plugin_mgr_test.cpp
+++ b/be/test/plugin/plugin_mgr_test.cpp
@@ -18,11 +18,11 @@
 #include "plugin/plugin_mgr.h"
 
 #include <gtest/gtest.h>
-#include <libgen.h>
 
 #include "plugin/plugin.h"
 #include "plugin/plugin_loader.h"
 #include "string"
+#include "test_util/test_util.h"
 
 namespace doris {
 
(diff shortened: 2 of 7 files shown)


morningman (approved):
LGTM

PR #5205, opened 2021-01-07
[Load] Load job should not begin transaction when task queue in loadingLoadTaskScheduler is full to avoid txn timeout

### LINKED ISSUE #5204: [Broker Load] Load job should not begin transaction when task queue in loadingLoadTaskScheduler is full
Now, many user use broker load to load hdfs data into doris, and they will submit many broker job at one time, like #5150  and loading task usually takes long  time to finish task while pending task cost little time to finish task. 
when task queue in loadingLoadTaskScheduler is full, there is no need to begin txn for a new load job, which will just wait for
other broker job to finish their loading task. Once start a new transaction, the broker job's status will be transformed from pending to loading, which may be confused for user that it is not load slowly but for the limited thread resource.
So I will make two changes, 
1. split async_load_task_pool_size config into two configs: async_pending_load_task_pool_size, async_loading_load_task_pool_size
2. check the task queue in loadingLoadTaskScheduler is full, once full, we should not begin transaction immediately for bulk load job

caiconghui:
## Proposed changes

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #5204), and have described the bug/feature there in detail
- [] Compiling and unit tests pass locally with my changes

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/Config.java
@@ -593,8 +593,8 @@ public class Config extends ConfigBase {
     public static int max_running_txn_num_per_db = 100;
 
     /**
-     * just for compatible with old version, this config has been replaced by async_pending_load_task_pool_size and
-     * async_loading_load_task_pool_size, it will be removed in next version
+     * This configuration is just for compatible with old version, this config has been replaced by async_pending_load_task_pool_size
+     * and async_loading_load_task_pool_size, it will be removed in next version
      */
     @ConfField(mutable = false, masterOnly = true)
     public static int async_load_task_pool_size = 10;
--- a/fe/fe-core/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/Config.java
@@ -593,19 +593,26 @@ public class Config extends ConfigBase {
     public static int max_running_txn_num_per_db = 100;
 
     /**
+     * just for compatible with old version, this config has been replaced by async_pending_load_task_pool_size and
+     * async_loading_load_task_pool_size, it will be removed in next version
+     */
+    @ConfField(mutable = false, masterOnly = true)
+    public static int async_load_task_pool_size = 10;
+
+    /**
      * The pending_load task executor pool size. This pool size limits the max running pending_load tasks.
      * Currently, it only limits the pending_load task of broker load and spark load.
      * It should be less than 'max_running_txn_num_per_db'
      */
     @ConfField(mutable = false, masterOnly = true)
-    public static int async_pending_load_task_pool_size = 5;
+    public static int async_pending_load_task_pool_size = async_load_task_pool_size;
 
     /**
      * The loading_load task executor pool size. This pool size limits the max running loading_load tasks.
      * Currently, it only limits the loading_load task of broker load.
      */
     @ConfField(mutable = false, masterOnly = true)
-    public static int async_loading_load_task_pool_size = 20;
+    public static int async_loading_load_task_pool_size = async_load_task_pool_size;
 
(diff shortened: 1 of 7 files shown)


fe/fe-core/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -593,12 +593,19 @@
...
     /**
-     * The load task executor pool size. This pool size limits the max running load tasks.
-     * Currently, it only limits the load task of broker load, pending and loading phases.
+     * The pending_load task executor pool size. This pool size limits the max running pending_load tasks.
+     * Currently, it only limits the pending_load task of broker load and spark load.
      * It should be less than 'max_running_txn_num_per_db'
      */
     @ConfField(mutable = false, masterOnly = true)
-    public static int async_load_task_pool_size = 10;
+    public static int async_pending_load_task_pool_size = 5;
```
morningman, line 601: Better to keep the config name. Or made some compatibility.
Some other users may modify this config to enhance their load ability.
If the name changed, their config will be disable and fall back to default value, this will cause problem

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -593,12 +593,19 @@
...
+
+    /**
+     * The loading_load task executor pool size. This pool size limits the max running loading_load tasks.
+     * Currently, it only limits the loading_load task of broker load.
+     */
+    @ConfField(mutable = false, masterOnly = true)
+    public static int async_loading_load_task_pool_size = 20;
```
morningman, line 608: I think pending number should be larger loading number？

fe/fe-core/src/main/java/org/apache/doris/load/loadv2/LoadJobScheduler.java
```diff
@@ -63,12 +64,18 @@ protected void runAfterCatalogReady() {
 
     private void process() throws InterruptedException {
         while (true) {
-            // take one load job from queue
-            LoadJob loadJob = needScheduleJobs.poll();
-            if (loadJob == null) {
+            if (!needScheduleJobs.isEmpty()) {
+                if (needScheduleJobs.peek() instanceof BrokerLoadJob && Catalog.getCurrentCatalog().getLoadingLoadTaskScheduler().isTaskQueueFull()) {
```
morningman, line 68: Broker or Spark Load？

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
morningman: And the `desired_max_waiting_jobs` is useless?

fe/fe-core/src/main/java/org/apache/doris/load/loadv2/LoadJobScheduler.java
caiconghui: Spark Load not use async loading task pool

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
caiconghui: pending task not execute heavy task，while loading task cost much time

caiconghui: pending task is used to collect some schema info or preparement work， loading task is used to execute loading job

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
caiconghui: > And the `desired_max_waiting_jobs` is useless?

desired_max_waiting_jobs is used to limit blocking queue size

morningman (approved):
LGTM

PR #5226, opened 2021-01-12
[Bug] Fix row_number and group by have inconsistent partition results for (0.0, -0.0)

### LINKED ISSUE #5225: [BUG] row_number and group by have inconsistent partition results for (0.0, -0.0)
**Describe the bug**
row_number and group by have inconsistent partition results for (0.0, -0.0)

**To Reproduce**
A.k1 contains -0.0 and 0.0
The result of `select k1, count(1) from A group by k1` is as follows.
+------+----------+
| k1   | count(1) |
+------+----------+
|    0 |        2 |
|   -0 |        1 |
+------+----------+
The result of `select k1, row_number() over (partition by k1 order by k2) as t from A` is as follows. 
+------+------+
| k1   | t   |
+------+------+
|    0 |    1 |
|    0 |    2 |
|   -0 |    3 |
+------+------+

**Expected behavior**
We hope to unify the partition results of group by and row_number

xinyiZzz:
## Problem analysis
The essence of the problem is behavior of negative zero (- 0.0) in comparison with positive zero (+ 0.0).
Currently in GroupBy and HashPartition, -0.0 is not equal to 0.0 (result of Hash function), so the -0.0 and 0.0 are divided into 2 partitions.

In row_number analytic function, for the sorted data, a new partition will be opened when the values ​​​​of the upper and lower rows are not equal. But in C++ the comparison 0.0 == -0.0 is true, so 0.0 and -0.0 are divided into the same partition for row_number.

(Floating point arithmetic in C++ is often IEEE-754. This norm defines two different representations for the value zero: positive zero and negative zero. It is also defined that those two representations must compare equals. Refer to https://stackoverflow.com/questions/45795397)

## Fix method
1. (Deprecated) Modifies the eq comparison of two Doubles in BinaryPredicate, and returns -0.0 == 0.0 as false when both sides of expr are SlotRef.
Problems:
  (1) When order by is still considered equal to 0.0 and -0.0, the order of 0.0 and -0.0 in the result is random
  (2) The reason for restricting both sides of BinaryPredicate to be SlotRef is that the constant cannot be defined as -0.0. For example, the expression of `where k1 = -0.0` in SQL in actual calculation is `where k1 = 0.0`.

2. Modify the Hash function in `hash_util.hpp`, when data is negative zero, rewrite it to 0.0, which will be used in `DataStreamSender::send HashPartition` and `GroupBy`, because the original value of data is covered, so -0.0 in the original data after Hash will be changed to 0.0.

Our goal is that -0.0 will not appear in Doris, and all will be replaced with 0.0. Therefore, you need to modify the Broker Load in the future, and change -0.0 to 0.0 when importing. Mysql and SparkSQL are currently converted to 0.0 when Load -0.0 to Double column. But there is a situation that may not be avoided, that is, `select round(-0.2,0);`, the returned `-0.0` is meaningful, refer to https://www.johndcook.com/blog/2010/06/15/why-computers-have-signed-zero/

## Types of changes

## Checklist

- [x] I have create an issue on (Fix #5225), and have described the bug/feature there in detail

### DIFF
--- a/be/src/util/hash_util.hpp
+++ b/be/src/util/hash_util.hpp
@@ -29,6 +29,7 @@
 #include <nmmintrin.h>
 #endif
 #include <zlib.h>
+#include <cmath>
 #include "util/cpu_info.h"
 #include "util/murmur_hash3.h"
 #include "gen_cpp/Types_types.h"
@@ -58,6 +59,10 @@ public:
         uint32_t words = bytes / sizeof(uint32_t);
         bytes = bytes % sizeof(uint32_t);
 
+        // When data is negative zero, rewrite it to 0.0, see Doris#5226
+        if (*(double*)data == 0.0 and std::signbit(*(double*)data) == true) {
+            *(double*)data = 0.0;
+        }
         const uint32_t* p = reinterpret_cast<const uint32_t*>(data);
 
         while (words--) {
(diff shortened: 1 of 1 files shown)


EmmyMiao87 (approved):
LGTM

PR #5254, opened 2021-01-17
[Bug] Remove schema hash and fix bug of calculating table signature

morningman:
## Proposed changes

1. Schema hash is useless long time ago
    Currently, schema hash can only be generated as a random integer, no need to calculated
    from real schema.

2. The CRC32 algo is not enough to generate the table' signature.
    Table's signature is used to determine whether the tables have the same schema.
    And current CRC32 algo may return same signature even if table's schema are different.

    So I change it to calculate the md5 of a signature string assambled by schema info of a table.

## Types of changes


EmmyMiao87 (approved):
LGTM

PR #5277, opened 2021-01-21
[Load Parallel][1/3] Broker Load supports setting the load parallelism

morningman:
## Proposed changes

Similar to the parallel_fragment_exec_instance_num parameter,
it allows the user to set the parallelism of the load execution plan
on a single node when the broker load is submitted.

eg:
```
...
properties (
"load_parallelism" = "4";
...
)
```

This parameter is currently only used to support the load parallelism setting,
but it cannot significantly improve the load speed for the time being.
The speed increase will be completed in subsequent code submissions.
Documents will also be added in subsequent submissions.

This PR also update the FE meta version.

#5281

## Types of changes


yangzhg (approved):
LGTM

PR #5278, opened 2021-01-21
[Bug] fix bug that dead lock may occur when drop tablet concurrent with tablet distribution interface

weizuo93:
## Proposed changes

Fix bug that dead lock may occur when drop tablet concurrent with calling tablet distribution interface.

## Types of changes

### DIFF
--- a/be/src/olap/tablet_manager.cpp
+++ b/be/src/olap/tablet_manager.cpp
@@ -1492,6 +1492,9 @@ void TabletManager::get_tablets_distribution_on_different_disks(
     std::vector<DataDir*> data_dirs = StorageEngine::instance()->get_stores();
     std::map<int64_t, std::set<TabletInfo>> partition_tablet_map;
     {
+        // When drop tablet, '_partition_tablet_map_lock' is locked in 'tablet_shard_lock'.
+        // To avoid locking 'tablet_shard_lock' in '_partition_tablet_map_lock', we lock and
+        // copy _partition_tablet_map here.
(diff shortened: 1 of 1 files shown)


be/src/olap/tablet_manager.cpp
```diff
@@ -1490,9 +1490,13 @@ void TabletManager::get_tablets_distribution_on_different_disks(
                     std::map<int64_t, std::map<DataDir*, int64_t>> &tablets_num_on_disk,
                     std::map<int64_t, std::map<DataDir*, std::vector<TabletSize>>> &tablets_info_on_disk) {
     std::vector<DataDir*> data_dirs = StorageEngine::instance()->get_stores();
-    ReadLock rlock(&_partition_tablet_map_lock);
-    std::map<int64_t, std::set<TabletInfo>>::iterator partition_iter = _partition_tablet_map.begin();
-    for (; partition_iter != _partition_tablet_map.end(); partition_iter++) {
+    std::map<int64_t, std::set<TabletInfo>> partition_tablet_map;
+    {
+        ReadLock rlock(&_partition_tablet_map_lock);
```
morningman, line 1495: Could you please to add comment to explain why we need lock and copy the `_partition_tablet_map`?

weizuo93: > Could you please to add comment to explain why we need lock and copy the `_partition_tablet_map`?

Done.

morningman (approved):
LGTM

PR #5279, opened 2021-01-21
[Bug] Fix bug that recover table throw NPE

morningman:
## Proposed changes

This bug is introduced from #3775

## Types of changes


EmmyMiao87 (approved):
LGTM

PR #5286, opened 2021-01-22
[Doc] Update README.md: a reword suggestion

Sunt-ing:
## Proposed changes
A reword suggestion.

Target:
1. Remove the redundant words.
2. Make the statement more friendly.
3. "must be" here is a grammar error.

Nothing will be broken by this reword change.


yangzhg (commented):
you'd better commit those docs in one PR

PR #5288, opened 2021-01-22
[Doc] Update README.md: a reword suggestion

Sunt-ing:
## Proposed changes
Make the path more readable as the convention of Markdown suggests.

It will not break any existing build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ After you installed above all, you also must be set them to environment variable
 If your GCC version is lower than 7.3.0, you can run:
 
 ```
-sudo yum install devtoolset-4-toolchain -y
+sudo yum install devtoolset-7-toolchain -y
 ```
 
 and then, set the path of GCC (e.g `/opt/rh/devtoolset-4/root/usr/bin`) to the environment variable PATH.
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ If your GCC version is lower than 7.3.0, you can run:
 sudo yum install devtoolset-4-toolchain -y
 ```
 
-and then, set the path of GCC (e.g /opt/rh/devtoolset-4/root/usr/bin) to the environment variable PATH.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -131,7 +131,7 @@ If your GCC version is lower than 7.3.0, you can run:
 sudo yum install devtoolset-4-toolchain -y
```
yangzhg, line 131:
```suggestion
sudo yum install devtoolset-7-toolchain -y
```

README.md
```diff
@@ -131,7 +131,7 @@ If your GCC version is lower than 7.3.0, you can run:
...
 ```
 
-and then, set the path of GCC (e.g /opt/rh/devtoolset-4/root/usr/bin) to the environment variable PATH.
+and then, set the path of GCC (e.g `/opt/rh/devtoolset-4/root/usr/bin`) to the environment variable PATH.
```
yangzhg, line 134:
```suggestion
and then, set the path of GCC (e.g `/opt/rh/devtoolset-7/root/usr/bin`) to the environment variable PATH.
```

yangzhg: you'd better change the Chinese doc too.

Sunt-ing:
@yangzhg I don't find any Chinese version of README.md. Yes, there are at least three readme files: /README.md, /docs/en/README.md, /docs/zh-CN/README.md.
My change here is located in /README.md, which has no paired Chinese version yet.
I globally searched the whole project with the key words "/opt/rh/devtoolset", and only found one match: in /README.md.

Sunt-ing:
By the way, maybe we can mix Chinese and English version docs together in the same directory, and only left the file name different. The benefits are we can find the mismatch of these two docs more easily in this way.
Take Apache Flink as am example: https://github.com/apache/flink/tree/master/docs/internals

Sunt-ing:
@yangzhg I have merged PR#5286 here

yangzhg (approved):
LGTM

PR #5314, opened 2021-01-28
[Backup]Support content, exclude and whole database in backup

### LINKED ISSUE #5298: Backup and Restore support view and external table
# Background 

Existing function

Only olap table backup is supported, and the name of the table to be backed up must be manually declared, which is inconvenient for full database backup.

# Requirement

1. backup full database, data or metadata, partial table
2. backup an restore view, external table, related resources etc.
3. When restoring, overwrite all data in the existing table, if it does not exist, create a new table directly
4. backup the entire database but only restores individual tables during restore.
5. The backup support excludes some tables that are too large, and backs up the others.

# API design

```

BACKUP SNAPSHOT [db_name].{snapshot_name}

TO 'repo_name'

[on clause]

[EXCLUDE (

    'table_name' [partition (p1,...)]

)]

[properties (

    "content" = "metadata_only|data_only"

)]



RESTORE SNAPSHOT [db_name].{snapshot_name}

TO 'repo_name'

[on clause]

[EXCLUDE (

    'table_name' [partition (p1,...)]

)]

[properties (

    "content" = "metadata_only|data_only"

)]
```

EmmyMiao87:
## Proposed changes
This PR support following functions:
1. Support content properties in backup stmt. It means user can backup only metadata or meta+data which use content [METADATA_ONLY| ALL]attribute to distinguish.
2. Support exclude some tables in backup and restore stmt. This means that some very large and unimportant tables can be excluded when the entire database is backed up.
3. Support backup and restore whole database instead of declaring each table name in the backup and restore statement.

The backup and restore api has changed as following:
```
BACKUP SNAPSHOT [db_name].{snapshot_name}
TO 'repo_name'
[ON|EXCLUDE (
    'table_name' [partition (p1,...)]
)]
[properties (
    "content" = "metadata_only|all"
)]

RESTORE SNAPSHOT [db_name].{snapshot_name}
TO 'repo_name'
[EXCLUDE|ON (
    'table_name' [partition (p1,...)]
)]
[properties (
)]
```

## Types of changes

## Checklist

- [] I have created an issue on (Fix #5298 ) and described the bug/feature there in detail
- [TODO] If these changes need document changes, I have updated the document

## Further comments

I will continue to update the documentation, please review my code first.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/AbstractBackupTableRefClause.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/AbstractBackupTableRefClause.java
@@ -1,0 +1,87 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package org.apache.doris.analysis;
+
+import org.apache.doris.common.AnalysisException;
+import org.apache.doris.common.UserException;
+
+import com.google.common.base.Joiner;
+import com.google.common.collect.Maps;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.util.List;
+import java.util.Map;
+
+public class AbstractBackupTableRefClause implements ParseNode {
+    private static final Logger LOG = LogManager.getLogger(AbstractBackupTableRefClause.class);
+
+    private boolean isExclude;
+    private List<TableRef> tableRefList;
+
+    public AbstractBackupTableRefClause(boolean isExclude, List<TableRef> tableRefList) {
+        this.isExclude = isExclude;
+        this.tableRefList = tableRefList;
+    }
+
+    @Override
+    public void analyze(Analyzer analyzer) throws UserException {
+        // normalize
+        // table name => table ref
+        Map<String, TableRef> tblPartsMap = Maps.newTreeMap(String.CASE_INSENSITIVE_ORDER);
+        for (TableRef tblRef : tableRefList) {
+            String tblName = tblRef.getName().getTbl();
+            if (!tblPartsMap.containsKey(tblName)) {
+                tblPartsMap.put(tblName, tblRef);
+            } else {
+                throw new AnalysisException("Duplicated restore table: " + tblName);
+            }
+        }
+
+        // update table ref
+        tableRefList.clear();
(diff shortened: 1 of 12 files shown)


fe/fe-core/src/main/java/org/apache/doris/analysis/AbstractBackupTableRefClause.java
```diff
@@ -0,0 +1,87 @@
...
+            }
+        }
+
+        // update table ref
+        tableRefList.clear();
+        for (TableRef tableRef : tblPartsMap.values()) {
+            tableRefList.add(tableRef);
+        }
+
+        LOG.debug("table refs after normalization: \n{}", Joiner.on("\n").join(tableRefList));
```
morningman, line 63: do not use `\n`, for easy log greping.

fe/fe-core/src/main/java/org/apache/doris/backup/BackupHandler.java
```diff
@@ -269,10 +271,30 @@ private void backup(Repository repository, Database db, BackupStmt stmt) throws
                     + " is read only");
         }
 
+        // Determine the tables to be backed up
+        Set<String> tableNames = Sets.newHashSet();
+        AbstractBackupTableRefClause abstractBackupTableRefClause = stmt.getAbstractBackupTableRefClause();
+        if (abstractBackupTableRefClause == null) {
+            tableNames = db.getTableNamesWithLock();
+        } else if (abstractBackupTableRefClause != null && abstractBackupTableRefClause.isExclude()) {
```
morningman, line 279:
```suggestion
        } else if (abstractBackupTableRefClause.isExclude()) {
```

fe/fe-core/src/main/java/org/apache/doris/backup/BackupHandler.java
```diff
@@ -269,10 +271,30 @@ private void backup(Repository repository, Database db, BackupStmt stmt) throws
...
+            tableNames = db.getTableNamesWithLock();
+            for (TableRef tableRef : abstractBackupTableRefClause.getTableRefList()) {
+                tableNames.remove(tableRef.getName().getTbl());
```
morningman, line 282: Do we need to check if there "excluded" tables exist in database?

eg:
```suggestion
                if (!tableNames.remove(tableRef.getName().getTbl())) {
                    throw exception()
                }
```

fe/fe-core/src/main/java/org/apache/doris/backup/BackupHandler.java
```diff
@@ -269,10 +271,30 @@ private void backup(Repository repository, Database db, BackupStmt stmt) throws
...
+            }
+        }
+        List<TableRef> tblRefs = Lists.newArrayList();
+        if (tableNames.isEmpty()) {
```
morningman, line 286: There may be no table in database at all, and in this case, the `tableNames` is empty, and `abstractBackupTableRefClause` is null. So NPE will thrown.

fe/fe-core/src/main/java/org/apache/doris/backup/BackupJobInfo.java
```diff
@@ -204,6 +278,7 @@ public String getOrginNameByAlias(String alias) {
         public String database;
         @SerializedName("backup_time")
         public long backupTime;
+        public BackupContent content;
```
morningman, line 281: No need `@SerializedName`?

EmmyMiao87: No, it will be serialized by default, but some serialized names and attribute names are different.

morningman (approved):
LGTM

PR #5315, opened 2021-01-28
Adding a column with SUM aggregation to table should restrict the default value

### LINKED ISSUE #5316: Inconsistent between three replicas
```
create table t2(c1 int, c2 int, c3 int sum) aggregate key(c1, c2) 
distributed by hash(c1) 
buckets 1;
```
insert five patches data

If the three replica replica compaction progress is not at sync.
```
alter table t2 add column c4 column sum default '1';
```

Then the result between three replicas is not consistent.

chaoyli:
One Tablet have three replicas, the compaction progress is different.
Considering the following scenario, replica A have 3 versions(1, 2, 3);
replica B have 2 versions(1-2, 3);
replica C have 1 versions(1-3).
Now a column named city been added with default zero 1.
Replica A will be resulted as 3, replica B results as 2,
replica C results as 1.

So there is a necessity to restrict the default value to zero for SUM aggregation column.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/alter/SchemaChangeHandler.java
@@ -753,7 +753,8 @@ public class SchemaChangeHandler extends AlterHandler {
                 throw new DdlException("Can not assign aggregation method on key column: " +  newColName);
             } else if (null == newColumn.getAggregationType()) {
                 newColumn.setIsKey(true);
-            } else if (newColumn.getAggregationType() == AggregateType.SUM && !newColumn.getDefaultValue().equals("0")) {
+            } else if (newColumn.getAggregationType() == AggregateType.SUM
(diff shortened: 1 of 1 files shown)


fe/fe-core/src/main/java/org/apache/doris/alter/SchemaChangeHandler.java
```diff
@@ -753,6 +753,8 @@ private void addColumnInternal(OlapTable olapTable, Column newColumn, ColumnPosi
                 throw new DdlException("Can not assign aggregation method on key column: " +  newColName);
             } else if (null == newColumn.getAggregationType()) {
                 newColumn.setIsKey(true);
+            } else if (newColumn.getAggregationType() == AggregateType.SUM && !newColumn.getDefaultValue().equals("0")) {
```
morningman, line 756: `newColumn.getDefaultValue()` may be `null` here?

morningman (approved):
LGTM

morningman:
We will change the case in NewDorisTest to meet this issue.

Issue #5317, opened 2021-01-29
move common codes form fe-core to fe-common


copperybean:
Multiple module project structure is a good project structure and adapted by many popular projects. Now, many common codes are integrated in fe-core, it's not convenient for new modules which will need these common codes, so we'd better move some common codes from fe-core to fe-common first.

PR #5318, opened 2021-01-29
[Enhancement] move common codes from fe-core to fe-common and remove log4j1 (#5317)

copperybean:
The io related codes may be used by new modules, so It's better to move them to fe-common.

The modification to fe-core is frequent, but there are many generated java files by thrift will slow down the compilation, so It's better to move thrift generation process to fe-common.

Currently both log4j1 and log4j2 are used, which leads to logs are written to wrong files. Our modification will remove log4j1 from dependency, use slf4j + slf4j -> log4j2 instead.


morningman (dismissed):
LGTM
But fe plugins build failed

morningman (dismissed):
LGTM

morningman (approved):
LGTM

Issue #5326, opened 2021-02-01
[Feature] Support variable sql_select_limit


ccoffline:
Implement sql_select_limit like [MySQL](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_sql_select_limit).
This may not be completely consistent with MySQL for a simple Implemention. For example, the limit is not determined by whether there is a limit in the select statement.

PR #5332, opened 2021-02-01
[Feature] Support variable sql_select_limit (#5326)

ccoffline:
## Proposed changes

as #5326
This is mainly used for stopping big rowset. The ReturnRows is not exactly as `sql_select_limit`, but stop after the last RowBatch sended.

## Types of changes

## Checklist

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
@@ -776,6 +776,10 @@ public class StmtExecutor {
                     channel.sendOnePacket(row);
                 }
                 context.updateReturnRows(batch.getBatch().getRows().size());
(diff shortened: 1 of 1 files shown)


HangyuanLiu (commented):
I think if we want to implement this feature, we should try our best to ensure that the behavior is consistent with mysql 。
mysql: "If a SELECT has a LIMIT clause, the LIMIT takes precedence over the value of sql_select_limit"

ccoffline:
@HangyuanLiu
The MySQL documentation is unclear whether the LIMIT clause is in the outer clause or anywhere in the sql that takes over `sql_select_limit`.
I would try to figure it out. This commit is an emergency fix for short of memory on FE caused by large result set queries, so I didn't choose a complex implementation.
Besides, this is not a variable that would make the app depends on it. The scenario of Doris queries is not exactly the same as MySQL. I think we should figure out in what case the LIMIT clause needs to take over.

PR #5336, opened 2021-02-01
[Spark-Doris-Connector][Bug-Fix] Resolve deserialize exception when Spark Doris Connector in aync deserialize mode

### LINKED ISSUE #5335: [Bug] ScalaValueReader throw TApplicationException: get_next failed: out of sequence response: expected 21 but got 20
**Describe the bug**

When I use spark on doris with config `doris.deserialize.arrow.async=true`，ScalaValueReader throw TApplicationException: get_next failed: out of sequence response: expected 21 but got 20.

And then, ScalaValueReader will **block forever**

This is because, when `doris.deserialize.arrow.async=true`, `get_next()` will invoke in **asyncThread**，but `close()` will invoke in **current thread**, the sequence will encounter race condition 

**To Reproduce**
run
``` scala
val sparkSettings = new SparkSettings(sparkConf)
sparkSettings.setProperty("doris.deserialize.arrow.async", "true")
val partitions = RestService.findPartitions(sparkSettings, logger)
val reader = new ScalaDorisRowValueReader(partitions.get(0), partitions.get(0).settings())
while (reader.hasNext) {
  val nextBatch = reader.next // invoke in asyncThread
}
reader.close() // invoke in current thread
```

**Expected behavior**
No exception

**Screenshots**
image

**Desktop (please complete the following information):**
None

**Smartphone (please complete the following information):**
None

**Additional context**
None

924060929:
## Proposed changes

fix #5335

## Types of changes

## Checklist

## Further comments
None

### DIFF
--- a/extension/spark-doris-connector/src/main/scala/org/apache/doris/spark/rdd/ScalaValueReader.scala
+++ b/extension/spark-doris-connector/src/main/scala/org/apache/doris/spark/rdd/ScalaValueReader.scala
@@ -124,9 +124,12 @@ class ScalaValueReader(partition: PartitionDefinition, settings: Settings) {
     params
   }
 
-  protected val openResult: TScanOpenResult = clientLock.synchronized {
-    client.openScanner(openParams)
-  }
+  protected val openResult: TScanOpenResult =
+    if (deserializeArrowToRowBatchAsync) {
+      clientLock.synchronized {
+        client.openScanner(openParams)
+      }
+    } else client.openScanner(openParams)
   protected val contextId: String = openResult.getContext_id
   protected val schema: Schema =
     SchemaUtils.convertToSchema(openResult.getSelected_columns)
@@ -197,9 +200,12 @@ class ScalaValueReader(partition: PartitionDefinition, settings: Settings) {
         val nextBatchParams = new TScanNextBatchParams
         nextBatchParams.setContext_id(contextId)
         nextBatchParams.setOffset(offset)
-        val nextResult = clientLock.synchronized {
-          client.getNext(nextBatchParams)
-        }
+        val nextResult =
+          if (deserializeArrowToRowBatchAsync) {
+            clientLock.synchronized {
+              client.getNext(nextBatchParams)
+            }
+          } else client.getNext(nextBatchParams)
         eos.set(nextResult.isEos)
         if (!eos.get) {
           rowBatch = new RowBatch(nextResult, schema)
@@ -225,7 +231,11 @@ class ScalaValueReader(partition: PartitionDefinition, settings: Settings) {
   def close(): Unit = {
     val closeParams = new TScanCloseParams
     closeParams.context_id = contextId
-    clientLock.synchronized {
+    if (deserializeArrowToRowBatchAsync) {
+      clientLock.synchronized {
+        client.closeScanner(closeParams)
+      }
+    } else {
       client.closeScanner(closeParams)
     }
   }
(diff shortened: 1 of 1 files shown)


924060929:
Gentle ping @wuyunfeng

wuyunfeng:
@924060929 This is a nice capture. I am curious about whether  the lock is necessary when running in sync mode?

924060929:
> @924060929 This is a nice capture. I am curious about whether the lock is necessary when running in sync mode （default mode）?

Unnecessary, just for simplify. Synchronized has little effect on performance in sync mode.

Shall we remove synchronized in sync mode?

wuyunfeng:
@924060929 I suggest that there can be such an optimization

924060929:
@wuyunfeng OK, I will add this optimization soon

924060929:
> @924060929 I suggest that there can be such an optimization

@wuyunfeng done

wuyunfeng:
I just using `Java` describe the implementation which I would prefer:

First define a `NoOpLock`:

```
protected static final class NoOpLock implements Lock {

        @Override
        public void lock() {
        }

        @Override
        public void lockInterruptibly() throws InterruptedException {
        }

        @Override
        public boolean tryLock() {
            return true;
        }

        @Override
        public boolean tryLock(long time, TimeUnit unit) throws InterruptedException {
            return true;
        }

        @Override
        public void unlock() {
        }

        @Override
        public Condition newCondition() {
            throw new UnsupportedOperationException("NoOpLock can't provide a condition");
        }
    }
``` 
We just define a `private final Lock lock = new NoOpLock();`

If using `async` mode,  let the lock using the real `Lock`, If not, we can replace the `lock` with `NoOpLock`.

If we do this, the first code is more elegant, and second, we can avoid excessive branch switching.

 I think your approach is fine, would you like to think about my suggestion

924060929:
@wuyunfeng OK, first I think the same solution as you note, but don't know where the place the NoOpLock, we can add a inner class, or add some public class in any package, or any third-party package has same implement?

So I add the if condition without Lock implement.

Can you give more advice?

wuyunfeng:
I think you can just use a inner class

924060929:
> I think you can just use a inner class

done

wuyunfeng (dismissed):
To be honest, I don’t like this function name  `lockClient` too much, but this  looks good to me

extension/spark-doris-connector/src/main/scala/org/apache/doris/spark/rdd/ScalaValueReader.scala
```diff
@@ -46,6 +47,9 @@ class ScalaValueReader(partition: PartitionDefinition, settings: Settings) {
   protected val logger = Logger.getLogger(classOf[ScalaValueReader])
 
   protected val client = new BackendClient(new Routing(partition.getBeAddress), settings)
+  protected val clientLock =
+    if (deserializeArrowToRowBatchAsync) new ReentrantLock()
```
Youngwb, line 51: I have a question， deserializeArrowToRowBatchAsync is init below,  whether the use is valid here

924060929: Is valid, equals to c++:
```c++
bool deserializeArrowToRowBatchAsync() {
  return ...
}
```

924060929: My fault, `protected var deserializeArrowToRowBatchAsync` should be `protected def deserializeArrowToRowBatchAsync`

924060929:
@Youngwb already fix bug which use deserializeArrowToRowBatchAsync before initialize

Youngwb (approved):
+1

wuyunfeng (approved):
LGTM

wuyunfeng:
@924060929 I have merge this into master. Thanks

PR #5344, opened 2021-02-02
[Compaction] Cache compaction score for tablet to avoid calculating each tablet score in every production

weizuo93:
## Proposed changes

Keep tablet `compaction score` in memory to avoid calculating tablet score in every production. After `load`, `clone` and `compaction`，updating the tablet `compaction score`in memory.

## Types of changes

## Checklist

### DIFF
--- a/be/src/olap/compaction.cpp
+++ b/be/src/olap/compaction.cpp
@@ -127,8 +127,11 @@ OLAPStatus Compaction::do_compaction_impl(int64_t permits) {
     int64_t now = UnixMillis();
     if (compaction_type() == ReaderType::READER_CUMULATIVE_COMPACTION) {
         _tablet->set_last_cumu_compaction_success_time(now);
+        _tablet->update_cumulative_compaction_score(_tablet->calc_compaction_score(CompactionType::CUMULATIVE_COMPACTION));
+        _tablet->update_base_compaction_score(_tablet->calc_compaction_score(CompactionType::BASE_COMPACTION));
     } else {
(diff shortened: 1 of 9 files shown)


be/src/olap/compaction.cpp
```diff
@@ -127,8 +127,11 @@ OLAPStatus Compaction::do_compaction_impl(int64_t permits) {
     int64_t now = UnixMillis();
     if (compaction_type() == ReaderType::READER_CUMULATIVE_COMPACTION) {
         _tablet->set_last_cumu_compaction_success_time(now);
+        _tablet->update_cumulative_compaction_score(_tablet->calc_compaction_score(CompactionType::CUMULATIVE_COMPACTION));
```
acelyc111, line 130: I found every call of  `update_cumulative_compaction_score()` 's parameter is `calc_compaction_score()`, what's reason to do this?

acelyc111, line 130: I found every call of  `update_cumulative_compaction_score()` 's parameter is `calc_compaction_score()`, what's reason to do this?

acelyc111, line 130: I found every call of  `update_cumulative_compaction_score()` 's parameter is `calc_compaction_score()`, what's reason to do this?

weizuo93: > I found every call of `update_cumulative_compaction_score()` 's parameter is `calc_compaction_score()`, what's reason to do this?

Calculating compaction score before updating metrics.  Of course, these two steps can be encapsulated as a function.

PR #5347, opened 2021-02-02
Fix bug that sql with limit query statistics are wrong

### LINKED ISSUE #5340: The scan row in the audit log is incorrect when querying SQL with limit
**Describe the bug**
The query stastics may be enlarged by `merge_pb` when querying SQL is with `limit`.
The query stastics collected by BE is not delta. 
Thus, we should not merge them in merge_pb.

SQL likes that `select xxx from a cross join b limit xxx`.
,If the `_collect_query_statistics_with_every_batch` is true,  the query stastics will be collected by batch in `PlanFragmentExecutor::open_internal()`
image

However, the query stastics of each batch is not a delta. 
I think we should not merge the query stastics from the same sender in exchange nodes
in  this [pr](https://github.com/apache/incubator-doris/pull/1354/files#diff-f12fb6abe2807800b22c9a07c08295fc7d46de5b877ba651456ae9dc7086778b).

**To Reproduce**
SQL like that `select xxx from a cross join b limit xxx`.
Take scan rows as an example:
The scan rows in audit log is 6w+.
However, in query profile the sum of RawScanRows is only 3w+.

spaces-X:
When querying sql is with `limit`, the query statistics will be enlarged by merging the query statistics of each batch.

fix #5340

### DIFF
--- a/be/src/runtime/query_statistics.cpp
+++ b/be/src/runtime/query_statistics.cpp
@@ -33,7 +33,7 @@ void QueryStatisticsRecvr::insert(const PQueryStatistics& statistics, int sender
     } else {
         query_statistics = iter->second;
     }
-    query_statistics->merge_pb(statistics);
+    query_statistics->from_pb(statistics);
 }
 
 QueryStatisticsRecvr::~QueryStatisticsRecvr() {
--- a/be/src/runtime/query_statistics.h
+++ b/be/src/runtime/query_statistics.h
@@ -65,10 +65,10 @@ public:
         statistics->set_returned_rows(returned_rows);
     }
 
-    void merge_pb(const PQueryStatistics& statistics) {
-        scan_rows += statistics.scan_rows();
-        scan_bytes += statistics.scan_bytes();
-        cpu_ms += statistics.cpu_ms();
+    void from_pb(const PQueryStatistics& statistics) {
+        scan_rows = statistics.scan_rows();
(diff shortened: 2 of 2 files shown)


spaces-X:
@imay

be/src/runtime/query_statistics.cpp
```diff
@@ -33,7 +33,7 @@ void QueryStatisticsRecvr::insert(const PQueryStatistics& statistics, int sender
     } else {
         query_statistics = iter->second;
     }
-    query_statistics->merge_pb(statistics);
+    query_statistics->from_pb(statistics);
```
EmmyMiao87, line 36: Shouldn't it be accumulation here?

spaces-X: The query statistics such as scan_rows refers to ```RuntimeProfile::Counter _raw_rows_counter``` have been already accumlated in counter and the values in query statistics are not delta.
The query statistics sent by different senders should be merged in Exchange Nodes. 
However, the query statistics sent by the same sender why should we merge them here.
If merged, the query statistics will be enlarged by repeat accumulation when sql with limit.

EmmyMiao87: I see, I don’t think it’s a problem

EmmyMiao87 (approved):
LGTM

PR #5352, opened 2021-02-03
remove log4j1 from dependency in fe, use slf4j + slf4j -> log4j2 instead

vinson0526:
To keep Doris logging system simple and easy to use:
1. Remove log4j.log4j and org.slf4j.slf4j-log4j12 from FE dependency.
2. Replace org.apache.log4j.Logger by org.slf4j.Logger in every file that use  log4j 1.x version.


fe/fe-core/src/test/java/org/apache/doris/analysis/CreateRoutineLoadStmtTest.java
```diff
@@ -43,10 +41,12 @@
 import mockit.Mock;
 import mockit.MockUp;
 import mockit.Mocked;
+import org.slf4j.Logger;
```
EmmyMiao87, line 44: please pay attention order of import

EmmyMiao87 (approved):
LGTM

PR #5359, opened 2021-02-04
[UT] Modify to avoid some UT failure accidentally

morningman:
## Proposed changes

After an Alter job finished, the job's state is FINISHED, but table's state
may not be NORMAL for a while.
We need to make sure that table's state become NORMAL to continue next UT.


yangzhg (approved):
LGTM

PR #5360, opened 2021-02-04
[Bug] Fix the problem of floating point precision when importing parquet data

morningman:
The double data "4206.9" in parquet is converted to decimal data "4206.8999" in Doris,
which is not right.


yangzhg (approved):
LGTM

PR #5364, opened 2021-02-05
[Performance Optimization and Refactor]  Delete push to segment v2

### LINKED ISSUE #5358: [Proposal] Support column predicate replace delete condition
## Motivation
At present,  Doris support `delete statement`

For example,  ```delete table test where a = 10```

Now Doris support an asynchronous delete operation，data will be filtered in query. But Now Delete condition filter data slowly. We need a better way to do same thing.

## How column predicate replace detele condition

image

it make storage engine delete data early

### It can bring us the following benefits:

1. First, we can filter data more early, it make storage engine save mem and cpu to do unless work.

2. Secondly, segment v2 can use delete condition to do `Lazy Materization` which will reduce IO of Storage greatly.

3. Last, the data filter in column predicate it eazy for compiler to do `Query Verctorion`.

### The key idea and challenge

* Now ```delete from test where a = 1 and b = 10```.  need to change to
    ``` select * from test where a != 1 or b != 10``` 

* The key problem is to support or operation in column predicate.

### 1. Support New Interface Column Block Predicate 

We need an interface that supports `BlockColumnPredicate`, which supports  and or semantics. At the same time, it has two subclasses, one is `SingleColumnBlockPredicta`, which is a simple wrapper of the original column predict. The other is `MutilcolumnBlockPredicate`, which is a public interface and derives the subclasses of `AndBlockColumnPredicate` and `OrBlockColumnpredicate` to implement the semantics of and or

image

### 2 Support Opposite flag in column predicate

When filtering data when reading, `a! = 1` can not simply rewrite to  `a = 1`, because there may be a NULL value here.
So we need to add a flag: opposite to identify the inverse relationship.
image

### 3 Support and or interface in column predicate
`Column predicate` also supports the logical and semantic interfaces of or and, thus forming a tree structure we need.

image

#### 4: TODO
1. Remove segment v1, and no need to two way to check whether to use `delete condition`
2. Support do predicate on Bitmap and ZoneMap, So we can use index of column to do predicate on
page and segment and remove code of `delete condition` totally.

HappenLee:
1. Add BlockColumnPredicate support OR and AND column predicate in RowBlockV2
2. Support evaluate vectorization delete predicate in storage engine not in Reader in SegmentV2

## Proposed changes

fixed #5358

## Types of changes

## Checklist


morningman (approved):
LGTM

PR #5367, opened 2021-02-06
[Bug] Fix bug of NPE when replaying spark load job.

morningman:
## Proposed changes

The resourceDesc in spark load job may be null because it is not persisted.
So when replaying the job, we should check it.

## Types of changes


yangzhg (approved):
LGTM

wangbo (approved):
+1

PR #5368, opened 2021-02-06
[bug] Support if not exists in create table like stmt

xxiao2018:
## Proposed changes

Currently we support syntax `create table xx if not exists like xxx`, but `if not exists` does not work well.

## Types of changes


morningman (approved):
LGTM

PR #5385, opened 2021-02-10
[optimization] use inline optimize ExprContext::get_value

### LINKED ISSUE #5383: [optimization] Optimize ExprContext::get_value with inline
This function is simple enough, but it is not an inline function. When the root expression is a simple expression, calling get_value can see a certain overhead (very small) in the cpu profile, but in fact it is completely unnecessary here, and the compiler can take further optimization.

make this function to inline

**Additional context**
Here is my test:
```
select LO_ORDERKEY from LINE_ORDER_V2 order by LO_ORDERKEY desc limit 10;
```
| inline | no-inline | limt x |
| ------ | --------- | ------ |
| 1      | 1.05      | 10     |
| 2.88   | 2.94      | 1000   |

From these test results, the performance improvement is very small, but there is indeed a small part of the improvement

stdpain:
## Proposed changes
will close #5383 
## Types of changes

## Checklist

### DIFF
--- a/be/src/exec/base_scanner.cpp
+++ b/be/src/exec/base_scanner.cpp
@@ -19,6 +19,7 @@
 
 #include "common/logging.h"
 #include "exec/exec_node.h"
+#include "exprs/expr_context.h"
 #include "runtime/descriptors.h"
 #include "runtime/mem_tracker.h"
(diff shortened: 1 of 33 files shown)


morningman (approved):
LGTM

PR #5408, opened 2021-02-22
[Bug][Cache] Version should not be considered when find the latest partition

acelyc111:
## Proposed changes

When a table has multiple partitions, each partition has it's own
version, the version doesn't represent whether it's newer or not. When a
partition has a larger version, it may be considered as the latest one
currently, this will cause incorrect query result.
Suppose there are 2 partitions:
```
PartitionName | VisibleVersion |  VisibleVersionTime
p1            |            123 | 2021-02-17 23:31:32
p2            |             23 | 2021-02-22 11:39:19
```
Partition p1 will be considered as the lastest partition, and there is a
cache before p2's last update time, the cache will hit and return an
error result.

## Types of changes

## Checklist


marising:
LGTM，thanks

caiconghui (approved):
LGTM

PR #5410, opened 2021-02-22
Column Pruning under count start

### LINKED ISSUE #5409: Column Pruning of count(*)
# Background

When there is count(*) function in query, we only need to scan the smallest column.
But currently Doris does not column pruning in some cases.

Table structure:
```
MySQL [test]> desc base;
+-------+--------+------+-------+---------+-------+
| Field | Type   | Null | Key   | Default | Extra |
+-------+--------+------+-------+---------+-------+
| k1    | INT    | Yes  | true  | NULL    |       |
| k2    | BIGINT | Yes  | true  | NULL    |       |
| k3    | DOUBLE | Yes  | false | NULL    | NONE  |
+-------+--------+------+-------+---------+-------+
3 rows in set (0.011 sec)

```

Situation1: 
```select count(*) from base```
Only k1 is materialized.

Situation2:
```select count(*) from (select * from base) tmp```
**Column pruning becomes invalid**
All columns( k1, k2,k3 ) which belongs to base table are materialized.

# Expectation
The column pruning could be effect in situation2.
Only k1 should be materialized in situation2.

EmmyMiao87:
## Proposed changes

When there is count(*) function in query, we only need to scan the smallest column.
For example:
Query: select count(*) from (select k1, k2, k3 from base) tmp;
Only k1 which is the smallest column should be scanned.
The remaining columns (k2, k3) should be pruning.

This pr achieves this optimazation of column pruning.
Fixed #5409

Change-Id: Ie6fefe3e54f1c311bcdb35919da09668a8be5b3e

## Types of changes

## Checklist

- [x] I have created an issue on (Fix ##5409) and described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/AggregateInfo.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/AggregateInfo.java
@@ -745,8 +745,7 @@ public final class AggregateInfo extends AggregateInfoBase {
         materializedSlots_.clear();
         List<Expr> exprs = Lists.newArrayList();
         exprs.addAll(groupingExprs_);
-     
-        boolean hasCountStar = false;
+        
         int aggregateExprsSize = aggregateExprs_.size();
         int groupExprsSize = groupingExprs_.size();
         boolean isDistinctAgg = isDistinctAgg();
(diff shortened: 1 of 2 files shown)


morningman (approved):
LGTM

PR #5419, opened 2021-02-25
Remove order by from `insert into select`

### LINKED ISSUE #3803: [Proposal] remove order by for subquery in set opertion clause
In clause like 
```
select a from t order by a union all select b from t
```

The ORDER BY clause in subquery `select a from t order by a` is unmeaningful if there is no limit, so a query expression that precedes a set operator should not contain an ORDER BY clause.  We can remove the order by clause to impove the preformance.
This also applies to subquery in from clause， the ORDER BY  also need to remove.

yangzhg:
## Proposed changes

remove unnecessary order by of  `insert into select` stmt

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #3803) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/QueryStmt.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/QueryStmt.java
@@ -301,7 +301,7 @@ public abstract class QueryStmt extends StatementBase {
         sortInfo = new SortInfo(orderingExprs, isAscOrder, nullsFirstParams);
         // order by w/o limit and offset in inline views, set operands and insert statements
         // are ignored.
-        if (!hasLimit() && !hasOffset() && !analyzer.isRootAnalyzer()) {
+        if (!hasLimit() && !hasOffset() && (!analyzer.isRootAnalyzer() || fromInsert)) {
             evaluateOrderBy = false;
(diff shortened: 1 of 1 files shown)


fe/fe-core/src/main/java/org/apache/doris/analysis/QueryStmt.java
```diff
@@ -301,7 +301,7 @@ protected void createSortInfo(Analyzer analyzer) throws AnalysisException {
         sortInfo = new SortInfo(orderingExprs, isAscOrder, nullsFirstParams);
         // order by w/o limit and offset in inline views, set operands and insert statements
         // are ignored.
-        if (!hasLimit() && !hasOffset() && !analyzer.isRootAnalyzer()) {
+        if (!hasLimit() && !hasOffset() && (!analyzer.isRootAnalyzer() || fromInsert)) {
```
EmmyMiao87, line 304: May be we could just remove the order by directly. 
Because the value of Insert is actually useless after order by.

EmmyMiao87: I read it wrong, it's okay

EmmyMiao87 (approved):
LGTM

Issue #5426, opened 2021-02-25
[Enhancement] move FeMetaVersion.java from fe-common to fe-core


copperybean:
Currently, FeMetaVersion.java is in fe-common, users may forget to copy fe-common.jar when upgrading the service. It's really dangerous because the data may be corrupted and can not be recovered.

PR #5427, opened 2021-02-25
[Enhancement] Move FeMetaVersion back to fe-core to avoid invalid version upgrade (#5426)

copperybean:
Currently, java codes in Doris are separated to several modules, users may forget to copy some jar file when upgrading the service. Such as fe-common.jar, and it's really dangerous because the fe meta version may not be updated.


morningman (commented):
LGTM

yangzhg (approved):
LGTM

Issue #5443, opened 2021-03-02
[Rewrite] Rewrite from_unixtime to reduce calling this function


xinghuayu007:
**Is your feature request related to a problem? Please describe.**

from_unxitime is a cpu-exhausted function. SQL: `select filed from table where from_unixtime(field) > '2021-03-02'`, if there are one million rows of data. Function from_unixtime will be called one million times, which will make query very slow.

**Describe the solution you'd like**
 SQL: `select filed from table where from_unixtime(field) > '2021-03-02'`, 
we can rewrite this sql into `select filed from table where field > 1614614400`; 
that means we only need to transfer '2021-03-02' into timestamp, and compare field with this timestamp, it is no need to call function from_unixtime.

**Describe alternatives you've considered**

**Additional context**

PR #5444, opened 2021-03-02
[Rewrite]Rewrite from_unixtime to reduce calling this function

xinghuayu007:
## Proposed changes

from_unxitime is a cpu-exhausted function. SQL: `select filed from table where from_unixtime(field) > '2021-03-02'`, if there are one million rows of data. Function from_unixtime will be called one million times, which will make query very slow.

In issue #5443, we try to rewrite from_unixtime into timestamp to reduce calling this function. This rewriting can bring 2 times query performance improvment.

## Types of changes

## Checklist

## Further comments


fe/fe-core/src/main/java/org/apache/doris/rewrite/FromUnixTimeRule.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+        Expr right = bp.getChild(1);
+        if (!(right instanceof LiteralExpr)) {
+            return expr;
+        }
+        LiteralExpr le = (LiteralExpr) right;
+        SimpleDateFormat format =   new SimpleDateFormat( "yyyy-MM-dd" );
+        //default format is "yyyy-MM-dd HH:mm:ss"
+        if (params.exprs().size() == 1) {
+            format =   new SimpleDateFormat( "yyyy-MM-dd HH:mm:ss" );
```
morningman, line 76:
```suggestion
            format = new SimpleDateFormat( "yyyy-MM-dd HH:mm:ss" );
```

morningman:
I'm not sure whether this change can guarantee the semantic correctness.
Are there other systems with similar modifications?

Also, how much performance improvement can this change bring?

fe/fe-core/src/main/java/org/apache/doris/rewrite/FromUnixTimeRule.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.doris.analysis.LiteralExpr;
+import org.apache.doris.analysis.SlotRef;
+import org.apache.doris.catalog.Type;
+import org.apache.doris.common.AnalysisException;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public class FromUnixTimeRule implements ExprRewriteRule {
```
kangkaisen, line 36: Add some comment for how this rewrite rule work

kangkaisen:
> I'm not sure whether this change can guarantee the semantic correctness.
> Are there other systems with similar modifications?
> 
> Also, how much performance improvement can this change bring?

In, Meituan, their ad-hoc query engine do this improvement.

fe/fe-core/src/main/java/org/apache/doris/rewrite/FromUnixTimeRule.java
kangkaisen, line 36: Would better rename the class name.

xinghuayu007: ok

xinghuayu007:
> I'm not sure whether this change can guarantee the semantic correctness.
> Are there other systems with similar modifications?
> 
> Also, how much performance improvement can this change bring?

The rewriting can bring 2 times query performance improvement. The performance test is in issue: #5443

fe/fe-core/src/main/java/org/apache/doris/rewrite/RewriteFromUnixTimeRule.java
```diff
@@ -0,0 +1,107 @@
...
+        //default format is "yyyy-MM-dd HH:mm:ss"
+        if (params.exprs().size() == 1) {
+            format = new SimpleDateFormat( "yyyy-MM-dd HH:mm:ss" );
+        } else {
+            LiteralExpr fm = (LiteralExpr) params.exprs().get(1);
+            format = new SimpleDateFormat(fm.getStringValue());
+        }
+        try {
+            Date date = format.parse(le.getStringValue());
+            //it is better to add low bound 0, because when a field contains negative data like -100, it will be queried as a result
```
kangkaisen, line 90: Why do we need to add low bound or upper bound explicitly?

fe/fe-core/src/main/java/org/apache/doris/rewrite/RewriteFromUnixTimeRule.java
```diff
@@ -0,0 +1,107 @@
...
+        if (!sr.getColumn().getType().isIntegerType()) {
+            return new BoolLiteral(false);
+        }
+        Expr right = bp.getChild(1);
+        if (!(right instanceof LiteralExpr)) {
+            return expr;
+        }
+        LiteralExpr le = (LiteralExpr) right;
+        SimpleDateFormat format =   new SimpleDateFormat( "yyyy-MM-dd" );
+        //default format is "yyyy-MM-dd HH:mm:ss"
```
kangkaisen, line 81:
```suggestion
        // default format is "yyyy-MM-dd HH:mm:ss"
```

fe/fe-core/src/main/java/org/apache/doris/rewrite/RewriteFromUnixTimeRule.java
```diff
@@ -0,0 +1,107 @@
...
+        SlotRef sr = (SlotRef) paramSlot;
+        if (!sr.getColumn().getType().isIntegerType()) {
+            return new BoolLiteral(false);
+        }
+        Expr right = bp.getChild(1);
+        if (!(right instanceof LiteralExpr)) {
+            return expr;
+        }
+        LiteralExpr le = (LiteralExpr) right;
+        SimpleDateFormat format =   new SimpleDateFormat( "yyyy-MM-dd" );
```
kangkaisen, line 80: You should format the code.

fe/fe-core/src/main/java/org/apache/doris/rewrite/RewriteFromUnixTimeRule.java
xinghuayu007: rewrite sql: select * from table where from_unixtime(query_time) < '2021-03-02 12:12:23'
 into: select * from table where query_time < 1614658343

if query_time contains a data like -100, query: select * from table where query_time < 1614658343 will output -100, that is incorrect.

the timestamp range is [ 0,  253402271999],  1970-01-01 00:00:00 ~ 9999-12-31 23:59:59

kangkaisen: OK, I see.  The comment "it will be queried as a result" maybe need to change.  "it is better " should change to must.

xinghuayu007: ok

fe/fe-core/src/main/java/org/apache/doris/rewrite/RewriteFromUnixTimeRule.java
xinghuayu007: done

kangkaisen (approved):
+1

PR #5468, opened 2021-03-05
[Enhancement] Support Pallralel Merge In Exchange Node

### LINKED ISSUE #5467: [Feature] Support  Parallel Merge Sort in Exchange Node
## Motivation
Currently, Doris supports merging and sorting the data sent by the underlying sorted nodes on the exchange node to ensure that the final data is in order. This is a multi-channel merge process, but when the amount of underlying data is too large and there are too many sending nodes, the single thread merge sort will become the bottleneck of query performance.

So，We Should Support Parallel Merge in Exchange Node to speed up query. 
## Implementation

This picture show the design of parallel merge

 There we chose parallel merge, we should make thread execute more parallel to minimized the computation of top merger
 TopMerger: have child merger to supplier data
 ChildMerger: have sender queue to supplier data, each merger start a thread to merge data firstly
 SenderQueue: the data from other node
 Before parallel merge, if we have 81 sender queue, data is 1000, the computation is 1000 * log(81)
 After parallel merge, the computation is MAX(1000 * log(2), 500 * log(41))

#### How many child merger and thread
 Now we only support max 3 merge child, each child use 1 thread, because:
 we have N _sender_queue, M merge child. the best way is log(N / M) = M * log(M)
 So if N = 8, M = 2
  	 N = 81, M = 3
         N = 1024, M = 4
 normally the N is lower than 1024, so we chose 8 <= N < 81, M = 2
                                                                                     N >= 81, M = 3

## Test

3000kw data test `order by` 

| Merge Sort | Parallel Merge Sort|
| :-----: | :----: |
| 3.3s | 1.49s |

HappenLee:
## Proposed changes

Support Pallralel Merge In Exchange Node issue:#5467

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #5467 and described the bug/feature there in detail

## Further comments

### DIFF
--- a/be/src/runtime/sorted_run_merger.cc
+++ b/be/src/runtime/sorted_run_merger.cc
@@ -17,6 +17,7 @@
 
 #include "runtime/sorted_run_merger.h"
 
+#include <condition_variable>
 #include <vector>
 
 #include "exprs/expr.h"
@@ -46,10 +47,10 @@ public:
               _input_row_batch_index(-1),
               _parent(parent) {}
 
-    ~BatchedRowSupplier() {}
+    virtual ~BatchedRowSupplier() = default;
 
     // Retrieves the first batch of sorted rows from the run.
-    Status init(bool* done) {
+    virtual Status init(bool* done) {
         *done = false;
         RETURN_IF_ERROR(_sorted_run(&_input_row_batch));
         if (_input_row_batch == NULL) {
@@ -62,7 +63,7 @@ public:
 
     // Increment the current row index. If the current input batch is exhausted fetch the
     // next one from the sorted run. Transfer ownership to transfer_batch if not NULL.
-    Status next(RowBatch* transfer_batch, bool* done) {
+    virtual Status next(RowBatch* transfer_batch, bool* done) {
         DCHECK(_input_row_batch != NULL);
         ++_input_row_batch_index;
         if (_input_row_batch_index < _input_row_batch->num_rows()) {
@@ -83,7 +84,9 @@ public:
 
     TupleRow* current_row() const { return _input_row_batch->get_row(_input_row_batch_index); }
 
-private:
+    RowBatch* get_row_batch() const { return _input_row_batch; }
+
+protected:
     friend class SortedRunMerger;
 
     // The run from which this object supplies rows.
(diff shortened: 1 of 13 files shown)


be/src/runtime/sorted_run_merger.cc
```diff
@@ -99,6 +102,110 @@ class SortedRunMerger::BatchedRowSupplier {
...
+        while (true) {
+            _batch_prepared_cv.wait(lock, [this]() { return !_backup_ready.load(); });
+            if (_cancel) {
+                break;
+            }
+
+            // do merge from sender queue data
+            _status_backup = _sorted_run(&_input_row_batch_backup);
+            _backup_ready = true;
+            std::unique_ptr<NotifyCondition> notify_condition(
```
stdpain, line 198: may be we could use defer_op  instead of this ?

be/src/runtime/sorted_run_merger.cc
```diff
@@ -155,8 +263,17 @@ Status SortedRunMerger::prepare(const vector<RunBatchSupplier>& input_runs) {
...
+        auto row_batch = batched_row_supplier->get_row_batch();
+        if (row_batch != nullptr) {
+            row_batch->transfer_resource_ownership(transfer_resource_batch);
+        }
+    }
+}
+
 Status SortedRunMerger::get_next(RowBatch* output_batch, bool* eos) {
-    ScopedTimer<MonotonicStopWatch> timer(_get_next_timer);
+	ScopedTimer<MonotonicStopWatch> timer(_get_next_timer);
```
stdpain, line 276: code-format here

be/src/runtime/sorted_run_merger.cc
```diff
@@ -129,16 +236,17 @@ SortedRunMerger::SortedRunMerger(const TupleRowComparator& compare_less_than,
                                  bool deep_copy_input)
         : _compare_less_than(compare_less_than),
           _input_row_desc(row_desc),
-          _deep_copy_input(deep_copy_input) {
-    _get_next_timer = ADD_TIMER(profile, "MergeGetNext");
-    _get_next_batch_timer = ADD_TIMER(profile, "MergeGetNextBatch");
+		  _deep_copy_input(deep_copy_input) {
```
stdpain, line 239: need a code format

fe/fe-core/src/main/java/org/apache/doris/qe/SessionVariable.java
```diff
@@ -133,6 +134,9 @@
     @VariableMgr.VarAttr(name = ENABLE_SPILLING)
     public boolean enableSpilling = false;
 
+    @VariableMgr.VarAttr(name = ENABLE_EXCHANGE_NODE_PARALLEL_MERGE)
+    public boolean enableExchangeNodeParallelMerge = true;
```
morningman, line 138: default should be false?
If these are multi sort runs, but each run has a few rows, maybe this will increase the overhead.

HappenLee: yes，I will change the default value.

morningman (approved):
LGTM

PR #5452, opened 2021-03-03
[Audit][Stream Load] Support audit function for stream load

weizuo93:
## Proposed changes

Record finished  stream load job (both successful job and failed job) into audit log so that we can see when the stream load job was executed and check the details of stream load jobs.

```
2021-03-09 14:11:51,821 [stream_load] |Label=1615270311|Db=test_db_1|Table=table_test|User=root|ClientIp=xxx|Status=Success|Message=OK|Url=N/A|TotalRows=200|LoadedRows=200|FilteredRows=0|UnselectedRows=0|LoadBytes=8926|StartTime=2021-03-09 14:11:51.692|FinishTime=2021-03-09 14:11:51.819
2021-03-09 14:11:52,969 [stream_load] |Label=1615270312|Db=test_db_2|Table=table_test|User=root|ClientIp=xxx|Status=Success|Message=OK|Url=N/A|TotalRows=200|LoadedRows=200|FilteredRows=0|UnselectedRows=0|LoadBytes=8926|StartTime=2021-03-09 14:11:52.848278|FinishTime=2021-03-09 14:11:52.968169
2021-03-09 14:11:54,122 [stream_load] |Label=1615270313|Db=test_db_3|Table=table_test|User=root|ClientIp=xxx|Status=Success|Message=OK|Url=N/A|TotalRows=200|LoadedRows=200|FilteredRows=0|UnselectedRows=0|LoadBytes=8926|StartTime=2021-03-09 14:11:53.995|FinishTime=2021-03-09 14:11:54.120
2021-03-09 14:12:00,494 [stream_load] |Label=1615270320|Db=test_db_0|Table=table_test|User=root|ClientIp=xxx|Status=Success|Message=OK|Url=N/A|TotalRows=200|LoadedRows=200|FilteredRows=0|UnselectedRows=0|LoadBytes=8947|StartTime=2021-03-09 14:12:00.314894|FinishTime=2021-03-09 14:12:00.493533
2021-03-09 14:12:01,648 [stream_load] |Label=1615270321|Db=test_db_1|Table=table_test|User=root|ClientIp=xxx|Status=Success|Message=OK|Url=N/A|TotalRows=200|LoadedRows=200|FilteredRows=0|UnselectedRows=0|LoadBytes=8947|StartTime=2021-03-09 14:12:01.525|FinishTime=2021-03-09 14:12:01.649
[...]

PR #5488, opened 2021-03-09
[Feature][Stream Load] Add "show stream load" to show stream load record

weizuo93:
## Proposed changes

Add the sql of `show stream load` to show the result of recent stream load jobs. 

img_58a449aa-9d4b-4bbc-85a8-cbd64fc7659l

## Types of changes

## Checklist

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/Config.java
@@ -600,6 +600,18 @@ public class Config extends ConfigBase {
     public static int fetch_stream_load_record_interval_second = 120;
 
     /**
+     * Default max number of recent stream load record that can be stored in memory.
+     */
+    @ConfField(mutable = true, masterOnly = true)
+    public static int max_stream_load_record_size = 5000;
+
+    /**
+     * Whether to disable show stream load and clear stream load records in memory.
+     */
+    @ConfField(mutable = true, masterOnly = true)
+    public static boolean disable_show_stream_load = false;
+
+    /**
      * maximum concurrent running txn num including prepare, commit txns under a single db
      * txn manager will reject coming txns
      */
--- a/docs/en/sql-reference/sql-statements/Data Manipulation/SHOW STREAM LOAD.md
+++ b/docs/en/sql-reference/sql-statements/Data Manipulation/SHOW STREAM LOAD.md
@@ -1,0 +1,68 @@
+---
+{
+    "title": "SHOW STREAM LOAD",
+    "language": "en"
+}
+---
+
+<!-- 
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# SHOW STREAM LOAD
+## Description
+This statement is used to show the execution of the specified import task
+Grammar:
+SHOW STREAM LOAD
+[FROM both names]
+[
+WHERE
+[LABEL [ = "your_label" | LIKE "label_matcher"]]
+[STATUS = ["SUCCESS"|"FAIL"|]]
+]
+[ORDER BY ...]
+[LIMIT limit][OFFSET offset];
+
+Explain:
+1) If db_name is not specified, use the current default DB
+2) If you use LABEL LIKE, the label that matches the task contains the STREAM LOAD task of label_matcher
+3) If LABEL = is used, the specified label is matched accurately.
+4) If STATUS is specified, the STREAM LOAD status is matched
+5) Arbitrary column combinations can be sorted using ORDER BY
+6) If LIMIT is specified, the limit bar matching record is displayed. Otherwise, all of them will be displayed.
+7) If OFFSET is specified, the query results are displayed from offset. By default, the offset is 0.
+
+## example
+1. Show all STREAM LOAD tasks of default DB
+SHOW STREAM LOAD;
+
+2. Show the STREAM LOAD task of the specified db. The label contains the string "2014_01_02", showing the oldest 10
+SHOW STREAM LOAD FROM example_db WHERE LABEL LIKE "2014_01_02" LIMIT 10;
+
+3. Show the STREAM LOAD task of the specified db, specify label as "load_example_db_20140102"
+SHOW STREAM LOAD FROM example_db WHERE LABEL = "load_example_db_20140102";
+
+4. Show the STREAM LOAD task of the specified db, specify status as "success", and sort it in descending order by LoadStartTime
+SHOW STREAM LOAD FROM example_db WHERE STATUS = "success" ORDER BY LoadStartTime DESC;
+
+5. Show the STREAM LOAD task of the specified dB and sort it in descending order by LoadStartTime, and display 10 query results starting with offset 5
+SHOW STREAM LOAD FROM example_db ORDER BY LoadStartTime DESC limit 5,10;
+SHOW STREAM LOAD FROM example_db ORDER BY LoadStartTime DESC limit 10 offset 5;
+
+## keyword
+SHOW,STREAM LOAD
--- a/docs/zh-CN/sql-reference/sql-statements/Data Manipulation/SHOW STREAM LOAD.md
+++ b/docs/zh-CN/sql-reference/sql-statements/Data Manipulation/SHOW STREAM LOAD.md
@@ -1,0 +1,69 @@
+---
+{
+    "title": "SHOW STREAM LOAD",
+    "language": "zh-CN"
+}
+---
+
+<!-- 
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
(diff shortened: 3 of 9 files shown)


be/src/http/action/stream_load.cpp
```diff
@@ -132,9 +134,19 @@ void StreamLoadAction::handle(HttpRequest* req) {
     str = str + '\n';
     HttpChannel::send_reply(req, str);
 
+    auto stream_load_record = StorageEngine::instance()->get_stream_load_record();
+    if (stream_load_record != nullptr) {
+        auto st = stream_load_record->put(ToStringFromUnixMicros(ctx->start_micros + ctx->load_cost_micros) + "_" + ctx->label, str);
+        if (st.ok()) {
+            LOG(INFO) << "put stream_load_record rocksdb successfully. label: " << ctx->label;
```
acelyc111, line 141: Better to log the whole key, then we can check it in rocksdb.

be/src/http/action/stream_load.cpp
```diff
@@ -132,9 +134,19 @@ void StreamLoadAction::handle(HttpRequest* req) {
...
+        }
+    } else {
+        LOG(WARNING) << "stream_load_record is null.";
```
acelyc111, line 144:
```suggestion
        LOG(WARNING) << "put stream_load_record rocksdb failed, status = " << st.ToString();
```
And log the key.

be/src/http/action/stream_load.cpp
```diff
@@ -200,6 +212,16 @@ int StreamLoadAction::on_header(HttpRequest* req) {
...
         HttpChannel::send_reply(req, str);
+
+        auto stream_load_record = StorageEngine::instance()->get_stream_load_record();
+        if (stream_load_record != nullptr) {
+            auto st = stream_load_record->put(ToStringFromUnixMicros(ctx->start_micros + ctx->load_cost_micros) + "_" + ctx->label, str);
+            if (st.ok()) {
+                LOG(INFO) << "put stream_load_record rocksdb successfully. label: " << ctx->label;
+            }
+        } else {
+            LOG(WARNING) << "stream_load_record is null.";
```
acelyc111, line 223: Same as above.

be/src/olap/olap_server.cpp
```diff
@@ -118,6 +118,13 @@ Status StorageEngine::start_bg_threads() {
         LOG(INFO) << "path scan/gc threads started. number:" << get_stores().size();
     }
 
+    // stream load record clean thread
+    RETURN_IF_ERROR(Thread::create(
```
acelyc111, line 122: No need to create a thread to gc expired data, you can create a Compaction-Filter on this rocksdb instance.
https://github.com/facebook/rocksdb/wiki/Compaction-Filter

be/src/runtime/stream_load/stream_load_record.cpp
```diff
@@ -0,0 +1,137 @@
...
+        LOG(WARNING) << "rocks db put key:" << key << " failed, reason:" << s.ToString();
+        return Status::InternalError("Stream load record rocksdb put failed");
+    }
+    return Status::OK();
+}
+
+Status StreamLoadRecord::get_batch(const std::string& start, const int batch_size, std::map<std::string, std::string> &stream_load_records) {
+    rocksdb::ColumnFamilyHandle* handle = _handles[1];
+    std::unique_ptr<rocksdb::Iterator> it(_db->NewIterator(rocksdb::ReadOptions(), handle));
+    if (start == "") {
```
acelyc111, line 86:
```suggestion
    if (start.empty()) {
```

be/src/runtime/stream_load/stream_load_record.cpp
```diff
@@ -0,0 +1,137 @@
...
+        it->SeekToFirst();
+    } else {
+        it->Seek(start);
+        rocksdb::Status status = it->status();
+        if (!status.ok()) {
+            it->SeekToFirst();
```
acelyc111, line 92: why seek to first in this case?

be/src/runtime/stream_load/stream_load_record.cpp
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+    rocksdb::Status status = it->status();
+    if (!status.ok()) {
+        LOG(WARNING) << "rocksdb seek failed. reason:" << status.ToString();
+        return Status::InternalError("Stream load record rocksdb seek failed");
+    }
+    int num = 0;
+    for (it->Next(); it->Valid(); it->Next()) {
```
acelyc111, line 101: Have you skipped the first k-v? should it be 
```
for (; it->Valid(); it->Next()) {
```
?

be/src/runtime/stream_load/stream_load_record.h
```diff
@@ -0,0 +1,57 @@
...
+
+    Status clean_expired_stream_load_record();
+
+private:
+    std::string _root_path;
+    rocksdb::DB* _db;
+    std::vector<rocksdb::ColumnFamilyHandle*> _handles;
+
+    enum ColumnFamilyIndex {
+        DEFAULT_COLUMN_FAMILY_INDEX = 0,
```
acelyc111, line 49: where does the default column family used for?

be/src/olap/storage_engine.cpp
```diff
@@ -225,6 +226,35 @@ Status StorageEngine::_init_store_map() {
...
+    if (!st.ok()) {
+        LOG(WARNING) << "status=" << st.to_string();
+    }
+
+    return Status::OK();
+}
+
+Status StorageEngine::_init_stream_load_record() {
+    std::string stream_load_record_path;
+    if (_store_map.size() > 0) {
```
acelyc111, line 240: Is it valid when _store_map is empty? Would it be better to explicit define the path in config file?

be/src/runtime/stream_load/stream_load_context.cpp
```diff
@@ -89,6 +103,91 @@ std::string StreamLoadContext::to_json() const {
...
+void StreamLoadContext::parse_stream_load_record(std::string stream_load_record, TStreamLoadRecord& stream_load_item) {
+
+    rapidjson::Document document;
+    if (document.Parse(stream_load_record.data()).HasParseError()) {
+        return;
+    }
+
+    if (document.HasMember("Label")) {
+        const rapidjson::Value& label = document["Label"];
+        stream_load_item.__set_label(label.GetString());
```
acelyc111, line 115:
```suggestion
        stream_load_item.__set_label(document["Label"].GetString());
```

kangkaisen:
@weizuo93 Hi, Thanks for your contribution.
I don't understand why we need to store the stream load meta both in FE and BE.  why we couldn't get stream load info from FE meta?

weizuo93:
> @weizuo93 Hi, Thanks for your contribution.
> I don't understand why we need to store the stream load meta both in FE and BE. why we couldn't get stream load info from FE meta?
Our goal is to store the stream load meta in FE. The reason why the stream load meta is stored on BE first is to send the meta to FE in bulk to avoid calling RPC every time the stream load finished. After a certain period of time, stream load meta will be removed from BE.

weizuo93:
> @weizuo93 Hi, Thanks for your contribution.
> I don't understand why we need to store the stream load meta both in FE and BE. why we couldn't get stream load info from FE meta?

Our goal is to store the stream load meta in FE. The reason why the stream load meta is stored on BE first is to send the meta to FE in bulk to avoid calling RPC every time the stream load finished. After a certain period of time, stream load meta will be removed from BE.

kangkaisen:
@weizuo93 
Hi, I mean In `LoadManager`,  `dbIdToLabelToLoadJobs` already has the stream load join info, why we still need to store stream load job meta in BE?  in other words, I think we don't need to store stream load job meta in BE.

weizuo93:
> join

@kangkaisen As far as I know, there is no info about stream load in `LoadManager`and `dbIdToLabelToLoadJobs`. `LoadManager` only manages bulk load.

weizuo93:
> @weizuo93
> Hi, I mean In `LoadManager`, `dbIdToLabelToLoadJobs` already has the stream load join info, why we still need to store stream load job meta in BE? in other words, I think we don't need to store stream load job meta in BE.

@kangkaisen As far as I know, there is no info about stream load in LoadManagerand dbIdToLabelToLoadJobs. LoadManager only manages bulk load.

weizuo93:
> @weizuo93 You are right. I am wrong. This PR could go ahead. (But store some load job meta in FE and store some load job meta in BE, which seems not very good)

@kangkaisen 
Thanks for your comment. There is code overlap between this PR and PR #5452(https://github.com/apache/incubator-doris/pull/5452). After discussion with Mingyu Chen, we first push forward PR #5452. About storing stream load meta in BE, it's a tradeoff. If BE reports the stream load result to FE  when each stream load job finished, it may result in too much stress for FE RPC. So the stream load results are temporarily stored on BE and then sent to FE by batch. After finished data sending, stream load meta will be removed from BE soon. Do you have any better idea?

kangkaisen:
@weizuo93 Ok, Thanks for your explain. which is OK for me

yangzhg:
`show stream load` is better, like `show routine load`

weizuo93:
> `show stream load` is better, like `show routine load`

OK, thank you.

yangzhg (dismissed):
LGTM

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -605,6 +605,12 @@
     @ConfField(mutable = true, masterOnly = true)
     public static int max_stream_load_record_size = 5000;
 
+    /**
+     * Whether to close show stream load and clear stream load records in memory.
+     */
+    @ConfField(mutable = true, masterOnly = true)
+    public static boolean close_show_stream_load = false;
```
yangzhg, line 612: `disable_show_stream_load` is better

weizuo93: > `disable_show_stream_load` is better

OK, I have modified it.

yangzhg (approved):
LGTM

PR #5489, opened 2021-03-09
[Outfile] Support exporting query result to local disk

### LINKED ISSUE #5490: [Outfile] Support export query result to local disk

Currently, we only support writing query results to remote storage through the broker, but some users want to export them directly to the local and process them by themselves. So I plan to support the function of exporting to local.
At the same time, the user wants to be able to write a mark file after the export is completed to notify the downstream data that the export has been successful.

Because I cannot control which node the query result is output from, I also modified the return result of the outfile statement. The returned result shows the ip of the machine where the local file is located.

morningman:
## Proposed changes

1.
User can export query result to local disk like:

`select * from tbl into outfile ("file:///disk1/result_");`

And modify the return result to show the details of export:

```
mysql> select * from tbl1 limit 10 into outfile "file:///home/work/path/result_";
+------------+-----------+----------+--------------+
| FileNumber | TotalRows | FileSize | URL          |
+------------+-----------+----------+--------------+
|          1 |         2 |        8 | 192.168.1.10 |
+------------+-----------+----------+--------------+
```

2.
Support create a mark file after export successfully finished.

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #5490) and described the bug/feature there in detail

### DIFF
--- a/be/src/runtime/file_result_writer.cpp
+++ b/be/src/runtime/file_result_writer.cpp
@@ -23,11 +23,15 @@
 #include "exprs/expr.h"
 #include "exprs/expr_context.h"
 #include "gen_cpp/PaloInternalService_types.h"
+#include "runtime/buffer_control_block.h"
 #include "runtime/primitive_type.h"
 #include "runtime/row_batch.h"
 #include "runtime/runtime_state.h"
 #include "runtime/tuple_row.h"
+#include "service/backend_options.h"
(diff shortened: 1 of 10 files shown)


yangzhg (dismissed):
LGTM

yangzhg (approved):
Lgtm

PR #5518, opened 2021-03-13
Use O_SYNC instead of O_DIRECT to be fs agnostic.

amosbird:
## Proposed changes

Some fs might not support O_DIRECT and O_SYNC is semantically the same to be used for disk checking.


morningman (approved):
LGTM

PR #5527, opened 2021-03-16
 [Bug] Fix bug that the stale rowset file will not be deleted

### LINKED ISSUE #5526: [Bug] The stale rowset files are not deleted
**Describe the bug**
We found that in some cases, a large number of rowsets that have expired for a long time
(such as stale rowsets after compaction) remain on the disk.
These rowsets will not be deleted, which will occupy disk space.

And If we restart BE, these rowsets will be deleted.

**Reason**

The current compaction logic is that after the compaction is completed, the old rowset will be put into `_stale_rowset_meta_map` and wait to be deleted. The following case will trigger the problem:

1.
Current data version:
rowsets: [0-217][218-218 A]
Where A is the rowset id of [218-218] rowset.

2.
After rowset [218-218] triggers cumulative compaction:
rowsets: [0-217][218-218 B]
stale_rowsets: [218-218 A]

3.
Then after base compaction is triggered:
rowsets: [0-218]
stale_rowsets: [218-218 A], [0-217], [218-218 B]

Because _stale_rowset_meta_map uses the `Version` of Rowset as the key.
This will cause [218-218 B] to directly overwrite [218-218 A],
resulting in the rowset [218-218 A] cannot be accessed and cannot be deleted.

morningman:
## Proposed changes

1. If cumulative compaction compact only one rowset, the old rowset will not be put into `stale_rowset_meta_map`

2. Show rowset id in `/api/compaction/show`

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #5526 ) and described the bug/feature there in detail

### DIFF
--- a/be/src/olap/schema_change.cpp
+++ b/be/src/olap/schema_change.cpp
@@ -1478,7 +1478,8 @@ OLAPStatus SchemaChangeHandler::_do_process_alter_tablet_v2(const TAlterTabletRe
                 rowsets_to_delete.push_back(rowset);
             }
         }
-        new_tablet->modify_rowsets(std::vector<RowsetSharedPtr>(), rowsets_to_delete);
+        std::vector<RowsetSharedPtr> empty_vec;
+        new_tablet->modify_rowsets(empty_vec, rowsets_to_delete);
         // inherit cumulative_layer_point from base_tablet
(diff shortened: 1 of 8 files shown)


yangzhg (approved):
LGTM

PR #5534, opened 2021-03-18
[Bug] Fix bug that the last column may be null when using multibytes separator

### LINKED ISSUE #5537: [Bug] When using multi bytes separator in load process, the last column can not be read
**Describe the bug**
When using multi bytes separator in load process, such as "column_separator: AAAA"
the last column of source file can not be read and NULL will be filled into table.

morningman:
## Proposed changes

When using multibytes separator in load process, the last column of a line may be read as null.

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #5537 ) and described the bug/feature there in detail


yangzhg (approved):
+1

PR #5597, opened 2021-04-03
[Enhance] Support to configure RocksDB to improve tablet meta store

acelyc111:
## Proposed changes

Supported options:
- rocksdb_thread_count: Thread count of RocksDB uses for background flush and compaction
- rocksdb_block_cache_mb: Block cache size
- rocksdb_compression_type: compression type, now support LZ4 and snappy

And also do some refactor:
- Remove some  deprecated meta store related code which never used now
- Call OptimizeForPointLookup() to optimize options include bloom filter and block cache

## Types of changes

## Checklist

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -246,8 +246,8 @@ CONF_mInt32(base_compaction_write_mbytes_per_sec, "5");
 // size_based policy, a optimization version of cumulative compaction, targeting the use cases requiring
 // lower write amplification, trading off read amplification and space amplification.
 CONF_mString(cumulative_compaction_policy, "size_based");
-CONF_Validator(cumulative_compaction_policy, [](const std::string config) -> bool {
-    return config == "size_based" || config == "num_based";
+CONF_Validator(cumulative_compaction_policy, [](const std::string& value) -> bool {
+  return value == "size_based" || value == "num_based";
 });
 
 // In size_based policy, output rowset of cumulative compaction total disk size exceed this config size,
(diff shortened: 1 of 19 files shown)


be/src/common/config.h
```diff
@@ -516,10 +514,23 @@ CONF_Int32(flush_thread_num_per_store, "2");
 // config for tablet meta checkpoint
 CONF_mInt32(tablet_meta_checkpoint_min_new_rowsets_num, "10");
 CONF_mInt32(tablet_meta_checkpoint_min_interval_secs, "600");
+// Thread count of RocksDB uses for background flush and compaction, -1 means the number of cores.
+CONF_mInt32(rocksdb_thread_count, "-1");
+CONF_mInt32(rocksdb_block_cache_mb, "1024");
+CONF_Validator(rocksdb_block_cache_mb, [](int value) -> bool {
+  return value > 0;
+});
+CONF_String(rocksdb_compression_type, "LZ4");
```
morningman, line 523: What is the difference between LZ4 and snappy? Is it necessary to config this?

be/src/common/config.h
```diff
@@ -516,10 +514,23 @@ CONF_Int32(flush_thread_num_per_store, "2");
 // config for tablet meta checkpoint
 CONF_mInt32(tablet_meta_checkpoint_min_new_rowsets_num, "10");
 CONF_mInt32(tablet_meta_checkpoint_min_interval_secs, "600");
+// Thread count of RocksDB uses for background flush and compaction, -1 means the number of cores.
+CONF_mInt32(rocksdb_thread_count, "-1");
+CONF_mInt32(rocksdb_block_cache_mb, "1024");
```
morningman, line 519: Is 1024MB the default value of rocksdb cache? Maybe it is too big?
BTW, `rocksdb_thread_count` and `rocksdb_block_cache_mb` are not mutable configuration, use `CONF_Int32` instead.

be/src/common/config.h
acelyc111: > Use options.compression to specify the compression to use. By default it is Snappy. We believe LZ4 is almost always better than Snappy. We leave Snappy as default to avoid unexpected compatibility problems to previous users.

https://github.com/facebook/rocksdb/wiki/Compression

We encountered a problem that snappy decompress cost too much time in product environment.
Of course, the root cause is that the rocksdb value is too large, some tablet meta which have thounds of rowsets may reach hundreds of MB. But I think we can improve it by use LZ4.

be/src/common/config.h
acelyc111: The default cache is 8M which is too small.
I've updated code to make the block cache shared by all rocksdb instances, and it's capacity is 64M now.

PR #5608, opened 2021-04-07
[Bug] Fix bug that isPreAggregation is incorrectly set

### LINKED ISSUE #5609: [Bug] BE crash on SegmentIterator
**Describe the bug**
BE crash with error:

image

**To Reproduce**

Hard to reproduce.
The direct reason is because the read parameters passed by the BE side does not contain columns
related to the deletion conditions, which results in the corresponding column iterator being null when
preparing to filter the data according to the deletion conditions.

The root cause is that the FE side does not correctly set the `isPreAggregate` attribute during query planning,
which causes the BE side to skip the delete condition column by mistake.

 - Version [0.14.7]

morningman:
## Proposed changes

1. The MaterializedViewSelector should be reset for each scan node
2. On the BE side, columns with delete conditions must be added to the return column.

## Types of changes

## Checklist

- [ ] I have created an issue on (Fix #5609 ) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/be/src/olap/reader.cpp
+++ b/be/src/olap/reader.cpp
@@ -458,7 +458,8 @@ OLAPStatus Reader::_init_params(const ReaderParams& read_params) {
 OLAPStatus Reader::_init_return_columns(const ReaderParams& read_params) {
     if (read_params.reader_type == READER_QUERY) {
         _return_columns = read_params.return_columns;
-        if (!_delete_handler.empty() && read_params.aggregation) {
+        if (!_delete_handler.empty()) {
+            // We need to fetch columns which there are deletion conditions on them.
             set<uint32_t> column_set(_return_columns.begin(), _return_columns.end());
             for (const auto& conds : _delete_handler.get_delete_conditions()) {
                 for (const auto& cond_column : conds.del_cond->columns()) {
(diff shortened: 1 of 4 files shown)


be/src/olap/reader.cpp
```diff
@@ -458,7 +458,8 @@ OLAPStatus Reader::_init_params(const ReaderParams& read_params) {
 OLAPStatus Reader::_init_return_columns(const ReaderParams& read_params) {
     if (read_params.reader_type == READER_QUERY) {
         _return_columns = read_params.return_columns;
-        if (!_delete_handler.empty() && read_params.aggregation) {
```
morningman, line 461: For reviewer:
Remove `read_params.aggregation` because this is just a tricky optimization.
That is, if `aggregation` is false, than all key columns has already been added to the `_return_columns`.
And delete conditions can only be applied on key columns, so no need to iterate and add them again.

But this is not always right, because for DUPLICATION KEY table, the delete conditions can be applied on any columns. Although the query optimizer will guarantee that for  DUPLICATION KEY table, the aggregation is always true, but we should not depend on it.

EmmyMiao87 (approved):
LGTM

PR #5618, opened 2021-04-09
[Compaction] Use separate thread pool for base and cumulative compaction

weizuo93:
## Proposed changes

We hope that `cumulative compaction` can be implemented as soon as possible, but `base compaction` tasks are likely to  execute for a long time and occupy most of the threads in thread pool, which may result in high CPU load. 

This patch submits `cumulative compaction` and `base compaction` into separate thread pool. So we have more flexibility to adjust the execution of `cumulative compaction` and `base compaction`according to the status of cluster.
## Types of changes

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -283,8 +283,10 @@ CONF_mInt32(cumulative_compaction_skip_window_seconds, "30");
 CONF_mInt64(min_compaction_failure_interval_sec, "600"); // 10 min
 
 // This config can be set to limit thread number in compaction thread pool.
-CONF_mInt32(min_compaction_threads, "10");
-CONF_mInt32(max_compaction_threads, "10");
+CONF_mInt32(min_base_compaction_threads, "10");
+CONF_mInt32(max_base_compaction_threads, "10");
+CONF_mInt32(min_cumulative_compaction_threads, "10");
+CONF_mInt32(max_cumulative_compaction_threads, "10");
 
 // The upper limit of "permits" held by all compaction tasks. This config can be set to limit memory consumption for compaction.
 CONF_mInt64(total_permits_for_compaction_score, "10000");
@@ -293,7 +295,8 @@ CONF_mInt64(total_permits_for_compaction_score, "10000");
 CONF_mInt32(generate_compaction_tasks_min_interval_ms, "10")
 
 // Compaction task number per disk.
-CONF_mInt32(compaction_task_num_per_disk, "2");
+CONF_mInt32(max_base_compaction_task_num_per_disk, "2");
+CONF_mInt32(max_cumulative_compaction_task_num_per_disk, "2");
 
 // How many rounds of cumulative compaction for each round of base compaction when compaction tasks generation.
 CONF_mInt32(cumulative_compaction_rounds_for_each_base_compaction_round, "9");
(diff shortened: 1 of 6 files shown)


be/src/common/config.h
```diff
@@ -283,8 +283,10 @@ CONF_mInt32(cumulative_compaction_skip_window_seconds, "30");
 CONF_mInt64(min_compaction_failure_interval_sec, "600"); // 10 min
 
 // This config can be set to limit thread number in compaction thread pool.
-CONF_mInt32(min_compaction_threads, "10");
-CONF_mInt32(max_compaction_threads, "10");
+CONF_mInt32(min_base_compaction_threads, "10");
+CONF_mInt32(max_base_compaction_threads, "10");
+CONF_mInt32(min_cumulative_compaction_threads, "10");
+CONF_mInt32(max_cumulative_compaction_threads, "10");
```
acelyc111, line 289: They are not mutable, use CONF_Int32 instead.

be/src/olap/tablet_manager.cpp
```diff
@@ -693,12 +694,16 @@ TabletSharedPtr TabletManager::find_best_tablet_to_compaction(
...
+                std::set<TTabletId>::iterator it_tablet =
+                        (*tablet_submitted_base_compaction).find(tablet_ptr->tablet_id());
+                if (it_tablet != (*tablet_submitted_base_compaction).end()) {
                     continue;
                 }
+                it_tablet = (*tablet_submitted_cumulative_compaction).find(tablet_ptr->tablet_id());
+                if (it_tablet != (*tablet_submitted_cumulative_compaction).end()) {
+                    continue;
+                }
+
```
acelyc111, line 706: You can use `ContainsKey` in be/src/gutil/map-util.h instead to simplify code.

be/src/olap/storage_engine.h
```diff
@@ -334,12 +336,15 @@ class StorageEngine {
...
+    std::unique_ptr<ThreadPool> _cumulative_compaction_thread_pool;
 
     CompactionPermitLimiter _permit_limiter;
 
-    std::mutex _tablet_submitted_compaction_mutex;
-    std::map<DataDir*, vector<TTabletId>> _tablet_submitted_compaction;
+    RWMutex _tablet_submitted_base_compaction_mutex;
+    std::map<DataDir*, std::set<TTabletId>> _tablet_submitted_base_compaction;
+    RWMutex _tablet_submitted_cumulative_compaction_mutex;
+    std::map<DataDir*, std::set<TTabletId>> _tablet_submitted_cumulative_compaction;
```
acelyc111, line 347: How about use a two elements array, one for base compaction, the other for cumulative compaction, to simplify code?

be/src/common/config.h
weizuo93: > They are not mutable, use CONF_Int32 instead.

OK, thanks.

be/src/olap/tablet_manager.cpp
weizuo93: > You can use `ContainsKey` in be/src/gutil/map-util.h instead to simplify code.

We need to judge whether an element exists in `set`rather than `map`.

be/src/olap/storage_engine.h
weizuo93: > How about use a two elements array, one for base compaction, the other for cumulative compaction, to simplify code?

OK, Thank you.

Issue #5621, opened 2021-04-09
[Improvement] using KeyMayExist instead of Get when visit rocksdb


wangbo:
Acoording to [db.h](https://github.com/facebook/rocksdb/blob/v5.12.4/include/rocksdb/db.h) comment
KeyMayExist is lighter-weight than Get
```
// If the key definitely does not exist in the database, then this method
  // returns false, else true. If the caller wants to obtain value when the key
  // is found in memory, a bool for 'value_found' must be passed. 'value_found'
  // will be true on return if value has been set properly.
  // This check is potentially lighter-weight than invoking DB::Get(). One way
  // to make this lighter weight is to avoid doing any IOs.
  // Default implementation here returns true and sets 'value_found' to false
  virtual bool KeyMayExist(const ReadOptions& /*options*/,
                           ColumnFamilyHandle* /*column_family*/,
                           const Slice& /*key*/, std::string* /*value*/,
                           bool* value_found = nullptr) {
    if (value_found != nullptr) {
      *value_found = false;
    }
    return true;
  }
```

PR #5622, opened 2021-04-09
(#5621) using KeyMayExist instead of Get when visit rocksdb

wangbo:
## Proposed changes

using KeyMayExist instead of Get when visit rocksdb.
see #5621

### DIFF
--- a/be/src/olap/rowset/rowset_meta_manager.cpp
+++ b/be/src/olap/rowset/rowset_meta_manager.cpp
@@ -38,11 +38,7 @@ bool RowsetMetaManager::check_rowset_meta(OlapMeta* meta, TabletUid tablet_uid,
                                           const RowsetId& rowset_id) {
     std::string key = ROWSET_PREFIX + tablet_uid.to_string() + "_" + rowset_id.to_string();
     std::string value;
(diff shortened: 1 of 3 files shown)


be/src/olap/rowset/rowset_meta_manager.cpp
```diff
@@ -38,11 +38,7 @@ bool RowsetMetaManager::check_rowset_meta(OlapMeta* meta, TabletUid tablet_uid,
                                           const RowsetId& rowset_id) {
     std::string key = ROWSET_PREFIX + tablet_uid.to_string() + "_" + rowset_id.to_string();
     std::string value;
```
acelyc111, line 40: No need to get value?

wangbo: For doris, no need to get a value kere.
But rocksdb's ```KeyMayExist``` need a value.

acelyc111: Yes,you are right

acelyc111 (approved):

caiconghui (approved):
+1

PR #5624, opened 2021-04-09
[Docs] Update README.md about GCC version

weizuo93:
## Proposed changes

`BE`code can compile pass using `GCC 7.3.0`, but  compiling `BE UT`code will encounter error using `GCC 7.3.0`.  After upgrade gcc version to `10.2.1` which is the same version as in docker image, `BE UT` can be compiled successfully.

## Types of changes


morningman (approved):
LGTM

PR #5651, opened 2021-04-14
[Compaction] Support to switching cumulative compaction policy dynamically

weizuo93:
## Proposed changes

Support to switching cumulative compaction policy between`num_based` and `size_based` dynamically without restarting BE.

## Types of changes

### DIFF
--- a/be/src/common/configbase.cpp
+++ b/be/src/common/configbase.cpp
@@ -41,6 +41,8 @@ std::map<std::string, std::string>* full_conf_map = nullptr;
 
 std::mutex custom_conf_lock;
 
+std::mutex mutable_string_config_lock;
+
 // trim string
 std::string& trim(std::string& s) {
     // rtrim
@@ -416,11 +418,18 @@ Status set_config(const std::string& field, const std::string& value, bool need_
     UPDATE_FIELD(it->second, value, int32_t, need_persist);
     UPDATE_FIELD(it->second, value, int64_t, need_persist);
     UPDATE_FIELD(it->second, value, double, need_persist);
+    {
+        // add lock to ensure thread safe
+        std::lock_guard<std::mutex> lock(mutable_string_config_lock);
+        UPDATE_FIELD(it->second, value, std::string, need_persist);
+    }
 
     // The other types are not thread safe to change dynamically.
     return Status::NotSupported(strings::Substitute(
             "'$0' is type of '$1' which is not support to modify", field, it->second.type));
 }
+
+std::mutex* get_mutable_string_config_lock() { return &mutable_string_config_lock; }
 
 } // namespace config
 } // namespace doris
--- a/be/src/olap/olap_server.cpp
+++ b/be/src/olap/olap_server.cpp
@@ -64,9 +64,6 @@ Status StorageEngine::start_bg_threads() {
         data_dirs.push_back(tmp_store.second);
     }
 
-    // check cumulative compaction config
-    _check_cumulative_compaction_config();
-
     int32_t max_thread_num = config::max_compaction_threads;
     int32_t min_thread_num = config::min_compaction_threads;
     ThreadPoolBuilder("CompactionTaskThreadPool")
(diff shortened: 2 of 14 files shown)


be/src/common/configbase.cpp
```diff
@@ -338,7 +340,11 @@ bool init(const char* conf_file, bool fill_conf_map, bool must_exist, bool set_t
         SET_FIELD(it.second, int32_t, fill_conf_map, set_to_default);
         SET_FIELD(it.second, int64_t, fill_conf_map, set_to_default);
         SET_FIELD(it.second, double, fill_conf_map, set_to_default);
-        SET_FIELD(it.second, std::string, fill_conf_map, set_to_default);
+        {
+            // add lock to ensure thread safe
+            std::lock_guard<std::mutex> lock(mutable_string_config_lock);
+            SET_FIELD(it.second, std::string, fill_conf_map, set_to_default);
+        }
```
acelyc111, line 347: `init` is called in main when no other threads created, is it needed to use lock?

be/src/olap/tablet.cpp
```diff
@@ -714,7 +718,7 @@ bool Tablet::can_do_compaction() {
     return true;
 }
 
-const uint32_t Tablet::calc_compaction_score(CompactionType compaction_type) const {
+const uint32_t Tablet::calc_compaction_score(CompactionType compaction_type) {
```
acelyc111, line 721:
```suggestion
uint32_t Tablet::calc_compaction_score(CompactionType compaction_type) {
```

be/src/olap/tablet.h
```diff
@@ -162,7 +162,7 @@ class Tablet : public BaseTablet {
 
     // operation for compaction
     bool can_do_compaction();
-    const uint32_t calc_compaction_score(CompactionType compaction_type) const;
+    const uint32_t calc_compaction_score(CompactionType compaction_type);
```
acelyc111, line 165:
```suggestion
   uint32_t calc_compaction_score(CompactionType compaction_type);
```

be/src/olap/tablet.cpp
```diff
@@ -88,6 +88,10 @@ OLAPStatus Tablet::_init_once_action() {
             << ", version_size=" << _tablet_meta->version_count();
 
     // init cumulative compaction policy by type
+    if (_cumulative_compaction_type == "") {
```
acelyc111, line 91: Now _cumulative_compaction_policy will be created in _calc_cumulative_compaction_score, is it needed to create it here initially?

be/src/olap/olap_server.cpp
```diff
@@ -205,29 +205,32 @@ void StorageEngine::_disk_stat_monitor_thread_callback() {
 }
 
 void StorageEngine::_check_cumulative_compaction_config() {
-    std::string cumulative_compaction_type = config::cumulative_compaction_policy;
+    std::string cumulative_compaction_type = "";
```
acelyc111, line 208: Most of these configs are mutable, check them firstly here may be useless. A better way is to check them when any of them are mutated?

be/src/common/configbase.cpp
weizuo93: > `init` is called in main when no other threads created, is it needed to use lock?

Thanks, you are right.

be/src/olap/tablet.cpp
weizuo93: > Now _cumulative_compaction_policy will be created in _calc_cumulative_compaction_score, is it needed to create it here initially?

I also konw that we don't need to execute`create_cumulative_compaction_policy()`here, but UT will have some problems if we remove it. Next, I will refacor UT.

be/src/olap/olap_server.cpp
weizuo93: > Most of these configs are mutable, check them firstly here may be useless. A better way is to check them when any of them are mutated?

OK.

be/src/olap/tablet.cpp
weizuo93: > 

OK.

be/src/olap/tablet.h
weizuo93: > 

OK.

be/src/olap/tablet.cpp
```diff
@@ -725,7 +727,23 @@ const uint32_t Tablet::calc_compaction_score(CompactionType compaction_type) con
     }
 }
 
-const uint32_t Tablet::_calc_cumulative_compaction_score() const {
+const uint32_t Tablet::_calc_cumulative_compaction_score() {
+#ifndef BE_TEST
+    std::string current_policy = "";
+    {
+        std::lock_guard<std::mutex> lock(*config::get_mutable_string_config_lock());
```
morningman, line 734: It's better to check the config at the beginning of each `_compaction_tasks_producer_callback` loop,
instead of check it for each tablet.

weizuo93: > It's better to check the config at the beginning of each `_compaction_tasks_producer_callback` loop,
> instead of check it for each tablet.

Thanks, it sounds reasonable.

morningman (dismissed):
LGTM

morningman (commented):
LGTM

morningman (approved):
LGTM

PR #5654, opened 2021-04-14
[Optimize] Make tablet meta checkpoint to be threadpool model

xinghuayu007:
## Proposed changes

Currently Tablet meta checkpoint is a memory-exhausted operation. If a host has 12 disks, it will start 12 threads to do tablet meta checkpoint. In our experience, the data size of one tablet can be as high as 2G. If 12 threads do the checkpoint at the same time, it maybe cause OOM.
Therefore, this PR try to solve this problem. Firstly, it only start one thread to produce table meta checkpoint tasks. Secondly, it creates a threadpool to handle these tasks. You can configure the size of the thread pool to control the parrallelism  in case of OOM. It is a prducer-customer model.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -286,6 +286,9 @@ CONF_mInt64(min_compaction_failure_interval_sec, "600"); // 10 min
 CONF_mInt32(min_compaction_threads, "10");
 CONF_mInt32(max_compaction_threads, "10");
 
+// Thread count to do tablet meta checkpoint, -1 means use the data directories count.
+CONF_Int32(max_meta_checkpoint_threads, "-1");
+
 // The upper limit of "permits" held by all compaction tasks. This config can be set to limit memory consumption for compaction.
 CONF_mInt64(total_permits_for_compaction_score, "10000");
 
@@ -516,6 +519,7 @@ CONF_Int32(flush_thread_num_per_store, "2");
 // config for tablet meta checkpoint
 CONF_mInt32(tablet_meta_checkpoint_min_new_rowsets_num, "10");
 CONF_mInt32(tablet_meta_checkpoint_min_interval_secs, "600");
+CONF_Int32(generate_tablet_meta_checkpoint_tasks_interval_secs, "600");
 
 // config for default rowset type
 // Valid configs: ALPHA, BETA
--- a/be/src/olap/olap_server.cpp
+++ b/be/src/olap/olap_server.cpp
@@ -81,16 +81,19 @@ Status StorageEngine::start_bg_threads() {
             &_compaction_tasks_producer_thread));
     LOG(INFO) << "compaction tasks producer thread started";
 
-    // tablet checkpoint thread
-    for (auto data_dir : data_dirs) {
-        scoped_refptr<Thread> tablet_checkpoint_thread;
-        RETURN_IF_ERROR(Thread::create(
-                "StorageEngine", "tablet_checkpoint_thread",
-                [this, data_dir]() { this->_tablet_checkpoint_callback(data_dir); },
-                &tablet_checkpoint_thread));
-        _tablet_checkpoint_threads.emplace_back(tablet_checkpoint_thread);
+    int32_t max_checkpoint_thread_num = config::max_meta_checkpoint_threads;
+    if (max_checkpoint_thread_num < 0) {
+        max_checkpoint_thread_num = data_dirs.size();
     }
-    LOG(INFO) << "tablet checkpoint thread started";
+    ThreadPoolBuilder("TabletMetaCheckpointTaskThreadPool")
+            .set_max_threads(max_checkpoint_thread_num)
+            .build(&_tablet_meta_checkpoint_thread_pool);
+
+    RETURN_IF_ERROR(Thread::create(
+            "StorageEngine", "tablet_checkpoint_tasks_producer_thread",
+            [this, data_dirs]() { this->_tablet_checkpoint_callback(data_dirs); },
+            &_tablet_checkpoint_tasks_producer_thread));
+    LOG(INFO) << "tablet checkpoint tasks producer thread started";
 
     // fd cache clean thread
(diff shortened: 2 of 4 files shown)


be/src/common/config.h
```diff
@@ -516,6 +520,7 @@ CONF_Int32(flush_thread_num_per_store, "2");
 // config for tablet meta checkpoint
 CONF_mInt32(tablet_meta_checkpoint_min_new_rowsets_num, "10");
 CONF_mInt32(tablet_meta_checkpoint_min_interval_secs, "600");
+CONF_mInt32(generate_tablet_meta_checkpoint_tasks_interval_secs, "10");
```
acelyc111, line 523: Would produce checkpoint task every 10 seconds cause BE do checkpoint task too frequently? better to keep the behavior similar as before to set it to 600?

be/src/olap/olap_server.cpp
```diff
@@ -290,21 +292,19 @@ void StorageEngine::_path_scan_thread_callback(DataDir* data_dir) {
...
-            interval = config::tablet_meta_checkpoint_min_interval_secs - used_time;
-        } else {
-            interval = 1;
+        LOG(INFO) << "begin to produce tablet meta checkpoint tasks.";
+        for (auto data_dir : data_dirs) {
+            auto st =_tablet_meta_checkpoint_thread_pool->submit_func([=]() {
+                CgroupsMgr::apply_system_cgroup();
+                _tablet_manager->do_tablet_meta_checkpoint(data_dir);
+            });
         }
```
acelyc111, line 308: Could you keep the interval config mutable as before?

be/src/olap/olap_server.cpp
```diff
@@ -81,16 +81,18 @@ Status StorageEngine::start_bg_threads() {
...
-                &tablet_checkpoint_thread));
-        _tablet_checkpoint_threads.emplace_back(tablet_checkpoint_thread);
-    }
-    LOG(INFO) << "tablet checkpoint thread started";
+    int32_t max_checkpoint_thread_num = config::max_meta_checkpoint_threads;
+    int32_t min_checkpoint_thread_num = config::min_meta_checkpoint_threads;
+    ThreadPoolBuilder("TabletMetaCheckpointTaskThreadPool")
+            .set_min_threads(min_checkpoint_thread_num)
+            .set_max_threads(max_checkpoint_thread_num)
+            .build(&_tablet_meta_checkpoint_thread_pool);
```
acelyc111, line 89: _tablet_meta_checkpoint_thread_pool need shutdown in deconstructor.

be/src/olap/storage_engine.h
```diff
@@ -313,8 +313,8 @@ class StorageEngine {
     std::vector<scoped_refptr<Thread>> _path_gc_threads;
     // threads to scan disk paths
     std::vector<scoped_refptr<Thread>> _path_scan_threads;
-    // threads to run tablet checkpoint
-    std::vector<scoped_refptr<Thread>> _tablet_checkpoint_threads;
+    // threads to produce tablet checkpoint tasks
```
acelyc111, line 316:
```suggestion
    // thread to produce tablet checkpoint tasks
```

be/src/common/config.h
```diff
@@ -286,6 +286,10 @@ CONF_mInt64(min_compaction_failure_interval_sec, "600"); // 10 min
 CONF_mInt32(min_compaction_threads, "10");
 CONF_mInt32(max_compaction_threads, "10");
 
+// This config can be set to limit thread number in tablet meta checkpoint thread pool.
+CONF_mInt32(min_meta_checkpoint_threads, "10");
+CONF_mInt32(max_meta_checkpoint_threads, "10");
```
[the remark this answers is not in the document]
xinghuayu007: It is good idea to be compatible with previous config.

be/src/common/config.h
```diff
@@ -286,6 +286,9 @@ CONF_mInt64(min_compaction_failure_interval_sec, "600"); // 10 min
 CONF_mInt32(min_compaction_threads, "10");
 CONF_mInt32(max_compaction_threads, "10");
 
+// This config can be set to limit thread number in tablet meta checkpoint thread pool.
```
acelyc111, line 289:
```suggestion
// Thread count to do tablet meta checkpoint, -1 means use the data directories count.
```
Or some words like this.

be/src/olap/olap_server.cpp
```diff
@@ -81,16 +81,20 @@ Status StorageEngine::start_bg_threads() {
...
-    // tablet checkpoint thread
-    for (auto data_dir : data_dirs) {
-        scoped_refptr<Thread> tablet_checkpoint_thread;
-        RETURN_IF_ERROR(Thread::create(
-                "StorageEngine", "tablet_checkpoint_thread",
-                [this, data_dir]() { this->_tablet_checkpoint_callback(data_dir); },
-                &tablet_checkpoint_thread));
-        _tablet_checkpoint_threads.emplace_back(tablet_checkpoint_thread);
+    int32_t max_checkpoint_thread_num = config::max_meta_checkpoint_threads;
+    // compatible with before behavior
```
acelyc111, line 85:
```suggestion
```

be/src/common/config.h
xinghuayu007: done

be/src/olap/olap_server.cpp
```diff
@@ -290,22 +293,24 @@ void StorageEngine::_path_scan_thread_callback(DataDir* data_dir) {
     } while (!_stop_background_threads_latch.wait_for(MonoDelta::FromSeconds(interval)));
 }
 
-void StorageEngine::_tablet_checkpoint_callback(DataDir* data_dir) {
+void StorageEngine::_tablet_checkpoint_callback(std::vector<DataDir*> data_dirs) {
```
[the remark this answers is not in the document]
xinghuayu007: done

morningman (approved):
LGTM

PR #5660, opened 2021-04-15
[TRACE] Add trace info for all kinds of task

xinghuayu007:
## Proposed changes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/be/src/agent/task_worker_pool.cpp
+++ b/be/src/agent/task_worker_pool.cpp
@@ -73,6 +73,18 @@ DEFINE_GAUGE_METRIC_PROTOTYPE_2ARG(agent_task_queue_size, MetricUnit::NOUNIT);
 const uint32_t TASK_FINISH_MAX_RETRY = 3;
 const uint32_t PUBLISH_VERSION_MAX_RETRY = 3;
 
+#define TRACE_POINT() \
+    scoped_refptr<Trace> trace(new Trace); \
+    MonotonicStopWatch watch; \
+    watch.start(); \
+    SCOPED_CLEANUP({ \
+        if (config::agent_task_trace_threshold_sec > 0 && \
+            watch.elapsed_time() / 1e9 > config::agent_task_trace_threshold_sec) { \
+            LOG(WARNING) << "Trace:" << std::endl << trace->DumpToString(Trace::INCLUDE_ALL); \
+        } \
+    }); \
+    ADOPT_TRACE(trace.get());
+
 std::atomic_ulong TaskWorkerPool::_s_report_version(time(NULL) * 10000);
 Mutex TaskWorkerPool::_s_task_signatures_lock;
 map<TTaskType::type, set<int64_t>> TaskWorkerPool::_s_task_signatures;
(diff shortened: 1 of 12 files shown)


be/src/agent/task_worker_pool.cpp
```diff
@@ -390,6 +391,7 @@ void TaskWorkerPool::_drop_tablet_worker_thread_callback() {
         TDropTabletReq drop_tablet_req;
         {
             lock_guard<Mutex> worker_thread_lock(_worker_thread_lock);
+            TRACE("get worker thread lock");
```
acelyc111, line 394: Same as above

be/src/agent/task_worker_pool.cpp
```diff
@@ -402,6 +404,17 @@ void TaskWorkerPool::_drop_tablet_worker_thread_callback() {
...
         }
 
+        scoped_refptr<Trace> trace(new Trace);
+        MonotonicStopWatch watch;
+        watch.start();
+        SCOPED_CLEANUP({
+            if (watch.elapsed_time() / 1e9 > config::agent_task_trace_threshold_sec) {
+                LOG(WARNING) << "Trace:" << std::endl << trace->DumpToString(Trace::INCLUDE_ALL);
+            }
+        });
```
acelyc111, line 414: Better to define a macro for these code, it's duplicate with code in ceate tablet, and if you add more traces for other tasks, there will be manu duplicate code.

be/src/agent/task_worker_pool.cpp
```diff
@@ -402,6 +404,17 @@ void TaskWorkerPool::_drop_tablet_worker_thread_callback() {
...
+        ADOPT_TRACE(trace.get());
+        
+        TRACE("begin to drop tablet");
```
acelyc111, line 417: How about add another TRACE after line 422 when tablet got?

be/src/agent/task_worker_pool.cpp
xinghuayu007: update

acelyc111 (dismissed):

morningman (commented):
The default `agent_task_trace_threshold_sec` is only 2 seconds,
which may lead to a lot of tracing dump log.

Maybe we can modify the default value to a large number.

xinghuayu007:
> The default `agent_task_trace_threshold_sec` is only 2 seconds,
> which may lead to a lot of tracing dump log.
> 
> Maybe we can modify the default value to a large number.

how about set to 10s?

xinghuayu007:
> The default `agent_task_trace_threshold_sec` is only 2 seconds,
> which may lead to a lot of tracing dump log.
> 
> Maybe we can modify the default value to a large number.

`agent_task_trace_threshold_sec` is modified to 86400 seconds (one day). Anyone who wants to open trace feathure, just config it lower.

acelyc111:
> > The default `agent_task_trace_threshold_sec` is only 2 seconds,
> > which may lead to a lot of tracing dump log.
> > Maybe we can modify the default value to a large number.
> 
> `agent_task_trace_threshold_sec` is modified to 86400 seconds (one day). Anyone who wants to open trace feathure, just config it lower.

How about set default to -1 which means disable it?

xinghuayu007:
> > > The default `agent_task_trace_threshold_sec` is only 2 seconds,
> > > which may lead to a lot of tracing dump log.
> > > Maybe we can modify the default value to a large number.
> > 
> > 
> > `agent_task_trace_threshold_sec` is modified to 86400 seconds (one day). Anyone who wants to open trace feathure, just config it lower.
> 
> How about set default to -1 which means disable it?

good idea.

PR #5665, opened 2021-04-16
[Metrics] Add metrics to show the distribution of tablets version count

weizuo93:
## Proposed changes

 Add metrics (P50, P75, P90, P95, P99, etc.) to show the distribution of tablets version count.

```
# TYPE doris_be_tablet_version_num_distribution histogram
doris_be_tablet_version_num_distribution{quantile="0.50"} 9.21429
doris_be_tablet_version_num_distribution{quantile="0.75"} 11.7949
doris_be_tablet_version_num_distribution{quantile="0.90"} 13
doris_be_tablet_version_num_distribution{quantile="0.95"} 13
doris_be_tablet_version_num_distribution{quantile="0.99"} 13
doris_be_tablet_version_num_distribution_sum 950
doris_be_tablet_version_num_distribution_count 100
```

## Types of changes

### DIFF
--- a/be/src/olap/tablet_manager.cpp
+++ b/be/src/olap/tablet_manager.cpp
@@ -47,6 +47,7 @@
 #include "olap/utils.h"
 #include "util/doris_metrics.h"
 #include "util/file_utils.h"
+#include "util/histogram.h"
 #include "util/path_util.h"
 #include "util/pretty_printer.h"
 #include "util/scoped_cleanup.h"
@@ -965,7 +966,7 @@ OLAPStatus TabletManager::build_all_report_tablets_info(std::map<TTabletId, TTab
     LOG(INFO) << "find expired transactions for " << expire_txn_map.size() << " tablets";
 
     DorisMetrics::instance()->report_all_tablets_requests_total->increment(1);
-
+    HistogramStat tablet_version_num_hist;
     for (const auto& tablets_shard : _tablets_shards) {
         ReadLock rlock(tablets_shard.lock.get());
         for (const auto& item : tablets_shard.tablet_map) {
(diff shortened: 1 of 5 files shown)


be/src/util/doris_metrics.h
```diff
@@ -137,6 +137,20 @@ class DorisMetrics {
...
+    UIntGauge* tablet_version_num_distribution_sum;
+    DoubleGauge* tablet_version_num_distribution_median;
+    DoubleGauge* tablet_version_num_distribution_average;
+    DoubleGauge* tablet_version_num_distribution_standard_deviation;
+    DoubleGauge* tablet_version_num_distribution_p50;
+    DoubleGauge* tablet_version_num_distribution_p75;
+    DoubleGauge* tablet_version_num_distribution_p90;
+    DoubleGauge* tablet_version_num_distribution_p95;
+    DoubleGauge* tablet_version_num_distribution_p99;
+
```
acelyc111, line 152: Use HistogramMetric directly, its output is standard now.

weizuo93: > Use HistogramMetric directly, its output is standard now.

OK,  I see it too. PR #5671 has been merged.

morningman (dismissed):
LGTM

be/src/olap/tablet_manager.cpp
```diff
@@ -976,13 +978,17 @@ OLAPStatus TabletManager::build_all_report_tablets_info(std::map<TTabletId, TTab
                     expire_txn_map.erase(find);
                 }
                 t_tablet.tablet_infos.push_back(tablet_info);
+                if (tablet_ptr->tablet_id() == tablet_id) {
+                    DorisMetrics::instance()->tablet_version_num_distribution->add(tablet_ptr->version_count());
```
acelyc111, line 982: It's not atomic and may last a long time before all values added, that means we may get a middle state which is not very correct.

weizuo93: > It's not atomic and may last a long time before all values added, that means we may get a middle state which is not very correct.

Thank you, you are right.

acelyc111 (dismissed):
LGTM

morningman (approved):
LGTM

PR #5675, opened 2021-04-19
[Optimize] Remove expired txns in batch to avoid holding lock for too long

### LINKED ISSUE #5679: [Bug]  the report queue size exceeds the limit: 100. current: 101
**Describe the bug**
An error is reported in the Master FE log:

`the report queue size exceeds the limit: 100. current: 101`

This error will cause all report task from BE to FE fail, resulting in a variety of exceptions in the cluster, such as a large number of transactions in the COMMITTED state that cannot be completed, resulting in data loading blocked.

**To Reproduce**

One possible reason is that the regularly cleans up thread of expired transactions in FE may clean up a large number of expired transactions during a certain check, and write an editlog when cleaning up each transaction.
This is an IO operation.
The entire processing flow hold the TransactionMgr write lock, which causes other operations blocked.

morningman:
## Proposed changes

This CL mainly changes:

1.  Add a config to control the expire time of load job

    Add a new FE config "streaming_label_keep_max_second" to control
    the expire time of some high frequency load job such as INSERT and STREAM LOAD.

2. Remove expired txn in batch to avoid holding transaction lock for a long time

## Types of changes

## Checklist

- [X] I have created an issue on (Fix #5679 ) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
@@ -45,6 +45,7 @@ import org.apache.doris.common.util.TimeUtils;
 import org.apache.doris.common.util.Util;
 import org.apache.doris.metric.MetricRepo;
 import org.apache.doris.mysql.privilege.PrivPredicate;
+import org.apache.doris.persist.BatchRemoveTransactionsOperation;
 import org.apache.doris.persist.EditLog;
 import org.apache.doris.qe.ConnectContext;
 import org.apache.doris.task.AgentBatchTask;
@@ -55,16 +56,16 @@ import org.apache.doris.task.PublishVersionTask;
 import org.apache.doris.thrift.TTaskType;
 import org.apache.doris.thrift.TUniqueId;
 
-import org.apache.commons.collections.CollectionUtils;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import java.io.DataOutput;
 import java.io.IOException;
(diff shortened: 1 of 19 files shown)


fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
```diff
@@ -590,15 +594,44 @@ public boolean publishTransaction(Database db, long transactionId, long timeoutM
...
+
+    public void replayBatchRemoveTransaction(List<Long> txnIds) {
+        writeLock();
+        try {
+            for (Long txnId : txnIds) {
+                // here we only delete the oldest element, so if element exist in finalStatusTransactionStateDeque,
+                // it must at the front of the finalStatusTransactionStateDeque
+                // check both "short" and "long" queue.
+                if (!finalStatusTransactionStateDequeShort.isEmpty() &&
+                        txnId == finalStatusTransactionStateDequeShort.getFirst().getTransactionId()) {
```
caiconghui, line 627: use check isShortTxn to make code more concise？ use a queue reference to point to finalStatusTransactionStateDeque

fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
```diff
@@ -1436,9 +1501,8 @@ public void unprotectWriteAllTransactionStates(DataOutput out) throws IOExceptio
             entry.getValue().write(out);
         }
 
-        for (TransactionState transactionState : finalStatusTransactionStateDeque) {
+        for (TransactionState transactionState : idToFinalStatusTransactionState.values()) {
```
caiconghui, line 1504: use two finalStatusTransactionStateDeques instead of idToFinalStatusTransactionState to keep order for transaction state

morningman: Done

fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
morningman: If using `isShortTxn()`, I need to get TransactionState from `idToFinalStatusTransactionState`  first.
This change does not reduce the amount of code.

fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
```diff
@@ -1076,36 +1113,53 @@ private void clearBackendTransactions(TransactionState transactionState) {
...
-                TransactionState transactionState = finalStatusTransactionStateDeque.getFirst();
-                if (transactionState.isExpired(currentMillis)) {
-                    finalStatusTransactionStateDeque.pop();
-                    clearTransactionState(transactionState);
-                    editLog.logDeleteTransactionState(transactionState);
-                    LOG.info("transaction [" + transactionState.getTransactionId() + "] is expired, remove it from transaction manager");
-                } else {
-                    break;
-                }
+            unprotectedRemoveExpiredTxns(currentMillis, expiredTxnIds, finalStatusTransactionStateDequeShort);
```
caiconghui, line 1120: I suggest that we should do some experiments to make sure log BatchRemoveTransactionsOperation performance when there are many txns expired as the same time? I doubt that too many txns expired would cause bad performance when write log for bdbje?

morningman: We tested but writing a value of size 50MB to bdbje cost about 500ms.
And the number of txn here should be much smaller than it.
So I think we don't need to wrong about it.

caiconghui (approved):
LGTM

PR #5694, opened 2021-04-22
[Enhancement] Zonemap compatible with previous version-0.12

xinghuayu007:
## Proposed changes

Doris version-0.12 V1 segment  only supports creating zone map index for key columns. Doris version-0.13 supports creating zone map index for key columns and duplicate model columns. Lastest version supports creating zone map for key columns in AGG_KEYS or all columns in UNIQUE_KEYS or DUP_KEYS. In V1 Segment file, the tablet meta contains zone map index. But in V2 Segment file, it does not.

When upgrade Doris from version-0.12 to version-0.13, we found the memory is obviousely increasing in table with V1 segement file. Because a lots value columns create zone map index. Therefore, it is better to be compatible with version-0.12. The config `enable_value_column_zonemap` is default false. If it is need to add zone map index for value colums, you can set it true.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_group.cpp
+++ b/be/src/olap/rowset/segment_group.cpp
@@ -729,9 +729,6 @@ const TabletSchema& SegmentGroup::get_tablet_schema() {
 }
 
 int SegmentGroup::get_num_zone_map_columns() {
-    if (_schema->keys_type() != KeysType::AGG_KEYS) {
-        return _schema->num_columns();
-    }
     return _schema->num_key_columns();
 }
 


morningman (dismissed):
LGTM

be/src/common/config.h
```diff
@@ -598,6 +598,8 @@ CONF_mInt32(zone_map_row_num_threshold, "20");
 //    Trace = 6
 CONF_Int32(aws_log_level, "3");
 
+CONF_mBool(enable_value_column_zonemap, "false");
```
morningman, line 601: I think we can just remove this config, and just not add zone map for v1's value column, to make things simple.
Do we have any reason to add zonemap for v1's value column now?

xinghuayu007: ok

morningman (approved):
LGTM

PR #5706, opened 2021-04-26
[optimize] provide a better defer operator

### LINKED ISSUE #5576: [optimize] a better defer operater

Our current defer operation is based on std:: function But in most C + + standard libraries, when creating a function, there will be a new overhead, which is not necessary for the defer operation. But in fact, if we pass in a lambda expression, we can determine it during compilation, so we can avoid a new / delete overhead during the defer operation

STL std::funcional implement
```c++ 
private:
  static void
  _M_init_functor(_Any_data& __functor, _Functor&& __f, true_type)
  { ::new (__functor._M_access()) _Functor(std::move(__f)); }

  static void
  _M_init_functor(_Any_data& __functor, _Functor&& __f, false_type)
  { __functor._M_access<_Functor*>() = new _Functor(std::move(__f)); }
```

we could use this code instread of DeferOP
```c++
template <class T>
class Defer {
public:
    Defer(T& closure) : _closure(closure) {}
    ~Defer() { _closure(); }

private:
    T& _closure;
};
```

Here was the bench mark: https://quick-bench.com/q/4IKpxAA5VEbXVziV1G2Po-ZupSE

image

stdpain:
## Proposed changes

will close #5576 

## Types of changes

## Checklist

- [x] I have created an issue on (#5576) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/be/src/util/defer_op.h
+++ b/be/src/util/defer_op.h
@@ -34,6 +34,24 @@ private:
     DeferFunction _func;
 };
 
+// A Better Defer operator #5576
+// for C++17
+// Defer defer {[]{ call something }};
+//
+// for C++11
+// auto op = [] {};
+// Defer<decltype<op>> (op);
+template <class T>
+class Defer {
+public:
+    Defer(T& closure) : _closure(closure) {}
+    Defer(T&& closure) : _closure(std::move(closure)) {}
+    ~Defer() { _closure(); }
+
+private:
+    T _closure;
+};
+
 } // namespace doris
 
 #endif
--- a/be/src/util/defer_op.h
+++ b/be/src/util/defer_op.h
@@ -23,17 +23,6 @@
 namespace doris {
 
 // This class is used to defer a function when this object is deconstruct
-class DeferOp {
-public:
-    typedef std::function<void()> DeferFunction;
-    DeferOp(const DeferFunction& func) : _func(func) {}
-
-    ~DeferOp() { _func(); };
-
-private:
-    DeferFunction _func;
-};
-
 // A Better Defer operator #5576
 // for C++17
 // Defer defer {[]{ call something }};
--- a/be/src/util/file_utils.cpp
+++ b/be/src/util/file_utils.cpp
@@ -240,16 +240,15 @@ Status FileUtils::copy_file(const std::string& src_path, const std::string& dest
     }
 
     const int64_t BUF_SIZE = 8192;
-    char* buf = new char[BUF_SIZE];
-    DeferOp free_buf(std::bind<void>(std::default_delete<char[]>(), buf));
+    std::unique_ptr<char[]> buf(new char[BUF_SIZE]);
     int64_t src_length = src_file.length();
     int64_t offset = 0;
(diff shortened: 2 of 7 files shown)


be/src/util/file_utils.cpp
```diff
@@ -240,16 +240,15 @@ Status FileUtils::copy_file(const std::string& src_path, const std::string& dest
     }
 
     const int64_t BUF_SIZE = 8192;
-    char* buf = new char[BUF_SIZE];
-    DeferOp free_buf(std::bind<void>(std::default_delete<char[]>(), buf));
+    std::unique_ptr<char[]> buf(new char[BUF_SIZE]);
```
yangzhg, line 243: std::unique_ptr<char[]> buf = `make_unique<char[]>(BUF_SIZE);`  may be better

be/src/util/defer_op.h
```diff
@@ -34,6 +34,24 @@ class DeferOp {
...
+// for C++17
+// Defer defer {[]{ call something }};
+//
+// for C++11
+// auto op = [] {};
+// Defer<decltype<op>> (op);
+template <class T>
+class Defer {
+public:
+    Defer(T& closure) : _closure(closure) {}
```
HappenLee, line 47: may only Defer(T&& closure) : _clousure(std::forward<T>(closure)) is better

stdpain: std::forward<T>(closure) will got a compile error
such as:
```
error: cannot bind rvalue reference of type 'main()::<lambda()>&&' to lvalue of type 'main()::<lambda()>'
```

be/src/util/file_utils.cpp
stdpain: fixed

be/src/util/defer_op.h
HappenLee: I test the code，do not have the compile error.

be/src/util/defer_op.h
```diff
@@ -34,6 +34,24 @@ class DeferOp {
...
 
+// A Better Defer operator #5576
+// for C++17
+// Defer defer {[]{ call something }};
+//
+// for C++11
+// auto op = [] {};
+// Defer<decltype<op>> (op);
+template <class T>
+class Defer {
```
HappenLee, line 45: USE Defer replace DeferOP directly, do not create a new class,

HappenLee (changes requested):
little problem

acelyc111:
Seems it is similar with `class ScopedCleanup` in be/src/util/scoped_cleanup.h? why not use it?

stdpain:
@acelyc111 
emm... I think "Defer" has a better semantics than "ScopedCleanup"

HappenLee:
> Seems it is similar with `class ScopedCleanup` in be/src/util/scoped_cleanup.h? why not use it?

it is similar with `class ScopedCleanup`,  but more effectively. we shoule union the code chose only one way to use

HappenLee (approved):
LGTM

morningman (approved):
LGTM

PR #5731, opened 2021-04-28
[Apache] Change download link to archive page

EmmyMiao87:
Only the latest package should appear in svn.
The old release package already has been archived, so the download link should be replaced by archive page.


yangzhg (dismissed):
LGTM

yangzhg (approved):
LGTM

PR #5737, opened 2021-04-29
[Compaction] Modify compaction logic

### LINKED ISSUE #5715: Difficult to find Issue Tracker
It's not at all easy to find how to report bugs.

The only link I could find required navigating several pages.
The link on the page http://doris.incubator.apache.org/master/en/community/how-to-contribute.html#if-a-bug-or-problem-is-found leads to  https://github.com/apache/incubator-doris/issues/new/select which does not actually work.

I would expect to find a direct link to the Issue tracker in one of the menus as well.

### LINKED ISSUE #5738: [Compaction] A tablet cannot execute base and cumulative compaction at the same time
**Describe the bug**
As mentioned above, in a high-concurrency import scenario, if a tablet is doing base compaction and cannot perform cumulative compaction at the same time, and base compaction may run for a long time, this will cause the tablet to accumulate too many delta versions.

We can strictly guarantee that the cumulative point will not fall back, so it is safe for a tablet to do two compactions at the same time

morningman:
## Proposed changes

1. Add /api/compaction/run_status to show the running compaction tasks.
2. Support do base and cumulative compaction for one tablet at same time.
3. Modify some log level.
4. Add a feedback document. [Close #5715 ]

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] Documentation Update (if none of the other choices apply)

## Checklist

- [X] I have created an issue on (Fix #5738) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -270,7 +270,6 @@ CONF_mInt64(cumulative_size_based_compaction_lower_size_mbytes, "64");
 // cumulative compaction policy: min and max delta file's number
 CONF_mInt64(min_cumulative_compaction_num_singleton_deltas, "5");
 CONF_mInt64(max_cumulative_compaction_num_singleton_deltas, "1000");
-CONF_mInt64(cumulative_compaction_budgeted_bytes, "104857600");
 // cumulative compaction skips recently published deltas in order to prevent
 // compacting a version that might be queried (in case the query planning phase took some time).
 // the following config set the window size
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -204,8 +204,6 @@ CONF_mInt32(max_percentage_of_error_disk, "0");
 CONF_mInt32(default_num_rows_per_column_file_block, "1024");
 // pending data policy
 CONF_mInt32(pending_data_expire_time_sec, "1800");
-// inc_rowset expired interval
-CONF_mInt32(inc_rowset_expired_sec, "1800");
 // inc_rowset snapshot rs sweep time interval
 CONF_mInt32(tablet_rowset_stale_sweep_time_sec, "1800");
 // garbage sweep policy
@@ -283,8 +281,7 @@ CONF_mInt32(cumulative_compaction_skip_window_seconds, "30");
 CONF_mInt64(min_compaction_failure_interval_sec, "600"); // 10 min
 
 // This config can be set to limit thread number in compaction thread pool.
-CONF_mInt32(min_compaction_threads, "10");
-CONF_mInt32(max_compaction_threads, "10");
+CONF_Int32(max_compaction_threads, "10");
 
 // Thread count to do tablet meta checkpoint, -1 means use the data directories count.
 CONF_Int32(max_meta_checkpoint_threads, "-1");
--- a/be/src/http/action/compaction_action.cpp
+++ b/be/src/http/action/compaction_action.cpp
@@ -180,7 +180,7 @@ Status CompactionAction::_handle_run_status_compaction(HttpRequest* req, std::st
             "compact_type" : "$4"
 })";
     
-        std::string msg = "this tablet_id is not running";
+        std::string msg = "compaction task for this tablet is not running";
         std::string compaction_type = "";
         bool run_status = 0;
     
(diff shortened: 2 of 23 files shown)


be/src/common/config.h
```diff
@@ -283,8 +281,7 @@ CONF_mInt32(cumulative_compaction_skip_window_seconds, "30");
 CONF_mInt64(min_compaction_failure_interval_sec, "600"); // 10 min
 
 // This config can be set to limit thread number in compaction thread pool.
-CONF_mInt32(min_compaction_threads, "10");
-CONF_mInt32(max_compaction_threads, "10");
+CONF_Int32(max_compaction_threads, "10");
```
weizuo93, line 284: Now, `max_threads` and `min_threads`of Thread Pool can not be adjusted dynamically. Recently, I am looking into the implementation of the ThreadPool, and I consider supporting dynamic adjustment of `max_threads` and `min_threads`for thread pool later. How do you like it?

be/src/http/action/compaction_action.cpp
```diff
@@ -155,59 +154,66 @@ Status CompactionAction::_handle_run_status_compaction(HttpRequest* req, std::st
...
-        MutexLock lock_cumulative(tablet->get_cumulative_lock(), TRY_LOCK);
-        if (!lock_cumulative.own_lock()) {
-            msg = "this tablet_id is running";
-            compaction_type = "cumulative";
-            run_status = 1;
-            *json_result = strings::Substitute(json_template, run_status, msg, tablet_id,
-                                               schema_hash, compaction_type);
-            return Status::OK();
+    
+        std::string msg = "this tablet_id is not running";
```
weizuo93, line 183: I think `compaction task for this tablet_id is not running` may be better, and `this tablet_id is not running`may be misunderstood to mean that this tablet is unhealthy.

be/src/http/action/compaction_action.cpp
```diff
@@ -155,59 +154,66 @@ Status CompactionAction::_handle_run_status_compaction(HttpRequest* req, std::st
...
+        std::string compaction_type = "";
+        bool run_status = 0;
+    
+        {
+            // use try lock to check this tablet is running cumulative compaction
+            MutexLock lock_cumulative(tablet->get_cumulative_lock(), TRY_LOCK);
+            if (!lock_cumulative.own_lock()) {
+                msg = "this tablet_id is running";
```
weizuo93, line 191: `this tablet_id is running compaction task`

be/src/http/action/compaction_action.cpp
```diff
@@ -155,59 +154,66 @@ Status CompactionAction::_handle_run_status_compaction(HttpRequest* req, std::st
...
-            run_status = 1;
-            *json_result = strings::Substitute(json_template, run_status, msg, tablet_id,
-                                               schema_hash, compaction_type);
-            return Status::OK();
+    
+        {
+            // use try lock to check this tablet is running base compaction
+            MutexLock lock_base(tablet->get_base_lock(), TRY_LOCK);
+            if (!lock_base.own_lock()) {
+                msg = "this tablet_id is running";
```
weizuo93, line 204: `this tablet_id is running compaction task`?

weizuo93:
+1

be/src/common/config.h
morningman: It's OK for me~
But I think the `min_compaction_threads` is meaning less for user, we should reduce unnecessary configuration items as much as possible

be/src/olap/storage_engine.cpp
```diff
@@ -1035,4 +1035,76 @@ void StorageEngine::create_base_compaction(TabletSharedPtr best_tablet,
     base_compaction.reset(new BaseCompaction(best_tablet, tracker_label, _compaction_mem_tracker));
 }
 
+// Return json:
+// {
+//   "CumulativeCompaction": {
+//          "/home/disk1" : [10001, 10002],
```
weizuo93, line 1041: Can we add segment number that be merged in this compaction task for each thread? Of course, we could add the segment number info in another PR.

be/src/http/action/compaction_action.cpp
morningman: ok

be/src/common/config.h
weizuo93: > It's OK for me~
> But I think the `min_compaction_threads` is meaning less for user, we should reduce unnecessary configuration items as much as possible

OK.

EmmyMiao87 (approved):
LGTM

PR #5749, opened 2021-04-30
[Bug-fix] Pause routine load when data size exceeds quota

### LINKED ISSUE #5745: Routine load is stuck and no new task is generated
**Describe the bug**

The status of routine load task is running, but the number of abort tasks and commit tasks remains unchanged.
And it won't work if the task is paused and restarted, or the same task is recreated.
The task is still stuck.

After opening the debug log of routine load, you will find the following line：
```
2021-04-30 13:54:14,687 DEBUG (Routine load task scheduler|40) [RoutineLoadTaskScheduler.allocateTaskToBe():303] ROUTINE_LOAD_TASK=39f4653fd0dd4b93-8bd8ae6d069654a7, job_id={33427}, be_id={10002}, msg={task has been allocated to be}
2021-04-30 13:54:14,687 DEBUG (Routine load task scheduler|40) [RoutineLoadTaskInfo.beginTxn():183] begin txn failed for routine load task: 39f4653fd0dd4b93-8bd8ae6d069654a7, errCode = 2, detailMessage = Database[default_cluster:iaas] data size exceeds quota[3.000 TB]
```
Version 0.13.16

**To Reproduce**
1. Fully use the quota of the database.
2. create routine load
3. show routine load
4. routine load is running but no task is executed.

**Expected behavior**
When the database capacity is full, the routine load task should be paused immediately.
 The error information should be exposed to the end user.

EmmyMiao87:
## Proposed changes

In the previous code, the routine load task did not catch the exception of opening the transaction.
As a result, although the task cannot be executed,
no exceptions can be seen during show routine load, only the routine load job is stuck.

The PR catch the QuotaExceedException when opening a transaction.
If the routine load task cannot be executed due to the exhaustion of the quota,
the routine load will be paused and an error message will be presented to the user.

Similarly, other load method will also catch similar exceptions and cancel job.

Fixed #5745

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #5745 ) and described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/common/MetaNotFoundException.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/MetaNotFoundException.java
@@ -17,14 +17,12 @@
 
 package org.apache.doris.common;
 
-import static org.apache.doris.common.InternalErrorCode.META_NOT_FOUND_ERR;
-
 /**
  * Exception for meta info is null, like db table partition tablet replica job
  */
 public class MetaNotFoundException extends UserException {
     public MetaNotFoundException(String msg) {
-        super(META_NOT_FOUND_ERR, msg);
+        super(InternalErrorCode.META_NOT_FOUND_ERR, msg);
     }
 
     public MetaNotFoundException(InternalErrorCode errcode, String msg) {
--- a/fe/fe-core/src/main/java/org/apache/doris/common/QuotaExceedException.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/QuotaExceedException.java
@@ -17,21 +17,12 @@
 
 package org.apache.doris.common;
 
-import static org.apache.doris.common.InternalErrorCode.META_NOT_FOUND_ERR;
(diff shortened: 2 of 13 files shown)


fe/fe-core/src/main/java/org/apache/doris/common/MetaNotFoundException.java
```diff
@@ -17,12 +17,14 @@
 
 package org.apache.doris.common;
 
+import static org.apache.doris.common.InternalErrorCode.META_NOT_FOUND_ERR;
```
morningman, line 20: Do not use static import

EmmyMiao87: Changed, but why?

morningman: Just to follow the existing code style

morningman (approved):
LGTM

PR #5768, opened 2021-05-07
[Bug-fix] When query cancel, transfer_thread does not continue to schedule scanner_thread

### LINKED ISSUE #5767: [BUG] The OlapScanNode thread on be will not end quickly after query cancel
**Describe the bug**
For some large queries, after canceling the query in the mysql client, it takes a long time for the scan thread on be to completely exit. During this period, the I/O will be full.

**To Reproduce**
Table data Size: 12T, ReplicaCount: 4032.
The query is similar: select * from T where c1 = xxx, where the value of xxx does not exist in column c1, so the entire table will be scanned.

**Expected behavior**
After the query was canceled during the running process, the actual be did not stop executing the query, which caused the I/O to be full.

xinyiZzz:
## Proposed changes

The cause of the problem is that after query cancel, OlapScanNode::transfer_thread still continues to schedule OlapScanNode::scanner_thread until all tasks are scheduled.

Although each task does not scan data and exits quickly, it still consumes a lot of resources.
(Guess)This may be the cause of the BUG (#5767) causing the I/O to be full.

So after query cancel, immediately exit the scheduling loop in transfer_thread, and after waiting for the end of all scanner_threads, transfer_thread will also exit.

## Types of changes

## Checklist
- [x] I have created an issue on (Fix #5767) and described the bug/feature there in detail

### DIFF
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -1252,6 +1252,13 @@ void OlapScanNode::transfer_thread(RuntimeState* state) {
     }
     // read from scanner
     while (LIKELY(status.ok())) {
+        // When query cancel, _transfer_done is set to true at OlapScanNode::close,
+        // and the loop is exited at this time, and the current thread exits after
+        // waiting for _running_thread to decrease to 0.
+        if (UNLIKELY(_transfer_done)) {
+            LOG(INFO) << "Transfer thread cancelled, wait for the end of scan thread.";
+            break;
+        }
         int assigned_thread_num = 0;
         // copy to local
         {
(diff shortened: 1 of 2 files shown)


be/src/exec/olap_scan_node.cpp
```diff
@@ -1368,9 +1375,14 @@ void OlapScanNode::scanner_thread(OlapScanner* scanner) {
...
     bool eos = false;
+    if (UNLIKELY(_transfer_done)) {
+        eos = true;
+        status = Status::Cancelled("Cancelled");
+        LOG(INFO) << "Scan thread cancelled, cause query done, scan thread started to exit";
+    } 
     RuntimeState* state = scanner->runtime_state();
     DCHECK(NULL != state);
-    if (!scanner->is_open()) {
+    if (!eos && !scanner->is_open()) {
```
morningman, line 1385: If scanner is not opened, we need to take care of "close()" method

xinyiZzz: Modified to when _transfer_done` is True, the scan thread will exit directly after starting, 
and no longer execute the subsequent logic including the scanner's `close()` method.

be/src/exec/olap_scan_node.cpp
```diff
@@ -1361,6 +1368,14 @@ void OlapScanNode::transfer_thread(RuntimeState* state) {
 }
 
 void OlapScanNode::scanner_thread(OlapScanner* scanner) {
+    if (UNLIKELY(_transfer_done)) {
+        _scanner_done = true;
+        std::unique_lock<std::mutex> l(_scan_batches_lock);
+        _running_thread--;
+        _scan_thread_exit_cv.notify_one();
```
morningman, line 1375: I think _scan_batch_added_cv should be notified too.

xinyiZzz: Ok,
I thought that `notify_all` was already in `OlapScanNode::close`……but I forgot that `_transfer_done` may also be set to true in `scanner_thread`.

morningman (approved):
LGTM

PR #5770, opened 2021-05-08
[Feature] Support show data order by

DinoZhang:
## Proposed changes

Currently, the `show data` does not support sorting. When the number of tables increases, it is inconvenient to manage. Need to support sorting

like:
```
mysql>  show data order by ReplicaCount desc,Size asc;
+-----------+-------------+--------------+
| TableName | Size        | ReplicaCount |
+-----------+-------------+--------------+
| table_c   | 3.102 KB    | 40           |
| table_d   | .000        | 20           |
| table_b   | 324.000 B   | 20           |
| table_a   | 1.266 KB    | 10           |
| Total     | 4.684 KB    | 90           |
| Quota     | 1024.000 GB | 1073741824   |
| Left      | 1024.000 GB | 1073741734   |
+-----------+-------------+--------------+
```

## Types of changes

## Checklist

- [x] I have created an issue on (Fix issue #5836 ) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/docs/zh-CN/sql-reference/sql-statements/Data Manipulation/SHOW DATA.md
+++ b/docs/zh-CN/sql-reference/sql-statements/Data Manipulation/SHOW DATA.md
@@ -33,7 +33,7 @@ under the License.
 语法：
 
 ```
-SHOW DATA [FROM db_name[.table_name]];
+SHOW DATA [FROM db_name[.table_name]] [ORDER BY ...];
 ```
     
 说明：
@@ -47,6 +47,8 @@ SHOW DATA [FROM db_name[.table_name]];
 4. 结果集中的 `Total` 行表示汇总行。`Quota` 行表示当前数据库设置的配额。`Left` 行表示剩余配额。
 
 5. 如果想查看各个 Partition 的大小，请参阅 `help show partitions`。
+
+6. 可以使用 ORDER BY 对任意列组合进行排序。
 
 ## example
 
--- a/docs/zh-CN/sql-reference/sql-statements/Data Manipulation/SHOW DATA.md
+++ b/docs/zh-CN/sql-reference/sql-statements/Data Manipulation/SHOW DATA.md
@@ -74,7 +74,9 @@ SHOW DATA [FROM db_name[.table_name]] [ORDER BY ...];
 
     ```
     SHOW DATA FROM example_db.test;
+    ```
     
+    ```
     +-----------+-----------+-----------+--------------+----------+
     | TableName | IndexName | Size      | ReplicaCount | RowCount |
     +-----------+-----------+-----------+--------------+----------+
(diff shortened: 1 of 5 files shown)


fe/fe-core/src/main/java/org/apache/doris/analysis/ShowDataStmt.java
```diff
@@ -136,13 +160,27 @@ public int compare(Table t1, Table t2) {
                     String readableSize = DebugUtil.DECIMAL_FORMAT_SCALE_3.format(tableSizePair.first) + " "
                             + tableSizePair.second;
 
-                    List<String> row = Arrays.asList(table.getName(), readableSize, String.valueOf(replicaCount));
-                    totalRows.add(row);
+					if (orderBySize == null) {
+						List<String> row = Arrays.asList(table.getName(), readableSize, String.valueOf(replicaCount));
+						totalRows.add(row);
+					} else {
+						List<String> row = Arrays.asList(tableSize + "", table.getName(), readableSize, String.valueOf(replicaCount));
```
morningman, line 167: Please using 4 spaces instead of `\t`

fe/fe-core/src/main/java/org/apache/doris/analysis/ShowDataStmt.java
```diff
@@ -211,24 +249,51 @@ public int compare(Table t1, Table t2) {
...
-                        row = Arrays.asList(tableName,
-                                olapTable.getIndexNameById(indexId),
-                                readableSize, String.valueOf(indexReplicaCount),
-                                String.valueOf(indexRowCount));
-                    } else {
-                        row = Arrays.asList("",
-                                olapTable.getIndexNameById(indexId),
-                                readableSize, String.valueOf(indexReplicaCount),
-                                String.valueOf(indexRowCount));
+						if (orderBySize == null) {
```
morningman, line 252: Please notice the code foramt

morningman (commented):
I think this function can be more general. The current `show data` is sorted by table name by default, which is equivalent to `order by table_name`. Based on this, we can expand `order by size` and `order by replica_count`.

We can first use a list to store the original values, then sort them according to the `order by` column, and finally convert them to readable format for output. This can avoid a lot of if else judgments.

DinoZhang:
@morningman Thank you for your review. Based on your comments, I think it is possible to use ORDER BY to sort any combination of columns.

docs/zh-CN/sql-reference/sql-statements/Data Manipulation/SHOW DATA.md
```diff
@@ -48,6 +48,8 @@ SHOW DATA [FROM db_name[.table_name]];
 
 5. 如果想查看各个 Partition 的大小，请参阅 `help show partitions`。
 
+6. 可以使用 ORDER BY 对任意列组合进行排序。
+
 ## example
 
 1. 展示默认 db 的各个 table 的数据量，副本数量，汇总数据量和汇总副本数量。
```
EmmyMiao87, line 55: 可以增加一个使用order by 的实际例子。

fe/fe-core/src/main/java/org/apache/doris/analysis/ShowDataStmt.java
```diff
@@ -17,7 +17,12 @@
 
 package org.apache.doris.analysis;
 
+import com.google.common.base.Strings;
```
EmmyMiao87, line 20: pay attention to the import order

fe/fe-core/src/main/java/org/apache/doris/analysis/ShowDataStmt.java
```diff
@@ -270,6 +371,15 @@ public String toSql() {
             return builder.toString();
```
EmmyMiao87, line 371: Maybe it should not be returned here.
The reason is that even if there is no tableName, there may be an order by clause.

EmmyMiao87:
By the way,  unless it is a very small pr, there should be an issue to describe it ~

morningman (approved):
LGTM

PR #5781, opened 2021-05-10
[Compaction][ThreadPool]Support adjust compaction threads num at runtime

### LINKED ISSUE #5779: [ThreadPool] Adjust max_thread_num and min_thread_num of thread pool at runtime
Now, the maximum and minimum number of threads (`max_thread_num` and `min_thread_num`) in the thread pool are initialized once and cannot be changed at runtime. Sometimes, we need to adjust `max_threads` or `min_threads`dynamically according to  machine load. （For example, we need to adjust compaction thread number according to memory usage or cpu load sometimes）

The current mechanism of thread pool is as follows:

(1) When the thread pool is initialized, multiple threads (which we call `permanent threads`) are created based on the `min_thread_num`. 
(2) When more tasks need to be executed and `permanent threads` cannot meet the demand, new threads are created which we call `non-permanent threads`. The total number of threads will not exceed `max_thread_num`, that is:
```
permanent_threads_num + non-permanent_threads_num <= max_thread_num
```
(3)When the number of tasks reduced and there are idle threads, idle `non-permanent threads`will be released but idle `permanent threads`will never be released even if there is no task in `task queue`.

We can easily adjust  `max_thread_num`and `min_thread_num`at runtime. New `max_thread_num` can work well, but sometimes new`min_thread_num`is hard to work because `permanent threads`, which have been created when initialization, can not be released. So can we modify the thread pool model as follows:

(1) When the thread pool is initialized, multiple threads  (which we call `initial threads`) are created based on the `min_thread_num`. 
(2) When more tasks need to be executed and current threads cannot meet the demand, new threads are created which we call `subsequent threads`. The total number of thread will not exceed `max_thread_num` , that is:
```
initial_threads_num + subsequent_threads_num <= max_thread_num
```
(3) When the number of tasks reduced and there are idle threads, both `initial threads`and `subsequent threads`could be released so long as total number of thread is not less than `min_thread_num`. 

In this model, `initial threads` and `subsequent threads`will have equal status for thread pool.

weizuo93:
## Proposed changes

The number of threads in compaction thread pool remains constant(`max_thread_num` is the same as `min_thread_num` in compaction thread pool). Sometimes, we need to adjust compaction thread number dynamically according to memory usage and cpu load of BE.

The method of adjusting `max_thread_num` and `min_thread_num` of thread pool at runtime has been described in ISSUE #5779 .

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #5779) and described the bug/feature there in detail

### DIFF
--- a/be/src/olap/olap_server.cpp
+++ b/be/src/olap/olap_server.cpp
@@ -333,6 +333,30 @@ void StorageEngine::_compaction_tasks_producer_callback() {
     int64_t interval = config::generate_compaction_tasks_min_interval_ms;
     do {
         if (!config::disable_auto_compaction) {
+            VLOG_CRITICAL << "compaction thread pool. num_threads: " << _compaction_thread_pool->num_threads()
+                      << ", num_threads_pending_start: " << _compaction_thread_pool->num_threads_pending_start()
+                      << ", num_active_threads: " << _compaction_thread_pool->num_active_threads()
+                      << ", max_threads: " << _compaction_thread_pool->max_threads()
+                      << ", min_threads: " << _compaction_thread_pool->min_threads()
+                      << ", num_total_queued_tasks: " << _compaction_thread_pool->get_queue_size();
+
+            if(_compaction_thread_pool->max_threads() != config::max_compaction_threads) {
+                int old_max_threads = _compaction_thread_pool->max_threads();
+                Status status = _compaction_thread_pool->set_max_threads(config::max_compaction_threads);
+                if (status.ok()) {
+                    LOG(INFO) << "update compaction thread pool max_threads from "
+                              << old_max_threads << " to " << config::max_compaction_threads;
+                }
+            }
+            if(_compaction_thread_pool->min_threads() != config::max_compaction_threads) {
+                int old_min_threads = _compaction_thread_pool->min_threads();
+                Status status = _compaction_thread_pool->set_min_threads(config::max_compaction_threads);
+                if (status.ok()) {
+                    LOG(INFO) << "update compaction thread pool min_threads from "
+                              << old_min_threads << " to " << config::max_compaction_threads;
+                }
+            }
+
             bool check_score = false;
             int64_t cur_time = UnixMillis();
             if (round < config::cumulative_compaction_rounds_for_each_base_compaction_round) {
--- a/be/src/util/threadpool.cpp
+++ b/be/src/util/threadpool.cpp
@@ -673,7 +673,7 @@ Status ThreadPool::set_max_threads(int max_threads) {
     }
 
     _max_threads = max_threads;
-    if (_max_threads > _num_threads - _num_threads_pending_start) {
+    if (_max_threads > _num_threads + _num_threads_pending_start) {
         int addition_threads = _max_threads - _num_threads - _num_threads_pending_start;
         addition_threads = std::min(addition_threads, _total_queued_tasks);
         _num_threads_pending_start += addition_threads;
--- a/be/src/util/threadpool.cpp
+++ b/be/src/util/threadpool.cpp
@@ -472,7 +472,7 @@ Status ThreadPool::do_submit(std::shared_ptr<Runnable> r, ThreadPoolToken* token
             }
             // If we failed to create a thread, but there are still some other
             // worker threads, log a warning message and continue.
-            LOG(ERROR) << "Thread pool failed to create thread: " << status.to_string();
+            LOG(WARNING) << "Thread pool failed to create thread: " << status.to_string();
         }
     }
 
@@ -509,9 +509,6 @@ void ThreadPool::dispatch_thread() {
     DCHECK_GT(_num_threads_pending_start, 0);
     _num_threads++;
     _num_threads_pending_start--;
-    // If we are one of the first '_min_threads' to start, we must be
-    // a "permanent" thread.
-    bool permanent = _num_threads <= _min_threads;
 
     // Owned by this worker thread and added/removed from _idle_threads as needed.
     IdleThread me(&_lock);
@@ -520,6 +517,10 @@ void ThreadPool::dispatch_thread() {
         // Note: Status::Aborted() is used to indicate normal shutdown.
         if (!_pool_status.ok()) {
             VLOG_CRITICAL << "DispatchThread exiting: " << _pool_status.to_string();
+            break;
+        }
+
+        if (_num_threads + _num_threads_pending_start > _max_threads) {
             break;
         }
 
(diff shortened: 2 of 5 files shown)


be/src/olap/olap_server.cpp
```diff
@@ -334,6 +334,30 @@ void StorageEngine::_compaction_tasks_producer_callback() {
...
+            VLOG_CRITICAL << "compaction thread pool. num_threads: " << _compaction_thread_pool->num_threads()
+                      << ", num_threads_pending_start: " << _compaction_thread_pool->num_threads_pending_start()
+                      << ", num_active_threads: " << _compaction_thread_pool->num_active_threads()
+                      << ", max_threads: " << _compaction_thread_pool->max_threads()
+                      << ", min_threads: " << _compaction_thread_pool->min_threads()
+                      << ", num_total_queued_tasks: " << _compaction_thread_pool->num_total_queued_tasks();
+
+            if(_compaction_thread_pool->max_threads() != config::max_compaction_threads) {
+                Status status = _compaction_thread_pool->set_max_threads(config::max_compaction_threads);
+                if (status.ok()) {
```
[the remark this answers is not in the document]
weizuo93: > what if it returns failure?

Max and min thread num for compaction thread pool are both determined by config `max_compaction_threads`. It means we turn down the config `max_compaction_threads` if it returns failure. when we call `set_max_threads()`, `_min_threads` hasn't been adjusted yet and` _max_threads` can not be less than `_min_threads`, so it would return failure. We don't need to worry. `set_min_threads()`will be executed and turn down `_min_threads` later in this round of compaction producer. `set_max_threads()`will be executed successfully next round and turn down ` _max_threads`.

be/src/olap/olap_server.cpp
```diff
@@ -334,6 +334,30 @@ void StorageEngine::_compaction_tasks_producer_callback() {
...
+                    LOG(INFO) << "update compaction thread pool max_threads from "
+                              << _compaction_thread_pool->max_threads() << " to "
+                              << config::max_compaction_threads;
+                }
+            }
+            if(_compaction_thread_pool->min_threads() != config::max_compaction_threads) {
+                Status status = _compaction_thread_pool->set_min_threads(config::max_compaction_threads);
+                if (status.ok()) {
```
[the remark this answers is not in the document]
weizuo93: > what if it returns failure?

There are two possibilities if it returns failure.
(1)  We turn up the config `max_compaction_threads`. Similar to the above, `_max_threads` hasn't been adjusted yet.`set_min_threads()` will be executed again next round in compaction producer and turn up `_min_threads`.
(2) Failed to create new thread. We don't need to worry. When compaction task is submitted to thread pool, a new thread will be created.

be/src/util/threadpool.cpp
```diff
@@ -610,6 +607,35 @@ void ThreadPool::check_not_pool_thread_unlocked() {
...
+        int old_min_threads = _min_threads;
+        _min_threads = min_threads;
+        if (min_threads > _num_threads + _num_threads_pending_start) {
+            int addition_threads = min_threads - _num_threads - _num_threads_pending_start;
+            _num_threads_pending_start += addition_threads;
+            for (int i = 0; i < addition_threads; i++) {
+                Status status = create_thread();
+                if (!status.ok()) {
+                    _num_threads_pending_start--;
+                    LOG(ERROR) << "Thread pool failed to create thread: " << status.to_string();
```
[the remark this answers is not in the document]
weizuo93: > use WARNING

OK.

be/src/util/threadpool.cpp
```diff
@@ -610,6 +607,35 @@ void ThreadPool::check_not_pool_thread_unlocked() {
     }
 }
 
+Status ThreadPool::set_min_threads(int min_threads) {
+    if (min_threads <= _max_threads) {
+        int old_min_threads = _min_threads;
```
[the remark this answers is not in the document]
weizuo93: > `old_min_threads` seems useless.

Sorry, it's my carelessness.

be/test/util/threadpool_test.cpp
```diff
@@ -778,6 +778,85 @@ TEST_F(ThreadPoolTest, TestLIFOThreadWakeUps) {
 }
 */
 
+TEST_F(ThreadPoolTest, TestThreadPoolDynamicAdjustMaximumMinimum) {
+ASSERT_TRUE(rebuild_pool_with_builder(ThreadPoolBuilder(kDefaultPoolName)
```
acelyc111, line 782: indent

be/src/util/threadpool.cpp
```diff
@@ -610,6 +607,47 @@ void ThreadPool::check_not_pool_thread_unlocked() {
...
+                }
+            }
+        }
+        return Status::OK();
+    }
+    return Status::InternalError("set thread pool min_threads failed");
+}
+
+Status ThreadPool::set_max_threads(int max_threads) {
+    if (_min_threads <= max_threads) {
```
acelyc111, line 631: Maybe Use short circuit return would make you code more elegant

be/src/util/threadpool.cpp
```diff
@@ -610,6 +607,47 @@ void ThreadPool::check_not_pool_thread_unlocked() {
...
+                if (!status.ok()) {
+                    _num_threads_pending_start--;
+                    LOG(WARNING) << "Thread pool failed to create thread: " << status.to_string();
+                    return status;
+                }
+            }
+        }
+        return Status::OK();
+    }
+    return Status::InternalError("set thread pool min_threads failed");
```
acelyc111, line 627: Better to clarify why it failed.

be/src/util/threadpool.cpp
```diff
@@ -610,6 +607,47 @@ void ThreadPool::check_not_pool_thread_unlocked() {
     }
 }
 
+Status ThreadPool::set_min_threads(int min_threads) {
+    if (min_threads <= _max_threads) {
+        _min_threads = min_threads;
```
acelyc111, line 612: Even though `_min_threads` is type of `AtomicInt32`, this function is not thread safe.

be/test/util/threadpool_test.cpp
weizuo93: > indent

OK, Thank you.

be/src/util/threadpool.cpp
weizuo93: > Maybe Use short circuit return would make you code more elegant

OK, Thank you.

be/src/util/threadpool.cpp
weizuo93: > Better to clarify why it failed.

OK, Thank you.

be/src/util/threadpool.cpp
weizuo93: > Even though `_min_threads` is type of `AtomicInt32`, this function is not thread safe.

You're right, thank you.

morningman (approved):
LGTM

PR #5783, opened 2021-05-10
[MemTracker] Make some MemTracker have reasonable parent MemTracker to eazily to trace. Add show level of MemTracker

### LINKED ISSUE #5782: [Enhance] Make MemTracker More Reasonable
### 2 Make Each MemTracker can be trace:
        
The current memory consumption is mainly concentrated on the following operations:

- query
- load
- compaction
- schema change

But it's really difficult to Identify the memory consumption caused by that particular operation, such as three queries. We don't know which query occupies the most memory.

So we need a way to identify it

- query ：each memtracker of same instance the label have same instance id
- load：same as above
- compaction ： each memtracker of same tablet of compaction the label have same tablet id
- schema change： each memtracker of same tablet of schema change the label have old tablet id - new tablet id

HappenLee:
1 Make some MemTracker have reasonable parent MemTracker not the root tracker
2 Make each MemTracker can be eazily to trace.
3 Add show level of MemTracker to reduce the MemTracker show in the web page to have a way to control show how many tracker in web page.

## Proposed changes

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #5782 ) and described the bug/feature there in detail

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -598,6 +598,12 @@ CONF_Int32(aws_log_level, "3");
 // the buffer size when read data from remote storage like s3
 CONF_mInt32(remote_storage_read_buffer_mb, "256");
 
+// Default level of MemTracker to show in web page
+// now MemTracker support two level:
+//      RELEASE: 0
+//      DEBUG: 1
+// the level equal or lower than mem_tracker_level will show in web page
+CONF_Int16(mem_tracker_level, "0");
 } // namespace config
 
 } // namespace doris
--- a/be/src/olap/memtable.cpp
+++ b/be/src/olap/memtable.cpp
@@ -40,7 +40,7 @@ MemTable::MemTable(int64_t tablet_id, Schema* schema, const TabletSchema* tablet
           _slot_descs(slot_descs),
           _keys_type(keys_type),
           _row_comparator(_schema),
-          _mem_tracker(MemTracker::CreateTracker(-1, "MemTable:" + std::to_string(tablet_id), parent_tracker)),
+          _mem_tracker(MemTracker::CreateTracker(-1, "MemTable", parent_tracker)),
           _buffer_mem_pool(new MemPool(_mem_tracker.get())),
           _table_mem_pool(new MemPool(_mem_tracker.get())),
           _schema_size(_schema->schema_size()),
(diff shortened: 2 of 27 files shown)


be/src/common/config.h
```diff
@@ -597,6 +597,8 @@ CONF_mInt32(zone_map_row_num_threshold, "20");
 //    Trace = 6
 CONF_Int32(aws_log_level, "3");
 
+// default level of MemTracker to show in webpage
```
morningman, line 600: Explain whether this value is higher and more detailed or lower and more detailed.

be/src/olap/memtable.cpp
```diff
@@ -40,7 +40,7 @@ MemTable::MemTable(int64_t tablet_id, Schema* schema, const TabletSchema* tablet
           _slot_descs(slot_descs),
           _keys_type(keys_type),
           _row_comparator(_schema),
-          _mem_tracker(MemTracker::CreateTracker(-1, "MemTable:" + std::to_string(tablet_id), parent_tracker)),
+          _mem_tracker(MemTracker::CreateTracker(-1, "MemTable", parent_tracker)),
```
morningman, line 43: Why remove tablet id?

be/src/runtime/tablets_channel.cpp
```diff
@@ -34,7 +34,7 @@ std::atomic<uint64_t> TabletsChannel::_s_tablet_writer_count;
 TabletsChannel::TabletsChannel(const TabletsChannelKey& key,
                                const std::shared_ptr<MemTracker>& mem_tracker)
         : _key(key), _state(kInitialized), _closed_senders(64) {
-    _mem_tracker = MemTracker::CreateTracker(-1, "TabletsChannel:" + key.to_string(), mem_tracker);
+    _mem_tracker = MemTracker::CreateTracker(-1, "TabletsChannel", mem_tracker);
```
morningman, line 37: Why remove key?

be/src/olap/memtable.cpp
HappenLee: Beacause the tracker will use the parent label id

be/src/runtime/tablets_channel.cpp
HappenLee: The same to up

morningman (approved):
LGTM

PR #5791, opened 2021-05-11
Add buffer to s3 reader and fix bug in BufferedReader

### LINKED ISSUE #5789: [BUG] BufferedReader return wrong data when read size is larger than buffer
When read data size is larger than buffer, the BufferedReader will return data directly without buffer, but it will not update _current_offset, so when call read again, it will return same data. The read method will not return the read bytes either.

yangzhg:
## Proposed changes

1. add buffer when use s3 reader
2. fix bug (#5789 ) in BufferedReader
3. change the read method in  FileReader, separate buf_len and read bytes

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #5789) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/be/src/exec/buffered_reader.cpp
+++ b/be/src/exec/buffered_reader.cpp
@@ -45,21 +45,18 @@ Status BufferedReader::open() {
         return Status::InternalError(ss.str());
     }
     RETURN_IF_ERROR(_reader->open());
-    RETURN_IF_ERROR(_fill());
     return Status::OK();
 }
 
 //not support
-Status BufferedReader::read_one_message(std::unique_ptr<uint8_t[]>* buf, size_t* length) {
+Status BufferedReader::read_one_message(std::unique_ptr<uint8_t[]>* buf, int64_t* length) {
     return Status::NotSupported("Not support");
-
 }
 
-Status BufferedReader::read(uint8_t* buf, size_t* buf_len, bool* eof) {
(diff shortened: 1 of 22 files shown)


be/src/exec/buffered_reader.cpp
```diff
@@ -116,13 +117,8 @@ Status BufferedReader::_read_once(int64_t position, int64_t nbytes, int64_t* byt
...
     if (_buffer_offset >= 0) {
-        int64_t bytes_read;
-        // retry for new content
-        int retry_times = 1;
-        do {
-            // fill the buffer
-            RETURN_IF_ERROR(_reader->readat(_buffer_offset, _buffer_size, &bytes_read, _buffer));
-        } while (bytes_read == 0 && retry_times++ < 2);
+        int64_t bytes_read = 0;
+        RETURN_IF_ERROR(_reader->readat(_buffer_offset, _buffer_size, &bytes_read, _buffer));
```
morningman, line 121: Why the previous code had "retry" logic?

yangzhg: @xy720 has confirmed, this retry is useless

xy720: This retry is used to check whether the file has been updated after the first read of end. (nearly useless..)

xy720 (approved):
LGTM

morningman (approved):
LGTM

PR #5800, opened 2021-05-12
abstract some codes to a new parse method for further scalability

copperybean:
abstract some codes to a new parse method for further scalability


morningman (approved):
LGTM

PR #5811, opened 2021-05-13
[Enhancement] Reduce memory consumption by releasing readers earier

acelyc111:
## Proposed changes

We created multiple rowset readers to read data of one tablet, after one rowset reader has reached EOF, it can be released to reduce resource (typicallly memory) comsumption. As the same, we can release segment reader when it reach EOF.

## Types of changes

## Checklist

### DIFF
--- a/be/src/olap/rowset/alpha_rowset_reader.cpp
+++ b/be/src/olap/rowset/alpha_rowset_reader.cpp
@@ -37,6 +37,15 @@ AlphaRowsetReader::AlphaRowsetReader(int num_rows_per_row_block, AlphaRowsetShar
 AlphaRowsetReader::~AlphaRowsetReader() {
     delete _dst_cursor;
     _rowset->release();
+    while (!_merge_heap.empty()) {
+        auto ctx = _merge_heap.top();
+        _merge_heap.pop();
+        delete ctx;
+    }
+    for (auto ctx : _sequential_ctxs) {
+        delete ctx;
+    }
+    _sequential_ctxs.clear();
 }
 
 OLAPStatus AlphaRowsetReader::init(RowsetReaderContext* read_context) {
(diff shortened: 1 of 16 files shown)


be/src/olap/rowset/alpha_rowset_reader.cpp
```diff
@@ -120,20 +129,24 @@ int64_t AlphaRowsetReader::filtered_rows() {
...
         } else if (status != OLAP_SUCCESS) {
             return status;
         } else {
-            (*block) = _merge_ctxs[_ordinal].row_block;
+            (*block) = _cur_ctx->row_block;
             return OLAP_SUCCESS;
         }
     }
-    if (_ordinal == _merge_ctxs.size()) {
+    if (_sequential_ctxs.empty()) {
```
morningman, line 149: Seems that this can be a `CHECK(_init_merge_heap.empty())`

be/src/olap/rowset/alpha_rowset_reader.cpp
```diff
@@ -37,6 +37,15 @@ AlphaRowsetReader::AlphaRowsetReader(int num_rows_per_row_block, AlphaRowsetShar
 AlphaRowsetReader::~AlphaRowsetReader() {
     delete _dst_cursor;
     _rowset->release();
+    while (!_merge_heap.empty()) {
```
morningman, line 40: Why in previous implementation, we don't need to clear the `_merge_heap` here?

be/src/olap/rowset/alpha_rowset_reader.cpp
morningman: OK

be/src/olap/rowset/alpha_rowset_reader.cpp
morningman: OK

morningman (approved):
LGTM

PR #5820, opened 2021-05-14
[Debug Enhancement] Associate the old and new query id by retry

EmmyMiao87:
## Proposed changes

When a query is retried, the FE log cannot quickly associate the new and old queries by query id.
This will increase the complexity of troubleshooting.
Modify the log printing logic of FE to associate the new and old query ids, and the print log looks like this:
Query {old_query_id} {retry_times} times with new query id: {new_query_id}


morningman (approved):
LGTM

PR #5831, opened 2021-05-17
[BE] Optimize version retrieval efficiency.

### LINKED ISSUE #5830: [Enhance] Optimize version retrieval efficiency in high-frequency import/compaction scenarios.
## Backgroud

For high-frequency import and high-frequency Compaction scenarios, the number of versions of the Tablet will be large(maybe thousands), and the version graph will be more complex. In this case, the current implementation of VersionGraph.capture_consistent_versions will have relatively large performance problems.

For example, the flowing is a flame diagram of one of our online scenes

image

After analyzing the relevant code, find the current version capture algorithm is BFS, and there are two main reasons for the poor performance
1. Space complexity is O (n), which is related to the versions number of Tablet
2. In most scenarios, the version retrieving needs to traverse all versions.

Suppose StreamLoad is performed once in 1 second and Compaction is performed once in 2 seconds, the version graph of the tablet may be as the following figure (the edges represents rowset, the nodes represents version)

image

If we want to retrieve the version (0 -> 8), the retrieval process is as follows (we need to traverse all nodes and edges)

`0 -> (1, 2, 4, 6) -> (3, 5, 7) -> (8)`

## Solution

One solution is to change the retrieval algorithm from BFS to greedy strategy. when retrieving, we only find the version with the largest step size in each time, so that the space complexity can be reduced to O (1), and the nodes that need to be traversed when retrieving are also greatly reduced.

For example, for the above scenario, the retrieval process is as follows:

`0 -> 6 -> 7 -> 8`

The following is the flame diagram of BE in the same scene after switching to greedy strategy. The proportion of capture_consistent_versions has almost disappeared.

image

killxdcj:
Optimize version retrieval efficiency in high-frequency import/compaction scenarios.

## Proposed changes

More detail: #5830 

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #5830 ) and described the bug/feature there in detail

### DIFF
--- a/be/src/olap/version_graph.cpp
+++ b/be/src/olap/version_graph.cpp
@@ -580,7 +580,7 @@ OLAPStatus VersionGraph::capture_consistent_versions(const Version& spec_version
         for (const auto& it : _version_graph[cur_idx].edges) {
             // Only consider incremental versions
             if (_version_graph[it].value < _version_graph[cur_idx].value) {
-                continue;
+                break;
             }
 
             if (_version_graph[it].value > end_value) {
--- a/be/src/olap/version_graph.cpp
+++ b/be/src/olap/version_graph.cpp
@@ -452,6 +452,13 @@ void VersionGraph::construct_version_graph(const std::vector<RowsetMetaSharedPtr
         // Add reverse edge from end_version to start_version.
         _version_graph[end_vertex_index].edges.push_front(start_vertex_index);
     }
+
+    // Sort edges by version in descending order.
+    for (auto& vertex : _version_graph) {
+        vertex.edges.sort([this](const int& vertex_idx_a, const int& vertex_idx_b) {
+            return _version_graph[vertex_idx_a].value > _version_graph[vertex_idx_b].value;
+        });
+    }
 }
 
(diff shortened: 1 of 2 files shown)


morningman:
Thanks for your contribution. I will take time to review the PR.. maybe later...

be/src/olap/version_graph.cpp
```diff
@@ -537,95 +560,58 @@ OLAPStatus VersionGraph::capture_consistent_versions(const Version& spec_version
...
-                }
-                visited[it] = true;
-                predecessor[it] = top_vertex_index;
-                bfs_queue.push(it);
+    int64_t end_value = spec_version.second + 1;
+    while (_version_graph[cur_idx].value < end_value) {
+        int64_t next_idx = -1;
+        for (const auto& it : _version_graph[cur_idx].edges) {
+            // Only consider incremental versions
+            if (_version_graph[it].value < _version_graph[cur_idx].value) {
```
morningman, line 582: If we already sort edges in descending order, can we "break" here?

killxdcj: resolved~

morningman (approved):
LGTM

PR #5872, opened 2021-05-21
[Bug] NULL value in `where` and `on` clause should return EmptySetNode

### LINKED ISSUE #5859: [BUG] NULL = NULL in the join expression is different from the value of mysql
**Describe the bug**
1) query 1:
```
SELECT * FROM tab1 INNER JOIN tab0 ON NULL = NULL
```
`NULL = NULL` in Mysql is `false`
`NULL = NULL` in Doris is `true`

2) In contrast, query 2:
```
SELECT * FROM tab1 where NULL = NULL
```
NULL = NULL is `false` in both Doris and Mysql

In addition, the conclusion is the same in NULL NOT BETWEEN NULL AND (NULL ), NULL <> NULL, NULL <= NULL.

Found through the sqllogictest test case.

HappenLee:
## Proposed changes

RT

Fix the issue#5859

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #5859 ) and described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/Analyzer.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/Analyzer.java
@@ -1251,6 +1251,14 @@ public class Analyzer {
                     }
                     markConjunctAssigned(conjunct);
                 }
+                if (newConjunct instanceof NullLiteral) {
+                    if (fromHavingClause) {
+                        hasEmptyResultSet_ = true;
+                    } else {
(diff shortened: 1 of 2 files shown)


EmmyMiao87 (dismissed):
Although it is very unreasonable to add it here, but at this stage it will have to do.

morningman:
UT failed

morningman (approved):
LGTM

PR #5807, opened 2021-05-13
[Command] [SQL] Add show database/table/partition id command

xinghuayu007:
## Proposed changes

In BE, when a problem happened, in the log, we can find the database id, table id, partition id, but no database name, table name, partition name.

_2021-05-18 19:07:52,331 INFO 193033 [DatabaseTransactionMgr.commitTransaction():512] transaction:[TransactionState. transaction id: 285498579, label: index_subtask_0_time_1621336072089_count_1_48cfe999-7460-439e-a284-d71b2943e203, db id: 19595062, table id list: 169245759, callback id: -1, coordinator: BE: host:51692, transaction status: COMMITTED, error replicas num: 0, replica ids: , prepare time: 1621336072307, commit time: 1621336072329, finish time: -1, reason: ] successfully committed_

 In FE, there also no way to find database name/table name/partition name accourding to database id/table id/partition id. Therefore, this patch add 3 new commands:
1. show database id;
mysql> show database 10002;
+----------------------+
| DbName               |
+----------------------+
| default_cluster:test |
+----------------------+
2. show table id;
mysql> show table 11100;
+----------------------+-----------+-------+
| DbName               | TableName | DbId  |
+----------------------+-----------+-------+
| default_cluster:test | table2    | 10002 |
+----------------------+-----------+-------+

3. show partition id;
mysql> show partition 11099;
+----------------------+-----------+---------------+-------+---------+
| DbName               | TableName | PartitionName | DbId  | TableId |
+----------------------+-----------+---------------+-------+---------+
| default_cluster:test | table2    | p201708       | 10002 | 11100   |
+----------------------+-----------+---------------+-------+---------+
## Types of changes

What types of changes does your code introduce to Doris?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Code refactor (Modify the code structure, format the code, etc...)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [ ] I have created an issue on (Fix #ISSUE) and described the bug/feature there in detail
- [x] Compiling and unit tests pass locally with my changes
[...]

PR #5880, opened 2021-05-22
[Optimize] Add db,table information in transaction exception log

e0c9:
Improve transaction exception log.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/transaction/TransactionState.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/transaction/TransactionState.java
@@ -18,7 +18,9 @@
 package org.apache.doris.transaction;
 
 import org.apache.doris.catalog.Catalog;
+import org.apache.doris.catalog.Database;
 import org.apache.doris.catalog.OlapTable;
+import org.apache.doris.catalog.Table;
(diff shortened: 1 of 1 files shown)


fe/fe-core/src/main/java/org/apache/doris/transaction/TransactionState.java
```diff
@@ -526,11 +530,23 @@ public synchronized void addTableIndexes(OlapTable table) {
...
+        Database db = Catalog.getCurrentCatalog().getDb(dbId);
+        String dbName = null;
+        String tableNameList = null;
+        if (db != null) {
+           dbName = db.getFullName();
+           tableNameList = tableIdList.stream()
+                   .map(db::getTable)
+                   .filter(Objects::nonNull)
+                   .map(Table::getName).collect(Collectors.joining(","));
+        }
```
morningman, line 542: We have add `show database id` stmt to find database name from database id.
So there is no need to print database name here.
And a database may have thousands of tables, so print all table here may be disaster.

morningman: This PR #5807 may solve your problem

PR #5908, opened 2021-05-26
[BUG] Fix "in memory table" may cause a lot of CPU consumption when LRU Cache evict

### LINKED ISSUE #5907: [BUG] "In memory table" may cause a lot of CPU consumption when LRU Cache evict
**Describe the bug**
When `in_memory=true` is specified in `create table`, currently doris will specify the LRU priority as `DURABLE` when reading the table data and caching it through the `page cache`.

After the `page cache` is filled, the LRU strategy will be used to traverse the `lru list` to try to evict the longest unused `NORMAL` priority page. The page with priority `DURABLE` will be skipped. After traversing the `lru' List` and evict all `NORMAL pages` and try to evict `DURABLE page` if there is still insufficient space.

This will cause the page cache of `in memory table` to always accumulate at the end of the `lru list`, and it will be traversed every time LRU evict, which will consume a lot of CPU when the cache is full.

**To Reproduce**
1) Try to fill up `DURABLE page` in `page cache` in `StoragePageCache::StoragePageCache`:
2) Comment out the statement of `cache->lookup` in `PageIO::read_and_decompress_page`

**Expected behavior**
After submitting the query, `perf` can see that the `LRUCache::_evict_from_lru` method consumes a lot of CPU.

xinyiZzz:
## Proposed changes

According to the LRU priority, the `lru list` is split into `lru normal list` and `lru durable list`, and the two lists are traversed in sequence during LRU evict, avoiding invalid cycles.

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #5907) and described the bug/feature there in detail

## Further comments

At present, `in memory table` will always occupy part of the page cache even if this `in memory table` has not been accessed for a long time, causing a waste of cache resources.

Later, we can consider that the page of `in memory table` can be cached for a longer period of time instead of permanent cache, such as increasing the timestamp or weight.

### DIFF
--- a/be/src/olap/lru_cache.cpp
+++ b/be/src/olap/lru_cache.cpp
@@ -177,8 +177,10 @@ void HandleTable::_resize() {
 
 LRUCache::LRUCache() {
     // Make empty circular linked list
-    _lru.next = &_lru;
-    _lru.prev = &_lru;
+    _lru_normal.next = &_lru_normal;
+    _lru_normal.prev = &_lru_normal;
+    _lru_durable.next = &_lru_durable;
+    _lru_durable.prev = &_lru_durable;
 }
 
 LRUCache::~LRUCache() {
@@ -244,7 +246,11 @@ void LRUCache::release(Cache::Handle* handle) {
                 last_ref = true;
             } else {
                 // put it to LRU free list
-                _lru_append(&_lru, e);
+                if (e->priority == CachePriority::NORMAL) {
+                    _lru_append(&_lru_normal, e);
+                } else if (e->priority == CachePriority::DURABLE) {
+                    _lru_append(&_lru_durable, e);
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


be/src/olap/lru_cache.h
```diff
@@ -300,6 +300,7 @@ class LRUCache {
     Cache::Handle* lookup(const CacheKey& key, uint32_t hash);
     void release(Cache::Handle* handle);
     void erase(const CacheKey& key, uint32_t hash);
+    void prune_one(LRUHandle* old);
```
morningman, line 303: should be a private method, named `_prune_one`

be/src/olap/lru_cache.cpp
```diff
@@ -244,7 +246,11 @@ void LRUCache::release(Cache::Handle* handle) {
                 last_ref = true;
             } else {
                 // put it to LRU free list
-                _lru_append(&_lru, e);
+                if (e->priority ==  CachePriority::NORMAL) {
```
morningman, line 249:
```suggestion
                if (e->priority == CachePriority::NORMAL) {
```

be/src/olap/lru_cache.cpp
```diff
@@ -244,7 +246,11 @@ void LRUCache::release(Cache::Handle* handle) {
...
+                    _lru_append(&_lru_normal, e);
+                } else if (e->priority ==  CachePriority::DURABLE) {
```
morningman, line 251:
```suggestion
                } else if (e->priority == CachePriority::DURABLE) {
```

be/src/olap/lru_cache.h
xinyiZzz: Ok, pay attention later

morningman (approved):
LGTM

PR #5916, opened 2021-05-26
[BUG] Fix parse error when the memory parameter is a float value percentage

### LINKED ISSUE #5818: [BUG]  The float value percentage of `storage_page_cache_limit` will cause the page_cache capacity calculation result to be wrong
**Describe the bug**
For example, when setting `storage_page_cache_limit=0.1%` in `be.conf`, the method `ParseUtil::parse_mem_spec` in `ExecEnv::_init_mem_tracker()` returns `storage cache capacity = -1`.

After the parameters are passed to `StoragePageCache::create_global_cache`, `storage cache capacity ` will be converted from `int64_t` to `size_t` and then equal to `18446744073709551615`

xinyiZzz:
## Proposed changes

When parsing memory parameters in `ParseUtil::parse_mem_spec`, convert the percentage to `double` instead of `int`.

The currently affected parameters include `mem_limit` and `storage_page_cache_limit`

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #5818) and described the bug/feature there in detail

### DIFF
--- a/be/src/util/parse_util.cpp
+++ b/be/src/util/parse_util.cpp
@@ -69,8 +69,8 @@ int64_t ParseUtil::parse_mem_spec(const std::string& mem_spec_str, bool* is_perc
     StringParser::ParseResult result;
     int64_t bytes;
 
-    if (multiplier != -1) {
-        // Parse float - MB or GB
+    if (multiplier != -1 || *is_percent) {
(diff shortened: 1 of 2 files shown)


morningman (commented):
add some unit test

morningman (approved):
LGTM

PR #5925, opened 2021-05-27
The storage format should not be converted from v1 to v2 by default when alter table's schema

### LINKED ISSUE #5924: [Bug] Alter table add column operation becomes slow when upgrade doris from 0.12 to 0.13, 0.14
**Describe the bug**
 Alter table add column operation becomes slow when upgrade doris from 0.12 to 0.13, 0.14, 
before upgrade, alter table add column operation may cost several minutes
after upgrade, alter table add column opertion may easily failed due to timeout 

**To Reproduce**
upgrade doris version from 0.12 to 0.13 or 0.14

**Expected behavior**
alter table add column succeed

**Additional context**
we should consider other way to convert storage format from v1 to v2 instead of by alter table, because it may cause many
failure when user alter table, which may have a great negative impact on the business

caiconghui:
## Proposed changes

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #5924 ) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/be/src/olap/tablet_manager.cpp
+++ b/be/src/olap/tablet_manager.cpp
@@ -1402,12 +1402,15 @@ OLAPStatus TabletManager::_create_tablet_meta_unlocked(const TCreateTabletReq& r
     OLAPStatus res = TabletMeta::create(request, TabletUid::gen_uid(), shard_id, next_unique_id,
                                         col_idx_to_unique_id, tablet_meta);
 
-    // TODO(lingbin): when beta-rowset is default, should remove it
-    if (request.__isset.storage_format && request.storage_format == TStorageFormat::V2) {
-        (*tablet_meta)->set_preferred_rowset_type(BETA_ROWSET);
-    } else {
-        (*tablet_meta)->set_preferred_rowset_type(ALPHA_ROWSET);
+    auto rowset_type = StorageEngine::instance()->default_rowset_type();
+    if (request.__isset.storage_format) {
+        if (request.storage_format == TStorageFormat::V1) {
+            rowset_type = ALPHA_ROWSET;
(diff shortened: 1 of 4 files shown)


stdpain:
SegmentV1 will be deprecated in the future, so I don't think this change is necessary.
Insisting on maintaining SegmentV1 will be a big burden in future vectorization support

caiconghui:
> SegmentV1 will be deprecated in the future, so I don't think this change is necessary.
> Insisting on maintaining SegmentV1 will be a big burden in future vectorization support

Yes , I agree with you that SegmentV1 will be deprecated in the future, segment V2 will replace all segment v1 in the future.

This PR is not Insisting on maintaining SegmentV1 but to be compatible with older doris version.

For new doris user , This PR doesn't take any effect, all newer table will be created by segment v2

For older doris user, there are many tables with segment v1, and it need to gradually change V1 to v2,
actually, for many older doris user, the most important thing is stable on production enviroment, but now, Very few alter operations succeed, most fail due to timeout, because convert v1 to v2 is a huge task which would cost a lot of time while there are many tablets in older table. and once timeout,  all operations that convert v1 to v2 would rollback, which is not what we expect, and make doris not friendly to older user

If we always alter table failed,  it is meanless to  convert v1 to v2 when do alter table, I think it is not a good idea to convert v1 to v2 when user just want to add column for table, which now seems an impossible goal to achive

yangzhg:
We need to update to segment V2 and remove the v1 code from the code base. If this is done, this will be delayed indefinitely. If it is for the stability of online users, this part of users can stay in A certain version, segment v2 has gone through 3 versions from 0.12 to the current 0.14. If we should not have any way to create a new v1 table in the current code, we should keep the v1 table read-only

PR #5947, opened 2021-05-31
[Feature] Support recording custom number of backup and restore task information

e0c9:
## Proposed changes
Doris currently only supports saving the last backup/restore task information. This PR supports to save a custom number of task information.

SQL Syntax 
```SQL
SHOW backup/restore FROM dbname [WHERE LABEL = "" | WHERE LABEL LIKE "matcher"]
```
Lark20210531-182218

## Types of changes

## Checklist


fe/fe-core/src/main/java/org/apache/doris/backup/BackupHandler.java
```diff
@@ -392,11 +410,49 @@ private void restore(Repository repository, Database db, RestoreStmt stmt) throw
...
+        addBackupOrRestoreJob(db.getId(), restoreJob);
 
         LOG.info("finished to submit restore job: {}", restoreJob);
     }
 
+    private void addBackupOrRestoreJob(long dbId, AbstractJob job) {
+        jobLock.lock();
+        try {
+            Deque<AbstractJob> jobs = dbIdToBackupOrRestoreJobs.computeIfAbsent(dbId, k -> Lists.newLinkedList());
+            if (jobs.size() == Config.max_backup_job_num_per_db) {
```
morningman, line 422:
```suggestion
            where (jobs.size() >= Config.max_backup_job_num_per_db) {
            ...
            }
```

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -1396,4 +1396,10 @@
      */
     @ConfField(mutable = true, masterOnly = true)
     public static int max_dynamic_partition_num = 500;
+
+    /*
+     * Control the max num of backup job per db
+     */
+    @ConfField(mutable = true, masterOnly = true)
+    public static int max_backup_job_num_per_db = 100;
```
morningman, line 1404:
```suggestion
    public static int max_backup_restore_job_num_per_db = 100;
```
And I think 100 by default is too large. How about 10?

fe/fe-core/src/main/java/org/apache/doris/backup/BackupHandler.java
```diff
@@ -392,11 +410,49 @@ private void restore(Repository repository, Database db, RestoreStmt stmt) throw
...
+                jobs.removeFirst();
+            }
+            AbstractJob lastJob = jobs.peekLast();
+            // only save the latest job
```
morningman, line 426: Better add more comment to explain why we need to remove the duplicate job at the last of the queue.
I think the reason is because we may add same job with different jobs state when replaying edit log.

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
e0c9: done

fe/fe-core/src/main/java/org/apache/doris/backup/BackupHandler.java
```diff
@@ -17,6 +17,7 @@
 
 package org.apache.doris.backup;
 
+import org.apache.commons.collections.CollectionUtils;
```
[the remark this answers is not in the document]
e0c9: solved

fe/fe-core/src/main/java/org/apache/doris/backup/BackupHandler.java
```diff
@@ -621,8 +681,16 @@ public static BackupHandler read(DataInput in) throws IOException {
     public void write(DataOutput out) throws IOException {
         repoMgr.write(out);
 
-        out.writeInt(dbIdToBackupOrRestoreJob.size());
-        for (AbstractJob job : dbIdToBackupOrRestoreJob.values()) {
+        jobLock.lock();
```
morningman, line 684: Actually, we don't need to lock here, cause the `write()` method can only be called in Checkpoint thread, so there is no concurrency issue. 
And by the way, I think we can just use:
`for (AbstractJob job : dbIdToBackupOrRestoreJob.values())`

morningman (approved):
LGTM

PR #5948, opened 2021-05-31
[Feature] Support setting the number and time threshold for displaying query profile

e0c9:
## Proposed changes
Sometimes we only need to care about big queries, so we can filter out queries that we don’t need to care about by setting a time threshold.

Usage:
```
set is_report_success = true;
set report_query_time_threshold = 100;
--------
set global report_query_array_size = 150;
```

## Types of changes

### DIFF
--- a/docs/zh-CN/administrator-guide/variables.md
+++ b/docs/zh-CN/administrator-guide/variables.md
@@ -68,6 +68,7 @@ SET GLOBAL exec_mem_limit = 137438953472
 * `wait_timeout`
 * `sql_mode`
 * `is_report_success`
+* `report_query_time_threshold`
 * `query_timeout`
 * `exec_mem_limit`
 * `batch_size`
@@ -234,7 +235,13 @@ SELECT /*+ SET_VAR(query_timeout = 1, enable_partition_cache=true) */ sleep(3);
     
     `fe_host:fe_http_port/query`
     
-    其中会显示最近100条，开启 `is_report_success` 的查询的 profile。
+    其中默认会显示最近100条（可以通过`reporty_query_array_size`进行设置），开启 `is_report_success` 的查询的 profile。
+
+* `report_query_time_threshold`
+    用于设置展示查询 profile 的耗时阈值，单位 ms，默认为 -1，展示所有查询 profile。
+
+* `report_query_array_size` 
+    用于设置展示查询 profile 的数量，默认为100。
     
 * `language`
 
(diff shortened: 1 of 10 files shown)


docs/zh-CN/administrator-guide/variables.md
```diff
@@ -234,7 +235,13 @@ SELECT /*+ SET_VAR(query_timeout = 1, enable_partition_cache=true) */ sleep(3);
     
     `fe_host:fe_http_port/query`
     
-    其中会显示最近100条，开启 `is_report_success` 的查询的 profile。
+    其中默认会显示最近100条（可以通过`reporty_query_array_size`进行设置），开启 `is_report_success` 的查询的 profile。
```
morningman, line 238: I think it's better to be a fe configuration, not session variable?

docs/zh-CN/administrator-guide/variables.md
```diff
@@ -68,6 +68,7 @@ SET GLOBAL exec_mem_limit = 137438953472
 * `wait_timeout`
 * `sql_mode`
 * `is_report_success`
+* `report_query_time_threshold`
```
morningman, line 71:
```suggestion
* `report_query_time_threshold_second`
```
Using second is more clear and easy to use.

fe/fe-core/src/main/java/org/apache/doris/common/util/ProfileManager.java
```diff
@@ -132,8 +134,15 @@ public void pushProfile(RuntimeProfile profile) {
         if (profile == null) {
             return;
         }
-        
+
+        // filter profile that take more time than threshold
         ProfileElement element = createElement(profile);
```
morningman, line 139: Currently, Doris will call this `pushProfile()` at 3 phases:
1. The beginning of the query.
2. During the query execution(report incremental)
3. The end of the query.

At phase 1 and 2, the `element.totalTimeMs` may be very small. so you may not be able to see the profile of current running queries. This is not a very big problem because it will be pushed successfully at phase 3, finally.
But can we change it like:

1. Allow pushing at phase 1 and 2, no matter what `element.totalTimeMs` is.
2. At phase 3, check the threshold and determine to push the final profile, or to delete it if the query time is too small.

e0c9: I understand that pushing profile does not need to care about the different phase, it will make the code look more difficult to understand.

PR #5966, opened 2021-06-03
[Bug] Fix bug that start time is null when SQL is forward to master

e0c9:
WjJb1GujSl
When SQL need to forward to master,  the master node does not set the start time when it is processed.


fe/fe-core/src/main/java/org/apache/doris/qe/ConnectProcessor.java
```diff
@@ -324,7 +324,6 @@ private void dispatch() throws IOException {
             return;
         }
         ctx.setCommand(command);
-        ctx.setStartTime();
```
morningman, line 327: Other stmt also need to set startime. So I think we can't move this method here?

morningman (approved):
LGTM

PR #5991, opened 2021-06-08
[Bug] Fix range partition pruner bug

huangmengbin:
## Proposed changes
- Fix the bug in issue [5989](https://github.com/apache/incubator-doris/issues/5989).
```sql
CREATE TABLE xxx ( ..... ) 
UNIQUE KEY (`a`,`b`,`c`) 
PARTITION BY RANGE (`a`,`b`,`c`) (
    PARTITION p VALUES LESS THAN ('0')
)   ......
```
```sql
INSERT INTO xxx VALUES ('-1', '-1', '-1' .... );
```
```sql
EXPLAIN SELECT * FROM xxx where a > 0;
EXPLAIN SELECT * FROM xxx where a >= 0;
EXPLAIN SELECT * FROM xxx where a = 0;
EXPLAIN SELECT * FROM xxx where a in (0);
```
- Before fix the bug，the result of explain contains "partitions=1/1"
- After fix the bug，the result of explain contains "partitions=0/1"

## Types of changes

## Checklist

- [x] I have created an issue on [5989](https://github.com/apache/incubator-doris/issues/5989) and described the bug there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/RangePartitionPruner.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/RangePartitionPruner.java
@@ -122,11 +122,20 @@ public class RangePartitionPruner implements PartitionPruner {
             if (filter.lowerBound != null) {
                 minKey.pushColumn(filter.lowerBound, keyColumn.getDataType());
                 pushMinCount++;
-                if (filter.lowerBoundInclusive && columnIdx != lastColumnId) {
-                    Column column = partitionColumns.get(columnIdx + 1);
-                    Type type = Type.fromPrimitiveType(column.getDataType());
-                    minKey.pushColumn(LiteralExpr.createInfinity(type, false), column.getDataType());
-                    pushMinCount++;
+                if (filter.lowerBoundInclusive) {
+                    for (int i = columnIdx + 1; i <= lastColumnId; i++) {
+                        Column column = partitionColumns.get(i);
+                        Type type = Type.fromPrimitiveType(column.getDataType());
+                        minKey.pushColumn(LiteralExpr.createInfinity(type, false), column.getDataType());
+                        pushMinCount++;
+                    }
+                } else {
+                    for (int i = columnIdx + 1; i <= lastColumnId; i++) {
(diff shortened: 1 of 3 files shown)


huangmengbin:
- Sorry, I haven't fixed the bug completely this time. In the next step I will review all the usage of the "LiteralExpr.createInfinity" function to see if they are necessary to be included in the loop. It is expected that a new PR will be released this week.

huangmengbin:
Hi, @morningman, the bug has been fixed. Please review again. Thanks!

fe/fe-core/src/main/java/org/apache/doris/planner/RangePartitionPruner.java
```diff
@@ -75,19 +75,21 @@ public RangePartitionPruner(Map<Long, PartitionItem> rangeMap,
         Column keyColumn = partitionColumns.get(columnIdx);
         PartitionColumnFilter filter = partitionColumnFilters.get(keyColumn.getName());
         if (null == filter) {
-            minKey.pushColumn(LiteralExpr.createInfinity(Type.fromPrimitiveType(keyColumn.getDataType()), false),
-                    keyColumn.getDataType());
-            maxKey.pushColumn(LiteralExpr.createInfinity(Type.fromPrimitiveType(keyColumn.getDataType()), true),
-                    keyColumn.getDataType());
+            int pushMinCount = minKey.fillWithInfinity(partitionColumns, false);
```
EmmyMiao87, line 78: If there is no condition in the first column, should  we directly add the infinity range of the first column, right?

huangmengbin: > If there is no condition in the first column, should we directly add the infinity range of the first column, right?

Yes, it might be true. If the columnIdx is 0, we only need to add an infinity range of the first column. there is no need to add the infinity range for all columns. Thanks for your review !

huangmengbin: > If there is no condition in the first column, should we directly add the infinity range of the first column, right?

Done.

PR #6029, opened 2021-06-15
[Bug-fix] Fix wrong data distribution judgment

### LINKED ISSUE #6028: [Bug] Duplicate results in 'Group By' query
**Describe the bug**
When the query contains a group by clause, the result is wrong. 
There are duplicate groups in result.

**To Reproduce**
1. create a table with more than one partition
```
create table test (k1 int ,k2 int) partition by range(k1) (partition p1 values less than 1, partition p2 values less than 2)
distributed by hash(k1) buckets 32;
```
2. insert into some values in both partitions
3. query, duplicate results 
```
select k1 from test group by k1;
```
4. There is only one aggregation node in the plan tree. The aggregation node is wrongly evaluated as colocate.
```
explain select k1 from test group by k1;
```

**Expected behavior**
There is no duplicate k1 in result.

**Version**
Doris: 0.14.12

EmmyMiao87:
## Proposed changes

The Fragment where OlapScanNode is located has two data distribution possibilities.
1. RANDOM: Involving multi-partitioned tables in OlapScanNode.
2. HASH_PARTITIONED: The involving table is in the colocate group.

For a multi-partition table, although the data in each individual partition is distributed according to the bucketing column,
the same bucketing column between different partitions is not necessarily in the same be.
So the data distribution is RANDOM.

If Doris wrongly plan RANDOM as HASH_PARTITIONED, it will lead to the wrong colocate agg node.
The result of query is incorrect.

Fixed #6028

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #6028 ) and described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/OlapScanNode.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/OlapScanNode.java
@@ -795,9 +795,7 @@ public class OlapScanNode extends ScanNode {
     }
 
     public DataPartition constructInputPartitionByDistributionInfo() {
-        if (getNumInstances() == 1) {
-            return DataPartition.UNPARTITIONED;
-        } else if (Catalog.getCurrentColocateIndex().isColocateTable(olapTable.getId())) {
+        if (Catalog.getCurrentColocateIndex().isColocateTable(olapTable.getId())) {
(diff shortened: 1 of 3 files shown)


morningman (dismissed):
LGTM

morningman (approved):

PR #6034, opened 2021-06-15
[optimization] open gcc strict-aliasing optimization

stdpain:
## Proposed changes

use flag `-fno-strict-aliasing` won't generate SIMD Instructions for some vectorized case, so we need open it

## Types of changes

## Checklist

### DIFF
--- a/be/src/util/binary_cast.hpp
+++ b/be/src/util/binary_cast.hpp
@@ -1,0 +1,84 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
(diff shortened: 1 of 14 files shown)


be/src/util/binary_cast.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+namespace doris {
+union TypeConverter {
+    uint64_t u64;
+    int64_t i64;
+    uint32_t u32[2];
+    int32_t i32[2];
+    uint8_t u8[8];
+    float flt[2];
+    double db;
```
yangzhg, line 34:
```suggestion
    double dbl;
```

HappenLee (dismissed):
LGTM

EmmyMiao87:
If we remove this flag '-fno-strict-aliasing', will forced conversion of different types of pointers produce unexpected results?

stdpain:
@EmmyMiao87 
For example, an "unsigned int" can alias an "int", but not a "void*" or a "double".  A character type may alias any other type.

if we break this rules, gcc will generate a warning. I have fixed it.

EmmyMiao87:
> @EmmyMiao87
> For example, an "unsigned int" can alias an "int", but not a "void*" or a "double". A character type may alias any other type.
> 
> if we break this rules, gcc will generate a warning. I have fixed it.

So if there is no similar warning in the compile stage, there will be no problems in the run stage, right?

stdpain:
@EmmyMiao87 
right

morningman (approved):
LGTM

PR #6036, opened 2021-06-16
[optimization] _Tuple_ptrs into mempool when RowBatch is initialized

### LINKED ISSUE #6037: [Proposal] _Tuple_ptrs into mempool when RowBatch is initialized
When RowBatch is initialized, it will choose to allocate memory for `_tuple_ptrs` through malloc or mempool according to the value of `enable_partitioned_aggregation`.

stick _tuple_ptrs into mempool.

xinyiZzz:
## Proposed changes

Memory is allocated from MemPool, need to investigate the repercussions.

In the past, there were malloc'd and MemPool memory allocation methods. Malloc'd memory belongs to RowBatch itself, and the latter belongs to MemPool management. The memory ownership depends on whether legacy joins and aggs are enabled.

At present, it is uniformly allocated from MemPool, and tuple pointers are not transferred and do not have to be re-created in every Reset(), which has better performance.

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #6037) and described the bug/feature there in detail

## Further comments

need to investigate the repercussions


morningman (approved):
LGTM

PR #6047, opened 2021-06-16
[Enhance] Add config to limit the max size of simple tablet

weizuo93:
## Proposed changes

Add a new config `max_tablet_size_mb` to limit the max size of simple tablet. If the size of a tablet exceed limitation, load process will reject new incoming load job of this tablet. There is no limit to the size of the tablet if`max_tablet_size_mb` is set to 0.

## Types of changes

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -578,6 +578,11 @@ CONF_Int32(query_cache_max_partition_count, "1024");
 // This is to avoid too many version num.
 CONF_mInt32(max_tablet_version_num, "500");
 
+// Maximum size of a tablet, the unit is MB. If the size of a tablet exceed limit, the load process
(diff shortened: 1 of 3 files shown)


be/src/common/config.h
```diff
@@ -578,6 +578,11 @@ CONF_Int32(query_cache_max_partition_count, "1024");
 // This is to avoid too many version num.
 CONF_mInt32(max_tablet_version_num, "500");
 
+// Maximum size of a tablet, the unit is MB. If the size of a tablet exceed limit, the load process
+// will reject new incoming load job of this tablet. This is to avoid too large tablet.
+// There is no limit to tablet size when the parameter is set to "0".
+CONF_mInt32(max_tablet_size_mb, "0");
```
morningman, line 584: What can user do if a tablet's size exceed the limit?

PR #6051, opened 2021-06-17
[Bug-fix] Decimal Divide, Mod Zero Result Should be NULL.

### LINKED ISSUE #6049: [Bug] Decimal Value Divide, Mod Zero Result Should Be Null
**Describe the bug**
Now, Decimal Divide or Mod zero doris will get 0, but not null.

```
select 10.2 / 0.0;
+------------+
| 10.2 / 0.0 |
+------------+
|          0 |
+------------+
```

But in Mysql, no metter `int` or `decimal`, divide, mod zero always get Null
```
select 10.2 % 0.0;
+------------+
| 10.2 % 0.0 |
+------------+
|       NULL |
+------------+
```

HappenLee:
## Proposed changes

RT

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #6049) and described the bug/feature there in detail

### DIFF
--- a/be/src/runtime/decimalv2_value.cpp
+++ b/be/src/runtime/decimalv2_value.cpp
@@ -192,7 +192,7 @@ DecimalV2Value operator/(const DecimalV2Value& v1, const DecimalV2Value& v2) {
     int128_t x = v1.value();
     int128_t y = v2.value();
 
-    //todo: return 0 for divide zero
+    DCHECK(y != 0);
     if (x == 0 || y == 0) return DecimalV2Value(0);
     bool is_positive = (x > 0 && y > 0) || (x < 0 && y < 0);
     do_div(abs(x), abs(y), &result);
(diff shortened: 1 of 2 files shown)


be/src/runtime/decimalv2_value.cpp
```diff
@@ -192,7 +192,7 @@ DecimalV2Value operator/(const DecimalV2Value& v1, const DecimalV2Value& v2) {
     int128_t x = v1.value();
     int128_t y = v2.value();
 
-    //todo: return 0 for divide zero
+    DCHECK(y != 0);
     if (x == 0 || y == 0) return DecimalV2Value(0);
```
morningman, line 196:
```suggestion
    if (x == 0) return DecimalV2Value(0);
```

be/src/runtime/decimalv2_value.cpp
```diff
@@ -207,7 +207,7 @@ DecimalV2Value operator%(const DecimalV2Value& v1, const DecimalV2Value& v2) {
...
```
morningman, line 211:
```suggestion
    if (x == 0) return DecimalV2Value(0);
```

be/src/runtime/decimalv2_value.cpp
HappenLee: we have to do this check to prevent divide / zero which will cause core in BE.

morningman (approved):
LGTM

PR #6067, opened 2021-06-21
[Feature] ADD: show create routine load

### LINKED ISSUE #5645: Add show create routine load support
There is currently no`show create routine load`, so we need to save the routineLoad statements ourselves

Add show create routineload command

stalary:
## Proposed changes

Add show create routine load

Fixed #5645


stalary:
https://github.com/apache/incubator-doris/issues/5645

EmmyMiao87:
In fact, these information will be displayed in the results of show routine load.
Under what circumstances will a show create routine load syntax be needed?

morningman:
> In fact, these information will be displayed in the results of show routine load.
> Under what circumstances will a show create routine load syntax be needed?

Just like `show create table`, user may may need to create routine load job again.

stalary:
For example, I stopped it and then tried to restart it

fe/fe-core/src/main/java/org/apache/doris/analysis/ShowCreateRoutineLoadStmt.java
```diff
@@ -0,0 +1,73 @@
...
+        return labelName.getDbName();
+    }
+    
+    public String getLabel() {
+        return labelName.getLabelName();
+    }
+    
+    @Override
+    public void analyze(Analyzer analyzer) throws AnalysisException {
+        if (labelName == null) {
```
EmmyMiao87, line 52: Label_name in the grammar file is not optional, so it will not be empty.

fe/fe-core/src/main/java/org/apache/doris/qe/ShowExecutor.java
```diff
@@ -1834,7 +1837,51 @@ private void handleShowQueryProfile() throws AnalysisException {
...
     }
-
+    
+    private void handleShowCreateRoutineLoad() throws AnalysisException {
+        ShowCreateRoutineLoadStmt showCreateRoutineLoadStmt = (ShowCreateRoutineLoadStmt) stmt;
+        List<List<String>> rows = Lists.newArrayList();
+        // if job exists
+        RoutineLoadJob routineLoadJob;
+        try {
+            routineLoadJob = Catalog.getCurrentCatalog().getRoutineLoadManager()
```
EmmyMiao87, line 1847: getRoutineLoadManager().checkPrivAndGetJob()

fe/fe-core/src/main/java/org/apache/doris/qe/ShowExecutor.java
```diff
@@ -1834,7 +1837,51 @@ private void handleShowQueryProfile() throws AnalysisException {
...
+                    .getJob(showCreateRoutineLoadStmt.getDb(),
```
EmmyMiao87, line 1848: getJob() function will not displayed the routine load in the stopped state.

EmmyMiao87:
You need to consider a question. 
Whether the routine load after stopping also needs to be displayed.

The routine load after stopping may have the same name. 
So if show, it may produce multiple results.
The one way is that using 'show create routine load' to display running routine load, using 'show all create routine load' to displays all routine loads with the same name include stopped.

EmmyMiao87:
Fixed #5645

stalary:
> You need to consider a question.
> Whether the routine load after stopping also needs to be displayed.
> 
> The routine load after stopping may have the same name.
> So if show, it may produce multiple results.
> The one way is that using 'show create routine load' to display running routine load, using 'show all create routine load' to displays all routine loads with the same name include stopped.
At the moment my requirement is just to show running, do we need to support all of them？Show all create routine load is not very useful to me right now.

stalary:
> You need to consider a question.
> Whether the routine load after stopping also needs to be displayed.
> 
> The routine load after stopping may have the same name.
> So if show, it may produce multiple results.
> The one way is that using 'show create routine load' to display running routine load, using 'show all create routine load' to displays all routine loads with the same name include stopped.

Well, I've decided to support both

EmmyMiao87:
> At the moment my requirement is just to show running, do we need to support all of them？Show all create routine load is not very useful to me right now.

This is no problem. 
It needs to be stated on the document that only the routine load of running or paused is displayed.

fe/fe-core/src/main/java/org/apache/doris/analysis/ShowCreateRoutineLoadStmt.java
stalary: OK, thanks for the warning

stalary:
PTAL @EmmyMiao87

docs/zh-CN/sql-reference/sql-statements/Data Manipulation/SHOW CREATE ROUTINE LOAD.md
```diff
@@ -0,0 +1,42 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# SHOW CREATE ROUTINE LOAD
+## description
+    该语句用于展示用户自定义例行导入作业的创建语句
```
EmmyMiao87, line 29:
```suggestion
    该语句用于展示例行导入作业的创建语句
```

fe/fe-core/src/main/java/org/apache/doris/load/routineload/KafkaRoutineLoadJob.java
```diff
@@ -17,6 +17,7 @@
 
 package org.apache.doris.load.routineload;
 
+import java.util.HashMap;
```
EmmyMiao87, line 20: Pay attention to the ordering of imports

fe/fe-core/src/main/java/org/apache/doris/qe/ShowExecutor.java
```diff
@@ -1834,7 +1837,56 @@ private void handleShowQueryProfile() throws AnalysisException {
...
+    private void handleShowCreateRoutineLoad() throws AnalysisException {
+        ShowCreateRoutineLoadStmt showCreateRoutineLoadStmt = (ShowCreateRoutineLoadStmt) stmt;
+        List<List<String>> rows = Lists.newArrayList();
+        String dbName = showCreateRoutineLoadStmt.getDb();
+        String labelName = showCreateRoutineLoadStmt.getLabel();
+        // if include history return all create load
+        if (showCreateRoutineLoadStmt.isIncludeHistory()) {
+            List<RoutineLoadJob> routineLoadJobList = new ArrayList<>();
+            try {
+                routineLoadJobList = Catalog.getCurrentCatalog().getRoutineLoadManager().getJob(dbName, labelName, true);
```
EmmyMiao87, line 1850: The 'getJob' function will return null when there is no routine load has been found.

fe/fe-core/src/main/java/org/apache/doris/qe/ShowExecutor.java
```diff
@@ -1834,7 +1837,56 @@ private void handleShowQueryProfile() throws AnalysisException {
...
+                } catch (MetaNotFoundException e) {
+                    LOG.warn(new LogBuilder(LogKey.ROUTINE_LOAD_JOB, job.getId())
+                            .add("error_msg", "The table name for this routine load does not exist")
+                            .build(), e);
+                }
+                if (!Catalog.getCurrentCatalog().getAuth().checkTblPriv(ConnectContext.get(),
+                        dbName,
+                        tableName,
+                        PrivPredicate.LOAD)) {
+                    ErrorReport.reportAnalysisException(ErrorCode.ERR_TABLEACCESS_DENIED_ERROR, "LOAD",
```
EmmyMiao87, line 1869: If user don’t have permission, we can simply not display it without reporting an error

fe/fe-core/src/main/java/org/apache/doris/qe/ShowExecutor.java
```diff
@@ -1834,7 +1837,56 @@ private void handleShowQueryProfile() throws AnalysisException {
...
+                rows.add(Lists.newArrayList(String.valueOf(job.getId()), showCreateRoutineLoadStmt.getLabel(), job.getShowCreateInfo()));
+            }
+        } else {
+            // if job exists
+            RoutineLoadJob routineLoadJob;
+            try {
+                routineLoadJob = Catalog.getCurrentCatalog().getRoutineLoadManager().checkPrivAndGetJob(dbName, labelName);
+                // get routine load info
+                rows.add(Lists.newArrayList(String.valueOf(routineLoadJob.getId()), showCreateRoutineLoadStmt.getLabel(), routineLoadJob.getShowCreateInfo()));
+            } catch (Exception e) {
```
EmmyMiao87, line 1883: The exception caught here is too broad

EmmyMiao87:
Please add some UT~

fe/fe-core/src/main/java/org/apache/doris/load/routineload/KafkaRoutineLoadJob.java
stalary: ok, I'll adjust it later

PR #6083, opened 2021-06-22
[Feature] Extract wide common factors

### LINKED ISSUE #6082: [Feature] Extract wide common factors

If the where expression contains consecutive or clauses, the predicate can only be executed in higher-level operators. such as
```
select * from a, b where (1<a.k1<3 and b.k1 in ('a')) or (2<a.k1<4 and b.k1 in ('b'))
```
Because the two or clauses of where involve two tables respectively. So the expression can only be calculated in the join operator. This will cause the query to scan all tables a and b.

In fact, although these two clauses do not have a common factor, we can extract a wider range to filter the data in tables a and b in advance. Reduce the number of filtered rows.
Finally, perform precise filtering on the join operator.
such as:
For a.k1, the ranges in the two clauses can be combined to obtain:
1<a.k1<4, in this way, table a can be filtered in advance to reduce the amount of join data in table a b.

After rewriting, the query looks like:
```
select * from a,b where (1<a.k1<4) and (b.k1 in('a','b')) and (1<a.k1<3 and b.k1 in ('a' )) or (2<a.k1<4 and b.k1 in ('b'))
```

EmmyMiao87:
## Proposed changes
This PR mainly adds a rewrite rule 'ExtractCommonFactorsRule'
  used to extract wide common factors in the planning stage for 'Expr'.
The main purpose of this rule is to extract (Range or In) expressions
  that can be combined from each or clause.
E.g:
  Origin expr: (1<a<3 and b in ('a') ) or (2<a<4 and b in ('b'))
  Rewritten expr: (1<a<4 ) and (b in ('a', 'b')) and ((1<a<3 and b in ('a') ) or (2<a<4 and b in ('b')))
Although the range of the wide common factors is larger than the real range,
  the wide common factors only involve a single column, so it can be pushed down to the scan node,
  thereby reducing the amount of scanned data in advance and improving the query speed.

It should be noted that this optimization strategy is not for all scenarios.
When filter rate of the wide common factor is too low,
  the query will consume an extra time to calculate the wide common factors.

So this strategy can be switched by configuring session variables 'extract_wide_range_expr'.
The default policy is enabled which means this strategy takes effect.
If you encounter unsatisfactory filtering rate, you can set the variable to false.
It will turn off the strategy.

Fixed #6082

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #6082) and described the bug/feature there in detail

### DIFF
--- a/docs/zh-CN/administrator-guide/variables.md
+++ b/docs/zh-CN/administrator-guide/variables.md
@@ -383,3 +383,7 @@ SELECT /*+ SET_VAR(query_timeout = 1, enable_partition_cache=true) */ sleep(3);
     在一个排序的查询之中，一个上层节点接收下层节点有序数据时，会在exchange node上进行对应的排序来保证最终的数据是有序的。但是单线程进行多路数据归并时，如果数据量过大，会导致exchange node的单点的归并瓶颈。
 
     Doris在这部分进行了优化处理，如果下层的数据节点过多。exchange node会启动多线程进行并行归并来加速排序过程。该参数默认为False，即表示 exchange node 不采取并行的归并排序，来减少额外的CPU和内存消耗。
+
+* `extract_wide_range_expr`
+
+    用于控制是否开启 「宽泛公因式提取」的优化。取值有两种：true 和 false 。默认情况下开启。
--- a/fe/fe-core/src/main/java/org/apache/doris/rewrite/ExprRewriter.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/rewrite/ExprRewriter.java
@@ -17,12 +17,13 @@
 
 package org.apache.doris.rewrite;
 
-import java.util.List;
-
 import org.apache.doris.analysis.Analyzer;
 import org.apache.doris.analysis.Expr;
 import org.apache.doris.common.AnalysisException;
+
 import com.google.common.collect.Lists;
+
+import java.util.List;
 
 /**
  * Helper class that drives the transformation of Exprs according to a given list of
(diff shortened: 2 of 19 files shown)


docs/zh-CN/administrator-guide/variables.md
```diff
@@ -383,3 +383,7 @@ SELECT /*+ SET_VAR(query_timeout = 1, enable_partition_cache=true) */ sleep(3);
     在一个排序的查询之中，一个上层节点接收下层节点有序数据时，会在exchange node上进行对应的排序来保证最终的数据是有序的。但是单线程进行多路数据归并时，如果数据量过大，会导致exchange node的单点的归并瓶颈。
 
     Doris在这部分进行了优化处理，如果下层的数据节点过多。exchange node会启动多线程进行并行归并来加速排序过程。该参数默认为False，即表示 exchange node 不采取并行的归并排序，来减少额外的CPU和内存消耗。
+
+* `extract_wide_range_expr`
```
yangzhg, line 387: `extract_wide_range_expr` is the name of who this implementation, not the name of the function， you may change the name

EmmyMiao87: Maybe extract_common_factors?

fe/fe-core/src/main/java/org/apache/doris/rewrite/ExprRewriter.java
```diff
@@ -36,11 +37,17 @@
 public class ExprRewriter {
     private int numChanges_ = 0;
     private final List<ExprRewriteRule> rules_;
+    private List<ExprRewriteRule> onceRules_ = Lists.newArrayList();
```
morningman, line 40: Add comment to explain what is `onceRules`

yangzhg (dismissed):
LGTM

morningman (approved):
LGTM

fe/fe-core/src/main/java/org/apache/doris/rewrite/ExprRewriter.java
EmmyMiao87: OK

PR #6099, opened 2021-06-26
[Enhance] improve performance of init_scan_key by sharing the schema

huangmengbin:
## Proposed changes

- We find that the execution efficiency of statements like `where a in (...) and b in (...) ......` is very poor, especially when the `max_scan_key_num` and the number of column are large.
- Every time this function `RowCursor::init_scan_key` is called in the for-loop, a duplicate object, `schema`, which owns a `vector<Field*>` inside, will be created.
- So we want to construct this object outside the for-loop, and make it be shared by multiple `row_cursor`

## Types of changes

## Checklist

- [x] I have created an issue on (Enhancement #6098 ) and described the feature there in detail

## Further comments

- I am not sure if this change will break the original code structure.


be/src/olap/reader.cpp
```diff
@@ -548,14 +549,36 @@ OLAPStatus Reader::_init_keys_param(const ReaderParams& read_params) {
...
+                << "Input param are invalid. Column count is bigger than num_columns of schema. "
+                << "column_count=" << scan_key_size
+                << ", schema.num_columns=" << _tablet->tablet_schema().num_columns();
+        return OLAP_ERR_INPUT_PARAMETER_ERROR;
+    }
+
+    std::vector<uint32_t> columns;
+    for (size_t i = 0; i < scan_key_size; ++i) {
+        columns.push_back(i);
+    }
```
yangzhg, line 565: it is better to extract those to a function

huangmengbin: - Done. 
- I used the `std::iota` to replace the for-loop.
- Do I need to extra the "if (scan_key_size > _tablet->tablet_schema().num_columns()) { LOG(WARNING)...." to a function, and wrap it in the macro "RETURN_NOT_OK()" ?

huangmengbin: > it is better to extract those to a function, this is duplicted with line 158

- Done.
- I used the std::iota to replace the for-loop.
- Do I need to extra the "if (scan_key_size > _tablet->tablet_schema().num_columns()) { LOG(WARNING)...." to a function, and wrap it in the macro "RETURN_NOT_OK()" ?

yangzhg (approved):
LGTM

PR #6105, opened 2021-06-28
Align Date and Datetime by 16 bytes

EmmyMiao87:
The previous alignment of Doris is up to 8 bytes.
For types with more than 8 bytes, such as Date, Datetime is not aligned.
This PR is mainly to relax the maximum 8-byte limitation
Also, because the data type Decimal v1 no longer exists,
the logic of the 40-byte Decimal v1 is also discarded.


stdpain:
LGTM

yangzhg (approved):
LGTM

PR #6187, opened 2021-07-08
[Log] Optimize delete error message

qidaye:
## Proposed changes

When a delete error occurs, the error message is ambiguous.

```sql
mysql> DELETE FROM nebula_trade_health_trade PARTITION q3_2021 WHERE event_day = '20210706';
ERROR 1064 (HY000): errCode = 2, detailMessage = failed to execute delete. transaction id 7215554, timeout(ms) 160000, unfinished replicas: 4718319=7345841, 4718319=7345873, 4718319=7345749, 4718319=7345797, 4718319=7345829
```

We do not kown the meaning of `4718319=7345829`.

Actually the former is `BackendId` and the latter is `TabletId`.

I'll add a instruction here to help locate the problem quickly. The error message will be 

```sql
ERROR 1064 (HY000): errCode = 2, detailMessage = failed to execute delete. transaction id 7215554, timeout(ms) 160000, unfinished replicas [BackendId=TabletId]: 4718319=7345841, 4718319=7345873, 4718319=7345749, 4718319=7345797, 4718319=7345829
```


EmmyMiao87 (approved):
LGTM

PR #6196, opened 2021-07-09
[Feature] Add trash_used_capacity statistics at 'show backends'

### LINKED ISSUE #6195: [Feature] Add trash_used_capacity statistics at 'show backends'
Sometimes, garbage data will accumulate and take up a lot of space, but users don't know it. 

Add a column called 'Trash Used Capacity' when user query ‘show backends’ .
Trash Used Capacity represents the space occupied by the user trash data.

**Describe alternatives you've considered**
Users can manually delete trash after noticing that trash takes up a lot of space 

**Additional context**
图片

BiteTheDDDDt:
## Proposed changes

Add trash_used_capacity statistics at 'show backends' . 
Fix #6195

## Types of changes

- [X] New feature (non-breaking change which adds functionality)
- [X] Documentation Update (if none of the other choices apply)

## Checklist

- [X] I have created an issue on (Fix #6195 ) and described the bug/feature there in detail
- [X] Compiling and unit tests pass locally with my changes
- [X] If these changes need document changes, I have updated the document

### DIFF
--- a/be/src/agent/task_worker_pool.cpp
+++ b/be/src/agent/task_worker_pool.cpp
@@ -1150,6 +1150,9 @@ void TaskWorkerPool::_report_disk_state_worker_thread_callback() {
             disk.__set_disk_total_capacity(root_path_info.disk_capacity);
             disk.__set_data_used_capacity(root_path_info.data_used_capacity);
             disk.__set_disk_available_capacity(root_path_info.available);
+            std::filesystem::path trash_path(root_path_info.path + TRASH_PREFIX);
+            disk.__set_trash_used_capacity(
+                    _env->storage_engine()->get_file_or_directory_size(trash_path));
             disk.__set_used(root_path_info.is_used);
             disks[root_path_info.path] = disk;
         }
--- a/be/src/agent/task_worker_pool.cpp
+++ b/be/src/agent/task_worker_pool.cpp
@@ -1149,10 +1149,8 @@ void TaskWorkerPool::_report_disk_state_worker_thread_callback() {
             disk.__set_storage_medium(root_path_info.storage_medium);
             disk.__set_disk_total_capacity(root_path_info.disk_capacity);
             disk.__set_data_used_capacity(root_path_info.data_used_capacity);
+            disk.__set_trash_used_capacity(root_path_info.trash_used_capacity);
             disk.__set_disk_available_capacity(root_path_info.available);
-            std::filesystem::path trash_path(root_path_info.path + TRASH_PREFIX);
-            disk.__set_trash_used_capacity(
-                    _env->storage_engine()->get_file_or_directory_size(trash_path));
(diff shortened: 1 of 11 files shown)


be/src/agent/task_worker_pool.cpp
```diff
@@ -1150,6 +1150,9 @@ void TaskWorkerPool::_report_disk_state_worker_thread_callback() {
             disk.__set_disk_total_capacity(root_path_info.disk_capacity);
             disk.__set_data_used_capacity(root_path_info.data_used_capacity);
             disk.__set_disk_available_capacity(root_path_info.available);
+            std::filesystem::path trash_path(root_path_info.path + TRASH_PREFIX);
+            disk.__set_trash_used_capacity(
```
morningman, line 1154: Why not update the trash dir size in `DataDir::update_capacity()`, just like `data_used_capacity`?

fe/fe-core/src/main/java/org/apache/doris/common/util/DebugUtil.java
```diff
@@ -27,7 +27,7 @@
 import java.util.UUID;
 
 public class DebugUtil {
-    public static final DecimalFormat DECIMAL_FORMAT_SCALE_3 = new DecimalFormat("#.000");
+    public static final DecimalFormat DECIMAL_FORMAT_SCALE_3 = new DecimalFormat("0.000");
```
morningman, line 30: Why change this? Is it some thing wrong with origin `#.000`?

BiteTheDDDDt: It will display incorrectly when the value is less than 1. For example, if the value is 0.12, it will display **.120** instead of **0.120** .

be/src/agent/task_worker_pool.cpp
BiteTheDDDDt: Now I moved it to StorageEngine::get_all_data_dir_info. In fact, DataDir::update_capacity() only updates disk_capacity and available, while data_used_capacity and trash_used_capacity have different update methods.

be/src/olap/storage_engine.cpp
```diff
@@ -357,6 +357,12 @@ OLAPStatus StorageEngine::get_all_data_dir_info(vector<DataDirInfo>* data_dir_in
     size_t tablet_count = 0;
     _tablet_manager->update_root_path_info(&path_map, &tablet_count);
 
+    // update trash used capacity
+    for (auto& path : path_map) {
+        std::filesystem::path trash_path(path.first + TRASH_PREFIX);
```
stdpain, line 362: I recommend not using the right value here to construct std::filesystem::path directly because it might trigger one of gcc 10.1.0

BiteTheDDDDt: fixed.

PR #6200, opened 2021-07-10
[Performance] Optimize the performance of tabletReport

### LINKED ISSUE #6201: [Performance] ReportHandler has poor performance for tabletReport when there are many tablets on one backend
We found that ReportHandler has poor performance for tabletReport when there are many tablets on one backend. For example, in our production environment, one backend may have 27000+ tablet, and tabletReport for one backend may cost 1000+ms, which indicate poor performance.

After we doing some optimization for tabletReport,  tabletReport only cost 65ms - 75ms

caiconghui:
## Proposed changes
1. Use parallelStream to speed up tabletReport.
2. Add partitionIdInMemorySet to speed up tabletToInMemory check.
3. Add disable_storage_medium_check to disable storage medium check when user doesn't care what tablet's storage medium is, and remove enable_strict_storage_medium_check config to fix some potential migration task failures.

## Types of changes

- [ ] Other.

## Checklist

- [x] I have created an issue on (Fix #6201) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/docs/en/administrator-guide/config/fe_config.md
+++ b/docs/en/administrator-guide/config/fe_config.md
@@ -374,7 +374,7 @@ IsMutable：true
 
 MasterOnly：true
 
-If disable_storage_medium_check is true, ReportHandler would not check tablet's storage medium and disable storage cool down function, the default value is false. You can set the value false when you don't care what the storage medium of the tablet is.
+If disable_storage_medium_check is true, ReportHandler would not check tablet's storage medium and disable storage cool down function, the default value is false. You can set the value true when you don't care what the storage medium of the tablet is.
   
 ### drop_backend_after_decommission
 
--- a/docs/en/administrator-guide/config/fe_config.md
+++ b/docs/en/administrator-guide/config/fe_config.md
@@ -366,7 +366,7 @@ MasterOnly：true
 
 Whether to enable spark load temporarily, it is not enabled by default
 
-### enable_strict_storage_medium_check
+### disable_storage_medium_check
 
 Default：false
 
@@ -374,12 +374,8 @@ IsMutable：true
 
 MasterOnly：true
 
-This configuration indicates that when the table is being built, it checks for the presence of the appropriate storage medium in the cluster. For example, when the user specifies that the storage medium is' SSD 'when the table is built, but only' HDD 'disks exist in the cluster,
-
-If this parameter is' True ', the error 'Failed to find enough host in all Backends with storage medium with storage medium is SSD, need 3'.
-
-If this parameter is' False ', no error is reported when the table is built. Instead, the table is built on a disk with 'HDD' as the storage medium
-
+If disable_storage_medium_check is true, ReportHandler would not check tablet's storage medium and disable storage cool down function, the default value is false. You can set the value false when you don't care what the storage medium of the tablet is.
+  
 ### drop_backend_after_decommission
 
 Default：false
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/PartitionInfo.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/PartitionInfo.java
@@ -133,7 +133,7 @@ public class PartitionInfo implements Writable {
 
         idToDataProperty.put(partitionId, desc.getPartitionDataProperty());
         idToReplicationNum.put(partitionId, desc.getReplicationNum());
-        idToInMemory.put(partitionId, desc.isInMemory());
+        setIsInMemory(partitionId, desc.isInMemory());
 
         return partitionItem;
     }
@@ -148,7 +148,7 @@ public class PartitionInfo implements Writable {
         setItemInternal(partitionId, isTemp, partitionItem);
         idToDataProperty.put(partitionId, dataProperty);
         idToReplicationNum.put(partitionId, replicationNum);
-        idToInMemory.put(partitionId, isInMemory);
+        setIsInMemory(partitionId, isInMemory);
     }
 
     public List<Map.Entry<Long, PartitionItem>> getPartitionItemEntryList(boolean isTemp, boolean isSorted) {
@@ -264,7 +264,7 @@ public class PartitionInfo implements Writable {
                              boolean isInMemory) {
         idToDataProperty.put(partitionId, dataProperty);
         idToReplicationNum.put(partitionId, replicationNum);
-        idToInMemory.put(partitionId, isInMemory);
+        setIsInMemory(partitionId, isInMemory);
     }
 
     public static PartitionInfo read(DataInput in) throws IOException {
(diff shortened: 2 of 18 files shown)


morningman:
How much improvement does this new implementation have?

caiconghui:
> How much improvement does this new implementation have?

About seven to ten times improvement for tabletReport of TabletInvertedIndex, and operation outside this function cost less than 1 ms when all tablet are normal and we skip tablet storage medium check.

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -735,9 +735,17 @@
...
     @ConfField(mutable = true, masterOnly = true)
     public static int max_backend_down_time_second = 3600; // 1h
+
+    /**
+     * If disable_storage_medium_check is true, ReportHandler would not check tablet's storage medium
+     * and disable storage cool down function, the default value is false.
+     * You can set the value true when you don't care what the storage medium of the tablet is.
+     */
+    @ConfField(mutable = true, masterOnly = true)
+    public static boolean disable_storage_medium_check = false;
```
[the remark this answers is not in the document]
caiconghui: enable_strict_storage_medium_check make the user confused, if we really need to check storage_medium, we should not create table without consider storage medium and then produce many migration task usually failed because that user not use this feature. If we not care the storage medium, just disable this check, which can save a lot of time for tablet report, because it is meanless to distinguish ssd from hdd for some users while user not need cool down feature.

fe/fe-core/src/main/java/org/apache/doris/catalog/TabletInvertedIndex.java
```diff
@@ -484,40 +512,40 @@ public void setNewSchemaHash(long partitionId, long indexId, int newSchemaHash)
         if (Catalog.isCheckpointThread()) {
             return;
         }
-        writeLock();
+        readLock();
```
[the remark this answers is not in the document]
caiconghui: actually here not modify any structure for TabletInvertedIndex, and when we modify schemahash we hold the tablet meta write lock of tablet meta

fe/fe-core/src/main/java/org/apache/doris/master/ReportHandler.java
```diff
@@ -17,6 +17,7 @@
 
 package org.apache.doris.master;
 
+import com.google.common.collect.Sets;
```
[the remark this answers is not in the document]
caiconghui: ok

fe/fe-core/src/main/java/org/apache/doris/master/ReportHandler.java
```diff
@@ -660,51 +679,64 @@ private static void deleteFromBackend(Map<Long, TTablet> backendTablets,
...
-                    LOG.warn("delete tablet[" + tabletId + " - " + backendTabletInfo.getSchemaHash()
-                            + "] from backend[" + backendId + "] because not found in meta");
-                    ++deleteFromBackendCounter;
-                }
-            } // end for tabletInfos
-
-            if (foundTabletsWithInvalidSchema.containsKey(tabletId)) {
-                // this tablet is found in meta but with invalid schema hash.
-                // delete it.
+        if (foundTabletsWithValidSchema.size() + foundTabletsWithInvalidSchema.size() == backendTablets.size()) {
```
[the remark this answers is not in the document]
caiconghui: ok

fe/fe-core/src/main/java/org/apache/doris/catalog/PartitionInfo.java
```diff
@@ -282,6 +282,14 @@ public void moveFromTempToFormal(long tempPartitionId) {
         }
     }
 
+    public void resetPartitionIdForRestore(long newPartitionId, long oldPartitionId, short restoreReplicationNum) {
+        idToDataProperty.put(newPartitionId, idToDataProperty.remove(oldPartitionId));
+        idToReplicationNum.remove(oldPartitionId);
+        idToReplicationNum.put(newPartitionId, restoreReplicationNum);
+        idToItem.put(newPartitionId, idToItem.remove(oldPartitionId));
```
morningman, line 289: For unpartitioned table, the `idToItem` is empty.

fe/fe-core/src/main/java/org/apache/doris/master/PartitionInMemoryInfoCollector.java
```diff
@@ -0,0 +1,95 @@
...
+        super("PartitionInMemoryInfoCollector", Config.partition_in_memory_update_interval_secs * 1000);
+    }
+
+    @Override
+    protected void runAfterCatalogReady() {
+        updatePartitionInMemoryInfo();
+    }
+
+    private void updatePartitionInMemoryInfo() {
+        Catalog catalog = Catalog.getCurrentCatalog();
```
morningman, line 47: I think we can just clear the `partitionIdInMemorySet` and set it again .
So that we don't need to care about the partition's adding and dropping operation.
That will make things more easy.

morningman (approved):
LGTM

docs/en/administrator-guide/config/fe_config.md
```diff
@@ -366,20 +366,16 @@ MasterOnly：true
 
 Whether to enable spark load temporarily, it is not enabled by default
 
-### enable_strict_storage_medium_check
+### disable_storage_medium_check
```
acelyc111, line 369: It's not a good idea to change a config key for compatiable reason. Suppose a user set `enable_strict_storage_medium_check` to true, after he/she upgraded, the old config has no effect and would complain.

caiconghui: Actually, now, this config is a bad design, so I forcibly modified this configuration. For users, it can be divided into two categories, one is concerned about the storage media, the other is not concerned. if enable_strict_storage_medium_check is true, the behavior is same as that disable_storage_medium_check is false. If  enable_strict_storage_medium_check is false, it would cause continuous migration task failure which is a bug that should be fixed.

acelyc111: It's not necessary to change the config name when you fix the bug.

caiconghui: enable_strict_storage_medium_check is not same as  enable_storage_medium_check, if we don't remove this config, it is still a bug, you could see the comment about this config. and enable_strict_storage_medium_check -> enable_storage_medium_check or disable_storage_medium_check, I think it is all right. for older user, new config would let user know the wrong usage for the storage medium, and for new user, for new user, we have already correct usage

acelyc111: Does the bug exits in all cases?

acelyc111: further more, if it's a bug, better to describe it on PR title or description, or lable.

caiconghui: > Does the bug exist in all cases?

for enable_strict_storage_medium_check=false, on condition that most user cannot be aware this problem, but it may cause unstabitily of fe.

caiconghui: > BTW, if it's a bug, better to describe it on PR title or description, or add a `bug` lable.

ok

acelyc111 (approved):
LGTM

PR #6207, opened 2021-07-12
[New Feature][Meta][Image] Add file header and footer for image

### LINKED ISSUE #6206: [New Feature][Meta] Add file header and footer for image.
At present, our image file does not have file header/footer. When we need to change the image format (such as adding different journal versions to the image), there is no way to distinguish different image formats.

Therefore, we suggest adding file header and footer to the image. By the new image format, we can freely distinguish and define different image reading ways.

The format of the image is as follows:
```
 Image Format:
 |- Image --------------------------------------|
 | - Magic String (4 bytes)                     |
 | - Header Length (4 bytes)                    |
 | |- Header -----------------------------|     |
 | | |- Json Header ---------------|      |     |
 | | | - version                   |      |     |
 | | | - other key/value(undecided)|      |     |
 | | |-----------------------------|      |     |
 | |--------------------------------------|     |
 |                                              |
 | |- Image Body -------------------------|     |
 | | Object a                             |     |
 | | Object b                             |     |
 | | ...                                  |     |
 | |--------------------------------------|     |
 |                                              |
 | |- Footer -----------------------------|     |
 | | | - Checksum (8 bytes)               |     |
 | | |- object index --------------|      |     |
 | | | - index a                   |      |     |
 | | | - index b                   |      |     |
 | | | ...                         |      |     |
 | | |-----------------------------|      |     |
 | | - other value(undecided)             |     |
 | |--------------------------------------|     |
 | - Footer Length (8 bytes)                    |
 | - Magic String (4 bytes)                     |
 |----------------------------------------------|
 ```

1. Magic Nuumber
One image format is identified by one magic string and one version field. The magic string is save in the first 4 bytes and last 4 bytes in the images.

2. Image Header:
The version is save in the header with json format now.

3. Image Body:
Equal to the original image.

4.Image Footer:
Image footer stores the file offset(index) of many image objects.

### PR DESCRIPTION
**Pull request by xy720**
## Proposed changes

#6206 

At present, our image file does not have file header/footer. When we need to change the image format (such as adding different journal versions to the image), there is no way to distinguish different image formats.

Therefore, we suggest adding file header and footer to the image. By the new image format, we can freely distinguish and define different image reading ways.

The format of the image is as follows:

```
/**
 * Image Format:
 * |- Image --------------------------------------|
 * | - Magic String (4 bytes)                     |
 * | - Header Length (4 bytes)                    |
 * | |- Header -----------------------------|     |
 * | | |- Json Header ---------------|      |     |
 * | | | - version                   |      |     |
 * | | | - other key/value(undecided)|      |     |
 * | | |-----------------------------|      |     |
 * | |--------------------------------------|     |
 * |                                              |
 * | |- Image Body -------------------------|     |
 * | | Object a                             |     |
 * | | Object b                             |     |
 * | | ...                                  |     |
 * | |--------------------------------------|     |
 * |                                              |
 * | |- Footer -----------------------------|     |
 * | | | - Checksum (8 bytes)               |     |
 * | | |- object index --------------|      |     |
 * | | | - index a                   |      |     |
 * | | | - index b                   |      |     |
 * | | | ...                         |      |     |
 * | | |-----------------------------|      |     |
 * | | - other value(undecided)             |     |
 * | |--------------------------------------|     |
 * | - Footer Length (8 bytes)                    |
 * | - Magic String (4 bytes)                     |
 * |----------------------------------------------|
 */
```
1. Magic Nuumber
One image format is identified by one magic string and one version field. The magic string is save in the first 4 bytes and last 4 bytes in the images.

2. Image Header:
The version is save in the header with json format now.

3. Image Body:
Equal to the original image.

4.Image Footer:
Image footer stores the file offset(index) of many image objects. If necessary, we can read some objects in the image by the footer.

## Types of changes

## Checklist

- [x] I have created an issue on (Fix #6206 ) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/common/FeMetaFormat.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/FeMetaFormat.java
@@ -1,0 +1,39 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package org.apache.doris.common;
+
+public enum FeMetaFormat {
+    COR1("COR1", "v1"),
+    ETL1("ETL1", "v1");
+
+    private final String magicString;
+    private final String version;
+
+    private FeMetaFormat(String magicString, String version) {
+        this.magicString = magicString;
+        this.version = version;
+    }
+
+    public String getMagicString() {
+        return magicString;
+    }
+
+    public String getVersion() {
+        return version;
+    }
+}
--- a/fe/fe-core/src/main/java/org/apache/doris/common/FeMetaFormat.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/FeMetaFormat.java
@@ -18,8 +18,7 @@
 package org.apache.doris.common;
 
 public enum FeMetaFormat {
-    COR1("COR1", "v1"),
-    ETL1("ETL1", "v1");
+    COR1("COR1", "v1");
 
     private final String magicString;
     private final String version;
--- a/fe/fe-core/src/main/java/org/apache/doris/common/MetaFooter.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/MetaFooter.java
@@ -1,0 +1,86 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package org.apache.doris.common;
+
+import com.google.common.collect.Lists;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.RandomAccessFile;
+import java.util.Arrays;
+import java.util.List;
+
+public class MetaFooter {
+    private static final Logger LOG = LogManager.getLogger(MetaFooter.class);
+
+    private static final long FOOTER_LENGTH_SIZE = 8L;
+    public static final MetaFooter EMPTY_FOOTER = new MetaFooter(null, 0L);
+
+    // length of footer
+    private long length;
+    // meta indices
+    public List<MetaIndex> metaIndices;
+
+    public static MetaFooter read(File imageFile) throws IOException {
+        try(RandomAccessFile raf = new RandomAccessFile(imageFile, "r")) {
+            long fileLength = raf.length();
+            long footerLengthIndex = fileLength - FOOTER_LENGTH_SIZE - MetaMagicNumber.MAGIC_STR.length();
+            raf.seek(footerLengthIndex);
+            long footerLength = raf.readLong();
+            MetaMagicNumber magicNumber = MetaMagicNumber.read(raf);
+            if (!Arrays.equals(MetaMagicNumber.MAGIC, magicNumber.getBytes())) {
+                LOG.warn("Image file {} format mismatch. Expected magic number is {}, actual is {}",
+                        imageFile.getPath(), Arrays.toString(MetaMagicNumber.MAGIC), Arrays.toString(magicNumber.getBytes()));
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: fe/fe-core/src/main/java/org/apache/doris/common/MetaFooter.java [hunk 1]
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.RandomAccessFile;
+import java.util.Arrays;
+import java.util.List;
+
+public class MetaFooter {
```
morningman, line 31: You can add comment of the format of this footer

fe/fe-core/src/main/java/org/apache/doris/common/MetaFooter.java
```diff
@@ -0,0 +1,86 @@
...
+        try(RandomAccessFile raf = new RandomAccessFile(imageFile, "r")) {
+            long fileLength = raf.length();
+            long footerLengthIndex = fileLength - FOOTER_LENGTH_SIZE - MetaMagicNumber.MAGIC_STR.length();
+            raf.seek(footerLengthIndex);
+            long footerLength = raf.readLong();
+            MetaMagicNumber magicNumber = MetaMagicNumber.read(raf);
+            if (!Arrays.equals(MetaMagicNumber.MAGIC, magicNumber.getBytes())) {
+                LOG.warn("Image file {} format mismatch. Expected magic number is {}, actual is {}",
+                        imageFile.getPath(), Arrays.toString(MetaMagicNumber.MAGIC), Arrays.toString(magicNumber.getBytes()));
+                return EMPTY_FOOTER;
```
morningman, line 52: I think we should throw an exception here, because this is an invalid file, not a "empty" one.
And why put MagicNumber in both header and footer? I think one place is enough.

fe/fe-core/src/main/java/org/apache/doris/common/MetaHeader.java
```diff
@@ -0,0 +1,103 @@
...
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.RandomAccessFile;
+import java.util.Arrays;
+
+public class MetaHeader {
```
morningman, line 28: Add format in comment

fe/fe-core/src/main/java/org/apache/doris/common/MetaHeader.java
```diff
@@ -0,0 +1,103 @@
...
+    public MetaJsonHeader metaJsonHeader;
+
+    public static MetaHeader read(File imageFile) throws IOException {
+        try(RandomAccessFile raf = new RandomAccessFile(imageFile, "r")) {
+            raf.seek(0);
+            MetaMagicNumber magicNumber = MetaMagicNumber.read(raf);
+            if (!Arrays.equals(MetaMagicNumber.MAGIC, magicNumber.getBytes())) {
+                LOG.warn("Image file {} format mismatch. Expected magic number is {}, actual is {}",
+                        imageFile.getPath(), Arrays.toString(MetaMagicNumber.MAGIC), Arrays.toString(magicNumber.getBytes()));
+                return EMPTY_HEADER;
```
morningman, line 48: throw exception

fe/fe-core/src/main/java/org/apache/doris/common/MetaWriter.java
```diff
@@ -0,0 +1,131 @@
...
+            checksum = catalog.saveRoutineLoadJobs(dos, checksum);
+            metaIndices.add(new MetaIndex("loadJobV2", dos.getCount()));
+            checksum = catalog.saveLoadJobsV2(dos, checksum);
+            metaIndices.add(new MetaIndex("smallFiles", dos.getCount()));
+            checksum = catalog.saveSmallFiles(dos, checksum);
+            metaIndices.add(new MetaIndex("plugins", dos.getCount()));
+            checksum = catalog.savePlugins(dos, checksum);
+            metaIndices.add(new MetaIndex("deleteHandler", dos.getCount()));
+            checksum = catalog.saveDeleteHandler(dos, checksum);
+            dos.writeLong(checksum);
```
morningman, line 122: checksum should be wrote in footer.

fe/fe-core/src/main/java/org/apache/doris/common/MetaFooter.java
xy720: When we read a old version image, the magic number is not exist, if we throw a exception here, it will fail to read the image. So my idea is, if we can't read the correct magic number here, then read this file as an old version of the image, and the old version of the image has no header or footer, so an empty header or footer is returned here.

fe/fe-core/src/main/java/org/apache/doris/common/MetaHeader.java
xy720: The same as above.

fe/fe-core/src/main/java/org/apache/doris/common/MetaWriter.java
xy720: ok

fe/fe-core/src/main/java/org/apache/doris/common/FeMetaFormat.java
```diff
@@ -0,0 +1,44 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package org.apache.doris.common;
+
+public enum FeMetaFormat {
+    COR1("COR1", "v1"),
+    ETL1("ETL1", "v1");
```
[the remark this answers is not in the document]
xy720: May be shorter is better? Magic string is 4 bytes just like parquet file.

fe/fe-core/src/main/java/org/apache/doris/common/MetaHeader.java
xy720: done

fe/fe-core/src/main/java/org/apache/doris/common/MetaFooter.java
xy720: done

morningman (approved):
LGTM

PR #6236, opened 2021-07-14
Optimized the reading performance of Duplicate tables with severe data skew

yangzhg:
## Proposed changes

When dividing the scanner, there is no longer a scanner for each
tablet, but according to the file size of the tablet, the tablet is
divided into multiple scanners for reading. By default, each scanner
reads 1GB. If the data is stored on a SATA disk, this value can be
increased, and if it is an SSD disk, the value can be decreased.

## Types of changes

- [ ] Other.

## Checklist

- [x] I have created an issue on (Fix #skew) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/be/src/olap/tablet.cpp
+++ b/be/src/olap/tablet.cpp
@@ -700,10 +700,10 @@ OLAPStatus Tablet::capture_rs_readers_by_range(const Version& spec_version,
 
 OLAPStatus Tablet::capture_rs_readers_by_rowsets(const std::vector<Version>& version_path,
                                                  std::vector<RowsetReaderSharedPtr>* rs_readers,
-                                                 const std::set<RowsetId>& segments,
+                                                 const std::set<RowsetId>& rowset_ids,
                                                  std::shared_ptr<MemTracker> parent_tracker) const {
     DCHECK(rs_readers != nullptr && rs_readers->empty());
-    if (segments.empty()) {
+    if (rowset_ids.empty()) {
         return capture_rs_readers(version_path, rs_readers, parent_tracker);
     }
     for (auto version : version_path) {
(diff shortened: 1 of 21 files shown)


be/src/olap/tablet.cpp
```diff
@@ -675,6 +676,62 @@ OLAPStatus Tablet::capture_rs_readers(const std::vector<Version>& version_path,
...
+    bool has_segment_range = false;
+    for (auto key_range : key_ranges) {
+        if (key_range->range_type == ScanRangeType::SEGMENT) {
+            for (const auto& s : key_range->segments) {
+                rowsets.emplace(s.first);
+            }
+            has_segment_range = true;
+        }
+    }
+    if (!has_segment_range) {
```
morningman, line 693: Is it possible to have both KEY and SEGMENT key range in `key_ranges`?
If yes, we should clear the `rs_readers` before calling `capture_rs_readers()`

be/src/olap/tablet_meta.h
```diff
@@ -285,6 +289,38 @@ inline void TabletMeta::set_cumulative_layer_point(int64_t new_point) {
     _cumulative_layer_point = new_point;
 }
 
+inline bool TabletMeta::contains_delete(const Version& version) const {
+    for (auto& rs : _rs_metas) {
```
morningman, line 293: I am not sure if rs in _rs_metas are ordered by version?

yangzhg: no need to order by version, if contains delete version this tablet will not split by segmnent

be/src/olap/tablet.cpp
yangzhg: if key_ranges  have SEGMENT key range, it will have KEY key range

be/src/olap/tablet.cpp
```diff
@@ -675,6 +676,62 @@ OLAPStatus Tablet::capture_rs_readers(const std::vector<Version>& version_path,
...
+        return capture_rs_readers(spec_version, rs_readers, parent_tracker);
+    }
+    std::vector<Version> version_path;
+    RETURN_NOT_OK(capture_consistent_versions(spec_version, &version_path));
+    return capture_rs_readers_by_rowsets(version_path, rs_readers, rowsets, parent_tracker);
+}
+
+OLAPStatus Tablet::capture_rs_readers_by_rowsets(const std::vector<Version>& version_path,
+                                                 std::vector<RowsetReaderSharedPtr>* rs_readers,
+                                                 const std::set<RowsetId>& segments,
```
morningman, line 703:
```suggestion
                                                 const std::set<RowsetId>& rowset_ids,
```

PR #6237, opened 2021-07-14
[Feature] Support setting concurrency for thread pool token

morningman:
## Proposed changes

Now we can submit a group of tasks using thread pool token, and limit
the max concurrency of this task group

## Types of changes

- [ ] Other.

## Checklist

## Further comments


be/src/util/threadpool.h
```diff
@@ -362,6 +362,13 @@ class ThreadPoolToken {
     // Returns true if all submissions are complete, false otherwise.
     bool wait_for(const MonoDelta& delta);
 
+    bool need_dispatch();
```
yangzhg, line 365: need_dispatch is a bit weird. Idle may be better or have other better names

weizuo93:
Based on this patch, base compaction and cumulative compaction could be submitted to thread pool by different tokens. So we could resolve the problem that too many threads are occupied by  base compaction.

morningman:
> Based on this patch, base compaction and cumulative compaction could be submitted to thread pool by different tokens. So we could resolve the problem that too many threads are occupied by base compaction.

Actually, currently implementation of compaction task generation has already reserved at least one thread for cumulative compaction. But base on thread token, it can be implemented easier.

yangzhg (approved):
LGTM

be/src/util/threadpool.h
morningman: The purpose of this `need_dispatch` is to express whether the token needs to be put into the queue, so it does not actually mean idle.

PR #6274, opened 2021-07-19
[Feature] Modify the cost evaluation algorithm of Broadcast and Shuffle Join

### LINKED ISSUE #6272: [Proposal] Modify the cost evaluation algorithm of Broadcast and Shuffle Join
## Motivation
When planning the Join type, the estimated memory size required for Broadcast Join HashTable is inaccurate, which is found in the TPC-DS test.

The reason for OOM is that during FE planning, it is estimated that the memory value required by Broadcast Join to construct HashTable is too small, which leads to OOM when constructing HashTable after planning into Broadcast Join.

Currently, it is estimated that the memory cost constructed by Hash Table = `cardinality * avgRowSize` in the right table,  `avgRowSize` is equal to the average data size of one row of the right table, that is `the amount of data in the right table/the number of rows`, but the actual Join in the HashTable construction is part of the column of the right table.

Reasons that can be passed before opening Join Reorder, is that the value of `avgRowSize` in the right table of Join is larger than the actual value. Because `Cardinality * avgRowSize > memory limit`, it is planned as Shuffle Join, thus avoiding OOM. 

In addition, the memory cost calculated in this way only contains part of the actual cost, without considering the cost of the node and bucket pointers, and the additional cost of the node itself, so the final Join planning result is uncontrollable.

## Design
When planning the join type, modify the cost evaluation algorithm of Broadcast and Shuffle Join to choose between the two.

Memory cost of constructing Hash Table in Broadcast Join, the modified evaluation algorithm is as follows:
```
memory cost by the Hash Table = 
    ((cardinality/0.75[1']) * 8[2'])[3'] 
    + (cardinality * avgRowSize)[4'] 
    + (nodeArrayLen[5'] * 16[6'])[7'] 
    + (nodeArrayLen * tupleNum[8'] * 8[9'])[10']


Corresponding to four parts:
    - All bucket pointers. 
    - Length of the node array. 
    - Overhead of all nodes. 
    - Tuple pointers of all nodes.

The meaning of each value in the formula:
    - [1'] Expansion factor of the number of HashTable buckets;
    - [2'] The pointer length of each bucket of HashTable;
    - [3'] bucketPointerSpace: The memory cost by all bucket pointers of HashTable;
    - [4'] rhsDataSize: The memory cost by all nodes of HashTable, equal to the amount of data that the right table participates in the construction of HashTable;
    - [5'] HashTable stores the length of the node array, which is larger than the actual cardinality. 
        The initial value is 4096. When the storage is full, one-half of the current array length is added each time. 
        The length of the array after each increment is actually a sequence of numbers:
            4096 = pow(3/2, 0) * 4096,
            6144 = pow(3/2, 1) * 4096,
            9216 = pow(3/2, 2) * 4096,
            13824 = pow(3/2, 3) * 4096,
        finally need to satisfy len(node array)> cardinality, so the number of increments = int((ln(cardinality/4096) / ln(3/2)) + 1), finally len(node array) = pow(3/2, int((ln(cardinality/4096) / ln(3/2)) + 1) * 4096
    - [6'] The overhead length of each node of HashTable, including the next node pointer, Hash value, and a bool type variable;
    - [7'] nodeOverheadSpace: The memory cost by the overhead of all nodes in the HashTable;
    - [8'] Number of Tuples participating in the build;
    - [9'] The length of each Tuple pointer;
    - [10'] nodeTuplePointerSpace: The memory cost by Tuple pointers of all nodes in HashTable;
```

When broadcastCost and partitionCost are equal, there is no uniform standard for which join implementation is better. Some scenarios are suitable for broadcast join, and some scenarios are suitable for shuffle join. 

Therefore, provide a SessionVariable to help users choose a better join implementation.
```
set prefer_join_method=broadcast;
```

## future work
Design Fragment memory cost estimation algorithm

At present, only a single Join can estimate whether the memory cost will reach the Fragment memory-limit. There is no memory cost evaluation method for Fragments. 

If a Fragment contains multiple Joins, or contains other operations that consume a lot of memory, it may be due to insufficient Fragment memory. Lead to OOM.

xinyiZzz:
## Proposed changes
issue #6272

After modifying the Hash Table memory cost estimation algorithm, TPC-DS query-72 etc. can be passed, avoiding OOM.

## Types of changes

- [ ] Other.

## Checklist

- [x] I have created an issue on (Fix #6272) and described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/DistributedPlanner.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/DistributedPlanner.java
@@ -348,53 +348,7 @@ public class DistributedPlanner {
             return leftChildFragment;
         }
 
-        // broadcast: send the rightChildFragment's output to each node executing
-        // the leftChildFragment; the cost across all nodes is proportional to the
-        // total amount of data sent
-
-        // NOTICE:
-        // for now, only MysqlScanNode and OlapScanNode has Cardinality.
-        // OlapScanNode's cardinality is calculated by row num and data size,
-        // and MysqlScanNode's cardinality is always 0.
-        // Other ScanNode's cardinality is -1.
-        //
-        // So if there are other kind of scan node in join query, it won't be able to calculate the cost of
-        // join normally and result in both "broadcastCost" and "partitionCost" be 0. And this will lead
-        // to a SHUFFLE join.
-        PlanNode rhsTree = rightChildFragment.getPlanRoot();
-        long rhsDataSize = 0;
-        long broadcastCost = 0;
-        if (rhsTree.getCardinality() != -1 && leftChildFragment.getNumNodes() != -1) {
-            rhsDataSize = Math.round((double) rhsTree.getCardinality() * rhsTree.getAvgRowSize());
-            broadcastCost = rhsDataSize * leftChildFragment.getNumNodes();
-        }
-        if (LOG.isDebugEnabled()) {
-            LOG.debug("broadcast: cost=" + Long.toString(broadcastCost));
-            LOG.debug("card=" + Long.toString(rhsTree.getCardinality()) + " row_size="
-                    + Float.toString(rhsTree.getAvgRowSize()) + " #nodes="
-                    + Integer.toString(leftChildFragment.getNumNodes()));
-        }
-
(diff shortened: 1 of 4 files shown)


morningman (approved):
LGTM

PR #6292, opened 2021-07-21
[Optimize] Make light schema change complete more faster

### LINKED ISSUE #6291: [Optimize] Make light schema change such as add column more faster
Many OLAP System like kudu, hive, add column operation is very fast while add column in doris waste much time. in our environment, add column is normal use case, and user want to see the result in short time, so I think we can refactor some code to make make light schema change more faster

caiconghui:
## Proposed changes

## Types of changes

- [ ] Other.

## Checklist

- [x] I have created an issue on (Fix #6291 ) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -87,7 +87,7 @@ CONF_Int32(clear_transaction_task_worker_count, "1");
 // the count of thread to delete
 CONF_Int32(delete_worker_count, "3");
 // the count of thread to alter table
-CONF_Int32(alter_tablet_worker_count, "3");
+CONF_Int32(alter_tablet_worker_count, "6");
 // the count of thread to clone
 CONF_Int32(clone_worker_count, "3");
 // the count of thread to clone
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -87,7 +87,7 @@ CONF_Int32(clear_transaction_task_worker_count, "1");
 // the count of thread to delete
 CONF_Int32(delete_worker_count, "3");
 // the count of thread to alter table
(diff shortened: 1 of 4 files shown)


be/src/common/config.h
```diff
@@ -87,7 +87,7 @@ CONF_Int32(clear_transaction_task_worker_count, "1");
 // the count of thread to delete
 CONF_Int32(delete_worker_count, "3");
 // the count of thread to alter table
-CONF_Int32(alter_tablet_worker_count, "3");
+CONF_Int32(alter_tablet_worker_count, "6");
```
morningman, line 90: better not changing the default value if it is not that much necessary.

morningman (commented):
I don't get it, why it get faster than before?
I notice that you put a limit number of active jobs in a new thread pool and let them run concurrently?

be/src/common/config.h
caiconghui: ok. maybe it is better decided by user's business scenario, I will restore the value to its original state

caiconghui:
> I don't get it, why it get faster than before?
> I notice that you put a limit number of active jobs in a new thread pool and let them run concurrently?

on our production environment, there are usually some tables added columns at the same time, and now these table are loaded data all time by stream load, so We hope that adding column operation can be complished as soon as possible, or stream load can be more slower when it should write two copies of data. now, it waste time to sleep and schema change jobs will affect each other, so here I change the interval for schema change check and add threadpool to make schema change jobs wouldn't affect each other.

morningman (approved):
LGTM

PR #6293, opened 2021-07-22
[Feature] Support config max length of zone map index

### LINKED ISSUE #6276: [Proposal] Support large variable-length string type
## Background

There are currently two types of strings: CHAR and VARCHAR. Char stores fixed-length strings and VARCHAR stores variable-length strings. The maximum length of VARCHAR is 65533. This length can meet most demand scenarios, but for some scenarios. In the scenario of storing larger strings in doris, it is not enough, so we need to add a new data type String. String can correspond to blob or text storage in mysql. The maximum length is 2GB, but we still don't recommend it. Store more than 64K strings in DORIS

## Other system implementation

* MYSQL: Mysql uses blob or TEXT as the storage type for very long strings. MySQL can perform string operations on these types, but performance is not guaranteed. In actual storage, the data will be stored in the overflow page. And according to the version and storage engine in the data page, the first n characters will be stored for indexing
* parquet/ORC: These two pairs and large strings are directly stored in the data area, and there is no special processing and only dictionary encoding

## Design

* Added the String type, which represents a string of any length. In order to be compatible with mysql, the maximum length is set to 2G-4, and 4 bytes are used to store the length of the string
* The data storage is similar to the varchar type, the previous length identifier is changed to 4 bytes
* Indexes are not currently supported, and zonemap indexes will be enabled after the zonemap length limit is ready.

HappenLee:
## Proposed changes

Like postgresql's Brin index, by default, records 128 rows of data once. The current Doris default page size is 64k, so a reasonable zonemap index length set 512byte.

## Types of changes

- [ ] Other.

## Checklist

- [x] I have created an issue on (Fix #6276) and described the bug/feature there in detail

### DIFF
--- a/be/src/common/configbase.cpp
+++ b/be/src/common/configbase.cpp
@@ -22,7 +22,6 @@
 #include <iostream>
 #include <list>
 #include <map>
-#include <sstream>
 
 #define __IN_CONFIGBASE_CPP__
 #include "common/config.h"
--- a/be/src/olap/field.h
+++ b/be/src/olap/field.h
@@ -66,12 +66,23 @@ public:
     inline const std::string& name() const { return _name; }
 
     virtual inline void set_to_max(char* buf) const { return _type_info->set_to_max(buf); }
+    virtual inline void set_to_zone_map_max(char* buf) const {
+        set_to_max(buf);
+    }
+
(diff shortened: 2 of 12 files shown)


morningman (commented):
Do we need to change the compare method of zonemap index?

morningman (approved):
LGTM

Issue #6307, opened 2021-07-23
[BUG]Broker can't load OSS/S3A files


huozhanfeng:
**Describe the bug**
Currently, Doris supports loading OSS/S3A files by using params like `fs.s3a.access.key`, but there is a bug when using it to load such type files.

Here is the error log when loading OSS files
<pre>
java.lang.UnsupportedOperationException: Byte-buffer read unsupported by input stream
    at org.apache.hadoop.fs.FSDataInputStream.read(FSDataInputStream.java:146)
    at org.apache.doris.broker.hdfs.FileSystemManager.readByteBuffer(FileSystemManager.java:678)
    at org.apache.doris.broker.hdfs.FileSystemManager.readByteBufferFully(FileSystemManager.java:689)
    at org.apache.doris.broker.hdfs.FileSystemManager.pread(FileSystemManager.java:570)
    at org.apache.doris.broker.hdfs.HDFSBrokerServiceImpl.pread(HDFSBrokerServiceImpl.java:157)
    at org.apache.doris.thrift.TPaloBrokerService$Processor$pread.getResult(TPaloBrokerService.java:937)
    at org.apache.doris.thrift.TPaloBrokerService$Processor$pread.getResult(TPaloBrokerService.java:922)
    at org.apache.thrift.ProcessFunction.process(ProcessFunction.java:39)
    at org.apache.thrift.TBaseProcessor.process(TBaseProcessor.java:39)
    at org.apache.thrift.server.TThreadPoolServer$WorkerProcess.run(TThreadPoolServer.java:286)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
    at java.lang.Thread.run(Thread.java:748)
</pre>
**To Reproduce**
We can simply use such load configuration to reproduce the bug
<pre>
WITH BROKER hdfs_broker (
    "fs.s3a.access.key"="...",
    "fs.s3a.secret.key"="...",
    "fs.s3a.endpoint"="..."
  )
</pre>

**Additional context**
It's a bug related to this issue.

S3A input stream to support ByteBufferReadable
https://issues.apache.org/jira/browse/HADOOP-14603

PR #6308, opened 2021-07-23
[Bug] Fix bug that broker can't load OSS/S3A files

huozhanfeng:
## Proposed changes

Currently, Doris supports loading OSS/S3A files by using params like fs.s3a.access.key, but there is a bug when using it to load such type files. The root cause is broker can not handle FSDataInputStream which does not implement ByteBufferReadable.

See Issue #6307
S3A input stream to support ByteBufferReadable
https://issues.apache.org/jira/browse/HADOOP-14603

## Types of changes
- [ ] Other.

## Checklist

### DIFF
--- a/fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
+++ b/fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
@@ -561,22 +561,25 @@ public class FileSystemManager {
                             currentStreamOffset, offset);
                 }
             }
-            ByteBuffer buf;
+            // Avoid using the ByteBuffer based read for Hadoop because some FSDataInputStream
+            // implementations are not ByteBufferReadable,
+            // See https://issues.apache.org/jira/browse/HADOOP-14603
+            byte[] buf;
             if (length > readBufferSize) {
-                buf = ByteBuffer.allocate(readBufferSize);
+                buf = new byte[readBufferSize];
             } else {
-                buf = ByteBuffer.allocate((int) length);
+                buf = new byte[(int) length];
             }
             try {
-                int readLength = readByteBufferFully(fsDataInputStream, buf);
+                int readLength = readBytesFully(fsDataInputStream, buf);
                 if (readLength < 0) {
                     throw new BrokerException(TBrokerOperationStatusCode.END_OF_FILE,
                             "end of file reached");
                 }
                 if (logger.isDebugEnabled()) {
-                    logger.debug("read buffer from input stream, buffer size:" + buf.capacity() + ", read length:" + readLength);
+                    logger.debug("read buffer from input stream, buffer size:" + buf.length + ", read length:" + readLength);
                 }
-                return buf;
+                return ByteBuffer.wrap(buf, 0, readLength);
             } catch (IOException e) {
                 logger.error("errors while read data from stream", e);
                 throw new BrokerException(TBrokerOperationStatusCode.TARGET_STORAGE_SERVICE_ERROR,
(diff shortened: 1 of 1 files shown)


fs_brokers/apache_hdfs_broker/src/main/java/org/apache/doris/broker/hdfs/FileSystemManager.java
```diff
@@ -561,22 +561,25 @@ public ByteBuffer pread(TBrokerFD fd, long offset, long length) {
...
                 }
             }
-            ByteBuffer buf;
+            // Avoid using the ByteBuffer based read for Hadoop because some FSDataInputStream
+            // implementations are not ByteBufferReadable,
+            // See https://issues.apache.org/jira/browse/HADOOP-14603
+            byte[] buf;
             if (length > readBufferSize) {
-                buf = ByteBuffer.allocate(readBufferSize);
+                buf = new byte[readBufferSize];
```
morningman, line 569: Could you help to test if the `readBufferSize` is larger than 128kb, can this `pread()` method read enough bytes that we want?
The reason why we use `bytebuffer` is that when using `byte array`, it seems that it can not read more than 128kb.

huozhanfeng: Ehh...I think `ByteBuffer` can't solve such a problem, it's only related to what size of the buffer we inited and whether the buffer can read enough bytes. In this way, the `ByteBuffer` should same as `byte array`.

I tested it with both `ByteBuffer` and `byte array` and the behavior are same when `readBufferSize` is larger than 128kb. All two of them can't read more the 128k data. Here is the debug code and part of the log.
<pre>
logger.debug("read buffer from input stream, request.length " + length + ", readBufferSize:" + readBufferSize +", buffer size:" + buf.length + ", read length:" + readLength);
                
2021-07-27 09:57:04  [ pool-2-thread-4:31261 ] - [ INFO ]  read buffer from input stream, request.length 131072, readBufferSize:1048576, buffer size:131072, read length:131072
2021-07-27 09:57:04  [ pool-2-thread-4:31268 ] - [ INFO ]  read buffer from input stream, request.length 131072, readBufferSize:1048576, buffer size:131072, read length:131072
2021-07-27 09:57:04  [ pool-2-thread-4:31273 ] - [ INFO ]  read buffer from input stream, request.length 17612, readBufferSize:1048576, buffer size:17612, read length:17612
2021-07-27 09:57:04  [ pool-2-thread-4:31275 ] - [ INFO ]  read buffer from input stream, request.length 186, readBufferSize:1048576, buffer size:186, read length:186
2021-07-27 09:57:04  [ pool-2-thread-4:31277 ] - [ INFO ]  read buffer from input stream, request.length 680, readBufferSize:1048576, buffer size:680, read length:680
</pre>

I guess the root cause is `TBrokerPReadRequest.length` in RPC request is no more than 128k which is controlled by the client. I have no BE dev env now, maybe you can help to take a look😁

huozhanfeng: I saw the default value of `readBufferSize` is 128k, maybe there is a tricky logic in Doris.
<dev>
    private int readBufferSize = 128 << 10; // 128k
    private int writeBufferSize = 128 << 10; // 128k
</dev>

huozhanfeng: I have tested a case by using ORC file format file, it can read enough bytes when `readBufferSize` is larger than 128kb. And in ORC mode, the limitation of 128k is controlled by logic as follows:
<pre>
file: be/src/exec/orc_scanner.cpp

/**
 * Get the natural size for reads.
 * @return the number of bytes that should be read at once
 */
uint64_t getNaturalReadSize() const override {
    return 128 * 1024;
}
</pre>

See https://github.com/apache/orc/blob/main/c++/src/OrcFile.cc to get more details.

morningman (approved):
LGTM

PR #6306, opened 2021-07-23
[Metrics][Load] Add metrics to monitor the size of queued tasks in load thread pool

weizuo93:
## Proposed changes

(1) Add metrics to monitor the size of queued tasks in load thread pool.
(2) Change some log level to VLOG_NOTICE

## Types of changes

What types of changes does your code introduce to Doris?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Code refactor (Modify the code structure, format the code, etc...)
- [ ] Optimization. Including functional usability improvements and performance improvements.
- [ ] Dependency. Such as changes related to third-party components.
- [ ] Other.

PR #6342, opened 2021-07-29
[Bug][RoutineLoad] Avoid TOO_MANY_TASKS error

### LINKED ISSUE #6340: [Bug][RoutineLoad] TOO_MANY_TASKS error
**Describe the bug**

Routine load job will meet TOO_MANY_TASKS errors occasionally.
And I found it is because after the transaction being committed, the task will call `kafka->commit()` to commit the offset.
But it is a blocking method, so that is kafka server has some problem, it may block for a long time.
But the task is considered as FINISHED on FE after the  transaction being committed, but the task actually not finished due 
to the blocking `commit` method.

morningman:
## Proposed changes

Use `commitAsync` to commit offset to kafka, instread of using `commitSync`, which may block for a long time.
Also assign a group.id to routine load if user not specified "property.group.id" property, so that all consumer of
this job will use same group.id instead of a random id for each consume task.

## Types of changes

- [ ] Other.

## Checklist

- [ ] I have created an issue on (Fix #6340) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/load/routineload/RoutineLoadManager.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/routineload/RoutineLoadManager.java
@@ -307,6 +307,7 @@ public class RoutineLoadManager implements Writable {
     // get the BE id with minimum running task on it
     // return -1 if no BE is available.
     // throw exception if unrecoverable errors happen.
+    // ATTN: this is only used for unit test now.
     public long getMinTaskBeId(String clusterName) throws LoadException {
         List<Long> beIdsInCluster = Catalog.getCurrentSystemInfo().getClusterBackendIds(clusterName, true);
         if (beIdsInCluster == null) {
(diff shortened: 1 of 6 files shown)


morningman:
#6306 will cause BE crash in DEBUG mode, remove this service will fix that bug.

EmmyMiao87:
Please add a issue link

fe/fe-core/src/main/java/org/apache/doris/load/routineload/RoutineLoadManager.java
```diff
@@ -344,30 +345,54 @@ public long getMinTaskBeId(String clusterName) throws LoadException {
...
             }
-            if (idleTaskNum > 0) {
-                return true;
+
+            // 2. The given BE id does not have available slots, find a BE with min tasks
+            updateBeIdToMaxConcurrentTasks();
+            int idleTaskNum = 0;
+            long resultBeId = -1L;
+            int maxIdleSlotNum = 0;
+            for (Long beId : beIdsInCluster) {
```
EmmyMiao87, line 380: This logic is consistent with manager's getMinTaskBeId function logic

morningman: i merged the 2 methods to avoid call getBeCurrentTasksNumMap twice

EmmyMiao87 (approved):
LGTM

PR #6443, opened 2021-08-14
[ResourceLimit] Limit CPU resource usage of each query

### LINKED ISSUE #6442: [ResourceLimit] Limit CPU resource usage of each query

At present, some users are executing complex queries, and the CPU or IO may be full, which affects the execution of other queries.

Users may not pay attention to the latency of these queries, but pay more attention to whether they can be successfully executed without affecting the normal use of the cluster. So we need to provide a mechanism to limit the resource overhead of a single query.

Doris currently only supports limiting the memory overhead of a single query, but does not support limiting the CPU and IO.

Achieving CPU and IO limits is a complicated matter, but we can use a simple way to achieve it according to the characteristics of Doris' query execution plan.

Doris's execution plan is a volcano model. And use the push method to get data from top to bottom, and the bottom olap scan node is usually the final node. If we can limit the resource usage of the scan node, it will indirectly limit the resource overhead of the entire query (because the data is not reached, the upper node will not work)

The scan node runs the scan task through a thread pool, and the scan task of a query will be split into several execution fragments for execution. Here we limit the number of scanning threads that can be used by a query to slow down the data scanning speed, thereby limiting the overall resource overhead.

morningman:
## Proposed changes

Details see #6442 

I add a new session variable "cpu_resource_limit" and a new user property "cpu_resource_limit".

The default value is -1, which is unlimited. And user can set limit for a single query in a session,
or set the limit for a certain user, so that all queries the user execute will use this limit.

## Types of changes

- [ ] Other.

## Checklist

- [x] I have created an issue on (Fix #6442) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/UserProperty.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/UserProperty.java
@@ -32,13 +32,13 @@ import org.apache.doris.common.io.Writable;
 import org.apache.doris.load.DppConfig;
 import org.apache.doris.system.SystemInfoService;
 
+import org.apache.commons.lang.StringUtils;
+
 import com.google.common.base.Joiner;
 import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
-
-import org.apache.commons.lang.StringUtils;
 
 import java.io.DataInput;
 import java.io.DataOutput;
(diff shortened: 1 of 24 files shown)


fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/UserProperty.java
```diff
@@ -339,6 +363,9 @@ public String getDefaultLoadCluster() {
         // sql block rules
         result.add(Lists.newArrayList(PROP_SQL_BLOCK_RULES, commonProperties.getSqlBlockRules()));
 
+        // sql block rules
```
killxdcj, line 366: The comment is wrong

gensrc/thrift/PaloInternalService.thrift
```diff
@@ -77,6 +77,10 @@ struct TLoadErrorHubInfo {
     3: optional TBrokerErrorHubInfo broker_info;
 }
 
+struct TResourceLimit {
+    1: optional i32 cpu_limit
```
killxdcj, line 81: Maybe It is better to use a name such as scannode_max_parallelism xxx

morningman: I want to decouple the struct with the certain implementation.
So that we can extend this feature in future.

wuyunfeng: maybe use max_concurrency is better

morningman: > maybe use max_concurrency is better

In fact, what I want to express is that this value is not bound to any implementation. It does not represent the number of threads or the number of concurrency, but a relative value.
It is more like a priority concept, or similar to the concept of weight in cgroups.
In this way, we can continue to use this parameter when we implement more refined resource isolation in the future.

yangzhg (approved):
LGTM

PR #6490, opened 2021-08-23
Flink reads multiple data sources to doris

caoliang-web:
Flink Demo example:
1.Flink consumes Kafka data and uses Stream load to store data (dorisSink method)
2.Flink reads the file into the library (hdfs file, the format can be text or other), and uses stream load to import
3.Use flink connector to read doris data to build DataStream for sum analysis


hf200012 (dismissed):
LGTM

morningman (approved):
LGTM

PR #6501, opened 2021-08-25
add config to support long table name 

ikaruga4600:
## Proposed changes

Support length of table name large than 64

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)
- [ ] Other.

## Checklist

- [x ] Compiling and unit tests pass locally with my changes


morningman (commented):
The max table name length of MySQL is 64, If we use a longer name, I am not sure
if some of BI tools can be compatible with this.

morningman:
> > The max table name length of MySQL is 64, If we use a longer name, I am not sure
> > if some of BI tools can be compatible with this.
> 
> Thanks for your reply ! because we need to copy some tables from hive, and in hive max size is 128.
> may better let user change this value ?

Sounds reasonable

morningman (approved):
LGTM

PR #6513, opened 2021-08-26
[Optimize][Clone] Take version count into consideration when choosing src replica for clone task

### LINKED ISSUE #6512: [Optimize][Clone] Optimize clone task replica selector
If there is  missing replica for a tablet, clone task will be executed to restore missing replica from a healthy replica. Src replica selector will randomly choose a healthy replica as src replica.

 It's better to choose the health replica with minmun version count as src replica so that it could avoid repetitive compaction task. In addition, replica with less version count is good for query performance.

weizuo93:
## Proposed changes

Fix #6512 

If there is missing replica for a tablet, clone task will be executed to restore missing replica from a healthy replica. Src replica selector will randomly choose a healthy replica as src replica.

It's better to choose the health replica with minmun version count as src replica so that it could avoid repetitive compaction task. In addition, replica with less version count is good for query performance.

## Types of changes

- [ ] Other.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/clone/TabletSchedCtx.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/clone/TabletSchedCtx.java
@@ -490,7 +490,7 @@ public class TabletSchedCtx implements Comparable<TabletSchedCtx> {
             throw new SchedException(Status.UNRECOVERABLE, "unable to find source replica");
         }
 
-        long minVersionCount = -1;
+        long minVersionCount = -2;
         boolean findSrcReplica = false;
         // choose a replica which slot is available from candidates.
         for (Replica srcReplica : candidates) {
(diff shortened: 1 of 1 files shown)


morningman (approved):
LGTM

fe/fe-core/src/main/java/org/apache/doris/clone/TabletSchedCtx.java
```diff
@@ -499,10 +501,26 @@ public void chooseSrcReplica(Map<Long, PathSlot> backendsWorkingSlots) throws Sc
...
+                    if (curVerCount < minVersionCount) {
+                        minVersionCount = curVerCount;
+                        setSrc(srcReplica);
+                        findSrcReplica = true;
+                    }
+                }
             }
         }
+
+        if (findSrcReplica) {
```
caiconghui, line 521:
```suggestion
        if (!findSrcReplica) {
             throw new SchedException(Status.SCHEDULE_FAILED, "unable to find source slot");
        }

       
```

fe/fe-core/src/main/java/org/apache/doris/clone/TabletSchedCtx.java
```diff
@@ -499,10 +501,26 @@ public void chooseSrcReplica(Map<Long, PathSlot> backendsWorkingSlots) throws Sc
             
             long srcPathHash = slot.takeSlot(srcReplica.getPathHash());
             if (srcPathHash != -1) {
-                setSrc(srcReplica);
-                return;
+                if (!findSrcReplica) {
```
caiconghui, line 504: remove the first if block

fe/fe-core/src/main/java/org/apache/doris/clone/TabletSchedCtx.java
weizuo93: It's necessary to record first candidate replica in the `if` block.

fe/fe-core/src/main/java/org/apache/doris/clone/TabletSchedCtx.java
weizuo93: It's necessary to record first candidate replica in the `if` block.

PR #6515, opened 2021-08-26
[Memory] BitShufflePageDecoder use memory allocated by ChunkAllocator instead of Faststring

### LINKED ISSUE #6285: [Proposal] Low performance when BitShufflePageDecoder uses faststring store decoded data
BitShufflePageDecoder uses faststring as a continuous memory to store decoded data, which has low performance at high concurrency.

## Motivation

In a user test, a simple `select *` query is slow, and analysis of the reason found that `~Reader()` takes a long time to destruct.

After preliminary analysis, it is found that the time-consuming `tcmalloc::CentralFreeList::Release` in `faststring` is the release of the memory requested from tcmalloc. `faststring` is used in `BitShufflePageDecoder` to save the data of `decompress_lz4` decompress.

Consider allowing `faststring` to apply for space directly from `mempool` to avoid the time-consuming tcmalloc allocate and release.

## Simple verification
Test environment: 1 FE, 3 BE
Test table: 346 columns, 3000w rows, column types include char, int, decimal, varchar (same as the above user test)
Test SQL: select * from test_table limit 10;
(Note: The problem cannot be reproduced on a table with a small number of columns. The reason has not been investigated...)
Doris Version: 

jmeter stress testing:
```
<stringProp name="ThreadGroup.num_threads">10</stringProp>
<stringProp name="ThreadGroup.ramp_time">1</stringProp>
<boolProp name="ThreadGroup.scheduler">true</boolProp>
<stringProp name="ThreadGroup.duration">100</stringProp>
```

trunk branch average time: 1193 (2021.10.08)
image
image

xinyiZzz:
## Proposed changes

BitShufflePageDecoder reuses the memory for storing decoder results, allocate memory directly from the `ChunkAllocator`, the performance is improved to a certain extent.

In the case of #6285, the total time consumption is reduced by 13.5%, and the time consumption ratio of `~Reader()` has also been reduced from 17.65% to 1.53%, and the memory allocation is unified to `ChunkAllocator` for centralized management , Which is conducive to subsequent memory optimization.

which can avoid the memory waste caused by `Mempool`, because the chunk can be free at any time, but the performance is lower than the allocation from `Mempool`. The guess is that there is no `Mempool` after secondary allocation of large chunks , Will directly apply for a large number of small chunks from `ChunkAllocator`, and it takes longer to lock in `pop_free_chunk` and `push_free_chunk` (but this is not proven from the flame graphs of BE's cpu and contention).

Average time: 1032s (jmeter test, see issue #6285 for environment and configuration,  2021.10.08)
image

### 1. Memory allocated by Mempool instead of Faststring.

Average time: 971s
image

But this will cause the memory requested in BitShufflePageDecoder to be released uniformly when FileColumnIterator is destructed, resulting in a waste of memory.

### 2. Reuse the Faststring object

Create a faststring object in FileColumnIterator, and then pass the pointer to each page BitShufflePageDecoder. The length of each decode in BitShufflePageDecoder is variable. If the length is insufficient during faststring resize, the capacity will be expanded, otherwise it will be reused directly. 

In theory, the results of all page decodes in a FileColumnIterator are roughly the same, so there should be no memory waste problem. However, the performance of reusing faststring is lower than using mempool to allocate cached memory. 

image
image

### 3. Modify Faststring to allocate space from Mempool

Currently Faststring allocates space through `new`

faststring_test_all_no_leng
faststring_test_faststring_noleng

## Types of changes

- [ ] Other.

## Checklist

- [x] I have created an issue on (Fix #6285) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_v2/bitshuffle_page.h
+++ b/be/src/olap/rowset/segment_v2/bitshuffle_page.h
@@ -33,6 +33,7 @@
 #include "olap/rowset/segment_v2/page_builder.h"
 #include "olap/rowset/segment_v2/page_decoder.h"
 #include "olap/types.h"
+#include "runtime/memory/chunk_allocator.h"
 #include "util/coding.h"
 #include "util/faststring.h"
 #include "util/slice.h"
@@ -214,6 +215,12 @@ public:
               _size_of_element(0),
               _cur_index(0) {}
 
+    ~BitShufflePageDecoder() {
+        if (_chunk.size != 0) {
+            ChunkAllocator::instance()->free(_chunk);
+        }
+    }
+
     Status init() override {
         CHECK(!_parsed);
         if (_data.size < BITSHUFFLE_PAGE_HEADER_SIZE) {
@@ -302,7 +309,7 @@ public:
         // - left == _num_elements when not found (all values < target)
         while (left < right) {
             size_t mid = left + (right - left) / 2;
-            mid_value = &_decoded[mid * SIZE_OF_TYPE];
+            mid_value = &_chunk.data[mid * SIZE_OF_TYPE];
             if (TypeTraits<Type>::cmp(mid_value, value) < 0) {
                 left = mid + 1;
             } else {
@@ -312,7 +319,7 @@ public:
         if (left >= _num_elements) {
             return Status::NotFound("all value small than the value");
         }
-        void* find_value = &_decoded[left * SIZE_OF_TYPE];
+        void* find_value = &_chunk.data[left * SIZE_OF_TYPE];
         if (TypeTraits<Type>::cmp(find_value, value) == 0) {
             *exact_match = true;
         } else {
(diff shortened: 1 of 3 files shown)


morningman (commented):
1. We should implement a `buffer` class to encapsulate the underlying memory allocation logic, instead of directly calling the pool's allocate method.

2. The life cycle of the buffer in PageDecoder is the same as that of PageDecoder, no need to follow ColumnReader

be/src/olap/rowset/segment_v2/bitshuffle_page.h
```diff
@@ -385,7 +394,7 @@ class BitShufflePageDecoder : public PageDecoder {
 
     int _size_of_element;
     size_t _cur_index;
-    faststring _decoded;
+    Chunk chunk;
```
HappenLee, line 397: '_chunk` not ‘chunk’

xinyiZzz: done

HappenLee:
LGTM

HappenLee (dismissed):

HappenLee (approved):
LGTM

morningman (approved):
LGTM

PR #6577, opened 2021-09-06
Custom config handler #6508

### LINKED ISSUE #6508: [Feature] Custom config handler, support callback when set mutable config at FE and BE
We need to custom query traffic limit rules in FE, but config can only be basic type and no callback. We need to compile the limit rules when set new limit rules config, and calculate all the rules at each query.

Add a handler field at `ConfField` named `handler` to specify the class to use as callback function.
Call the handler when setting new config value. The default handler is the same as current.
Support all types config.

ccoffline:
## Proposed changes

Support custom config handler callback and types.

## Types of changes

- [ ] Other.

## Checklist

- [x] I have created an issue on (Fix #6508) and described the bug/feature there in detail

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/common/ConfigBase.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/ConfigBase.java
@@ -17,6 +17,7 @@
 
 package org.apache.doris.common;
 
+import org.apache.doris.catalog.Catalog;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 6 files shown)


morningman:
Could your explain more about `query traffic limit` in your issue #6508 ?

ccoffline:
> Could your explain more about `query traffic limit` in your issue #6508 ?

I didn't put `query traffic limit` in this pr. This is still experimental on one of our clusters.

Some users want to limit query more flexibly than `maxQueryInstances` #6159. For example, all query should be rejected when queryInstances of this user reach 30000, and big queries that queryInstances more than 1000 should be rejected when queryInstances of this user reach 20000. Limiting rules like this will be complex and need to be modified at any time.

So we add a config that support this feature, using `SpEL` to evaluate whether to limit a query or not. The limiting expressions will be reset at any time, and the `Expression` needed to be compile as soon as setting the config. This is what this pr for, I think more configs will need callback in the future.

We are using 0.13 and #6159 is not supported yet. In master, there is a similar feature `sql block rule` #6192 which can support `query traffic limit`, maybe add a `limit` property. So I won't push our traffic limit pr to master, but just add config handler to support config setting callback.

ccoffline:
I just looked into #6192 and `sql block rule` cannot support traffic limit, but I think traffic limit is better set in the user properties.

morningman (approved):
LGTM

PR #6644, opened 2021-09-14
Support hdfs in select outfile clause

EmmyMiao87:
## Proposed changes

Support hdfs in select outfile clause without broker.
This PR implement a HDFS writer in BE which is used to write HDFS file directly without using broker.
Also the hdfs outfile  clause syntax check has been added in FE.
The syntax:
```
select * from xx into outfile "hdfs://user/outfile_" format as csv properties ("hdfs.fs.dafultFS" = "xxx", "hdfs.hdfs_user" = "xxx");
```
Note that all hdfs configurations need to carry a prefix `hdfs.`.

## Types of changes

- [ ] Other.

## Checklist

### DIFF
--- a/be/src/runtime/file_result_writer.cpp
+++ b/be/src/runtime/file_result_writer.cpp
@@ -21,6 +21,7 @@
 #include "exec/local_file_writer.h"
 #include "exec/parquet_writer.h"
 #include "exec/s3_writer.h"
+#include "exec/hdfs_writer.h"
 #include "exprs/expr.h"
 #include "exprs/expr_context.h"
 #include "gen_cpp/PaloInternalService_types.h"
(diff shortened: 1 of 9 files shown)


be/src/runtime/file_result_writer.cpp
```diff
@@ -131,12 +132,14 @@ Status FileResultWriter::_create_next_file_writer() {
...
         _file_writer = new LocalFileWriter(file_name, 0 /* start offset */);
-    } else if (_storage_type == TStorageBackendType::BROKER){
+    } else if (_storage_type == TStorageBackendType::BROKER) {
         _file_writer =
                 new BrokerWriter(_state->exec_env(), _file_opts->broker_addresses,
                                  _file_opts->broker_properties, file_name, 0 /*start offset*/);
     } else if (_storage_type == TStorageBackendType::S3) {
         _file_writer =  new S3Writer(_file_opts->broker_properties, file_name, 0 /* offset */);
+    } else if (_storage_type == TStorageBackendType::HDFS) {
+        _file_writer = new HDFSWriter(const_cast<std::map<std::string, std::string>&>(_file_opts->broker_properties), file_name);
```
morningman, line 142: Why need a `const_cast` here but `S3Writer` don't?

be/src/runtime/result_file_sink.cpp
```diff
@@ -119,6 +116,9 @@ Status ResultFileSink::prepare(RuntimeState* state) {
 
     }
     RETURN_IF_ERROR(_writer->init(state));
+    for (int i = 0; i < _channels.size(); ++i) {
```
morningman, line 119: So this is a bug?

be/src/runtime/file_result_writer.cpp
EmmyMiao87: The HDFS writer needs to parse properties so here is a ```const_cast```

be/src/runtime/result_file_sink.cpp
EmmyMiao87: Yes, if the _writer failed to init, the channel should not be inited before closed. Otherwise, the be will core in destructor

morningman (approved):
LGTM

PR #6702, opened 2021-09-18
[Optimize] Use compact mode to send query plan thrift data structure.

morningman:
## Proposed changes

In some cases, the query plan thrift structure of a query may be very large
(for example, when there are many columns in SQL), resulting in a large number
of "send fragment timeout" errors.

This PR adds an FE config  `use_compact_thrift_rpc` to control whether to transmit
the query plan in a compressed format.

Using compressed format transmission can reduce the size by ~50%. But it may reduce
the concurrency by ~10%. Therefore, in the high concurrency small query scenario,
you can choose to turn off compaction.

## Types of changes

- [ ] Other.

## Checklist

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/rpc/BackendServiceProxy.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/rpc/BackendServiceProxy.java
@@ -37,34 +37,35 @@ import org.apache.thrift.protocol.TCompactProtocol;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Future;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.concurrent.locks.ReentrantLock;
 
 public class BackendServiceProxy {
     private static final Logger LOG = LogManager.getLogger(BackendServiceProxy.class);
-    private static volatile BackendServiceProxy INSTANCE;
+    // use exclusive lock to make sure only one thread can add or remove client from serviceMap.
+    // use concurrent map to allow access serviceMap in multi thread.
+    private ReentrantLock lock = new ReentrantLock();
     private final Map<TNetworkAddress, BackendServiceClient> serviceMap;
-    private ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
 
     public BackendServiceProxy() {
-        serviceMap = Maps.newHashMap();
+        serviceMap = Maps.newConcurrentMap();
+    }
+
+    private static class SingletonHolder {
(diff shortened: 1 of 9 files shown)


caiconghui:
should we turn off or turn on use_compact_thrift_rpc  by adaptive mode， so we can get both query concurrent performance and avoid send fragment timeout

caiconghui:
> > should we turn off or turn on use_compact_thrift_rpc by adaptive mode， so we can get both query concurrent performance and avoid send fragment timeout
> 
> That may be a good choice. But I have no idea how to make it adaptive to the runtime situation...

according to the request size？

morningman:
> according to the request size？

The key point is how to determine the threshold of the size...

caiconghui:
> > according to the request size？
> 
> The key point is how to determine the threshold of the size...

I think that it is better to use  threshold of the size than to use use_compact_thrift_rpc as a config.

morningman:
> > > according to the request size？
> > 
> > 
> > The key point is how to determine the threshold of the size...
> 
> I think that it is better to use threshold of the size than to use use_compact_thrift_rpc as a config.

You mean use size as config and using a big or small value as default?

caiconghui:
> > > > according to the request size？
> > > 
> > > 
> > > The key point is how to determine the threshold of the size...
> > 
> > 
> > I think that it is better to use threshold of the size than to use use_compact_thrift_rpc as a config.
> 
> You mean use size as config and using a big or small value as default?

yeah, because the request size is different with each other,  it is not flexible for users to just simply turn on or turn off the use_compact_thrift_rpc config

fe/fe-core/src/main/java/org/apache/doris/rpc/BackendServiceProxy.java
```diff
@@ -59,23 +61,19 @@ public static BackendServiceProxy getInstance() {
...
 
     public void removeProxy(TNetworkAddress address) {
-        lock.writeLock().lock();
-        try {
-            serviceMap.remove(address);
-        } finally {
-            lock.writeLock().unlock();
+        LOG.warn("begin to remove proxy: {}", address);
+        BackendServiceClient service;
+        synchronized (this) {
```
caiconghui, line 66: why use synchronized here and use lock for getProxy? how about use ConcurrentHashMap and use computeIfAbsent method

morningman: fixed.
but `computeIfAbsent` is not suitable because we need to make sure that only one thread can create client for a certain BE.

caiconghui (approved):
LGTM

PR #6736, opened 2021-09-24
[Bug] When using view, make toSql method generates the final sql

### LINKED ISSUE #6735: [Bug] Cache cannot sense the modification of the view
### Search before asking

### Version

rebase on branch-0.13

### What's Wrong?

Modifying the `VIEW` is equivalent to modifying the SQL, but because the SQL string has not changed, the query will hit the cached result before the `VIEW` is modified.

PR: [6687](https://github.com/apache/incubator-doris/pull/6687), by adding `updateTime` to `VIEW` and using `originSql + viewUpdateTime` as the cache key, this problem can be solved, but it is not very good.

### What You Expected?

> Note: There are two types of views, one is the view created by `create view ...`, which will be saved in the FE metadata. The other is the use of the `WITH` statement in SQL. After the SQL is parsed, a `localView` will be generated, which will not be saved in the metadata.

1. For `localView`, it does not need to be expanded into a statement, because `toSql` method already contains `WITH` statement. (However, there should be a bug in the with content display. When a `UNION` is encountered, the with statement is not displayed. The PR has been fixed)
2. For normal VIEW, it needs to be converted into a final statement
3. For the normal VIEW contained in `localView`, the normal VIEW will be expanded in the WITH statement

### How to Reproduce?

Omit

### Anything Else?

_No response_

### Code of Conduct

zenoyang:
Fix #6735 

## Proposed changes

1. Fix the problem that the WITH statement cannot be printed when `UNION` is included in SQL
2. In the `toSql` method, convert the normal VIEW into the final statement
3. Replace `selectStmt.originSql` with `selectStmt.toSql`

## Types of changes

- [ ] Other.

## Checklist

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/cache/SqlCache.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/cache/SqlCache.java
@@ -41,7 +41,7 @@ public class SqlCache extends Cache {
 
     public InternalService.PFetchCacheResult getCacheData(Status status) {
         InternalService.PFetchCacheRequest request = InternalService.PFetchCacheRequest.newBuilder()
-                .setSqlKey(CacheProxy.getMd5(selectStmt.getOrigStmt().originStmt))
+                .setSqlKey(CacheProxy.getMd5(selectStmt.toSql()))
                 .addParams(InternalService.PCacheParam.newBuilder()
                         .setPartitionKey(latestTable.latestPartitionId)
                         .setLastVersion(latestTable.latestVersion)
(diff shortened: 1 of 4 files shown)


morningman (commented):
Please add UT for this case

zenoyang:
> Please add UT for this case

Done. Please review again, thank you 😀

fe/fe-core/src/main/java/org/apache/doris/qe/cache/SqlCache.java
```diff
@@ -41,7 +41,7 @@ public void setCacheInfo(CacheAnalyzer.CacheTable latestTable) {
 
     public InternalService.PFetchCacheResult getCacheData(Status status) {
         InternalService.PFetchCacheRequest request = InternalService.PFetchCacheRequest.newBuilder()
-                .setSqlKey(CacheProxy.getMd5(selectStmt.getOrigStmt().originStmt))
+                .setSqlKey(CacheProxy.getMd5(selectStmt.toSql()))
```
morningman, line 44: Why use `toSql()` can solve the issue #6735?

xinghuayu007:
This solution is elegant！

morningman (approved):
lgtm

PR #6748, opened 2021-09-26
Revert "[Enhancement] Modify the method of calculating compaction score"

morningman:
Reverts apache/incubator-doris#6252

This commit may cause tablet which segments are all empty never to compaction, and results in -235 error.
I will revert this commit, and the problem will be solved in #6671


morningman:
cc @weizuo93 , please review.

weizuo93 (approved):

PR #6793, opened 2021-10-04
[Revert] "[Bug] When using view, make toSql method generates the final sql (#6736)"

### LINKED ISSUE #6792: [Bug] Failed query a logic view which based on another view.
### Search before asking

### Version

trunk 5f3559a

### What's Wrong?

```
CREATE TABLE `tbl1` (
  `k1` int(11) NULL COMMENT "",
  `k2` bigint(20) NULL COMMENT ""
) ENGINE=OLAP
DUPLICATE KEY(`k1`, `k2`)
COMMENT "OLAP"
PROPERTIES (
"in_memory" = "false",
"storage_format" = "V2"
);

create view v1 as select * from tbl1;
create view v2 as select * from v2;

select * from v2;
ERROR 1248 (42000): errCode = 2, detailMessage = Every derived table must have its own alias
```

### What You Expected?

Working well

### How to Reproduce?

as above

### Anything Else?

This bug is introduced from #6736
This is because when creating a view based on another view, the inner view is expanded (after modification of #6736).

For `select * from v2`,
After #6736, it will be expanded to :
`select * from (select * from (select * from tbl1))`
But the subquery does not have alias, so the error returns.

But even if we add alias to the subquery, such as:

`select * from (select * from (select * from tbl1) v1) v2`

it will still return error: `Unknown column 'k1' in 'test.nv1'`

This is because the query is expanded to this:
```
SELECT `db1`.`v1`.`k1` AS `k1`, `db1`.`v1`.`k2` AS `k2` FROM (SELECT `default_cluster:db1`.`tbl1`.`k1` AS `k1`, `default_cluster:db1`.`tbl1`.`k2` AS `k2` FROM `default_cluster:db1`.`tbl1`) v1;
```

Notice that the qualifier of `k1` is `db1.v1`, not `v1`. But there is no `db1.v1` in this query.

I have no idea how to fix this bug. Maybe I have to revert the #6736 , and find another way to solve the problem.

### Code of Conduct

morningman:
## Proposed changes

This reverts part of commit 11ec38dd6fd9f86632d83c47bd9d8bc05db69a2b(#6736)
Because it will cause view query problem described in #6792 

The following bug fix remained:
1. Fix the problem that the WITH statement cannot be printed when UNION is included in SQL

## Types of changes

- [ ] Other.

## Checklist

- [ ] I have created an issue on (Fix #6792) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/conf/be.conf
+++ b/conf/be.conf
@@ -22,7 +22,6 @@ sys_log_level = INFO
 
 # ports for admin, web, heartbeat service 
 be_port = 9060
-be_rpc_port = 9070
 webserver_port = 8040
 heartbeat_service_port = 9050
 brpc_port = 8060
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/InlineViewRef.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/InlineViewRef.java
@@ -423,23 +423,17 @@ public class InlineViewRef extends TableRef {
         // This is needed for view compatibility between Impala and Hive.
         String aliasSql = null;
         String alias = getExplicitAlias();
-        if (alias != null) aliasSql = ToSqlUtils.getIdentSql(alias);
-
-        StringBuilder sb = new StringBuilder();
-        if (view != null) {
-            // When it is a local view, it does not need to be expanded into a statement,
-            // because toSql function already contains `with` statement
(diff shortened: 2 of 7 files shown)


yangzhg (approved):
LGTM

PR #6829, opened 2021-10-13
[Memory Usage] Implement segment lru cache to save memory of BE

morningman:
## Proposed changes

This CL mainly changes:

1. Use a lru cache to cache all opened segments(Only for beta rowset, alpha rowset remain unchanged)

    In the previous implementation, the segment instance is an object belonging to the rowset,
    so if the rowset exists, the segment will always exist, resulting in memory usage.
    In fact, the segment instance is only used in the data reading process. In principle, the segment
    can be closed after the data reading is completed.

    So here I use a lru cache to cache all open segments, and use SegmentLoader to uniformly manage
    segment loading and closing.

    Before:
    image

    After using cache, the memory will no longer increase continously:
    image

    In addition to the cached segments being eliminated by the lru algorithm, there will also be threads
    that regularly clean up segments that have not been used for a period of time.

    Also, I modify the lru cache implementation, to add a `type` for the lru cache.
    The type is either SIZE or NUMBER. SIZE is the default type, and NUMBER means the capacity of cache
    is based on the number of cached entries, not the size of cached entries.

    This is because currently we can not precisely calculate the memory usage of each opened segments.
    So we have to use "number" to limit the max cached segments. But if later we get size, better to use SIZE.

2. Optimize the usage statistics of lru cache

    In the previous implementation, the usage of lru cache is only based on the size of cache value.
    But for some cases, the cache key is larger than the value. So it is better to use both key and value
    size to calculate the usage of lru.

2. Let BE report immediately the first time it receive heartbeat from FE

    If not do so, even if the backend is alive with success heartbeat, user can not create table because the
    disk info is not reported yet.
    And also add a random sleep for the report process to avoid too many reports proceed at same time.

## Types of changes

- [ ] Other.

## Checklist

- [ ] I have created an issue on (Fix #6828 ) and described the bug/feature there in detail

## Further comments


xinyiZzz:
`segment_cache_capacity` only limits the number of Rowsets, the memory occupied by Segment Cache is uncontrollable.

After that, we can consider improving the statistics of _mem_tracker in the Segment to indicate the actual memory occupied by each Segment, thereby limiting the size of the Segment Cache.

xinyiZzz:
The `prune_if` method of LRU cache is good!, I understand that other caches can also be cleaned regularly, using the same strategy.

yangzhg (approved):
LGTM

PR #6861, opened 2021-10-17
[Optimize][Runtime Filter] Init left node asynchronously in join node

Gabriel39:
## Proposed changes

Currently, when runtime filters is specified, left child node will be initialed after building process is done in HashJoinNode. Obviously it will hurt performance and it should be changed to asynchronous mode.

## Types of changes

- [ ] Other.

## Checklist

## Further comments

### DIFF
--- a/be/src/exec/hash_join_node.cpp
+++ b/be/src/exec/hash_join_node.cpp
@@ -183,6 +183,12 @@ void HashJoinNode::build_side_thread(RuntimeState* state, boost::promise<Status>
     status->set_value(construct_hash_table(state));
 }
 
+
+void HashJoinNode::probe_side_initial_thread(RuntimeState* state, ExecNode* child,
+                                             boost::promise<Status>* status) {
+    status->set_value(child->open(state));
+}
+
 Status HashJoinNode::construct_hash_table(RuntimeState* state) {
     // Do a full scan of child(1) and store everything in _hash_tbl
     // The hash join node needs to keep in memory all build tuples, including the tuple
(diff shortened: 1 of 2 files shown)


Gabriel39:
cc @stdpain @morningman

EmmyMiao87:
This is actually a strategic trade-off. Move to execute left node or wait for RT and push to the storage layer. It depends on the filtering rate of RT at the storage layer and the possible waiting time for synchronous execution.

Gabriel39:
> @Gabriel39 Thank you for your contribute! If the open phase is completed before the RF is generated, it may result in the generated expressions not being pushed down to the storage layer. It looks like this could lead to worse performance. We first need to provide runtime expressions to push down to the storage layer.

Hi @stdpain @EmmyMiao87  , I got your point but one more thing I m not sure is whether open phase of scan node will be completed before RF is generated. For example, open() in `OlapScanNode` will be blocked until all RFs are ready. So it seems like scan node will execute with RFs without performance degradation.

stdpain:
The reason why we have to block to wait for the arrival of the filter is that there is no ability to dynamically push down predicates on the scan. This is the first thing we need to address, rather than forcing down-push expressions.

The ability to index may not be available without pushing down expressions. Also the segment layer will use the zone map index when init is called for the first time, so this is the first thing we do.

If the olap scan node does not wait for any runtime filter, then there seems to be no benefit in the current implementation. If it does not wait for a runtime filter to arrive it means that the query has a higher probability of being a big query. The overhead of waiting at this time is not significant compared to the overall overhead.

So we need to provide the ability to normailize predicates on the runtime filter. This way there is no need to wait for the filter to be sent

Translated with www.DeepL.com/Translator (free version)

Gabriel39:
> The reason why we have to block to wait for the arrival of the filter is that there is no ability to dynamically push down predicates on the scan. This is the first thing we need to address, rather than forcing down-push expressions.
> 
> The ability to index may not be available without pushing down expressions. Also the segment layer will use the zone map index when init is called for the first time, so this is the first thing we do.
> 
> If the olap scan node does not wait for any runtime filter, then there seems to be no benefit in the current implementation. If it does not wait for a runtime filter to arrive it means that the query has a higher probability of being a big query. The overhead of waiting at this time is not significant compared to the overall overhead.
> 
> So we need to provide the ability to normailize predicates on the runtime filter. This way there is no need to wait for the filter to be sent

Hi @stdpain I think you misunderstood me. 

In this PR, scan node will still execute only after the runtime filter is ready, so in essence, it just changes the place of the blocking for RF from join node to scan node.

That is to say, scan node does not change its behavior after this PR. It still step by step to init , wait for RF arrival, and execute.

Gabriel39:
still any questions about this PR? @stdpain @EmmyMiao87

morningman:
> still any questions about this PR? @stdpain @EmmyMiao87

Yes, there are something need to be resolved. You can talk to @HappenLee @BiteTheDDDDt

PR #6964, opened 2021-10-28
Prevent invalid expr continue parse

zbtzbtzbt:
## Proposed changes

In some cases, when expr == null, continue parse statement may be visit invalid memory, which will casuse BE process failure.

## Types of changes

- [ ] Other.

## Checklist

## Further comments

### DIFF
--- a/be/src/exprs/expr.cpp
+++ b/be/src/exprs/expr.cpp
@@ -283,6 +283,9 @@ Status Expr::create_tree_from_thrift(ObjectPool* pool, const std::vector<TExprNo
     Expr* expr = NULL;
     RETURN_IF_ERROR(create_expr(pool, nodes[*node_idx], &expr));
     DCHECK(expr != NULL);
+    if (expr == NULL) {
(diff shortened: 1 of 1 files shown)


be/src/exprs/expr.cpp
```diff
@@ -283,6 +283,9 @@ Status Expr::create_tree_from_thrift(ObjectPool* pool, const std::vector<TExprNo
     Expr* expr = NULL;
     RETURN_IF_ERROR(create_expr(pool, nodes[*node_idx], &expr));
     DCHECK(expr != NULL);
```
morningman, line 285: Here is the DCHECK here. So I think if expr is null here, this is a fatal error?
Could you show me the case which can result in expr as null there?

zbtzbtzbt: some bad cases which lead to Doris-BE process fail in meituan are fixed by [https://github.com/apache/incubator-doris/pull/5799](https://github.com/apache/incubator-doris/pull/5799).
so this pr can be closed.

PR #6965, opened 2021-10-28
optimize for single rowset reading

GoGoWen:
## Proposed changes
when read the tablet with all columns for single rowset which !has_overlapping && nonoverlapping_count == 1 && !has_delete_rowset,  we should use _direct_agg_key_next_row, which can reduce useless aggregation when reading from rowset,   however  if we read less columns(most olap query should be),  use _agg_key_next_row should be better,  which  will do aggregation before return to next caller.  

## Types of changes

- [X] Optimization. Including functional usability improvements and performance improvements.
- [ ] Other.

## Checklist

## Further comments


be/src/olap/reader.cpp
```diff
@@ -130,10 +130,16 @@ OLAPStatus Reader::init(const ReaderParams& read_params) {
...
-
+    // optimize for single rowset reading without do aggregation when reading all columns, 
+    // and otherwise should use _agg_key_next_row for AGG_KEYS
     if (_optimize_for_single_rowset(rs_readers)) {
-        _next_row_func = _tablet->keys_type() == AGG_KEYS ? &Reader::_direct_agg_key_next_row
-                                                          : &Reader::_direct_next_row;
+        if(_tablet->keys_type() == AGG_KEYS && _return_columns.size() == _tablet->tablet_schema().num_columns()) {
+            _next_row_func = &Reader::_direct_agg_key_next_row;
+        } else if (_tablet->keys_type() == AGG_KEYS) {
+            _next_row_func = &Reader::_agg_key_next_row;
```
morningman, line 139: I didn't get the point.
if `_optimize_for_single_rowset()` is true, there should be only one rowset, why we need to call `agg`?

GoGoWen: For example,  the table test (k int ,k1 int ,k2 int ,k3 int, v int sum),  if we read all columns (k,k1,k2,k3,v),  then it definitely no need to call agg.  However,  if we just need (k,k1,v),  then if we do not call agg, there are many  many duplicate-key rows for next caller(may be aggregation node),   it's not a good idea to pass these rows without aggregation to aggregation node for a query.

morningman (dismissed):
LGTM  @GoGoWen , please rebase to solve the conflict

Issue #6969, opened 2021-10-29
[Feature] Add stream load info to system info page of web site


tianhui5:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Description

In my circumstance, the Doris cluster is usually facing pressure from massive Stream Load. I think it should be more convient if there is Stream Load info in system info of web site. 
Maybe I will develop a flow restraint plugin based on this feature some days later.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #6970, opened 2021-10-29
Add stream load info to system info of web site

MorningLight5:
## Proposed changes
#6969 

## Types of changes

- [ ] Other.

## Checklist


fe/fe-core/src/main/java/org/apache/doris/common/proc/ProcService.java
```diff
@@ -38,7 +38,7 @@ private ProcService() {
         root.register("backends", new BackendsProcDir(Catalog.getCurrentSystemInfo()));
         root.register("dbs", new DbsProcDir(Catalog.getCurrentCatalog()));
         root.register("jobs", new JobsDbProcDir(Catalog.getCurrentCatalog()));
-        root.register("statistic", new StatisticProcDir(Catalog.getCurrentCatalog()));
```
morningman, line 41: better not to modify the default value. Some online system rely on it. We should make it forward-compatible, in most cases.

tianhui5: That makes sense, I'll change it back right now.

morningman (approved):
LGTM

PR #6971, opened 2021-10-29
hll optimize: trace memory usage, new explicit data when really need

zuochunwei:
reduce hll memory occupied：

replace uint64_t _explicit_data[160*2] with uint64_t*
new memory for explicit data when really needed
trace HLL memory usage


morningman:
Hi @zuochunwei Could you please create an issue to describe more detail about how this works and how much memory can be reduced after this PR?

zuochunwei:
> Hi @zuochunwei Could you please create an issue to describe more detail about how this works and how much memory can be reduced after this PR?

each HLL object will reduce nearly 2k bytes，because 160*2*sizeof(uint64_t) result 2560，almost 2k bytes.
so when use tool to import data，if the table have hundreds of millions records,  we generate one hll object for each record,  the memory usage will increase hundreds Gigabytes

yangzhg (dismissed):
LGTM

be/src/olap/hll.h
```diff
@@ -33,6 +33,7 @@ class Slice;
 const static int HLL_COLUMN_PRECISION = 14;
 const static int HLL_ZERO_COUNT_BITS = (64 - HLL_COLUMN_PRECISION);
 const static int HLL_EXPLICIT_INT64_NUM = 160;
+const static int HLL_EXPLICIT_INT64_NUM_DOUBLE = HLL_EXPLICIT_INT64_NUM*2;
```
HappenLee, line 36: HLL_EXPLICIT_INT64_NUM * 2

be/src/olap/hll.h
```diff
@@ -154,7 +174,7 @@ class HyperLogLog {
     HllDataType _type = HLL_DATA_EMPTY;
 
     uint32_t _explicit_data_num = 0;
-    uint64_t _explicit_data[HLL_EXPLICIT_INT64_NUM * 2];
+    uint64_t* _explicit_data = nullptr;
```
HappenLee, line 177: directly use the raw ptr is dangerous and complex. Rethink shoule we replace is use std::unique_ptr

zuochunwei: no need replace with unique_ptr,  keep style consistency with uint8_t *_registers

be/src/olap/hll.h
zuochunwei: fixed it

morningman (dismissed):
LGTM

morningman (approved):

PR #6976, opened 2021-10-30
generage compile_command.json

ucasfl:
Set cmake to generate `compile_commands.json`, which is useful for lsp like clangd, cquery, et.


morningman (approved):
LGTM

PR #6988, opened 2021-11-02
mem_tracker optimization

zuochunwei:
optimise HighWaterMarkCounter::add(),  UpdateMax only if delta greate than 0 to reduce function call times
delete useless code lines to keep MemTrack clean

some member datas never be set, but check its value，the if condition never be satisified, so clean these codes


morningman (approved):
LGTM

PR #7024, opened 2021-11-05
fix core problem: use object replace pointer

zuochunwei:
old version:
NodeChannel::_node_info is a pointer member data， this member pointer is set by _parent->_nodes_info->find_node in the function NodeChannel::init(), _parent->_nodes_info is managed by object_pool, the object_pool‘ lifetime is the same with PlanFragmentExecutor.

but _node_info will be used in _add_batch_closure' success handler, it's a async callback, in this callback, PlanFragmentExecutor created for other RPC may be destroy for some reasons, such as timeout or canceled, so continue use _node_info is wrong.

refactor: use NodeInfo _node_info replace NodeInfo *_node_info, just copy NodeInfo in init function

### DIFF
--- a/be/src/exec/tablet_info.h
+++ b/be/src/exec/tablet_info.h
@@ -243,6 +243,8 @@ struct NodeInfo {
     std::string host;
     int32_t brpc_port;
 
+    NodeInfo() = default;
+
     NodeInfo(const TNodeInfo& tnode)
             : id(tnode.id),
               option(tnode.option),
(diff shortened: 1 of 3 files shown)


morningman (dismissed):
LGTM

morningman:
Conflict

zuochunwei:
> Conflict

done

morningman:
> > Conflict
> 
> done

Unit test compile failed. You need to run `run-be-ut.sh`

morningman (approved):
LGTM

Issue #7009, opened 2021-11-04
[Proposal] Add tablets number limit to affend wrong usage


tianhui5:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Description
A user input a SQL like
 `alter table sample_table add partition  tmp20211103 VALUES [("20211102"),("20211103")) DISTRIBUTED BY HASH(hour,keyword,tagId) BUCKETS 20211103`
So there is 20211103 tablets, costs very large number of FE memory.
The monitor is below:
![image](https://user-images.githubusercontent.com/15223868/140300625-7a4d1f66-7c44-4864-8acb-efc6f8dd1aa0.png)
![image](https://user-images.githubusercontent.com/15223868/140300832-78856c3d-1c7a-44f0-a212-5294ed1262f5.png)
As above image shows, too many tablets cause frontend OOM. I think it's necessary to add tablets number limit.
 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

xinghuayu007:
up

PR #7025, opened 2021-11-05
Add tablets number limit to affend wrong usage

MorningLight5:
## Proposed changes

#7009 

## Types of changes

- [ ] Other.

## Checklist

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
@@ -3254,6 +3254,16 @@ public class Catalog {
         DataProperty dataProperty = singlePartitionDesc.getPartitionDataProperty();
         Preconditions.checkNotNull(dataProperty);
 
+        // check replica quota if this operation done
+        long indexNum = indexIdToMeta.size();
+        long bucketNum = distributionInfo.getBucketNum();
+        long replicaNum = singlePartitionDesc.getReplicaAlloc().getTotalReplicaNum();
+        long totalReplicaNum = indexNum * bucketNum * replicaNum;
+        if (totalReplicaNum >= db.getReplicaQuotaLeftWithLock()) {
+            throw new DdlException("Database " + db.getFullName() + " table " + tableName
+                    + " add partition increasing " + totalReplicaNum
+                    + " of replica exceeds quota[" + db.getReplicaQuota() + "]");
+        }
         Set<Long> tabletIdSet = new HashSet<Long>();
         try {
             long partitionId = getNextId();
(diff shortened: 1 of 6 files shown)


morningman (commented):
I think we can use replicaQuotaSize in Database to limit.

tianhui5:
> I think we can use replicaQuotaSize in Database to limit.

You are right, it can reach out same result. But the threshold value can only be changed in db level, I'll change the value from constant to config, so that we can change it in cluster level.

tianhui5:
The previous replica check can only check the history quota, I add the current operation quota check.

fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -3254,6 +3254,16 @@ public void addPartition(Database db, String tableName, AddPartitionClause addPa
         DataProperty dataProperty = singlePartitionDesc.getPartitionDataProperty();
         Preconditions.checkNotNull(dataProperty);
 
+        // check replica quota if this operation done
```
morningman, line 3257: The check here, and check at line L3851-3858 and L3895-3907 can be done in `Catalog.createPartitionWithIndices()`, before `createTablets()`

https://github.com/apache/incubator-doris/blob/4c6cbdf4635e48a3465bdfd7302cd3a0faabba42/fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java#L3527-L3537

fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -6691,6 +6726,19 @@ public void truncateTable(TruncateTableStmt truncateTableStmt) throws DdlExcepti
         // tabletIdSet to save all newly created tablet ids.
         Set<Long> tabletIdSet = Sets.newHashSet();
         try {
+            // check replica quota if this operation done
+            long totalReplicaNum = 0;
```
morningman, line 6730: For `truncate` operation. I don't think we should check replica quota. Otherwise user may not be able to truncate table.

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -1384,6 +1384,12 @@
     @ConfField(mutable = true, masterOnly = true)
     public static long default_db_data_quota_bytes = 1024L * 1024 * 1024 * 1024 * 1024L; // 1PB
 
+    /**
+     * Used to set default db replica quota num.
+     */
+    @ConfField(mutable = true, masterOnly = true)
+    public static long default_db_replica_quota_size = 1024 * 1024 * 1024;
```
morningman, line 1391: Add this new config in
```
docs/zh-CN/administrator-guide/config/fe_config.md
docs/en/administrator-guide/config/fe_config.md
```

fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
tianhui5: L3895-3907,when there are multi partitions, I don't think we can check quota in single partition.

morningman (approved):
LGTM

PR #7144, opened 2021-11-18
bugfix: ReportProfile

zuochunwei:
issue description:
1. setting _report_thread_active to false is not necessary protected by _report_thread_lock, because _report_thread_active's type is bool, writing data is multi-threadly safety if size <= marchine word length
2. report_profile thread terminates early is possiable, in the function report_profile(), while (_report_thread_active) may break if  _report_thread_active is false,  the thread of calling open() may be scheduled out between _report_thread_started_cv.wait(l) and _report_thread_active = true, we should not assume that how long time elapsed between a thread be scheduled twice


yangzhg (approved):
LGTM

PR #7145, opened 2021-11-18
no static_cast

zuochunwei:
provider interface in abstract-class replace static_cast
static_cast is not good for oop

### DIFF
--- a/be/src/exec/exec_node.h
+++ b/be/src/exec/exec_node.h
@@ -66,7 +66,7 @@ public:
     // Init conjuncts.
     ExecNode(ObjectPool* pool, const TPlanNode& tnode, const DescriptorTbl& descs);
 
-    virtual ~ExecNode();
+    virtual ~ExecNode() {}
 
     /// Initializes this object from the thrift tnode desc. The subclass should
     /// do any initialization that can fail in Init() rather than the ctor.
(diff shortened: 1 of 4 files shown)


be/src/exec/exec_node.h
```diff
@@ -135,6 +135,12 @@ class ExecNode {
     // each implementation should start out by calling the default implementation.
     virtual Status close(RuntimeState* state);
 
+    virtual void set_num_senders(int num_senders) {}
```
morningman, line 138: I think it is not good to add the subclass method to the parent class.
These methods are indeed just for subclass.

zuochunwei: when you need cast by hand writing in C++,  it's almost means existed some design issues. try resolving this problem by cast is wrong, the problem of design should resolved by re-design.

when you hold a pointer or reference, but you want to use the interfaces designed for subclass, so the problem is type info missed, using cast is similar said to compiler: "don't trouble me, don't help me to check security" we should write code fit for compiler, not discard compiler, it's dangours.

this viewpoint is not mine, but a major standpoint in OOP, you can find so much related articles, such as CppCoreGuildlines, edited by Bjarne Stroustrup（the father of C++） and Herb Sutter(C++ committee chairman)

morningman: OK, I see.
But I wonder is it possible to solve this problem by modifying `ExecNode::collect_scan_nodes` to return the real node type pointer by request?

PR #7183, opened 2021-11-23
[BUG] Fix `curdate()+0` to behave like mysql

littleeleventhwolf:
## Proposed changes

I proposed this point in [PR-6005](https://github.com/apache/incubator-doris/pull/6005).
Now I try to make `curdate()+0` behave like mysql.
NOW✅:
```
mysql> SELECT CURDATE() + 0;
+-----------------+
| CURDATE() + 0.0 |
+-----------------+
|        20211123 |
+-----------------+
```
BEFORE❌:
```
mysql> select CURDATE() + 0;
+-----------------+
| curdate() + 0.0 |
+-----------------+
|  20211123104151 |
+-----------------+
```

Also, I add `current_date` annotated as a FEFunction.

## Types of changes

- [ ] Other.

## Checklist

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/DateLiteral.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/DateLiteral.java
@@ -374,7 +374,7 @@ public class DateLiteral extends LiteralExpr {
         if (!getType().equals(expr.getType())) {
             if (getType().equals(Type.DATE)) {
                 thisExprLongValue *= BASIC_ZERO_TIME;
-            } else {
+            } else if (expr.getType().equals(Type.DATE)) {
                 otherExprLongValue *= BASIC_ZERO_TIME;
             }
         }
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/DateLiteral.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/DateLiteral.java
@@ -361,8 +361,18 @@ public class DateLiteral extends LiteralExpr {
         if (expr == MaxLiteral.MAX_VALUE) {
             return -1;
         }
+
+        long thisExprLongValue = getLongValue();
+        long otherExprLongValue = expr.getLongValue();
+        if (!getType().equals(expr.getType())) {
+            if (getType().equals(Type.DATE)) {
+                thisExprLongValue *= 1000000;
+            } else {
+                otherExprLongValue *= 1000000;
+            }
+        }
         // date time will not overflow when doing addition and subtraction
-        return Long.signum(getLongValue() - expr.getLongValue());
+        return Long.signum(thisExprLongValue - otherExprLongValue);
     }
 
     @Override
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/DateLiteral.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/DateLiteral.java
@@ -66,6 +66,13 @@ public class DateLiteral extends LiteralExpr {
     private static final int MAX_MICROSECOND = 999999;
     private static final int DATETIME_TO_MINUTE_STRING_LENGTH = 16;
     private static final int DATETIME_TO_HOUR_STRING_LENGTH = 13;
+    /*
+     * A long value datetime in the form of 'yyyyMMddHHmmss' divided by BASIC_ZERO_TIME
+     * should obtain a long value date in the form of 'yyyyMMdd'.
+     * Conversely, a long value date in the form of 'yyyyMMdd' times BASIC_ZERO_TIME
+     * should obtain a long value datetime in the form of 'yyyyMMddHHmmss', means zero o'clock of that day.
+     */
+    public static final long BASIC_ZERO_TIME = 1000000L;
 
     private static DateTimeFormatter DATE_TIME_FORMATTER = null;
     private static DateTimeFormatter DATE_TIME_FORMATTER_TO_HOUR = null;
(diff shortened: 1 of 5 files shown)


fe/fe-core/src/main/java/org/apache/doris/analysis/DateLiteral.java
```diff
@@ -361,8 +361,18 @@ public int compareLiteral(LiteralExpr expr) {
         if (expr == MaxLiteral.MAX_VALUE) {
             return -1;
         }
+
+        long thisExprLongValue = getLongValue();
+        long otherExprLongValue = expr.getLongValue();
+        if (!getType().equals(expr.getType())) {
+            if (getType().equals(Type.DATE)) {
+                thisExprLongValue *= 1000000;
```
morningman, line 369: Better define this `1000000` as a final static variable, and explain what this number's mean.

morningman (commented):
Thanks for your PR.
And could you please add ut for this case. You can refer to QueryPlanTest.java

fe/fe-core/src/main/java/org/apache/doris/analysis/DateLiteral.java
littleeleventhwolf: Thanks for your reply.
I have modified it according to your suggestion.

fe/fe-core/src/main/java/org/apache/doris/qe/cache/PartitionRange.java
```diff
@@ -48,6 +48,8 @@
 import java.util.List;
 import java.util.Map;
 
+import static org.apache.doris.analysis.DateLiteral.BASIC_ZERO_TIME;
```
morningman, line 51: not using static import:
http://doris.incubator.apache.org/master/zh-CN/developer-guide/java-format-code.html#import-order

fe/fe-core/src/test/java/org/apache/doris/planner/QueryPlanTest.java
```diff
@@ -17,6 +17,7 @@
 
 package org.apache.doris.planner;
 
+import com.google.common.base.Splitter;
```
morningman, line 20: import order:
http://doris.incubator.apache.org/master/zh-CN/developer-guide/java-format-code.html#import-order

fe/fe-core/src/main/java/org/apache/doris/analysis/DateLiteral.java
```diff
@@ -361,8 +368,18 @@ public int compareLiteral(LiteralExpr expr) {
         if (expr == MaxLiteral.MAX_VALUE) {
             return -1;
         }
+
+        long thisExprLongValue = getLongValue();
+        long otherExprLongValue = expr.getLongValue();
+        if (!getType().equals(expr.getType())) {
+            if (getType().equals(Type.DATE)) {
```
morningman, line 375: Are you sure the `expr` is either `DATE` or `DATETIME` type?
May it be a Long value? for example.

littleeleventhwolf: Thank you for reminding.
`Expr` could be other type besides `DATE` and `DATETIME`. I find the constructor or `init` method separately carry out `DATE` type. Other types are treated as `DATETIME` indeed, even though the assigned type is not `DATETIME`.

littleeleventhwolf: Thank you for reminding.
`Expr` could be other type besides `DATE` and `DATETIME`. I find the constructor or `init` method separately carry out `DATE` type. Other types are treated as `DATETIME` indeed, even though the assigned type is not `DATETIME`.

littleeleventhwolf: Could you please give me some advice on how to process other types reasonably?
What I can think of now is to process `DATE` type separately, because the constructor or init-method or getLongValue-method only process `DATE` type separately too. And other types (e.g. LONG), the assigned type is not `DATEIME`, but in fact they are still processed as `DATETIME`.

fe/fe-core/src/main/java/org/apache/doris/qe/cache/PartitionRange.java
```diff
@@ -238,7 +238,11 @@ public long realValue() {
         }
 
         private Date getDateValue(LiteralExpr expr) {
-            value = expr.getLongValue() / 1000000;
+            if (expr.getType().equals(Type.DATE)) {
+                value = expr.getLongValue();
+            } else {
+                value = expr.getLongValue() / DateLiteral.BASIC_ZERO_TIME;
```
morningman, line 244: You already handle the type in `DateLiteral.getLongValue()`, why divide `BASIC_ZERO_TIME` here again?

littleeleventhwolf: Yeah, you're right. Method `getDateValue` is only invoked in the case of `DATE`, the division is unnecessary.

littleeleventhwolf: In `PartitionRange` class, the first parameter `type` of init-method cannot represent the second parameter `expr`'s real type, as illustrated in following figure:
image
 
So I reset to last commit, still handle the `expr`'s type with if-else statement, non-`DATE` type should divide `BASIC_ZERO_TIME`.

What about your opinion?

PR #7216, opened 2021-11-25
[Feature]. Improve DeltaWriter's performance.

### LINKED ISSUE #7214: [Feature] Improve DeltaWriter's performance
### Search before asking

### Description

# Background

DeltaWriter may encounter serious performance problems when write concurrency is high, such as the following case
- There are a large number of insert-into-select tasks in the cluster
- When use BrokerLoad to load  large quantities of data with high concurrency

The following is a flame graph of a cluster which contains a large number of insert-into-select tasks

image

The following is a flame graph when  executing brokerload. The parallel of brokerload is set to 4, data to load: 200GB, 40 billion

image

# Solution

- When writing data to the DeltaWriter, group the data by tablet first and then write in batch, which can greatly reduce the lock/unlock operation

- Use mutex instead of SpinLock, because DeltaWriter.write is not a lightweight operation(It may hold the lock for a long time), using SpinLock while consumes too much CPU when the write concurrency is high

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

killxdcj:
## Proposed changes

Improve DeltaWriter's performance, detail in #7214

- Support batch write for DeltaWriter.

- Use mutex instead of SpinLock.

## Types of changes

- [ ] Other.

## Checklist

- [x] I have created an issue on (Fix #7214) and described the bug/feature there in detail

### DIFF
--- a/be/src/olap/delta_writer.cpp
+++ b/be/src/olap/delta_writer.cpp
@@ -24,6 +24,8 @@
 #include "olap/schema.h"
 #include "olap/schema_change.h"
 #include "olap/storage_engine.h"
+#include "runtime/row_batch.h"
+#include "runtime/tuple_row.h"
 
 namespace doris {
 
@@ -153,7 +155,7 @@ OLAPStatus DeltaWriter::init() {
 }
 
 OLAPStatus DeltaWriter::write(Tuple* tuple) {
-    std::lock_guard<SpinLock> l(_lock); 
+    std::lock_guard<std::mutex> l(_lock);
     if (!_is_init && !_is_cancelled) {
         RETURN_NOT_OK(init());
     }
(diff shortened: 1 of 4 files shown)


be/src/olap/delta_writer.cpp
```diff
@@ -179,6 +180,26 @@ OLAPStatus DeltaWriter::write(Tuple* tuple) {
...
+        RETURN_NOT_OK(init());
+    }
+
+    if (_is_cancelled) {
+        return OLAP_ERR_ALREADY_CANCELLED;
+    }
+
+    for (const auto& row_idx : row_idxs) {
+        _mem_table->insert(row_batch->get_row(row_idx)->get_tuple(0));
+        if (_mem_table->memory_usage() >= config::write_buffer_size) {
```
morningman, line 195: I think we can move this memory check outside the for loop.
Maybe inaccurate, but its ok~

killxdcj: Comment resolved~

morningman (dismissed):
LGTM

morningman:
Hi @killxdcj , please fix BE ut.

killxdcj:
> 

@morningman BE ut fixed~

morningman (approved):

PR #7225, opened 2021-11-25
[optimize]Partition column and bucket column should not have duplicates

renzhimin7:
Even if the field is a partition column and a bucket column, there will be performance problems when the in operation is performed according to the field  https://github.com/apache/incubator-doris/issues/7165

## Proposed changes

## Types of changes

- [ ] Other.

## Checklist

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/CreateTableStmt.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/CreateTableStmt.java
@@ -406,6 +406,12 @@ public class CreateTableStmt extends DdlStmt {
                 throw new AnalysisException("Create olap table should contain distribution desc");
             }
             distributionDesc.analyze(columnSet);
+            List<String> distributionColNames = ((HashDistributionDesc)distributionDesc).getDistributionColumnNames();
+            if (partitionDesc != null) {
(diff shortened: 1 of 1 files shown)


renzhimin7:
When querying, cut the partition first. After determining the partition, cut the bucket. When the column is both a bucket column and a partition column, if you use the in operation, you will select more buckets during bucket cutting. Moreover, each zone column is of little significance as a bucket column, so it is limited when creating a table.

fe/fe-core/src/main/java/org/apache/doris/analysis/CreateTableStmt.java
```diff
@@ -406,6 +406,12 @@ public void analyze(Analyzer analyzer) throws UserException {
                 throw new AnalysisException("Create olap table should contain distribution desc");
             }
             distributionDesc.analyze(columnSet);
+            List<String> distributionColNames = ((HashDistributionDesc)distributionDesc).getDistributionColumnNames();
+            if (partitionDesc != null) {
+                if (checkIsDuplicate(partitionDesc.partitionColNames, distributionColNames)) {
+                    throw new AnalysisException("Partition column and bucket column should not have duplicates");
```
morningman, line 412: I don't think we should restrict users from this kind of use.
The problem you described in issue should be fixed by modifying the prune logic?

PR #7240, opened 2021-11-29
[Load] Transaction for load job with no data for all partitions should be considered as normal and should not be aborted

### LINKED ISSUE #7239: [Feature] Enable Load job with all partitions have no data succeed and be compatible with Mysql
### Search before asking

### Description

Transaction for load job with no data for all partitions should be considered as normal and should not be aborted

### Use case

for some stream load case, it would be normal when stream load have no data for all partitions, and transaction should not be failed.
for  insert job, we should keep insert job succeed which behave like mysql 

### Related issues

_No response_

### Code of Conduct

caiconghui:
## Proposed changes

## Types of changes

- [ ] Other.

## Checklist

- [x] I have created an issue on (Fix #7239 ) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/load/routineload/KafkaRoutineLoadJob.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/routineload/KafkaRoutineLoadJob.java
@@ -240,16 +240,6 @@ public class KafkaRoutineLoadJob extends RoutineLoadJob {
             return true;
         }
 
-        if (txnStatusChangeReason != null && txnStatusChangeReason == TransactionState.TxnStatusChangeReason.NO_PARTITIONS) {
-            // Because the max_filter_ratio of routine load task is always 1.
-            // Therefore, under normal circumstances, routine load task will not return the error "too many filtered rows".
-            // If no data is imported, the error "all partitions have no load data" may only be returned.
-            // In this case, the status of the transaction is ABORTED,
-            // but we still need to update the offset to skip these error lines.
-            Preconditions.checkState(txnState.getTransactionStatus() == TransactionStatus.ABORTED, txnState.getTransactionStatus());
-            return true;
-        }
-
         // Running here, the status of the transaction should be ABORTED,
         // and it is caused by other errors. In this case, we should not update the offset.
         LOG.debug("no need to update the progress of kafka routine load. txn status: {}, " +
--- a/fe/fe-core/src/main/java/org/apache/doris/load/update/UpdateStmtExecutor.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/update/UpdateStmtExecutor.java
@@ -44,7 +44,6 @@ import org.apache.doris.thrift.TUniqueId;
 import org.apache.doris.transaction.BeginTransactionException;
 import org.apache.doris.transaction.GlobalTransactionMgr;
 import org.apache.doris.transaction.TabletCommitInfo;
-import org.apache.doris.transaction.TransactionCommitFailedException;
 import org.apache.doris.transaction.TransactionState.LoadJobSourceType;
 import org.apache.doris.transaction.TransactionState.TxnCoordinator;
 import org.apache.doris.transaction.TransactionState.TxnSourceType;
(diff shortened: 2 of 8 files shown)


fe/fe-core/src/main/java/org/apache/doris/load/routineload/KafkaRoutineLoadJob.java
```diff
@@ -240,16 +240,6 @@ protected boolean checkCommitInfo(RLTaskTxnCommitAttachment rlTaskTxnCommitAttac
...
         }
 
-        if (txnStatusChangeReason != null && txnStatusChangeReason == TransactionState.TxnStatusChangeReason.NO_PARTITIONS) {
-            // Because the max_filter_ratio of routine load task is always 1.
-            // Therefore, under normal circumstances, routine load task will not return the error "too many filtered rows".
-            // If no data is imported, the error "all partitions have no load data" may only be returned.
-            // In this case, the status of the transaction is ABORTED,
-            // but we still need to update the offset to skip these error lines.
-            Preconditions.checkState(txnState.getTransactionStatus() == TransactionStatus.ABORTED, txnState.getTransactionStatus());
-            return true;
```
morningman, line 250: We need to return true to skip those unloaded rows.

fe/fe-core/src/main/java/org/apache/doris/load/update/UpdateStmtExecutor.java
```diff
@@ -177,21 +176,7 @@ private void executePlan() throws Exception {
...
-            LOG.info("abort transaction for update stmt, query id:{}, reason: {}", DebugUtil.printId(queryId),
-                    TransactionCommitFailedException.NO_DATA_TO_LOAD_MSG);
-            globalTransactionMgr.abortTransaction(dbId, txnId, TransactionCommitFailedException.NO_DATA_TO_LOAD_MSG);
-            StringBuilder sb = new StringBuilder();
-            sb.append("{'label':'").append(label);
-            sb.append(", 'txnId':'").append(txnId).append("'");
-            sb.append(", 'queryId':'").append(DebugUtil.printId(queryId)).append("'");
-            sb.append("}");
-            analyzer.getContext().getState().setOk(effectRows, 0, sb.toString());
-            return;
```
morningman, line 191: I remember that Doris do not support commit empty rowset. Have you tested this case yet?

fe/fe-core/src/main/java/org/apache/doris/load/routineload/KafkaRoutineLoadJob.java
caiconghui: when we can commit empty rowset, txnStatusChangeReason would be null

fe/fe-core/src/main/java/org/apache/doris/load/update/UpdateStmtExecutor.java
caiconghui: yes, i have tested stream load and insert already, it works well. actually, we should keep same behavior with mysql, whether update or insert 0 rows

yiguolei: Yes， Doris support empty rowset. Because when there are 10 tablets, but user could insert 1 row, and this row only exist in one tablet and other tablets have empty rowset.

But, we should optimize this. If all tablets have empty rowsets, then just return success, not commit it, and abort the txn.

caiconghui: @yiguolei now, commit and publish txn without any tablet  almost cost nothing, we should unity it to avoid user to load data again, because no partition with data load may have some cases for normal and unnormal situation, and we may write extra code to handle this case for insert and update, which make code look like ugly and may not easy to handle

yiguolei: Yes, I agree with you.

Then LGTM.

morningman (approved):
LGTM

PR #7265, opened 2021-11-30
[BUG] Rewrite Bigint SlotRef to compare DecimalLiteral in Binary predicate

### LINKED ISSUE #7264: [Bug] Bigint key column in where clause equal double constant, bucket cannot be hit
### Search before asking

### Version

trunk

### What's Wrong?

When the bigint key column in the where clause is equal to the double constant, the bucket cannot be hit.

Reproduce on the ssb test set:
`select * from LINEORDER where LO_ORDERKEY = 33.0 limit 1;`

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

xinyiZzz:
## Proposed changes

Convert the binary predicate of the form
`<CastExpr<SlotRef(ResultType=BIGINT)>> <op><DecimalLiteral>` 
to the binary predicate of 
`<SlotRef(ResultType=BIGINT)> <new op> <new DecimalLiteral>`, 
thereby allowing the binary predicate The predicate pushes down and completes the bucket clipped.

For query `select * from T where t1 = 2.0`, when the ResultType of column t1 is equal to BIGINT, in the binary predicate analyze, the type will be unified to DECIMALV2, so the binary predicate will be converted to `<CastExpr<SlotRef>> <op In the form of ><DecimalLiteral>`, because Cast wraps the t1 column, it cannot be pushed down, resulting in poor performance.We convert it to the equivalent query `select * from T where t1 = 2` to push down and improve performance.

SSB test:
1. query `select * from LINEORDER3 where LO_ORDERKEY <2.2`
Performance improvement: `1.587s` -> `0.012s`,
The result and performance of `select * from LINEORDER3 where LO_ORDERKEY <3` are equivalent, and the other comparison methods are the same.
2. query `select * from LINEORDER3 where LO_ORDERKEY = 2.2`
Performance improvement: `0.012s` -> `0.006`.

## Types of changes

- [ ] Other.

## Checklist

- [x] I have created an issue on (Fix #7264) and described the bug/feature there in detail

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/DecimalLiteral.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/DecimalLiteral.java
@@ -231,6 +231,14 @@ public class DecimalLiteral extends LiteralExpr {
         return fracPart.intValue();
     }
 
+    public void roundCeiling() {
+        value = value.setScale(0, 2);
+    }
+
+    public void roundFloor() {
+        value = value.setScale(0, 3);
+    }
+
     @Override
     protected Expr uncheckedCastTo(Type targetType) throws AnalysisException {
         if (targetType.isDecimalV2()) {
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/DecimalLiteral.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/DecimalLiteral.java
@@ -35,6 +35,7 @@ import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
 import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.util.Objects;
@@ -232,11 +233,11 @@ public class DecimalLiteral extends LiteralExpr {
     }
 
     public void roundCeiling() {
-        value = value.setScale(0, 2);
+        value = value.setScale(0, RoundingMode.CEILING);
     }
 
     public void roundFloor() {
-        value = value.setScale(0, 3);
+        value = value.setScale(0, RoundingMode.FLOOR);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


yangzhg:
We need more discuss

yangzhg (changes requested):
We need more discuss

xinyiZzz:
> We need more discuss

Modified according to discussion

fe/fe-core/src/main/java/org/apache/doris/analysis/DecimalLiteral.java
```diff
@@ -231,6 +231,14 @@ public int getFracValue() {
         return fracPart.intValue();
     }
 
+    public void roundCeiling() {
+        value = value.setScale(0, 2);
+    }
```
yangzhg, line 236: use RoundingMode.ROUND_CEILING

fe/fe-core/src/main/java/org/apache/doris/analysis/DecimalLiteral.java
```diff
@@ -231,6 +231,14 @@ public int getFracValue() {
         return fracPart.intValue();
     }
 
+    public void roundCeiling() {
+        value = value.setScale(0, 2);
```
yangzhg, line 235:
```suggestion
        value = value.setScale(0, RoundingMode.CEILING);
```

fe/fe-core/src/main/java/org/apache/doris/analysis/DecimalLiteral.java
```diff
@@ -231,6 +231,14 @@ public int getFracValue() {
...
+    }
+
+    public void roundFloor() {
+        value = value.setScale(0, 3);
```
yangzhg, line 239:
```suggestion
        value = value.setScale(0, RoundingMode.FLOOR);
```

yangzhg:
You should add some unit test to cover your 12 scenarios

fe/fe-core/src/main/java/org/apache/doris/analysis/DecimalLiteral.java
xinyiZzz: Thats right

fe/fe-core/src/main/java/org/apache/doris/analysis/DecimalLiteral.java
xinyiZzz: Thats right

yangzhg (approved):
LGTM

PR #7284, opened 2021-12-03
[fix](executor)  set the len of StringValue to 0 when it is null

zuochunwei:
## Proposed changes

brief description:
the tuple String Slot's ptr and len are not assigned appropriately on send side, the receive side may crash in some situation. 

detail description:
on send side, when we call RowBatch::serialize(PRowBatch* output_batch) to pack RowBatch, the Tuple::deep_copy() will be called,  for each String Slot,  only String Slots that is not null will set ptr and len with proper value, the null String Slots will keep original status, the ptr member will point randomly and the len member may unexpect. 

on recv side, unpack is processed by RowBatch::RowBatch(const RowDescriptor&, const PRowBatch&...), in this function, each String Slot will transfer offset to valid string_val->ptr whether the String Slot is null or not. 

but some business logic depends on string_val->len=0, such as AggregateFuncTraits::init(), HyperLogLog::deserialize() will return correctly if slice.size<=0. so if string_val->len is set to 0 in send side, everything will be ok, otherwise server may crash.

by netcomm viewpoint, we should make sure transfer correct data, it's sender's responsibility to set data with proper value, and do not make any presume which way the recv side will use it.

## Types of changes

- [ x] Bugfix (non-breaking change which fixes an issue)
- [ ] Other.

## Checklist

- [ x] Compiling and unit tests pass locally with my changes

## Further comments

### DIFF
--- a/be/src/runtime/tuple.cpp
+++ b/be/src/runtime/tuple.cpp
@@ -73,14 +73,17 @@ void Tuple::deep_copy(Tuple* dst, const TupleDescriptor& desc, MemPool* pool, bo
          i != desc.string_slots().end(); ++i) {
         DCHECK((*i)->type().is_string_type());
 
+        StringValue* string_v = dst->get_string_slot((*i)->tuple_offset());
         if (!dst->is_null((*i)->null_indicator_offset())) {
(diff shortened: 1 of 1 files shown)


morningman (approved):
LGTM

PR #7312, opened 2021-12-06
[revert] "clean too many bdbje log"

morningman:
Reverts apache/incubator-doris#7273
Because there is no EnvironmentConfig.RESERVED_DISK.


yangzhg (approved):
LGTM

Issue #7355, opened 2021-12-08
[Bug] NullPointerException will be in `org/apache/doris/analysis/InsertStmt.java`


jackwener:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Version

latest

### What's Wrong?

line 607 of `Nullorg/apache/doris/analysis/InsertStmt.java`

entry is null, it's wrong to call `entry.first`;

``` java
if (entry == null) {
    extentedRow.add(extentedRow.get(entry.first));
}
```

### What You Expected?

fix this bug

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #7360, opened 2021-12-09
[fix] skip the condition of null

jackwener:
## Proposed changes

Close related #7355 

when entry is null, there will be `NullPointerException`.

## Types of changes

- [ ] Other.

## Checklist

## Further comments

None

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/InsertStmt.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/InsertStmt.java
@@ -595,7 +595,7 @@ public class InsertStmt extends DdlStmt {
         ArrayList<Expr> row = rows.get(rowIdx);
         if (!origColIdxsForExtendCols.isEmpty()) {
             /**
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/org/apache/doris/analysis/InsertStmt.java
```diff
@@ -604,9 +604,7 @@ private void analyzeRow(Analyzer analyzer, List<Column> targetColumns, List<Arra
             extentedRow.addAll(row);
             
             for (Pair<Integer, Column> entry : origColIdxsForExtendCols) {
-                if (entry == null) {
```
jackwener, line 607: I don't know whether it's right solution. But this problem didn't happen past, so I just remove it.

morningman: indeed this is a wrong logic

morningman (approved):
LGTM

PR #7427, opened 2021-12-17
Revert "[improvement](reader) optimize for single rowset reading  (#7351)"

morningman:
Reverts apache/incubator-doris#7351

This commit will cause wrong result with agg table.
For example, an agg table `(k1, k2, v1 sum)` with single non-overlapping rowset

`select count(k1) from tbl1;` should using `_direct_agg_key_next_row` instead of `_agg_key_next_row`.

Otherwise it return less rows than expected.(because `_agg_key_next_row` will only do aggregation with `k1`)


morningman:
hi @GoGoWen , this PR #7351 has to be reverted because the logic is incorrect.
We may need to find another way to optimize it.

yangzhg (approved):
lgtm

PR #7442, opened 2021-12-20
cancel export job

wangxiaobaidu11:
**Description**
When exporting a large number data, users should cancel the operation during the operation to release resources as soon as possible.


morningman:
Still work in progress.

PR #7445, opened 2021-12-20
[Feature#7441][Doc] Seatunnel Suppoorts Doris connector

xtr1993:


hf200012:
It is recommended to give a complete usage example, including the installation and deployment of SeaTunnel (document connection can be given), combined with a Mysql or other data source for ETL and access to Doris, which can be used as a reference for users in the future

PR #7451, opened 2021-12-21
[Compile] Support build with clang

yiguolei:
Support build Doris BE with clang

Currently, thirdparty could only be build with GCC

### DIFF
--- a/be/CMakeLists.txt
+++ b/be/CMakeLists.txt
@@ -18,13 +18,6 @@
 cmake_minimum_required(VERSION 3.19.2)
 
 # set CMAKE_C_COMPILER, this must set before project command
-if (DEFINED ENV{DORIS_GCC_HOME})
-    set(CMAKE_C_COMPILER "$ENV{DORIS_GCC_HOME}/bin/gcc")
-    set(CMAKE_CXX_COMPILER "$ENV{DORIS_GCC_HOME}/bin/g++")
-    set(GCC_HOME $ENV{DORIS_GCC_HOME})
-else()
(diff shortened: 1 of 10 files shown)


wangbo:
Maybe provide clang in thrid-party or docker image?
And make ```build with clang```  an optional args?

be/CMakeLists.txt
```diff
@@ -18,13 +18,6 @@
 cmake_minimum_required(VERSION 3.19.2)
 
 # set CMAKE_C_COMPILER, this must set before project command
-if (DEFINED ENV{DORIS_GCC_HOME})
```
morningman, line 21: I think this should be an optional choice?

yiguolei: No, We should remove GCC HOME env. 
User should only define CC and CXX env, 
for example
export CC= XXXXXXXXXX
export CXX=xxxxxxxxxxxxxx

I will remove GCC HOME and rewrite bitshuffle build method.

yiguolei:
> Maybe provide clang in thrid-party or docker image? And make `build with clang` an optional args?

Currently, could only build be with clang, And I don't want it to be an optional args because I think we should use clang as default build tool chain in the future.

be/CMakeLists.txt
yiguolei: If we want to build with clang, then we export CC=XXXX/clang  and CXX=XXXXXXX/clang++
If we want to build with gcc, then we export CC=xxxx/gcc and CXX=xxxxx/g++

wangbo:
It seems that ```--clean```  and ```incremental compilation``` is not support in this pr;
Every time I need to clean manually

PR #7198, opened 2021-11-23
[Memory] Refactor MemTracker and new memory statistics framework based on TCMalloc Hook

xinyiZzz:
## Proposed changes

1. Add thread local variable TheadContext, save task info, MemTracker, etc.;
2. Implement a new memory statistics framework based on TCMalloc Hook and TheadContext, and it is expected that all memory consume/release of the BE process can be counted; (Introduction: https://shimo.im/docs/DT6JXDRkdTvdyV3G)
3. Refactor MemTracker, streamline the code, and adapt to the new statistical framework;
4. Add Task MemTracker Pool, manage all Query, Load, and Compaction tasks on BE, a Task ID is unique in the BE process, and judge whether it exceeds the Query memory limit based on the sum of the memory consumption of all fragments of a Task.
(Unlike before, The query memory limit was previously used as a fragment instance level limit. Therefore, some previously successful large queries may fail, and the query memory limit configuration needs to be increased.)
5. Add new Process MemTracker, ChunkAllocator MemTracker, and modify the memory statistics method of MemPool and BufferPool.

[ impact of testing on performance ]
The current default untracked_mem_limit_mbytes = 1M, the test found no impact on performance
Config `untracked_mem_limit_mbytes` impact on performance:
master, jmeter thread=50, Avg1: 23120, Avg2: 22505, Avg3: 23459
untracked_mem_limit_mbytes=4M, jmeter thread=50, Avg1: 23024, Avg2: 23391 22681
untracked_mem_limit_mbytes=2M, jmeter thread=50, Avg1: 23426, Avg2: 22674 23658
untracked_mem_limit_mbytes=64K, jmeter thread=50, Avg1: 24121, Avg2: 24042 23762 22939 23294
untracked_mem_limit_mbytes=4K, jmeter thread=50, Avg1: 24942, Avg2: 24983 23305 22611 23506

## Types of changes

What types of changes does your code introduce to Doris?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Code refactor (Modify the code structure, format the code, etc...)
- [ ] Optimization. Including functional usability improvements and performance improvements.
- [ ] Dependency. Such as changes related to third-party components.
- [ ] Other.

## Checklist

[...]

PR #7474, opened 2021-12-23
[Feature] Limit frequency of query and load in system level

### LINKED ISSUE #7129: [Feature] Limit cluster resource usage in user granularity
### Search before asking

### Description

In productive environment, the Doris cluster is often facing pressure from many aspects (mainly from stream load and query), cause many resource shortage problem like OOM, especially in shared cluster.
image
As above picture shows, the memory usage waves too big.
I think it's better to have a way to limit the resource usage of each user. Maybe limit the usage frequency is a proper way.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

MorningLight5:
## Proposed changes

Fix #7129 

## Types of changes

- [ ] Other.

## Checklist


xinyiZzz:
good idea,
If FE can get the memory consumption of each query, we can limit the memory used by users, which is what I plan to do in the future.

The actual AdmissionController of Impala does a similar thing, refer to: https://shimo.im/docs/6qxjctpyDHJgPwtw

In #7198, I have counted the real memory consumption of each Query, which may be implemented based on this PR later

morningman:
Hi @tianhui5 , thanks for your contribution.
But better to add document to explain the usage or best practice.
Otherwise, it is hard to review.

tianhui5:
> Hi @tianhui5 , thanks for your contribution. But better to add document to explain the usage or best practice. Otherwise, it is hard to review.

Thanks for remind me. I've add relative doc and unit test, please review this PR

tianhui5:
> Good idea, If FE can get the memory consumption of each Query and Load, we can limit the memory used by users, which is what I plan to do in the future.
> 
> Impala’s AdmissionController does a similar thing, Introduction is here https://shimo.im/docs/6qxjctpyDHJgPwtw
> 
> In #7198, I have counted the real memory consumption of each Query, which may be implemented based on your PR later

Nice work.

fe/fe-core/src/main/java/org/apache/doris/system/RateLimiterMgr.java
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * RateLimiterMgr collect all rate limiter of the system.
+ */
+public class RateLimiterMgr {
+    private RateLimiter systemQPSLimiter;
+    private RateLimiter systemStreamLoadLimiter;
+    private Map<String, RateLimiter> userToQPSLimiter;
+    private Map<String, RateLimiter> userToStreamLoadLimiter;
+
+    public RateLimiterMgr(double QPS, double streamLoadPS) {
```
morningman, line 37: It is not recommended to pass these values through the constructor's parameters. If we subsequently add more limiter, there will be more arguments to the constructor.

fe/fe-core/src/main/java/org/apache/doris/system/RateLimiterMgr.java
```diff
@@ -0,0 +1,100 @@
...
+        this.systemQPSLimiter = RateLimiter.create(QPS);
+        this.systemStreamLoadLimiter = RateLimiter.create(streamLoadPS);
+        this.userToQPSLimiter = Maps.newConcurrentMap();
+        this.userToStreamLoadLimiter = Maps.newConcurrentMap();
+    }
+
+    public synchronized boolean checkRateLimiter(TLimiterType type, String user) {
```
morningman, line 44: why using `synchronized`? the RateLimiter is thread safe

morningman:
There are two main issues

1. all query requests need to send an RPC to Master FE, which is too much overhead.
2. Limiter is not set in the right place. For example, for query, it should be handled at the place where the query plan is finally sent to BE. For load, the frequency of transactions should be limited.

tianhui5:
> There are two main issues
> 
> 1. all query requests need to send an RPC to Master FE, which is too much overhead.
> 2. Limiter is not set in the right place. For example, for query, it should be handled at the place where the query plan is finally sent to BE. For load, the frequency of transactions should be limited.

Yes, that seems to have too much RPC. I have rewrote the feature with statistic sychronize based on bdbje. Because I think it's fine to use bdbje to store some temperory statistics, and there is too much work to develop a middleware myself.

Please review this patch, thanks for your time!

morningman:
Hi @tianhui5 

Although we had some private discussions about this feature, we eventually needed a complete design and usage document to help all developers and users understand how this feature works and how to use it.

It is recommended that the final design be fully described in the issue #7129  (and the previous design be marked as Deprecated).

Also, I suggest a separate user document describing the functionality from a usage perspective, including the followings:

1. the purpose of the feature (Motivation)
2. what scenarios the feature is applicable to, how to use it, and what impact it will have.
3. whether there are shortcomings for further improvement.

In the previous Doris development, we also had a lot of features that were difficult to maintain due to lack of documentation, and I hope to be able to improve the process in this area in future projects. Thanks!

tianhui5:
> 

Done.

tianhui5:
> Hi @tianhui5
> 
> Although we had some private discussions about this feature, we eventually needed a complete design and usage document to help all developers and users understand how this feature works and how to use it.
> 
> It is recommended that the final design be fully described in the issue #7129 (and the previous design be marked as Deprecated).
> 
> Also, I suggest a separate user document describing the functionality from a usage perspective, including the followings:
> 
> 1. the purpose of the feature (Motivation)
> 2. what scenarios the feature is applicable to, how to use it, and what impact it will have.
> 3. whether there are shortcomings for further improvement.
> 
> In the previous Doris development, we also had a lot of features that were difficult to maintain due to lack of documentation, and I hope to be able to improve the process in this area in future projects. Thanks!
Done.

tianhui5:
The local current query number is reported to Master every certain time, and then send to all FE.
image

PR #7515, opened 2021-12-28
optimization reading for single rowset

GoGoWen:
## Proposed changes

this pr try to find a collect fix as  #7427.   the wrong fix is #7351.

## Types of changes

- [ X] Optimization. Including functional usability improvements and performance improvements.
- [ ] Other.

## Checklist

## Further comments

### DIFF
--- a/be/src/olap/tuple_reader.cpp
+++ b/be/src/olap/tuple_reader.cpp
@@ -80,8 +80,15 @@ OLAPStatus TupleReader::init(const ReaderParams& read_params) {
     if (status != OLAP_SUCCESS) { return status; }
 
     if (_optimize_for_single_rowset(rs_readers)) {
-        _next_row_func = _tablet->keys_type() == AGG_KEYS ? &TupleReader::_direct_agg_key_next_row
-                                                          : &TupleReader::_direct_next_row;
+        if (_tablet->keys_type() == AGG_KEYS) {
+            if (_aggregation) {
+                _next_row_func = &Reader::_agg_key_next_row;
+            } else {
+                _next_row_func = &Reader::_direct_agg_key_next_row;
+            }
+        } else {
+            _next_row_func = &Reader::_direct_next_row;
+        }
         return OLAP_SUCCESS;
     }
 
--- a/be/src/olap/tuple_reader.cpp
+++ b/be/src/olap/tuple_reader.cpp
@@ -84,7 +84,7 @@ OLAPStatus TupleReader::init(const ReaderParams& read_params) {
             if (_aggregation) {
                 _next_row_func = &TupleReader::_agg_key_next_row;
             } else {
-                _next_row_func = &Reader::_direct_agg_key_next_row;
+                _next_row_func = &TupleReader::_direct_agg_key_next_row;
             }
         } else {
             _next_row_func = &Reader::_direct_next_row;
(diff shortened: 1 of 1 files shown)


be/src/olap/tuple_reader.cpp
```diff
@@ -80,8 +80,15 @@ OLAPStatus TupleReader::init(const ReaderParams& read_params) {
     if (status != OLAP_SUCCESS) { return status; }
 
     if (_optimize_for_single_rowset(rs_readers)) {
-        _next_row_func = _tablet->keys_type() == AGG_KEYS ? &TupleReader::_direct_agg_key_next_row
-                                                          : &TupleReader::_direct_next_row;
+        if (_tablet->keys_type() == AGG_KEYS) {
+            if (_aggregation) {
+                _next_row_func = &Reader::_agg_key_next_row;
```
jackwener, line 85:
```suggestion
                _next_row_func = &TupleReader::_agg_key_next_row;
```

be/src/olap/tuple_reader.cpp
```diff
@@ -80,8 +80,15 @@ OLAPStatus TupleReader::init(const ReaderParams& read_params) {
...
+            } else {
+                _next_row_func = &Reader::_direct_agg_key_next_row;
```
jackwener, line 87:
```suggestion
                _next_row_func = &TupleReader::_direct_agg_key_next_row;
```

be/src/olap/tuple_reader.cpp
jackwener, line 85:
```suggestion
                _next_row_func = &TupleReader::_agg_key_next_row;
```

be/src/olap/tuple_reader.cpp
```diff
@@ -80,8 +80,15 @@ OLAPStatus TupleReader::init(const ReaderParams& read_params) {
...
+            }
+        } else {
+            _next_row_func = &Reader::_direct_next_row;
```
jackwener, line 90:
```suggestion
            _next_row_func = &TupleReader::_direct_next_row;
```

jackwener (commented):
why not use `TupleReader`?

The build is failed.

GoGoWen:
> why not use `TupleReader`?
> 
> The build is failed.

Hi jackwener,   thanks for your reply.   It should be TupleReader.  Changed!

morningman:
Hi, @GoGoWen , Sorry for the late reply, I've rethought the question.

Essentially, the purpose of the `_aggregation` parameter is to force the storage layer to return data according to the **aggregation semantics of the table itself** when the **aggregation semantics in the original SQL** do not match the **aggregation semantics of the table itself**.

In the case of single rowset, we can use `_direct_agg_key_next_row` to return the data directly because we can confirm that the data has been aggregated according to the table's own aggregation semantics.

The optimization you want to do is to use the aggregation function at the storage layer to reduce the amount of data returned from the storage layer in the single rowset case. But in essence, the amount of computation is the same when aggregation is done at the storage layer and aggregation is done at the query layer (there is no network overhead, it's all done locally within the same process).

And the parameter `_aggregation` is used to determine whether to optimize or not. However, the `_aggregation` parameter is not essentially used for this purpose, which may make the code less readable.

So have you come across a specific case and found that optimizing this way can significantly improve the efficiency of the query?

Issue #7525, opened 2021-12-29
[Bug] the session variable changed back after being globally set


morningman:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Version

0.15

### What's Wrong?

1. origin `exec_mem_limit` is 1234.
1. `set global exec_mem_limit=5678;`
2. reconnect to Doris, and sometime you will see that the `exec_mem_limit` changed back to `1234`

### What You Expected?

Not changed

### How to Reproduce?

Not easy to trigger.

1. `set global exec_mem_limit=5678;`
2. edit log roll.
3. `set global exec_mem_limit=1234;`
4. checkpoint triggered.
5. `exec_mem_limit` changed back to 5678

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #7526, opened 2021-12-29
[fix](session-variable) fix bug that checkpoint may overwrite the global variables

morningman:
## Proposed changes

Close related #7525 (replace it with issue number if it exists).

We should create temporary object for some static fields when doing checkpoint,
to avoid these variables to be overwritten by the checkpoint process.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/VariableMgr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/VariableMgr.java
@@ -28,7 +28,6 @@ import org.apache.doris.common.ErrorCode;
 import org.apache.doris.common.ErrorReport;
 import org.apache.doris.common.FeMetaVersion;
 import org.apache.doris.common.PatternMatcher;
-import org.apache.doris.persist.EditLog;
 import org.apache.doris.persist.GlobalVarPersistInfo;
 
 import com.google.common.collect.ImmutableMap;
@@ -38,6 +37,7 @@ import com.google.common.collect.Lists;
 import org.apache.commons.lang.SerializationUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.jetbrains.annotations.NotNull;
 import org.json.JSONObject;
 
 import java.io.DataInputStream;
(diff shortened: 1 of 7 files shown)


fe/fe-core/src/main/java/org/apache/doris/qe/VariableMgr.java
```diff
@@ -304,17 +286,30 @@ public static void setLowerCaseTableNames(int mode) throws DdlException {
 
     // global variable persistence
     public static void write(DataOutputStream out) throws IOException {
-        defaultSessionVariable.write(out);
+        SessionVariable variablesToWrite = defaultSessionVariable;
+        if (Catalog.isCheckpointThread()) {
```
xy720, line 290: There should not be non-checkpoint-thread calling this method. Why not directly use defaultSessionVariableForCkpt instead.

fe/fe-core/src/main/java/org/apache/doris/qe/VariableMgr.java
```diff
@@ -304,17 +286,30 @@ public static void setLowerCaseTableNames(int mode) throws DdlException {
...
+        GlobalVarPersistInfo info = new GlobalVarPersistInfo(variablesToWrite, varNames);
         info.write(out);
     }
 
     public static void read(DataInputStream in) throws IOException, DdlException {
         wlock.lock();
         try {
-            defaultSessionVariable.readFields(in);
+            SessionVariable variablesToRead = defaultSessionVariable;
+            if (Catalog.isCheckpointThread()) {
```
xy720, line 306: The same as above

fe/fe-core/src/main/java/org/apache/doris/qe/VariableMgr.java
morningman: Not exactly, the `write` method may be called in:
1. Checkpoint thread
2. when calling `/dump` http api

fe/fe-core/src/main/java/org/apache/doris/qe/VariableMgr.java
morningman: No, the `read` method will be called:
1. when FE start
2. checkpoint thread

xy720 (approved):
LGTM

PR #7529, opened 2021-12-29
Support remote storage, step2, only for be: hot data trans to cold data. clean cold data when drop table

### LINKED ISSUE #7097: [Feature] Support storage of remote cluster(BOS/S3) for doris data.
### Search before asking

### Description

Now, doris only store data in local disk, it makes you can read and write data on disk quickly. But not all data in database is read/written usually, most data is used when it is a new one. When the data is not hot, it will still cost the space of the disk.You can delete it, however some data maybe useful again some time.

So, the cold data need to be saved on some cheaper storage, such as BOS/S3/HDFS, etc. It will be cheaper.

Then the cold data can also be read when it is necessary, just from remote storage.

### Overall

1. Support remote storage, data will be move to remote storage(BOS/S3) when it is cold.
2. Dynamic partition need to be set to cold continuously by the create time, so we can set them cold continuously.
3. Meta need to be local, so we can read it quickly. Then read data by the meta.
4. When cold data need to be read, get it from remote storage.
5. remote storage need to be similar to local storage, cold data can be read, moved to trash and deleted, but cant't be appended.

### Detail design
BE will resovle the relation of local disk and remote storage.
Local disk will hold the meta, which will be used to find which data is needed.
Remote storage will hold the cold data, which will be read by be.
```
                                     FE
                                      |
                                     BE
                          |                        |
                         META                     DATA
                      LOCAL DISK              REMOTE STORAGE
```
1. Support remote storage
remote storage configure will be set in the properties of Create/Alter Table
a. storage_medium is the storage for hot data.
b. storage_cold_medium is the destination storage which cold data will be moved to.
c. storage_cooldown_time is the time for cold data.
```
CREATE TABLE TblPxy
(
    aa BIGINT
)
ENGINE=olap
DISTRIBUTED BY HASH (aa) BUCKETS 32
PROPERTIES(
    "storage_medium" = "SSD",
    "storage_cold_medium" = "S3",
    "storage_cooldown_time" = "2021-11-08 11:52:00"
);
```
2. Dynamic partition cold data
Dynamic partition is created continuously, so the cold time must be set by the partition time.
a. dynamic_partition.hot_partition_num means how many hot partition will relay, the older partition will be set to cold.
b. dynamic_partition.storage_medium is the storage holding hot data.
c. dynamic_partition.storage_cold_medium is the dest storage for cold data.
```
CREATE TABLE TblPxy (
    k1 DATE,
    aa BIGINT
) ENGINE=olap PARTITION BY RANGE (k1) ()
DISTRIBUTED BY HASH (aa) BUCKETS 1
PROPERTIES(
    "dynamic_partition.hot_partition_num" = "3",
    "dynamic_partition.storage_medium" = "HDD",
    "dynamic_partition.storage_cold_medium" = "S3",
    "dynamic_partition.time_unit" = "DAY",
    "dynamic_partition.start" = "-3",
    "dynamic_partition.end" = "3",
    "dynamic_partition.prefix" = "p",
    "dynamic_partition.buckets" = "32"
);
```
3. Read cold data, meta will be local
When you are calling select and the data is cold. BE will get meta of local disck first, choose which data is needed.
Then the matched remote data will be read and return to BE.
```
SELECT * FROM TblPxy;
```
4. Cold data trash
When cold data need to be dropped, move it to trash path on remote storage, and the trash path will be set in local trash path.
Cleaner will check local trash path, if it's time to delete, remote data will be deleted first, and then local.

### Code of Conduct

pengxiangyu:
## Proposed changes

1. When the hot data need to be changed to cold data, MigrationHandler will create a new Job of SchemaChangeV2, call the CreateReplicaTask and AlterReplicaTask. It will do something:
    1.1 FE create a new shadow tablet with remote path, send this request to BE.(MigrationHandler.java)
    1.2 BE will create local cache dir for remote path, upload data from local to s3. (schema_change.cpp  beta_rowset.cpp)
    1.3 Write tablet_uid(used for remote path) in cache path.  (schema_change.cpp) tablet_uid will be used to create the remote path on S3. When the remote path need to be deleted, tablet_uid is useful.
    1.4 update meta to olap meta.  (schema_change.cpp)
2. hot data to cold data, for FE:
    2.1 Add StorageColdMedium for DataProperty. change every DataProperty() function.
    2.2 Add Type S3 to Cold Storage. (PropertyAnalyzer.java)
3. Whe the cold data need to be read:
    3.1 RemoteBlockManager will be used, all the operations will be done in it.
    3.2 When a select operation arrived, it will be send to RemoteBlockManager, download the files from remote path on S3 to cache path, read the files in cache path and return.
5. When the cold data need to be dropped:
    3.1 move remote data to trash path on s3.（data_dir.cpp  move_to_trash()）
    3.2 move local data to trash path on local disk.
    3.3 delete files in trash dir on remote and local.（storage_engine.cpp start_trash_sweep()   _do_sweep())

## Types of changes

- [ ] Other.

## Checklist

- [x] I have created an issue on (Fix #7097) and described the bug/feature there in detail

## Further comments

After this patch, get remote data when select is called.


morningman:
link to #7575

yiguolei:
I have two questions:
1. Shoud  set the partition to freeze state to avoid insert data to cold partitions? 
2. How to deal with schema change for the data in S3?

morningman:
Please update the PR comment to describe the new implementation

pengxiangyu:
> I have two questions:
> 
> 1. Shoud  set the partition to freeze state to avoid insert data to cold partitions?
> 2. How to deal with schema change for the data in S3?

1. partition need not to freeze, when migration, old tablet and new shadow tablet will both be inserted new data. Just like schema_change.
2. data in S3 is cold data, this type of data is not writable, schema_change is not supported. If cold data is in a partition, the partition can't be writable, but partition of hot data is writable.

fe/fe-core/src/main/java/org/apache/doris/alter/MigrationHandler.java
```diff
@@ -0,0 +1,420 @@
...
+    public int cycle_count = 0;
+
+    public MigrationHandler() {
+        super("migration", Config.default_schema_change_scheduler_interval_millisecond);
+    }
+
+    private void createJob(long dbId, OlapTable olapTable, Collection<Long> partitionIds,
+                           TStorageMedium destStorageMedium) throws UserException {
+        if (olapTable.getState() == OlapTableState.ROLLUP) {
+            throw new DdlException("Table[" + olapTable.getName() + "]'s is doing ROLLUP job");
```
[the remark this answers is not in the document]
pengxiangyu: DONE

fe/fe-core/src/main/java/org/apache/doris/alter/MigrationHandler.java
```diff
@@ -0,0 +1,420 @@
...
+
+            for (Long tableId : tableIdToStorageMedium.keySet()) {
+                Table table = db.getTableNullable(tableId);
+                if (table == null) {
+                    continue;
+                }
+                OlapTable olapTable = (OlapTable) table;
+                // use try lock to avoid blocking a long time.
+                // if block too long, backend report rpc will timeout.
+                if (!olapTable.tryWriteLock(Table.TRY_LOCK_TIMEOUT_MS, TimeUnit.MILLISECONDS)) {
```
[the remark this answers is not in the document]
pengxiangyu: DONE

be/src/olap/rowset/beta_rowset.h
```diff
@@ -53,9 +53,10 @@ class BetaRowset : public Rowset {
 
     OLAPStatus link_files_to(const FilePathDesc& dir_desc, RowsetId new_rowset_id) override;
 
-    OLAPStatus copy_files_to(const std::string& dir) override;
+    OLAPStatus copy_files_to(const std::string& dir, RowsetId new_rowset_id) override;
 
-    OLAPStatus upload_files_to(const FilePathDesc& dir_desc) override;
+    OLAPStatus upload_files_to(const FilePathDesc& dir_desc,
```
yiguolei, line 58: why need this api?

be/src/olap/rowset/alpha_rowset.cpp
```diff
@@ -110,7 +110,7 @@ OLAPStatus AlphaRowset::link_files_to(const FilePathDesc& dir_desc, RowsetId new
     return OLAP_SUCCESS;
```
yiguolei, line 110: not support alpha rowset any more, just return error and stop migration job if this rowset is alpharowset

be/src/olap/base_tablet.cpp
```diff
@@ -28,9 +29,10 @@ extern MetricPrototype METRIC_query_scan_bytes;
 extern MetricPrototype METRIC_query_scan_rows;
 extern MetricPrototype METRIC_query_scan_count;
 
-BaseTablet::BaseTablet(TabletMetaSharedPtr tablet_meta, DataDir* data_dir)
+BaseTablet::BaseTablet(TabletMetaSharedPtr tablet_meta, const StorageParamPB& storage_param, DataDir* data_dir)
```
yiguolei, line 32: Not add a new parameter, StorageParam should be part of TabletMeta

be/src/olap/storage_engine.cpp
```diff
@@ -776,18 +783,18 @@ void StorageEngine::_clean_unused_txns() {
     }
 }
 
-OLAPStatus StorageEngine::_do_sweep(const string& scan_root, const time_t& local_now,
+OLAPStatus StorageEngine::_do_sweep(const FilePathDesc& scan_root_desc, const time_t& local_now,
                                     const int32_t expire) {
```
yiguolei, line 787: Cold data will have no trash or garbage, not deal with it.

be/src/olap/tablet_manager.cpp
```diff
@@ -232,6 +236,19 @@ OLAPStatus TabletManager::create_tablet(const TCreateTabletReq& request,
...
               << ", schema_hash=" << schema_hash;
 
+    if (FilePathDesc::is_remote(request.storage_medium)) {
+        FilePathDesc path_desc;
+        path_desc.storage_medium = request.storage_medium;
+        path_desc.storage_name = request.storage_param.storage_name;
+        std::shared_ptr<Env> env = Env::get_env(path_desc);
+        if (env == nullptr) {
+            LOG(INFO) << "remote storage is not exist, create it. storage_name: " << request.storage_param.storage_name;
+            RETURN_WITH_WARN_IF_ERROR(Env::get_remote_mgr()->create_remote_storage(
```
yiguolei, line 246: Remote Storage Object should be shared, do not create remote storage object for every tablet. For example, we may monitor the remote storage's performance, if there are too many object, it is too hard.

be/src/olap/tablet_manager.cpp
```diff
@@ -1091,7 +1131,48 @@ void TabletManager::try_delete_unused_tablet_path(DataDir* data_dir, TTabletId t
     // TODO(ygl): may do other checks in the future
     if (Env::Default()->path_exists(schema_hash_path).ok()) {
         LOG(INFO) << "start to move tablet to trash. tablet_path = " << schema_hash_path;
-        OLAPStatus rm_st = move_to_trash(schema_hash_path, schema_hash_path);
+        FilePathDesc segment_desc(schema_hash_path);
```
yiguolei, line 1134: Tablet GC could not be performance on any BE, because it does not know when to delete data.
It should be performed on FE. Be only care about its local data not remote data.

yiguolei:
I have reviewed the code, it is too complicated, I should be simplified.
1. I find that you reused the schema change logic. It's ok to reuse the logic but not reuse the code, we should not modify schema change code since it is very serious. We should rewrite a new job very like schema change.
2. I think if user set the partition to COLD, then it should be freezed, not allow to insert new data. I think this is acceptable. But I think It is very important to allow SCHEMA CHANGE!!! because schema change is table level.
3. If the tablet is a new tablet and it is not writeable, or do compaction, then there should be no rubbish or garbage, so that we do not need to care about gc or sweep logic.
4. Remote Storage Medium is set to tablet level is ok. But not a separate parameter, it should be part of table meta.
5. We should write clone logic on FE maybe it is just a create tablet logic on BE, BE just load the header file from remote storage.
6. There are also many other logics like backup, restore, snapshot logic. Maybe we could disable some feature at current step.

yiguolei:
I also think Env is not used properly. There are many code like if Env is xxx Env then do something. I think we could bind tablet to specific Env, then not check env type and just call env object's method directly.

yiguolei:
Some method in Env is useless, for example `new_sequential_file` it is only used in 

`Status get_thread_stats(int64_t tid, ThreadStats* stats) {
    DCHECK(stats != nullptr);
    if (kTicksPerSec <= 0) {
        return Status::NotSupported("ThreadStats not supported");
    }
    faststring buf;
    RETURN_IF_ERROR(env_util::read_file_to_string(
            Env::Default(), strings::Substitute("/proc/self/task/$0/stat", tid), &buf));

    return parse_stat(buf.ToString(), nullptr, stats);
}`

It does not make sense to be part of Env. What about just use posix api do to this and simplify Env Interface.

be/src/olap/tablet_manager.cpp
pengxiangyu: Every remote data has a local meta, When the local meta is deleted, remote data will be deleted too.
If deleted by fe, when a migration operation is failed and did not notify fe, how can fe know it？

be/src/olap/tablet_manager.cpp
pengxiangyu: Remote storage is shared in RemoteEnvMgr. Env::get_env() will get the shared_ptr from RemoteEnvMgr, path_desc will only take the storage_name and medium type.

be/src/olap/storage_engine.cpp
pengxiangyu: It may be deleted later, but I think is safer when migration operation is online for the first time.

morningman:
This PR is closed, please refer to #8663

PR #7544, opened 2021-12-30
[refactor & fix](exce & olap) refactor reader: rename Reader to TabletReader

zuochunwei:
## Proposed changes
1. Consider the responsibility of Reader,  Rename Reader to TabletReader, I think the new name TabletReader can represent its function exactly,  it is more suitable and meaningful
2. add virtual keyword for the destructor of OlapScanner, because VOlapScanner is derived from it
3. refactor struct ReaderParams and KeysParam as TabletReader's inner struct，guard by TabletReader name scope, it's also more reasonable
4. reduce OlapScanner's member data amount, just use _parent->member_data is simpler
5. bugfix: TupleReader has the same memeber data _collect_iter to its parent class Reader, this usage is dangerous, the writer may make some mistake, so i delete TupleReader::_collect_iter to fix it.
6. call set_tablet_reader() in OlapScanner::prepare() to setup _tablet_reader, VOlapScanner should override set_tablet_reader to new BlockReader instead,  use this way to avoid new Reader twice by reset unique_ptr _tablet_reader
7. if the member data is a inseparable part of a class, i suggest using normal variable while not pointer variable, because pointer bring a indirect lay and must handle coping and destructing carefully, it's not necessary
8. some other small changes for readbility or design

Close related #issue (replace it with issue number if it exists).

## Types of changes

- [ x] Bugfix (non-breaking change which fixes an issue)
- [ x] Code refactor (Modify the code structure, format the code, etc...)
- [ ] Other.

## Checklist

## Further comments


wangshuo128:
Maybe the filename of reader.h and reader.cpp should be renamed to table_reader.h and table_reader.cpp as well.

zuochunwei:
> 

yes，but rename file will result in massive changes for files included reader.h, so let's do it next time!

HappenLee (approved):
LGTM

PR #7558, opened 2021-12-31
[refactor](olap) prefer to malloc for _variable_buf once rather than malloc repeatedly

zuochunwei:
## Proposed changes

1. calc the total buf len before mallocing in the function _alloc_buf()
2. delete unnecessary member data such as _long_text_buf
3. use std::iota() to set std::vector<int> columns  consistently

I have completed the comparing test, result shows that using a whole buf is quicker

Close related #issue (replace it with issue number if it exists).

## Types of changes

- [ x] Code refactor (Modify the code structure, format the code, etc...)
- [ ] Other.

## Checklist

## Further comments

### DIFF
--- a/be/src/olap/row_cursor.cpp
+++ b/be/src/olap/row_cursor.cpp
@@ -28,19 +28,6 @@ using std::string;
 using std::vector;
 
 namespace doris {
-RowCursor::RowCursor()
-        : _fixed_len(0), _variable_len(0), _string_field_count(0), _long_text_buf(nullptr) {}
-
-RowCursor::~RowCursor() {
-    delete[] _owned_fixed_buf;
-    delete[] _variable_buf;
-    if (_string_field_count > 0 && _long_text_buf != nullptr) {
-        for (int i = 0; i < _string_field_count; ++i) {
-            free(_long_text_buf[i]);
-        }
-        free(_long_text_buf);
-    }
-}
 
 OLAPStatus RowCursor::_init(const std::vector<uint32_t>& columns) {
     _variable_len = 0;
@@ -99,7 +86,7 @@ OLAPStatus RowCursor::_init_scan_key(const TabletSchema& schema,
     RETURN_NOT_OK(_alloc_buf());
     char* fixed_ptr = _fixed_buf;
     char* variable_ptr = _variable_buf;
-    char** long_text_ptr = _long_text_buf;
+    char** long_text_ptr = (char**)(_variable_buf + _variable_len);
     for (auto cid : _schema->column_ids()) {
         const TabletColumn& column = schema.column(cid);
         fixed_ptr = _fixed_buf + _schema->column_offset(cid);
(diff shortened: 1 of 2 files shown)


be/src/olap/row_cursor.h
```diff
@@ -36,10 +36,13 @@ class RowCursor {
...
     static const int DEFAULT_TEXT_LENGTH = 128;
 
-    RowCursor();
+    RowCursor() = default;
 
     // 遍历销毁field指针
-    ~RowCursor();
+    ~RowCursor() {
+        delete [] _owned_fixed_buf;
+        delete [] _variable_buf;
```
wangbo, line 44: If _variable_buf holds a ref of an Object such as Slice,can ```delete [] _variable_buf ``` trigger Slice's destructor.

zuochunwei: no, because the type of _variable_buf is char*, so delete[] _variable_buf will just release the memory allocated by corresponding new[] not matter whatever the _variable_buf is used for.

wangbo:
I think there is another optimization point for ```row_cursor```;
When only ```string type``` is needed,  it's clear that only ```_long_text_buf``` need to be allocated memory, ```_variable_buf``` won't be used, so in ```_alloc_buf()```, allocation for _variable_buf is not necessary;
I think the allocation for ```_long_text_buf``` and ```_variable_buf``` should be separated;
Of course, this optimization can be done later.
So I think ```_long_text_buf``` should not be removed;

be/src/olap/row_cursor.h
zuochunwei: i have appended comments to explain why i do this

wangbo:
I have another question,Is using a continuous memory to store some strings  ```cpu-cache-friendly```?
In current implementation, strings is kept in memory divided.

zuochunwei:
> I have another question,Is using a continuous memory to store some strings `cpu-cache-friendly`? In current implementation, strings is kept in memory divided.

reply the buf's size,  level1 cache size may 32/64k，this size may big enought

zuochunwei:
> 

i got your meaning

zuochunwei:
> I think there is another optimization point for `row_cursor`; When only `string type` is needed, it's clear that only `_long_text_buf` need to be allocated memory, `_variable_buf` won't be used, so in `_alloc_buf()`, allocation for _variable_buf is not necessary; I think the allocation for `_long_text_buf` and `_variable_buf` should be separated; Of course, this optimization can be done later. So I think `_long_text_buf` should not be removed;

if all fields are long_text_buf, the _variable_len calculcted by init function will be 0, so memory will not allocated for var

wangbo (approved):
LGTM

be/src/olap/row_cursor.cpp
```diff
@@ -28,19 +28,6 @@ using std::string;
...
 namespace doris {
-RowCursor::RowCursor()
-        : _fixed_len(0), _variable_len(0), _string_field_count(0), _long_text_buf(nullptr) {}
-
-RowCursor::~RowCursor() {
-    delete[] _owned_fixed_buf;
-    delete[] _variable_buf;
-    if (_string_field_count > 0 && _long_text_buf != nullptr) {
-        for (int i = 0; i < _string_field_count; ++i) {
-            free(_long_text_buf[i]);
```
HappenLee, line 39: the long_text_buf should release one by one. because string type will realloc the mem

zuochunwei: i see, realloc in agg_update

PR #7569, opened 2022-01-02
[Improvement] support build thirdparty and BE using clang

yiguolei:
1. Support build thirdparty and BE using clang
2. User could export CC and CXX env  to switch the compiler from clang to gcc
3. If user not set CC and CXX env, then it will use gcc from DORIS_GCC_HOME or the system's default gcc.

In the future, I think we should only support Clang.


morningman:
#7590

morningman:
I think we need to document to explain when and how to use clang or gcc.

be/src/olap/column_vector.h
```diff
@@ -51,6 +51,37 @@ class DataBuffer {
...
+            std::free(buf_old);
+        } else {
+            buf = reinterpret_cast<T*>(std::malloc(sizeof(T) * new_size));
+        }
+        current_capacity = new_size;
+    }
+    current_size = new_size;
+}
+
 template class DataBuffer<bool>;
```
amosbird, line 85: "explicit template instantiation" is definition and should be in source file (translation unit), or else in theory it's ODR violation. In practice it doesn't make sense to have them in header files (outline defined) which slows down compilation and leads to problems you're trying to work around.

PR #7631, opened 2022-01-06
[Fix] Support compiling by gcc 11.

biohazard4321:
Use gcc 11 could improve performance. But there's some compiling problems, i try to fix them.

"_serialize" is an gcc internal macro, so i removed the underscore


yiguolei:
I find some issues is being fixed in #7569

zuochunwei:
good job

morningman:
Resolved in #7712

PR #7754, opened 2022-01-13
[improvement](routine-load) Reduce the probability that the routine load task rpc timeout

### LINKED ISSUE #7755: [Enhancement] routine load task may failed due to RPC timeout
### Search before asking

### Description

When there are some routine load jobs and a large number of broker load jobs.
The routine load jobs may be failed due to RPC timeout like:

```
Reached Timeout@xxxxx
``

This is because we may perform some time-consuming operations inside the RPC,
such as flush memtable to reduce memory usage, etc.
And these operations may be affected by different load jobs, which eventually lead to rpc timeouts.

### Solution

_No response_

### Code of Conduct

morningman:
# Proposed changes

Issue Number: close #7755

## Problem Summary:

If an load task has a relatively short timeout, then we need to ensure that
each RPC of this task does not get blocked for a long time.
And an RPC is usually blocked for two reasons.

1. handling "memory exceeds limit" in the RPC
    
    If the system finds that the memory occupied by the load exceeds the threshold,
    it will select the load channel that occupies the most memory and flush the memtable in it.
    this operation is done in the RPC, which may be more time consuming.

2. close the load channel

    When the load channel receives the last batch, it will end the task.
    It will wait for all memtables flushes to finish synchronously. This process is also time consuming.

Therefore, this PR solves this problem by.

1. Use timeout to determine whether it is a high-priority load task

    If the timeout of an load task is relatively short, then we mark it as a high-priority task.

2. not processing "memory exceeds limit" for high priority tasks
3. use a separate flush thread to flush memtable for high priority tasks.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -515,6 +515,8 @@ CONF_mInt32(storage_flood_stage_usage_percent, "90"); // 90%
 CONF_mInt64(storage_flood_stage_left_capacity_bytes, "1073741824"); // 1GB
 // number of thread for flushing memtable per store
 CONF_Int32(flush_thread_num_per_store, "2");
+// number of thread for flushing memtable per store, for high priority load task
+CONF_Int32(high_priority_flush_thread_num_per_store, "1");
 
 // config for tablet meta checkpoint
 CONF_mInt32(tablet_meta_checkpoint_min_new_rowsets_num, "10");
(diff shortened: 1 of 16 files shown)


yangzhg (approved):
LGTM

PR #7773, opened 2022-01-16
[fix][chore](thrift) Fix warning when generate cpp code by thrift IDL file and use strict mode

caiconghui:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/gensrc/thrift/FrontendService.thrift
+++ b/gensrc/thrift/FrontendService.thrift
@@ -718,32 +718,32 @@ struct TWaitingTxnStatusResult {
 }
 
 service FrontendService {
-    TGetDbsResult getDbNames(1:TGetDbsParams params)
-    TGetTablesResult getTableNames(1:TGetTablesParams params)
-    TDescribeTableResult describeTable(1:TDescribeTableParams params)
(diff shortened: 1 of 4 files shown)


morningman (approved):
LGTM

yangzhg:
After this pr is merged, will the thrift data of different versions cannot be parsed during the upgrade or in the mixed distribution scenario?

caiconghui:
> After this pr is merged, will the thrift data of different versions cannot be parsed during the upgrade or in the mixed version cluster?

The only thing I'm not sure about is if List<byte> and binary are compatible
but according to https://stackoverflow.com/questions/40886279/apache-thrift-difference-between-byte-and-binary-types
It seems that it is ok

yangzhg:
Have you test this in a cluster that have new and old nodes ?

caiconghui:
> Have you test this in a cluster that have new and old nodes ?



> Have you test this in a cluster that have new and old nodes ?

follower with new version，master with older version，I test forward function, but not test list<byte>

PR #7788, opened 2022-01-17
[fix](bitmap-index) Fix bug that bitmap index may return wrong result.

### LINKED ISSUE #7765: [Bug] if the query based on a column that with index, the result  does not meet the expectation
### Search before asking

### Version

0.15.1-rc09

### What's Wrong?

A query sql like this: 

```sql
select count(*) from table_name where id>1234
```
If the `id` column with index, the query result  does not meet the expectation. 

### What You Expected?

got the right count number.

### How to Reproduce?

1. create a table, sql like this:
```sql
CREATE TABLE `table1` (
  `_i_create_time` datetime NULL COMMENT "",
  `_i_random_val` smallint(6) NULL COMMENT "",
  `_src_id` varchar(64) NULL COMMENT "",
  `name` varchar(128) NULL COMMENT "",
  `country` varchar(128) NULL COMMENT "",
  `city` varchar(128) NULL COMMENT "",
  `_id` bigint(20) NULL COMMENT "联系人id",
  `_create_time` datetime NULL COMMENT "",
  `_update_time` datetime NULL COMMENT "",
  `_id_init` bigint(20) NULL COMMENT "",
  INDEX idx__i_create_time (`_i_create_time`) USING BITMAP COMMENT '',
  INDEX idx__i_random_val (`_i_random_val`) USING BITMAP COMMENT '',
  INDEX idx__src_id (`_src_id`) USING BITMAP COMMENT '',
  INDEX idx_name (`name`) USING BITMAP COMMENT '',
  INDEX idx_country (`country`) USING BITMAP COMMENT '',
  INDEX idx_city (`city`) USING BITMAP COMMENT '',
  INDEX idx__id (`_id`) USING BITMAP COMMENT '',
  INDEX idx__create_time (`_create_time`) USING BITMAP COMMENT '',
  INDEX idx__update_time (`_update_time`) USING BITMAP COMMENT '',
  INDEX idx__id_init (`_id_init`) USING BITMAP COMMENT ''
) ENGINE=OLAP
DISTRIBUTED BY HASH(`_src_id`) BUCKETS 10
```

2. import 100 million data to the table. 

3. confirm the total without any conditions, 
```sql
select count(*) from table1
```
it will return the right number, 100000000

But, if the query based on `id` column , 
```sql
select count(*) from table_name where id>1234
```
the result  does not meet the expectation, and every time is different.

### Anything Else?

_No response_

### Code of Conduct

morningman:
# Proposed changes

Issue Number: close #7765

## Problem Summary:

Fix the following bugs.

1. `column1` created a bitmap index.
2. `column1` has a lot index items in the bitmap index, and the index page is divided into two levels.
3. `column1`'s value range is `[1000, 10000000]`.
4. the query condition is `column1 > 0`
5. the empty result will be returned, while the expected value should be 9999000 rows.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_v2/ordinal_page_index.h
+++ b/be/src/olap/rowset/segment_v2/ordinal_page_index.h
@@ -71,7 +71,7 @@ public:
 
     // the returned iter points to the largest element which is less than `ordinal`,
     // or points to the first element if all elements are greater than `ordinal`,
-    // or points to "end" if all elementss are smaller than `ordinal`.
+    // or points to "end" if all elements are smaller than `ordinal`.
     OrdinalPageIndexIterator seek_at_or_before(ordinal_t ordinal);
     inline OrdinalPageIndexIterator begin();
     inline OrdinalPageIndexIterator end();
--- a/be/src/olap/rowset/segment_v2/ordinal_page_index.h
+++ b/be/src/olap/rowset/segment_v2/ordinal_page_index.h
@@ -69,6 +69,9 @@ public:
     // load and parse the index page into memory
     Status load(bool use_page_cache, bool kept_in_memory);
 
+    // the returned iter points to the largest element which is less than `ordinal`,
+    // or points to the first element if all elements are greater than `ordinal`,
+    // or points to "end" if all elementss are smaller than `ordinal`.
     OrdinalPageIndexIterator seek_at_or_before(ordinal_t ordinal);
     inline OrdinalPageIndexIterator begin();
     inline OrdinalPageIndexIterator end();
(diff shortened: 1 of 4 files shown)


be/src/olap/rowset/segment_v2/ordinal_page_index.h
```diff
@@ -69,6 +69,9 @@ class OrdinalIndexReader {
     // load and parse the index page into memory
     Status load(bool use_page_cache, bool kept_in_memory);
 
+    // the returned iter points to the largest element which is less than `ordinal`,
+    // or points to the first element if all elements are greater than `ordinal`,
+    // or points to "end" if all elementss are smaller than `ordinal`.
```
yangzhg, line 74:
```suggestion
    // or points to "end" if all elements are smaller than `ordinal`.
```

yangzhg (approved):
LGTM

PR #7852, opened 2022-01-24
(refactor)(vectorized)refactor SegmentIterator::BitmapRangeIterator and SegmentIterator

zuochunwei:
# Proposed changes
1. BitmapRangeIterator refactor: using static buf replace new, improving BitmapRangeIterator::next_range 
2. SegmentIterator refactor: prefer using empty() to 'size() comparing with 0', delete some useless codes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_v2/segment_iterator.cpp
+++ b/be/src/olap/rowset/segment_v2/segment_iterator.cpp
@@ -46,13 +46,10 @@ namespace segment_v2 {
 //   output ranges: [0,2), [4,8), [10,11), [15,18), [18,20) (when max_range_size=3)
 class SegmentIterator::BitmapRangeIterator {
 public:
-    explicit BitmapRangeIterator(const roaring::Roaring& bitmap)
-            : _last_val(0), _buf(new uint32_t[256]), _buf_pos(0), _buf_size(0), _eof(false) {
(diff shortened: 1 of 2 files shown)


morningman (approved):
LGTM

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -872,13 +867,13 @@ Status SegmentIterator::next_batch(vectorized::Block* block) {
         RETURN_IF_ERROR(_init(true));
         _inited = true;
         if (!_vec_pred_column_ids.empty() || !_short_cir_pred_column_ids.empty()) {
-            _block_rowids.reserve(_opts.block_row_max);
+            _block_rowids.resize(_opts.block_row_max);
```
wangbo, line 870: When first to init SegmentIterator, _block_rowids needs to reserve;
Because _block_rowids is accessed by operator[] directly

zuochunwei: reverve just alloc memory for optimization reason.

there is one basic rule: any reserve calls can be removed, this removing operation will not influence the correctness of result.

according this rule, if we remove this _block_rowids.reserve(), the will crash, so we should use resize instead.

wangbo (approved):
LGTM

PR #7866, opened 2022-01-24
[fix](toolchain) Allow building when system contains libunwind

amosbird:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments

If the hosting machine contains libunwind, glog will introduce needed symbols. However, it doesn't handle static libraries properly, which in turn breaks other dependent libs, such as brpc. There is a related issue 
https://github.com/google/glog/issues/57 , however it's not fixed for more than 6 years. Let's simply disable using libunwind for glog.


amosbird:
Btw, this is a strong indication to hosting all thirdparty libraries as cmake modules. We should gradually adopt the `contrib` folder from ClickHouse.

morningman:
> Btw, this is a strong indication to hosting all thirdparty libraries as cmake modules. We should gradually adopt the `contrib` folder from ClickHouse.

This can be done after we graduated from Apache Incubator. So that we can use a new repo to save all contribs

morningman (approved):
LGTM

PR #7867, opened 2022-01-24
[build][scripts] add set -e to download_thirdparty.sh

ucasfl:
Let the build process stop immediately if some command execute faild. For example, patch failed due to command not found, otherwise it will stop until compile error.


morningman (approved):
LGTM

PR #7911, opened 2022-01-27
[Help] gcc11 compiles `thread_local` variable, BE start: version `GLIBC_2.18' not found

xinyiZzz:
## Problem Summary:

Error details in be.out:
````
/home/disk3/zxy/baidu/bdg/doris/core/output/be/lib/palo_be: /lib64/libc.so.6: version `GLIBC_2.18' not found (required by /home/disk3/zxy/ baidu/bdg/doris/core/output/be/lib/palo_be)
````

nm looks at palo_be and finds that the method that depends on GLIBC_2.18 is `__cxa_thread_atexit_impl`
3be717fca2b8563d447aa1e1c8f1f39a

I found that the `__cxa_thread_atexit_impl` method was introduced by glibc to support thread_local variables
https://sourceware.org/glibc/wiki/Destructor%20support%20for%20thread_local%20variables

It was no problem to compile through gcc10 before

So, how can I avoid this problem, ask for help.

### DIFF
--- a/be/src/runtime/thread_status.h
+++ b/be/src/runtime/thread_status.h
@@ -1,0 +1,35 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#pragma once
+
+#include <string>
+
+namespace doris {
+
+class ThreadStatus {
+
+private:
+    // When `thread_local` is a constant int/bool, no problem.
+    // When `thread_local` is a class, if the member variable is `int`, there is no problem.
+    // When `thread_local` is a class, if the member variable is `string`, error: lib64/libc.so.6: version `GLIBC_2.18' not found
+    std::string _query_id; // error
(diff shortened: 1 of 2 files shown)


morningman:
PTAL @amosbird , if you have time, thanks

amosbird:
Is it for some future PRs with thread local?

morningman:
> Is it for some future PRs with thread local?

Yes, @xinyiZzz is developing memory tracking and management based on thread local.

amosbird:
I did some investigation. It's not easy to mask the symbol, which is from libc++abi https://libcxxabi.llvm.org/spec.html . It's also not safe to use the polyfill implementation from some libc++api directly. I'd suggest the following:

1. use clang toolchain
2. keep thread_local with trivial destructor if gcc is used.

xinyiZzz:
> I did some investigation. It's not easy to mask this symbol, which is meant to be added to the libc++abi. It's also not safe to use the polyfill implementation from some libc++api directly. I'd suggest the following:
> 
> 1. use clang toolchain
> 2. keep thread_local with trivial destructor if gcc is used.

Thanks for the answer~ @amosbird 
Compiling with clang is indeed fine.
Regarding the second point, do you mean changing the thread_local variable to a member variable of the class? For example, abstract a parent class to save this variable. or other meanings.

My pr: https://github.com/apache/incubator-doris/pull/7198
I want to track all memory requests during a thread lifetime via thread_local + TCMalloc hook, like Clickhouse does.

amosbird:
> Regarding the second point, do you mean changing the thread_local variable to a member variable of the class? For example, abstract a parent class to save this variable. or other meanings.

You can redesign the implementation of your thread_local structs to have trivial destructor. The definition is in https://en.cppreference.com/w/cpp/language/destructor .

xinyiZzz:
> > Regarding the second point, do you mean changing the thread_local variable to a member variable of the class? For example, abstract a parent class to save this variable. or other meanings.
> 
> You can redesign the implementation of your thread_local structs to have trivial destructor. The definition is in https://en.cppreference.com/w/cpp/language/destructor .

I think I understand trivial destructor,
For non-trivial classes in thread_local, such as std::string, I plan to store them as pointers.
To ensure that thread_local is trivial, these non-trivial pointers will uniformly call destructors elsewhere.

thanks again, very helpful @amosbird

PR #7951, opened 2022-02-07
add performance test framework for ci/cd

dataroaring:
This patch contributes a performance test framework to doris for ci/cd
by learning from ClickHouse's. Performance results is computed by
using clickhouse, maybe we should use doris instead.

# PerformanceTestFramework

Issue Number: close #xxx

## Add a performance test framework for ci/cd

This patch contributes a performance test framework to doris for ci/cd
by learning from ClickHouse's. Performance results is computed by
using clickhouse, maybe we should use doris instead.

## Checklist(Required)

## Further comments

### DIFF
--- a/tests/ci/README.md
+++ b/tests/ci/README.md
@@ -1,0 +1,34 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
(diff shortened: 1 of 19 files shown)


dataroaring:
Here is an example of result. https://doris-build.oss-cn-beijing.aliyuncs.com/doris-test-reports/1/192ec94f1541b1d0c053f3ba3f885ceca9a045a6/performance/output.7z

morningman:
Hi @dataroaring , thanks for you contribution.
This is a huge PR, I think it is better to discuss about it, or at least add document to explain how it works and how to use it.
Also take care about the license header, and if it is copied from other project, better announce it after the license header.

For example:

https://github.com/apache/incubator-doris/blob/8207883319b0ab4b5fae709098c65867298b4118/be/src/vec/core/block.cpp#L1-L20

dataroaring:
> Hi @dataroaring , thanks for you contribution. This is a huge PR, I think it is better to discuss about it, or at least add document to explain how it works and how to use it. Also take care about the license header, and if it is copied from other project, better announce it after the license header.
> 
> For example:
> 
> https://github.com/apache/incubator-doris/blob/8207883319b0ab4b5fae709098c65867298b4118/be/src/vec/core/block.cpp#L1-L20

done

PR #7963, opened 2022-02-07
[fix] fix ltrim result may incorrect in some case

yangzhg:
## Problem Summary:

fix ltrim result may incorrect in some case
according to https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html
Built-in Function: int __builtin_cl/tz (unsigned int x)
If x is 0, the result is undefined.
So we handle the case of 0 separately

this function return different between gcc and clang when x is 0

## Checklist(Required)

## Further comments


xinghuayu007:
please add some unit tests for case:  __builtin_cl/tz (0)

amosbird:
Similar fixes has been done in https://github.com/apache/incubator-doris/pull/7712 , which was caught by current unit tests. It's an indication that test coverage should be improved.

yangzhg:
> please add some unit tests for case: __builtin_cl/tz (0)

Unit test has already contained this situation

yangzhg:
> Similar fixes has been done in #7712 , which was caught by current unit tests. It's an indication that test coverage should be improved.

I see, this is in another function, we may merge the those implements together

yangzhg:
> Similar fixes has been done in #7712 , which was caught by current unit tests. It's an indication that test coverage should be improved.

Fix in #7712 will add an offset of 128 when mask is 0, but offset is 32 when mask is 0 in this pr, I not sure 128 is a correct value @amosbird ?

xinghuayu007:
LGTM

morningman (approved):
LGTM

PR #7995, opened 2022-02-09
[Improvement] print log foreground if not use --daemon to start fe

yiguolei:
Doris does not support start_fe.sh and print all logs to console. It maybe a bug.


morningman (approved):
LGTM

PR #8000, opened 2022-02-09
[fix](compatibility) Fix compatibility issue of PRowBatch and some tablet sink bugs

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

1. set both `tuple_offsets` and `new_tuple_offsets` in PRowBatch for compatibility
2. set FE config `repair_slow_replica` default to false
   Avoid impacting the load process after upgrading.
   Eg, if there are only 2 replicas, one is with high version count. After upgrade,
   that replica will be set to bad, so that the load process will be stopped
   because only 1 replica is alive.
3. Fix a bug that NodeChannel may be blocked at `close_wait()`
   Forget to set `add_batch_finish` flag after the last rpc finished.
4. Fix a NPE of RoutineLoadScheduler

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/exec/tablet_sink.cpp
+++ b/be/src/exec/tablet_sink.cpp
@@ -185,7 +185,7 @@ Status NodeChannel::open_wait() {
     _add_batch_closure = ReusableClosure<PTabletWriterAddBatchResult>::create();
     _add_batch_closure->addFailedHandler([this](bool is_last_rpc) {
         // If rpc failed, mark all tablets on this node channel as failed
-        _index_channel->mark_as_failed(this, _add_batch_closure->cntl.ErrorText(), -1);
+        _index_channel->mark_as_failed(this->node_id(), this->host(), _add_batch_closure->cntl.ErrorText(), -1);
         Status st = _index_channel->check_intolerable_failure();
         if (!st.ok()) {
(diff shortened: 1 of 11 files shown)


be/src/exec/tablet_sink.cpp
```diff
@@ -571,7 +577,7 @@ void IndexChannel::add_row(Tuple* tuple, int64_t tablet_id) {
         // if this node channel is already failed, this add_row will be skipped
         auto st = channel->add_row(tuple, tablet_id);
         if (!st.ok()) {
-            mark_as_failed(channel, st.get_error_msg(), tablet_id);
+            mark_as_failed(channel.get(), st.get_error_msg(), tablet_id);
```
yangzhg, line 580: Why not change `mark_as_failed` signature to accept shared_ptr ?

morningman: I think it is not necessary, and it will cause extra shared ptr copy.

yangzhg: But I don't think it's a good practice to mix smart pointers and raw pointers. When we use smart pointers, we already imply that object ownership is managed by smart pointers instead of accessing raw pointers.

yangzhg (dismissed):
LGTM

yangzhg (approved):
LGTM

PR #8008, opened 2022-02-10
[Refactor] remove plugin folder in be since it is useless 

yiguolei:
Remove plugin folder in be since it is useless  and it need fPIC tag to build and we will remove all fPIC tag in the future.


morningman (approved):
LGTM

PR #8026, opened 2022-02-11
[Refactor] Remove snapshot converter and unused Protobuf Definitions

yiguolei:
1. remove snapshot converter
2. remove unused protobuf definitions
3. move some macro as const variables


morningman (approved):
LGTM

PR #8042, opened 2022-02-14
[Improvement] check simd instructions before start

yiguolei:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

Sometimes BE is build on a machine with SIMD instruction suck as AVX2. But the BE binary will be copied to a machine without AVX2. It will crashed without any error message.

This PR will check the required SIMD instructions and print error messages during startup.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/service/doris_main.cpp
+++ b/be/src/service/doris_main.cpp
@@ -20,8 +20,12 @@
 #include <unistd.h>
 
 #include <condition_variable>
+#include <cstring>
+#include <errno.h>
 #include <mutex>
+#include <setjmp.h>
 #include <thread>
+#include <unistd.h>
 #include <unordered_map>
 
 #if defined(LEAK_SANITIZER)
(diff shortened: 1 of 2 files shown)


be/src/service/doris_main.cpp
```diff
@@ -74,6 +78,194 @@ static void thrift_output(const char* x) {
...
+        case InstructionFail::AVX512:
+            ret("AVX512");
+    }
+    __builtin_unreachable();
+}
+
+
+sigjmp_buf jmpbuf;
+
+[[noreturn]] void sig_ill_check_handler(int, siginfo_t *, void *)
```
HappenLee, line 128: the funciton is void， why need `noreturn`

be/src/service/doris_main.cpp
```diff
@@ -74,6 +78,194 @@ static void thrift_output(const char* x) {
...
+{
+    siglongjmp(jmpbuf, 1);
+}
+
+/// Check if necessary SSE extensions are available by trying to execute some sse instructions.
+/// If instruction is unavailable, SIGILL will be sent by kernel.
+void check_required_instructions_impl(volatile InstructionFail & fail)
```
HappenLee, line 135: why here need `volatile`, avoid instuction reoder?

be/src/service/doris_main.cpp
```diff
@@ -74,6 +78,194 @@ static void thrift_output(const char* x) {
...
+    AVX512 = 8
+};
+
+#define ARRAY_SIZE(a) (sizeof(a)/sizeof((a)[0]))
+
+auto instruction_fail_to_string(InstructionFail fail)
+{
+    switch (fail)
+    {
+#define ret(x) return std::make_tuple(STDERR_FILENO, x, ARRAY_SIZE(x) - 1)
```
HappenLee, line 102: Add `ARM_NEON` to Support detect arm SIMD

be/src/service/doris_main.cpp
```diff
@@ -74,6 +78,194 @@ static void thrift_output(const char* x) {
 
 } // namespace doris
 
+// These code is referenced from clickhouse
+// It is used to check the SIMD instructions
+enum class InstructionFail
+{
```
HappenLee, line 84: obey the code format `InstructionFail {

be/src/service/doris_main.cpp
```diff
@@ -74,6 +78,194 @@ static void thrift_output(const char* x) {
...
+    SSSE3 = 2,
+    SSE4_1 = 3,
+    SSE4_2 = 4,
+    POPCNT = 5,
+    AVX = 6,
+    AVX2 = 7,
+    AVX512 = 8
+};
+
+#define ARRAY_SIZE(a) (sizeof(a)/sizeof((a)[0]))
```
zuochunwei, line 96: replace this macro to common header is better!

be/src/service/doris_main.cpp
zuochunwei: 因为用了sigsetjmp和siglongjmp，在异步信号的环境下，使用volatile确保每次都是都从内存里读取fail变量的值，而不用寄存器里的缓存吧

be/src/service/doris_main.cpp
zuochunwei: noreturn主要用来抑制警告：即声明为非void返回类型，并非在所有路径都有return语句。

be/src/service/doris_main.cpp
```diff
@@ -74,6 +79,199 @@ static void thrift_output(const char* x) {
 
 } // namespace doris
 
+// These code is referenced from clickhouse
```
morningman, line 82: Can these code be moved to a separated file. like `util/instruction_utils.hpp`?

yiguolei: 1. These code is not utils, they will be only used in the main function and could not in other functions. There maybe some signal catching problem if running in other functions I have not test it.
2. These code should run at the beginning of the program so I add __attribute__(priority = 101) to checker. If other functions running before these code meet SIMD code and could not find the SSE instruction the program may crash we could not catch the signal.

be/src/service/doris_main.cpp
```diff
@@ -74,6 +79,199 @@ static void thrift_output(const char* x) {
...
+    }
+
+    volatile InstructionFail fail = InstructionFail::NONE;
+
+    if (sigsetjmp(jmpbuf, 1))
+    {
+        WRITE_ERROR("Instruction check fail. The CPU does not support ");
+        if (!std::apply(write_retry, instruction_fail_to_string(fail)))
+            _Exit(1);
+        WRITE_ERROR(" instruction set.\n");
```
morningman, line 250: Add following to guide the user:
```
WRITE_ERROR("For AVX2 support, you need to rebuild the Doris BE with: ");
WRITE_ERROR("USE_AVX2=0 sh build.sh --be\n");
```

be/src/service/doris_main.cpp
morningman: Ok I see.
I've checked and it works.

morningman (approved):
LGTM

PR #8109, opened 2022-02-17
[fix](storage) Fix the bug of coredump when vec exec engine with delete condition

### LINKED ISSUE #8104: [Bug] [Vectorized] Core dump with query table have a `delete condition`
### Search before asking

### Version

vectorized

### What's Wrong?

vec block size is different with RowBlock v2 size， the mismatch cause the problem

### What You Expected?

query normal

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

HappenLee:
# Proposed changes

Issue Number: close #8104 

## Problem Summary:

Fix coredump

## Checklist(Required)

4. Does it need to update dependencies: （No)

## Further comments

### DIFF
--- a/be/src/olap/row_block2.cpp
+++ b/be/src/olap/row_block2.cpp
@@ -329,7 +329,8 @@ Status RowBlockV2::_copy_data_to_column(int cid, doris::vectorized::MutableColum
 }
 
 Status RowBlockV2::convert_to_vec_block(vectorized::Block* block) {
-    for (int i = 0; i < _schema.column_ids().size(); ++i) {
+    DCHECK_LE(block->columns(), _schema.column_ids().size());
+    for (int i = 0; i < block->columns(); ++i) {
         auto cid = _schema.column_ids()[i];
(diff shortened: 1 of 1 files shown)


be/src/olap/row_block2.cpp
```diff
@@ -329,7 +329,8 @@ Status RowBlockV2::_copy_data_to_column(int cid, doris::vectorized::MutableColum
 }
 
 Status RowBlockV2::convert_to_vec_block(vectorized::Block* block) {
-    for (int i = 0; i < _schema.column_ids().size(); ++i) {
+    DCHECK_LE(block->columns(), _schema.column_ids().size());
+    for (int i = 0; i < block->columns(); ++i) {
```
wangbo, line 333: Maybe using ```schema.num_column_ids()```  instead of  ```block->columns()``` is better?

HappenLee: vec block column size is different with RowBlockV2 size, the mismatch cause the problem

wangbo (approved):
LGTM

PR #8112, opened 2022-02-17
[fix](load) Fix InsertStmt prepareExpressions

### LINKED ISSUE #8111: [Bug] insert into cause error result
### Search before asking

### Version

trunk

### What's Wrong?

Create the table:
```
CREATE TABLE test.result_exprs (
  `aid` int(11) NULL,
  `bid` int(11) NULL
) ENGINE=OLAP
DUPLICATE KEY(`aid`)
COMMENT "OLAP"
DISTRIBUTED BY HASH(`aid`) BUCKETS 7
PROPERTIES (
"replication_num" = "1",
"in_memory" = "false",
"business_key_column_name" = "",
"storage_medium" = "HDD",
"storage_format" = "V2"
);
```

insert into data:
```
INSERT INTO result_exprs
SELECT a.aid,
       b.bid
FROM
  (SELECT 3 AS aid)a
RIGHT JOIN
  (SELECT 4 AS bid)b ON (a.aid=b.bid);
```
and then query the result:
```
select * from result_exprs;
```

the result is:
image

### What You Expected?

image

### How to Reproduce?

As shown above.

### Anything Else?

I think the insert plan is wrong, It uses 3 instand of the right exprs.

image

Maybe this is right:
image

When Doris generates resultExprs in InsertStmt, I think using `queryStmt.getResultExprs()` is more resonable than `queryStmt.getBaseTblResultExprs()`.

### Code of Conduct

arthuryangcs:
Use queryStmt.getResultExprs() instead of queryStmt.getBaseTblResultExprs() in InsertStmt prepareExpressions func.

# Proposed changes

Issue Number: close #8111

## Problem Summary:

Use queryStmt.getResultExprs() instead of queryStmt.getBaseTblResultExprs() in InsertStmt prepareExpressions func.

## Checklist(Required)

1. Does it affect the original behavior: Yes
2. Has unit tests been added:Yes
4. Does it need to update dependencies:  No

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/InsertStmt.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/InsertStmt.java
@@ -667,7 +667,7 @@ public class InsertStmt extends DdlStmt {
     }
 
     public void prepareExpressions() throws UserException {
-        List<Expr> selectList = Expr.cloneList(queryStmt.getBaseTblResultExprs());
+        List<Expr> selectList = Expr.cloneList(queryStmt.getResultExprs());
         // check type compatibility
         int numCols = targetColumns.size();
         for (int i = 0; i < numCols; ++i) {
--- a/fe/fe-core/src/main/java/org/apache/doris/common/profile/PlanTreeBuilder.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/profile/PlanTreeBuilder.java
@@ -56,7 +56,12 @@ public class PlanTreeBuilder {
             PlanTreeNode sinkNode = null;
             if (sink != null) {
                 StringBuilder sb = new StringBuilder();
-                sb.append("[").append(sink.getExchNodeId().asInt()).append(": ").append(sink.getClass().getSimpleName()).append("]");
+                if (sink.getExchNodeId() != null) {
+                    sb.append("[").append(sink.getExchNodeId().asInt()).append(": ")
+                            .append(sink.getClass().getSimpleName()).append("]");
+                } else {
+                    sb.append("[").append(sink.getClass().getSimpleName()).append("]");
+                }
                 sb.append("\n[Fragment: ").append(fragment.getId().asInt()).append("]");
                 sb.append("\n").append(sink.getExplainString("", TExplainLevel.BRIEF));
                 sinkNode = new PlanTreeNode(sink.getExchNodeId(), sb.toString());
--- a/fe/fe-core/src/test/java/org/apache/doris/planner/QueryPlanTest.java
+++ b/fe/fe-core/src/test/java/org/apache/doris/planner/QueryPlanTest.java
@@ -433,6 +433,20 @@ public class QueryPlanTest {
                 "\"replication_num\" = \"1\"" +
                 ");");
 
+        createTable("CREATE TABLE test.result_exprs (\n" +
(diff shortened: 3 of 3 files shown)


arthuryangcs:
@weizuo93 PTAL, THX~

fe/fe-core/src/main/java/org/apache/doris/analysis/InsertStmt.java
```diff
@@ -667,7 +667,7 @@ private void analyzePlanHints(Analyzer analyzer) throws AnalysisException {
     }
 
     public void prepareExpressions() throws UserException {
-        List<Expr> selectList = Expr.cloneList(queryStmt.getBaseTblResultExprs());
+        List<Expr> selectList = Expr.cloneList(queryStmt.getResultExprs());
```
yangzhg, line 670: can you add some comment to explain why ?

arthuryangcs: I'm not sure this is right, but this can solve the issues :)
I don't know why using getBaseTblResultExprs rather than getResultExprs there, are there some other considerations?

arthuryangcs: I think that generating InserStmt's resultExprs from queryStmt's resultExprs is more reasonable, or maybe I make a mistake.

fe/fe-core/src/main/java/org/apache/doris/common/profile/PlanTreeBuilder.java
```diff
@@ -56,7 +56,11 @@ private void buildFragmentPlans() {
             PlanTreeNode sinkNode = null;
             if (sink != null) {
                 StringBuilder sb = new StringBuilder();
-                sb.append("[").append(sink.getExchNodeId().asInt()).append(": ").append(sink.getClass().getSimpleName()).append("]");
+                if(sink.getExchNodeId() != null) {
```
EmmyMiao87, line 59: Pay attention to the code format ~

fe/fe-core/src/main/java/org/apache/doris/analysis/InsertStmt.java
EmmyMiao87: The ```ResultExprs``` is the finally expr of querystmt.
So, this change is correct.

fe/fe-core/src/main/java/org/apache/doris/common/profile/PlanTreeBuilder.java
arthuryangcs: done, PTAL

fe/fe-core/src/main/java/org/apache/doris/common/profile/PlanTreeBuilder.java
```diff
@@ -17,6 +17,7 @@
 
 package org.apache.doris.common.profile;
 
+import com.clearspring.analytics.util.Lists;
```
EmmyMiao87, line 20: Error import ~

arthuryangcs: fixed~

EmmyMiao87 (approved):
LGTM

EmmyMiao87:
Please waiting for the P0 test and unit test ~

PR #8118, opened 2022-02-17
[fix](rowset-meta) Fix bug that rowset meta is not deleted

### LINKED ISSUE #8120: [Bug] Too many invalid rowset meta left in rocksdb
### Search before asking

### Version

0.15

### What's Wrong?

In some cases, a large number of logs appear in be.INFO when starting BE:

```
could not find tablet id xxx, schema hash xxxx, for rowset xxxx. skip the rowset
```

Sometimes the number can reach a million.

### What You Expected?

There should not be so many rowset meta left in rocksdb that do not find the corresponding tablet.

### How to Reproduce?

A easy way to reproduce is to:
1. disable compaction
2. create a table
3. insert many times
4. drop table xx force()

restart BE, you may see those log.

### Anything Else?

_No response_

### Code of Conduct

morningman:
# Proposed changes

Issue Number: close #8120 

## Problem Summary:

As described in #8120, a large number of rowset meta remain in rocksdb, which may be generated by:

1. drop tablet

    The drop tablet task itself just sets the state of the tablet meta to `SHUTDOWN`
    and moves the tablet to `_shutdown_tablets` vectorm then the background thread
    will periodically clean up the tablet in `_shutdown_tablets` (that's why even if we execute
    the `drop table xx force`, the tablet may be delayed by 10min to 1 hour before it goes into the trash directory).

    The regular cleanup thread in the background saves the complete tablet meta as a `.hdr` file
    when deleting the tablet, and then moves it to the trash directory along with the data files.

    But this process does not process the rowset meta (before doing the checkpoint of the tablet meta,
    the rowset meta is stored independently in rocksdb as a key-value). So this results in a residual rowset meta.

2. clone task

    The clone task may migrate back and forth between BEs, which may result in a situation
    where the tablet id is the same on the BE, but the tablet uuid is different.
    This leads to some rowset meta can not find the corresponding tablet, but there is no thread
    to process these rowsets, and eventually lead to residual.

This is PR, I handled it in the regular cleanup thread with method `_clean_unused_rowset_metas()`.
I did not delete rowset meta along with "drop tablet" task, because "drop tablet" itself is not a synchronous operation.
It also relies on a background thread to clean up the tablet periodically.
So I put this operation in the background cleanup thread. 

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/data_dir.cpp
+++ b/be/src/olap/data_dir.cpp
@@ -527,7 +527,7 @@ OLAPStatus DataDir::load() {
     // 1. add committed rowset to txn map
     // 2. add visible rowset to tablet
     // ignore any errors when load tablet or rowset, because fe will repair them after report
-    int64_t tablet_not_found = 0;
+    int64_t invalid_rowset_counter = 0;
     for (auto rowset_meta : dir_rowset_metas) {
         TabletSharedPtr tablet = _tablet_manager->get_tablet(rowset_meta->tablet_id(),
                                                              rowset_meta->tablet_schema_hash());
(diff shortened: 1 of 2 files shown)


yiguolei:
LGTM

yangzhg (approved):
LGTM

PR #8123, opened 2022-02-17
[improvement] optimize compaction too slow replica process, will set to ok if the replica compaction is done

yangzhg:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

optimize compaction too slow replica process, will set to ok if the replica compaction is done. And will not set bad if more than half repilca is too slow,
fix field in grouping_id() is not the same name will cause error lile  `select list expression not produced by aggregation output (missing from GROUP BY clause?): id`

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/Replica.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/Replica.java
@@ -45,10 +45,12 @@ public class Replica implements Writable {
         SCHEMA_CHANGE,
         CLONE,
         ALTER, // replica is under rollup or schema change
-        DECOMMISSION; // replica is ready to be deleted
+        DECOMMISSION, // replica is ready to be deleted
+        COMPACTION_TOO_SLOW; // replica version count is too large
+
 
         public boolean canLoad() {
-            return this == NORMAL || this == SCHEMA_CHANGE || this == ALTER;
+            return this == NORMAL || this == SCHEMA_CHANGE || this == ALTER || this == COMPACTION_TOO_SLOW;
         }
 
         public boolean canQuery() {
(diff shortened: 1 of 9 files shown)


morningman:
In this PR, you also fix a grouping set bug.
Please add ut for that case, and add it to the PR comment.

fe/fe-core/src/main/java/org/apache/doris/clone/TabletScheduler.java
```diff
@@ -647,7 +647,10 @@ private void handleReplicaMissing(TabletSchedCtx tabletCtx, AgentBatchTask batch
         RootPathLoadStatistic destPath = chooseAvailableDestPath(tabletCtx, tag, false /* not for colocate */);
         Preconditions.checkNotNull(destPath);
         tabletCtx.setDest(destPath.getBeId(), destPath.getPathHash());
-
+        // check compaction too slow file is recovered
+        if (tabletCtx.compactionRecovered()) {
```
morningman, line 651: why not check it before `chooseAvailableDestPath()`?

fe/fe-core/src/main/java/org/apache/doris/catalog/Replica.java
```diff
@@ -46,10 +46,12 @@
...
         CLONE,
         ALTER, // replica is under rollup or schema change
-        DECOMMISSION; // replica is ready to be deleted
+        DECOMMISSION, // replica is ready to be deleted
+        COMPACTION_TOO_SLOW; // replica version count is too large
+
 
         public boolean canLoad() {
-            return this == NORMAL || this == SCHEMA_CHANGE || this == ALTER;
+            return this == NORMAL || this == SCHEMA_CHANGE || this == ALTER || this == COMPACTION_TOO_SLOW;
```
morningman, line 54: This is meaningless.
Because every time you set the state to `COMPACTION_TOO_SLOW`, you also set replica `isBad` to `true`.
I think we can only set replica's state to `COMPACTION_TOO_SLOW`, and change logic in `replica.isAlive()`.

fe/fe-core/src/main/java/org/apache/doris/clone/TabletScheduler.java
```diff
@@ -1023,17 +1042,15 @@ private boolean handleReplicaTooSlow(TabletSchedCtx tabletCtx) throws SchedExcep
...
-            try {
-                Catalog.getCurrentCatalog().setReplicaStatus(tabletCtx.getTabletId(), chosenReplica.getBackendId(),
-                        Replica.ReplicaStatus.BAD);
-                throw new SchedException(Status.FINISHED, "set slow replica as bad");
-            } catch (MetaNotFoundException e) {
-                LOG.warn("set slow replica bad failed:", e);
-                return false;
-            }
+        if (chosenReplica != null && !chosenReplica.equals(minReplica) && minReplica.isAlive()
+                && aliveReplicaCount >= (replicaCount / 2 + 1)) {
```
morningman, line 1046: If there the replica num is 2, then we can not handle this situation.

morningman (approved):
LGTM

PR #8131, opened 2022-02-18
[Improve]Profile logging is enabled by default

hf200012:
Many community users do not know how to set this, they are confused, it is recommended to enable it by default

# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments


caiconghui:
The default value of enableProfile should be false, this only be used for trace query performance, not used on line for better query performance

yangzhg (commented):
Why set this variables to true?  it may be cause performance problem when  qps  is very high

yangzhg (changes requested):
we may need more discuss

hf200012:
Mainly when many users in the community use it, they do not set this, and the web interface cannot see the information, and this problem is raised

hf200012:
> > > information
> 
> > 
> 
> > Maybe should prompt that user should set the variable when entered the web interface.
> 
> 
> 
> Agree, I think we can just add a hint on UI.

good idea

PR #8148, opened 2022-02-19
make asan work as much as possible

dataroaring:
# Proposed changes
make asan work as much as possible

Issue Number: close #xxx

## Problem Summary:
Manually ASAN poisoning is complicated and it is hard to make it work right. There are illustrated examples in http://blog.hostilefork.com/poison-memory-without-asan/.

Stacks of use after poison do not provide enough information to resolve bug, while stacks of use afer free provide. https://github.com/google/sanitizers/issues/191

We'd better implement a mempool using malloc/free directly, thus asan works natively. However we cannot do it in a short time, so we make manual poisoning work as much as possible.

I refers to https://github.com/mcgov/asan_alignment_example.

## Checklist(Required)

## Further comments

We need much more work to make asan work, for example we need fix manual poison in ./be/src/olap/rowset/segment_v2/bitshuffle_page.h and I am not sure if there are any places.

### DIFF
--- a/be/src/runtime/mem_pool.h
+++ b/be/src/runtime/mem_pool.h
@@ -203,26 +203,35 @@ private:
         return chunks_[current_chunk_idx_].allocated_bytes;
     }
 
+    uint8_t * allocate_from_current_chunk(int64_t size, int alignment) {
+        ChunkInfo& info = chunks_[current_chunk_idx_];
+        int64_t aligned_allocated_bytes =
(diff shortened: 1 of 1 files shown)


morningman (approved):
LGTM

yangzhg:
Hi @dataroaring  this pr will cause  `be/test/runtime/mem_pool_test.cpp` failed

PR #8202, opened 2022-02-23
[improvment] show export support label like

wunan1210:
#  Proposed changes

using `show export where label like 'xxx%'` to list more results.

## Checklist(Required)

just copy blindly...  😓

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/ShowExportStmt.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/ShowExportStmt.java
@@ -46,12 +46,13 @@ public class ShowExportStmt extends ShowStmt {
     private static final Logger LOG = LogManager.getLogger(ShowExportStmt.class);
 
     private String dbName;
-    private Expr whereClause;
-    private LimitElement limitElement;
-    private List<OrderByElement> orderByElements;
+    private final Expr whereClause;
+    private final LimitElement limitElement;
+    private final List<OrderByElement> orderByElements;
 
     private long jobId = 0;
     private String label = null;
+    private boolean isLabelUseLike = false;
     private String stateValue = null;
 
     private JobState jobState;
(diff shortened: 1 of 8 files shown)


fe/fe-core/src/main/java/org/apache/doris/analysis/ShowExportStmt.java
```diff
@@ -132,82 +137,47 @@ private void analyzePredicate(Expr whereExpr) throws AnalysisException {
...
-            
-            // right child
-            if (hasState) {
-                if (!(whereExpr.getChild(1) instanceof StringLiteral)) {
-                    valid = false;
-                    break CHECK;
+        boolean valid = false;
+
+        // enumerate all possible conditions
+        if (whereExpr.getChild(0) instanceof SlotRef) {
```
EmmyMiao87, line 143: Could you please add some test

fe/fe-core/src/main/java/org/apache/doris/load/ExportMgr.java
```diff
@@ -147,7 +155,9 @@ private ExportJob createJob(long jobId, ExportStmt stmt) throws Exception {
                     continue;
                 }
```
EmmyMiao87, line 157:
```suggestion
If ( label not null or empty) {
    if (isLabelUseLike) {
      check pattern;
    } else {
       check job label;
    }
}
```

fe/fe-core/src/main/java/org/apache/doris/load/ExportMgr.java
```diff
@@ -147,7 +155,9 @@ private ExportJob createJob(long jobId, ExportStmt stmt) throws Exception {
...
 
-                if (!Strings.isNullOrEmpty(label) && !jobLabel.equals(label)) {
+                if (!Strings.isNullOrEmpty(label) &&
```
EmmyMiao87, line 158: The above way of writing is more readable

morningman (approved):
LGTM

PR #8209, opened 2022-02-23
[fix](replica) handle replica version missing info to avoid -214 error

### LINKED ISSUE #8208: [Bug] query failed with -214
### Search before asking

### Version

0.15

### What's Wrong?

Some time query may failed with error code: -214

```
error Error 5025: errCode = 2, detailMessage = failed to initialize storage reader. tablet=9929137.1435609437.e64a4f69d3271f54-98005e0571265fb7, res=-214, backend=192.168.1.1
```

-214 is error code from BE side: `OLAP_ERR_VERSION_NOT_EXIST`
which means there are missing version of tablet.

### What You Expected?

Although the replica with missing version will be repaired automatically, but it may takes a long time.
We should let FE know the replica's missing version info and do not choose that replica to do the query,
to avoid -214 error during the repair.

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

morningman:
# Proposed changes

Issue Number: close #8208 

## Problem Summary:

In the original tablet reporting information, the version missing information is done by combining
two pieces of information as follows:

1. the maximum consecutive version number
2. the `version_miss` field

The logic of this approach is confusing and inconsistent with the logic of checking for missing versions when querying.

After the change, we directly use the version checking logic used in the query, and set `version_miss` to true
if a missing version is found

and on the FE processing side. Originally, only the **bad replica** information was syncronized among FEs,
but not the **version missing** information. As a result, the non-master FE is not aware of the missing version information.

In the new design, we deprecate the original log persistence class `BackendTabletsInfo` and use the new 
`BackendReplicasInfo` to record replica reporting information and write both **bad** and **version missing**
information to metadata so that other FEs can synchronize these information.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/tablet.cpp
+++ b/be/src/olap/tablet.cpp
@@ -553,10 +553,10 @@ bool Tablet::_reconstruct_version_tracker_if_necessary() {
 
 OLAPStatus Tablet::capture_consistent_versions(const Version& spec_version,
                                                std::vector<Version>* version_path,
-                                               bool quite) const {
+                                               bool quiet) const {
     OLAPStatus status =
             _timestamped_version_tracker.capture_consistent_versions(spec_version, version_path);
-    if (status != OLAP_SUCCESS && !quite) {
+    if (status != OLAP_SUCCESS && !quiet) {
         std::vector<Version> missed_versions;
         calc_missed_versions_unlocked(spec_version.second, &missed_versions);
         if (missed_versions.empty()) {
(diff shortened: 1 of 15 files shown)


be/src/olap/tablet.cpp
```diff
@@ -577,9 +578,9 @@ OLAPStatus Tablet::capture_consistent_versions(const Version& spec_version,
     return status;
 }
 
-OLAPStatus Tablet::check_version_integrity(const Version& version) {
+OLAPStatus Tablet::check_version_integrity(const Version& version, bool quite) {
```
yangzhg, line 581: what does quite mean?

morningman: if !quite, it will print log for missing versions.

yangzhg (dismissed):
LGTM

yangzhg (dismissed):
LGTM

be/src/olap/tablet.cpp
yangzhg: You mean quiet? slient may be better

yangzhg (approved):
LGTM

PR #8233, opened 2022-02-24
[Vectorized][HashJoin] add probe timer

awakeljw:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

add three probe timer. No more than 5% impact on Performance.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/vec/exec/join/vhash_join_node.cpp
+++ b/be/src/vec/exec/join/vhash_join_node.cpp
@@ -166,8 +166,56 @@ struct ProcessHashTableProbe {
               _items_counts(join_node->_items_counts),
               _build_block_offsets(join_node->_build_block_offsets),
               _build_block_rows(join_node->_build_block_rows),
-              _rows_returned_counter(join_node->_rows_returned_counter) {}
+              _rows_returned_counter(join_node->_rows_returned_counter),
+              _search_hashtable_timer(join_node->_search_hashtable_timer),
+              _build_side_output_timer(join_node->_build_side_output_timer),
+              _probe_side_output_timer(join_node->_probe_side_output_timer) {}
 
+    // output build side result column
+    void build_side_output_column(MutableColumns& mcol, int column_offset, int column_length, int size) {
+        constexpr auto is_semi_anti_join = JoinOpType::value == TJoinOp::RIGHT_ANTI_JOIN ||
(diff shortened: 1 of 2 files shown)


be/src/vec/exec/join/vhash_join_node.cpp
```diff
@@ -166,8 +166,58 @@ struct ProcessHashTableProbe {
...
               _build_block_rows(join_node->_build_block_rows),
-              _rows_returned_counter(join_node->_rows_returned_counter) {}
+              _rows_returned_counter(join_node->_rows_returned_counter),
+              _search_hashtable_timer(join_node->_search_hashtable_timer),
+              _build_side_output_timer(join_node->_build_side_output_timer),
+              _probe_side_output_timer(join_node->_probe_side_output_timer) {}
+
+    // output build side result column
+    Status build_side_output_column(MutableColumns& mcol, int column_offset, int column_length, int size) {
+        constexpr auto is_semi_anti_join = JoinOpType::value == TJoinOp::RIGHT_ANTI_JOIN ||
```
HappenLee, line 176: The Function only return OK ？ seems not need to return Status.

be/src/vec/exec/join/vhash_join_node.cpp
```diff
@@ -166,8 +166,58 @@ struct ProcessHashTableProbe {
...
+                        }
+                    }
+                }
+            }
+        }
+        return Status::OK();
+    }
+
+    // output probe side result column
+    Status probe_side_output_column(MutableColumns& mcol, int column_length, int size) {
```
HappenLee, line 214: same to the up

HappenLee (commented):
may should add timer both in other join conjunt

be/src/vec/exec/join/vhash_join_node.cpp
awakeljw: i agree.

be/src/vec/exec/join/vhash_join_node.cpp
awakeljw: i agree.

HappenLee (dismissed):
LGTM

HappenLee (approved):
LGTM

PR #8234, opened 2022-02-24
[Feature] Support pre-aggregation for quantile type

### LINKED ISSUE #7782: [Proposal][Feature] enable Quantile pre-aggregation
### Search before asking

### Description

In current doris-0.15 or elder version, the quantile value is calculated by the detailed data from duplicated module, whose latency is unfriendly under the large scale of data.

Proposed to enable `quantile pre-aggregation` to reduce query latency, already implemented in ClickHouse as follows.

```
SELECT quantileState(number) AS st  -- st is a quantileState generated by 0~9
FROM numbers(10)

Query id: cbde1c1b-e20a-430d-b34a-67c9833be6af

┌─st─────────────────────────────┐
│
6364136223846793005 0 123459     │
└────────────────────────────────┘

1 rows in set. Elapsed: 0.002 sec.

-------------------------------------------------------

SELECT quantileMerge(0.8)(st) -- use quantileMerge function to calculate quantile
FROM
(
    SELECT quantileState(number) AS st
    FROM numbers(10)
)

Query id: 1c25beb5-f6c5-4f32-a6ce-7bbd6d0429ef

┌─quantileMerge(0.8)(st)─┐
│                    7.2 │
└────────────────────────┘
```

Referring to the existing **HLL and bitmap** implementations, the **intermediate state** of the quantile function can be **stored** by TDigest serialization in stream-load step. 

The changes are roughly as follows.

1. A new column named `quantilestate` and corresponding agg function `quantile_union` , `quantile_cal` are supposed to added.
   - quantile_union: add a value into quantilestate
   - quantile_cal(float: percentage):  calculate the quantile of percentage  by quantilestate
   - to_quantile(double: value): transfer value to quantilestate

2. Support for `QuantileState` in query and load step.
   
3. Refactor  `PercentileApproxState` and `TDigest`

### Use case

create table sql like:
```
CREATE TABLE `QuantileState_Test` (
  `keys` bigint(20) NULL COMMENT "keys",
  `quantile_value` quantilestate quantile_union NOT NULL COMMENT "qualite calue"
) ENGINE=OLAP
AGGREGATE KEY(`brand_id`, `dt`, `poi_type`)
COMMENT "bitmap load 测试#OWNER#lihuigang"
PARTITION BY RANGE(`dt`) (
   xxx
)
DISTRIBUTED BY HASH(`keys`) BUCKETS 3
PROPERTIES (
   xxx
);
```

stream load cmd like:

```
curl --location-trusted -u root -H "columns: k1, k2, v1=to_quantilestate(v1)" -T testData http://host:port/api/testDb/testTbl/_stream_load

```

### Related issues

_No response_

### Code of Conduct

spaces-X:
# Proposed changes

close #7782 
## Problem Summary:
Add a new column-type to speed up the approximation of quantiles.
1. The  new column-type is named `quantile_state` with fixed aggregation function `quantile_union`, which stores the intermediate results of pre-aggregated approximation calculations for quantiles.
2. support pre-aggregation of new column-type and quantile_state related functions.

I will update this page to add some test results about this column-type later.

## Checklist(Required)

1. Does it affect the original behavior: (Almost No)
4. Has unit tests been added: (Yes)
5. Has document been added or modified: (Yes)

## Further comments

## SOME TEST RESULTS
**Cluster** : 1 FE + 1 BE   in docker
### Table Schema

bucket num: 3
replica num: 1

Table A: with the new column-type `QuantileState`
```
CREATE TABLE IF NOT EXISTS $db_name.$table_name
(
  k1 int(11) NULL comment "date",
  k2 INT NULL comment "k2 int",
  v1 quantile_state quantile_union,
  v2 INT SUM
)
AGGREGATE KEY(k1, k2)
PARTITION BY RANGE (k1)
(
  PARTITION p1 VALUES [("20210101"), ("20210102")),
  PARTITION p2 VALUES [("20210102"), ("20210103"))
)
DISTRIBUTED BY HASH(k2) BUCKETS $bucket_num
PROPERTIES (
  "replication_num" = "$replica_num",
  "in_memory" = "false",
  "storage_format" = "V2"
);
```
                       
Query SQL of Table A：
```select k1,quantile_percent(quantile_union(v1), 0.95) from test_quantile where k1= 20210101 group by k1;```

Table B: use `BigInt Sum` to replace `QuantileState` 

```
CREATE TABLE IF NOT EXISTS $db_name.$table_name
(
  k1 int(11) NULL comment "date",
  k2 INT NULL comment "k2 int",
  v1 BigInt SUM,
  v2 INT SUM
)
AGGREGATE KEY(k1, k2)
PARTITION BY RANGE (k1)
(
  PARTITION p1 VALUES [("20210101"), ("20210102")),
  PARTITION p2 VALUES [("20210102"), ("20210103"))
)
DISTRIBUTED BY HASH(k2) BUCKETS $bucket_num
PROPERTIES (
  "replication_num" = "$replica_num",
  "in_memory" = "false",
  "storage_format" = "V2"
);
```

Query SQL of Table B: 
```select k1,sum(v2) from test_quantile where k1 = 20210101 group by k1;```

Table C:  store the duplicated data of `k3`
```
CREATE TABLE IF NOT EXISTS $db_name.$details_test_table
(
  k1 int(11) NULL comment "date",
  k2 INT NULL comment "k2 int",
  k3 BIGINT NULL comment " k3 is int key not value",
  v2 INT SUM
)

AGGREGATE KEY(k1, k2, k3)

PARTITION BY RANGE (k1)
(
  PARTITION p1 VALUES [("20210101"), ("20210102")),
  PARTITION p2 VALUES [("20210102"), ("20210103"))
)
DISTRIBUTED BY HASH(k2) BUCKETS $bucket_num
PROPERTIES (
  "replication_num" = "$replica_num",
  "in_memory" = "false",
  "storage_format" = "V2"
);
```

Query SQL of Table C:
```select k1, percentile_approx(k3,0.95) from test_quantile_detail_test where k1 = 20210101 group by k1;```

### Results Sets

number of raw data | number of aggregated data | duration of stream load for table A | duration of stream load for table B | partition data size of table A | partition data size of table B | query time of table A | query time of table B | query time of table C
----- | ----- | ----- | ----- | ----- | ----- | ----- | ----- | -----
4 million | 50,000 | 15s | 14s | 16MB | 255KB | 0.25s | 0.20s | 0.85s
4 million | 500 | 15s | 14s | 18MB | 5KB | 0.15s | 0.01s | 0.75s
4 million | 50 | 15s | 14s | 10MB | 2KB | 0.11s | 0.01s | 0.69s
40 million | 50,000 | 126s | 122s | 176MB | 282KB | 1.87s | 0.02s | 6s
40 million | 500 | 126s | 122s | 104MB | 5.8KB | 1.02s | 0.01s | 5.8s
40 million | 50 | 126s | 122s | 11MB | 2.5KB | 0.09s | 0.01s | 6s

### CODE REVIEW

File: be/src/util/quantile_state.cpp [hunk 1]
```diff
@@ -0,0 +1,377 @@
...
+            break;
+        case SINGLE:
+            _type = EXPLICIT;
+            _explicit_data.swap(other._explicit_data);
+            add_value(_single_data);
+            break;
+        case EXPLICIT:
+            if (_explicit_data.size() + other._explicit_data.size() > QUANTILE_STATE_EXPLICIT_NUM) {
+                _type = TDIGEST;
+                tdigest_ptr = new TDigest(compression);
```
spaces-X, line 273: use `new TDigest(compression,compression,compression)` will save lots of memory.

be/src/exprs/quantile_function.cpp
```diff
@@ -0,0 +1,127 @@
...
+}
+
+StringVal QuantileStateFunctions::quantile_state_serialize(FunctionContext* ctx, const StringVal& src) {
+    if(src.is_null) {
+        return src;
+
+    }
+    auto tmp_ptr = reinterpret_cast<QuantileState<double>*>(src.ptr);
+    StringVal result = serialize(ctx, tmp_ptr);
+    delete tmp_ptr;
```
wangbo, line 123: Why delete src.ptr here? Where is its memory allocated?

HappenLee:
before we release the feature, maybe we should support it in vec exec engine both.

be/src/exprs/quantile_function.cpp
spaces-X: This function will be called in the PartitionedAggregationNode.
The memory is allocated from `quantile_state_init` initialize with a `empty quantile_state object`.
The `empty quantile_state` object is for aggregate calculations and it can be deleted after the aggregated results are serialized.

spaces-X:
> before we release the feature, maybe we should support it in vec exec engine both.

yeap, i think it is more reasonable to put it in another pr and i will implement it in the near future.

be/src/olap/row_block.cpp
```diff
@@ -89,8 +89,7 @@ void RowBlock::_compute_layout() {
         _field_offset_in_memory.push_back(memory_size);
 
         // All field has a nullbyte in memory
-        if (column.type() == OLAP_FIELD_TYPE_VARCHAR || column.type() == OLAP_FIELD_TYPE_HLL ||
-            column.type() == OLAP_FIELD_TYPE_CHAR || column.type() == OLAP_FIELD_TYPE_OBJECT ||column.type() == OLAP_FIELD_TYPE_STRING) {
+        if (column.is_slice_stored()) {
```
wangbo, line 92: Maybe ```is_variable_type```  be better than ```is_slice_stored``` ?

be/src/olap/tablet_schema.cpp
```diff
@@ -246,6 +256,7 @@ uint32_t TabletColumn::get_field_length_by_type(TPrimitiveType::type type, uint3
         return 4;
     case TPrimitiveType::DOUBLE:
         return 8;
+    case TPrimitiveType::QUANTILE_STATE:
```
wangbo, line 259: Why ```QUANTILE_STATE``` share the same type length with ```Object Type```?
Is QUANTILE variable length type or fixed length?

spaces-X: QUANTILE_STATE is variable length type, which is same to BITMAP.

It only store `ptr`(8 byte) and `length`(8 byte) in the TupleRow.

spaces-X: QUANTILE_STATE is variable length type, which is same to BITMAP.

It only store ptr(8 byte) and length(8 byte) in the TupleRow.

be/src/olap/row_block.cpp
spaces-X: +1

gensrc/thrift/PlanNodes.thrift
```diff
@@ -157,8 +157,8 @@ struct TBrokerRangeDesc {
 }
 
 struct TBrokerScanRangeParams {
-    1: required i8 column_separator;
-    2: required i8 line_delimiter;
+    1: required byte column_separator;
```
wangbo, line 160: Why using byte instead of i8 here?

be/src/olap/aggregate_func.h
```diff
@@ -580,6 +581,53 @@ struct AggregateFuncTraits<OLAP_FIELD_AGGREGATION_BITMAP_UNION, OLAP_FIELD_TYPE_
...
+                     ObjectPool* agg_pool) {
+        DCHECK_EQ(src_null, false);
+        dst->set_not_null();
+
+        auto* src_slice = reinterpret_cast<const Slice*>(src);
+        auto* dst_slice = reinterpret_cast<Slice*>(dst->mutable_cell_ptr());
+
+        // we use zero size represent this slice is a agg object
+        dst_slice->size = 0;
+        auto* quantile_state = new QuantileState<double>(*src_slice);
```
wangbo, line 596:
```suggestion
        auto* dst_quantile_state = new QuantileState<double>(*src_slice);
```

be/src/olap/aggregate_func.h
```diff
@@ -580,6 +581,53 @@ struct AggregateFuncTraits<OLAP_FIELD_AGGREGATION_BITMAP_UNION, OLAP_FIELD_TYPE_
...
+
+        mem_pool->mem_tracker()->Consume(quantile_state->get_serialized_size());
```
wangbo, line 598: Why not using ```sizeof(QuantileState)``` instead of ```quantile_state->get_serialized_size()```?
Which one is more precise?

be/src/exprs/quantile_function.cpp
```diff
@@ -0,0 +1,127 @@
...
+    }
+    
+    if(!src.is_null) {
+        StringParser::ParseResult parse_result = StringParser::PARSE_SUCCESS;
+        double double_value = StringParser::string_to_float<double>(
+                reinterpret_cast<char*>(src.ptr), src.len, &parse_result);
+        if (UNLIKELY(parse_result != StringParser::PARSE_SUCCESS)) {
+            std::stringstream error_msg;
+            error_msg << "The input: " << std::string(reinterpret_cast<char*>(src.ptr), src.len)
+                      << " is not valid, to_bitmap only support bigint value from 0 to "
```
wangbo, line 64: why ```to_bitmap``` appears here?

be/src/util/quantile_state.cpp
```diff
@@ -0,0 +1,377 @@
...
+template<typename T>
+QuantileState<T>::QuantileState():_type(EMPTY),compression(2048){}
+
+template<typename T>
+QuantileState<T>::QuantileState(float compression):_type(EMPTY),compression(compression){}
+
+
+template<typename T>
+QuantileState<T>::QuantileState(const Slice& slice) {
+    if (!deserialize(slice)) {
```
wangbo, line 35: Add a ```DCHECK``` for deserialize result.

be/src/util/quantile_state.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+template<typename T>
+void QuantileState<T>::clear() {
+    _type = EMPTY;
+    if (tdigest_ptr != nullptr) {
+        delete tdigest_ptr;
+        tdigest_ptr = nullptr;
+    }
+    _explicit_data.clear();
+    _explicit_data.shrink_to_fit();
```
wangbo, line 371: ```_explicit_data``` is  a field of quantitle_state, is it necessary to clear it manually?

be/src/util/quantile_state.cpp
```diff
@@ -0,0 +1,377 @@
...
+template<typename T>
+QuantileState<T>::QuantileState(const Slice& slice) {
+    if (!deserialize(slice)) {
+        _type = EMPTY;
+    }
+}
+
+template<typename T>
+QuantileState<T>::~QuantileState() {
+    clear();
```
wangbo, line 42: Maybe just delete ```tdigest_ptr```  is enough.

be/src/util/quantile_state.cpp
```diff
@@ -0,0 +1,377 @@
...
+    auto type = (QuantileStateType)*ptr++;
+    switch (type) {
+    case EMPTY:
+        break;
+    case SINGLE:{
+        if ((ptr + sizeof(T)) > end) {
+            return false;
+        }
+        ptr += sizeof(T);
+        // _single_data = *reinterpret_cast<T*>(ptr)++;
```
wangbo, line 93: remove this

be/src/util/quantile_state.cpp
```diff
@@ -0,0 +1,377 @@
...
+        ptr += sizeof(uint16_t);
+        ptr += num_explicits * sizeof(T);
+        break;
+    }
+    case TDIGEST: {
+        if ((ptr + sizeof(uint32_t)) > end) {
+            return false;
+        }
+        uint32_t tdigest_serialized_length = decode_fixed32_le(ptr);
+        ptr += tdigest_serialized_length;
```
wangbo, line 110: Why not check content of tdigest? Such as ```isDirty()```.

gensrc/thrift/PlanNodes.thrift
spaces-X: i got a compile error before when use i8 type.
i will try again in the next commit.

be/src/olap/aggregate_func.h
spaces-X: The result of sizeof() is not incorrect, which is much lower than actual memory usage, cause we have vector or obj ptr in the member variables.
The result of  get_serialized_size is closer to real memory usage.

be/src/util/quantile_state.cpp
spaces-X: Here just check whether the total length is correct.
If check the content , we should call `deserialize()` first. It will take lots of time.

be/src/exprs/anyval_util.cpp
```diff
@@ -146,6 +147,10 @@ FunctionContext::TypeDesc AnyValUtil::column_type_to_type_desc(const TypeDescrip
     case TYPE_OBJECT:
         out.type = FunctionContext::TYPE_OBJECT;
         // FIXME(cmy): is this fallthrough meaningful?
+        break;    //TODO(weixiang):why no break here?
```
wangshuo128, line 150: If not sure why no `break` here, it's better to not change the original behavior here. 
Avoid introducing weird and unrelated bug by this feature.

be/src/util/quantile_state.cpp
```diff
@@ -0,0 +1,377 @@
...
+        break;
+    }
+
+}
+
+template<typename T>
+void QuantileState<T>::clear() {
+    _type = EMPTY;
+    if (tdigest_ptr != nullptr) {
+        delete tdigest_ptr;
```
wangshuo128, line 367: It's safe to delete a `nullptr`, no need to check here.

be/src/util/quantile_state.h
```diff
@@ -0,0 +1,72 @@
...
+    void clear();
+    bool is_valid(const Slice& slice);
+    size_t get_serialized_size();
+    T get_value_by_percentile(float percentile);
+    T get_explicit_value_by_percentile(float percentile);
+    ~QuantileState();
+
+private:
+    QuantileStateType _type = EMPTY;
+    TDigest* tdigest_ptr = nullptr;
```
wangshuo128, line 63:
```suggestion
    TDigest* _tdigest_ptr = nullptr;
```

be/src/util/quantile_state.h
```diff
@@ -0,0 +1,72 @@
...
+    T _single_data;
+    std::vector<T> _explicit_data;
+    float compression;
```
wangshuo128, line 66:
```suggestion
    float _compression;
```

be/src/exprs/quantile_function.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+static StringVal serialize(FunctionContext* ctx, QuantileState<double>* value) {
+    StringVal result(ctx, value->get_serialized_size());
+    value->serialize(result.ptr);
+    return result;
+}
+
+StringVal QuantileStateFunctions::to_quantile_state(FunctionContext* ctx, const StringVal& src) {
+    QuantileState<double> quantile_state;
+    quantile_state.set_compression(2048);
```
wangshuo128, line 48: It's better to use a named constant to represent `2048` and `10000`, etc, rather than these magic numbers.

be/src/exprs/quantile_function.cpp
```diff
@@ -0,0 +1,127 @@
...
+    if(src.len == 0) {
+        dst_quantile->merge(*reinterpret_cast<QuantileState<double>*>(src.ptr));
+    } else {
+        QuantileState<double> state(Slice(src.ptr, src.len));
+        dst_quantile->merge(state);
+    }
+}
+
+DoubleVal QuantileStateFunctions::quantile_percent(FunctionContext* ctx, StringVal& src) {
+    const AnyVal* percentile = ctx->get_constant_arg(1);
```
wangshuo128, line 88: I'm afraid we should add a `quantile_percent _prepare` function to do some preparation work, e.g., handle constant arguments checks.

fe/fe-core/src/main/java/org/apache/doris/analysis/CreateTableStmt.java
```diff
@@ -386,25 +386,18 @@ public void analyze(Analyzer analyzer) throws UserException {
...
         }
 
-        if (hasHll && keysDesc.getKeysType() != KeysType.AGG_KEYS) {
-            throw new AnalysisException("HLL must be used in AGG_KEYS");
-        }
-
-        if (hasBitmap && keysDesc.getKeysType() != KeysType.AGG_KEYS) {
-            throw new AnalysisException("BITMAP_UNION must be used in AGG_KEYS");
+        if (hasObjectStored && keysDesc.getKeysType() != KeysType.AGG_KEYS) {
+            throw new AnalysisException("object stored column must be used in AGG_KEYS table, object stored column:" + objectStoredColumn);
```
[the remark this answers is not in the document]
spaces-X: ok i will remove the object stored key words just keep the column name in message

be/src/olap/aggregate_func.h
```diff
@@ -580,6 +581,53 @@ struct AggregateFuncTraits<OLAP_FIELD_AGGREGATION_BITMAP_UNION, OLAP_FIELD_TYPE_
...
+        dst_slice->data = reinterpret_cast<char*>(quantile_state);
+
+        agg_pool->add(quantile_state);
+    }
+
+    static void update(RowCursorCell* dst, const RowCursorCell& src, MemPool* mem_pool) {
+        DCHECK_EQ(src.is_null(), false);
+
+        auto* dst_slice = reinterpret_cast<Slice*>(dst->mutable_cell_ptr());
+        auto* src_slice = reinterpret_cast<const Slice*>(src.cell_ptr());
```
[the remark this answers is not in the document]
spaces-X: It will meet compile error cause `srr.cell_ptr` is const.

be/src/exprs/quantile_function.cpp
spaces-X: I did not get the benfit of adding a quantile_percent_prepare function. May be this can be implemented in support vec-functions of quantile_state.

be/src/olap/aggregate_func.h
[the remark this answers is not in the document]
wangshuo128: Ah, I see, my mistake.

be/src/exprs/quantile_function.cpp
wangshuo128: Please ignore the comments before.
Hmm... I think we should do arguments' validation check in FE, i.e., `FunctionCall.analyzeImpl`, rather than in BE.

be/src/util/quantile_state.h
```diff
@@ -0,0 +1,74 @@
...
+
+namespace doris {
+
+
+class Slice;
+class TDigest;
+
+const static int QUANTILE_STATE_EXPLICIT_NUM = 2048;
+const static int QUANTILE_STATE_COMPRESSION_MAX = 2048;
+const static int QUANTILE_STATE_COMPRESSION_MIN = 10000;
```
spaces-X, line 35: messed up of min and max

spaces-X:
@wangbo @wangshuo128 
Already accepted and completed some of your review comments.

wangshuo128:
Please format the code style. Maybe https://doris.apache.org/branch-0.13/zh-CN/developer-guide/format-code.html#%E4%BB%A3%E7%A0%81%E9%A3%8E%E6%A0%BC%E5%AE%9A%E5%88%B6 would help.

spaces-X:
@HappenLee @wangbo 
I've made the modifications as suggested by the reviewer.
Please review it soon otherwise I need to keep resolving conflicts with master.

wangbo (approved):
LGTM

PR #8243, opened 2022-02-25
[improvment] Ignore the case of column name, when do the stream load in json format

stgztsw:
# Proposed changes

Issue Number: close #8070

## Problem Summary:

stream load in json format is case sensitive, In order to be consistent with sql standard, change it to  case insensitive

## Checklist(Required)

## Further comments


morningman (approved):
LGTM

caiconghui:
I have two question:
1. this solution is only for simple json, and other json would cause different behavior?
2. it seems that if user not use fuzzy_parse for optimization that would cause different result?

morningman (changes requested):
> 

This is indeed a problem.
Json is case sensitive, but column name in Doris is case insensitive.
We need to make a clear definition about how to map json key to Doris column.

PR #8250, opened 2022-02-26
[revert] "[improvement]upgrade-grpc-version"

morningman:
Reverts apache/incubator-doris#8218

Because when using grpc 1.44.1, the corresponding `protoc-gen-grpc-java` plugin
requried GLIBC_2.14, which is not found in CentOS 6.

So I suggest to revert this commit this time. And considering upgrading this component
after most systems have reached glibc version 2.14.

And for Mac M1, you may have to change this version manually for now


morningman:
CC @wunan1210

hf200012 (approved):
LGTM

PR #8289, opened 2022-03-01
[refactor] remove types_test

yangzhg:
# Proposed changes

1. remove types_test, it will cause core dump in higher version gcc or
   clang, becuse of memory align, some code will be vectorize in higher
   gcc or clang
2. change string type length to 2GB instead of -1
3. modify unaccessable code

## Problem Summary:

## Checklist(Required)

## Further comments


fe/fe-core/src/main/java/org/apache/doris/catalog/ScalarType.java
```diff
@@ -291,7 +291,8 @@ public static ScalarType createVarcharType(String lenStr) {
     public static ScalarType createStringType() {
         // length checked in analysis
         ScalarType type = new ScalarType(PrimitiveType.STRING);
-        type.len = -1;
+        // 2GB - 4
+        type.len = 2147483643;
```
morningman, line 295: Define this magic number some where. Maybe in Type.java?
And explain why minus 4 in comment.

fe/fe-core/src/main/java/org/apache/doris/catalog/ScalarType.java
```diff
@@ -69,6 +66,12 @@
     // Keep consistent with backend ColumnType::CHAR_INLINE_LENGTH
     public static final int CHAR_INLINE_LENGTH = 128;
 
+    // Max length of String types, in be storage layer store string length
+    // using int32, the max length is 2GB, the first 4 bytes store the length
+    // so the max available length is 2GB - 4
+
```
HappenLee, line 72: no need the space？

fe/fe-core/src/main/java/org/apache/doris/catalog/ScalarType.java
```diff
@@ -69,6 +66,12 @@
...
+    public static final int MAX_STRING_LENGTH = 0x7fffffff - 4;
```
levy5307, line 73: I think `public static final int MAX_STRING_LENGTH = 2 << 30 - 1 - 4;` is more clear here. 
And you should explain why did you minus 1 here.

yangzhg: It is explained in comments that the value is 2GB - 4, why  use `public static final int MAX_STRING_LENGTH = 2 << 30 - 1 - 4;` ?  How to compute 2GB - 4 is not important and  2 << 30 does not have performance advantages

morningman (approved):
LGTM

PR #8297, opened 2022-03-02
[build](fe) Support building an individual maven module.

wangshuo128:
# Proposed changes

Current, if we build an individual FE maven module, e.g., build `fe-core` moudle via `mvn clean package -pl fe-core`, the below error would occur
```
[ERROR] Failed to execute goal on project fe-core: Could not resolve dependencies for project org.apache.doris:fe-core:jar:0.15-SNAPSHOT: Failed to collect dependencies at org.apache.doris:spark-dpp:jar:0.15-SNAPSHOT: Failed to read artifact descriptor for org.apache.doris:spark-dpp:jar:0.15-SNAPSHOT: Could not find artifact org.apache.doris:fe:pom:${revision}
```
It's because the maven module version is a property defined in the root pom.xml:
https://github.com/apache/incubator-doris/blob/246ac4e37aa4da6836b7850cb990f02d1c3725a3/fe/pom.xml#L121

This PR tried to fix this issue.

1. Use the version value in every pom.xml file instead of a property in the root pom.xml
2. If we want to change the project version, we could run the following commands in the `fe` dir:
``` bash
mvn versions:set -DnewVersion=${new_version}
mvn versions:commit
```
Please see https://www.mojohaus.org/versions-maven-plugin/examples/set.html for more details.
## Problem Summary:

## Checklist(Required)

1. Does it affect the original behavior: no
4. Has unit tests been added: no
5. Has document been added or modified: no
6. Does it need to update dependencies: no
7. Are there any changes that cannot be rolled back: no

## Further comments

### DIFF
--- a/fe/fe-common/pom.xml
+++ b/fe/fe-common/pom.xml
@@ -22,7 +22,7 @@ under the License.
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.apache.doris</groupId>
-        <version>${revision}</version>
+        <version>1.0-SNAPSHOT</version>
         <artifactId>fe</artifactId>
         <relativePath>../pom.xml</relativePath>
     </parent>
--- a/fe/fe-core/pom.xml
+++ b/fe/fe-core/pom.xml
@@ -22,7 +22,7 @@ under the License.
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.apache.doris</groupId>
-        <version>${revision}</version>
+        <version>1.0-SNAPSHOT</version>
         <artifactId>fe</artifactId>
         <relativePath>../pom.xml</relativePath>
     </parent>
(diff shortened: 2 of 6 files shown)


yangzhg:
I think this is not a good way, the purpose of using reversion is to unify version management, has added maven flatten plugin, in the release will play flat version, as a dependency should also rely on maven on the determination of the version of the package

wangshuo128:
> I think this is not a good way, the purpose of using reversion is to unify version management, has added maven flatten plugin, in the release will play flat version, as a dependency should also rely on maven on the determination of the version of the package



> I think this is not a good way, the purpose of using reversion is to unify version management, has added maven flatten plugin, in the release will play flat version, as a dependency should also rely on maven on the determination of the version of the package
> you can use `mvn -Drevision=0.15-SNAPSHOT` to compile child module
> So here is my -1

Ah, I see, make sense to me, closing this PR.

wangshuo128:
> you can use `mvn -Drevision=0.15-SNAPSHOT` to compile child module

It seems that using `-Drevision` does not work. So I reopen this PR. Looking forward to more thoughts about this.
cc @yangzhg @EmmyMiao87 @morrySnow @924060929 @morningman :)

PR #8318, opened 2022-03-03
[improvement](storage) Low cardinality string optimization in storage layer

### LINKED ISSUE #8315: [Enhancement] Low cardinality string optimization in storage layer
### Search before asking

### Description

For the current low-cardinality string column, when loading data in the storage layer, it is necessary to traverse all the codes in the data_page, convert it into the original string value through the dictionary in the dict_page, and then insert it into the ColumnString column. Subsequent predicate calculation, filtering, etc. are based on ColumnString for processing.

If you continue to use encoding during predicate calculation, it is equivalent to converting string comparisons to int type comparisons, resulting in greater performance and less memory consumption.

### Solution

Storage layer read call: `SegmentIterator::next_batch-> ... -> PageDecoder::next_batch`, the default batch reads 4096 rows, PageDecoder reads part of data_page, data_page may be bitshuffle encoding or plain encoding.

After enabling low cardinality optimization, optimize for string columns with predicate calculations. The resulting column is of type ColumnDictionary and is used to store encodings and dictionaries. If the current batch of data_pages are all bitshuffle encoded, that is, directly insert the encoded value and dictionary into the result column, in the process of PageDecode next_batch, the encoding to String overhead (there is the overhead of constructing a dictionary) is saved. After the column data is loaded in the Segment layer, the predicate calculation is performed based on the encoded value and the dictionary, that is, the predicate constant is converted into an int value through the dictionary, and the int value is compared with the int in the encoded column, which is equivalent to comparing the original String. Converted to int comparison, there will be a significant speedup effect. After the predicate is calculated, the encoding column and dictionary are converted into ColumnString columns to end the segment layer reading.

If there is plain encoded data_page in the current batch, the dictionary cannot be used and is converted to a PredicateColumn column for processing.

### Performance
env: 1 FE, 3 BE
version: c66a9bf

| Id        | Doris vectorization (including storage layer)                | After optimization                                           | Remark |
| --------- | ------------------------------------------------------------ | ------------------------------------------------------------ | ---- |
| Q2.1 flat | - Total: 769ms<br/>    - BlockLoadTime: 10s636ms<br/>    - BlockSeekTime: 3s199ms<br/>        - ReadColByIdxTime: 3s732ms<br/>        - **ShortPredEvalTime: 3s31ms**<br/>        - OutputColBySelIdxTime: 72.968ms<br/>        - OutputNonPredColTime: 9.814ms<br/>        - ReadColByRowIdTime: 3s589ms<br/>        - InitCurBlockTime: 42.203ms<br/>        - VectorPredEvalTime: 104.169ms | - Total: 427ms<br/>    - BlockLoadTime: 5s720ms<br/>    - BlockSeekTime: 1s798ms<br/>        - ReadColByIdxTime: 3s32ms<br/>        - **ShortPredEvalTime: 277.569ms**<br/>        - OutputColBySelIdxTime: 67.17ms<br/>        - OutputNonPredColTime: 6.655ms<br/>        - ReadColByRowIdTime: 2s171ms<br/>        - InitCurBlockTime: 37.303ms<br/>        - VectorPredEvalTime: 98.790ms |      |
| Q2.2 flat | - Total: 645ms<br/>    - BlockLoadTime: 9s352ms<br/>    - BlockSeekTime: 1s145ms<br/>        - ReadColByIdxTime: 2s957ms<br/>        - **ShortPredEvalTime: 5s10ms**<br/>        - OutputColBySelIdxTime: 27.5ms<br/>        - OutputNonPredColTime: 7.275ms<br/>        - ReadColByRowIdTime: 1s193ms<br/>        - InitCurBlockTime: 32.54ms<br/>        - VectorPredEvalTime: 92.931ms | - Total: 407ms<br/>    - BlockLoadTime: 5s385ms<br/>    - BlockSeekTime: 1s365ms<br/>        - ReadColByIdxTime: 3s219ms<br/>        - **ShortPredEvalTime: 559.475ms**<br/>        - OutputColBySelIdxTime: 31.517ms<br/>        - OutputNonPredColTime: 7.107ms<br/>        - ReadColByRowIdTime: 1s408ms<br/>        - InitCurBlockTime: 28.264ms<br/>        - VectorPredEvalTime: 94.499ms |      |
| Q2.3 flat | - Total: 511ms<br/>    - BlockLoadTime: 6s997ms<br/>    - BlockSeekTime: 939.491ms<br/>        - ReadColByIdxTime: 2s831ms<br/>        - **ShortPredEvalTime: 3s65ms**<br/>        - OutputColBySelIdxTime: 12.812ms<br/>        - OutputNonPredColTime: 7.697ms<br/>        - ReadColByRowIdTime: 937.718ms<br/>        - InitCurBlockTime: 19.722ms<br/>        - VectorPredEvalTime: 93.701ms | - Total: 352ms<br/>    - BlockLoadTime: 4s748ms<br/>    - BlockSeekTime: 1s218ms<br/>        - ReadColByIdxTime: 3s105ms<br/>        - **ShortPredEvalTime: 258.50ms**<br/>        - OutputColBySelIdxTime: 15.446ms<br/>        - OutputNonPredColTime: 6.700ms<br/>        - ReadColByRowIdTime: 1s211ms<br/>        - InitCurBlockTime: 20.622ms<br/>        - VectorPredEvalTime: 95.373ms |      |

In the SSB fact-table test Q2.x, after the low-cardinality string optimization is completed, the short-circuit predicate calculation performance is improved by about 90%, the storage layer is improved by about 40%, and the overall improvement is about 37%.

### Code of Conduct

zenoyang:
# Proposed changes

Issue Number: close #8315 

Low cardinality string optimization in storage layer

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/comparison_predicate.cpp
+++ b/be/src/olap/comparison_predicate.cpp
@@ -161,7 +161,7 @@ COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(GreaterEqualPredicate, >=)
                 if constexpr (std::is_same_v<type, StringValue>) {                                 \
                     auto* nested_col_ptr = vectorized::check_and_get_column<                       \
                             vectorized::ColumnDictionary<vectorized::Int32>>(nested_col);          \
-                    auto code = nested_col_ptr->get_code(_value);                                  \
+                    auto code = nested_col_ptr->find_code(_value);                                 \
                     auto& data_array = nested_col_ptr->get_data();                                 \
                     for (uint16_t i = 0; i < *size; i++) {                                         \
                         uint16_t idx = sel[i];                                                     \
@@ -192,7 +192,7 @@ COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(GreaterEqualPredicate, >=)
                         reinterpret_cast<vectorized::ColumnDictionary<vectorized::Int32>&>(        \
                                 column);                                                           \
                 auto& data_array = dict_col.get_data();                                            \
-                auto code = dict_col.get_code(_value);                                             \
+                auto code = dict_col.find_code(_value);                                            \
                 for (uint16_t i = 0; i < *size; ++i) {                                             \
                     uint16_t idx = sel[i];                                                         \
                     sel[new_size] = idx;                                                           \
--- a/be/src/olap/comparison_predicate.cpp
+++ b/be/src/olap/comparison_predicate.cpp
@@ -146,6 +146,7 @@ COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(LessEqualPredicate, <=)
 COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(GreaterPredicate, >)
 COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(GreaterEqualPredicate, >=)
 
+// todo(zeno) define interface in IColumn to simplify code
 #define COMPARISON_PRED_COLUMN_EVALUATE(CLASS, OP, IS_RANGE)                                       \
     template <class type>                                                                          \
     void CLASS<type>::evaluate(vectorized::IColumn& column, uint16_t* sel, uint16_t* size) const { \
@@ -161,10 +162,8 @@ COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(GreaterEqualPredicate, >=)
                 if constexpr (std::is_same_v<type, StringValue>) {                                 \
                     auto* nested_col_ptr = vectorized::check_and_get_column<                       \
                             vectorized::ColumnDictionary<vectorized::Int32>>(nested_col);          \
-                    auto code = nested_col_ptr->find_code(_value);                                 \
-                    if (code < 0 && IS_RANGE) {                                                    \
-                        code = nested_col_ptr->find_bound_code(_value, 0 OP 1, 1 OP 1 );           \
-                    }                                                                              \
+                    auto code = IS_RANGE ? nested_col_ptr->find_bound_code(_value, 0 OP 1, 1 OP 1) \
+                                         : nested_col_ptr->find_code(_value);                      \
                     auto& data_array = nested_col_ptr->get_data();                                 \
                     for (uint16_t i = 0; i < *size; i++) {                                         \
                         uint16_t idx = sel[i];                                                     \
@@ -194,10 +193,8 @@ COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(GreaterEqualPredicate, >=)
                 auto& dict_col =                                                                   \
                         reinterpret_cast<vectorized::ColumnDictionary<vectorized::Int32>&>(column);\
                 auto& data_array = dict_col.get_data();                                            \
-                auto code = dict_col.find_code(_value);                                            \
-                if (code < 0 && IS_RANGE) {                                                        \
-                    code = dict_col.find_bound_code(_value, 0 OP 1, 1 OP 1);                       \
-                }                                                                                  \
+                auto code = IS_RANGE ? dict_col.find_bound_code(_value, 0 OP 1, 1 OP 1)            \
+                                     : dict_col.find_code(_value);                                 \
                 for (uint16_t i = 0; i < *size; ++i) {                                             \
                     uint16_t idx = sel[i];                                                         \
                     sel[new_size] = idx;                                                           \
--- a/be/src/olap/comparison_predicate.cpp
+++ b/be/src/olap/comparison_predicate.cpp
@@ -21,6 +21,7 @@
 #include "olap/schema.h"
 #include "runtime/string_value.hpp"
 #include "runtime/vectorized_row_batch.h"
+#include "vec/columns/column_dictionary.h"
 #include "vec/columns/column_nullable.h"
 #include "vec/columns/column_vector.h"
 #include "vec/columns/predicate_column.h"
@@ -150,23 +151,58 @@ COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(GreaterEqualPredicate, >=)
     void CLASS<type>::evaluate(vectorized::IColumn& column, uint16_t* sel, uint16_t* size) const { \
         uint16_t new_size = 0;                                                                     \
         if (column.is_nullable()) {                                                                \
-            auto* nullable_column =                                                                \
+            auto* nullable_col =                                                                   \
                     vectorized::check_and_get_column<vectorized::ColumnNullable>(column);          \
-            auto& null_bitmap = reinterpret_cast<const vectorized::ColumnVector<uint8_t>&>(        \
-                                        *(nullable_column->get_null_map_column_ptr()))             \
+            auto& null_bitmap = reinterpret_cast<const vectorized::ColumnUInt8&>(                  \
+                                        nullable_col->get_null_map_column())                       \
                                         .get_data();                                               \
-            auto* nest_column_vector =                                                             \
-                    vectorized::check_and_get_column<vectorized::PredicateColumnType<type>>(       \
-                            nullable_column->get_nested_column());                                 \
-            auto& data_array = nest_column_vector->get_data();                                     \
-            for (uint16_t i = 0; i < *size; i++) {                                                 \
-                uint16_t idx = sel[i];                                                             \
-                sel[new_size] = idx;                                                               \
-                const type& cell_value = reinterpret_cast<const type&>(data_array[idx]);           \
-                bool ret = !null_bitmap[idx] && (cell_value OP _value);                            \
-                new_size += _opposite ? !ret : ret;                                                \
+            auto& nested_col = nullable_col->get_nested_column();                                  \
+            if (nested_col.is_column_dictionary()) {                                               \
+                if constexpr (std::is_same_v<type, StringValue>) {                                 \
+                    auto* nested_col_ptr = vectorized::check_and_get_column<                       \
+                            vectorized::ColumnDictionary<vectorized::Int32>>(nested_col);          \
+                    auto code = nested_col_ptr->get_code(_value);                                  \
+                    auto& data_array = nested_col_ptr->get_data();                                 \
+                    for (uint16_t i = 0; i < *size; i++) {                                         \
+                        uint16_t idx = sel[i];                                                     \
+                        sel[new_size] = idx;                                                       \
+                        const auto& cell_value =                                                   \
+                                reinterpret_cast<const vectorized::Int32&>(data_array[idx]);       \
+                        bool ret = !null_bitmap[idx] && (cell_value OP code);                      \
+                        new_size += _opposite ? !ret : ret;                                        \
+                    }                                                                              \
+                }                                                                                  \
+            } else {                                                                               \
+                auto* nested_col_ptr =                                                             \
+                        vectorized::check_and_get_column<vectorized::PredicateColumnType<type>>(   \
+                                nested_col);                                                       \
+                auto& data_array = nested_col_ptr->get_data();                                     \
+                for (uint16_t i = 0; i < *size; i++) {                                             \
+                    uint16_t idx = sel[i];                                                         \
+                    sel[new_size] = idx;                                                           \
+                    const type& cell_value = reinterpret_cast<const type&>(data_array[idx]);       \
+                    bool ret = !null_bitmap[idx] && (cell_value OP _value);                        \
+                    new_size += _opposite ? !ret : ret;                                            \
+                }                                                                                  \
             }                                                                                      \
             *size = new_size;                                                                      \
+        } else if (column.is_column_dictionary()) {                                                \
+            if constexpr (std::is_same_v<type, StringValue>) {                                     \
+                auto& dict_col =                                                                   \
+                        reinterpret_cast<vectorized::ColumnDictionary<vectorized::Int32>&>(        \
+                                column);                                                           \
+                auto& data_array = dict_col.get_data();                                            \
+                auto code = dict_col.get_code(_value);                                             \
+                for (uint16_t i = 0; i < *size; ++i) {                                             \
+                    uint16_t idx = sel[i];                                                         \
+                    sel[new_size] = idx;                                                           \
+                    const auto& cell_value =                                                       \
+                            reinterpret_cast<const vectorized::Int32&>(data_array[idx]);           \
+                    bool ret = cell_value OP code;                                                 \
+                    new_size += _opposite ? !ret : ret;                                            \
+                }                                                                                  \
+                *size = new_size;                                                                  \
+            }                                                                                      \
         } else {                                                                                   \
             auto& pred_column_ref =                                                                \
                     reinterpret_cast<vectorized::PredicateColumnType<type>&>(column);              \
--- a/be/src/olap/comparison_predicate.cpp
+++ b/be/src/olap/comparison_predicate.cpp
@@ -146,7 +146,7 @@ COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(LessEqualPredicate, <=)
 COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(GreaterPredicate, >)
 COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(GreaterEqualPredicate, >=)
 
-#define COMPARISON_PRED_COLUMN_EVALUATE(CLASS, OP)                                                 \
+#define COMPARISON_PRED_COLUMN_EVALUATE(CLASS, OP, IS_RANGE)                                       \
     template <class type>                                                                          \
     void CLASS<type>::evaluate(vectorized::IColumn& column, uint16_t* sel, uint16_t* size) const { \
         uint16_t new_size = 0;                                                                     \
@@ -162,6 +162,9 @@ COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(GreaterEqualPredicate, >=)
                     auto* nested_col_ptr = vectorized::check_and_get_column<                       \
                             vectorized::ColumnDictionary<vectorized::Int32>>(nested_col);          \
                     auto code = nested_col_ptr->find_code(_value);                                 \
+                    if (code < 0 && IS_RANGE) {                                                    \
+                        code = nested_col_ptr->find_bound_code(_value, 0 OP 1, 1 OP 1 );           \
+                    }                                                                              \
                     auto& data_array = nested_col_ptr->get_data();                                 \
                     for (uint16_t i = 0; i < *size; i++) {                                         \
                         uint16_t idx = sel[i];                                                     \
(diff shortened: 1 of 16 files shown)


be/src/olap/rowset/segment_v2/binary_dict_page.cpp
```diff
@@ -132,7 +134,6 @@ void BinaryDictPageBuilder::reset() {
     } else {
         _data_page_builder->reset();
     }
-    _finished = false;
```
wangbo, line 135: Why remove finish here?

be/src/vec/columns/column_dictionary.h
```diff
@@ -0,0 +1,381 @@
...
+            for (uint32_t i = 0; i < dict_num; ++i) {
+                auto value = StringValue(dict_array[i].data, dict_array[i].size);
+                dict.insert_value(value);
+            }
+            dict_inited = true;
+        }
+
+        for (int i = 0; i < data_num; i++, start_index++) {
+            int32_t code = data_array[start_index];
+            insert_data(code);
```
wangbo, line 254: Using memcpy here can improve performance;
Now writing to this ColumnDictionary must be very slow

be/src/vec/columns/column_dictionary.h
```diff
@@ -0,0 +1,381 @@
...
+        LOG(FATAL) << "should not call replace_column_data in ColumnDictionary";
+    }
+
+    void replace_column_data_default(size_t self_row = 0) override {
+        LOG(FATAL) << "should not call replace_column_data_default in ColumnDictionary";
+    }
+
+    void insert_many_dict_data(const int32_t* data_array, size_t start_index,
+                               const StringRef* dict_array, size_t data_num,
+                               uint32_t dict_num) override {
```
wangbo, line 242: Currently ```insert_many_dict_data``` 's implementation is a little ugly.
May be we can make a temporary object(such as named dict_args)  to cover all the args and use dict_args as args.
And the dict_args is only used by insert_many_dict_data

be/src/olap/rowset/segment_v2/binary_dict_page.cpp
```diff
@@ -240,6 +241,16 @@ void BinaryDictPageDecoder::set_dict_decoder(PageDecoder* dict_decoder, StringRe
 
 Status BinaryDictPageDecoder::next_batch(size_t* n, vectorized::MutableColumnPtr &dst) {
     if (_encoding_type == PLAIN_ENCODING) {
+        auto* col_ptr = dst.get();
+        if (dst->is_nullable()) {
+            auto nullable_col = reinterpret_cast<vectorized::ColumnNullable*>(dst.get());
+            col_ptr = nullable_col->get_nested_column_ptr().get();
+        }
```
wangbo, line 248: I think it's a good idea to exposing ColumnDictionary to PageDecoder.
You should deal everything inside the ColumnDictionary.
Such as add a Method for ```IColumn``` named ```convertToCommonColumnIfNecessary```, or some other better way.
Of course, you can add a ```todo``` here.

be/src/olap/rowset/segment_v2/binary_dict_page.cpp
zenoyang: `_finished = false;` appears twice here

be/src/vec/columns/column_dictionary.h
zenoyang: Great idea, I'll optimize this right away

be/src/olap/rowset/segment_v2/binary_dict_page.cpp
zenoyang: ok

be/src/vec/columns/column_dictionary.h
zenoyang: ok, this is also a refactoring, add it to `todo`

be/src/runtime/string_value.h
```diff
@@ -60,10 +105,36 @@ struct StringValue {
     // this < other: -1
     // this == other: 0
     // this > other: 1
-    int compare(const StringValue& other) const;
+    int compare(const StringValue& other) const {
```
zbtzbtzbt, line 108: small function add key word `inline`，the same for following `eq`

be/src/runtime/string_value.h
```diff
@@ -60,10 +105,36 @@ struct StringValue {
...
+        int l = std::min(len, other.len);
+
+        if (l == 0) {
```
zbtzbtzbt, line 111: can we simplify this code to：
`if (l==0) return 0;`？

be/src/runtime/string_value.h
zenoyang: done

be/src/vec/columns/column_dictionary.h
```diff
@@ -0,0 +1,381 @@
...
+#include "vec/columns/column_impl.h"
+#include "vec/columns/column_string.h"
+#include "vec/columns/column_vector.h"
+#include "vec/columns/predicate_column.h"
+#include "vec/core/types.h"
+
+namespace doris::vectorized {
+
+/**
+ * For low cardinality string columns, using ColumnDictionary can reducememory
```
zbtzbtzbt, line 39: fix typo `reduce memory`

be/src/olap/comparison_predicate.cpp
```diff
@@ -145,28 +146,68 @@ COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(LessEqualPredicate, <=)
...
+                    new_size += _opposite ? !ret : ret;                                            \
+                }                                                                                  \
             }                                                                                      \
             *size = new_size;                                                                      \
+        } else if (column.is_column_dictionary()) {                                                \
+            if constexpr (std::is_same_v<type, StringValue>) {                                     \
+                auto& dict_col =                                                                   \
+                        reinterpret_cast<vectorized::ColumnDictionary<vectorized::Int32>&>(column);\
+                auto& data_array = dict_col.get_data();                                            \
+                auto code = dict_col.find_code(_value);                                            \
```
zbtzbtzbt, line 197: we better use `std::move`,  because dict_col.find_code return a large local var.
`auto code = std::move(dict_col.find_code(_value));  `

be/src/vec/columns/column_dictionary.h
```diff
@@ -0,0 +1,381 @@
...
+            size_t operator()(const StringValue& value) const {
+                return HashStringThoroughly(value.ptr, value.len);
+            }
+        };
+
+        StringValue::Comparator comparator;
+        // dict code -> dict value
+        DictContainer dict_data;
+        // dict value -> dict code
+        phmap::flat_hash_map<StringValue, T, HashOfStringValue> inverted_index;
```
wangshuo128, line 368: Why use a generic type for integer value? Do you meat to support int8, int32, int64, etc?

zenoyang: Because for low cardinality columns, int8 and int16 are often enough. But now the stored encoding is int32, so now the code uses int32 by default. Template is used to think that it can be optimized in the future. If you use less than 32, use 8 and 16.

be/src/olap/comparison_predicate.cpp
zenoyang: For function return value initialized variables,  can rely on the compiler's RVO optimization. 
Even create an int value doesn't cost much. Because it is outside the loop, temporary objects are not created frequently.
just my opinion.

be/src/vec/columns/column_dictionary.h
zenoyang: done

be/src/vec/columns/column_dictionary.h
```diff
@@ -0,0 +1,381 @@
...
+
+        void clear() {
+            dict_data.clear();
+            inverted_index.clear();
+            code_convert_map.clear();
+        }
+
+        void sort() {
+            size_t dict_size = dict_data.size();
+            std::sort(dict_data.begin(), dict_data.end(), comparator);
```
wangbo, line 346: How cost is sorting here?

be/src/vec/columns/column_dictionary.h
```diff
@@ -0,0 +1,381 @@
...
+        // dict code -> dict value
+        DictContainer dict_data;
+        // dict value -> dict code
+        phmap::flat_hash_map<StringValue, T, HashOfStringValue> inverted_index;
+        // data page code -> sorted dict code, only used for range comparison predicate
+        phmap::flat_hash_map<T, T> code_convert_map;
+    };
+
+private:
+    bool dict_inited = false;
```
wangbo, line 374: For Doris's code style, Class's field should add a prefix.
Such as
```
bool _dict_inited = false;
```

be/src/vec/columns/column_dictionary.h
```diff
@@ -0,0 +1,381 @@
...
+        inline T find_code(const StringValue& value) const {
+            auto it = inverted_index.find(value);
+            if (it != inverted_index.end()) {
+                return it->second;
+            }
+            return -1;
+        }
+
+        inline T find_bound_code(const StringValue& value, bool lower, bool eq) const {
+            if (lower) {
```
wangbo, line 318: Should we need to check whether value exists in dict_data?

be/src/runtime/string_value.h
```diff
@@ -22,9 +22,53 @@
...
+// that the trivial cases are already handled (i.e. one string is empty).
+// Returns:
+//   < 0 if s1 < s2
+//   0 if s1 == s2
+//   > 0 if s1 > s2
+// The SSE code path is just under 2x faster than the non-sse code path.
+//   - s1/n1: ptr/len for the first string
+//   - s2/n2: ptr/len for the second string
+//   - len: min(n1, n2) - this can be more cheaply passed in by the caller
+static inline int string_compare(const char* s1, int64_t n1, const char* s2, int64_t n2,
```
wangbo, line 43: Why move it here?

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -856,6 +857,18 @@ void SegmentIterator::_evaluate_short_circuit_predicate(uint16_t* vec_sel_rowid_
     for (auto column_predicate : _short_cir_eval_predicate) {
         auto column_id = column_predicate->column_id();
         auto& short_cir_column = _current_return_columns[column_id];
+        auto* col_ptr = short_cir_column.get();
```
wangbo, line 860: Please add a todo for code refactor here.

be/src/olap/in_list_predicate.cpp
```diff
@@ -122,21 +123,56 @@ IN_LIST_PRED_COLUMN_BLOCK_EVALUATE(NotInListPredicate, ==)
...
-                sel[new_size] = idx;                                                               \
-                const type& cell_value = reinterpret_cast<const type&>(data_array[idx]);           \
-                bool ret = !null_bitmap[idx] && (_values.find(cell_value) OP _values.end());       \
-                new_size += _opposite ? !ret : ret;                                                \
+            auto* nullable_col =                                                                   \
+                    vectorized::check_and_get_column<vectorized::ColumnNullable>(column);          \
+            auto& null_bitmap = reinterpret_cast<const vectorized::ColumnUInt8&>(                  \
+                                        nullable_col->get_null_map_column()).get_data();           \
+            auto& nested_col = nullable_col->get_nested_column();                                  \
+            if (nested_col.is_column_dictionary()) {                                               \
```
wangbo, line 131: Too many branch, need a todo for code refactor here.

be/src/olap/comparison_predicate.cpp
```diff
@@ -145,28 +146,68 @@ COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(LessEqualPredicate, <=)
...
+                if (code < 0 && IS_RANGE) {                                                        \
+                    code = dict_col.find_bound_code(_value, 0 OP 1, 1 OP 1);                       \
```
wangbo, line 199: If there is a range predicate here, we just need to use string sort rule to sort dict code.
This means ```data_array``` has the same order with string value.
Finally we can compare value and data_array directly here.
So why need ```find_bound_code``` here?

be/src/vec/columns/column_dictionary.h
zenoyang: There is no specific test for sort. However, according to the SSB single-table Q2.2 query (the dictionary size is 1000), after sorting and storing codes, it is faster than the original method.

The original way:
- ReadColByIdxTime: 2s957ms
- ShortPredEvalTime: 5s10ms

After low cardinality optimization:
- ReadColByIdxTime: 2s556ms      (Dictionary sorting at this stage)
- ShortPredEvalTime: 481.672ms

be/src/vec/columns/column_dictionary.h
zenoyang: done

be/src/olap/rowset/segment_v2/segment_iterator.cpp
zenoyang: done

be/src/vec/columns/column_dictionary.h
zenoyang: Check before function call. I refactored it to check inside the function.

be/src/runtime/string_value.h
zenoyang: Because of the error in the link stage: The impl of `compare(const StringValue& other)` could not be found, but I don't know the reason, I will just move it to the string_value.h

be/src/olap/comparison_predicate.cpp
zenoyang: For range comparison predicates, predicate constants in query sql may not be in the dictionary.

For example, the currently sorted dictionary is `a:0, c:1, e:2`, the predicate in sql is: `col > 'b' and col < 'd'`, `b` and `d` are not in the dictionary, but the result `c` needs to be returned , this is what `find_bound_code` does.

wangbo (dismissed):
LGTM

be/src/olap/comparison_predicate.cpp
```diff
@@ -145,28 +146,65 @@ COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(LessEqualPredicate, <=)
...
+                    new_size += _opposite ? !ret : ret;                                            \
+                }                                                                                  \
             }                                                                                      \
             *size = new_size;                                                                      \
+        } else if (column.is_column_dictionary()) {                                                \
+            if constexpr (std::is_same_v<type, StringValue>) {                                     \
+                auto& dict_col =                                                                   \
+                        reinterpret_cast<vectorized::ColumnDictionary<vectorized::Int32>&>(column);\
+                auto& data_array = dict_col.get_data();                                            \
+                auto code = IS_RANGE ? dict_col.find_bound_code(_value, 0 OP 1, 1 OP 1)            \
```
HappenLee, line 196: why we do not dispose value before really do predicate, not need hash string value again and again, to speed up query

be/src/olap/comparison_predicate.cpp
```diff
@@ -145,28 +146,65 @@ COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(LessEqualPredicate, <=)
...
-                            nullable_column->get_nested_column());                                 \
-            auto& data_array = nest_column_vector->get_data();                                     \
-            for (uint16_t i = 0; i < *size; i++) {                                                 \
-                uint16_t idx = sel[i];                                                             \
-                sel[new_size] = idx;                                                               \
-                const type& cell_value = reinterpret_cast<const type&>(data_array[idx]);           \
-                bool ret = !null_bitmap[idx] && (cell_value OP _value);                            \
-                new_size += _opposite ? !ret : ret;                                                \
+            auto& nested_col = nullable_col->get_nested_column();                                  \
+            if (nested_col.is_column_dictionary()) {                                               \
```
HappenLee, line 161: if `nested_col.is_column_dictionary()` is true, the `std::is_same_v<type, StringValue>` is always true. so here should be 
`if (nested_col.is_column_dictionary() && std::is_same_v<type, StringValue>)`

be/src/olap/in_list_predicate.cpp
```diff
@@ -117,26 +118,62 @@ IN_LIST_PRED_EVALUATE(NotInListPredicate, ==)
...
```
HappenLee, line 132: same to up

be/src/olap/in_list_predicate.cpp
```diff
@@ -117,26 +118,62 @@ IN_LIST_PRED_EVALUATE(NotInListPredicate, ==)
...
+                if constexpr (std::is_same_v<type, StringValue>) {                                 \
+                    auto* nested_col_ptr = vectorized::check_and_get_column<                       \
+                            vectorized::ColumnDictionary<vectorized::Int32>>(nested_col);          \
+                    auto code_set = nested_col_ptr->find_codes(_values);                           \
```
HappenLee, line 136: same to up

be/src/olap/comparison_predicate.cpp
zenoyang: Great idea, I'm about to improve it too, by rewriting the predicate so that it runs vectorized predicate computations.

zenoyang: Great idea, I'm about to improve it too, by rewriting the predicate so that it runs vectorized predicate computations.

be/src/olap/comparison_predicate.cpp
zenoyang: I thought about it, but the compilation will report an error, because the `find_code` function requires that `_value` must be of type StringValue, so it must be specified at compile time through `if constexpr` that it must be of type stringValue

be/src/vec/columns/column_dictionary.h
```diff
@@ -0,0 +1,386 @@
...
+ * according to the dictionary, so that encoding comparisons are used instead
+ * of string comparisons to improve performance.
+ * For range comparison predicates, it is necessary to sort the dictionary
+ * contents, convert the encoding column, and then compare the encoding directly.
+ * If the read data page contains plain-encoded data pages, the dictionary
+ * columns are converted into PredicateColumn for processing.
+ * Currently ColumnDictionary is only used for storage layer.
+ */
+template <typename T>
+class ColumnDictionary final : public COWHelper<IColumn, ColumnDictionary<T>> {
```
HappenLee, line 51: static_assert T is singed integer

be/src/vec/columns/column_dictionary.h
```diff
@@ -0,0 +1,386 @@
...
+            }
+        };
+
+        StringValue::Comparator comparator;
+        // dict code -> dict value
+        DictContainer dict_data;
+        // dict value -> dict code
+        phmap::flat_hash_map<StringValue, T, HashOfStringValue> inverted_index;
+        // data page code -> sorted dict code, only used for range comparison predicate
+        phmap::flat_hash_map<T, T> code_convert_map;
```
HappenLee, line 375: why need the struct ? seems useless?

be/src/vec/columns/column_dictionary.h
```diff
@@ -0,0 +1,386 @@
...
+        size_t byte_size() { return dict_data.size() * sizeof(dict_data[0]); }
+
+    private:
+        struct HashOfStringValue {
+            size_t operator()(const StringValue& value) const {
+                return HashStringThoroughly(value.ptr, value.len);
+            }
+        };
+
+        StringValue::Comparator comparator;
```
HappenLee, line 369: the param in class why not use _dict_data?

zenoyang: I have refactored and haven't committed the code yet. wait for this pr to merge.

be/src/vec/columns/column_dictionary.h
zenoyang: ok, add it to the todo list

HappenLee (approved):
LGTM

be/src/vec/columns/column_dictionary.h
zenoyang: `DictContainer _dict_data` is an array. dict_code obtained from data_page is the index of this array. When the order changes after the dictionary is sorted, the dict_code must also be converted, so `code_convert_map` is introduced.

PR #8336, opened 2022-03-04
[improvment](broker load) fix broker load check logic

### LINKED ISSUE #8335: [Enhancement] broker load should support regular expressions when type is like hdfs
### Search before asking

### Description

in verison 0.12.21, broker load allowed users to write path with regular expressions. for example:
```
LOAD LABEL test.test_load_01(
DATA INFILE ('hdfs://ip:port/test/test/data/{20220131,20220201}/*')
...
);
```
but in version 0.15, it is not allowed.

According to hdfs, the path could be write in regular expressions. So doris should support it.

And if the path is not valid, exceptions can be thrown as expected.

### Solution

broker load should support regular expressions when type is like hdfs

### Code of Conduct

Henry2SS:
# Proposed changes

Issue Number: close #8335 

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/ExportStmt.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/ExportStmt.java
@@ -232,16 +232,17 @@ public class ExportStmt extends StatementBase {
         if (Strings.isNullOrEmpty(path)) {
             throw new AnalysisException("No dest path specified.");
         }
-
+        if (type == StorageBackend.StorageType.BROKER) {
+            if (!path.toLowerCase().contains("bos") && !path.toLowerCase().contains("afs")
+                    && !path.toLowerCase().contains("hdfs")) {
+                throw new AnalysisException("Invalid export path. please use valid 'HDFS://', 'AFS://' or 'BOS://' path.");
+            }
+            return path;
+        }
         try {
             URI uri = new URI(path);
             String schema = uri.getScheme();
-            if (type == StorageBackend.StorageType.BROKER) {
-                if (schema == null || (!schema.equalsIgnoreCase("bos") && !schema.equalsIgnoreCase("afs")
-                    && !schema.equalsIgnoreCase("hdfs"))) {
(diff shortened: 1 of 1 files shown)


fe/fe-core/src/main/java/org/apache/doris/analysis/ExportStmt.java
```diff
@@ -232,16 +232,17 @@ public static String checkPath(String path, StorageBackend.StorageType type) thr
...
             throw new AnalysisException("No dest path specified.");
         }
-
+        if (type == StorageBackend.StorageType.BROKER) {
+            if (!path.toLowerCase().contains("bos") && !path.toLowerCase().contains("afs")
+                    && !path.toLowerCase().contains("hdfs")) {
+                throw new AnalysisException("Invalid export path. please use valid 'HDFS://', 'AFS://' or 'BOS://' path.");
+            }
+            return path;
+        }
```
yangzhg, line 241: How  about a path  `nvalid_schema://xxbos/xxafs_path/path-contians-hdfs`

Henry2SS:
> 

It is. 
URI not support '{}' '[]'. And when the path contains '{}' '[]', `URISyntaxException` will be thrown out. 
This is the reason for the problem described in #8335.

Henry2SS:
> Seems like the code you changed in this PR does not related the problem you described in #8335 ?

It is.
URI doesn't support '{}' '[]', and when a path contains '{}', '[]', a `URISyntaxException` will be thrown out.
It is the reason for the problem described in #8335.

fe/fe-core/src/main/java/org/apache/doris/analysis/ExportStmt.java
Henry2SS: `checkPath` only do a pre-check, it checks the input style.
If the path is not a valid one, an error will be thrown out as expected.
```
 type:ETL_RUN_FAIL; msg:errCode = 2, detailMessage = Broker list path failed.
```

And maybe change `path.contains` to `path.startsWith`, is it better? What's your opinions?

PR #8352, opened 2022-03-06
[improvement]support build with parallel parameter only

dataalive:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

when sh build.sh -j n, which is followed by parallel parameter only , the output will be null because the script must follow --be or --fe etc... explicitly

so, the behavior is not the same as make -j n.

with this pr, we will support this.

## Checklist(Required)

## Further comments

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -120,6 +120,8 @@ BUILD_UI=
 BUILD_SPARK_DPP=
 CLEAN=
 HELP=0
+PARAMETER_COUNT=$#
+PARAMETER_FLAG=0
 if [ $# == 1 ] ; then
     # default
     BUILD_BE=1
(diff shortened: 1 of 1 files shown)


hf200012 (approved):
LGTM

yiguolei:
Hi everybody, I think we should think about this feature.
I think we should set the parallel parameter to half of the build machine's CPU core. Because many user feel build process is too slow and we have to told them that there is parameter to speed up. But many user will not ask for help and they will treat the slow build process as normal.

morningman:
> Hi everybody, I think we should think about this feature. I think we should set the parallel parameter to half of the build machine's CPU core. Because many user feel build process is too slow and we have to told them that there is parameter to speed up. But many user will not ask for help and they will treat the slow build process as normal.

Currently, the default value is 1/4 of cpu core. Using 1/2 may cause OOM exception(I am not really sure about that).
It there any "best practice" that we can refer to?

yiguolei:
> 
If too many build thread will cause OOM, then it is ok for me. And I will track this issue.

PR #8364, opened 2022-03-07
[Function] Add compare_version function and ut

carlvinhust2012:
# Proposed changes
Add compare_version function and ut. the function will return the compared result of two version.
Issue Number: close #xxx

## Problem Summary:

1.Add compare_version function and ut.
2.add md files for this compare_version function.

## Checklist(Required)

4. Does it need to update dependencies:No

## Further comments

### DIFF
--- a/be/src/exprs/utility_functions.cpp
+++ b/be/src/exprs/utility_functions.cpp
@@ -39,4 +39,82 @@ BooleanVal UtilityFunctions::sleep(FunctionContext* ctx, const IntVal& seconds) 
     return BooleanVal(true);
 }
 
+static void convert_to_string(doris_udf::StringVal& src, std::string& str_dst) {
+    for (int i = 0; i < src.len; i++) {
+        str_dst += src.ptr[i];
+    }
+    return;
+}
+
+static BooleanVal split_string(std::string& str_src, std::vector<std::string>& str_vec_dst) {
+    std::string s;
+    std::stringstream ss;
+    ss << str_src;
+    while (std::getline(ss, s, '.')) {
+        // return when find a non-digital separated by "." , such as all.8.9 or Chinese, etc.
+        if (s.find_first_not_of("0123456789") != std::string::npos) {
+            return BooleanVal(false);
+        }
+        str_vec_dst.push_back(s);
+    }
+    return BooleanVal(true);
+}
+
+static BooleanVal compare_splitted_string(std::vector<std::string>& v1, 
+                                          std::vector<std::string>& v2,
+                                          std::string& str_op) {
+    if (v1.size() == 0 || v2.size() == 0) {
+        return BooleanVal(false);
+    }
+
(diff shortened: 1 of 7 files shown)


be/src/exprs/utility_functions.cpp
```diff
@@ -39,4 +39,82 @@ BooleanVal UtilityFunctions::sleep(FunctionContext* ctx, const IntVal& seconds)
     return BooleanVal(true);
 }
 
+static void convert_to_string(doris_udf::StringVal& src, std::string& str_dst) {
```
yangzhg, line 42: why called `convert_to_string` ？ just a concat ?

be/src/exprs/utility_functions.cpp
```diff
@@ -39,4 +39,82 @@ BooleanVal UtilityFunctions::sleep(FunctionContext* ctx, const IntVal& seconds)
...
+    for (int i = 0; i < src.len; i++) {
+        str_dst += src.ptr[i];
+    }
+    return;
+}
+
+static BooleanVal split_string(std::string& str_src, std::vector<std::string>& str_vec_dst) {
```
yangzhg, line 49: there is already have split in the util, refer to be/src/util/string_util.h or gutils

yangzhg (commented):
Why not `compare_version(ver1 , ver2)`  return 0 if equal, 1 if ver1 > ver2, -1 if ver1 < ver2 ？
In addition, there is still a lot of room for optimization in the implementation of the program

be/src/exprs/utility_functions.cpp
carlvinhust2012: just want to convert the type from 'StringVal' to 'string'.

be/src/exprs/utility_functions.cpp
carlvinhust2012: I will re-check this split function.

carlvinhust2012:
> Why not `compare_version(ver1 , ver2)` return 0 if equal, 1 if ver1 > ver2, -1 if ver1 < ver2 ？ In addition, there is still a lot of room for optimization in the implementation of the program

carlvinhust2012:
> Why not `compare_version(ver1 , ver2)` return 0 if equal, 1 if ver1 > ver2, -1 if ver1 < ver2 ？ In addition, there is still a lot of room for optimization in the implementation of the program

We discussed this approach with the business, but they still prefer to use the current approach.

carlvinhust2012:
> > Why not `compare_version(ver1 , ver2)` return 0 if equal, 1 if ver1 > ver2, -1 if ver1 < ver2 ？ In addition, there is still a lot of room for optimization in the implementation of the program

We discussed this approach with the business, but they still prefer to use the current approach.

carlvinhust2012:
the customized versions have been provided to customer, so this pr could be closed.

PR #8399, opened 2022-03-08
[improvement]avoid read bitmap index when no index works

### LINKED ISSUE #8388: [Enhancement] Avoid read index bitmap when no index is provided
### Search before asking

### Description

When a query hits index, then index bitmap is reset;
Storage layer read rows by index bitmap.
In the current implementation, even a query doesn't hit any index, storage still need to read rows by index bitmap.
This could bring two costs:
1 Read data from roaring bitmap.
2 The foreach to read rowids from roaring bitmap.

There is still optimization room for query without indexes.

### Solution

_No response_

### Code of Conduct

wangbo:
# Proposed changes

Issue Number: close #8388

## Problem Summary:

## Checklist(Required)

## Further comments

Performance test has been added, see #8388

### DIFF
--- a/be/src/olap/rowset/segment_v2/segment_iterator.cpp
+++ b/be/src/olap/rowset/segment_v2/segment_iterator.cpp
@@ -53,9 +53,18 @@ public:
 
     bool has_more_range() const { return !_eof; }
 
+    template<ReadIndexType T>
+    bool next_range(uint32_t max_range_size, uint32_t* from, uint32_t* to) {
+        if constexpr (T == READ_WITH_BITMAP) {
+            return get_next_range_with_bitmap_index(max_range_size, from, to);
+        } else if constexpr (T == READ_WITHOUT_BITMAP) {
+            return get_next_range_without_bitmap_index(max_range_size, from, to);
+        }
+    }
+
     // read next range into [*from, *to) whose size <= max_range_size.
     // return false when there is no more range.
-    bool next_range(uint32_t max_range_size, uint32_t* from, uint32_t* to) {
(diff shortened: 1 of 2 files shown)


be/src/olap/rowset/segment_v2/segment_iterator.h
```diff
@@ -45,6 +45,11 @@ class BitmapIndexIterator;
 class BitmapIndexReader;
 class ColumnIterator;
 
+enum ReadIndexType {
+    READ_WITH_BITMAP,
+    READ_WITHOUT_BITMAP,
```
zbtzbtzbt, line 50: we can remove the end `,`

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -76,6 +85,27 @@ class SegmentIterator::BitmapRangeIterator {
         return true;
     }
 
+    bool get_next_range_without_bitmap_index(uint32_t max_range_size, uint32_t* from, uint32_t* to) {
+        if (_curr_pos == _max_value) {
+            return false;
+        }
+        if (_curr_pos + max_range_size <= _max_value) {
```
zbtzbtzbt, line 92: should we add `_eof` like `get_next_range_with_bitmap_index()`? ( i am not sure)

wangbo: ```_eof``` is used to confirm whether reachs the end of ```roaring_read_uint32_iterator```.
If you don't use ```roaring_read_uint32_iterator```, then ```_eof``` is useless.

zbtzbtzbt: i get it

PR #8436, opened 2022-03-10
[improvement] Avoid print large string in error log

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

1. Avoid print large string in error log
    If user load a unqualified large string, the all string will be saved in error log,
    so the error log is too big that can not be shown be using `show load warnings on "url"`.
    Err: `Got packet bigger than 'max_allowed_packet' bytes`

2. Remove duplicate help doc
    Do not allow doc with same title, or error thrown when starting FE:
    `java.lang.IllegalArgumentException: Multiple entries with same key:`

## Checklist(Required)

## Further comments


EmmyMiao87 (approved):
LGTM

PR #8438, opened 2022-03-11
[Feature] [vectorized]support vectorized compaction

starocean999:
# Proposed changes

Issue Number: [#8445](https://github.com/apache/incubator-doris/issues/8445)

## Problem Summary:

support vectorized compaction

## Checklist(Required)

## Further comments


morningman:
Hi @starocean999 , thanks for your contribution.
1. Please take a look at: http://doris.incubator.apache.org/zh-CN/community/how-to-contribute/commit-format-specification.html  to change your PR title
2. Need to create an issue to describe:
    1. What problem it is?
    2. How it works?

HappenLee:
Also need a clear performance report about origin compation and vec compatcion.

be/src/vec/olap/olap_data_convertor.h
```diff
@@ -0,0 +1,256 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+#pragma once
+#include "olap/tablet_schema.h"
+#include "vec/core/block.h"
+
+namespace doris::vectorized {
+
+struct OlapFieldData {
```
HappenLee, line 24: why use value == nullpt as null value?

be/src/vec/olap/olap_data_convertor.h
```diff
@@ -0,0 +1,256 @@
...
+                    typed_column, row_pos, num_rows);
+            m_values.resize(num_rows);
+        }
+
+        const void* get_data() const override { return m_values.data(); }
+
+        OlapFieldData get_data_at(size_t offset) const override {
+            assert(offset < m_num_rows && m_num_rows == m_values.size());
+            UInt8 null_flag = 0;
+            auto null_map = get_nullmap();
```
HappenLee, line 198: each time have a virtual function call ? why not keep the `_null_map` ptr as Member variable?

be/src/vec/olap/olap_data_convertor.h
```diff
@@ -0,0 +1,256 @@
...
+                        // do nothing
+                    }
+                    ++value;
+                    ++data_cur;
+                    ++nullmap_cur;
+                }
+                assert(nullmap_cur == nullmap + m_row_pos + m_num_rows &&
+                       value == m_values.get_end_ptr());
+            } else {
+                while (data_cur != data_end) {
```
HappenLee, line 238: maybe use memcpy could be more elegant and efficient

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -88,6 +88,75 @@ OLAPStatus BetaRowsetWriter::init(const RowsetWriterContext& rowset_writer_conte
...
+    refresh_segment_capacity();
+    if (UNLIKELY(segment_capacity_in_bytes < row_avg_size_in_bytes ||
+                      segment_capacity_in_rows <= 0)) {
+        // no space for another signle row, need flush now
+        RETURN_NOT_OK(_flush_segment_writer(&_segment_writer));
+        RETURN_NOT_OK(_create_segment_writer(&_segment_writer));
+        refresh_segment_capacity();
+    }
+
+    assert(segment_capacity_in_bytes > row_avg_size_in_bytes && segment_capacity_in_rows > 0);
```
Gabriel39, line 120: In extreme cases, `row_avg_size_in_bytes` could be larger than `segment_capacity_in_bytes`. In old version `add_row` function, it will be written in a single segment which contains only one row. But in this new function, it seems like this row will never be processed.

Gabriel39: Btw, size of a segment file will be at least MAX_SEGMENT_SIZE in old way however at most MAX_SEGMENT_SIZE in this new way. So I just wonder should we keep origin behavior?

be/src/olap/rowset/segment_v2/segment_writer.cpp
```diff
@@ -96,6 +102,78 @@ Status SegmentWriter::init(uint32_t write_mbytes_per_sec __attribute__((unused))
...
+    }
+    
+    _row_count += num_rows;
+    _olap_data_convertor.clear_source_content();
+    return Status::OK();
+}
+
+std::string SegmentWriter::encode_short_keys(
+        const std::vector<const void*> key_column_fields, bool null_first) {
+    size_t num_key_columns = _tablet_schema->num_short_key_columns();
```
Gabriel39, line 154: how about to remove `num_key_columns` and just use `key_column_fields.size()` instead

PR #8451, opened 2022-03-11
[improvement](memory) fix olap table scan and sink memory usage problem

xiaokang:
# Proposed changes

Issue Number: NA

## Problem Summary:

Due to unlimited queue in OlapScanNode and NodeChannel, memory usage can be very large for reading and writting large table, e.g 'insert into tableB select * from tableA'.

## Checklist(Required)

## Further comments

1. Add bytes limit (2.5% of query_options.mem_limit) for _scan_row_batches queue in OlapScanNode. There was no limit for _scan_row_batches yet.
2. Add bytes limit (2.5% of exec_mem_limit) for _materialized_row_batches queue in OlapScanNode. There was only queue size limit for _materialized_row_batches before, and it's not enough for data with large average bytes of a single row.
3. Add bytes limit for reading a single batch. A new be config doris_scanner_row_bytes is added and its default value is 10MB. There was only row number limit before, and it's also not enough for data with large average bytes of a single row.
5. Add bytes limit (5% of load_mem_limit) for _max_pending_batches queue in NodeChannel. 
6. Fixed an exceptional zero value of max_thread in OlapScanNode.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -169,8 +169,10 @@ CONF_mInt32(doris_scan_range_row_count, "524288");
 CONF_mInt32(doris_scan_range_max_mb, "0");
 // size of scanner queue between scanner thread and compute thread
 CONF_mInt32(doris_scanner_queue_size, "1024");
-// single read execute fragment row size
+// single read execute fragment row number
 CONF_mInt32(doris_scanner_row_num, "16384");
+// single read execute fragment row bytes
+CONF_mInt32(doris_scanner_row_bytes, "10485760");
 // number of max scan keys
 CONF_mInt32(doris_max_scan_key_num, "1024");
 // the max number of push down values of a single column.
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -77,6 +77,8 @@ Status OlapScanNode::init(const TPlanNode& tnode, RuntimeState* state) {
         _max_pushdown_conditions_per_column = config::max_pushdown_conditions_per_column;
     }
 
+    _max_scanner_queue_size_bytes = query_options.mem_limit / 20; //TODO: session variable percent
+
     /// TODO: could one filter used in the different scan_node ?
     int filter_size = _runtime_filter_descs.size();
     _runtime_filter_ctxs.resize(filter_size);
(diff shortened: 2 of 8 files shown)


morningman:
LGTM, but there are 2 more things:
1. It is better to provide some data to illustrate the memory usage before and after the modification in order to give more information to the reviewer.
2. Also need to modify the `src/vec/exec/volap_scan_node.cpp`

xiaokang:
Hi @morningman, Thanks for your quick response!

> 
> 1. It is better to provide some data to illustrate the memory usage before and after the modification in order to give more information to the reviewer.

Test environment:
A single node cluster with 32g memory and 8 cpu cores. All configuration is default.

Table schema:
CREATE TABLE `tableA` (
  `uuid` text NULL,
  `no` text NULL,
  `f` text NULL,
  `m` text NULL,
  `data` text NULL,
  `time1` datetime NULL,
  `time2` datetime NULL,
  `p` text NOT NULL
) ENGINE=OLAP
DUPLICATE KEY(`uuid`)
PARTITION BY LIST(`p`)
(PARTITION p1 VALUES IN ("1"),
...
PARTITION p98 VALUES IN ("98"),
PARTITION p99 VALUES IN ("99"))
DISTRIBUTED BY HASH(`uuid`) BUCKETS 10
PROPERTIES (
"in_memory" = "false",
"storage_format" = "V2"
)

Test steps:
1. create tableA and tableB with 100 partition.
2. insert 1 million rows with large average size 100KB to tableA.
3. execute query: 'set batch_size = 128; set exec_mem_limit=8147483648; insert into tableB select * from table A;'

Result:
- master branch: be process quickly consume all 32g memory in about 20 seconds and killed by Linux oom-killer.
- fix branch: The memory usage of be process is 14g memory at peak and fall back to 8g when query finished. 8g is for be storage page cache (32g*20%=6g) and chunk_reserved_bytes (2g).

> 2. Also need to modify the `src/vec/exec/volap_scan_node.cpp`

I will try to do the same things for the vectorized code.

xiaokang:
@morningman volap_scan_node.cpp is done. The test result is almost the same as non-vectorized version.

morningman (approved):
LGTM

PR #8548, opened 2022-03-21
fix sort spill memory limit

xiaokang:
# Proposed changes

Issue Number: NA

## Problem Summary:

Sort spill does not work since spill is triggered when memory limit exceed but some more memory need to be allocated during spilling.

This pr fix sort spill memory limit by:
1.change block_mgr_limit from exec_mem_limit to half of block_mgr_limit
2.trigger spill when 1/2 of block_mgr_limit is used, implemented in should_spill() function
## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/runtime/buffered_block_mgr2.cc
+++ b/be/src/runtime/buffered_block_mgr2.cc
@@ -656,6 +656,7 @@ Status BufferedBlockMgr2::pin_block(Block* block, bool* pinned, Block* release_b
                             << endl
                             << block->debug_string();
                     _free_io_buffers.remove(block->_buffer_desc);
+                    _free_io_buffers_bytes -= block->_buffer_desc->len;
                 } else if (_unpinned_blocks.contains(block)) {
                     _unpinned_blocks.remove(block);
                 } else {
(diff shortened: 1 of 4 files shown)


xiaokang:
Hi @xinyiZzz , Thanks for you comment.

The purpose of the first 1/2 is to reserve half of exec_mem_limit for other components rather than block_mgr. The purpose of the second 1/2 is to trigger spill before block_mgr_limit is full. Maybe twice 1/2 seems to be conservative.

xinyiZzz:
> The purpose of the first 1/2 is to reserve half of exec_mem_limit for other components rather than block_mgr. The purpose of the second 1/2 is to trigger spill before block_mgr_limit is full. Maybe twice 1/2 seems to be conservative.

The first 1/2 I understand, great~, but I'm not sure if 1/2 is a good value.

The role of the second 1/2. Currently, `BufferedBlockMgr2` has a similar implementation. In `write_unpinned_blocks`, unpinned blocks will be spilled in advance to avoid waiting when full.
image

xinyiZzz:
The first 1/2 may be OK, you can add a TODO, and then we may need a complete Query memory estimation model.

PR #8553, opened 2022-03-21
[feature](balance) Support balance between disks on a single BE

### LINKED ISSUE #8550: [Feature] support disk rebalance
### Search before asking

### background
Current situation of doris is that the cluster is balanced, but the disks of a backend may be unbalanced.
for example, backend  A have two disks: disk1 and disk2, disk1's usage is 98%, but disk2's usage is only 40%. disk1 is unable to take more data, therefore only one disk of backend A can take new data, the avaiable write throughput of backend A is only half of its ability, and we can not resolve this through load or partition rebalace now.

so we introduce disk rebalancer, disk rebalancer is different from other rebalancer(load or partition) which take care of cluster-wide data balancing. it takes care about backend-wide data balancing.

### schedule strategies:
1. normally, disk rebalance task will be scheduled when cluster is balanced .
2. if user has set  some backends that have high priority of disk rebalance, then disk rebalance task of the prior backends will be scheduled, no matter if cluster is balanced.

#### defintion:
1. high load disk/path: 10% greater than avg disk usage
2. mid load disk/path: -10% < avg disk usage <+10%
3. low load disk/path: 10% less than avg disk usage

#### strategies:
1. if we have **high** load disks:
source : 
we only pick a tablet from **high** load disks
dest disk:
if we have **low** load disks, we only pick a disk from **low** load disks
else we will pick a disk from **mid** load disks

2. if we only have **mid** and **low** load disks:
source : 
we only pick a tablet from **mid** load disks
dest disk:
we only pick disk from **low** load disks

3. if we only have **mid** load disks:
it means backend is balanced, we have nothing to do.

#### 1. set some backends have high priority of disk rebalance
ADMIN REBALANCE DISK [ON ("BackendHost1:BackendHeartBeatPort1", "BackendHost2:BackendHeartBeatPort2", ...)];

#### 2. cancel high priority of disk rebalance of backends
ADMIN CANCEL REBALANCE DISK [ON ("BackendHost1:BackendHeartBeatPort1", "BackendHost2:BackendHeartBeatPort2", ...)];

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

yinzhijian:
# Proposed changes

Issue Number: close #8550

## Problem Summary:

Current situation of doris is that the cluster is balanced, but the disks of a backend may be unbalanced.
for example, backend A have two disks: disk1 and disk2, disk1's usage is 98%, but disk2's usage is only 40%. disk1 is unable to take more data, therefore only one disk of backend A can take new data, the avaiable write throughput of backend A is only half of its ability, and we can not resolve this through load or partition rebalace now.

So we introduce disk rebalancer, disk rebalancer is different from other rebalancer(load or partition) which take care of cluster-wide data balancing. it takes care about backend-wide data balancing.

[For more details see #8550](https://github.com/apache/incubator-doris/issues/8550) 

## Checklist(Required)

## Further comments


fe/fe-core/src/main/java/org/apache/doris/task/StorageMediaMigrationTask.java
```diff
@@ -36,9 +38,20 @@ public StorageMediaMigrationTask(long backendId, long tabletId, int schemaHash,
 
     public TStorageMediumMigrateReq toThrift() {
         TStorageMediumMigrateReq request = new TStorageMediumMigrateReq(tabletId, schemaHash, toStorageMedium);
+        if (dataDir != null && !dataDir.isEmpty()) {
```
morningman, line 41: Strings.isNullOrEmpty(dataDir)

caiconghui:
Is it possible that the work of disk balancing is only scheduled within the scope of be and fe does not need to handle？for fe，it has taken on too many responsibilities

yiguolei:
agree with @caiconghui , BE should have local balance mechanism.

yinzhijian:
@caiconghui @yiguolei 
It does increase FE's responsibilities, but putting it in FE has its benefits:
1. It can control disk balance globally and avoid competing with recovery and cluster balancing.
2. It reuses the Rebalance mechanism of FE, and  there is no need to add a new mechanism for FE. If it is implemented in BE, a set of balance mechanism needs to be repeatedly designed without additional income.

fe/fe-core/src/main/java/org/apache/doris/clone/TabletScheduler.java
```diff
@@ -1189,14 +1229,39 @@ private void selectTabletsForBalance() {
...
-        rebalancer.createBalanceTask(tabletCtx, backendsWorkingSlots, batchTask);
+        AgentTask task = null;
+        if (tabletCtx.getBalanceType() == TabletSchedCtx.BalanceType.DISK_BALANCE) {
+            task = diskRebalancer.createBalanceTask(tabletCtx, backendsWorkingSlots);
+            checkDiskBalanceLastSuccTime(tabletCtx.getSrcBackendId(), tabletCtx.getSrcPathHash());
+            checkDiskBalanceLastSuccTime(tabletCtx.getDestBackendId(), tabletCtx.getDestPathHash());
+        } else if (tabletCtx.getBalanceType() == TabletSchedCtx.BalanceType.BE_BALANCE) {
+            task = rebalancer.createBalanceTask(tabletCtx, backendsWorkingSlots);
+        } else {
+            throw new SchedException(Status.SCHEDULE_FAILED, "unknown balance type");
```
yangzhg, line 1262: it better to print  tabletCtx.getBalanceType() in log

fe/fe-core/src/main/java/org/apache/doris/clone/Rebalancer.java
```diff
@@ -93,4 +101,20 @@ public Long getToDeleteReplicaId(TabletSchedCtx tabletCtx) {
     public void updateLoadStatistic(Table<String, Tag, ClusterLoadStatistic> statisticMap) {
         this.statisticMap = statisticMap;
     }
+
+    public void addPrioBackends(List<Backend> backends, long timeoutS) {
+        for (Backend backend : backends) {
+            prioBackends.put(backend.getId(), System.currentTimeMillis() + timeoutS);
```
yangzhg, line 107: cache System.currentTimeMillis() outside for loop

caiconghui:
> @caiconghui @yiguolei It does increase FE's responsibilities, but putting it in FE has its benefits:
> 
> 1. It can control disk balance globally and avoid competing with recovery and cluster balancing.
> 2. It reuses the Rebalance mechanism of FE, and  there is no need to add a new mechanism for FE. If it is implemented in BE, a set of balance mechanism needs to be repeatedly designed without additional income.

I personally think that the scalability of this implementation is not good, especially when there are a large number of machines. If there is data skew in the internal be for each machine, then the repair bottleneck will be thrown to fe.
now cluster balancing is to deal with data skew between bes.
Then this implementation and cluster balance can be run together？how do they collaborate with each other? when the tablet is scheduled to a be, then the be also needs to consider the disk strategy

fe/fe-core/src/main/java/org/apache/doris/clone/BackendLoadStatistic.java
```diff
@@ -315,6 +315,60 @@ public BalanceStatus isFit(long tabletSize, TStorageMedium medium,
...
+        double newSrcPathScore = srcPathStat.getCapacityB() == 0
+            ? 0.0 : (srcPathStat.getUsedCapacityB() - tabletSize) / (double) srcPathStat.getCapacityB();
+        double newDestPathScore = destPathStat.getCapacityB() == 0
+            ? 0.0 : (destPathStat.getUsedCapacityB() + tabletSize) / (double) destPathStat.getCapacityB();
+
+        double currentDiff = Math.abs(currentSrcPathScore - avgUsedPercent)
+            + Math.abs(currentDestPathScore - avgUsedPercent);
+        double newDiff = Math.abs(newSrcPathScore - avgUsedPercent) + Math.abs(newDestPathScore - avgUsedPercent);
+
+        LOG.info("after migrate {}(size: {}) from {} to {}, medium: {}, the load score changed."
```
yangzhg, line 362: should this be debug level?

yinzhijian:
> 
1. Personally, I think that disk balancing does not affect the reliability of data like recover, so its priority is not high. If it is completed too quickly, it will affect the performance of BE, thereby affecting service stability.
2. By default, disk balancing and cluster balancing will not run together. Only when the cluster has been balanced, the disk balancing task be scheduled. Thereby avoiding unnecessary data migration.

morningman:
> @caiconghui @yiguolei It does increase FE's responsibilities, but putting it in FE has its benefits:
> 
> 1. It can control disk balance globally and avoid competing with recovery and cluster balancing.
> 2. It reuses the Rebalance mechanism of FE, and  there is no need to add a new mechanism for FE. If it is implemented in BE, a set of balance mechanism needs to be repeatedly designed without additional income.

Using FE to control balance is a more controllable solution, although it may not be the most efficient or the fastest.
If put into the BE side of the balance processing separately, we cannot globally control the balancing operation within a cluster, which may pose a bigger problem. (For example, how do we control individual BEs to stop balancing operations to reduce system load?)

If a better solution is subsequently available that can be balanced separately between disks on the BE, it is possible to switch to the new solution and there is no conflict with the current implementation.

be/src/agent/task_worker_pool.cpp
```diff
@@ -941,8 +941,8 @@ void TaskWorkerPool::_storage_medium_migrate_worker_thread_callback() {
         TStatusCode::type status_code = TStatusCode::OK;
         // check request and get info
         TabletSharedPtr tablet;
-        DataDir* dest_store;
-        if (_check_migrate_requset(storage_medium_migrate_req, tablet, &dest_store) !=
+        DataDir* dest_store = NULL;
```
adonis0147, line 944: DataDir* dest_store = nullptr;

adonis0147 (changes requested):

morningman (approved):
LGTM

PR #8567, opened 2022-03-21
[api-change] add soft limit of String type length

yangzhg:
# Proposed changes

1. add a config string_type_soft_limit to soft limit max length of string type
2. disable using String type in Key column, partition column and
   distribution column
3. Remove string type alias BLOB for future use

## Problem Summary:

## Checklist(Required)

5. Does it need to update dependencies: (Yes/No)
6. Are there any changes that cannot be rolled back: (Yes/No)

## Further comments

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -720,6 +720,13 @@ CONF_String(function_service_protocol, "h2:grpc");
 // use which load balancer to select server to connect
 CONF_String(rpc_load_balancer, "rr");
 
+// a soft limit of string type length, the hard limit is 2GB - 4, but if too long will cause very low performance,
+// so we set a soft limit, default is 1MB
+CONF_mInt32(string_type_length_soft_limit_bytes, "1048576");
+
+CONF_Validator(string_type_length_soft_limit_bytes,
+               [](const int config) -> bool { return config > 0 && config <= 2147483643; });
+
 } // namespace config
 
 } // namespace doris
--- a/be/src/olap/row_block2.cpp
+++ b/be/src/olap/row_block2.cpp
@@ -17,6 +17,7 @@
 
 #include "olap/row_block2.h"
 
+#include <algorithm>
 #include <sstream>
 #include <utility>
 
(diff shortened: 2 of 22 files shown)


be/src/olap/row_block2.cpp
```diff
@@ -188,15 +190,16 @@ Status RowBlockV2::_copy_data_to_column(int cid, doris::vectorized::MutableColum
...
             if (!nullable_mark_array[j]) {
                 uint16_t row_idx = _selection_vector[j];
                 auto slice = reinterpret_cast<const Slice*>(column_block(cid).cell_ptr(row_idx));
-                if (LIKELY(slice->size <= MAX_SIZE_OF_VEC_STRING)) {
+                if (LIKELY(slice->size <= limit)) {
                     column_string->insert_data(slice->data, slice->size);
                 } else {
-                    return Status::NotSupported("Not support string len over than 1MB in vec engine.");
+                    return Status::NotSupported(
+                            "Not support string len over than 1MB in vec engine.");
```
morningman, line 202: Change `1MB` to `limit`

be/src/common/config.h
```diff
@@ -720,6 +720,10 @@ CONF_String(function_service_protocol, "h2:grpc");
 // use which load balancer to select server to connect
 CONF_String(rpc_load_balancer, "rr");
 
+// a soft limit of string type length, the hard limit is 2GB - 4, but if too long will cause very low performance,
+// so we set a soft limit, default is 10MB
+CONF_mInt32(string_type_soft_limit, "10485760");
```
morningman, line 725: better use 1MB?

morningman: `string_type_length_soft_limit_bytes`

morningman (commented):
also need to modify schema change

fe/fe-core/src/main/java/org/apache/doris/analysis/CreateTableStmt.java
```diff
@@ -416,7 +416,7 @@ public void analyze(Analyzer analyzer) throws UserException {
             if (distributionDesc == null) {
                 throw new AnalysisException("Create olap table should contain distribution desc");
             }
-            distributionDesc.analyze(columnSet);
+            distributionDesc.analyze(columnSet, columnDefs);
```
morningman, line 419: There is another place we need to modify in this `CreateTableStmt`.
When user create a DUPLICATE KEY table without specify the key column, Doris will auto
select part of columns as KEY column.
And if it select a string column as KEY column, the error will return to user.
But this is not what user expected.
We should:
1. stop selecting KEY column when met String column.
2. return error msg to user when the first column is String and give a suggestion to user.

morningman (approved):
LGTM

PR #8574, opened 2022-03-22
[refactor](schema_hash) remove schema_hash since every tablet id in be is unique

caiconghui:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

1.Now, schema_hash is part of the data_dir path, but tablet id is unique, so we can read tablet info only by tablet_id,
so this PR only modify get_tablet and drop_tablet function in be, not refactor write schema_hash logic in be.
module in be who use get_tablet and drop_tablet function will be affected.
2. For users, this PR change two httpAction in be, one is metaAction, and the other is CompactionAction. so I also modify the doc and fe code.
3.The main Purpose of this PR is to simplify managing tablet, whether to remove all schema_hash still need more discussion, because we need to consider compatibility and correctness after refactor.
 
## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/agent/task_worker_pool.cpp
+++ b/be/src/agent/task_worker_pool.cpp
@@ -976,11 +976,9 @@ void TaskWorkerPool::_storage_medium_migrate_worker_thread_callback() {
 OLAPStatus TaskWorkerPool::_check_migrate_request(const TStorageMediumMigrateReq& req,
                                                   TabletSharedPtr& tablet, DataDir** dest_store) {
     int64_t tablet_id = req.tablet_id;
-    int32_t schema_hash = req.schema_hash;
-    tablet = StorageEngine::instance()->tablet_manager()->get_tablet(tablet_id, schema_hash);
+    tablet = StorageEngine::instance()->tablet_manager()->get_tablet(tablet_id);
     if (tablet == nullptr) {
-        LOG(WARNING) << "can't find tablet. tablet_id= " << tablet_id
-                     << " schema_hash=" << schema_hash;
+        LOG(WARNING) << "can't find tablet. tablet_id= " << tablet_id;
         return OLAP_ERR_TABLE_NOT_FOUND;
     }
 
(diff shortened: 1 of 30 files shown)


yiguolei:
Great.
LGTM

yangzhg:
1. Does the command line of meta_tool also need to remove schema_hash?
2.  Do all the http_action of be has schema_hash should  remove schema_hash? If so, you should change the url of fe in `show tablet`
3. if the above is yes, document is also need to change

caiconghui:
> 1. Does the command line of meta_tool also need to remove schema_hash?
> 2. Do all the http_action of be has schema_hash should  remove schema_hash? If so, you should change the url of fe in `show tablet`
> 3. if the above is yes, document is also need to change

@yangzhg now, for upgrade compatibility , it is not the time to modify meta_tool, and url of show tablet command,
it just optimize the logic in be to find tablet and drop tablet now

yangzhg:
> > 1. Does the command line of meta_tool also need to remove schema_hash?
> > 2. Do all the http_action of be has schema_hash should  remove schema_hash? If so, you should change the url of fe in `show tablet`
> > 3. if the above is yes, document is also need to change
> 
> @yangzhg now, for upgrade compatibility , it is not the time to modify meta_tool, and url of show tablet command, it just optimize the logic in be to find tablet and drop tablet now

I think it's understandable to bring some compatibility problems when a big version is released, and there are not many users of meta_tool and url, so it's better to modify them together

morningman:
Schema hash is currently widely used in various modules of BE and is strongly related to the data itself.
Strict review is required for schema hash modifications.

It is recommended to add the following content to the PR:

1. Which modules currently use the schema hash, and whether it will affect it after removing it.
2. Which modules are modified by this PR
3. Which modules have not been modified
4. What is the next step.

yiguolei:
> Schema hash is currently widely used in various modules of BE and is strongly related to the data itself. Strict review is required for schema hash modifications.
> 
> It is recommended to add the following content to the PR:
> 
> 1. Which modules currently use the schema hash, and whether it will affect it after removing it.
> 2. Which modules are modified by this PR
> 3. Which modules have not been modified
> 4. What is the next step.

This PR mainly modify following modules:
1. It removes schema hash from tablet manager's get_tablet method, and modified the related usage.
2. It removes schema hash from http service's api. [this may cause some usage problems but I think it is ok]

yiguolei (approved):
LGTM 
I've pull the pr to local and build successfully.

be/src/agent/task_worker_pool.cpp
```diff
@@ -1552,10 +1549,9 @@ void TaskWorkerPool::_move_dir_thread_callback() {
 Status TaskWorkerPool::_move_dir(const TTabletId tablet_id, const TSchemaHash schema_hash,
```
morningman, line 1549: remove `schema_hash` in parameter

PR #8597, opened 2022-03-22
[feature-wip](array-type)Add element_at and subscript functions

cambyzju:
# Proposed changes

Issue Number: close #7570

## Problem Summary:

1. add function element_at;
2. support element_subscript([]) to get element of array, col_array[N] <==> element_at(col_array, N);
3. return error message instead of BE crash while array function execute failed;

element_at(array, index) desc:
>   Returns element of array at given **(1-based)** index. 
  If **index < 0**, accesses elements from the last to the first. 
  Returns NULL if the index exceeds the length of the array or the array is NULL.

Usage example:
1. create table with ARRAY type column and insert some data:
```
+------+------+--------+
| k1   | k2   | k3     |
+------+------+--------+
|    1 |    2 | [1, 2] |
|    2 |    3 | NULL   |
|    4 | NULL | []     |
|    3 | NULL | NULL   |
+------+------+--------+
```
2. enable vectorized:
```
set enable_vectorized_engine=true;
```
3. element_subscript([]) usage example:
```
> select k1,k3,k3[1] from array_test;
+------+--------+----------------------------+
| k1   | k3     | %element_extract%(`k3`, 1) |
+------+--------+----------------------------+
|    3 | NULL   |                       NULL |
|    1 | [1, 2] |                          1 |
|    2 | NULL   |                       NULL |
|    4 | []     |                       NULL |
+------+--------+----------------------------+
```
4. element_at function usage example:
```
> select k1,k3 from array_test where element_at(k3, -1) = 2;
+------+--------+
| k1   | k3     |
+------+--------+
|    1 | [1, 2] |
+------+--------+
```

## Checklist(Required)

6. Has unit tests been added: (Yes/No/No Need)
7. Has document been added or modified: (Yes/No/No Need)
8. Does it need to update dependencies: (Yes/No)
9. Are there any changes that cannot be rolled back: (Yes/No)

## Further comments


be/src/vec/functions/array/function_array_element.h
```diff
@@ -0,0 +1,247 @@
...
+    template <typename ElementDataType>
+    ColumnPtr _execute_number(const ColumnsWithTypeAndName& arguments,
+                              const DataTypePtr& result_type, size_t input_rows_count,
+                              const UInt8* src_null_map, UInt8* dst_null_map) {
+        // check array nested column type and get data
+        auto array_column = check_and_get_column<ColumnArray>(*arguments[0].column);
+        DCHECK(array_column != nullptr);
+        const ColumnVector<ElementDataType>* nested_column = nullptr;
+        const UInt8* nested_null_map = nullptr;
+        auto nested_null_column = check_and_get_column<ColumnNullable>(array_column->get_data());
```
Gabriel39, line 191: Both argument column and column inside array column can be nullable? It means it's possible for this function to pass argument like this `ColumnNullable(ColumnArray(ColumnNullable(ColumnInt8)))` ? I think it's better to de-duplicate these two ColumnNullable

be/src/vec/functions/array/function_array_element.h
```diff
@@ -0,0 +1,247 @@
...
+        block.replace_by_position(
+                result, ColumnNullable::create(std::move(res_column), std::move(dst_null_column)));
+        return Status::OK();
+    }
+
+private:
+    ColumnPtr _perform(const ColumnsWithTypeAndName& arguments, const DataTypePtr& result_type,
+                       size_t input_rows_count, const UInt8* src_null_map, UInt8* dst_null_map) {
+        ColumnPtr res;
+        if (!((res = _execute_number<Int8>(arguments, result_type, input_rows_count, src_null_map,
```
Gabriel39, line 91: `_execute_number ` and `_execute_string` start with the same logic so could we move the common logics here. It will help us improve efficience. Now, for Float64 type, it have to be passed to each `_execute_number` call and return nullptr until it is passed to `_execute_number<Float64>`

be/src/vec/functions/array/function_array_element.h
```diff
@@ -0,0 +1,247 @@
...
+                dst_null_map[row] = true;
+                dst_data[row] = ElementDataType();
+            } else {
+                DCHECK(index >= 0 && index < nested_data.size());
+                dst_null_map[row] = false;
+                dst_data[row] = nested_data[index];
+            }
+        }
+
+        VLOG_DEBUG << "function:" << get_name() << ", array type:" << arguments[0].type->get_name()
```
Gabriel39, line 240: Is this log necessary?

Gabriel39 (changes requested):

be/src/vec/functions/array/function_array_element.h
cambyzju: Yes we support ColumnNullable(ColumnArray(ColumnNullable(ColumnInt8)))

be/src/vec/functions/array/function_array_element.h
cambyzju: you are right, I have refractor these codes, please review again, thanks.

be/src/vec/functions/array/function_array_element.h
cambyzju: already removed

be/src/vec/functions/array/function_array_element.h
Gabriel39: @cambyzju I'm still confused about this. If outer ColumnNullable(null=true) is enough to indicate ColumnArray is null, it seems like the inner  ColumnNullable is redundant. So could you give an example which outer ColumnNullable is not null and inner ColumnNullable is null? Thx

cambyzju: ColumnNullable(  // the array may be null or not
ColumnArray(
ColumnNullable(  // nested element may be null or not
ColumnInt8)))

for example:
```
> select * from array_test;
+------+------+--------+
| k1   | k2   | k3     |
+------+------+--------+
|    3 | NULL | NULL   |
|    1 |    2 | [1, 2, NULL] |
|    4 | NULL | []     |
+------+------+--------+
```
row 1 : array is null (outer ColumnNullable)
row 2: one element of the array is null (inner ColumnNullable)

be/src/vec/functions/array/function_array_index.h
```diff
@@ -126,21 +139,37 @@ class FunctionArrayIndex : public IFunction
         return true;
     }
 
-#define INTEGRAL_TPL_PACK UInt8, UInt16, UInt32, UInt64, Int8, Int16, Int32, Int64, Float32, Float64
+#define INTEGRAL_TPL_PACK Int8, Int16, Int32, Int64, Float32, Float64
```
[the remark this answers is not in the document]
cambyzju: 1. already change from intergal to number;
2. also right column data type change from Resulting to RightType;

cambyzju:
> Check the logic of Clickhouse
> 
> 1. array_string call `array_position`  with `int` param
> 2. array_float call  `array_position`  with `int` param
>    Confirm the behavior rules of different array types and different data types

Got it. We will do it while extend array subtype to float and double.

morningman (approved):
LGTM

PR #8603, opened 2022-03-22
fix string default value bug

### LINKED ISSUE #8602: [Bug] Broker Load StringType cant set default_value
### Search before asking

### Version

master branch

### What's Wrong?

`CREATE TABLE test_db( 
  aaa date,
  bbb string , 
  ccc string, 
`

`
LOAD xxxxx 
set (
  aaa='_col1',
  bbb='_col2',
  ccc=default_value("good")
)
`

When setting the default value for a string type field, the actual measurement only supports a string with a maximum length of 1.

报错如下：

> type:ETL_RUN_FAIL; msg:errCode = 2, detailMessage = errCode = 2, detailMessage = Default value is too long: good

### What You Expected?

StringType can set default value.

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

whutpencil:
# Proposed changes

Issue Number: close #8602 

## Problem Summary:

Set the length of the string type field for analysis to the maximum allowed length

## Checklist(Required)

## Further comments


morningman:
I test it with following case, and there is no error:
```
CREATE TABLE test_db( aaa date, bbb string , ccc string) distributed by hash(aaa) buckets 1;
insert into test_db values("2021-01-01", "adasdsad", "asdsadsad");
```

Which version are you using?

whutpencil:
> 

@morningman  My branch is master. Please look at the description in my issue. It is Load instead of insert stmt.

fe/fe-core/src/main/java/org/apache/doris/analysis/ColumnDef.java
```diff
@@ -157,7 +159,7 @@ public void analyze(boolean isOlap) throws AnalysisException {
             final ScalarType targetType = (ScalarType) typeDef.getType();
```
cambyzju, line 159: default len of StringType is MAX_STRING_LENGTH, it seems your change is the same as just remove codes between line 158 to line 163.

fe/fe-core/src/main/java/org/apache/doris/analysis/ColumnDef.java
```diff
@@ -283,11 +285,15 @@ public static void validateDefaultValue(Type type, String defaultValue) throws A
             case CHAR:
             case VARCHAR:
             case HLL:
-            case STRING:
                 if (defaultValue.length() > scalarType.getLength()) {
                     throw new AnalysisException("Default value is too long: " + defaultValue);
                 }
                 break;
+            case STRING:
```
cambyzju, line 292: if StringType length is MAX_STRING_LENGTH, your change is the same as before.
if we set StringType length while create table, such as set to 100, the default value length should not bigger than 100, not MAX_STRING_LENGTH.

whutpencil: The length cannot be specified when creating a string type field. The default is the maximum length

fe/fe-core/src/main/java/org/apache/doris/analysis/ColumnDef.java
whutpencil: > default len of StringType is MAX_STRING_LENGTH, it seems your change is the same as just remove codes between line 158 to line 163.

This is because I don't know where the length of string is set to 1, so I directly specify the maximum value here.

whutpencil: This code sets the length of string type without length to 1, which should be consistent with the maximum length.

fe/fe-core/src/main/java/org/apache/doris/analysis/ColumnDef.java
whutpencil: The length cannot be specified when creating a string type field. The default is the maximum length. 

This is also to avoid modifying the allowable length of string type in other places. If you debug your own local code, you will find that the maximum allowable length of string has become 1.

Issue #8487, opened 2022-03-15
[Feature] CSV import and export support header


chenlinzhong:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Description

- Add two new types to stream load boker load: **csv_with_names** and **csv_with_name_sand_types**
- Add two new types to export: **csv_with_names** and **csv_with_names_and_types**

### Use case

Suppose you have a table like this
```
MySQL [test_db]> show create table my_table
....
CREATE TABLE `my_table` (
  `event_day` date NULL COMMENT "",
  `siteid` int(11) NULL DEFAULT "10" COMMENT "",
  `citycode` smallint(6) NULL COMMENT "",
  `username` varchar(32) NULL DEFAULT "" COMMENT "",
  `pv` bigint(20) SUM NULL DEFAULT "0" COMMENT ""
) ENGINE=OLAP
AGGREGATE KEY(`event_day`, `siteid`, `citycode`, `username`)
COMMENT "OLAP"
DISTRIBUTED BY HASH(`siteid`) BUCKETS 10
PROPERTIES (
"replication_allocation" = "tag.location.default: 1",
"in_memory" = "false",
"storage_format" = "V2"
)
```
### case1:  csv file without header(default)
curl --location-trusted -u root -H "label:123" -H "format:csv" -T testData.csv http://host:port/api/test_db/my_table/_stream_load
testData.csv
```
2017-07-03,78,5,OlpJfkVz,176
2017-07-03,73,18,iVM0NyAH,49
2017-07-03,5,13,wvim4aqW,95
```

### case2: csv file with header:names
curl --location-trusted -u root -H "label:123"  -H "format:csv_with_names" -T testData.csv http://host:port/api/test_db/my_table/_stream_load
testData.csv
```
event_day,siteid,citycode,username,pv
2017-07-03,78,5,OlpJfkVz,176
2017-07-03,73,18,iVM0NyAH,49
2017-07-03,5,13,wvim4aqW,95
```
### case3: csv file with header:names and types
curl --location-trusted -u root -H "label:123"  -H "format:csv_with_names_and_types" -T testData.csv  http://host:port/api/test_db/my_table/_stream_load
testData.csv
```
event_day,siteid,citycode,username,pv
date,int,samllint,varchar,int
2017-07-03,78,5,OlpJfkVz,176
2017-07-03,73,18,iVM0NyAH,49
[...]

PR #8604, opened 2022-03-22
[Feature] CSV import and export support header 

chenlinzhong:
# Proposed changes

Issue Number: open #8487

## Problem Summary:

- Add two new types to stream load boker load: csv_with_names and csv_with_name_sand_types
- Add two new types to export: csv_with_names and csv_with_names_and_types

## Checklist(Required)

## Further comments


be/src/exec/broker_scanner.cpp
```diff
@@ -103,6 +103,10 @@ Status BrokerScanner::get_next(Tuple* tuple, MemPool* tuple_pool, bool* eof, boo
         const uint8_t* ptr = nullptr;
         size_t size = 0;
         RETURN_IF_ERROR(_cur_line_reader->read_line(&ptr, &size, &_cur_line_reader_eof));
+        if (_skip_rows > 0){
+            _skip_rows--;
+            continue;
+        }
         if (_skip_next_line) {
```
cambyzju, line 110: could we replace _skip_next_line by '_skip_rows=1' ?

be/src/exec/broker_scanner.cpp
```diff
@@ -161,6 +193,11 @@ Status BrokerScanner::open_file_reader() {
     if (start_offset != 0) {
         start_offset -= 1;
     }
+    //means first range 
+    if (start_offset == 0){
```
cambyzju, line 197: if start_offset=1, then in line 194, it will be set to start_offset=0.

it is still first range?

be/src/http/action/stream_load.cpp
```diff
@@ -98,6 +98,40 @@ static TFileFormatType::type parse_format(const std::string& format_str,
         if (compress_type.empty()) {
             format_type = TFileFormatType::FORMAT_JSON;
         }
+    } else if (boost::iequals(format_str, "CSV_WITH_NAMES")) {
```
cambyzju, line 101: looks these codes the same as boost::iequals(format_str, "CSV")？

be/src/http/action/stream_load.cpp
```diff
@@ -98,6 +98,40 @@ static TFileFormatType::type parse_format(const std::string& format_str,
...
+        if (compress_type.empty()) {
+            format_type = TFileFormatType::FORMAT_CSVWITHNAMES_PLAIN;
+        }
```
cambyzju, line 104: add else?

be/src/http/action/stream_load.cpp
```diff
@@ -558,6 +606,8 @@ Status StreamLoadAction::_process_put(HttpRequest* http_req, StreamLoadContext*
                      << ctx->brief();
         return plan_status;
     }
+    LOG(INFO) << "ThriftDebugString:" << apache::thrift::ThriftDebugString(ctx->put_result);
```
cambyzju, line 609: debug message use VLOG, and remove std::cout

be/src/exec/broker_scanner.cpp
chenlinzhong: Yes, the reason for subtracting 1 is that offset is plus 1 in the FE calculation (bytesperinstance = totalbytes / numinstances + 1), so for the first range  start_ Offset is always 0, and other range will not be 0

be/src/exec/broker_scanner.cpp
chenlinzhong: ok, the latest code has been replaced

be/src/http/action/stream_load.cpp
chenlinzhong: It has been optimized

be/src/http/action/stream_load.cpp
chenlinzhong: It has been optimized

be/src/http/action/stream_load.cpp
chenlinzhong: fixed

gensrc/thrift/PlanNodes.thrift
```diff
@@ -108,6 +108,22 @@ enum TFileFormatType {
     FORMAT_ORC,
     FORMAT_JSON,
     FORMAT_PROTO,
+    //csv withnames
+    FORMAT_CSVWITHNAMES_PLAIN,
```
yangzhg, line 112: may put header and compress type to another message is better

chenlinzhong: Yes, the more reasonable way is to enumerate the file type and compression type separately. However, since the previous logic is that the two are put together, the expansibility is not very good. If you want to separate them, there are many changes involved and the cycle will be relatively long. Therefore, the previous way is still temporarily continued here

Issue #7678, opened 2022-01-08
[Feature] max_by, min_by aggregate function support


hf200012:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Description

max_by(x, y) Query the corresponding x value when y is the maximum value.
max_by(x, y, n) Query the x values corresponding to the largest n y values, and return the result as an array.
min_by(x, y) Query the corresponding x value when y is the minimum value.
min_by(x, y, n) Query the x value corresponding to the smallest n y values. The result is returned as an array.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #8623, opened 2022-03-24
[Vectorized][Feature] Support min_by/max_by function.

zhannngchen:
# Proposed changes

Issue Number: open #7678

## Problem Summary:

Support min_by/max_by on vectorized engine.

## Checklist(Required)

## Further comments
1. This is an initial patch, I will add the document later.
2. For now, aggregate functions only supports 1 intermediate type, but min_by/max_by needs 2 type information, I've did a tricky work-around in this initial patch, but we need to discuss that do we really need to change the intermediate type to an intermediate type list? @morningman @HappenLee  @dataroaring. Another solution is to serialize the type information in the intermediate data.
3. To support these 2 functions in the non-vectorized code needs to pay much more efforts, we may need to do some refactor. Do we need to support this?

### DIFF
--- a/be/src/vec/aggregate_functions/aggregate_function_min_max_by.h
+++ b/be/src/vec/aggregate_functions/aggregate_function_min_max_by.h
@@ -1,0 +1,114 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#pragma once
+
+#include "common/logging.h"
+#include "vec/aggregate_functions/aggregate_function.h"
+#include "vec/columns/column_decimal.h"
+#include "vec/columns/column_vector.h"
+#include "vec/common/assert_cast.h"
+#include "vec/io/io_helper.h"
+
+namespace doris::vectorized {
+template <typename VT, typename KT>
+struct AggregateFunctionMaxByData {
+private:
+    VT value;
+    KT key;
+public:
+    using Self = AggregateFunctionMaxByData;
+
+    bool change_if_better(const IColumn& value_column, const IColumn& key_column, size_t row_num, Arena* arena) {
+        if (key.change_if_greater(key_column, row_num, arena)) {
+            value.change(value_column, row_num, arena);
+        }
+    }
+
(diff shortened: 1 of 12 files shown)


fe/fe-core/src/main/java/org/apache/doris/analysis/AggregateInfo.java
```diff
@@ -452,10 +452,14 @@ private void createMergeAggInfo(Analyzer analyzer)  {
         for (int i = 0; i < getAggregateExprs().size(); ++i) {
             FunctionCallExpr inputExpr = getAggregateExprs().get(i);
             Preconditions.checkState(inputExpr.isAggregateFunction());
-            Expr aggExprParam =
-                    new SlotRef(inputDesc.getSlots().get(i + getGroupingExprs().size()));
+            List<Expr> paramExprs = new ArrayList<>();
+            if (inputExpr.fn.functionName().equals("max_by")) {
```
zhannngchen, line 456: It's a tricky workaround, won't merge to production code.

zbtzbtzbt:
we can add docs for these new functions @zhannngchen

zhannngchen:
> we can add docs for these new functions @zhannngchen

Yes, I'll update the patch later today

be/test/vec/aggregate_functions/agg_min_max_by_test.cpp
```diff
@@ -0,0 +1,107 @@
...
+        }
+        agg_function->destroy(place);
+    }
+}
+
+INSTANTIATE_TEST_SUITE_P(Params, AggMinMaxByTest,
+                         ::testing::ValuesIn(std::vector<std::string> {"min_by", "max_by"}));
+} // namespace doris::vectorized
+
+int main(int argc, char** argv) {
```
yiguolei, line 104: We changed unit test running mechanism, not every unit test has a main function.

yiguolei (dismissed):
LGTM

be/test/vec/aggregate_functions/agg_min_max_by_test.cpp
zhannngchen: cleared

be/src/vec/aggregate_functions/aggregate_function_min_max_by.h
```diff
@@ -0,0 +1,166 @@
...
+    AggregateFunctionsMinMaxBy(const DataTypePtr& value_type_, const DataTypePtr& key_type_)
+            : IAggregateFunctionDataHelper<
+                      Data, AggregateFunctionsMinMaxBy<Data, AllocatesMemoryInArena>>(
+                      {value_type_, key_type_}, {}),
+              value_type(this->argument_types[0]),
+              key_type(this->argument_types[1]) {
+        if (StringRef(Data::name()) == StringRef("min_by") ||
+            StringRef(Data::name()) == StringRef("max_by")) {
+            if (!key_type_->is_comparable()) {
+                LOG(FATAL) << fmt::format(
```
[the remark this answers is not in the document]
zhannngchen: updated

fe/fe-core/src/main/java/org/apache/doris/analysis/AggregateInfo.java
zhannngchen: Discussed with @HappenLee and @yiguolei , we should change the intermediateType to a list, in another PR, in current PR, we just keep the hard code way.

morningman (approved):
LGTM

PR #8627, opened 2022-03-24
[refactor][optimize] Code optimization and refactoring for low-cardinality columns in storage layer

### LINKED ISSUE #8315: [Enhancement] Low cardinality string optimization in storage layer

zenoyang:
# Proposed changes

Issue Number: close #8315 
Code optimization and refactoring for low-cardinality columns in storage layer.

## Problem Summary:

Optimization:
- Add `dict_code` member variable to the predicate. Before the predicate is calculated, the conversion of the predicate constant into the dictionary code only needs to be done once

Refactor:
- Add `PredicateType` enum, add `type` method to the predicate, which is used to represent the predicate type, instead of the original less flexible way, for example: `is_in_predicate()` ...
- Added the following interface to `IColumn` to simplify the code of the storage layer and avoid exposing the specific type of column:
     - `convert_to_predicate_column_if_dictionary`
     - `convert_dict_codes_if_necessary`
     - `set_predicate_dict_code_if_necessary`
- Class member variables start with an underscore
- Specify the `int` type only supported by the specialized `ColumnDictionary` template

In addition to this, there are some priorities that I will provide in the coming period.

### DIFF
--- a/be/src/olap/column_predicate.h
+++ b/be/src/olap/column_predicate.h
@@ -33,12 +33,29 @@ class VectorizedRowBatch;
 class Schema;
 class RowBlockV2;
 
+enum class PredicateType {
+    Unknown = 0,
+    EQ = 1,
+    NE = 2,
+    LT = 3,
+    LE = 4,
+    GT = 5,
+    GE = 6,
+    InList = 7,
+    NotInList = 8,
+    IsNull = 9,
+    NotIsNull = 10,
+    BF = 11
+};
+
 class ColumnPredicate {
 public:
     explicit ColumnPredicate(uint32_t column_id, bool opposite = false)
             : _column_id(column_id), _opposite(opposite) {}
 
     virtual ~ColumnPredicate() = default;
+
+    virtual PredicateType type() const = 0;
 
     //evaluate predicate on VectorizedRowBatch
     virtual void evaluate(VectorizedRowBatch* batch) const = 0;
@@ -68,18 +85,6 @@ public:
     // a vectorized eval way
     virtual void evaluate_vec(vectorized::IColumn& column, uint16_t size, bool* flags) const {};
     uint32_t column_id() const { return _column_id; }
-
-    virtual bool is_in_predicate() { return false; }
-
-    virtual bool is_bloom_filter_predicate() { return false; }
-
-    virtual bool is_equal_comparison_predicate() { return false; }
-
-    virtual bool is_range_comparison_predicate() { return false; }
-
-    // Useful only with range comparison predicates
-    virtual bool contain_equal() { return false; }
-    virtual bool is_less() { return false; }
 
 protected:
     uint32_t _column_id;
--- a/be/src/olap/column_predicate.h
+++ b/be/src/olap/column_predicate.h
@@ -73,7 +73,13 @@ public:
 
     virtual bool is_bloom_filter_predicate() { return false; }
 
+    virtual bool is_equal_comparison_predicate() { return false; }
+
     virtual bool is_range_comparison_predicate() { return false; }
+
+    // Useful only with range comparison predicates
+    virtual bool contain_equal() { return false; }
+    virtual bool is_less() { return false; }
 
 protected:
     uint32_t _column_id;
--- a/be/src/olap/column_predicate.h
+++ b/be/src/olap/column_predicate.h
@@ -86,6 +86,8 @@ public:
     virtual void evaluate_vec(vectorized::IColumn& column, uint16_t size, bool* flags) const {};
     uint32_t column_id() const { return _column_id; }
 
+    virtual void set_dict_code_if_necessary(vectorized::IColumn& column) { }
+
 protected:
     uint32_t _column_id;
     bool _opposite;
--- a/be/src/olap/column_predicate.h
+++ b/be/src/olap/column_predicate.h
@@ -34,17 +34,17 @@ class Schema;
 class RowBlockV2;
 
 enum class PredicateType {
-    Unknown = 0,
+    UNKNOWN = 0,
     EQ = 1,
     NE = 2,
     LT = 3,
     LE = 4,
     GT = 5,
     GE = 6,
-    InList = 7,
-    NotInList = 8,
-    IsNull = 9,
-    NotIsNull = 10,
+    IN_LIST = 7,
+    NO_IN_LIST = 8,
+    IS_NULL = 9,
+    NOT_IS_NULL = 10,
     BF = 11, // BloomFilter
 };
 
--- a/be/src/olap/column_predicate.h
+++ b/be/src/olap/column_predicate.h
@@ -45,7 +45,7 @@ enum class PredicateType {
     NotInList = 8,
     IsNull = 9,
     NotIsNull = 10,
-    BF = 11
+    BF = 11, // BloomFilter
 };
 
 class ColumnPredicate {
--- a/be/src/olap/comparison_predicate.cpp
+++ b/be/src/olap/comparison_predicate.cpp
@@ -28,9 +28,9 @@
 
 namespace doris {
 
-#define COMPARISON_PRED_CONSTRUCTOR(CLASS)                                   \
-    template <class type>                                                    \
-    CLASS<type>::CLASS(uint32_t column_id, const type& value, bool opposite) \
+#define COMPARISON_PRED_CONSTRUCTOR(CLASS) \
+    template <class T>                     \
+    CLASS<T>::CLASS(uint32_t column_id, const T& value, bool opposite) \
             : ColumnPredicate(column_id, opposite), _value(value) {}
 
 COMPARISON_PRED_CONSTRUCTOR(EqualPredicate)
@@ -56,15 +56,15 @@ COMPARISON_PRED_CONSTRUCTOR_STRING(GreaterPredicate)
 COMPARISON_PRED_CONSTRUCTOR_STRING(GreaterEqualPredicate)
 
 #define COMPARISON_PRED_EVALUATE(CLASS, OP)                                           \
-    template <class type>                                                             \
-    void CLASS<type>::evaluate(VectorizedRowBatch* batch) const {                     \
+    template <class T>                                                                \
+    void CLASS<T>::evaluate(VectorizedRowBatch* batch) const {                        \
         uint16_t n = batch->size();                                                   \
         if (n == 0) {                                                                 \
             return;                                                                   \
         }                                                                             \
         uint16_t* sel = batch->selected();                                            \
-        const type* col_vector =                                                      \
-                reinterpret_cast<const type*>(batch->column(_column_id)->col_data()); \
+        const T* col_vector =                                                         \
+                reinterpret_cast<const T*>(batch->column(_column_id)->col_data());    \
         uint16_t new_size = 0;                                                        \
         if (batch->column(_column_id)->no_nulls()) {                                  \
             if (batch->selected_in_use()) {                                           \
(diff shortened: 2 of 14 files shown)


be/src/vec/columns/column_dictionary.h
```diff
@@ -242,18 +213,73 @@ class ColumnDictionary final : public COWHelper<IColumn, ColumnDictionary<T>> {
...
+            auto* in_pred = reinterpret_cast<doris::InListPredicate<StringValue>*>(predicate);
+            auto pred_values = in_pred->get_values();
+            auto code_set = _dict.find_codes(pred_values);
+            in_pred->set_dict_codes(code_set);
+            break;
+        }
+        default:
+            LOG(FATAL) << "PredicateType: " << static_cast<int>(predicate->type())
+                       <<  " not supported in ColumnDictionary";
+            break;
```
wangbo, line 281: ```LOG(FATAL)``` could core directly， so break is useless.

wangbo:
Is there a  performance drop after code refactoring?

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -830,23 +831,17 @@ void SegmentIterator::_evaluate_short_circuit_predicate(uint16_t* vec_sel_rowid_
...
+            col_ptr->convert_dict_codes_if_necessary();
         }
-
-        if (col_ptr->is_column_dictionary() && column_predicate->is_range_comparison_predicate()) {
-            auto& dict_col =
-                    reinterpret_cast<vectorized::ColumnDictionary<vectorized::Int32>&>(*col_ptr);
-            dict_col.convert_dict_codes();
-        }
-        column_predicate->evaluate(*short_cir_column, vec_sel_rowid_idx, selected_size_ptr);
+        col_ptr->set_predicate_dict_code_if_necessary(predicate);
```
wangbo, line 843: First, It seems that you do ```set_predicate_dict_code_if_necessary``` after read one block.
Maybe we can do ```set_predicate_dict_code_if_necessary``` in SegmentIter's init method.
Because I think dict is immutable after segment is loaded, and the same to Predicate.

be/src/vec/columns/column_dictionary.h
```diff
@@ -242,18 +213,73 @@ class ColumnDictionary final : public COWHelper<IColumn, ColumnDictionary<T>> {
...
+            for (size_t i = 0; i < size(); ++i) {
+                _codes[i] = _dict.convert_code(_codes[i]);
+            }
+            _dict_code_converted = true;
+        }
+    }
+
+    void set_predicate_dict_code_if_necessary(doris::ColumnPredicate* predicate) override {
+        switch (predicate->type()) {
+        case PredicateType::EQ:
```
wangbo, line 246: Can we move this logic to ColumnPredicate?
I think maybe we can using ColumnPredicate's different implementation to different case instead of a runtime switch.

be/src/vec/columns/column_dictionary.h
zenoyang: ok

be/src/olap/rowset/segment_v2/segment_iterator.cpp
zenoyang: It's a problem, but it can't be implemented in `SegmentIterator::init`, because when the init method, there is no dictionary.
But I can add a judgment, only set once in the block load stage.

be/src/vec/columns/column_dictionary.h
zenoyang: ok, it's more reasonable

zenoyang: done

be/src/olap/null_predicate.cpp
```diff
@@ -29,6 +29,13 @@ namespace doris {
 NullPredicate::NullPredicate(uint32_t column_id, bool is_null, bool opposite)
         : ColumnPredicate(column_id), _is_null(opposite != is_null) {}
 
+PredicateType NullPredicate::type() const {
+    if (_is_null)
```
wangbo, line 33:
```suggestion
    return _is_null ? PredicateType::IsNull : PredicateType::NotIsNull;
```

be/src/vec/columns/column_dictionary.h
```diff
@@ -264,127 +261,121 @@ class ColumnDictionary final : public COWHelper<IColumn, ColumnDictionary<T>> {
...
+        inline int32_t find_code(const StringValue& value) const {
+            auto it = _inverted_index.find(value);
+            if (it != _inverted_index.end()) {
                 return it->second;
             }
             return -1;
         }
 
-        inline T find_bound_code(const StringValue& value, bool lower, bool eq) const {
+        inline int32_t find_code_by_bound(const StringValue& value, bool lower, bool eq) const {
```
wangbo, line 310: why hard code ```int32_t``` here?

zenoyang: Because the `dict_code` in `ColumnPredicate` is fixed to `int32_t` type, the `dict_code` returned by `find_code...` needs to be compared with the `dict_code` in `ColumnPredicate`. `T` may be `int8_t` or `int16_t`, so cast is performed when find returns.
And if there is no cast here, there will be a compilation error, similar to `phmap::flat_hash_set<int8_t>` cannot be converted to `phmap::flat_hash_set<int32_t>`

be/src/vec/columns/column_dictionary.h
```diff
@@ -264,127 +261,121 @@ class ColumnDictionary final : public COWHelper<IColumn, ColumnDictionary<T>> {
...
-            sort_dict();
-        }
-
-        if (!is_dict_code_converted()) {
-            for (size_t i = 0; i < size(); ++i) {
-                codes[i] = dict.convert_code(codes[i]);
-            }
-            _dict_code_converted = true;
+    ColumnPtr convert_to_predicate_column_if_dictionary() override {
+        auto res = vectorized::PredicateColumnType<StringValue>::create();
```
wangbo, line 276: What's the difference between ```convert_to_predicate_column_if_dictionary``` and ```convert_to_predicate_column ```

zenoyang: No difference, `convert_to_predicate_column` is useless, I'll delete it right away

be/src/olap/column_predicate.h
```diff
@@ -33,13 +33,30 @@ class VectorizedRowBatch;
...
 
+enum class PredicateType {
+    Unknown = 0,
+    EQ = 1,
+    NE = 2,
+    LT = 3,
+    LE = 4,
+    GT = 5,
+    GE = 6,
+    InList = 7,
```
wangbo, line 44:
```suggestion
    IN_LIST = 7,
```
You can refer other ```enum class in BE code.

zenoyang: done

be/src/olap/null_predicate.cpp
zenoyang: done

be/src/vec/columns/column_dictionary.h
zenoyang: > No difference, `convert_to_predicate_column` is useless, I'll delete it right away

done

be/src/olap/comparison_predicate.cpp
```diff
@@ -159,60 +159,54 @@ COMPARISON_PRED_COLUMN_BLOCK_EVALUATE(GreaterEqualPredicate, >=)
                                         .get_data();                                               \
             auto& nested_col = nullable_col->get_nested_column();                                  \
             if (nested_col.is_column_dictionary()) {                                               \
-                if constexpr (std::is_same_v<type, StringValue>) {                                 \
+                if constexpr (std::is_same_v<T, StringValue>) {                                    \
```
wangbo, line 162: Is this ```std::is_same_v<T, StringValue>``` necessary?
If a column is ColumnDict, then its TYPE in predicate can only be StringValue.

zenoyang: Yes, if I remove this judgment, the compilation will report an error

wangbo (dismissed):
LGTM

wangbo (approved):
LGTM

PR #8633, opened 2022-03-24
[Opt] Optimize hotspot functions in pre-compaction

zbtzbtzbt:
## Proposed changes
**Compaction** is very cpu consuming, 
after test, I found partial bottleneck is in function `rowset_with_max_version` 
so made following targeted performance tuning:

- better time cost O(1) function`max_version()` rather than previous heavy and hotspot O(n)
- use faster `phmap::flat_hash_map` replace `rs_version_map` and `_stale_rs_version_map` data structure
- use key word `inline` for frequently used small function
- remove some unnecessary inline keywords 

reference:
https://llvm.org/docs/CodingStandards.html#don-t-use-inline-when-defining-a-function-in-a-class-definition 
https://isocpp.org/wiki/faq/inline-functions#where-to-put-inline-keyword

### Previous: 
`Tablet::can_do_compaction` 40%, like Health Kit in China, red means unhealth.

previous

### Enhancement:
`Tablet::can_do_compaction` 20%, like Health Kit in China, green means health.

max_version_V2

### DIFF
--- a/be/src/olap/tablet_meta.h
+++ b/be/src/olap/tablet_meta.h
@@ -123,6 +123,7 @@ public:
     // disk space occupied by tablet
     size_t tablet_footprint() const;
     size_t version_count() const;
+    void update_max_version();
     Version max_version() const;
 
     TabletState tablet_state() const;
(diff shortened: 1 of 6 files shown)


zbtzbtzbt:
I also considered using SIMD max, 
but O(1) is fastest, and size of vector is not 128X or 256X.
so this way is better.

be/src/olap/tablet_meta.h
```diff
@@ -200,7 +201,8 @@ class TabletMeta {
     // the reference of _schema may use in tablet, so here need keep
     // the lifetime of tablemeta and _schema is same with tablet
     std::shared_ptr<TabletSchema> _schema;
-
+    // _maxVersion record _rs_metas's max endVersion Version
+    Version _maxVersion = {-1, 0};
```
HappenLee, line 205: _max_version?

zbtzbtzbt: ok

Issue #8629, opened 2022-03-24
[Bug] [sql_block_rule] the limitations of sql_block_rule can be negative


DarvenDuan:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Version

master branch

### What's Wrong?

When I create or alter a sql_block_rule like : `create sql_block_rule test_rule properties("partition_num"="-100","tablet_num"="-100");`,it can be executed correctly. And I exec `show sql_block_rule for test_rule` got the result like this: 
+------------+------+---------+--------------+-----------+-------------+--------+--------+
| Name       | Sql  | SqlHash | PartitionNum | TabletNum | Cardinality | Global | Enable |
+------------+------+---------+--------------+-----------+-------------+--------+--------+
| test_rule2 | NULL | NULL    | -100            | -100         | 0           | false  | true   |
+------------+------+---------+--------------+-----------+-------------+--------+--------+

Obviously, this is unreasonable

### What You Expected?

Value of properties can't be negative

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Issue #8621, opened 2022-03-24
[Enhancement] [sql_block_rule]  Optimize the judgment conditions when a query hits a sql_block_rule


DarvenDuan:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Description

If we create a sql_block_rule like this stmt:
`create sql_block_rule test_rule properties("cardinality" = "100","global" = "true");`

When we exec a sql and the 'cardinality' of which is greater than 100, we got an AnalysisException as follow:
`ERROR 1064 (HY000): errCode = 2, detailMessage = sql hits sql block rule: test_rule, reach partition_num : 0`

But our expected err massage may be :
`ERROR 1064 (HY000): errCode = 2, detailMessage = sql hits sql block rule: test_rule, reach cardinality : 100`

### Solution

 Optimize the judgment conditions if a query hits a sql_block_rule

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #8634, opened 2022-03-24
[bug] (sql_block_rule) Fix sql block rule bug

DarvenDuan:
# Proposed changes

Issue Number: #8629 & #8621

## Problem Summary:

1. Check properties'  effectiveness of a sql_block_rule, can't set limitations of sql_block_rule to be negative.
2. Optimize the judgment conditions when a query hits a sql_block_rule
3. Check if sql_block_rule has already exist when exec `set property for "user" "sql_block_rule" = "xxx"`

## Checklist(Required)

4. Has unit tests been added: (No Need)
5. Has document been added or modified: (No)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/UserProperty.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/UserProperty.java
@@ -275,9 +275,9 @@ public class UserProperty implements Writable {
                     throw new DdlException(PROP_SQL_BLOCK_RULES + " format error");
                 }
                 // check if sql_block_rule has already exist
-                for (String rule_name : value.replaceAll(" ","").split(",")){
-                    if (!rule_name.equals("") && !Catalog.getCurrentCatalog().getSqlBlockRuleMgr().existRule(rule_name)){
-                        throw new DdlException("the sql block rule " + rule_name + " not exist");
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/UserProperty.java
```diff
@@ -274,6 +274,12 @@ public void update(List<Pair<String, String>> properties) throws UserException {
                 if (keyArr.length != 1) {
                     throw new DdlException(PROP_SQL_BLOCK_RULES + " format error");
                 }
+                // check if sql_block_rule has already exist
+                for (String rule_name : value.replaceAll(" ","").split(",")){
```
morningman, line 278:
```suggestion
                for (String ruleName : value.replaceAll(" ","").split(",")){
```

morningman (commented):
Please add some unit test in `SqlBlockRuleMgrTest.java`

fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/UserProperty.java
DarvenDuan: Ok, thanks for your suggestion, I will do it latter.

PR #8500, opened 2022-03-16
[Opt] speed up data load for duplicate table no-vec-load

zbtzbtzbt:
# Proposed changes

issue number：https://github.com/apache/incubator-doris/issues/8522

## Problem Summary:

similar pr：https://github.com/apache/incubator-doris/pull/7967

via `perf` command, I find data loading mainly cost is from `insert()` funtion's `get_type_info()`,
so I made some targeted performance tuning.

@HappenLee @yangzhg @spaces-X @zenoyang 

thanks.

PR #8680, opened 2022-03-27
[Refactor](type_info) use template and single instance to refactor get type info logic

caiconghui:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments


be/src/olap/column_vector.h
```diff
@@ -150,7 +150,7 @@ class ScalarColumnVectorBatch : public ColumnVectorBatch {
 class ArrayNullColumnVectorBatch : public ColumnVectorBatch {
 public:
     explicit ArrayNullColumnVectorBatch(ColumnVectorBatch* array)
-            : ColumnVectorBatch(get_scalar_type_info(FieldType::OLAP_FIELD_TYPE_TINYINT), false),
+            : ColumnVectorBatch(get_scalar_type_info<FieldType::OLAP_FIELD_TYPE_TINYINT>(), false),
```
zbtzbtzbt, line 153: we can also delete prefix`FieldType::` here, and other places.

be/src/olap/column_vector.h
```diff
@@ -177,7 +177,7 @@ class ArrayNullColumnVectorBatch : public ColumnVectorBatch {
 
 class ArrayColumnVectorBatch : public ColumnVectorBatch {
 public:
-    explicit ArrayColumnVectorBatch(std::shared_ptr<const TypeInfo> type_info, bool is_nullable,
```
yiguolei, line 180: Could use reference instead of pointer.

be/src/olap/rowset/segment_v2/indexed_column_reader.cpp
```diff
@@ -31,12 +31,12 @@ Status IndexedColumnReader::load(bool use_page_cache, bool kept_in_memory) {
     _use_page_cache = use_page_cache;
     _kept_in_memory = kept_in_memory;
 
-    _type_info = get_type_info((FieldType)_meta.data_type());
+    _type_info = get_scalar_type_info((FieldType)_meta.data_type());
```
zbtzbtzbt, line 34: question: can we make sure that the type here is scalar type?

be/src/olap/tablet_schema.cpp
```diff
@@ -286,7 +286,7 @@ TabletColumn::TabletColumn(FieldAggregationMethod agg, FieldType type) {
 TabletColumn::TabletColumn(FieldAggregationMethod agg, FieldType filed_type, bool is_nullable) {
     _aggregation = agg;
     _type = filed_type;
-    _length = get_type_info(filed_type)->size();
+    _length = get_scalar_type_info(filed_type)->size();
```
zbtzbtzbt, line 289: question: can we make sure that the type here is scalar type?

be/src/olap/rowset/segment_v2/bitmap_index_reader.h
```diff
@@ -50,13 +50,13 @@ class BitmapIndexReader {
...
 
-    std::shared_ptr<const TypeInfo> type_info() { return _typeinfo; }
+    const TypeInfo* type_info() { return _typeinfo; }
 
 private:
     friend class BitmapIndexIterator;
 
     FilePathDesc _path_desc;
-    std::shared_ptr<const TypeInfo> _typeinfo;
+    const TypeInfo* _typeinfo;
```
yiguolei, line 59: Use reference?

caiconghui: pointer is enough

zbtzbtzbt:
for some certainly types, I think all use unified `_typeinfo` do not reflect semantics,
such as `bf_typeinfo` is better than `_typeinfo`. we may can modify some certainly types variables names.

be/src/olap/rowset/segment_v2/indexed_column_reader.cpp
caiconghui: because array, map and struct cannot be got only with FiedType info

be/src/olap/tablet_schema.cpp
caiconghui: because array, map and struct cannot be got only with FiedType info

be/src/olap/column_vector.h
caiconghui: ok

caiconghui:
> for some certainly types, I think all use unified `_typeinfo` do not reflect semantics, such as `bf_typeinfo` is better than `_typeinfo`. we may can modify some certainly types variables names.

actually _type_info is a private variable of parent object , so I think it is clear now, we can easy get its usage by code

adonis0147:
I think this pr should be discussed sufficiently before merge it in.

caiconghui:
> I think this pr should be discussed sufficiently before we merge it in.
> 
> Unlike the scalar type, the type info of dynamic types (such as Array, Map and etc) can not be generated in compile time. For instance, in `get_type_info(const TabletColumn* col)`, the depth of the col should be calculated first and it is unknown until runtime. The optimization for `ArrayType` here is powered by small object optimization technique (cache objects with small depth) but it sacrifice the flexibility for objects with bigger depth.
> 
> I think there are some possible improvements (may hard to implement):
> 
> 1. It is reasonable to restrict the max depth to protect our system from suffering serious performance penalty. I was also planning to do this. However, if we can configure the max depth in runtime, it may be better than the hard coded number.
> 2. `ArrayTypeInfo` can be refactored to store both leaf type info and depth, therefore, we can construct an `ArrayTypeInfo` with any depth in a quick way. After that, we may cache the objects with small depth statically and use some cache strategies for objects with large depth to speed up.
> 
> All in all, the key problem is how to manage the resource of dynamic types if there are generated dynamically. I think we should also consider whether the refactor introduced by this pr is suitable for the type info of `MayType` (which may be more hard to manage than `ArrayType`).
> 
> For now, `ArrayType` feature is under development and we have not start the performance stuffs for `ArrayType`. Thanks for the contribution. What's more, how much the benefit after this modification for `ArrayType`? Is there any statistics?

In fact, my PR is optimized for scalar type, because get scalar type is used very frequently, so the operation of getting scalar type must be light enough. Here I use templates and static variables to optimize get scalar type, and I I think that for the design of composite types such as arrays and maps, my changes are only expedient. I think composite types may need to be managed with another set of logic in future, because scalar type is the most used type by users, so performance is very important, PR #8500 has already done some performance 
 optimization work for this operation

adonis0147:
Hi @caiconghui , I read the #8500 before and knew the motivation behind this pr, but the challenge is that some functions (such as `get_type_info(const TabletColumn* col)`) and some members in classes (such as `type_info` in  `ColumnVectorBatch`) are used for both scalar types and dynamic types. If we manage the resource created by dynamic types and ignore the static resource for static types, the unified `const TypeInfo*` returned by some functions is appropriate. I think some functions and some usage for `TypeInfo` should be **SEPARATED** first before apply the  optimization for static types.

be/test/runtime/array_test.cpp
```diff
@@ -51,6 +51,7 @@ namespace doris {
 template <typename... Ts>
 ColumnPB create_column_pb(const std::string& type, const Ts&... sub_column_types) {
     ColumnPB column;
+    std::cout << " type cch1333 " << type << std::endl;
```
zbtzbtzbt, line 54: Remove unused code

be/test/olap/column_vector_test.cpp
```diff
@@ -145,7 +145,7 @@ TEST_F(ColumnVectorTest, scalar_column_vector_test) {
         for (int i = 0; i < size; ++i) {
             val[i] = i;
         }
-        auto ti = get_scalar_type_info(OLAP_FIELD_TYPE_TINYINT);
+        const auto* ti = get_scalar_type_info<OLAP_FIELD_TYPE_TINYINT>();
```
zbtzbtzbt, line 148: unified name type_info

be/src/olap/types.cpp
```diff
@@ -95,110 +47,188 @@ bool is_olap_string_type(FieldType field_type) {
...
+    case OLAP_FIELD_TYPE_ARRAY: {
+        int32_t iterations = 0;
+        const auto* child_column = &col->get_sub_column(0);
+        while (child_column->type() == OLAP_FIELD_TYPE_ARRAY) {
+            iterations++;
+            child_column = &child_column->get_sub_column(0);
         }
+        return get_array_type_info(child_column->type(), iterations);
+    }
+    [[likely]] default:
```
zbtzbtzbt, line 230: `[[likely]]` is used for C++20
we can use `UNLIKELY` or  `LIKELY` for C++17 in doris currently

reference: https://en.cppreference.com/w/cpp/language/attributes/likely

be/src/olap/types.cpp
```diff
@@ -95,110 +47,188 @@ bool is_olap_string_type(FieldType field_type) {
...
+    case OLAP_FIELD_TYPE_ARRAY: {
+        int32_t iterations = 0;
+        const auto* child_column = &column_meta_pb->children_columns(0);
+        while (child_column->type() == OLAP_FIELD_TYPE_ARRAY) {
+            iterations++;
+            child_column = &child_column->children_columns(0);
         }
+        return get_array_type_info((FieldType) child_column->type(), iterations);
+    }
+    [[likely]] default:
```
zbtzbtzbt, line 213: here`[[likely]] ` +1

be/src/olap/aggregate_func.h
```diff
@@ -98,13 +98,28 @@ struct BaseAggregateFuncs {
...
+
+template <FieldType sub_type>
+struct BaseAggregateFuncs<OLAP_FIELD_TYPE_ARRAY, sub_type> {
+    static void init(RowCursorCell* dst, const char* src, bool src_null, MemPool* mem_pool,
+                     ObjectPool* agg_pool) {
+        dst->set_is_null(src_null);
+        if (src_null) {
+            return;
+        }
+        if (sub_type != OLAP_FIELD_TYPE_ARRAY) {
```
zbtzbtzbt, line 120: small question: Is this rewriting equivalent as before?

caiconghui: origin implementation not support collection<ARRAY>, because need more detail info

be/src/olap/types.cpp
caiconghui: it seems that we cannot use UNLIKELY or LIKELY for switch now， I will remove it

adonis0147:
Hi @caiconghui , I'm doing some work to separate some functions and usage for `ArrayType`. After that, you can employ the optimization to scalar types.

caiconghui:
> Hi @caiconghui , I'm doing some work to separate some functions and usage for `ArrayType`. After that, the optimization of scalar types may be more suitable to be employed.

the work you are doing now aims to merged in dev/1.0.1? if so, we can wait to be merged until your work done.

adonis0147:
Due to the timeline of release for 1.0.1, it is ok to merge it first.

morningman (approved):
LGTM

PR #8705, opened 2022-03-28
[fix](load) fix bug that NodeChannel can not be destroyed ontime

### LINKED ISSUE #8704: [Bug] routine load return too many task error
### Search before asking

### Version

master

### What's Wrong?

This bug may cause sevaral errors:
1.  lots of `FragmentMgr cancel worker going to cancel timeout fragment` in be.INFO
2. Routine load return TOO_MANY_TASKS error
3. pstack shows lots of thread blocked in `doris::stream_load::NodeChannel::~NodeChannel`

### What You Expected?

The above error should not happen

### How to Reproduce?

occasionally

### Anything Else?

_No response_

### Code of Conduct

morningman:
# Proposed changes

Issue Number: close #8704 

## Problem Summary:

After the ReusableClosure is reset, we can not call join() method, or it will blocked forever.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/exec/tablet_sink.cpp
+++ b/be/src/exec/tablet_sink.cpp
@@ -504,7 +504,6 @@ void NodeChannel::try_send_batch(RuntimeState* state) {
         }
     }
 
-    _add_batch_closure->reset();
     int remain_ms = _rpc_timeout_ms - _timeout_watch.elapsed_time() / NANOS_PER_MILLIS;
     if (UNLIKELY(remain_ms < config::min_load_rpc_timeout_ms)) {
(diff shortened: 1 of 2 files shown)


be/src/exec/tablet_sink.cpp
```diff
@@ -504,7 +504,6 @@ void NodeChannel::try_send_batch(RuntimeState* state) {
         }
     }
 
-    _add_batch_closure->reset();
```
morningman, line 507: For reviewer:
if call reset() here, the cancel() may be called after it.

yangzhg (approved):
LGTM

PR #8745, opened 2022-03-30
[Enhancement] generate runtime filter only for tuples with conjunct

starocean999:
# Proposed changes

Issue Number: [8744](https://github.com/apache/incubator-doris/issues/8744)

## Problem Summary:

Remove useless runtime filter in some primary-foreign key join scenario in TPCH case.

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilter.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilter.java
@@ -25,10 +25,12 @@ import org.apache.doris.analysis.SlotId;
 import org.apache.doris.analysis.TupleDescriptor;
 import org.apache.doris.analysis.TupleId;
 import org.apache.doris.analysis.TupleIsNullPredicate;
+import org.apache.doris.analysis.SlotRef;
 import org.apache.doris.catalog.PrimitiveType;
 import org.apache.doris.catalog.ScalarType;
 import org.apache.doris.common.FeConstants;
 import org.apache.doris.common.IdGenerator;
+import org.apache.doris.qe.ConnectContext;
 import org.apache.doris.thrift.TRuntimeFilterDesc;
 import org.apache.doris.thrift.TRuntimeFilterType;
 
@@ -43,6 +45,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.HashSet;
 
 /**
  * Representation of a runtime filter. A runtime filter is generated from
@@ -226,8 +229,9 @@ public final class RuntimeFilter {
      * or null if a runtime filter cannot be generated from the specified predicate.
      */
     public static RuntimeFilter create(IdGenerator<RuntimeFilterId> idGen, Analyzer analyzer,
-            Expr joinPredicate, int exprOrder, HashJoinNode filterSrcNode,
-            TRuntimeFilterType type, RuntimeFilterGenerator.FilterSizeLimits filterSizeLimits) {
+                                       Expr joinPredicate, int exprOrder, HashJoinNode filterSrcNode,
+                                       TRuntimeFilterType type, RuntimeFilterGenerator.FilterSizeLimits filterSizeLimits,
+                                       HashSet<TupleId> tupleHasConjuncts) {
         Preconditions.checkNotNull(idGen);
         Preconditions.checkNotNull(joinPredicate);
         Preconditions.checkNotNull(filterSrcNode);
@@ -268,6 +272,20 @@ public final class RuntimeFilter {
         if (LOG.isTraceEnabled()) {
             LOG.trace("Generating runtime filter from predicate " + joinPredicate);
         }
+        if (ConnectContext.get().getSessionVariable().enableRemoveNoConjunctsRuntimeFilterPolicy) {
+            if (srcExpr instanceof SlotRef) {
+                if (!tupleHasConjuncts.contains(((SlotRef) srcExpr).getDesc().getParent().getId())) {
+                    // src tuple has no conjunct, don't create runtime filter
+                    return null;
+                } else {
+                    // runtime filter itself is a valid conjunct, add all the target tuple ids
+                    for (TupleId tupleId: targetSlots.keySet()) {
+                        tupleHasConjuncts.add(tupleId);
+                    }
+                }
+            }
+        }
+
         return new RuntimeFilter(idGen.getNextId(), filterSrcNode, srcExpr, exprOrder,
                 targetExpr, targetSlots, type, filterSizeLimits);
     }
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilter.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilter.java
@@ -22,10 +22,10 @@ import org.apache.doris.analysis.BinaryPredicate;
 import org.apache.doris.analysis.Expr;
 import org.apache.doris.analysis.Predicate;
 import org.apache.doris.analysis.SlotId;
+import org.apache.doris.analysis.SlotRef;
 import org.apache.doris.analysis.TupleDescriptor;
 import org.apache.doris.analysis.TupleId;
 import org.apache.doris.analysis.TupleIsNullPredicate;
-import org.apache.doris.analysis.SlotRef;
 import org.apache.doris.catalog.PrimitiveType;
 import org.apache.doris.catalog.ScalarType;
 import org.apache.doris.common.FeConstants;
@@ -42,10 +42,10 @@ import org.apache.logging.log4j.Logger;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.HashSet;
 
 /**
  * Representation of a runtime filter. A runtime filter is generated from
@@ -228,10 +228,9 @@ public final class RuntimeFilter {
      * to the join node 'filterSrcNode'. Returns an instance of RuntimeFilter
      * or null if a runtime filter cannot be generated from the specified predicate.
      */
-    public static RuntimeFilter create(IdGenerator<RuntimeFilterId> idGen, Analyzer analyzer,
-                                       Expr joinPredicate, int exprOrder, HashJoinNode filterSrcNode,
-                                       TRuntimeFilterType type, RuntimeFilterGenerator.FilterSizeLimits filterSizeLimits,
-                                       HashSet<TupleId> tupleHasConjuncts) {
+    public static RuntimeFilter create(IdGenerator<RuntimeFilterId> idGen, Analyzer analyzer, Expr joinPredicate,
+            int exprOrder, HashJoinNode filterSrcNode, TRuntimeFilterType type,
+            RuntimeFilterGenerator.FilterSizeLimits filterSizeLimits, HashSet<TupleId> tupleHasConjuncts) {
         Preconditions.checkNotNull(idGen);
         Preconditions.checkNotNull(joinPredicate);
         Preconditions.checkNotNull(filterSrcNode);
@@ -279,7 +278,7 @@ public final class RuntimeFilter {
                     return null;
                 } else {
                     // runtime filter itself is a valid conjunct, add all the target tuple ids
-                    for (TupleId tupleId: targetSlots.keySet()) {
+                    for (TupleId tupleId : targetSlots.keySet()) {
                         tupleHasConjuncts.add(tupleId);
                     }
                 }
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilterGenerator.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilterGenerator.java
@@ -79,6 +79,8 @@ public final class RuntimeFilterGenerator {
     // Generator for filter ids
     private final IdGenerator<RuntimeFilterId> filterIdGenerator = RuntimeFilterId.createGenerator();
 
+    private HashSet<TupleId> tupleHasConjuncts = null;
+
     /**
      * Internal class that encapsulates the max, min and default sizes used for creating
      * bloom filter objects.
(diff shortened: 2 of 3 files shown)


fe/fe-core/src/main/java/org/apache/doris/planner/Planner.java
```diff
@@ -266,6 +271,57 @@ public void createPlanFragments(StatementBase statement, Analyzer analyzer, TQue
...
+    private class FindTuplePredicate implements com.google.common.base.Predicate<PlanNode> {
+        private final TupleId tupleId;
+
+        FindTuplePredicate(TupleId id) {
+            tupleId = id;
+        }
+
+        @Override
+        public boolean apply(PlanNode arg) {
+            return arg.getTupleIds().size() == 1 && arg.getTupleIds().get(0) == tupleId
```
xinyiZzz, line 283: I understand that filtering conjuncts to be empty is to avoid invalid filtering in some primary-foreign key joins, because the key values ​​on the left and right sides are the same at this time; limiting the number of Tuples to 1 skips the child join node.

It is recommended to add a note here, it is not very well understood;

fe/fe-core/src/main/java/org/apache/doris/planner/Planner.java
```diff
@@ -229,6 +231,9 @@ public void createPlanFragments(StatementBase statement, Analyzer analyzer, TQue
         if (!ConnectContext.get().getSessionVariable().getRuntimeFilterMode().toUpperCase()
                 .equals(TRuntimeFilterMode.OFF.name())) {
             RuntimeFilterGenerator.generateRuntimeFilters(analyzer, rootFragment.getPlanRoot());
+            if (ConnectContext.get().getSessionVariable().enableStarSchemaRuntimeFilterPolicy) {
+                removeUselessRuntimeFilters(rootFragment.getPlanRoot());
```
xinyiZzz, line 235: Why not filter when generating rf, but re-traverse remove.

fe/fe-core/src/main/java/org/apache/doris/planner/Planner.java
```diff
@@ -266,6 +271,57 @@ public void createPlanFragments(StatementBase statement, Analyzer analyzer, TQue
...
+        }
+    }
+
+    private boolean isNodeHasConjuncts(PlanNode rootNode, TupleId tupleId) {
+        List<PlanNode> nodes = Lists.newArrayList();
+        rootNode.collectAll(new FindTuplePredicate(tupleId), nodes);
+        return !nodes.isEmpty();
+    }
+
+    private void removeUselessRuntimeFilters(PlanNode root) {
```
xinyiZzz, line 294: I understand that this is a very infrequently used func and should have a recognizable name.

Alternatively, we can look for other scenarios that require less RF and extend the range.

fe/fe-core/src/main/java/org/apache/doris/qe/SessionVariable.java
```diff
@@ -176,6 +176,8 @@
 
     public static final String BLOCK_ENCRYPTION_MODE = "block_encryption_mode";
 
+    public static final String ENABLE_STAR_SCHEMA_RUNTIME_FILTER_POLICY = "enable_star_schema_runtime_filter_policy";
```
xinyiZzz, line 179: Under star schemas such as SSB, after `set enable_star_schema_runtime_filter_policy=true`.
Will performance overall be higher than the default RF generation rules?

fe/fe-core/src/main/java/org/apache/doris/planner/Planner.java
starocean999: Note is added, pls see the new commit

fe/fe-core/src/main/java/org/apache/doris/planner/Planner.java
starocean999: we can only remove the useless runtime filters after all needed infomation are set to the plan tree nodes. Because the runtime filter itself is a valid conjuction, we need this info is set on scan node to make a correct decision. During generating phase, we are not able to see the whole picture. Some valid runtime filter hasn't been set to scan node yet.( like tpch q21 ). So after generating phase, the runtime filter are all set, then we can find the invalid ones and remove them.

( The above note is also added as comments, pls see the new commit )

fe/fe-core/src/main/java/org/apache/doris/planner/Planner.java
starocean999: Yes, I agree. If other scenarios are found, we can re-think how to extend this policy.

fe/fe-core/src/main/java/org/apache/doris/qe/SessionVariable.java
starocean999: In SSB test set, this policy will benefit q2.1, q2.2, q2.3, q4.1 and q4.3, and don't have any impact on other sqls. In tpch test set, q9, q10, q17 and q18 would run faster after removing the useless runtime filters, no impact on other sqls either.

starocean999: In SSB test set, this policy will benefit q2.1, q2.2, q2.3, q4.1 and q4.3, and don't have any impact on other sqls. In tpch test set, q9, q10, q17 and q18 would run faster after removing the useless runtime filters, no impact on other sqls either.

xinyiZzz:
The strategy of RuntimeFilter should be placed in `RuntimeFilterGenerator`, not `Planner`, which will destroy the previous structure.

Three suggestions~:
1. Before `generateFilters`, use a Map to store whether all Tuples of PlanTree have conjuncts;

2. This strategy should be placed in `RuntimeFilter create` to filter `srcExpr`, similar to the one selected in the picture; the `removeUselessRuntimeFilters` Func is no longer needed;

3. Modify the name of `ENABLE_STAR_SCHEMA_RUNTIME_FILTER_POLICY`. This policy is not directly related to star schema, for example: `ENABLE_REMOVE_NO_CONJUNCTS_RUNTIME_FILTER`. Or add enough comments to explain that this strategy is beneficial in most star schemas, otherwise it is misleading;

xinyiZzz:
image

explain the second point.

starocean999:
> The strategy of RuntimeFilter should be placed in `RuntimeFilterGenerator`, not `Planner`, which will destroy the previous structure.
> 
> Three suggestions~:
> 
>     1. Before `generateFilters`, use a Map to store whether all Tuples of PlanTree have conjuncts;
> 
>     2. This strategy should be placed in `RuntimeFilter create` to filter `srcExpr`, similar to the one selected in the picture; the `removeUselessRuntimeFilters` Func is no longer needed;
> 
>     3. Modify the name of `ENABLE_STAR_SCHEMA_RUNTIME_FILTER_POLICY`. This policy is not directly related to star schema, for example: `ENABLE_REMOVE_NO_CONJUNCTS_RUNTIME_FILTER`. Or add enough comments to explain that this strategy is beneficial in most star schemas, otherwise it is misleading;

The code is modified based on the comments, pls help review again, thanks a lot.

fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilterGenerator.java
```diff
@@ -181,7 +203,7 @@ public static void generateRuntimeFilters(Analyzer analyzer, PlanNode plan) {
      * In the bottom-up traversal of the plan tree, the filters are assigned to destination
      * (scan) nodes. Filters that cannot be assigned to a scan node are discarded.
      */
-    private void generateFilters(PlanNode root) {
+    private void generateFilters(PlanNode root, HashSet<TupleId> tupleHasConjuncts) {
```
xinyiZzz, line 206: Make `tupleHasConjuncts` a member variable of `RuntimeFilterGenerator` to avoid passing it frequently.

fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilterGenerator.java
```diff
@@ -122,6 +122,24 @@ private RuntimeFilterGenerator(Analyzer analyzer) {
         bloomFilterSizeLimits = new FilterSizeLimits(sessionVariable);
     }
 
+    static class HasConjunctPredicate implements com.google.common.base.Predicate<PlanNode> {
+        @Override
+        public boolean apply(PlanNode arg) {
+            // for simplicity, skip join node( which contains more than 1 tuple id )
+            return arg.getTupleIds().size() == 1 && !arg.conjuncts.isEmpty();
```
xinyiZzz, line 129: Returns the tuple id that satisfies the condition.

fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilterGenerator.java
```diff
@@ -122,6 +122,24 @@ private RuntimeFilterGenerator(Analyzer analyzer) {
...
+        }
+    }
+
+    private static HashSet<TupleId> collectAllTupleIdsHavingConjunct(PlanNode rootNode) {
+        HashSet<TupleId> tupleIds = new HashSet<>();
+        List<PlanNode> nodes = Lists.newArrayList();
+        rootNode.collectAll(new HasConjunctPredicate(), nodes);
```
xinyiZzz, line 136: Replace `nodes` with `RuntimeFilterGenerator::tupleHasConjuncts` , which condenses `collectAllTupleIdsHavingConjunct` into one line of code and puts it into `generateRuntimeFilters`.

fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilter.java
```diff
@@ -223,6 +227,20 @@ public static RuntimeFilter create(IdGenerator<RuntimeFilterId> idGen, Analyzer
...
         }
+        if (ConnectContext.get().getSessionVariable().enableRemoveNoConjunctsRuntimeFilterPolicy) {
+            if (srcExpr instanceof SlotRef) {
+                if (!tupleHasConjuncts.contains(((SlotRef) srcExpr).getDesc().getParent().getId())) {
+                    // src tuple has no conjunct, don't create runtime filter
+                    return null;
+                } else {
+                    // runtime filter itself is a valid conjunct, add all the target tuple ids
+                    for (TupleId tupleId: targetSlots.keySet()) {
+                        tupleHasConjuncts.add(tupleId);
```
xinyiZzz, line 238: I don't understand why `targetSlots::tupleID` is put into `tupleHasConjuncts`, can you explain, thx.

xinyiZzz (commented):
great. I still have a few small questions.

fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilterGenerator.java
starocean999: Done

fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilterGenerator.java
starocean999: Done

fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilterGenerator.java
starocean999: Done

fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilter.java
starocean999: I assume the targetSlots is where the runtime filter goes to.( if not, pls corret me ) In that case, the runtime filter can be treated as a normal conjunct. So if hash join node's build side is a scan node with no conjunct originally, but it has a runtime filter, we can also create a new runtime filter for it and push it to probe side.

xinyiZzz: You are right, no problem.

xinyiZzz:
LGTM

fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilterGenerator.java
```diff
@@ -122,6 +124,33 @@ private RuntimeFilterGenerator(Analyzer analyzer) {
...
+        // 1. The node itself has conjunct
+        // 2. it's an aggregation or exchange node with no conjunct, but its descendant have conjuncts.
+        int tupleNumBeforeCheckingChildren = tupleIds.size();
+        for (PlanNode child : node.getChildren()) {
+            collectAllTupleIdsHavingConjunct(child, tupleIds);
+        }
+        if ((node.getTupleIds().size() == 1 && !node.conjuncts.isEmpty())) {
+            // The node itself has conjunct
+            tupleIds.add(node.getTupleIds().get(0));
+        } else if ((node instanceof AggregationNode || node instanceof ExchangeNode) && tupleIds.size() > tupleNumBeforeCheckingChildren) {
```
xinyiZzz, line 139: Why only add aggregation or exchange node here, whether other nodes can also generate runtime filter if child has conjunct is satisfied.

xinyiZzz:
The test results on SSB and TPC-H under different strategies can be given:
- initially
- only scan node has conjunct (optional)
- scan and agg and exchange node has conjunct
- all nodes have conjunct

fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilterGenerator.java
starocean999: Yes I agree. This is not a perfect solution to only check agg and exchange node's children' conjunct. I added this code because after testing all tpch 22 queries, I found one useful runtime filter of q2 was removed. The goal of this modification is trying to keep all possible useful runtime filters and remove ones which are absolutely useless. ( that means we might still have some invalid ones in some cases ). For other node types, I am not so sure if this logic is still valid. So I think we can add more node types here if we come across new scenarios in future.

starocean999:
> The test results on SSB and TPC-H under different strategies can be given:
> 
>     * initially
> 
>     * only scan node has conjunct (optional)
> 
>     * scan and agg and exchange node has conjunct
> 
>     * all nodes have conjunct

<p>I tested using TPCH 3G data based on current code. CPU: 6 core 12 threads * 1, memory: 16G. the policy impact 7 queries more or less</p>
<div class="table-wrap">

runtime filter remove | q8 | q9 | q10 | q11 | q17 | q18 | q21
-- | -- | -- | -- | -- | -- | -- | --
OFF | 6.76s | 63.27s | 2.27s | 0.39s | 3.36s | 2.4s | 5.27s
ON | 6.75s | 61.12s | 2.25s | 0.33s | 3.02s | 2.1s | 4.72s

</div>

fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilterGenerator.java
xinyiZzz: I think there is no need to limit node types, including join nodes, which will make this pr less tricky and easier to understand and use.

You can try to remove `node.getTupleIds().size() == 1` and `node instanceof AggregationNode || node instanceof ExchangeNode`, and then test it on SSB and TPCH

In addition to agg and exchange nodes, the nodes that satisfy the number of tuples equal to 1 and the children have conjunct are mainly: sort, union, topn, table function, it seems that they can be generated.

xinyiZzz:
> I tested using TPCH 3G data based on current code. CPU: 6 core 12 threads * 1, memory: 16G. the policy impact 7 queries more or less
> 
> runtime filter remove	q8	q9	q10	q11	q17	q18	q21
> OFF	6.76s	63.27s	2.27s	0.39s	3.36s	2.4s	5.27s
> ON	6.75s	61.12s	2.25s	0.33s	3.02s	2.1s	4.72s

seems to have some effect~

fe/fe-core/src/main/java/org/apache/doris/planner/RuntimeFilterGenerator.java
xinyiZzz: If don't need to limit the node type, this strategy can be placed after the entire runtime filter is generated, and you only need to traverse `runtimeFiltersByTid`.

starocean999: I tested on TPCH queries, only join node can be skipped when collecting the tuple ids of nodes which having conjunct. currently, join node has more then 1 descendant tuple ids, it's safe to skip it. And the first version is to remove the invalid runtime filters after entire runtime filter are generated, do you mean it's better to change it back to old way?

xinyiZzz:
LGTM
This still looks trick, but it does work on TPCH

xinyiZzz (approved):

PR #8764, opened 2022-03-30
[fix](OlapScanner)fix bitmap or hll's OOM

SleepyBear96:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:
The _agg_object_pool in row_batch is used to hold the pointer of bitmap or hll data, which will be released later.
When a tuple check direct conjuncts or check pushdown conjuncts fail, the memory holding by _agg_object_pool which belong to the tuple also should be released. Otherwise, it will lead up to OOM。

## Checklist(Required)

## Further comments


be/src/exec/olap_scanner.cpp
```diff
@@ -335,6 +335,7 @@ Status OlapScanner::get_batch(RuntimeState* state, RowBatch* batch, bool* eof) {
                         // make sure to reset null indicators since we're overwriting
                         // the tuple assembled for the previous row
                         tuple->init(_tuple_desc->byte_size());
+                        batch->agg_object_pool()->remove_last_one();
```
cambyzju, line 338: Could you give some more detail about the OOM?

If one thread(T2) add an new object into agg_object_pool, just before remove_last_one called by this thread(T1).
I think it may remove object added by T2, not the one added by T1?

SleepyBear96: _batch_ in _Status OlapScanner::get_batch(RuntimeState*, RowBatch*, bool*)_ was allocated by _void OlapScanNode::scanner_thread(OlapScanner*)_, every scanner_thread has its own _batch_, it not a multi-thread case。

cambyzju: Thanks for your reply.

1. Could you give me some detail steps? I want to try to reproduce the problem.
2. If one tuple has bitmap and hll column at the same time, remove_last_one should be called twice?

be/src/exec/olap_scanner.cpp
```diff
@@ -286,6 +286,8 @@ Status OlapScanner::get_batch(RuntimeState* state, RowBatch* batch, bool* eof) {
     int64_t raw_bytes_threshold = config::doris_scanner_row_bytes;
     {
         SCOPED_TIMER(_parent->_scan_timer);
+        ObjectPool tmp_object_pool;
```
cambyzju, line 289: please add more comments to explain why we need this tmp object pool.

SleepyBear96: Done.

be/src/common/config.h
```diff
@@ -729,6 +729,7 @@ CONF_String(rpc_load_balancer, "rr");
 // so we set a soft limit, default is 1MB
 CONF_mInt32(string_type_length_soft_limit_bytes, "1048576");
 
+CONF_Int32(object_pool_buffer_size, "100");
```
morningman, line 732: Add comment for this new config

be/src/common/object_pool.h
```diff
@@ -54,6 +54,10 @@ class ObjectPool {
         for (Element& elem : _objects) elem.delete_fn(elem.obj);
         _objects.clear();
     }
+    
+    uint64_t size() {
```
morningman, line 58: Need a `std::lock_guard<SpinLock> l(_lock);`?

be/src/common/config.h
SleepyBear96: Done.

be/src/common/object_pool.h
SleepyBear96: Done.

morningman:
See #9205

PR #8766, opened 2022-03-30
[feature-wip](array-type) support explode and explode_outer table function

cambyzju:
# Proposed changes

Issue Number: close #7570

## Problem Summary:
explode(ArrayColumn) desc:
> Create a row for each element in the array column. 

explode_outer(ArrayColumn) desc:
> Create a row for each element in the array column. Unlike explode, if the array is null or empty, it returns null.

Usage example:
1. create a table with array column, and insert some data;
2. open enable_lateral_view and enable_vectorized_engine;
```
set enable_lateral_view = true;
set enable_vectorized_engine=true;
```
3. use explode_outer
```
> select * from array_test;
+------+------+--------+
| k1   | k2   | k3     |
+------+------+--------+
|    3 | NULL | NULL   |
|    1 |    2 | [1, 2] |
|    2 |    3 | NULL   |
|    4 | NULL | []     |
+------+------+--------+

> select k1,explode_column from array_test LATERAL VIEW explode_outer(k3) TempExplodeView as explode_column;
+------+----------------+
| k1   | explode_column |
+------+----------------+
|    1 |              1 |
|    1 |              2 |
|    2 |           NULL |
|    4 |           NULL |
|    3 |           NULL |
+------+----------------+
```
4. explode usage example. explode return empty rows while the ARRAY is null or empty
```
> select k1,explode_column from array_test LATERAL VIEW explode(k3) TempExplodeView as explode_column;
+------+----------------+
| k1   | explode_column |
+------+----------------+
|    1 |              1 |
|    1 |              2 |
+------+----------------+
```

## Checklist(Required)

## Further comments


fe/fe-core/src/main/java/org/apache/doris/catalog/FunctionSet.java
```diff
@@ -2384,5 +2385,18 @@ private void initTableFunction() {
                 "_ZN5doris19DummyTableFunctions22explode_numbersEPN9doris_udf15FunctionContextERKNS1_9IntValE",
                 null, null, true));
         tableFunctions.put(EXPLODE_NUMBERS, explodeNumbers);
+
+        List<Function> explodes = Lists.newArrayList();
+        explodes.add(ScalarFunction.createBuiltin(
+                EXPLODE_OUTER, Type.INT, Function.NullableMode.DEPEND_ON_ARGUMENT,
```
EmmyMiao87, line 2391: The 'explode_outer' function should be 'always_nullable'

fe/fe-core/src/main/java/org/apache/doris/catalog/FunctionSet.java
```diff
@@ -2384,5 +2385,18 @@ private void initTableFunction() {
...
+                Lists.newArrayList(new ArrayType(Type.INT)), false,
+                "_ZN5doris19DummyTableFunctions13explode_outerEPN9doris_udf15FunctionContextERKNS1_13CollectionValE",
+                null, null, true));
+        explodes.add(ScalarFunction.createBuiltin(
+                EXPLODE_OUTER, Type.VARCHAR, Function.NullableMode.DEPEND_ON_ARGUMENT,
```
EmmyMiao87, line 2396: same as above

fe/fe-core/src/main/java/org/apache/doris/catalog/FunctionSet.java
cambyzju: done

fe/fe-core/src/main/java/org/apache/doris/catalog/FunctionSet.java
cambyzju: done

be/src/vec/exprs/table_function/vexplode.cpp
```diff
@@ -0,0 +1,112 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+#include "vec/exprs/table_function/vexplode.h"
+
+#include "vec/exprs/vexpr.h"
+
+namespace doris::vectorized {
+
+VExplodeTableFunction::VExplodeTableFunction() {
```
BiteTheDDDDt, line 24: Why this file do not named `vexplode_outer`?

be/src/vec/exprs/table_function/vexplode.cpp
```diff
@@ -0,0 +1,112 @@
...
+    _eos = false;
+
+    if (_array_null_map && _array_null_map[row_idx]) {
+        _is_current_empty = true;
+        _cur_size = 0;
+        _cur_offset = 0;
+        _pos = 0;
+    } else {
+        _cur_size =
+                _array_column->get_offsets()[row_idx] - _array_column->get_offsets()[row_idx - 1];
```
BiteTheDDDDt, line 71: Will there be an error when `row_idx==0`

be/src/vec/functions/function_fake.h
```diff
@@ -47,6 +49,15 @@ struct FunctionExplodeNumbersImpl {
     }
 };
 
+struct FunctionExplodeImpl {
+    static constexpr auto name = "explode_outer";
+    static DataTypePtr get_return_type_impl(const DataTypes& arguments) {
+        DCHECK(is_array(arguments[0])) << arguments[0]->get_name() << " not supported";
+        return make_nullable(
```
BiteTheDDDDt, line 56: Is it a `ALWAYS_NULLABLE` function?

be/src/vec/exprs/table_function/vexplode.cpp
cambyzju: offsets[-1] always has value 0.

// pod_array.h  
* The template parameter `pad_left` - always allocate memory before 0th element of the array (rounded up to the whole number of elements)
  *  and zero initialize -1th element. It allows to use -1th element that will have value 0.
  * This gives performance benefits when converting an array of offsets to array of sizes.

be/src/vec/exprs/table_function/vexplode.cpp
cambyzju: for Array type, I want to implement two table function: explode and explode_outer.
Explode_outer only has a little different with explode, they could reuse most codes.
So I use explode to name the file and class.

EmmyMiao87 (commented):
The FE part is OK

BiteTheDDDDt:
LGTM

cambyzju:
rebase upstream/master to avoid conflict.

be/src/vec/exprs/table_function/vexplode.h
```diff
@@ -43,4 +47,13 @@ class VExplodeTableFunction : public TableFunction {
     size_t _pos;
 };
 
+class VExplodeOuterTableFunction : public VExplodeTableFunction {
+public:
```
BiteTheDDDDt, line 51: You can use template specific on `TableFunctionCreator<VExplodeTableFunction>` to  reduce a class `VExplodeOuterTableFunction` just like `TableFunctionCreator<VExplodeJsonArrayTableFunction>`.
And so that we can have a unified design style.

BiteTheDDDDt:
LGTM

be/src/vec/exprs/table_function/vexplode.h
cambyzju: thanks for your advice, I have updated the codes.

morningman (approved):
LGMT

PR #8768, opened 2022-03-31
[improvement](storage) Add more detailed timer on SegmentIter in profile

zenoyang:
# Proposed changes

- Fix some profile indicators of vectorized query were not counted: ScanTime, VectorPredEvalTime
- Add some metrics for analyzing storage layer performance consumption: LazyReadTime, PredColumnReadTime, ShortPredEvalTime, OutputColumnTime

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -129,6 +129,7 @@ void OlapScanNode::_init_counter(RuntimeState* state) {
     _short_cond_timer = ADD_TIMER(_segment_profile, "ShortPredEvalTime");
     _pred_col_read_timer = ADD_TIMER(_segment_profile, "PredColumnReadTime");
     _lazy_read_timer = ADD_TIMER(_segment_profile, "LazyReadTime");
+    _output_col_timer = ADD_TIMER(_segment_profile, "OutputColumnTime");
 
     _stats_filtered_counter = ADD_COUNTER(_segment_profile, "RowsStatsFiltered", TUnit::UNIT);
     _bf_filtered_counter = ADD_COUNTER(_segment_profile, "RowsBloomFilterFiltered", TUnit::UNIT);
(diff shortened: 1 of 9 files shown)


wangbo:
1 Timer itself has an overhead, I don't think we need so many Timer by default and collect it in Prod Env.
We can use GenericTimer when we meet obvious performance bottleneck in Prod Env or Test Env.
We'd better not add more timer than before.
So I suggest you just add ```ShortPredEvalTime```  or using ```PredEvalTime``` to cover short and vec evaluation

2 ```GeneralDebugTimer```  is open by default currently, I think we can disable it by default.

zenoyang:
> 1 Timer itself has an overhead, I don't think we need so many Timer by default and collect it in Prod Env. We can use GenericTimer when we meet obvious performance bottleneck in Prod Env or Test Env. We'd better not add more timer than before. So I suggest you just add `ShortPredEvalTime` or using `PredEvalTime` to cover short and vec evaluation. For other Timer, you can disable it by default or using `GeneralDebugTimer` instead. Explaining how to Timer `SegmentIter` is helpful.
> 
> 2 `GeneralDebugTimer` is open by default currently, I think we can disable it by default.

I think there is a scenario that needs to be considered. For example, a certain sql query of a user in an online environment is slow. If the profile is relatively detailed, we can directly use the online environment to analyze the profile through the profile to quickly locate the bottleneck. Instead of adding relevant profile indicators by changing the code, then compiling, deploying, and analyzing the problem.
Of course, the timer itself has an overhead, so there needs to be a trade-off. ReadColumn is a very time consuming operation (shown below), please confirm whether it is necessary?

```
- BlockLoadTime: 8s695ms
- PredColumnReadTime: 2s536ms
- LazyReadTime: 5s422ms
- ShortPredEvalTime: 433.292ms
```

I will delete other new timers and turn off GeneralDebugTimer by default.

wangbo:
The purpose of this PR to help locate and analyze the query bottleneck is good.
I will do some code refactor to make lazy materialization easier to understand.
So this PR can be merged first.
Performance test would be will be included in my next PR.

wangbo (approved):
LGTM

zenoyang:
> The purpose of this PR to help locate and analyze the query bottleneck is good. I will do some code refactor to make lazy materialization easier to understand. So this PR can be merged first. Performance test would be will be included in my next PR.

ok, thanks

PR #8816, opened 2022-04-02
[Fix] fix restore issue when  meta version is too low

GoGoWen:
# Proposed changes

when restore snapshot from 0.13 to master,  the restore job is pending for long time.   However,  we get error "Could not set meta version to 93 since it is lower than minimum required version 100" in log。  we should cancel restore job once get that error.

## Problem Summary:
when restore snapshot from 0.13 to master,  the restore job is pending for long time.   However,  we get error "Could not set meta version to 93 since it is lower than minimum required version 100" in log。  we should cancel restore job once get that error.

## Checklist(Required)

## Further comments


morningman (approved):
LGTM

PR #8838, opened 2022-04-03
[fix](ut) fix fe run CreateTableAsSelectStmtTest ,UserPropertyTest,  ProjectPlannerFunctionTest and AggregateTest failed

caiconghui:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/FunctionCallExpr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/FunctionCallExpr.java
@@ -246,7 +246,7 @@ public class FunctionCallExpr extends Expr {
         if (params.isDistinct()) {
             sb.append("DISTINCT ");
         }
-        int len = params.exprs().size();
+        int len = children.size();
         List<String> result = Lists.newArrayList();
         if (fnName.getFunction().equalsIgnoreCase("json_array") ||
                 fnName.getFunction().equalsIgnoreCase("json_object")) {
@@ -265,7 +265,7 @@ public class FunctionCallExpr extends Expr {
                     fnName.getFunction().equalsIgnoreCase("sm4_encrypt"))) {
                 result.add("\'***\'");
             } else {
-                result.add(params.exprs().get(i).toSql());
+                result.add(children.get(i).toSql());
             }
         }
         sb.append(Joiner.on(", ").join(result)).append(")");
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/FunctionCallExpr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/FunctionCallExpr.java
@@ -236,14 +236,14 @@ public class FunctionCallExpr extends Expr {
                 && fnParams.isStar() == o.fnParams.isStar();
     }
 
-    private String paramsToSql(FunctionParams params) {
+    private String paramsToSql() {
         StringBuilder sb = new StringBuilder();
         sb.append("(");
 
-        if (params.isStar()) {
+        if (fnParams.isStar()) {
             sb.append("*");
         }
-        if (params.isDistinct()) {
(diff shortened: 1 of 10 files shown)


fe/fe-core/src/main/java/org/apache/doris/analysis/FunctionCallExpr.java
```diff
@@ -246,7 +246,7 @@ private String paramsToSql(FunctionParams params) {
         if (params.isDistinct()) {
             sb.append("DISTINCT ");
         }
-        int len = params.exprs().size();
```
morningman, line 249: Looks like we don't need `params` as the parameter of this method.
Because `fnParams` is a field of this class.

fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -3116,12 +3116,12 @@ public void createTableAsSelect(CreateTableAsSelectStmt stmt) throws DdlExceptio
                 }
                 TypeDef typeDef;
                 Expr resultExpr = resultExprs.get(i);
-                // varchar/char transfer to string
-                if (resultExpr.getType().isStringType()) {
+                if (resultExpr.getType().isStringType() && resultExpr.getType().getLength() < 0) {
```
morningman, line 3119: Why adding `resultExpr.getType().getLength() < 0` condition?
If this is a bug, please described in issue.

fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -3116,12 +3116,12 @@ public void createTableAsSelect(CreateTableAsSelectStmt stmt) throws DdlExceptio
...
                     typeDef = new TypeDef(Type.STRING);
-                } else {
+                }  else {
                     typeDef = new TypeDef(resultExpr.getType());
                 }
+                typeDef = new TypeDef(resultExpr.getType());
```
morningman, line 3124: Duplicate line?

fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
caiconghui: -1 means that we don't know the varchar length ,so we transfer it to string, but if we know the length of string, just keep it same as before

fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
caiconghui: I will remove it

fe/fe-core/src/main/java/org/apache/doris/analysis/SlotRef.java
```diff
@@ -74,9 +74,6 @@ public SlotRef(SlotDescriptor desc) {
         this.type = desc.getType();
         // TODO(zc): label is meaningful
         this.label = null;
-        if (this.type.equals(Type.CHAR)) {
```
morningman, line 77: Are you sure this is OK to remove this?

caiconghui: already do some test, but I am not sure for all case, because the motivation to transfer char to varchar in unknown for me

morningman: If it is unknown, better not modify it. Or it is very hard to track.
And please do not do 'bug fix' in this PR, because the title of this PR is `[fix](ut)`

caiconghui: ok

morningman (approved):
LGTM

PR #8855, opened 2022-04-06
[Refactor] Change ALL OLAPStatus to Status

yiguolei:
Currently, there are 2 status code in BE, one is common/Status.h, and the other is olap/olap_define.h called OLAPStatus.
OLAPStatus is just an enum type, it is very simple and could not save many informations, I will unify these code to common/Status.

# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments


jackwener (dismissed):
TL;DR

morningman:
We should separate this PR

be/CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ endif ()
 
 # Set boost/stacktrace use backtrace api to unwind
 add_definitions(-DBOOST_STACKTRACE_USE_BACKTRACE)
-add_definitions(-DPRINT_ALL_ERR_STATUS_STACKTRACE)
+#add_definitions(-DPRINT_ALL_ERR_STATUS_STACKTRACE)
```
caiconghui, line 54: why comment out this config？

be/src/agent/agent_server.cpp
```diff
@@ -208,17 +208,14 @@ void AgentServer::submit_tasks(TAgentResult& agent_result,
...
 void AgentServer::make_snapshot(TAgentResult& t_agent_result,
                                 const TSnapshotRequest& snapshot_request) {
-    Status ret_st;
     string snapshot_path;
     bool allow_incremental_clone = false;
-    OLAPStatus err_code =
+    Status err_code =
             SnapshotManager::instance()->make_snapshot(snapshot_request, &snapshot_path, &allow_incremental_clone);
-    if (err_code != OLAP_SUCCESS) {
+    if (!err_code) {
```
caiconghui, line 215: why not err_code.ok()? !err_code is same with !err_code.ok() ??

wangbo: I think maybe ret_code is better.
```err_code ``` is not a good name.It seems like ```!error_code``` means no error.

File: be/CMakeLists.txt (hunk 1 above)
Comment (yiguolei) [reply be/CMakeLists.txt:54]: If this marco is enabled, BE will print all error status to log, it may cause performance problem. So that I disabled it by default. Should enable it manually if it is needed.

File: be/src/agent/agent_server.cpp (hunk 2 above)
Comment (yiguolei) [reply be/src/agent/agent_server.cpp:215]: I think we should not modify variable name in this PR. Because there maybe multiple status variable in same function. Modify the variable name may cause logic bug. And also, there are too many variables, it may take lot of time.
We may refactor the status variable name in other PRs.

File: be/src/common/status.cpp [hunk 3]
```diff
@@ -4,10 +4,35 @@
...
+#include "gutil/strings/fastmem.h" // for memcpy_inlined
 namespace doris {
 
+constexpr int MAX_ERROR_NUM = 65536;
+struct ErrorCodeState {
+    int16_t error_code = 0;
+    bool stacktrace = true;
+    std::string description;
+    size_t count = 0;   // Used for count the number of error happens
+    std::mutex mutex;   // lock guard for count state
```
caiconghui, line 19: why not use atomic？

yiguolei: This filed is not used yet. Just add a filed and used in the future and I could remove it.

be/src/olap/task/engine_batch_load_task.cpp
```diff
@@ -80,17 +80,16 @@ OLAPStatus EngineBatchLoadTask::execute() {
...
         }
     } else if (_push_req.push_type == TPushType::DELETE) {
-        OLAPStatus delete_data_status = _delete_data(_push_req, _tablet_infos);
-        if (delete_data_status != OLAPStatus::OLAP_SUCCESS) {
-            OLAP_LOG_WARNING("delete data failed. status: %d, signature: %ld", delete_data_status,
-                             _signature);
+        Status delete_data_status = _delete_data(_push_req, _tablet_infos);
+        if (delete_data_status != Status::OK()) {
+            LOG(WARNING) << "delete data failed. status:" << delete_data_status << " signature:" << _signature;
             status = Status::InternalError("Delete data failed");
```
morningman, line 86:
```suggestion
            status = delete_data_status;
```

be/src/olap/task/engine_batch_load_task.cpp
```diff
@@ -257,12 +256,12 @@ Status EngineBatchLoadTask::_process() {
...
-        OLAPStatus push_status = _push(_push_req, _tablet_infos);
+        Status push_status = _push(_push_req, _tablet_infos);
         time_t push_finish = time(nullptr);
         LOG(INFO) << "Push finish, cost time: " << (push_finish - push_begin);
-        if (push_status == OLAPStatus::OLAP_ERR_PUSH_TRANSACTION_ALREADY_EXIST) {
+        if (push_status == Status::OLAPInternalError(OLAP_ERR_PUSH_TRANSACTION_ALREADY_EXIST)) {
             status = Status::OK();
-        } else if (push_status != OLAPStatus::OLAP_SUCCESS) {
+        } else if (push_status != Status::OK()) {
             status = Status::InternalError("Unknown");
```
morningman, line 265:
```suggestion
            status = push_status;
```

be/src/vec/exec/volap_scanner.cpp
```diff
@@ -57,12 +57,12 @@ Status VOlapScanner::get_block(RuntimeState* state, vectorized::Block* block, bo
         do {
             // Read one block from block reader
             auto res = _tablet_reader->next_block_with_aggregation(block, nullptr, nullptr, eof);
-            if (res != OLAP_SUCCESS) {
+            if (!res) {
                 std::stringstream ss;
-                ss << "Internal Error: read storage fail. res=" << res
+                ss << "Internal Error: read storage fail. res=" << res.to_string()
```
morningman, line 62: this `ss` can be removed

be/src/olap/rowset/CMakeLists.txt
```diff
@@ -46,3 +46,5 @@ add_library(Rowset STATIC
     beta_rowset.cpp
     beta_rowset_reader.cpp
     beta_rowset_writer.cpp)
+
+target_compile_options(Rowset PUBLIC "-Wno-error=maybe-uninitialized")
```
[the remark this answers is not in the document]
yiguolei: Some variables in rowset is not initialized, so that I has to add this option to compile successfully.  I will add another PR to init all variables and remove this option.

be/src/olap/delete_handler.cpp
```diff
@@ -48,20 +48,20 @@ using google::protobuf::RepeatedPtrField;
...
+        return Status::OLAPInternalError(OLAP_ERR_DELETE_INVALID_PARAMETERS);
     }
 
     // Check whether the delete condition meets the requirements
     for (const TCondition& condition : conditions) {
-        if (check_condition_valid(schema, condition) != OLAP_SUCCESS) {
+        if (check_condition_valid(schema, condition) != Status::OK()) {
             LOG(WARNING) << "invalid condition. condition=" << ThriftDebugString(condition);
-            return OLAP_ERR_DELETE_INVALID_CONDITION;
+            return Status::OLAPInternalError(OLAP_ERR_DELETE_INVALID_CONDITION);
```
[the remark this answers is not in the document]
yiguolei: Actually, I do not know. I think we'd better not change logic in this PR.

be/src/http/action/compaction_action.cpp
```diff
@@ -207,44 +207,42 @@ OLAPStatus CompactionAction::_execute_compaction_callback(TabletSharedPtr tablet
...
-    OLAPStatus status = OLAP_SUCCESS;
+    Status res = Status::OK();
     if (compaction_type == PARAM_COMPACTION_BASE) {
         BaseCompaction base_compaction(tablet);
-        OLAPStatus res = base_compaction.compact();
-        if (res != OLAP_SUCCESS) {
-            if (res == OLAP_ERR_BE_NO_SUITABLE_VERSION) {
+        res = base_compaction.compact();
+        if (!res) {
+            if (res == Status::OLAPInternalError(OLAP_ERR_BE_NO_SUITABLE_VERSION)) {
```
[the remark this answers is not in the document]
yiguolei: Yes. We'd better not change logic here. And there are many code like this because I use search and replace to change the code. Maybe we could use some code like isNoSuitableVersionError here to check it.

be/src/olap/delete_handler.cpp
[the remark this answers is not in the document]
yiguolei: And I think we should refactor the LOG in the future. We could add more detail info to status and do not log it at every method.

be/src/olap/rowset/alpha_rowset.cpp
```diff
@@ -175,63 +175,61 @@ OLAPStatus AlphaRowset::split_range(const RowCursor& start_key, const RowCursor&
...
     }
 
     step_pos = start_pos;
     VLOG_NOTICE << "start_pos=" << start_pos.segment << ", " << start_pos.index_offset;
 
     //find last row_block is end_key is given, or using last_row_block
-    if (largest_segment_group->find_short_key(end_key, &helper_cursor, false, &end_pos) !=
-        OLAP_SUCCESS) {
-        if (largest_segment_group->find_last_row_block(&end_pos) != OLAP_SUCCESS) {
+    if (!largest_segment_group->find_short_key(end_key, &helper_cursor, false, &end_pos)) {
```
wangbo, line 205:
```suggestion
    if (largest_segment_group->find_short_key(end_key, &helper_cursor, false, &end_pos) != Status::OK()) {
```

be/src/olap/rowset/alpha_rowset.cpp
```diff
@@ -175,63 +175,61 @@ OLAPStatus AlphaRowset::split_range(const RowCursor& start_key, const RowCursor&
...
+        if (!largest_segment_group->find_last_row_block(&end_pos)) {
```
wangbo, line 206:
```suggestion
        if (largest_segment_group->find_last_row_block(&end_pos) != Status::OK()) {
```

be/src/olap/rowset/bit_field_reader.cpp
```diff
@@ -30,40 +30,40 @@ BitFieldReader::~BitFieldReader() {
...
 }
 
-OLAPStatus BitFieldReader::_read_byte() {
-    OLAPStatus res = OLAP_SUCCESS;
+Status BitFieldReader::_read_byte() {
+    Status res = Status::OK();
 
     if (_byte_reader->has_next()) {
-        if (OLAP_SUCCESS != (res = _byte_reader->next(&_current))) {
+        if (!(res = _byte_reader->next(&_current))) {
```
wangbo, line 50:
```suggestion
        if (_byte_reader->next(&_current) != Status::OK()) {
```

be/src/olap/rowset/bit_field_reader.cpp
```diff
@@ -30,40 +30,40 @@ BitFieldReader::~BitFieldReader() {
...
 }
 
-OLAPStatus BitFieldReader::next(char* value) {
-    OLAPStatus res = OLAP_SUCCESS;
+Status BitFieldReader::next(char* value) {
+    Status res = Status::OK();
 
     if (0 == _bits_left) {
-        if (OLAP_SUCCESS != (res = _read_byte())) {
+        if (!(res = _read_byte())) {
```
wangbo, line 66:
```suggestion
        if (_read_byte() != res) {
```

be/src/olap/rowset/bit_field_reader.cpp
```diff
@@ -72,23 +72,23 @@ OLAPStatus BitFieldReader::next(char* value) {
...
+    return Status::OK();
 }
 
-OLAPStatus BitFieldReader::seek(PositionProvider* position) {
-    OLAPStatus res = OLAP_SUCCESS;
+Status BitFieldReader::seek(PositionProvider* position) {
+    Status res = Status::OK();
 
-    if (OLAP_SUCCESS != (res = _byte_reader->seek(position))) {
+    if (!(res = _byte_reader->seek(position))) {
```
wangbo, line 81:
```suggestion
    if (_byte_reader->seek(position) != res) {
```

be/src/olap/rowset/column_data.cpp
```diff
@@ -138,13 +138,13 @@ OLAPStatus ColumnData::_seek_to_block(const RowBlockPosition& block_pos, bool wi
...
-            return OLAP_ERR_MALLOC_ERROR;
+            return Status::OLAPInternalError(OLAP_ERR_MALLOC_ERROR);
         }
 
         _current_segment = block_pos.segment;
         auto res = _segment_reader->init(_is_using_cache);
-        if (OLAP_SUCCESS != res) {
-            OLAP_LOG_WARNING("fail to init segment reader. [res=%d]", res);
+        if (!res.ok()) {
+            LOG(WARNING) << "fail to init segment reader. res = " << res;
```
wangbo, line 147: Why not using OLAP_LOG_WARNING?

caiconghui (approved):
the logic LGTM, but we need to keep one check status style in the future.

yiguolei:
Add some performance test:

class Status {

public:

  Status() {
    _length = 0;
    _precise_code = 0;
  }

  Status(int i) {
    _precise_code = i % 17;
  }

  static Status OK() {
    return Status();
  }

  static Status ERROR(int i) {
    return Status(i);
  }

  bool operator == (const Status& st) { return _precise_code == st._precise_code; }

  union {
      char _state[2048];

      struct {
          // Message length == HEADER(7 bytes) + message size
          // Sometimes error message is empty, so that we could not use length==0 to indicate
          // whether there is error happens
          int64_t _length : 32;      
          int64_t _code : 8;          
          int64_t _precise_code : 16;
          int64_t _message : 8;       // save message since here
      };
  };

};

static void newstatus(benchmark::State& state) {
  // Code inside this loop is measured repeatedly
  bool result = false;
  int i = 0;
  for (auto _ : state) {
    Status status(i - 1);
    ++i;
    if (status == Status::ERROR(i)) {
      result = true;
    } else {
      result = false;
    }
  }
}
// Register the function as a benchmark
BENCHMARK(newstatus);

static void olapstatus(benchmark::State& state) {
  // Code before the loop is not measured
  bool result = false;
  int i = 0;
  for (auto _ : state) {
    //Status status(i - 1);
    ++i;
    if (i == 16) {
      result = true;
    } else {
      result = false;
    }
  }
}
BENCHMARK(olapstatus);

new status is slower than olapstatus about 3.6%. 

Since there are lot of other work during query and the status is returned batch by batch, so that the impact could be ignored.

yiguolei:
Add SSB benchmark:
Environment:
1. ONE BE, ONE FE
2. CREATE LINEORDER TABLE with bucket num == 1, lineorder table with 600037902 rows.
3. set vectoried_engine = true;
4. set scanner thread == 1;
5. using select count(lo_shipmode) from lineorder; to test

new status:
image

old olapstatus:
image

It is very curious new status is a litter better than olapstatus.

be/src/olap/rowset/column_data.cpp
yiguolei: OLAP_LOG_XXXX will be replaced with LOG(WARNING) in the future. LOGGING SYSTEM should also be uniformed.

morningman:
I tested with jmeter, 50 concurrency, 1 FE, 1 BE
`select * from tbl1 where k1=1 limit 1`

The avg qps decreased from 1400/s to 1300/s, about 7% performance impact at high concurrency senario.
So I think it deserved to improve it

chenlinzhong:
I have a question, why print the stack? 

- For stack, the common practice in the industry is to print when the process exits
- By default, the stack will be printed when an error is reported, resulting in a full screen of stack information，which is not good for viewing the log

image


image

PR #8877, opened 2022-04-07
[feature](function) Support nth value window function on vectorized e…

luwei16:
# Proposed changes

Issue Number: close #8877 

## Problem Summary:

## Checklist(Required)

## Further comments


fe/fe-core/src/main/java/org/apache/doris/catalog/FunctionSet.java
```diff
@@ -2336,6 +2336,10 @@ private void initAggregateBuiltins() {
                     prefix + OFFSET_FN_UPDATE_SYMBOL.get(t),
                     null, null, null, true));
 
+            addBuiltin(AggregateFunction.createAnalyticBuiltin(
+                     "nth_value", Lists.newArrayList(t, Type.BIGINT, t), t, t,
```
zhangstar333, line 2340: It seems that MySQL has no default value and directly returns null

luwei16: Default value have been removed

luwei16: > It seems that MySQL has no default value and directly returns null

Default value argument have been removed

be/src/vec/aggregate_functions/aggregate_function_window.h
```diff
@@ -363,6 +363,47 @@ struct WindowFunctionLastData : Data {
...
+        }
+
+        if (!is_set_frame_start) {
+            this->frame_start = std::max<int64_t>(frame_start, partition_start);
+            is_set_frame_start = true;
+        }
+        frame_end = std::min<int64_t>(frame_end, partition_end);
+
+        if (this->frame_start + offset > frame_end) {
+            this->set_is_null();
```
zhangstar333, line 382: Maybe this is the case:  window is above partition_start or under partition_end, should be null also

be/src/vec/aggregate_functions/aggregate_function_window.h
```diff
@@ -363,6 +363,47 @@ struct WindowFunctionLastData : Data {
...
+struct WindowFunctionNthData : Data {
+    void add_range_single_place(int64_t partition_start, int64_t partition_end, int64_t frame_start,
+                                int64_t frame_end, const IColumn** columns) {
+        int64_t offset = columns[1]->get64(partition_start);
+        if (this->has_set_value()) {
+            return;
+        }
+
+        if (!is_set_frame_start) {
+            this->frame_start = std::max<int64_t>(frame_start, partition_start);
```
zhangstar333, line 376: A little confused about this judgment is used to?

be/src/vec/aggregate_functions/aggregate_function_window.h
luwei16: > Maybe this is the case: window is above partition_start or under partition_end, should be null also

this->frame_start = std::max<int64_t>(frame_start, partition_start);
frame_end = std::min<int64_t>(frame_end, partition_end); 

Because of these two statements, the window must be inside the partition, so the above case should be no problem

be/src/vec/aggregate_functions/aggregate_function_window.h
luwei16: The frame_start parameter is not the real window start, only when reset is called, 
frame_start is equal to the real window start, so I use this->frame_start to save the real window start value

Below is an example

the simple table is：
|class | id | score| 
1           1    80   
1           2    90    
1           3    70   
1           4    95

the sql is：
select class_id, student_id, score, nth_value(id, 3) over (partition by class order by score desc) as 'nth' from table

The parameters for add_range_single_place to run four times are：

partition_start partition_end frame_start frame_end
0                             4                       0                1             
0                             4                       1                 2                 
0                             4                       2                3 
0                             4                       3                4

PR #8900, opened 2022-04-07
[Refactor-step1] Add OLAPInternalError to status

yiguolei:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

This job is part of replace OLAPStatus with common/Status. I add some method to Status, including:
1. constructor for OLAPInternalError
2. add stacktrace during construct error status, it is useful for unknown errors or cores
3. override << operator for log message
4. fix bug: length should be to total length instead of message length, since the code is error code but the error msg is empty.
## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/common/status.h
+++ b/be/src/common/status.h
@@ -329,7 +329,8 @@ private:
 };
 
 // Override the << operator, it is used during LOG(INFO) << "xxxx" << status;
-std::ostream & operator << (std::ostream & ostr, const Status & param)
+// Add inline here to dedup many includes
+inline std::ostream & operator << (std::ostream & ostr, const Status & param)
 {
     return ostr << param.to_string();
 }
--- a/be/src/common/status.h
+++ b/be/src/common/status.h
@@ -4,8 +4,12 @@
 
 #pragma once
 
+#include <iostream>
 #include <string>
 #include <vector>
+
+#include <boost/stacktrace.hpp>
+#include <glog/logging.h>
 
 #include "common/compiler_util.h"
 #include "common/logging.h"
@@ -38,7 +42,7 @@ public:
     // same as copy c'tor
     Status& operator=(const Status& rhs) {
         if (rhs._length) {
-            memcpy(_state, rhs._state, rhs._length + Status::HEADER_LEN);
+            memcpy(_state, rhs._state, rhs._length);
         } else {
             _length = 0;
         }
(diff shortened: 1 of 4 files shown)


be/src/common/status.h
```diff
@@ -151,8 +155,18 @@ class Status {
         return Status(TStatusCode::DATA_QUALITY_ERROR, msg, precise_code, msg2);
     }
 
+    // A wrapper for OLAPStatus
+    //      Precise code is for OLAPStatus's enum value
+    //      All Status Error is treated as Internal Error 
+    static Status OLAPInternalError(int16_t precise_code, const Slice& msg = Slice()) {
+        #ifdef PRINT_ALL_ERR_STATUS_STACKTRACE
+        LOG(WARNING) << "Error occurred, error code = " << precise_code << ", with message: " << msg
+                     << "\n" << boost::stacktrace::stacktrace();
```
xinyiZzz, line 164: Can the stacktrace be printed for all errors, but the same stacktrace will not be printed after printing several times, like JAVA, until the next restart of BE.

be/src/common/status.h
```diff
@@ -151,8 +155,18 @@ class Status {
...
+        #endif
+        return Status(TStatusCode::INTERNAL_ERROR, Slice(), precise_code, msg);
+    }
+
     bool ok() const { return _length == 0; }
```
wangbo, line 169: Because ```_length = size + HEADER_LEN;```, 
So ```ok()``` should mean ```_length == HEADER_LEN``` ?

be/src/common/status.h
yiguolei: Does java have this feature? Could you show me some demo code in JAVA?

morningman: Search jvm option `OmitStackTraceInFastThrow`

xinyiZzz: This is an optimization of HotSpot VM called fast throw, I'm not sure if C++ has a similar optimization. Or can we do it ourselves.

Ref: https://stackoverflow.com/questions/42115284/what-are-jvm-pre-allocated-exceptions
        https://www.jianshu.com/p/e87d166380eb

wangbo (approved):
LGTM

be/src/common/status.h
yiguolei: @xinyiZzz Currently, I won't implement this feature in Doris, because it depends on unify the error code work.
 I will do following task after this PR is closed:
1. Replace OLAPStatus with Status
2. Unify ErrorCode, for example TStatusCode, OLAPStatus, PStatus, I will unify them to only one error code
3. Do some statistics on ErrorCodes, for example, we could find how many IO Error happens.
4. After all these 3 tasks are finished, I think maybe we could implement OmitStackTraceInFastThrow. But I am not sure because it will ignore some errors.

xinyiZzz: Great, looking forward.

PR #8910, opened 2022-04-08
[feature](optimizer) : convert out join to inner join

### LINKED ISSUE #7698: [Enhancement] Elimination of outer joins, if the predicates can restrict the result sets so that all null-supplying rows are eliminated
### Search before asking

### Description

Elimination of outer joins, if the predicates can restrict the result sets s
all null-supplying rows are eliminated
- full outer -> inner if both sides have such predicates
- left outer -> inner if the right side has such predicates
- right outer -> inner if the left side has such predicates
- full outer -> left outer if only the left side has such predicates
- full outer -> right outer if only the right side has such predicates

### Solution

Reference implementation of Spark: https://github.com/apache/spark/blob/9c12c3765bda65caf0f7b55b3a0d2f2642e2ac4f/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/joins.scala#L136

### Code of Conduct

jackwener:
# Proposed changes

Issue Number: close #7698

## Problem Summary:

Add optimizer feature: outer join to inner join when there is not null predicate in joined table.

Such as 

```sql
select * 
from table1 left join table2 
on table1.siteid = table2.siteid 
where table2.siteid > 0;

->

select * 
from table1 inner join table2 
on table1.siteid = table2.siteid 
where table2.siteid > 0;
``` 

In this PR, if a table exist a non-null predicate in it, I will convert it join type
- full outer join -> right outer join
- left outer join -> inner join 

to its next table:
- full outer join -> left outer join
- right outer join -> inner join

## Checklist(Required)

1. Does it affect the original behavior: (NO)

## Further comments

in #7698. Just talk about two table join 

- full outer -> inner if both sides have such predicates
- left outer -> inner if the right side has such predicates
- right outer -> inner if the left side has such predicates
- full outer -> left outer if only the left side has such predicates
- full outer -> right outer if only the right side has such predicates

But for multi-table Join, it is worth noticing.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/SelectStmt.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/SelectStmt.java
@@ -63,6 +63,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.function.Consumer;
+import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
 
 /**
@@ -511,7 +513,7 @@ public class SelectStmt extends QueryStmt {
             analyzer.registerConjuncts(whereClause, false, getTableRefIds());
         }
 
-        convertOutJoinToInnerJoin();
+        convertOutJoinToInnerJoin(analyzer);
 
         // Change all outer join tuple to null here after analyze where and from clause
         // all solt desc of join tuple is ready. Before analyze sort info/agg info/analytic info
@@ -617,7 +619,7 @@ public class SelectStmt extends QueryStmt {
         }
     }
 
-    private void convertOutJoinToInnerJoin() {
+    private void convertOutJoinToInnerJoin(Analyzer analyzer) {
         // Use HashSet to ensure tableRefs are unique, such as avoid two table A in it;
         HashSet<String> notNullTables = Sets.newHashSet();
 
@@ -630,6 +632,7 @@ public class SelectStmt extends QueryStmt {
             return;
         }
 
+        // Find tables which is in not-null predicate
         for (Expr expr : flatExprs) {
             if (expr instanceof Predicate && ((Predicate) expr).isNotNullPred()) {
                 for (Expr child : expr.getChildren()) {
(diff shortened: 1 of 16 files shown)


kpfly:
Is there any test cases?

jakevin ***@***.***> 于2022年4月8日周五 10:37写道：

jackwener:
TODO：
Now, this rule apply just predicate all is `AND` like `expr1 And expr2 And expr3`.

The future ticklet is normalize `Expr`, we should conjunct all `OR`, like

```sql
(A and B) or C -> (A or C) and (B or C)
(A and B or C) or (D or E)-> (A and B and E) or (C and D) -> (A or C) and (A or D) and (B or C) and (B or D) and (E or C) and (E or D)
```

For each (`left` `OR` `right`), `left` and `right` must have same `column`.

fe/fe-core/src/main/java/org/apache/doris/analysis/SelectStmt.java
```diff
@@ -613,6 +617,102 @@ private void whereClauseRewrite() {
...
+                if (tableRef != null && tableRef.joinOp.isRightOuterJoin() || tableRef.joinOp.isInnerJoin()
+                        || (tableRef.leftTblRef != null
+                                && (tableRef.leftTblRef.joinOp.isLeftOuterJoin()
+                                        || tableRef.joinOp.isInnerJoin())))
+                    ids.add(id);
+            });
+            ids.forEach(tupleIds::remove);
+        };
+        remove.accept(analyzer.getFullOuterJoinedTupleIds());
+        remove.accept(analyzer.getOuterJoinedTupleIds());
```
morrySnow, line 713: 1. Do we need to remove conjuncts from `org.apache.doris.analysis.Analyzer.GlobalState#fullOuterJoinedConjuncts`?
2. Do we need to remove TupleId from `org.apache.doris.analysis.Analyzer.GlobalState#outerJoinedMaterializedTupleIds`?

@EmmyMiao87 could you help to check out whether it is correct?

PR #8849, opened 2022-04-06
[improvement](account) support to account management sql

zingdle:
# Proposed changes

Issue Number: close #8715

## Problem Summary:

Add `IF EXISTS` support to following statements:
```
CREATE [IF NOT EXISTS] USER
CREATE [IF NOT EXISTS] ROLE
DROP [IF EXISTS] USER
DROP [IF EXISTS] ROLE
```

## Checklist(Required)

1. Does it affect the original behavior: (No)
2. Has unit tests been added: (Yes)
3. Has document been added or modified: (Yes)
4. Does it need to update dependencies: (No)
5. Are there any changes that cannot be rolled back: (No)

## Further comments

Add `ignoreIfExists` param to `XXinternal`.

PR #8911, opened 2022-04-08
[fix](account) use LOG.info instead of LOG.debug

zingdle:
This complements (#8849)

## Problem Summary:

In #8849 I commented that I misused `LOG.debug`. However the PR has been merged before I fix it.


morningman (approved):
LGTM

PR #8937, opened 2022-04-09
[doc] Add 1.0.x version for build image

kkfnui:
apache/incubator-doris:build-env-ldb-toolchain-latest is the right env for 1.0.x.

# Proposed changes

The doc isn't friend for new develop. 
I think it's more readable for new develop by this pr.

## Further comments

It's not the final solution!
Can we build a new image for every great release?


hf200012 (approved):
LGTM

morningman (changes requested):
Actually, we should create a new tag for 1.0.0
I will do this later

Issue #7901, opened 2022-01-27
[Feature] [Vectorized] Some Join opt in vec exec engine


HappenLee:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Description

1. shoule adjust the out join column to nullable
    `select * from t1 left join t2 on t1.a = t2.a` 
    all slots of the t2 shoule be nullable change in join node.

2. Adjust the join column of non nullaware join to non nullable
    `select * from t1 join t2 on t1.a = t2.a`  if a is a nullable column shoule change to not nullable, because of join column `=` will skip the null.

3. Useless columns do not need to be output， here need a projection to reduce the column output to speed up
the join operator.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

EmmyMiao87:
I will resolve this issue together ~

EmmyMiao87:
# Join opt

## Reduce unnecessary memory copy

Gabriel39:
还有个场景我觉得可以做类似优化，就是比如agg+having过滤，having中的列不管之后有没有用到都会带下去，事实上如果之后用不到的话，在agg这里就可以裁剪掉了

EmmyMiao87:
This means that, in fact, the output columns of many operators may be a subset of the input columns. 
So not only the join operator can do this optimization.

Gabriel39:
@EmmyMiao87 Exactly. Btw, just FYI, it should be highlight that which operation benefit from this. In our practice, shuffling profit most from column pruning so we only do projection (including column pruning, some computing operations) in each datastreamsender instead of each node (join, agg, etc.).

EmmyMiao87:
>  shuffling profit most from column pruning so we only do projection (including column pruning, some computing operations) in each datastreamsender instead of each node (join, agg, etc.).

When be is implemented, it is possible to put the project logic in the sender uniformly. 
However, in the case of vectorization, if the project logic is pushed down to the hash join node, the time-consuming of **memory copying can also be reduced during node calculation**. Therefore, the project implementation of hash join is not uniformly placed in the sender.

EmmyMiao87:
# 关于 Outer join 导致的列属性从非 null 变更到 nullable 的问题

[...]

PR #8979, opened 2022-04-12
[Forbidden](Vec) Switch to non-vec engine when outer join + not null column

EmmyMiao87:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

Vectorized code will occur `core` in the case of ```outer join + not null column```, such as issue #7901
So we need to fall back from vectorized mode to non-vectorized mode when we encounter this situation.

If the nullside column of the outer join is a column that must return non-null like count(*)
then there is no way to force the column to be nullable.
At this time, vectorization cannot support this situation,
so it is necessary to fall back to non-vectorization for processing.
For example:
  Query: set enable_vectorized_engine=true
  Query: select * from t1 left join (select k1, count(k2) as count_k2 from t2 group by k1) tmp on t1.k1=tmp.k1
  Result: Query goes non-vectorized engine

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/Analyzer.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/Analyzer.java
@@ -666,10 +666,23 @@ public class Analyzer {
         return globalState.mvExprRewriter;
     }
 
+    /**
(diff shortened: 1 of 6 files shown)


morningman (approved):
LGTM

yiguolei:
#8690

PR #8987, opened 2022-04-12
[community](*): polish config about project

jackwener:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

- Add `.editorconfig`
- Polish `.gitignore`

## Checklist(Required)


fe/checkstyle.xml
```diff
@@ -1,33 +1,420 @@
...
+            <property name="severity" value="error"/>
+        </module>
+
+        <module name="LeftCurly">
+            <!-- Checks for placement of the left curly brace ('{'). -->
+            <property name="severity" value="error"/>
+        </module>
+
+        <module name="RightCurly">
+            <!-- Checks right curlies on CATCH, ELSE, and TRY blocks are on
```
jackwener, line 301: This rule needs attention. 
It will disable `if` without bracket.

fe/checkstyle.xml
```diff
@@ -1,33 +1,420 @@
...
+                     value="Redundant import {0}."/>
+        </module>
+
+        <module name="ImportOrder">
+            <property name="severity" value="error"/>
+            <!-- This ensures that static imports go first. -->
+            <property name="option" value="top"/>
+            <property name="sortStaticImportsAlphabetically" value="true"/>
+            <property name="tokens" value="STATIC_IMPORT, IMPORT"/>
+            <message key="import.ordering"
```
qidaye, line 122: Is it possible to customize import order instead of alphabetical order？

Like below:

```java
import org.apache.doris.*

import com.*

import org.*

import java.*

```

jackwener:
@morningman-cmy @EmmyMiao87 PTAL.

 Now it don't influence current behavior because I remove the change of `checkstyle.xml` in this PR.

EmmyMiao87 (dismissed):
LGTM

.gitignore
```diff
@@ -1,88 +1,77 @@
...
-fe/**/target
+derby.log
 dependency-reduced-pom.xml
-fe_plugins/**/.classpath
+package-lock.json
+yarn.lock
+
+# ignore all sub `/bin` and `/build` directory
+**/build/
+**/bin/
```
yangzhg, line 27: What if I want to modify start.sh in bin?

.gitignore
```diff
@@ -1,88 +1,77 @@
...
-be/.vscode
-be/src/gen_cpp/*.cc
-be/src/gen_cpp/*.cpp
-be/src/gen_cpp/*.h
+# FE
+fe/mocked/
+fe/fe-core/src/main/resources/static/
+
+# BE
+be/*build*/
```
yangzhg, line 72: his modification is too broad. Why don't we sort out the output directory and put them all in one directory?

jackwener:
@yangzhg Thanks review, fix this two point.

yangzhg:
> @yangzhg Thanks review, fixed this two point.

I mean we should put all be generate files to one dir. and ignore that dir, this need to modify build.sh and run_be_ut.sh

jackwener:
> > @yangzhg Thanks review, fixed this two point.
> 
> I mean we should put all be generate files to one dir. and ignore that dir, this need to modify build.sh and run_be_ut.sh

It is `api-break`, I think we should keep same with current situation.

This behavior will influence the developer. 

For example, `clangd` is depend on `compile_commands.json` which is under `build/`. if we change dir, it will make developer env is wrong.

EmmyMiao87 (approved):
LGTM

PR #9023, opened 2022-04-14
[Bug][Vectorized] Fix UB when doing ORDER BY.

amosbird:
priority_queue needs strict weak ordering, or else it's UB.


HappenLee (approved):
LGTM

PR #9050, opened 2022-04-15
[feature](optimizer): spilt implicit `is not null`

### LINKED ISSUE #9042: [Feature] Split the implict `is not null` from `whereclause`
### Search before asking

### Description

*It's original from the comment from #9018, more detail discussion in it*

> For comparison operator, semantics is different in select expr and where expr. Only where expr implicit include is not null.

I'd like to split the predicate with implicit `is not null` from `whereclause`, like:

```sql
select * from table where c1 > 0;
--->
select * from table where c1 > 0 and c1 is not null;
```

After this, I don't need to consider `implicit not null` in `whereclause predicate` 😍

Like #7698 #9016, we just need to consider the explicit `not null`.

### Use case

_No response_

### Related issues

- #9018 
- #7698 
- #9016

### Code of Conduct

jackwener:
# Proposed changes

Issue Number: close #9042

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/SelectStmt.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/SelectStmt.java
@@ -61,6 +61,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.function.BinaryOperator;
 import java.util.stream.Collectors;
 
 /**
@@ -492,7 +493,7 @@ public class SelectStmt extends QueryStmt {
             }
         }
 
-        whereClauseRewrite();
+        rewriteWhereClause();
         if (whereClause != null) {
             if (checkGroupingFn(whereClause)) {
                 throw new AnalysisException("grouping operations are not allowed in WHERE.");
@@ -603,7 +604,7 @@ public class SelectStmt extends QueryStmt {
         return result;
     }
 
-    private void whereClauseRewrite() {
+    private void rewriteWhereClause() {
         if (whereClause instanceof IntLiteral) {
             if (((IntLiteral) whereClause).getLongValue() == 0) {
                 whereClause = new BoolLiteral(false);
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/org/apache/doris/rewrite/RewriteInferNotNullRule.java
```diff
@@ -0,0 +1,115 @@
...
+
+        Predicate predicate = (Predicate) expr;
+        Expr leftExpr = predicate.getChild(0);
+        Expr rightExpr = predicate.getChild(1);
+        if (leftExpr == null || rightExpr == null) {
+            return expr;
+        }
+
+        if (predicate instanceof BinaryPredicate) {
+            /**
```
jackwener, line 59: TODO: Trim `same`

fe/fe-core/src/main/java/org/apache/doris/analysis/SelectStmt.java
```diff
@@ -603,14 +604,92 @@ public List<TupleId> collectTupleIds() {
...
+        whereClause = inferIsNotNull(whereClause);
+    }
+
+    /*-
+     * Infer `IS NOT NULL` from predicate in whereClause.
+     * And spilt it into two parts.
+     * Example:
+     * Where col > 0; -> Where col > 0 and col is not null;
+     */
+    private Expr inferIsNotNull(Expr expr) {
```
EmmyMiao87, line 625: It should be a new Rule ~

jackwener: Note: It isn't a `optimizer rule` 

It is because the following transformation can't terminate in Finite State Machine of rule.

```sql
select * from t1 where c1 > 0;  
-> 
select * from t1 where c1 > 0 and c1 is not null;
->
select * from t1 where c1 > 0 and c1 is not null and c1 is not null;
->
.....
```

So, In fact, this is a `rewrite` in planner build.

wangshuo128: What' the difference between rule and rewrite?

jackwener: This key point is `in planner build.`
Current rule is recursively optimized.

wangshuo128: We should make sure that a `SlotRef` has only one is-not-null predicate, no matter the infer-not-null logical is in a recursively applied rule or single-pass plan rewrite.

Please check the following cases:
1. There is already an is-not-null predicate, e.g., `SELECT a FROM t WHERE a > 1 AND a is not NULL`.
2. More than one where clauses refer to the same `SlotRef` that could infer an is-not-null predicate, e.g., `SELECT * FROM (SELECT * FROM t WHERE a > 1) t WHERE a < 2`.

One possible solution is that we track the relationship between a `SlotRef` and its is-not-null predicate in `Analyzer`.

PR #9106, opened 2022-04-19
[fix](transaction) Fix running transaction num may be zero if there are running transactions when when execute show proc '/transactions' stmt

caiconghui:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
@@ -124,10 +124,10 @@ public class DatabaseTransactionMgr {
 
 
     // count the number of running txns of database, except for the routine load txn
-    private int runningTxnNums = 0;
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
```diff
@@ -124,10 +124,10 @@ public class DatabaseTransactionMgr {
 
 
     // count the number of running txns of database, except for the routine load txn
-    private int runningTxnNums = 0;
+    private volatile int runningTxnNums = 0;
 
     // count only the number of running routine load txns of database
-    private int runningRoutineLoadTxnNums = 0;
+    private volatile int runningRoutineLoadTxnNums = 0;
```
yiguolei, line 130: Do we meet errors using int? If there are any errors, I think you'd better add comment here.

yiguolei (approved):
LGTM

PR #9113, opened 2022-04-19
[format](*): format FE code

jackwener:
# Proposed changes

Issue Number: None

## Problem Summary:

Format code for reviewing.

## Checklist(Required)

## Further comments


fe/fe-core/src/main/java/org/apache/doris/consistency/ConsistencyChecker.java
```diff
@@ -53,23 +52,23 @@
...
 
     private static final Comparator<MetaObject> COMPARATOR =
             (first, second) -> Long.signum(first.getLastCheckTime() - second.getLastCheckTime());
-    
+
     // tabletId -> job
     private Map<Long, CheckConsistencyJob> jobs;
 
-    /*
+    /*-
```
wangshuo128, line 64: Why use `-` in block comments?

jackwener: > Why use `-` in block comments?

Because indentation will be canceled by formatting.

We need `-` keep format

wangshuo128: What tool did you use to format the code style? 
Intellij IDEA code formator has no affects on the indentation of the comments.

EmmyMiao87: I had the same problem at first.
If you use Intellij IDEA , you should add a plugin firstly named 'checkstyle-idea'.
Then, you can import the 'checkstyle.xml' to format in IDE

wangshuo128: Oh. So do we need to add some docs about code style and guide users to use tools to check and format code style? :)
In particular,  some FAQs about the code style, e.g., we want to have indentations in comments, such as wrapped with the `<pre></pre>` HTML tag or with escape characters like `-`.

PR #9121, opened 2022-04-20
[fix](cache) Generate md5 value using utf8 encoding for sqlkey string

### LINKED ISSUE #9120: [Bug] Error cache results occur when sql contains Chinese
### Search before asking

### Version

master

### What's Wrong?

If the cache is turned on(enable_sql_cache=true), if the two sql contents are only different in Chinese, and the Chinese length is the same, the wrong cached result will be queried.

### What You Expected?

Correct cached results.

### How to Reproduce?

```sql

set enable_sql_cache=true;
ADMIN SET FRONTEND CONFIG ('cache_last_version_interval_second' = '0');

create database db1;
use db1;

CREATE TABLE `tbl1` (
    k1 INT DEFAULT '0',
    k2 VARCHAR(32) DEFAULT '',
    k3 BIGINT SUM DEFAULT '0'
)
AGGREGATE KEY(k1, k2)
DISTRIBUTED BY HASH(k1) BUCKETS 10

insert into tbl1 values(1, '北京', 1);

mysql> select * from tbl1 where k2='北京';
+------+--------+------+
| k1   | k2     | k3   |
+------+--------+------+
|    1 | 北京   |    1 |
+------+--------+------+

mysql> select * from tbl1 where k2='上海';
+------+--------+------+
| k1   | k2     | k3   |
+------+--------+------+
|    1 | 北京   |    1 |
+------+--------+------+

```

### Anything Else?

none

### Code of Conduct

zenoyang:
# Proposed changes

Issue Number: close #9120 
- Generate md5 value using utf8 encoding for sqlkey string
- Fix FE UT

## Problem Summary:

The current Chinese character strings with the same length and different contents have the same md5 value after encoding, which will cause different SQLs to generate the same cache key and hit the same cache.

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/cache/CacheProxy.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/cache/CacheProxy.java
@@ -25,6 +25,7 @@ import org.apache.doris.proto.Types;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 
 /**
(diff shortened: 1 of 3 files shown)


caiconghui (approved):
LGTM

PR #9123, opened 2022-04-20
[enhancement](load) optimize load string data and dict page write

yangzhg:
# Proposed changes

1.  reduce condition check when memtable insert
2. reduce virtual function call of BinaryDictPageBuilder
3. remove `_dict_items`  add to `dict_builder` directly

## Problem Summary:

## Checklist(Required)

4. Has unit tests been added: (Yes/No/No Need)
5. Has document been added or modified: (Yes/No/No Need)
6. Does it need to update dependencies: (Yes/No)

## Further comments

### DIFF
--- a/be/src/olap/memtable.cpp
+++ b/be/src/olap/memtable.cpp
@@ -51,18 +51,28 @@ MemTable::MemTable(int64_t tablet_id, Schema* schema, const TabletSchema* tablet
         _skip_list = nullptr;
         _vec_row_comparator = std::make_shared<RowInBlockComparator>(_schema);
         // TODO: Support ZOrderComparator in the future
-        _vec_skip_list = new VecTable(_vec_row_comparator.get(), _table_mem_pool.get(),
-                                      _keys_type == KeysType::DUP_KEYS);
+        _vec_skip_list = std::make_unique<VecTable>(
+                _vec_row_comparator.get(), _table_mem_pool.get(), _keys_type == KeysType::DUP_KEYS);
     } else {
         _vec_skip_list = nullptr;
+        if (_keys_type == KeysType::DUP_KEYS) {
+            _insert_fn = &MemTable::_insert_dup;
+        } else {
+            _insert_fn = &MemTable::_insert_agg;
+        }
+        if (_tablet_schema->has_sequence_col()) {
+            _aggregate_two_row_fn = &MemTable::_aggregate_two_row_with_sequence;
+        } else {
+            _aggregate_two_row_fn = &MemTable::_aggregate_two_row;
+        }
         if (tablet_schema->sort_type() == SortType::ZORDER) {
             _row_comparator = std::make_shared<TupleRowZOrderComparator>(
                     _schema, tablet_schema->sort_col_num());
         } else {
             _row_comparator = std::make_shared<RowCursorComparator>(_schema);
         }
-        _skip_list = new Table(_row_comparator.get(), _table_mem_pool.get(),
-                               _keys_type == KeysType::DUP_KEYS);
+        _skip_list = std::make_unique<Table>(_row_comparator.get(), _table_mem_pool.get(),
+                                             _keys_type == KeysType::DUP_KEYS);
     }
 }
 
@@ -83,24 +93,9 @@ void MemTable::_init_agg_functions(const vectorized::Block* block) {
         DCHECK(function != nullptr);
         _agg_functions[cid] = function;
     }
-    _skip_list = new Table(_row_comparator.get(), _table_mem_pool.get(),
-                           _keys_type == KeysType::DUP_KEYS);
-    if (_keys_type == KeysType::DUP_KEYS) {
-        _insert_fn = &MemTable::_insert_dup;
-    } else {
-        _insert_fn = &MemTable::_insert_agg;
-    }
-    if (_tablet_schema->has_sequence_col()) {
-        _aggregate_two_row_fn = &MemTable::_aggregate_two_row_with_sequence;
-    } else {
-        _aggregate_two_row_fn = &MemTable::_aggregate_two_row;
-    }
 }
 
 MemTable::~MemTable() {
-    delete _skip_list;
-    delete _vec_skip_list;
-
     std::for_each(_row_in_blocks.begin(), _row_in_blocks.end(), std::default_delete<RowInBlock>());
     _mem_tracker->release(_mem_usage);
 }
@@ -248,7 +243,7 @@ void MemTable::_aggregate_two_row_in_block(RowInBlock* new_row, RowInBlock* row_
     }
 }
 vectorized::Block MemTable::_collect_vskiplist_results() {
-    VecTable::Iterator it(_vec_skip_list);
+    VecTable::Iterator it(_vec_skip_list.get());
     vectorized::Block in_block = _input_mutable_block.to_block();
     // TODO: should try to insert data by column, not by row. to opt the the code
     if (_keys_type == KeysType::DUP_KEYS) {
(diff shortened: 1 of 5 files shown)


be/src/olap/memtable.cpp
```diff
@@ -65,44 +75,46 @@ int MemTable::RowCursorComparator::operator()(const char* left, const char* righ
     return compare_row(lhs_row, rhs_row);
 }
 
-void MemTable::insert(const Tuple* tuple) {
+// For non-DUP models, for the data rows passed from the upper layer, when copying the data,
+// we first allocate from _buffer_mem_pool, and then check whether it already exists in
+// _skiplist.  If it exists, we aggregate the new row into the row in skiplist.
+// otherwise, we need to copy it into _table_mem_pool before we can insert it.
+void MemTable::_insert_agg(const Tuple* tuple) {
```
zbtzbtzbt, line 82: we may change a name, such as `_insert_no_dup`? and put  `_insert_dup` before `_insert_no_dup`.
becasue this kind of `insert` agg and unique data.

be/src/olap/rowset/segment_v2/bitshuffle_page.h
```diff
@@ -93,13 +93,52 @@ class BitshufflePageBuilder : public PageBuilder {
...
         _count += to_add;
         _remain_element_capacity -= to_add;
         // return added number through count
         *count = to_add;
+        if constexpr (single) {
+            if constexpr (SIZE_OF_TYPE == 1) {
+                *reinterpret_cast<uint8_t*>(&_data[orig_size]) = *vals;
+                return Status::OK();
+            }
+            if constexpr (SIZE_OF_TYPE == 2) {
```
zbtzbtzbt, line 123: else if constexpr

be/src/olap/rowset/segment_v2/binary_plain_page.h
```diff
@@ -127,11 +125,16 @@ class BinaryPlainPageBuilder : public PageBuilder {
         return Status::OK();
     }
 
-    void update_prepared_size(size_t added_size) {
-        _prepared_size += added_size;
-        _prepared_size += sizeof(uint32_t);
+    inline Slice operator[](std::size_t idx) const {
```
zbtzbtzbt, line 128: size_t

be/src/olap/rowset/segment_v2/bitshuffle_page.h
```diff
@@ -93,13 +93,52 @@ class BitshufflePageBuilder : public PageBuilder {
     bool is_page_full() override { return _remain_element_capacity == 0; }
 
     Status add(const uint8_t* vals, size_t* count) override {
+        return add_internal<false>(vals, count);
+    }
+
+    Status single_add(const uint8_t* vals, size_t* count) {
```
zbtzbtzbt, line 99: can we delete this function？

be/src/olap/rowset/segment_v2/binary_dict_page.cpp
```diff
@@ -90,13 +92,18 @@ Status BinaryDictPageBuilder::add(const uint8_t* vals, size_t* count) {
...
+                    break;
+                }
                 _dictionary.emplace(dict_item, value_code);
-                _dict_items.push_back(dict_item);
-                _dict_builder->update_prepared_size(dict_item.size);
             }
             size_t add_count = 1;
-            RETURN_IF_ERROR(_data_page_builder->add(reinterpret_cast<const uint8_t*>(&value_code),
-                                                    &add_count));
+            RETURN_IF_ERROR(actual_builder->single_add(
```
zbtzbtzbt, line 105: use `add_internal<true>`

be/src/olap/rowset/segment_v2/bitshuffle_page.h
```diff
@@ -93,13 +93,52 @@ class BitshufflePageBuilder : public PageBuilder {
...
+                *reinterpret_cast<uint32_t*>(&_data[orig_size]) =
+                        *reinterpret_cast<const uint32_t*>(vals);
+                return Status::OK();
+            }
+            if constexpr (SIZE_OF_TYPE == 8) {
+                *reinterpret_cast<uint64_t*>(&_data[orig_size]) =
+                        *reinterpret_cast<const uint64_t*>(vals);
+                return Status::OK();
+            }
+            memcpy(&_data[orig_size], vals, to_add_size);
```
zbtzbtzbt, line 138: else

be/src/olap/rowset/segment_v2/bitshuffle_page.h
```diff
@@ -93,13 +93,52 @@ class BitshufflePageBuilder : public PageBuilder {
     bool is_page_full() override { return _remain_element_capacity == 0; }
 
     Status add(const uint8_t* vals, size_t* count) override {
+        return add_internal<false>(vals, count);
```
zbtzbtzbt, line 96: can we delete this function?

zbtzbtzbt (commented):
Thank you for your pr, I have learned a lot

be/src/olap/rowset/segment_v2/binary_plain_page.h
```diff
@@ -47,14 +47,13 @@ namespace segment_v2 {
...
-            : _size_estimate(0), _prepared_size(0), _options(options) {
+            : _size_estimate(0), _options(options) {
         reset();
     }
 
     bool is_page_full() override {
         // data_page_size is 0, do not limit the page size
-        return _options.data_page_size != 0 && (_size_estimate > _options.data_page_size ||
-                                                _prepared_size > _options.data_page_size);
+        return _options.data_page_size != 0 && (_size_estimate > _options.data_page_size);
```
zbtzbtzbt, line 56: better
`return (_options.data_page_size != 0 && _size_estimate > _options.data_page_size);`

yangzhg: why ？

be/src/olap/memtable.cpp
yangzhg: The unique table is essentially an Agg table where the aggregation function on the value column is replace, and the use of unique is only for the convenience of optimization.

be/src/olap/rowset/segment_v2/binary_dict_page.cpp
yangzhg: The reason for adding add internal is that I want to unify the logic of add and single add, and I don't want to add a separate function, so add internal doesn't want to be exposed for use

be/src/olap/rowset/segment_v2/bitshuffle_page.h
yangzhg: No，The reason for adding add internal is that I want to unify the logic of add and single add, and I don't want to add a separate function, so add internal doesn't want to be exposed for use

be/src/olap/rowset/segment_v2/bitshuffle_page.h
yangzhg: else is no need this is like switch case  every if has a return and，and there is no overlap between if statements

be/src/olap/memtable.cpp
zbtzbtzbt: ok

be/src/olap/rowset/segment_v2/binary_plain_page.h
zbtzbtzbt: we might remove the `( )`
`return _options.data_page_size != 0 && _size_estimate > _options.data_page_size;`

or
`return (_options.data_page_size != 0) && (_size_estimate > _options.data_page_size);`

be/src/olap/rowset/segment_v2/bitshuffle_page.h
yangzhg: no add is pure virtual

be/src/olap/rowset/segment_v2/bitshuffle_page.h
```diff
@@ -93,13 +93,48 @@ class BitshufflePageBuilder : public PageBuilder {
...
+    }
+
+    Status single_add(const uint8_t* vals, size_t* count) {
+        return add_internal<true>(vals, count);
+    }
+
+    template <bool single>
+    inline Status add_internal(const uint8_t* vals, size_t* count) {
```
morningman, line 104: Why this implement faster than `_data.append`?

yangzhg: use assign instead of memcpy

morningman (dismissed):
LGTM

morningman (approved):
LGTM

PR #9136, opened 2022-04-20
[fix](spark load) fix `getHashValue` of string type is always zero in…

### LINKED ISSUE #9135: [Bug]  getHashValue of string type is always zero in spark load
### Search before asking

### Version

From 0.13 to the latest.

### What's Wrong?

The bucket id is incorrect when distributed keys contains `string type` in spark load etl.

### What You Expected?

The bucket id is correct when distributed keys contains `string type` in spark load etl.

### How to Reproduce?

First, create a table whose all distributed keys are string type.
```
CREATE TABLE `table_destribute_by_string` (
  `dt` int(11) NULL COMMENT "日期分区字段，格式为datekey(yyyymmdd)",
  `phone_hash` varchar(512) NULL COMMENT "电话",
  `file_name` varchar(512) NULL COMMENT "文件名称",
  `stripe_index` varchar(10) NULL COMMENT "stripe下标",
  `row_index` varchar(10) NULL COMMENT "行下标"
) ENGINE=OLAP
DUPLICATE KEY(`dt`, `phone_hash`)
COMMENT "回归测试"
PARTITION BY RANGE(`dt`)
(PARTITION p20220412 VALUES [("19700101"), ("20220412")),
PARTITION p20220413 VALUES [("20220412"), ("20220413")))
DISTRIBUTED BY HASH(`phone_hash`) BUCKETS 100
PROPERTIES (
"replication_num" = "3",
"in_memory" = "false",
"storage_format" = "V2"
);
```

Then, start spark load of the created table.
Finally, you will find that there is only one task running to write the hdfs file

image

### Anything Else?

_No response_

### Code of Conduct

spaces-X:
# Proposed changes

close #9135 

This bug was introduced from #6087

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/spark-dpp/src/main/java/org/apache/doris/load/loadv2/dpp/DppUtils.java
+++ b/fe/spark-dpp/src/main/java/org/apache/doris/load/loadv2/dpp/DppUtils.java
@@ -182,7 +182,10 @@ public class DppUtils {
             byte value = (byte) (b ? 1 : 0);
             buffer.put(value);
         }
-        buffer.flip();
+        // do not flip buffer when the buffer was created by wrap()
+        if (!type.equals(DataTypes.StringType)) {
+            buffer.flip();
+        }
         return buffer;
     }
 


wangshuo128:
Would you please elaborate how this issue happen？

spaces-X:
> Would you please elaborate how this issue happen？

Buffer flip is used incorrectly.
When the hash key is string type, the hash value is always zero. The reason is that the buffer of string type is obtained by _wrap_, which is not needed to _flip_. If we do so, the buffer limit for read will be zero.

wangbo (approved):
LGTM

PR #9139, opened 2022-04-20
[refactor-wip] Add FileSystem, ReadStream, WriteStream

platoneko:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments

This is the first step to IO stack refactoring, we need to reach a consensus on the interface.


be/src/filesystem/local_file_system.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+namespace fs = std::filesystem;
+
+namespace doris {
+
+LocalFileSystem::LocalFileSystem(std::string root_path) : _root_path(std::move(root_path)) {}
+
+LocalFileSystem::~LocalFileSystem() = default;
+
+Status LocalFileSystem::exists(const std::string& path, bool* res) const {
```
pengxiangyu, line 36: This class is similar to PosixEnv, most functions of it is also defined in PosixEnv, is it necessary to create a new class.

be/src/filesystem/file_system.h
```diff
@@ -0,0 +1,73 @@
...
+    virtual Status is_directory(const std::string& path, bool* res) const = 0;
+
+    // Get all files under the `path` directory.
+    // If it's not a directory, return error
+    virtual Status list(const std::string& path, std::vector<FileStat>* files) = 0;
+
+    // Delete the directory recursively if it exists and not a regular file.
+    // If it's a file, return error
+    // If the directory doesn't exist, return ok
+    virtual Status delete_directory(const std::string& path) = 0;
```
pengxiangyu, line 53: it is different for path with children or not.  Sometimes, all the children need to be deleted, but other times, an error need to be returned.

be/src/filesystem/s3_file_system.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include <aws/s3/model/HeadObjectRequest.h>
+
+#include <filesystem>
+
+#include "util/s3_util.h"
+
+namespace fs = std::filesystem;
+
+namespace doris {
+
```
pengxiangyu, line 31: S3StorageBackend(extends StorageBackend) is used to manager remote file system.
Local file system and remote file system has more different, Each remote file system has similiar operations( such as download,  upload,  local cache), Don't implement them for every remote file system.

pengxiangyu:
Please describe the whole structure for this issue, I can't image how it works. The implement of this patch will cause a big cost for reading and writing remote files. Do you want to create a cache of the remote, or just create a amount of small files？how will you deal with them?

be/src/filesystem/s3_read_stream.cpp
```diff
@@ -0,0 +1,104 @@
...
+        *read_n = 0;
+        return Status::OK();
+    }
+    Aws::S3::Model::GetObjectRequest req;
+    req.SetBucket(_bucket);
+    req.SetKey(_key);
+    req.SetRange(fmt::format("bytes={}-{}", position, position + req_n - 1));
+    req.SetResponseStreamFactory(AwsWriteableStreamFactory(to, req_n));
+
+    auto outcome = _client->GetObject(req);
```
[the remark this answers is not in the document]
platoneko: Our plan is to replace BlockManager with FileSystem class，and unify `env`, `BlockManager` together.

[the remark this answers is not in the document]
platoneko: For more details see https://cwiki.apache.org/confluence/display/DORIS/DSIP-006%3A+Refactor+IO+stack

PR #9145, opened 2022-04-21
[feature-wip] (memory tracker) (step5) Fix track bthread, fix track vectorized query

### LINKED ISSUE #7196: [Feature] Refactored memory statistics framework MemTracker
### Search before asking

### Description

https://shimo.im/docs/DT6JXDRkdTvdyV3G

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

xinyiZzz:
# Proposed changes

Issue Number: close #7196 

## Problem Summary:

1. fix track bthread
- Bthread, a high performance M:N thread library used by brpc. In Doris, a brpc server response runs on one bthread, possibly on multiple pthreads. Currently, MemTracker consumption relies on pthread local variables (TLS).
- This caused pthread TLS MemTracker confusion when switching pthread TLS MemTracker in brpc server response. So replacing pthread TLS with bthread TLS in the brpc server response saves the MemTracker.
Ref: https://github.com/apache/incubator-brpc/blob/731730da85f6af5c25012b4c83ab5bb371320cf8/docs/en/server.md#bthread-local

2. fix track vectorized query
- Added track mmap. Currently, mmap allocates memory in many places of the vectorized execution engine.
- Refactored ThreadContext to avoid dependency conflicts and make it easier to debug.
- Fix some bugs.

## Checklist(Required)

### 1) mem_limit=10M
This means that the memory limit of the BE process is 10M. At this time, the BE process will still start normally, but cannot do anything, and the query will return an error:
This means that after BE is started, the process mem limit has left -156M.
````
ERROR 1105 (HY000): errCode = 2, detailMessage = Memory exceed limit. fragment=89c882f82fa74db1-b8010703e7974fed, details=PartitionedHashTableCtx::ExprValuesCache failed to allocate 53760 bytes, on backend=10.81.85.89. Memory left in process limit=-156100257.00 B . failed alloc=<Memory limit exceeded: ExecNode:Exprs:AGGREGATION_NODE (id=65): TryConsume failed, bytes=53760 process whole consumption=244424704 mem limit=10485760>. current tracker=ExecNode:Exprs:AGGREGATION_NODE (id=65) . If this is a query, can change the limit b
````

### 2) mem_limit=80% (default), set exec_mem_limit=10M
The query returns the error:
At this time, the query mem limit is 10485760B, 10464520B has been used, and the re-allocation of 32768B fails. 
The memory application information reported when switching the MemTracker is `aggregator, while execute get_next`.
The tracker belongs to `ExecNode:VAGGREGATION_NODE`, and the consumption of the tracker in the TCMalloc Hook fails.
````
ERROR 1105 (HY000): errCode = 2, detailMessage = Memory limit exceeded: Memory exceed limit. fragment=, details=In TCMalloc Hook, aggregator, while execute get_next., on backend=10.81.85.89. Memory left in process limit=300.81 GB. failed alloc=<Memory limit exceeded: label=queryId=fdb6664525fe40f9-8bc275d2c86d2546 TryConsume failed size=32768, used=10464520, limit=10485760>. current tracker=ExecNode:VAGGREGATION_NODE (id=24). If this is a query, can change the limit by session variable exec_mem_limit.
````

### 3) mem_limit=99%, set exec_mem_limit=20G
(waiting for the test..., expecting BE will not crash)

## 2. Performance
Similar to #8669  test conclusion on row storage, the new memory statistics framework will also bring about a 2% performance loss on vectorized queries.

For POC performance testing, consider turning off the detailed memory track `memory_verbose_track=false`, which will avoid a 1% performance loss, and further completely turn off the memory track `track_new_delete=false`, which will further avoid a 1% performance loss.

### 1) TEST 1 - SSB 600w
```
Env: 1 FE, 1 BE;
Test Set: ssb LINEORDER 600w;
Default session veriables;
jmeter thread=20;
```
| conf  |  Q1.1(qps, avg time ms) |  Q1.2 | Q1.3  | Q2,1  | Q2.2  | Q3.1  |  Q4.1 |
| ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ |
| track_new_delete=false  |  175.0/s, 110 ms |  214.1/s, 90 ms | 225.7/s, 86 ms  | 72.5/s, 268 ms  | 71.6/s, 272 ms  | 68.0/s, 286 ms  | 45.8/s, 424 ms  |
| track_new_delete=true | 170.1/s, 114 ms  |  211.6/s, 91 ms |  222.9/s, 87 ms | 72.1/s, 269 ms  | 71.3/s, 273 ms  | 67.5/s, 288 ms  |  45.7/s, 427 ms |

### 2) TEST 2 - SSB 60003w
```
Env: 1 FE, 1 BE;
Test Set: ssb LINEORDER 60003w;
Default session veriables;
jmeter thread=1;
```
| conf  |  Q1.1(qps, avg time ms) |  Q1.2 | Q1.3  | Q2,1  | Q2.2  | Q3.1  |  Q4.1 |
| ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ |
| track_new_delete=false  |  2.0/s, 490 ms | 12.7/s, 77 ms  | 13.8/s, 72 ms  | 0.0/s, 28306 ms  | 0.0/s, 20577 ms  | 0.0/s, 41081 ms  |  0.0/s, 48236 ms |
| track_new_delete=true | 1.9/s, 525 ms  |  11.8/s, 84 ms | 13.0/s, 76 ms  | 0.0/s, 28531 ms  | 0.0/s, 20970 ms  | 0.0/s, 42959 ms  | 0.0/s, 48723 ms  |

### 3) TEST 3 - small query
```
Env: 1 FE, 1 BE;
Test Set: ssb LINEORDER 600w;
Default session veriables;
jmeter thread=100;
```
|  conf | select LO_EXTENDEDPRICE from LINEORDER2 where LO_EXTENDEDPRICE = 5273584 limit 1;  |
| ------------ | ------------ |
|  track_new_delete=false | 8275.7/s, 11 ms  |
| track_new_delete=true  |  8203.6/s, 11 ms |

## 3. Observability
```
Env: 1 FE, 1 BE
Test Set: ssb LINEORDER 60003w
track_new_delete=true
memory_verbose_trace=true
set parallel_fragment_exec_instance_num=2;
```

```
Test SQL: SSB Q3.1
         SELECT C_NATION, S_NATION, D_YEAR,
          SUM(LO_REVENUE)  AS  REVENUE
          FROM customer, lineorder, supplier, dates
          WHERE  LO_CUSTKEY = C_CUSTKEY
          AND LO_SUPPKEY = S_SUPPKEY
          AND  LO_ORDERDATE = D_DATEKEY
          AND C_REGION = 'ASIA'
          AND S_REGION = 'ASIA'
          AND D_YEAR >= 1992 AND D_YEAR <= 1997
          GROUP BY C_NATION, S_NATION, D_YEAR
          ORDER BY D_YEAR ASC,  REVENUE DESC;
```

see: BeIP:HttpPort/mem_tracker

```
// The Level use to decide whether to show it in web page,
// each MemTracker have a Level less than or equal to parent, only be set explicit,
// TASK contains query, import, compaction, etc.
enum class MemTrackerLevel { OVERVIEW = 0, TASK, INSTANCE, VERBOSE };
````

1. mem_tracker_level=0 (default OVERVIEW)
> init
image

> query ing
image

> query done
image

2. mem_tracker_level=1 (TASK)
> query ing
image

3. mem_tracker_level=2 (INSTANCE)
> query ing
image

5. mem_tracker_level=3 (VERBOSE)
> query ing
image
image
image
image

> query done
image

> query ing,  set parallel_fragment_exec_instance_num=10;
```
ERROR 1105 (HY000): errCode = 2, detailMessage = Memory limit exceeded: Memory exceed limit. fragment=, details=In TCMalloc Hook, aggregator, while execute get_next., on backend=10.81.85.89. Memory left in process limit=293.08 GB. failed alloc=<Memory limit exceeded: label=queryId=5e72ea753ed5438b-89adf8b2a56f6bf1 TryConsume failed size=4096, used=2147853311, limit=2147483648>. current tracker=ExecNode:VAGGREGATION_NODE (id=7). If this is a query, can change the limit by session variable exec_mem_limit.
```

> query ing,  set parallel_fragment_exec_instance_num=10, set exec_mem_limit= 21474836480;
image

### DIFF
--- a/be/src/exec/exchange_node.cpp
+++ b/be/src/exec/exchange_node.cpp
@@ -80,6 +80,7 @@ Status ExchangeNode::prepare(RuntimeState* state) {
 Status ExchangeNode::open(RuntimeState* state) {
     SCOPED_TIMER(_runtime_profile->total_time_counter());
     SCOPED_SWITCH_TASK_THREAD_LOCAL_MEM_TRACKER(mem_tracker());
+    ADD_THREAD_LOCAL_MEM_TRACKER(_stream_recvr->mem_tracker());
     RETURN_IF_ERROR(ExecNode::open(state));
     if (_is_merging) {
         RETURN_IF_ERROR(_sort_exec_exprs.open(state));
@@ -215,7 +216,6 @@ Status ExchangeNode::get_next_merging(RuntimeState* state, RowBatch* output_batc
     RETURN_IF_CANCELLED(state);
     RETURN_IF_ERROR(state->check_query_state("Exchange, while merging next."));
 
-    ADD_THREAD_LOCAL_MEM_TRACKER(_stream_recvr->mem_tracker());
     RETURN_IF_ERROR(_stream_recvr->get_next(output_batch, eos));
     while ((_num_rows_skipped < _offset)) {
         _num_rows_skipped += output_batch->num_rows();
--- a/be/src/olap/byte_buffer.cpp
+++ b/be/src/olap/byte_buffer.cpp
@@ -20,6 +20,7 @@
 #include <sys/mman.h>
 
 #include "olap/utils.h"
+#include "runtime/thread_context.h"
 
 namespace doris {
 
@@ -42,6 +43,8 @@ void StorageByteBuffer::BufDeleter::operator()(char* p) {
         if (0 != munmap(p, _mmap_length)) {
             LOG(FATAL) << "fail to munmap: mem=" << p << ", len=" << _mmap_length
                        << ", errno=" << Errno::no() << ", errno_str=" << Errno::str();
+        } else {
+            RELEASE_THREAD_LOCAL_MEM_TRACKER(_mmap_length);
         }
     } else {
         delete[] p;
@@ -93,6 +96,7 @@ StorageByteBuffer* StorageByteBuffer::reference_buffer(StorageByteBuffer* refere
 
 StorageByteBuffer* StorageByteBuffer::mmap(void* start, uint64_t length, int prot, int flags,
                                            int fd, uint64_t offset) {
+    CONSUME_THREAD_LOCAL_MEM_TRACKER(length);
     char* memory = (char*)::mmap(start, length, prot, flags, fd, offset);
 
     if (MAP_FAILED == memory) {
@@ -128,6 +132,7 @@ StorageByteBuffer* StorageByteBuffer::mmap(FileHandler* handler, uint64_t offset
 
     size_t length = handler->length();
     int fd = handler->fd();
+    CONSUME_THREAD_LOCAL_MEM_TRACKER(length);
     char* memory = (char*)::mmap(nullptr, length, prot, flags, fd, offset);
 
     if (MAP_FAILED == memory) {
@@ -173,7 +178,7 @@ Status StorageByteBuffer::put(uint64_t index, char src) {
 }
 
 Status StorageByteBuffer::put(const char* src, uint64_t src_size, uint64_t offset,
-                                  uint64_t length) {
+                              uint64_t length) {
     //没有足够的空间可以写
     if (length > remaining()) {
         return Status::OLAPInternalError(OLAP_ERR_BUFFER_OVERFLOW);
--- a/be/src/olap/byte_buffer.cpp
+++ b/be/src/olap/byte_buffer.cpp
@@ -101,6 +101,7 @@ StorageByteBuffer* StorageByteBuffer::mmap(void* start, uint64_t length, int pro
 
     if (MAP_FAILED == memory) {
         OLAP_LOG_WARNING("fail to mmap. [errno='%d' errno_str='%s']", Errno::no(), Errno::str());
+        RELEASE_THREAD_LOCAL_MEM_TRACKER(length);
         return nullptr;
     }
 
@@ -112,6 +113,7 @@ StorageByteBuffer* StorageByteBuffer::mmap(void* start, uint64_t length, int pro
     if (nullptr == buf) {
         deleter(memory);
         OLAP_LOG_WARNING("fail to allocate StorageByteBuffer.");
+        RELEASE_THREAD_LOCAL_MEM_TRACKER(length);
         return nullptr;
     }
 
@@ -137,6 +139,7 @@ StorageByteBuffer* StorageByteBuffer::mmap(FileHandler* handler, uint64_t offset
 
     if (MAP_FAILED == memory) {
         OLAP_LOG_WARNING("fail to mmap. [errno='%d' errno_str='%s']", Errno::no(), Errno::str());
+        RELEASE_THREAD_LOCAL_MEM_TRACKER(length);
         return nullptr;
     }
 
@@ -148,6 +151,7 @@ StorageByteBuffer* StorageByteBuffer::mmap(FileHandler* handler, uint64_t offset
     if (nullptr == buf) {
         deleter(memory);
         OLAP_LOG_WARNING("fail to allocate StorageByteBuffer.");
+        RELEASE_THREAD_LOCAL_MEM_TRACKER(length);
         return nullptr;
     }
 
--- a/be/src/runtime/bufferpool/system_allocator.cc
+++ b/be/src/runtime/bufferpool/system_allocator.cc
@@ -22,6 +22,7 @@
 
 #include "common/config.h"
 #include "gutil/strings/substitute.h"
+#include "runtime/thread_context.h"
 #include "util/bit_util.h"
 #include "util/error_util.h"
 
@@ -75,6 +76,7 @@ Status SystemAllocator::AllocateViaMMap(int64_t len, uint8_t** buffer_mem) {
         // Map an extra huge page so we can fix up the alignment if needed.
         map_len += HUGE_PAGE_SIZE;
     }
+    CONSUME_THREAD_LOCAL_MEM_TRACKER(map_len);
     uint8_t* mem = reinterpret_cast<uint8_t*>(
             mmap(nullptr, map_len, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0));
     if (mem == MAP_FAILED) {
@@ -89,10 +91,12 @@ Status SystemAllocator::AllocateViaMMap(int64_t len, uint8_t** buffer_mem) {
         if (misalignment != 0) {
             uintptr_t fixup = HUGE_PAGE_SIZE - misalignment;
             munmap(mem, fixup);
+            RELEASE_THREAD_LOCAL_MEM_TRACKER(fixup);
             mem += fixup;
             map_len -= fixup;
         }
         munmap(mem + len, map_len - len);
+        RELEASE_THREAD_LOCAL_MEM_TRACKER(map_len - len);
         DCHECK_EQ(reinterpret_cast<uintptr_t>(mem) % HUGE_PAGE_SIZE, 0) << mem;
         // Mark the buffer as a candidate for promotion to huge pages. The Linux Transparent
         // Huge Pages implementation will try to back the memory with a huge page if it is
@@ -142,6 +146,7 @@ Status SystemAllocator::AllocateViaMalloc(int64_t len, uint8_t** buffer_mem) {
 void SystemAllocator::Free(BufferPool::BufferHandle&& buffer) {
     if (config::mmap_buffers) {
         int rc = munmap(buffer.data(), buffer.len());
+        RELEASE_THREAD_LOCAL_MEM_TRACKER(buffer.len());
         DCHECK_EQ(rc, 0) << "Unexpected munmap() error: " << errno;
     } else {
         bool use_huge_pages = buffer.len() % HUGE_PAGE_SIZE == 0 && config::madvise_huge_pages;
--- a/be/src/runtime/bufferpool/system_allocator.cc
+++ b/be/src/runtime/bufferpool/system_allocator.cc
@@ -80,6 +80,7 @@ Status SystemAllocator::AllocateViaMMap(int64_t len, uint8_t** buffer_mem) {
     uint8_t* mem = reinterpret_cast<uint8_t*>(
             mmap(nullptr, map_len, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0));
     if (mem == MAP_FAILED) {
+        RELEASE_THREAD_LOCAL_MEM_TRACKER(map_len);
         return Status::BufferAllocFailed("mmap failed");
     }
 
--- a/be/src/runtime/mem_tracker.cpp
+++ b/be/src/runtime/mem_tracker.cpp
@@ -65,9 +65,7 @@ static std::shared_ptr<MemTracker> brpc_server_tracker;
 static GoogleOnceType brpc_server_tracker_once = GOOGLE_ONCE_INIT;
 
 void MemTracker::create_brpc_server_tracker() {
-    brpc_server_tracker.reset(new MemTracker(-1, "Brpc", get_process_tracker(), MemTrackerLevel::OVERVIEW, nullptr));
-    get_process_tracker()->add_child_tracker(brpc_server_tracker);
-    brpc_server_tracker->init();
+    brpc_server_tracker = MemTracker::create_tracker(-1, "Brpc", get_process_tracker(), MemTrackerLevel::OVERVIEW);
 }
 
 std::shared_ptr<MemTracker> MemTracker::get_brpc_server_tracker() {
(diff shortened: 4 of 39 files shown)

### CODE REVIEW

**Comment by xinyiZzz** [comment]
cc @morningman @yangzhg

File: be/src/runtime/thread_context.h [hunk 1]
```diff
@@ -111,10 +123,15 @@ class ThreadContext {
     }
 
     const std::string& task_id() const { return _task_id; }
-    const std::thread::id& thread_id() const { return _thread_id; }
-    const std::string& thread_id_str() const { return _thread_id_str; }
+    const std::string& thread_id_str() const { return _thread_id; }
     const TUniqueId& fragment_instance_id() const { return _fragment_instance_id; }
 
+    std::string get_thread_id() {
+        std::stringstream ss;
```
morningman, line 130: better not using stringstream. use fmt instead

be/src/runtime/mem_tracker.cpp
```diff
@@ -60,6 +60,21 @@ MemTracker* MemTracker::get_raw_process_tracker() {
     return raw_process_tracker;
 }
 
+// Track memory for all brpc server responses.
+static std::shared_ptr<MemTracker> brpc_server_tracker;
+static GoogleOnceType brpc_server_tracker_once = GOOGLE_ONCE_INIT;
+
+void MemTracker::create_brpc_server_tracker() {
+    brpc_server_tracker.reset(new MemTracker(-1, "Brpc", get_process_tracker(), MemTrackerLevel::OVERVIEW, nullptr));
```
morningman, line 68: I think we should use `MemTracker::create_tracker`? And no need to call `add_child_tracker` and `init` manually.

be/src/runtime/thread_context.h
xinyiZzz: stringstream is here just to convert `std::thread::id` to `std::string`, which fmt doesn't seem to be able to do.

be/src/runtime/thread_mem_tracker_mgr.h
```diff
@@ -175,39 +170,70 @@ class ThreadMemTrackerMgr {
...
+    _untracked_mems.clear();
+    _untracked_mems[0] = 0;
+    _mem_tracker_labels.clear();
+    _mem_tracker_labels[0] = MemTracker::get_process_tracker()->label();
+}
+
+inline void ThreadMemTrackerMgr::init_bthread() {
+    init();
+    _mem_trackers[1] = MemTracker::get_brpc_server_tracker();
+    _untracked_mems[1] = 0;
```
morningman, line 186: Is it safe to use 0 and 1 as the tracker id for process tracker and brpc tracker?

be/src/runtime/thread_mem_tracker_mgr.h
```diff
@@ -127,17 +107,32 @@ class ThreadMemTrackerMgr {
...
-        if (_mem_trackers[_tracker_id]) {
-            return _mem_trackers[_tracker_id];
-        } else {
-            return MemTracker::get_process_tracker();
+    std::shared_ptr<MemTracker> mem_tracker();
+
+    int64_t switch_count = 0;
+
+    std::string print_debug_string() {
+        std::stringstream mem_trackers_str;
```
morningman, line 119: use `fmt`

be/src/runtime/thread_mem_tracker_mgr.h
```diff
@@ -227,21 +253,36 @@ inline void ThreadMemTrackerMgr::cache_consume(int64_t size) {
...
+        _mem_trackers[mem_tracker->id()] = mem_tracker;
+        DCHECK(_mem_trackers[mem_tracker->id()]) << print_debug_string();
+        _untracked_mems[mem_tracker->id()] = 0;
+        _mem_tracker_labels[_temp_tracker_id] = mem_tracker->label();
+    }
+}
+
+inline std::shared_ptr<MemTracker> ThreadMemTrackerMgr::mem_tracker() {
+    DCHECK(_mem_trackers.find(_tracker_id) != _mem_trackers.end()) << print_debug_string();
+    DCHECK(_mem_trackers[_tracker_id]) << print_debug_string();
```
morningman, line 284: These 2 `DCHECK` looks same?

be/src/util/doris_metrics.h
```diff
@@ -128,6 +128,8 @@ class DorisMetrics {
     IntCounter* attach_task_thread_count;
     IntCounter* switch_thread_mem_tracker_count;
     IntCounter* switch_thread_mem_tracker_err_cb_count;
+    // brpc server response count
+    IntCounter* switch_pthread_count;
```
morningman, line 132: Should it `switch_bthread_count`?

be/src/runtime/thread_mem_tracker_mgr.h
```diff
@@ -227,21 +253,36 @@ inline void ThreadMemTrackerMgr::cache_consume(int64_t size) {
...
-        exceeded(_untracked_mem, st);
+        mem_tracker()->consume(size);
+        exceeded(size, st);
     }
-    _untracked_mem = 0;
+}
+
+inline void ThreadMemTrackerMgr::add_tracker(const std::shared_ptr<MemTracker>& mem_tracker) {
+    DCHECK(_mem_trackers.find(mem_tracker->id()) == _mem_trackers.end()) << print_debug_string();
+    if (_mem_trackers.find(mem_tracker->id()) == _mem_trackers.end()) {
```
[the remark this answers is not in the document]
xinyiZzz: OK, I'll remove `find` first.

But this is intentional. Because duplication of `add tracker` in some places may be unavoidable.
(Not currently, future prs will include this, but I'll try to avoid it)

be/src/runtime/thread_mem_tracker_mgr.h
```diff
@@ -227,21 +253,36 @@ inline void ThreadMemTrackerMgr::cache_consume(int64_t size) {
...
+        _mem_trackers[mem_tracker->id()] = mem_tracker;
+        DCHECK(_mem_trackers[mem_tracker->id()]) << print_debug_string();
```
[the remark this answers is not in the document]
xinyiZzz: If there is no `DCHECK(_mem_trackers.find(mem_tracker->id()) == _mem_trackers.end())`, the add tracker may be repeated (currently prohibited, for the reasons above). This will, 1. waste time 2. `_untracked_mems[mem_tracker->id()] = 0` will cause the memory track to be lost.

If there is no `DCHECK(_mem_trackers[mem_tracker->id()])` , it may cause empty trackers to be added, which will cause exceptions to consume trackers later.
The reason why it is placed after `_mem_trackers[mem_tracker->id()] = mem_tracker;` is because the parameter is a reference to shared_ptr to avoid the tracker being destructed after DCHECK.

be/src/runtime/thread_mem_tracker_mgr.h
```diff
@@ -227,21 +253,36 @@ inline void ThreadMemTrackerMgr::cache_consume(int64_t size) {
...
+        _untracked_mems[mem_tracker->id()] = 0;
+        _mem_tracker_labels[_temp_tracker_id] = mem_tracker->label();
```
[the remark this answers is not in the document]
xinyiZzz: A reusable variable, Avoid memory allocation in functions and fall into an infinite loop. Has annotations.

be/src/exec/exchange_node.cpp
```diff
@@ -80,6 +80,7 @@ Status ExchangeNode::prepare(RuntimeState* state) {
 Status ExchangeNode::open(RuntimeState* state) {
     SCOPED_TIMER(_runtime_profile->total_time_counter());
     SCOPED_SWITCH_TASK_THREAD_LOCAL_MEM_TRACKER(mem_tracker());
+    ADD_THREAD_LOCAL_MEM_TRACKER(_stream_recvr->mem_tracker());
```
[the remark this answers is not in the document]
xinyiZzz: As above, avoid adding trackers repeatedly.

We expect a tracker to be added only once per thread. (The DCHECK BUG introduced by the last pr is here...)

be/src/runtime/thread_mem_tracker_mgr.h
```diff
@@ -175,39 +170,70 @@ class ThreadMemTrackerMgr {
...
+    DCHECK(mem_tracker) << print_debug_string();
     _temp_tracker_id = mem_tracker->id();
     if (_temp_tracker_id == _tracker_id) {
         return _tracker_id;
     }
     if (Existed) {
-        DCHECK(_mem_trackers.find(_temp_tracker_id) != _mem_trackers.end());
+        DCHECK(_mem_trackers.find(_temp_tracker_id) != _mem_trackers.end()) << print_debug_string();
     } else {
         if (_mem_trackers.find(_temp_tracker_id) == _mem_trackers.end()) {
```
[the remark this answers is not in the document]
xinyiZzz: The purpose of map find here:
If the tracker has already been added, avoid `_untracked_mems[_temp_tracker_id] = 0;` causing the memory track to be lost.
image

(So in order to avoid this map find, you should add tracker as early as possible, and then call update_tracker<Existed>)

be/src/olap/byte_buffer.cpp
```diff
@@ -93,6 +96,7 @@ StorageByteBuffer* StorageByteBuffer::reference_buffer(StorageByteBuffer* refere
 
 StorageByteBuffer* StorageByteBuffer::mmap(void* start, uint64_t length, int prot, int flags,
                                            int fd, uint64_t offset) {
+    CONSUME_THREAD_LOCAL_MEM_TRACKER(length);
     char* memory = (char*)::mmap(start, length, prot, flags, fd, offset);
 
     if (MAP_FAILED == memory) {
```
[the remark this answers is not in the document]
xinyiZzz: fix

be/src/olap/byte_buffer.cpp
```diff
@@ -128,6 +132,7 @@ StorageByteBuffer* StorageByteBuffer::mmap(FileHandler* handler, uint64_t offset
 
     size_t length = handler->length();
     int fd = handler->fd();
+    CONSUME_THREAD_LOCAL_MEM_TRACKER(length);
     char* memory = (char*)::mmap(nullptr, length, prot, flags, fd, offset);
 
     if (MAP_FAILED == memory) {
```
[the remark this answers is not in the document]
xinyiZzz: fix

be/src/runtime/bufferpool/system_allocator.cc
```diff
@@ -75,6 +76,7 @@ Status SystemAllocator::AllocateViaMMap(int64_t len, uint8_t** buffer_mem) {
         // Map an extra huge page so we can fix up the alignment if needed.
         map_len += HUGE_PAGE_SIZE;
     }
+    CONSUME_THREAD_LOCAL_MEM_TRACKER(map_len);
     uint8_t* mem = reinterpret_cast<uint8_t*>(
             mmap(nullptr, map_len, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0));
     if (mem == MAP_FAILED) {
```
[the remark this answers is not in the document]
xinyiZzz: fix

be/src/runtime/mem_tracker.cpp
xinyiZzz: done, you are right

be/src/runtime/thread_mem_tracker_mgr.h
xinyiZzz: Safe because the map key is the tracker id
id = (GetCurrentTimeMicros() % 1000000) * 100 + _label.length()

(Actually, I'm not 100% sure if the id could be duplicated... but it should be faster than converting to int after hash)

be/src/runtime/thread_mem_tracker_mgr.h
xinyiZzz: fmt doesn't seem to support an indeterminate number of string concatenations.

I changed to `string +=`, which seems to be faster than `stringstream`. `StringBuilder` seems to be a better solution, but not necessary here = =

be/src/runtime/thread_mem_tracker_mgr.h
xinyiZzz: Because these two DCHECK allocations cover two problems.

If there is no first DCHECK, then `_mem_trackers[_tracker_id] == false` has two meanings
1. The key _tracker_id exists in the map, but the value is null;
2. The key _tracker_id does not exist in the map;

be/src/util/doris_metrics.h
xinyiZzz: fix

be/src/runtime/thread_mem_tracker_mgr.h
morningman: You can search `fmt` in `be/src/exec/tablet_sink.cpp` to see example of concating string.

xinyiZzz: cool!
done, I learned.

morningman (approved):
LGTM

PR #9164, opened 2022-04-22
[Bug] Fix bug that new coming routine load tasks are rejected all the time and report TOO_MANY_TASK error

### LINKED ISSUE #9162: [Bug] New coming routine load tasks are always rejected and report TOO_MANY_TASK error
### Search before asking

### Version

latest version

### What's Wrong?

Routine load task cannot be submitted to any backends.
It looks like all backend's task pools are full.

Fe log:
failed to submit task. error code: TOO_MANY_TASKS, msg: 47cbfcef3f7a436b-a591c509911788b0.

Be log:
[routine_load_task_executor.cpp:138] too many tasks in thread pool. reject task: 47cbfcef3f7a436b-a591c509911788b0, job id: 207405, queue size: 0, current tasks num: 10

### What You Expected?

Act normal.

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

xy720:
# Proposed changes

Issue Number: close #9162 

## Problem Summary:

This bug is caused by the data consumer group which is get by the routine load task.
Sometimes, the kafka's begin offsets is not set by Fe, so that 0 data consumer will be assigned in the data consumer group.
Then the routine load task will be stuck in data consumer group's start_all function, and the fragment corresponding to this task will never be closed.
As the result, this routine load task will never be removed from task pool so that the task pool will full soon and reject any new coming routine load task.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/runtime/routine_load/data_consumer_pool.cpp
+++ b/be/src/runtime/routine_load/data_consumer_pool.cpp
@@ -67,12 +67,18 @@ Status DataConsumerPool::get_consumer_grp(StreamLoadContext* ctx,
                 "PAUSE: Currently only support consumer group for Kafka data source");
     }
     DCHECK(ctx->kafka_info);
+    DCHECK(ctx->kafka_info->begin_offset.size() > 0);
 
     std::shared_ptr<KafkaDataConsumerGroup> grp = std::make_shared<KafkaDataConsumerGroup>();
 
     // one data consumer group contains at least one data consumers.
     int max_consumer_num = config::max_consumer_num_per_group;
     size_t consumer_num = std::min((size_t)max_consumer_num, ctx->kafka_info->begin_offset.size());
+    if (consumer_num < 1) {
+        return Status::InternalError(
+                "one data consumer group contains at least one data consumers.");
+    }
+
     for (int i = 0; i < consumer_num; ++i) {
         std::shared_ptr<DataConsumer> consumer;
         RETURN_IF_ERROR(get_consumer(ctx, &consumer));
--- a/be/src/runtime/routine_load/data_consumer_pool.cpp
+++ b/be/src/runtime/routine_load/data_consumer_pool.cpp
@@ -67,7 +67,11 @@ Status DataConsumerPool::get_consumer_grp(StreamLoadContext* ctx,
                 "PAUSE: Currently only support consumer group for Kafka data source");
     }
     DCHECK(ctx->kafka_info);
-    DCHECK(ctx->kafka_info->begin_offset.size() > 0);
+
+    if (ctx->kafka_info->begin_offset.size() == 0) {
+        return Status::InternalError(
+                "PAUSE: The size of begin_offset of task should not be 0.");
+    }
 
     std::shared_ptr<KafkaDataConsumerGroup> grp = std::make_shared<KafkaDataConsumerGroup>();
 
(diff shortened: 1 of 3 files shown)


qidaye (dismissed):
LGTM

be/src/runtime/routine_load/data_consumer_pool.cpp
```diff
@@ -67,12 +67,18 @@ Status DataConsumerPool::get_consumer_grp(StreamLoadContext* ctx,
                 "PAUSE: Currently only support consumer group for Kafka data source");
     }
     DCHECK(ctx->kafka_info);
+    DCHECK(ctx->kafka_info->begin_offset.size() > 0);
```
morningman, line 70: `DCHECK` means it must `> 0`;
So there should not be `consumer_num < 1`.

Why the begin offset is empty? Is it normal?

xy720: In the create statement example below, you can see 
The user didn't specify the custom partitions.
So that 1、Fe will get all kafka partitions from server in routine load's scheduler.
The user set the default offset by datetime.
So that 2、Fe will get kafka offset by time from server in routine load's scheduler.

When 1 is success, meanwhile 2 is failed, the progress of this routine load may not contains any partitions and offsets.
Nevertheless, since newCurrentKafkaPartition which is get by kafka server may be always equal to currentKafkaPartitions, the wrong progress will never be updated.

xy720 (commented):
I have found a example create statement that may cause this problem.

CREATE ROUTINE LOAD iaas.dws_nat ON dws_nat 
WITH APPEND PROPERTIES ( 
"desired_concurrent_number"="2", 
"max_batch_interval" = "20", 
"max_batch_rows" = "400000", 
"max_batch_size" = "314572800", 
"format" = "json", 
"max_error_number" = "0" 
) 
FROM KAFKA ( 
"kafka_broker_list" = "xxxx:xxxx", 
"kafka_topic" = "nat_nsq", 
"property.kafka_default_offsets" = "2022-04-19 13:20:00"
)

xy720 (commented):
During the schedule process, the following error is reported in the step of getting offset for time in KafkaUtil.

483755220206e935b91415336458ca62

be/src/runtime/routine_load/data_consumer_pool.cpp
morningman: I think we should let the task fail directly if `ctx->kafka_info->begin_offset.size() == 0`, so that it can be easier to find error.

xy720: Ok. Now the task will fail in get_consumer_grp method if the size of begin_offset is equal to 0.

morningman (dismissed):
LGTM

morningman (approved):
LGTM

be/src/runtime/routine_load/data_consumer_pool.cpp
```diff
@@ -68,11 +68,17 @@ Status DataConsumerPool::get_consumer_grp(StreamLoadContext* ctx,
     }
     DCHECK(ctx->kafka_info);
 
+    if (ctx->kafka_info->begin_offset.size() == 0) {
+        return Status::InternalError(
```
pengxiangyu, line 72: This is not an error, if a pipe has not data for a long time, it will write a lot of error logs.
Check it before call get_consumer_grp maybe better. Otherwise check return value and don't write warning log each time.

pengxiangyu (changes requested):

PR #9172, opened 2022-04-22
[enhancement](k8s) Support fqdn mode for be  in k8s enviroment

### LINKED ISSUE #9171: [Enhancement] Support fqdn mode for be in k8s enviroment
### Search before asking

### Description

be ip would be changed when reconstruct pod, so doris need to fresh be's ip  when pod is reconstructed 

### Solution

_No response_

### Code of Conduct

caiconghui:
# Proposed changes

Issue Number: close #9171 

## Problem Summary:

In the k8s environment, the ip of the pod  can be changed, but the  hostname of pod is stable. When the host machine of the pod fails, the k8s can schedule the failed pod to the new host machine for reconstruction. After that, the newly created pod's hostname remains unchanged, and the ip address has been changed. The change of the be node's ip address can be detected by FQDNManager when enable_fqdn_mode is true

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/alter/SystemHandler.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/alter/SystemHandler.java
@@ -38,7 +38,6 @@ import org.apache.doris.common.Config;
 import org.apache.doris.common.DdlException;
 import org.apache.doris.common.ErrorCode;
 import org.apache.doris.common.ErrorReport;
-import org.apache.doris.common.Pair;
 import org.apache.doris.common.UserException;
 import org.apache.doris.ha.FrontendNodeType;
 import org.apache.doris.system.Backend;
@@ -48,6 +47,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import org.apache.commons.lang.NotImplementedException;
+import org.apache.commons.lang3.tuple.Triple;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -125,7 +125,7 @@ public class SystemHandler extends AlterHandler {
             if (!Strings.isNullOrEmpty(destClusterName) && Env.getCurrentEnv().getCluster(destClusterName) == null) {
                 throw new DdlException("Cluster: " + destClusterName + " does not exist.");
             }
-            Env.getCurrentSystemInfo().addBackends(addBackendClause.getHostPortPairs(), addBackendClause.isFree(),
+            Env.getCurrentSystemInfo().addBackends(addBackendClause.getIpHostPortTriples(), addBackendClause.isFree(),
                     addBackendClause.getDestCluster(), addBackendClause.getTagMap());
         } else if (alterClause instanceof DropBackendClause) {
             // drop backend
@@ -136,7 +136,7 @@ public class SystemHandler extends AlterHandler {
                         + "All data on this backend will be discarded permanently. "
                         + "If you insist, use DROPP instead of DROP");
             }
-            Env.getCurrentSystemInfo().dropBackends(dropBackendClause.getHostPortPairs());
+            Env.getCurrentSystemInfo().dropBackends(dropBackendClause.getIpHostPortTriples());
         } else if (alterClause instanceof DecommissionBackendClause) {
             // decommission
             DecommissionBackendClause decommissionBackendClause = (DecommissionBackendClause) alterClause;
@@ -197,7 +197,7 @@ public class SystemHandler extends AlterHandler {
 
     private List<Backend> checkDecommission(DecommissionBackendClause decommissionBackendClause)
             throws DdlException {
-        return checkDecommission(decommissionBackendClause.getHostPortPairs());
+        return checkDecommission(decommissionBackendClause.getIpHostPortTriples());
     }
 
     /*
@@ -206,15 +206,18 @@ public class SystemHandler extends AlterHandler {
      * 2. after decommission, the remaining backend num should meet the replication num.
      * 3. after decommission, The remaining space capacity can store data on decommissioned backends.
      */
-    public static List<Backend> checkDecommission(List<Pair<String, Integer>> hostPortPairs)
+    public static List<Backend> checkDecommission(List<Triple<String, String, Integer>> ipHostPortTriples)
             throws DdlException {
         SystemInfoService infoService = Env.getCurrentSystemInfo();
         List<Backend> decommissionBackends = Lists.newArrayList();
         // check if exist
-        for (Pair<String, Integer> pair : hostPortPairs) {
-            Backend backend = infoService.getBackendWithHeartbeatPort(pair.first, pair.second);
+        for (Triple<String, String, Integer> triple : ipHostPortTriples) {
+            Backend backend = infoService.getBackendWithHeartbeatPort(triple.getLeft(), triple.getMiddle(),
+                    triple.getRight());
             if (backend == null) {
-                throw new DdlException("Backend does not exist[" + pair.first + ":" + pair.second + "]");
+                throw new DdlException("Backend does not exist["
+                        + (Config.enable_fqdn_mode ? triple.getMiddle() : triple.getLeft())
+                        + ":" + triple.getRight() + "]");
             }
             if (backend.isDecommissioned()) {
                 // already under decommission, ignore it
@@ -232,22 +235,23 @@ public class SystemHandler extends AlterHandler {
     @Override
     public synchronized void cancel(CancelStmt stmt) throws DdlException {
         CancelAlterSystemStmt cancelAlterSystemStmt = (CancelAlterSystemStmt) stmt;
-        cancelAlterSystemStmt.getHostPortPairs();
-
         SystemInfoService infoService = Env.getCurrentSystemInfo();
         // check if backends is under decommission
         List<Backend> backends = Lists.newArrayList();
-        List<Pair<String, Integer>> hostPortPairs = cancelAlterSystemStmt.getHostPortPairs();
-        for (Pair<String, Integer> pair : hostPortPairs) {
+        List<Triple<String, String, Integer>> ipHostPortTriples = cancelAlterSystemStmt.getIpHostPortTriples();
+        for (Triple<String, String, Integer> triple : ipHostPortTriples) {
             // check if exist
-            Backend backend = infoService.getBackendWithHeartbeatPort(pair.first, pair.second);
+            Backend backend = infoService.getBackendWithHeartbeatPort(triple.getLeft(), triple.getMiddle(),
+                    triple.getRight());
             if (backend == null) {
-                throw new DdlException("Backend does not exists[" + pair.first + "]");
+                throw new DdlException("Backend does not exist["
+                        + (Config.enable_fqdn_mode ? triple.getMiddle() : triple.getLeft())
+                        + ":" + triple.getRight() + "]");
             }
 
             if (!backend.isDecommissioned()) {
                 // it's ok. just log
-                LOG.info("backend is not decommissioned[{}]", pair.first);
+                LOG.info("backend is not decommissioned[{}]", backend.getId());
                 continue;
             }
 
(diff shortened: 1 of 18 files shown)


fe/fe-core/src/main/java/org/apache/doris/alter/SystemHandler.java
```diff
@@ -189,15 +188,15 @@ private List<Backend> checkDecommission(DecommissionBackendClause decommissionBa
...
         SystemInfoService infoService = Catalog.getCurrentSystemInfo();
         List<Backend> decommissionBackends = Lists.newArrayList();
         // check if exist
-        for (Pair<String, Integer> pair : hostPortPairs) {
-            Backend backend = infoService.getBackendWithHeartbeatPort(pair.first, pair.second);
+        for (Triple<String, String, Integer> triple : ipHostPortTriples) {
+            Backend backend = infoService.getBackendWithHeartbeatPort(triple.getLeft(), triple.getMiddle(), triple.getRight());
             if (backend == null) {
-                throw new DdlException("Backend does not exist[" + pair.first + ":" + pair.second + "]");
+                throw new DdlException("Backend does not exist[" + triple.getLeft() + ":" + triple.getLeft() + "]");
```
stalary, line 199: `triple.getLeft()` repeat?

caiconghui: my fault, I will fix it

stalary:
`triple.getMiddle() == null ? triple.getLeft() : triple.getMiddle()`
This code appears many times, can we do it in the generated place?
`return Triple.of(ip, Config.enable_k8s_container_drift_mode ? hostName : ip, heartbeatPort)` whether it can be satisfied?

stalary:
LGTM

fe/fe-core/src/main/java/org/apache/doris/system/FQDNManager.java
```diff
@@ -0,0 +1,103 @@
...
+                    }
+                }
+            }
+        }
+
+        Map<String, List<FsBroker>> brokerMap = Maps.newHashMap(
+            Catalog.getCurrentCatalog().getBrokerMgr().getBrokerListMap());
+        for (Map.Entry<String, List<FsBroker>> entry : brokerMap.entrySet()) {
+            for (FsBroker broker : entry.getValue()) {
+               if (broker.hostName != null) {
```
yiguolei, line 76: using getter not access property directly.

fe/fe-core/src/main/java/org/apache/doris/system/FQDNManager.java
```diff
@@ -0,0 +1,103 @@
...
+                   try {
+                       InetAddress inetAddress = InetAddress.getByName(broker.hostName);
+                       if (!broker.ip.equalsIgnoreCase(inetAddress.getHostAddress())) {
```
yiguolei, line 79: And here.

fe/fe-core/src/main/java/org/apache/doris/catalog/BrokerMgr.java
```diff
@@ -195,19 +195,19 @@ public void addBrokers(String name, Collection<Pair<String, Integer>> addresses)
...
 
             List<FsBroker> addedBrokerAddress = Lists.newArrayList();
-            for (Pair<String, Integer> pair : addresses) {
-                List<FsBroker> addressList = brokerAddrsMap.get(pair.first);
+            for (Triple<String, String, Integer> triple : addresses) {
+                List<FsBroker> addressList = brokerAddrsMap.get(triple.getLeft());
                 for (FsBroker addr : addressList) {
-                    if (addr.port == pair.second) {
-                        throw new DdlException("Broker(" + pair.first + ":" + pair.second
+                    if (addr.port == triple.getRight()) {
```
yiguolei, line 201: if drift mode is set, should also check hostname here

fe/fe-core/src/main/java/org/apache/doris/catalog/BrokerMgr.java
```diff
@@ -273,7 +273,7 @@ public void replayDropBrokers(String name, List<FsBroker> addresses) {
         try {
             ArrayListMultimap<String, FsBroker> brokerAddrsMap = brokersMap.get(name);
             for (FsBroker addr : addresses) {
-                brokerAddrsMap.remove(addr.ip, addr);
+                brokerAddrsMap.remove(addr.hostName == null ? addr.ip : addr.hostName, addr);
```
yiguolei, line 276: we should always use ip:port to delete broker.

fe/fe-core/src/main/java/org/apache/doris/catalog/BrokerMgr.java
```diff
@@ -306,6 +306,22 @@ public void replayDropAllBroker(String name) {
...
     }
 
+    public void replayUpdateBrokerInfo(String name, List<FsBroker> brokers) {
+        lock.lock();
+        try {
+            for (FsBroker oldBroker : brokerListMap.get(name)) {
+                for (FsBroker newBroker : brokers) {
+                    if (oldBroker.hostName != null && oldBroker.hostName.equals(newBroker.hostName)
+                        && oldBroker.port == newBroker.port) {
+                        oldBroker.ip = newBroker.ip;
```
yiguolei, line 316: something is wrong here. brokermap has ip-->broker. If changed ip here, should also change the map.

morningman:
This PR needs some more detailed description:

1. Background motivation for the issue
2. Detail Solutions
3. Which interfaces and behaviors are affected
4. Documentation corresponding to the function
    1. What scenarios does this function apply to
    2. what problems does it solve
    3. whether there are any precautions.
5. Unit test

hello-stephen:
 the sum of best hot time: 35.24 seconds
 load time: 447 seconds
 storage size: 17122714230 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20221126155700_clickbench_pr_53507.html

caiconghui:
1. enable_fqdn_mode is mainly used in k8s environment, not be compatible with old cluster, when set true, the FQDN may not work, because origin's be hostName is null, but would not affect the behavior of the whole cluster.
2. when enable_fqdn_mode is true, the FDQNManager would detect be ip change every some seconds, and would check hostName needed by backend when add backend, you sholud specify "hostName:port" or still give "ip:port" with that cluster can get hostName by ip, otherwise, add add backend would failed.
3. now for be,  if drop or modify backend cluster would check backend's hostname and port is same or then ip and port is same.
4. check interval is 5 seconds,  it means FDQNManager would check every 5 seconds, and other value is also ok, it just affect the be ip detect time
5. if be's ip change, then the FDQNManager would invalidate all client cache to be, and set the new ip for the be.
6. finally all be and fe connection is still based all ip, this logic keep unchanged

fe/fe-core/src/main/java/org/apache/doris/system/SystemInfoService.java
```diff
@@ -901,7 +915,8 @@ public void clear() {
         this.idToReportVersionRef = null;
     }
 
-    public static Pair<String, Integer> validateHostAndPort(String hostPort) throws AnalysisException {
+    public static Triple<String, String, Integer> getIpHostAndPort(String hostPort, boolean strictCheck)
```
morningman, line 918: why not name `strictCheck` to `isFqdnMode`?

fe/fe-core/src/main/java/org/apache/doris/system/FQDNManager.java
```diff
@@ -0,0 +1,73 @@
...
+    @Override
+    protected void runAfterCatalogReady() {
+        for (Backend be : nodeMgr.getIdToBackend().values()) {
+            if (be.getHostName() != null) {
+                try {
+                    InetAddress inetAddress = InetAddress.getByName(be.getHostName());
+                    if (!be.getHost().equalsIgnoreCase(inetAddress.getHostAddress())) {
+                        String ip = be.getHost();
+                        if (!ip.equalsIgnoreCase("unknown")) {
+                            ClientPool.backendPool.clearPool(new TNetworkAddress(ip, be.getBePort()));
```
morningman, line 54: I think we should change BE'ip and write edit log before clearing the client pool.

morningman: And maybe we should extract a method called: `backend.onIPChanged()`.
So that we can support `alter backend modify ip` later.

fe/fe-core/src/main/java/org/apache/doris/system/FQDNManager.java
```diff
@@ -0,0 +1,73 @@
...
+                        be.setHost(inetAddress.getHostAddress());
+                        Env.getCurrentEnv().getEditLog().logBackendStateChange(be);
+                        LOG.warn("ip for {} of be has been changed from {} to {}", be.getHostName(), ip, be.getHost());
+                    }
+                } catch (UnknownHostException e) {
+                    LOG.warn("unknown host name for be, {}", be.getHostName(), e);
+                    if (!be.isAlive() && !be.getHost().equalsIgnoreCase("unknown")) {
+                        String ip = be.getHost();
+                        ClientPool.backendPool.clearPool(new TNetworkAddress(ip, be.getBePort()));
+                        be.setHost("unknown");
```
morningman, line 65: define this `"unknown"` somewhere.

fe/fe-core/src/main/java/org/apache/doris/system/FQDNManager.java
```diff
@@ -0,0 +1,73 @@
...
+                        if (!ip.equalsIgnoreCase("unknown")) {
+                            ClientPool.backendPool.clearPool(new TNetworkAddress(ip, be.getBePort()));
+                        }
+                        be.setHost(inetAddress.getHostAddress());
+                        Env.getCurrentEnv().getEditLog().logBackendStateChange(be);
+                        LOG.warn("ip for {} of be has been changed from {} to {}", be.getHostName(), ip, be.getHost());
+                    }
+                } catch (UnknownHostException e) {
+                    LOG.warn("unknown host name for be, {}", be.getHostName(), e);
+                    if (!be.isAlive() && !be.getHost().equalsIgnoreCase("unknown")) {
```
morningman, line 62: Why judge `!be.isAlive()`?
Add some comment directly in code.

fe/fe-core/src/main/java/org/apache/doris/system/SystemInfoService.java
caiconghui: because when pod is deconstructed，the ip of pod may not be obtained，to prevent that we cannot drop backend

fe/fe-core/src/main/java/org/apache/doris/system/FQDNManager.java
caiconghui: actually if ip of pod changed， it means the pod has been reconstructed already， so I think it is ok to clear the client pool before changing BE'ip and writing edit log.

And pod ip changed is driven by k8s, not by Manual operation，in other word，pod‘s ip is also stable，if the host machine not failed or there is not manual reconstructed pod operation

fe/fe-core/src/main/java/org/apache/doris/system/FQDNManager.java
caiconghui: ok

fe/fe-core/src/main/java/org/apache/doris/system/FQDNManager.java
caiconghui: here is to make ip work if dns has some problem when be is still alive.

fe/fe-core/src/main/java/org/apache/doris/system/SystemInfoService.java
yiguolei, line 918: Is it better to add a class instead of using pair or triple....

fe/fe-core/src/main/java/org/apache/doris/system/Backend.java
```diff
@@ -70,7 +69,9 @@ public enum BackendState {
     @SerializedName("id")
     private long id;
     @SerializedName("host")
-    private String host;
+    private volatile String host;
```
yiguolei, line 72: 我们最好在这里做一下区分，ipaddress 和 host name， 感觉现在这个host 和 hostname 区分不清楚。

yiguolei (approved):
LGTM

PR #9011, opened 2022-04-13
[Enhancement] (image) check image validity as soon as generated

Henry2SS:
load newly generated image file as soon as generated to check if it is valid.

# Proposed changes

Issue Number: close #9010 

## Problem Summary:

Describe the overview of changes.

## Checklist(Required)

1. Does it affect the original behavior: (Yes)
2. Has unit tests been added: (No Need)
3. Has document been added or modified: (No Need)
4. Does it need to update dependencies: (No)
5. Are there any changes that cannot be rolled back: (No)

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #9001, opened 2022-04-13
[bug-fix] fix two coredump of vec-stream-load

zbtzbtzbt:
# Proposed changes
 **fix two coredump of vec-stream-load**

@HappenLee @zhangstar333 @BiteTheDDDDt @spaces-X

### bug-1:

when stream-load `replace` table data, with varchar(20000) `union_id` as value column

```
CREATE TABLE `test_table_stream_load_vec` (
  `k1` int(11) NULL COMMENT "a",
  `k2` varchar(20000) NULL ,
  `k3` varchar(20000) NULL ,
  `union_id` varchar(20000) REPLACE NULL 
) ENGINE=OLAP
AGGREGATE KEY(`k1`, `k2`, `k3`)
...
```
it will coredump, because `Value` has been free, we need use `CopiedValue ` instead.
```
start time: Mon Apr 11 14:31:42 CST 2022
WARNING: Logging before InitGoogleLogging() is written to STDERR
I0411 14:31:42.030472 179842 env.cpp:46] Env init successfully.
*** Aborted at 1649658858 (unix time) try "date -d @1649658858" if you are using GNU date ***
*** SIGSEGV address not mapped to object (@0x7fc5f64a0d82) received by PID 179842 (TID 0x7fc7869f5700) from PID 18446744073546632578; stack trace: ***
0# doris::signal::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*) at /root/zbt/incubator-doris-stream-load-vec/be/src/common/signal_handler.h:420
1# 0x00007FC78CEC4400 in /lib64/[libc.so](http://libc.so/).6
2# memcpy at /root/zbt/incubator-doris-stream-load-vec/be/src/glibc-compatibility/memcpy/memcpy_x86_64.cpp:219
3# doris::vectorized::ColumnString::insert_data(char const*, unsigned long) at /root/zbt/incubator-doris-stream-load-vec/be/src/vec/columns/column_string.h:156
4# doris::vectorized::ColumnNullable::insert_data(char const*, unsigned long) at /root/zbt/incubator-doris-stream-load-vec/be/src/vec/columns/column_nullable.cpp:95
5# doris::MemTable::collect_skiplist_results() at /root/zbt/incubator-doris-stream-load-vec/be/src/olap/memtable.cpp:293
6# doris::MemTable::_vflush() at /root/zbt/incubator-doris-stream-load-vec/be/src/olap/memtable.cpp:332
[...]

PR #9180, opened 2022-04-23
avoiding a corrupt image file when there is image.ckpt with non-zero …

dataroaring:
…size

For now, saveImage writes data to image.ckpt via an append FileOutputStream,
when there is a non-zero size file named image.ckpt, a disaster would happen
due to a corrupt image file. Even worse, fe only keeps the lastest image file
and removes others.

BTW, image file should be synced to disk.

It is dangerous to only keep the latest image file, because an image file is
validated when generating the next image file. Then we keep an non validated
image file but remove validated ones. So I will issue a pr which keeps at least
2 image file.

# Proposed changes
I have not tested the change, so please do not merge now.

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
@@ -1966,8 +1966,13 @@ public class Catalog {
     }
 
     public void saveImage(File curFile, long replayedJournalId) throws IOException {
-        if (!curFile.exists()) {
-            curFile.createNewFile();
+        if (curFile.exists()) {
+            if (!curFile.delete()) {
+                throw new IOException(curFile.getName() + " can not be deleted.");
+            }
+        }
+        if (!curFile.createNewFile()) {
+            throw new IOException(curFile.getName() + " can not be created.");
         }
         MetaWriter.write(curFile, this);
     }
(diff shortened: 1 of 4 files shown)


morningman (dismissed):
LGTM

caiconghui:
> It is dangerous to only keep the latest image file, because an image file is
> validated when generating the next image file. Then we keep an non validated
> image file but remove validated ones.

now the old validated ones will not be removed when new images is invalid, pr #9011 provide this solution, maybe we need to find the situation why image would be broken.

dataroaring:
> > It is dangerous to only keep the latest image file, because an image file is
> > validated when generating the next image file. Then we keep an non validated
> > image file but remove validated ones.
> 
> now the old validated ones will not be removed when new images is invalid, pr #9011 provide this solution, maybe we need to find the situation why image would be broken.

actually, this patch fix a case in which image file is broken.  I have a look at #9001, IMHO, we should keep at least 2 image file, because if we kept the one generated just now, it is hard to make sure the data is flushed to disk, especially, if a power off happened. I will send out the code I wrote yesterday for discussing.

lide-reed (approved):
LGTM

zxealous:
The current PR will cause compilation failure and require dependency on #9217

PR #9181, opened 2022-04-23
assure transaction num in image file is right

dataroaring:
For now, dbTransactionManager::getTransactionNum is only used by
checkpoint to get transaction num to put into a image file. However,
transactions written into a image file do not come from the same
data structure as the num comes. Thus, we should pay much attention to
assure two data structue is consistent on size. Actually, it is
very difficult to do so.

This patch just let getTransactionNum get number from the same data
structure as write method.

The change was introduced by b93e841688.

# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments


morningman (approved):
LGTM

lide-reed:
Could you commit it into dev/1.0.1?

PR #9192, opened 2022-04-24
keep at least one validated image file

dataroaring:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

We should keep at least one validated image file and edit log.  #9011 achieve the goal by running loadImage just after the image file is generated, then we keep an image file generated just now.  It works too. However, we should assure the newly generated file is flushed to disk, actually it is hard due to cache on disk itself especially if a power off happened.

Because I wrote the code yesterday, and noted by @caiconghui today that #9011 did the similar job. So I send out the code for discussing.

## Checklist(Required)

## Further comments


lide-reed (approved):
LGTM, it's really a nice way to keep validated image rather than last image that may be corrupt.

PR #9226, opened 2022-04-26
[Enhancement]Avoid unnecessary calculations in HashTableGrower

### LINKED ISSUE #9224: [Enhancement] Avoid unnecessary calculations in `HashTableGrower` & `HashTableFixedGrower`
### Search before asking

### Description

Doris use bit operation In `HashTableGrower` & `HashTableFixedGrower` to get buffer size like this:
```cpp
size_t buf_size() const { return 1ULL << size_degree; }
```
The operation will be executed every time when `buf_size()` is called, but this is not necessary.

### Solution

To remove unnecessary calculations, we can save the result in a newly added field.

### Code of Conduct

mrhhsg:
# Proposed changes

Issue Number: close #9224 

## Problem Summary:

After using this enhancement, the time to build the hash table is slightly reduced.
I test this with TPC-H Q1 with 10GB and run Q1 4 times separately.
|seq|Without the enhancement|With the enhancement|
|--|---------|----------|
|1|3s480ms|3s449ms|
|2|3s274ms|3s277ms|
|3|3s283ms|3s273ms|
|4|3s331ms|3s281ms|

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/vec/common/hash_table/hash_table.h
+++ b/be/src/vec/common/hash_table/hash_table.h
@@ -245,11 +245,14 @@ struct HashTableGrower {
     /// The state of this structure is enough to get the buffer size of the hash table.
     doris::vectorized::UInt8 size_degree = initial_size_degree;
 
-    /// The size of the hash table in the cells.
-    size_t buf_size() const { return 1ULL << size_degree; }
+    size_t _buf_size = 1ULL << initial_size_degree;
+    size_t _max_fill = 1ULL << (initial_size_degree - 1);
 
-    size_t max_fill() const { return 1ULL << (size_degree - 1); }
-    size_t mask() const { return buf_size() - 1; }
+    /// The size of the hash table in the cells.
+    size_t buf_size() const { return _buf_size; }
+
+    size_t max_fill() const { return _max_fill; }
+    size_t mask() const { return _buf_size - 1; }
 
     /// From the hash value, get the cell number in the hash table.
     size_t place(size_t x) const { return x & mask(); }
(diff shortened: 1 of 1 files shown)


be/src/vec/common/hash_table/hash_table.h
```diff
@@ -245,11 +245,14 @@ struct HashTableGrower {
     /// The state of this structure is enough to get the buffer size of the hash table.
     doris::vectorized::UInt8 size_degree = initial_size_degree;
 
+    size_t _buf_size = 1ULL << initial_size_degree;
+    size_t _max_fill = 1ULL << (initial_size_degree - 1);
+
     /// The size of the hash table in the cells.
-    size_t buf_size() const { return 1ULL << size_degree; }
+    size_t buf_size() const { return _buf_size; }
```
yiguolei, line 252: if the size_degree is modified in increase_size method. The buffer size should changed but in your code, the buffer size is const.

be/src/vec/common/hash_table/hash_table.h
```diff
@@ -245,11 +245,14 @@ struct HashTableGrower {
     /// The state of this structure is enough to get the buffer size of the hash table.
     doris::vectorized::UInt8 size_degree = initial_size_degree;
 
+    size_t _buf_size = 1ULL << initial_size_degree;
+    size_t _max_fill = 1ULL << (initial_size_degree - 1);
```
yiguolei, line 249: I think we'd better not change this. I find that it only improved about 0.02%. But the readability of the code is affected. Maybe @HappenLee  could read this PR.

yangzhg:
The query performance change looks very small, more like normal query time consuming jitter, are you sure this pr has a performance improvement?

mrhhsg:
> The query performance change looks very small, more like normal query time consuming jitter, are you sure this pr has a performance improvement?

hi @yangzhg 
In fact, the impact is very small, but a bit operation will be performed each time when data is inserted into the hash table.
For example, it may take tens of milliseconds to perform 150,000 bit operations.

yiguolei:
I will close this PR since it only have 0.02% improvement.

Issue #8984, opened 2022-04-12
[Discussion] Remove AlphaRowset Related Code


yangzhg:
BetaRowset has been online for a long time, and has been the default format, performance and stability have been fully verified, currently in our code base still retains support for AlphaRowset, while taking into account the support of AlphaRowset for our follow-up Development has added a lot of unnecessary work, so I propose to remove the alphaRowset support and code in doris.

This work needs to be done in two steps.
1. add check and convert before 1.1
   1. add the stmt to detect alpharowset tables in version 1.1 #9037

   2. we also can add a configuration to control whether toautomatically compaction alpha rowset tables to betarowset format
   3. before 1.1 upgrade to 1.2 users need to convert all tables in the cluster to
     betarowset,
2. remove alpha rowset related code in 1.2 version

yiguolei:
It's great. If we could remove alpha rowset we could remove many unused code, it will accelerate our develop speed.

PR #9248, opened 2022-04-27
[Refactor] Forbidden rowset v1

yiguolei:
# Proposed changes

This is part of #8984.
FE will never create rowset v1(or alpha rowset) any more. Fe will ignore the storage format parameters if it is v1. Create table, schema change and rollup job will use storage format v2.

- For create table or rollup, will throw exception if use storage format v1.
- For existing table, if it is v1, it will be force converted to V2.

## Problem Summary:

## Checklist(Required)

## Further comments


fe/fe-core/src/main/java/org/apache/doris/task/CreateReplicaTask.java
```diff
@@ -187,7 +189,12 @@ public void setBaseTablet(long baseTabletId, int baseSchemaHash) {
     }
 
     public void setStorageFormat(TStorageFormat storageFormat) {
-        this.storageFormat = storageFormat;
+    	if (storageFormat == TStorageFormat.V1) {
+    		// Ignore v1 format, the storage format must not be v1
```
yangzhg, line 193: print log or throw an exception

yiguolei: done

yangzhg (dismissed):
LGTM

be/src/olap/storage_engine.cpp
```diff
@@ -916,8 +916,13 @@ void StorageEngine::_parse_default_rowset_type() {
     boost::to_upper(default_rowset_type_config);
     if (default_rowset_type_config == "BETA") {
         _default_rowset_type = BETA_ROWSET;
-    } else {
+    } else if (default_rowset_type_config == "ALPHA") {
         _default_rowset_type = ALPHA_ROWSET;
+        LOG(WARNING) << "default_rowset_type in be.conf should be set to beta, alpha is not "
```
morningman, line 921: I think we should use `CONF_Validator` to check this.
See `CONF_Validator` in be/src/common/config.h

fe/fe-core/src/main/java/org/apache/doris/system/HeartbeatFlags.java
```diff
@@ -28,18 +28,21 @@
...
-        long heartbeatFlags = 0;
-        if ("beta".equalsIgnoreCase(GlobalVariable.defaultRowsetType)) {
-            heartbeatFlags |= HeartbeatServiceConstants.IS_SET_DEFAULT_ROWSET_TO_BETA_BIT;
-        }
+	public long getHeartbeatFlags() {
+		long heartbeatFlags = 0;
+		if ("beta".equalsIgnoreCase(GlobalVariable.defaultRowsetType)) {
+			heartbeatFlags |= HeartbeatServiceConstants.IS_SET_DEFAULT_ROWSET_TO_BETA_BIT;
+		} else {
+			throw new IllegalArgumentException("DEFAULT_ROWSET_TYPE in global "
```
morningman, line 42: We should `ignore` "alpha" instead of throw exception. Otherwise, heartbeat may fail

fe/fe-core/src/main/java/org/apache/doris/task/CreateReplicaTask.java
```diff
@@ -187,7 +189,14 @@ public void setBaseTablet(long baseTabletId, int baseSchemaHash) {
...
+    		// V1 will be removed in the future
+    		LOG.error("StorageFormat.V1 is not supported any more, will use V2 automatically");
```
morningman, line 195: Throw exception instead of print error log.
No one will notice this log.

be/src/olap/storage_engine.cpp
yiguolei: Not use validator because it will throw exception during bootstrap and the process could not start. It's bad for upgrade process.

morningman (approved):
LGTM

PR #9290, opened 2022-04-28
[Enhancement](Vectorized)build hash table with new thread, as non-vec…

hello-stephen:
## info

in HashJoinNode, build hash table with new thread, so that child(0) can open() async. in this way,  hash table from different HashJoinNode in the plan tree can build at same time.

## Checklist(Required)

I made the test on tpch data which scale factor is 10. 1 fe and 3 be separately deployed on 4 cloud servers, 
set runtime_filter_mode=off, 
set enable_vectorized_engine=true,
set exec_mem_limit=50G;
set batch_size=4096;
other session values are default.
sql: `select count(*) from lineitem, orders, part where l_orderkey=o_orderkey and l_partkey = p_partkey and year(l_shipdate)=1995;`
run this query 3 times, 
| run | before | after |
| --| --| -- |
| first time | 2.79 | 2.67 | 
| second time | 2.26 | 2.11 | 
| third time | 2.24 | 2.02 |

### DIFF
--- a/be/src/vec/exec/join/vhash_join_node.cpp
+++ b/be/src/vec/exec/join/vhash_join_node.cpp
@@ -971,10 +971,29 @@ Status HashJoinNode::open(RuntimeState* state) {
         RETURN_IF_ERROR((*_vother_join_conjunct_ptr)->open(state));
     }
 
-    RETURN_IF_ERROR(_hash_table_build(state));
-    RETURN_IF_ERROR(child(0)->open(state));
+    if (_runtime_filter_descs.empty()) {
+        std::promise<Status> thread_status;
+        std::thread(bind(&HashJoinNode::_hash_table_build_thread, this, state, &thread_status))
+            .detach();
 
-    return Status::OK();
+        // Open the probe-side child so that it may perform any initialisation in parallel.
+        // Don't exit even if we see an error, we still need to wait for the build thread
+        // to finish.
+        // ISSUE-1247, check open_status after buildThread execute.
+        // If this return first, build thread will use 'thread_status'
+        // which is already destructor and then coredump.
+        Status open_status = child(0)->open(state);
+        RETURN_IF_ERROR(thread_status.get_future().get());
+        return open_status;
+    } else {
+        RETURN_IF_ERROR(_hash_table_build(state));
+        RETURN_IF_ERROR(child(0)->open(state));
+        return Status::OK();
+    }
+}
+
+void HashJoinNode::_hash_table_build_thread(RuntimeState* state, std::promise<Status>* status) {
+    status->set_value(_hash_table_build(state));
 }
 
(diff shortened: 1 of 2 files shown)


be/src/vec/exec/join/vhash_join_node.cpp
```diff
@@ -956,10 +956,19 @@ Status HashJoinNode::open(RuntimeState* state) {
...
     }
 
-    RETURN_IF_ERROR(_hash_table_build(state));
-    RETURN_IF_ERROR(child(0)->open(state));
+    std::promise<Status> thread_status;
+    std::thread(bind(&HashJoinNode::_hash_table_build_thread, this, state, &thread_status))
+            .detach();
 
-    return Status::OK();
+    Status open_status = child(0)->open(state);
```
HappenLee, line 963: why `RETURN_IF_ERROR(child(0)->open(state));` ?

hello-stephen: because I noticed this in non-vectorized hash_join_node.cpp, 
```
        // Open the probe-side child so that it may perform any initialisation in parallel.
        // Don't exit even if we see an error, we still need to wait for the build thread
        // to finish.
        Status open_status = child(0)->open(state);

        // Blocks until ConstructHashTable has returned, after which
        // the hash table is fully constructed and we can start the probe
        // phase.
        RETURN_IF_ERROR(thread_status.get_future().get());

        // ISSUE-1247, check open_status after buildThread execute.
        // If this return first, build thread will use 'thread_status'
        // which is already destructor and then coredump.
        RETURN_IF_ERROR(open_status);
```

morningman:
2 issues:
1. I think you forget to handle runtime filter in vectorized hash join
2. We should bring some performance test to see if this do improve the performance.

be/src/vec/exec/join/vhash_join_node.cpp
```diff
@@ -956,10 +956,44 @@ Status HashJoinNode::open(RuntimeState* state) {
...
-    RETURN_IF_ERROR(child(0)->open(state));
+    std::promise<Status> thread_status;
+    std::thread(bind(&HashJoinNode::_hash_table_build_thread, this, state, &thread_status))
+            .detach();
+
+    if (_runtime_filter_descs.empty()) {
+        // Open the probe-side child so that it may perform any initialisation in parallel.
+        // Don't exit even if we see an error, we still need to wait for the build thread
+        // to finish.
+        // ISSUE-1247, check open_status after buildThread execute.
```
yiguolei, line 967: std::promise<Status> thread_status;
    std::thread(bind(&HashJoinNode::_hash_table_build_thread, this, state, &thread_status))
            .detach();

yiguolei (approved):
LGTM

PR #9329, opened 2022-04-29
[fix](alter-job) Missing alter job when doing checkpoint image

### LINKED ISSUE #9330: [Bug] alter job is missing in iamge
### Search before asking

### Version

1.0.0

### What's Wrong?

in #8030, we removed lots of code about old meta data in FE, 
But the logic of saveAlterJob was modified incorrectly. Cause the alter job to be lost in the image.
But this may not bring data problems, the analysis is as follows:

### Case 1. Do checkpoint before job is finished.
* [image: alter job(pending)]: The job information is lost, but the data is not changed. Like nothing happened.
* [image: alter job(pending)] + [editlog: alter job(waiting_txn)]: The information in the pending stage is lost, but there is still complete job information in the edit log, and the execution can continue.

### Case 2. Do checkpoint after job is finished.
* [image: alter job(finished)]: Modifications to metadata are already stored in the image, the only thing missing is the job history.

So this error does not result in a data error.
But it remains a potential risk.

### What You Expected?

fix it

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

morningman:
# Proposed changes

Issue Number: close #9330

## Problem Summary:

This bug is introduced from #8030

## Checklist(Required)

2. Has unit tests been added: (Nod)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
@@ -2113,6 +2113,8 @@ public class Catalog {
             alterJobsV2 = this.getMaterializedViewHandler().getAlterJobsV2();
         } else if (type == JobType.SCHEMA_CHANGE) {
             alterJobsV2 = this.getSchemaChangeHandler().getAlterJobsV2();
+        } else {
+            throw new IOException("Invalid alter job type: " + type.name());
         }
 
         // alter jobs == 0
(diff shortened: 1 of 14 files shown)


fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -2113,6 +2109,13 @@ public long saveAlterJob(CountingDataOutputStream dos, long checksum) throws IOE
 
     public long saveAlterJob(CountingDataOutputStream dos, long checksum, JobType type) throws IOException {
         Map<Long, AlterJobV2> alterJobsV2 = Maps.newHashMap();
```
xy720, line 2111:
```suggestion
        Map<Long, AlterJobV2> alterJobsV2;
```

morningman: done

xy720 (approved):
LGTM

PR #9358, opened 2022-05-03
fix #9351 can't load parquet file with column name case sensitive with Doris column

### LINKED ISSUE #9351: [Bug] can't load parquet file with column name case sensitive with Doris column
### Search before asking

### Version

1.0.0-rc03

### What's Wrong?

doris table 
```
CREATE TABLE `record` (
 `id` varchar(50) NOT NULL ,
 `energySerial` varchar(3) NULL 
) ENGINE=OLAP
UNIQUE KEY(`id`)
DISTRIBUTED BY HASH(`id`) BUCKETS 10;
```

```
LOAD LABEL test.record_20220502
( 
 DATA INFILE ("hdfs://172.0.0.9/tmp/part0.parq")
 INTO TABLE record format as parquet ( id,energyserial) 
 SET
 ( 
 id = id, 
energySerial = energyserial
 )
) WITH BROKER 'Broker_Doris' ( "username" = "hadoop" );
```

to show the key difference more clearly in sample above,
```
energySerial.toLowerCase() = energyserial
```

we get an load error msg like 
```
unknown reference column, column=energySerial, reference=energyserial 
```

we can't load that parquet column to doris

### What You Expected?

it could works

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

deardeng:
…h Doris column

# Proposed changes

Issue Number: close #9351 

## Problem Summary:

1. change slotDescByName and exprMap to case insensitive map
2. change the realColName acquisition method

## Checklist(Required)

3. Has unit tests been added: No Need
4. Has document been added or modified: No Need
5. Does it need to update dependencies: No
6. Are there any changes that cannot be rolled back: No

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/load/Load.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/Load.java
@@ -1045,12 +1045,23 @@ public class Load {
             return;
         }
 
+        Set<String> tmpSet = Sets.newHashSet();
+        for (ImportColumnDesc importColumnDesc : copiedColumnExprs) {
+            if (importColumnDesc.getExpr() == null) {
+                tmpSet.add(importColumnDesc.getColumnName());
+            }
+        }
+
         // init slot desc add expr map, also transform hadoop functions
         for (ImportColumnDesc importColumnDesc : copiedColumnExprs) {
             // make column name case match with real column name
             String columnName = importColumnDesc.getColumnName();
-            String realColName = tbl.getColumn(columnName) == null ? columnName
-                    : tbl.getColumn(columnName).getName();
+            String realColName;
+            if (tbl.getColumn(columnName) == null || tmpSet.contains(columnName) ){
+                realColName = columnName;
+            } else {
+                    realColName = tbl.getColumn(columnName).getName();
+            }
             if (importColumnDesc.getExpr() != null) {
                 Expr expr = transformHadoopFunctionExpr(tbl, realColName, importColumnDesc.getExpr());
                 exprsByName.put(realColName, expr);
--- a/fe/fe-core/src/main/java/org/apache/doris/load/Load.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/Load.java
@@ -1045,6 +1045,7 @@ public class Load {
             return;
         }
 
+        // load sql '(a, b)set(c=a, d=b)', tmpSet save {a,b}.
         Set<String> tmpSet = Sets.newHashSet();
         for (ImportColumnDesc importColumnDesc : copiedColumnExprs) {
             if (importColumnDesc.getExpr() == null) {
@@ -1057,10 +1058,10 @@ public class Load {
             // make column name case match with real column name
             String columnName = importColumnDesc.getColumnName();
             String realColName;
-            if (tbl.getColumn(columnName) == null || tmpSet.contains(columnName) ){
+            if (tbl.getColumn(columnName) == null || tmpSet.contains(columnName)) {
                 realColName = columnName;
             } else {
-                    realColName = tbl.getColumn(columnName).getName();
+                realColName = tbl.getColumn(columnName).getName();
             }
             if (importColumnDesc.getExpr() != null) {
                 Expr expr = transformHadoopFunctionExpr(tbl, realColName, importColumnDesc.getExpr());
--- a/fe/fe-core/src/main/java/org/apache/doris/load/Load.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/Load.java
@@ -1045,20 +1045,12 @@ public class Load {
             return;
         }
 
-        // load sql '(a, b)set(c=a, d=b)', tmpSet save {a,b}.
-        Set<String> tmpSet = Sets.newHashSet();
-        for (ImportColumnDesc importColumnDesc : copiedColumnExprs) {
-            if (importColumnDesc.getExpr() == null) {
-                tmpSet.add(importColumnDesc.getColumnName());
-            }
-        }
-
         // init slot desc add expr map, also transform hadoop functions
         for (ImportColumnDesc importColumnDesc : copiedColumnExprs) {
             // make column name case match with real column name
             String columnName = importColumnDesc.getColumnName();
             String realColName;
-            if (tbl.getColumn(columnName) == null || tmpSet.contains(columnName)) {
+            if (tbl.getColumn(columnName) == null || importColumnDesc.getExpr() == null) {
                 realColName = columnName;
             } else {
                 realColName = tbl.getColumn(columnName).getName();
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/org/apache/doris/planner/StreamLoadScanNode.java
```diff
@@ -66,8 +66,8 @@ public class StreamLoadScanNode extends LoadScanNode {
     private TupleDescriptor srcTupleDesc;
     private TBrokerScanRange brokerScanRange;
 
-    private Map<String, SlotDescriptor> slotDescByName = Maps.newHashMap();
-    private Map<String, Expr> exprsByName = Maps.newHashMap();
+    private Map<String, SlotDescriptor> slotDescByName = Maps.newTreeMap(String.CASE_INSENSITIVE_ORDER);
```
morningman, line 69: Better add comment to explain why we need CASE_INSENSITIVE_ORDER

fe/fe-core/src/main/java/org/apache/doris/load/Load.java
```diff
@@ -1045,12 +1045,23 @@ private static void initColumns(Table tbl, List<ImportColumnDesc> columnExprs,
...
         for (ImportColumnDesc importColumnDesc : copiedColumnExprs) {
             // make column name case match with real column name
             String columnName = importColumnDesc.getColumnName();
-            String realColName = tbl.getColumn(columnName) == null ? columnName
-                    : tbl.getColumn(columnName).getName();
+            String realColName;
+            if (tbl.getColumn(columnName) == null || tmpSet.contains(columnName) ){
+                realColName = columnName;
+            } else {
+                    realColName = tbl.getColumn(columnName).getName();
```
morningman, line 1063: `tbl.getColumn(columnName)` may be null here.

fe/fe-core/src/main/java/org/apache/doris/load/Load.java
```diff
@@ -1045,12 +1045,23 @@ private static void initColumns(Table tbl, List<ImportColumnDesc> columnExprs,
             return;
         }
 
+        Set<String> tmpSet = Sets.newHashSet();
```
morningman, line 1048: Better add comment to explain this logic.
Give an example is better.

fe/fe-core/src/main/java/org/apache/doris/load/Load.java
```diff
@@ -1045,12 +1045,23 @@ private static void initColumns(Table tbl, List<ImportColumnDesc> columnExprs,
...
+        }
+
         // init slot desc add expr map, also transform hadoop functions
         for (ImportColumnDesc importColumnDesc : copiedColumnExprs) {
             // make column name case match with real column name
             String columnName = importColumnDesc.getColumnName();
-            String realColName = tbl.getColumn(columnName) == null ? columnName
-                    : tbl.getColumn(columnName).getName();
+            String realColName;
+            if (tbl.getColumn(columnName) == null || tmpSet.contains(columnName) ){
```
xy720, line 1060: It didn't make sense.
Only two case make tbl.getColumn(columnName) != null
1、The columnName match.
2、The columnName match with case insensitive.
And you already change exprMap to case insensitive map, so checking tmpSet.contains(columnName) seems no effect at all.

xy720 (commented):
This bug actually has nothing to do with parquet format.
It is caused by the behavior that the user specifies a expr in the Set-Statement with the same name but different case as one of the column in table.

fe/fe-core/src/main/java/org/apache/doris/load/Load.java
deardeng: such case, 

CREATE TABLE `record` (
 `id` varchar(50) NOT NULL ,
 `SS` varchar(3) NULL 
) ENGINE=OLAP
UNIQUE KEY(`id`)
DISTRIBUTED BY HASH(`id`) BUCKETS 10;

LOAD LABEL test.record
( 
 DATA INFILE ("hdfs://172.0.0.9/tmp/part0.parq")
 INTO TABLE record format as parquet
 (id, ss) 
 SET
 ( 
	id = id, 
	SS = ss
 )
) WITH BROKER 'Broker_Doris' ( "username" = "hadoop" );

copiedColumnExprs has entrys ("id", "ss", "id = id", "SS = ss"), 
exprMap has entrys ("id = id", "SS = ss"), 
tmpSet has keys ("id", "ss")

1. if don't check tmpSet.contains(columnName), realName will be SS. realName(SS) will be sent to BE. BE uses realName(SS) to match parquet file's SS column, and an error will be occur，parquet file's meta has only ss column
2. if use tmpSet.contains(columnName), realName will be ss

fe/fe-core/src/main/java/org/apache/doris/load/Load.java
deardeng: if tbl.getColumn(columnName) eq null, it will into line 1061, not here.

fe/fe-core/src/main/java/org/apache/doris/load/Load.java
deardeng: ok, I add some comments

fe/fe-core/src/main/java/org/apache/doris/planner/StreamLoadScanNode.java
deardeng: For example, the column name 「A」 in the table and the mapping '(a) set (A = a)' in load sql， Slotdescbyname stores「a」, If Slotdescbyname case sensitive, later will use 「a」to get table's 「A」 column info, will throw exception.

fe/fe-core/src/main/java/org/apache/doris/load/Load.java
xy720: How about :
if (tbl.getColumn(columnName) == null || importColumnDesc.getExpr() == null){
    realColName = columnName;
} else {
    realColName = tbl.getColumn(columnName).getName();
}

fe/fe-core/src/main/java/org/apache/doris/planner/StreamLoadScanNode.java
morningman: What I mean is it would be better to add this comment in source file, so that other developer can understand it.

fe/fe-core/src/main/java/org/apache/doris/load/Load.java
morningman: My mistake, but please fix the code format(indent)

deardeng: done

fe/fe-core/src/main/java/org/apache/doris/load/Load.java
deardeng: yes, it works and I change it.

fe/fe-core/src/main/java/org/apache/doris/planner/StreamLoadScanNode.java
deardeng: done

xy720 (approved):
LGTM

PR #9424, opened 2022-05-07
[fix] fix bug that replica can not be repaired duo to DECOMMISSION state

### LINKED ISSUE #9422: [Bug] Failed to load data because replica's state is not rignt
### Search before asking

### Version

1.0.0

### What's Wrong?

Load data with error:
```
Failed to commit txn xxx, Tablet [xxxx] success replica num 0 is less than quorum replica ...
```

And the replica state is:
One replica is in state DECOMMISSION, but version is correct, eg, 21.
Another replica is in state NORMAL, but version is stale, eg, 20, and last failed version is 21.

This is table with `replication_num = 1`. 3 BEs.

### What You Expected?

Replica can be recovered automatically and load can be succeed later.

### How to Reproduce?

The following step may lead to the error:
0. Tablet 10000 with 1 replica 10001 on Backend A, version is 20.
1. Begin transaction 100, which is about to write version 21.
2. begin a balance clone task, to clone from Backend A to Backend B.
3. clone task finished, now there are 2 replica(10001, and 10002) with version 20 on Backend A and B.
4. Tablet 10000 being scheduled again, and set replica 10001's state to DECOMMISSION.
5. Transaction 100 finished, and set replica 10001's version to 21. But 10002 failed to load, so its version remains 20.
6. For now, there are 2 replicas, one is in state DECOMMISSION and version is 21, one is in state NORMAL with version 20.
7. The following load job can not find a normal replica, so load will be failed.

In this situation, we can only restart FE to recover.

### Anything Else?

_No response_

### Code of Conduct

morningman:
# Proposed changes

Issue Number: close #9422

## Problem Summary:

Reset state of replica which state are in DECOMMISSION after finished scheduling.

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/clone/TabletSchedCtx.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/clone/TabletSchedCtx.java
@@ -1187,7 +1187,9 @@ public class TabletSchedCtx implements Comparable<TabletSchedCtx> {
         }
     }
 
-    // call this when releaseTabletCtx()
+    /**
+     * call this when releaseTabletCtx()
+     */
     public void resetReplicaState() {
         if (tablet != null) {
             for (Replica replica : tablet.getReplicas()) {
--- a/fe/fe-core/src/main/java/org/apache/doris/clone/TabletSchedCtx.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/clone/TabletSchedCtx.java
@@ -1163,6 +1163,8 @@ public class TabletSchedCtx implements Comparable<TabletSchedCtx> {
             sb.append(". to backend: ").append(destBackendId);
             sb.append(", dest path hash: ").append(destPathHash);
         }
+        sb.append(", visible version: ").append(visibleVersion);
+        sb.append(", committed version: ").append(committedVersion);
         if (errMsg != null) {
             sb.append(". err: ").append(errMsg);
         }
@@ -1184,4 +1186,24 @@ public class TabletSchedCtx implements Comparable<TabletSchedCtx> {
             }
         }
     }
+
+    // call this when releaseTabletCtx()
+    public void resetReplicaState() {
+        if (tablet != null) {
+            for (Replica replica : tablet.getReplicas()) {
+                // To address issue: https://github.com/apache/incubator-doris/issues/9422
+                // the DECOMMISSION state is set in TabletScheduler and not persist to meta.
+                // So it is reasonable to reset this state if we failed to scheduler this tablet.
+                // That is, if the TabletScheduler cannot process the tablet, then it should reset
+                // any intermediate state it set during the scheduling process.
+                if (replica.getState() == ReplicaState.DECOMMISSION) {
+                    replica.setState(ReplicaState.NORMAL);
+                    replica.setWatermarkTxnId(-1);
+                    LOG.debug("reset replica {} on backend {} of tablet {} state from DECOMMISSION to NORMAL",
+                            replica.getId(), replica.getBackendId(), tabletId);
+                }
+            }
+        }
+    }
+
 }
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/org/apache/doris/clone/TabletSchedCtx.java
```diff
@@ -1184,4 +1186,24 @@ public int compare(Replica r1, Replica r2) {
...
+            for (Replica replica : tablet.getReplicas()) {
+                // To address issue: https://github.com/apache/incubator-doris/issues/9422
+                // the DECOMMISSION state is set in TabletScheduler and not persist to meta.
+                // So it is reasonable to reset this state if we failed to scheduler this tablet.
+                // That is, if the TabletScheduler cannot process the tablet, then it should reset
+                // any intermediate state it set during the scheduling process.
+                if (replica.getState() == ReplicaState.DECOMMISSION) {
+                    replica.setState(ReplicaState.NORMAL);
+                    replica.setWatermarkTxnId(-1);
+                    LOG.debug("reset replica {} on backend {} of tablet {} state from DECOMMISSION to NORMAL",
```
pengxiangyu, line 1202: LOG.warn() is better，resetReplicaState will not be called frequently, so this log will not be too many, but we have to known which tablet is reset, to find out why it is be like this.

fe/fe-core/src/main/java/org/apache/doris/clone/TabletScheduler.java
```diff
@@ -1569,7 +1574,10 @@ public void handleRunningTablets() {
 
         // 2. release ctx
         timeoutTablets.stream().forEach(t -> {
-            releaseTabletCtx(t, TabletSchedCtx.State.CANCELLED);
+            // Set "resetReplicaState" to true because
+            // the timeout task should also be considered as UNRECOVERABLE,
+            // so need to reset replica state.
+            releaseTabletCtx(t, TabletSchedCtx.State.CANCELLED, true);
```
pengxiangyu, line 1580: It is better to add a log here, it is useful for us to find why resetReplicaState is called. This log will be not too many.

pengxiangyu (changes requested):

fe/fe-core/src/main/java/org/apache/doris/clone/TabletSchedCtx.java
morningman: It may print a lot log. And actually, this is not an error. It is a common case.

fe/fe-core/src/main/java/org/apache/doris/clone/TabletScheduler.java
morningman: No need, this is an origin logic, and we don't expect this log in the past.

xy720 (commented):
```
One replica is in state DECOMMISSION, but version is correct, eg, 21.
Another replica is in state NORMAL, but version is stale, eg, 20, and last failed version is 21.
```

In this situation, which status of this tablet will cause a repair task? 
REPLICA_MISSING OR VERSION_INCOMPLETE ?

morningman:
> In this situation, which status of this tablet will cause a repair task?
> REPLICA_MISSING OR VERSION_INCOMPLETE ?

REPLICA_MISSING

yiguolei (approved):
LGTM

wykjLDF:
Hi, I meet a bug w.r.t the 'be decommission' too these days, can you please help me to check whether it has some connections with this bug?
Before decommission, the clusters have 5 be, and I want to do decommission on one. And currently, there are no unhealthy tablets in the databases;
Then after commanding on decommission, there are some unhealthy tablets, but most of them recover in about 20 minutes. However, there are 3 tablets that can not recover, even after about 20 hours; Here I attached some screenshots about the status:
Using command "SHOW PROC '/statistic'';":
image
And check the cluster_balance jobs by command SHOW PROC "'/cluster_balance/history_tablets';":
image
It seems that the TabletChecker keeps trying to relocate the three left tablets, however it still not works.
Can you please help me to see this problem? I really appreciate it!

PR #9433, opened 2022-05-07
[feature-wip](parquet-vec) Support parquet scanner in vectorized engine

yinzhijian:
# Proposed changes

Issue Number: close #xxx

## Optimization Todo list：
1. FE generates the corresponding src slot desc and Expr through the parquet schema
2. BE supports direct conversion of arrow type into dest primitive type of similar type. For example, arrow type is INT32, and dest type is TYPE_BIGINT (int64), INT32=>TYPE_BIGINT. Instead of the current way: INT32=> TYPE_INT => TYPE_BIGINT

## Performance Testing:
load parquet file in vec version almost 1x faster than rowset version.
rows num:300k
test table schema:
CREATE TABLE `parquet` (
  `id` int(11) NOT NULL COMMENT "",
  `email` varchar(26) NOT NULL COMMENT "",
  `c_date32` DATE NOT NULL COMMENT "",
  `c_date64` DATETIME NOT NULL COMMENT "",
  `c_timestamp` DATETIME NOT NULL COMMENT "",
  `c_decimal128` DECIMAL(27, 9) NULL COMMENT "",
  `c_bool` BOOLEAN NULL COMMENT "",
  `c_float` FLOAT NULL COMMENT "",
  `c_double` DOUBLE NULL COMMENT "",
  `c_fixed_size_binary` CHAR(20) NULL COMMENT "",
  `c_binary` VARCHAR(32) NULL COMMENT "",
  `c_uint64` BIGINT NULL COMMENT ""
)
PROPERTIES (
"replication_num" = "1"
);

## Checklist(Required)

## Further comments


fe/fe-core/src/main/java/org/apache/doris/load/Load.java
```diff
@@ -1044,26 +1047,52 @@ private static void initColumns(Table tbl, List<ImportColumnDesc> columnExprs,
...
+                    // If column is not nullable, error will be thrown when filling the dest slot,
+                    // which is not nullable.
+                    slotDesc.setIsNullable(true);
+                } else {
+                    // in vectorized load,
+                    // columns from files like parquet files can be parsed as the type in table schema
+                    slotDesc.setType(tblColumn.getType());
+                    slotDesc.setColumn(new Column(realColName, tblColumn.getType()));
+                    // non-nullable column is allowed in vectorized load with parquet format
+                    slotDesc.setIsNullable(tblColumn.isAllowNull());
```
morningman, line 1093: Even if this column does not have expr, it may still be nullable in parquet file?
So I think this should be set to true, too?

morningman:
And please fix the FE ut.

fe/fe-core/src/main/java/org/apache/doris/load/Load.java
yinzhijian: This is for in future versions, where doris column is not nullable and parquet is nullable,we can filter out unwanted values ​​before executing expressions

fe/fe-core/src/main/java/org/apache/doris/task/StreamLoadTask.java
```diff
@@ -263,6 +263,9 @@ private void setOptionalFromTSLPutRequest(TStreamLoadPutRequest request) throws
             case FILE_STREAM:
```
HappenLee, line 263: case FILE_LOCAL:

be/test/vec/utils/arrow_column_to_doris_column_test.cpp
```diff
@@ -0,0 +1,613 @@
+
```
HappenLee, line 1: no need space

yinzhijian:
> **Code Quality Analysis / Build (pull_request) ** Failing after 3m — Build

done

yinzhijian:
> And please fix the FE ut.

done

be/src/vec/exec/vparquet_scanner.cpp
```diff
@@ -0,0 +1,311 @@
...
+Status VParquetScanner::_init_src_block(Block* block) {
+    size_t batch_pos = 0;
+    for (auto i = 0; i < _num_of_columns_from_file; ++i) {
+        SlotDescriptor* slot_desc = _src_slot_descs[i];
+        if (slot_desc == nullptr) {
+            continue;
+        }
+        auto* array = _batch->column(batch_pos++).get();
+        // let src column be nullable for simplify converting
+        auto is_nullable = true;
```
HappenLee, line 117: nullable will make exec slowly, why do not get nullable from slot_desc?

be/src/vec/exec/vparquet_scanner.cpp
```diff
@@ -0,0 +1,311 @@
...
+        _counter->num_rows_unselected += orig_rows - block->rows();
+    }
+    return Status::OK();
+}
+
+void VParquetScanner::_fill_columns_from_path(Block* block) {
+    const TBrokerRangeDesc& range = _ranges.at(_next_range - 1);
+    if (range.__isset.num_of_columns_from_file) {
+        int start = range.num_of_columns_from_file;
+        int rows = block->rows();
```
HappenLee, line 204: auto or size_t

be/src/vec/exec/vparquet_scanner.cpp
```diff
@@ -0,0 +1,311 @@
...
+        MutableColumnPtr data_column = data_type->create_column();
+        block->insert(
+                ColumnWithTypeAndName(std::move(data_column), data_type, slot_desc->col_name()));
+    }
+    return Status::OK();
+}
+
+Status VParquetScanner::get_next(vectorized::Block* block, bool* eof) {
+    // overall of type converting:
+    // arrow type ==arrow_column_to_doris_column==> primitive type(PT0) ==cast_src_block==>
```
HappenLee, line 133: Here we need more detailed comments on the reading process. Why are there two stages of pT0 and pT1

be/src/vec/exec/vparquet_scanner.cpp
yinzhijian: We don't support non-nullable now, but will support it in the future.

HappenLee (dismissed):
LGTM

fe/fe-core/src/main/java/org/apache/doris/load/Load.java
```diff
@@ -1044,30 +1047,61 @@ private static void initColumns(Table tbl, List<ImportColumnDesc> columnExprs,
         if (!needInitSlotAndAnalyzeExprs) {
             return;
         }
-
+        Set<String> exprArgsColumns = Sets.newTreeSet(String.CASE_INSENSITIVE_ORDER);
```
EmmyMiao87, line 1050: Named "exprSrcSlotName"

fe/fe-core/src/main/java/org/apache/doris/load/Load.java
```diff
@@ -1044,30 +1047,61 @@ private static void initColumns(Table tbl, List<ImportColumnDesc> columnExprs,
...
+            }
+        }
+        Set<String> excludedColumns = Sets.newTreeSet(String.CASE_INSENSITIVE_ORDER);
         // init slot desc add expr map, also transform hadoop functions
         for (ImportColumnDesc importColumnDesc : copiedColumnExprs) {
             // make column name case match with real column name
             String columnName = importColumnDesc.getColumnName();
             String realColName;
-            if (tbl.getColumn(columnName) == null || importColumnDesc.getExpr() == null) {
+            if (tblColumn == null || importColumnDesc.getExpr() == null) {
```
EmmyMiao87, line 1068: Where did you declare this variable?

fe/fe-core/src/main/java/org/apache/doris/load/Load.java
```diff
@@ -1044,30 +1047,61 @@ private static void initColumns(Table tbl, List<ImportColumnDesc> columnExprs,
...
                 Expr expr = transformHadoopFunctionExpr(tbl, realColName, importColumnDesc.getExpr());
                 exprsByName.put(realColName, expr);
             } else {
                 SlotDescriptor slotDesc = analyzer.getDescTbl().addSlotDescriptor(srcTupleDesc);
-                slotDesc.setType(ScalarType.createType(PrimitiveType.VARCHAR));
+                // only support parquet format now
+                if (useVectorizedLoad  && formatType == TFileFormatType.FORMAT_PARQUET
+                        && tblColumn != null) {
+                    // in vectorized load
+                    if (exprArgsColumns.contains(columnName)) {
```
EmmyMiao87, line 1082: How about following situation ?
table schema: a, b
src schema : a
src expr: a, b= a+1

yinzhijian: A is varchar type. Suppose A is datetime in the schema , INT in the expression, so it is not sure whether to use the type of the schema or the inferred type of the expression.

B follows the exprsByName logic, assuming that B is int in the schema, then cast((a+1) as int)

EmmyMiao87 (dismissed):
FE is ok

HappenLee (approved):
LGTM

PR #9436, opened 2022-05-07
[Spark Load]fix min_value of GlobalDict will be negative number in spark load

spaces-X:
# Proposed changes

The `row_number()` function in **spark** returns **an integer type value**.

It will cause two problems in **Spark Load** :

**case 1:  loading a large amount of data at one time causes `row_number()` overflow.** 

When the cardinality of the columns to be encoded in the **data imported at one time** is more than **2.1 billion**, `row_number` will return a negative number.

**case 2:  loading data by many times causes the maximum dict_value  in the global dictionary to exceed Integer, but we do not cast it to bigint.**

---

For case 1, I think it's a design flaw that causes a bottleneck on one-time loading and **case 1** has relatively few scenarios, which can be solved by importing in multiple batches in the short term.

For case 2,  it will be solved by this pr.

## Problem Summary:

## Checklist(Required)

## Further comments


spaces-X:
@wangshuo128  Is there any window function like **row_num** return-type is **Long** in spark?

wangshuo128:
> @wangshuo128 Is there any window function like **row_num** return-type is **Long** in spark?

The data type of Spark builtin function `row_number` is `Integer`, please see 
https://github.com/apache/spark/blob/2349f74866ae1b365b5e4e0ec8a58c4f7f06885c/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/windowExpressions.scala#L564

If you want to have a long type `row_number` function, you could implement a UDAF yourself.

wangbo (approved):
LGTM

PR #9488, opened 2022-05-10
fix test compile error in vtablet_sink_test.cpp when use clang

pingchunzhang:
fix test compile error in vtablet_sink_test.cpp when use clang


dataroaring (approved):
LGTM

PR #9492, opened 2022-05-11
[fix](resource-tag) Consider resource tags when assigning tasks for broker & routine load

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

This CL mainly changes:
1. Broker Load
    When assigning backends, use user level resource tag to find available backends.
    If user level resource tag is not set, broker load task can be assigned to any BE node,
    otherwise, task can only be assigned to BE node which match the user level tags.

2. Routine Load
    The current routine load job does not have user info, so it can not get user level tag when assigning tasks.
    So there are 2 ways:
    1. For old routine load job, use tags of replica allocation info to select BE nodes.
    2. For new routine load job, the user info will be added and persisted in routine load job.

## Checklist(Required)

4. Has unit tests been added: (No)
5. Has document been added or modified: (Yes)

## Further comments


fe/fe-core/src/main/java/org/apache/doris/planner/BrokerScanNode.java
```diff
@@ -395,10 +405,21 @@ protected void getFileStatus() throws UserException {
...
 
     private void assignBackends() throws UserException {
+        Set<Tag> tags = Sets.newHashSet();
+        if (userIdentity != null) {
+            tags = Catalog.getCurrentCatalog().getAuth().getResourceTags(userIdentity.getQualifiedUser());
+            if (tags == UserProperty.INVALID_RESOURCE_TAGS) {
+                throw new UserException("No valid resource tag for user: " + userIdentity.getQualifiedUser());
+            }
+        } else {
+            LOG.debug("user info in BrokerScanNode should not be null, add log to observer");
```
xy720, line 415: I think user info could be allowed null here in a BrokerScanNode used by query.

fe/fe-core/src/main/java/org/apache/doris/planner/BrokerScanNode.java
```diff
@@ -135,18 +144,26 @@ private static class ParamCreateContext {
 
     private List<ParamCreateContext> paramCreateContexts;
 
+    // For broker load and external broker table
     public BrokerScanNode(PlanNodeId id, TupleDescriptor destTupleDesc, String planNodeName,
                           List<List<TBrokerFileStatus>> fileStatusesList, int filesAdded) {
         super(id, destTupleDesc, planNodeName, NodeType.BROKER_SCAN_NODE);
         this.fileStatusesList = fileStatusesList;
         this.filesAdded = filesAdded;
+        if (ConnectContext.get() != null) {
```
xy720, line 153: May be we don't need to get user info for query here.

fe/fe-core/src/main/java/org/apache/doris/planner/BrokerScanNode.java
morningman: Why? I think the user info must be set for both query and load. If not, may be we missed somewhere to set it.

fe/fe-core/src/main/java/org/apache/doris/planner/BrokerScanNode.java
morningman: I think I known what you mean. Because for query, the actually tags we use is based on replica tag, not user level tag.
So may be we don't need user info for query.
But here I insist to add this assignment this, for unification purpose.

fe/fe-core/src/main/java/org/apache/doris/planner/BrokerScanNode.java
morningman: I see. I just want to see where do I miss to set the user info, so I add this debug log here.

yangzhg (approved):
LGTM

PR #9500, opened 2022-05-11
[fix](storage) fix core for string predicate in storage layer

wangbo:
# Proposed changes

Currently, for string read from plain page, we hold ptr from page to avoid memcpy.
But if page is released, the ptr from page is invalid, core dump happens.
So for string from plain page, memcpy is neccessary.

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/vec/columns/predicate_column.h
+++ b/be/src/vec/columns/predicate_column.h
@@ -238,7 +238,7 @@ public:
             for (size_t i = 0; i < num; i++) {
                 total_mem_size += len_array[i];
             }
-            
+
             char* destination = (char*)_pool->allocate(total_mem_size);
             for (size_t i = 0; i < num; i++) {
                 uint32_t len = len_array[i];
@@ -253,8 +253,8 @@ public:
 
     void insert_default() override { data.push_back(T()); }
 
-    void clear() override { 
-        data.clear(); 
+    void clear() override {
+        data.clear();
         if (_pool != nullptr) {
             _pool->clear();
         }
--- a/be/src/vec/columns/predicate_column.h
+++ b/be/src/vec/columns/predicate_column.h
@@ -230,17 +230,36 @@ public:
     void insert_many_binary_data(char* data_array, uint32_t* len_array,
                                  uint32_t* start_offset_array, size_t num) override {
         if constexpr (std::is_same_v<T, StringValue>) {
+            if (!_is_mem_pool_inited) {
+                _pool.reset(new MemPool("PredicateStringColumn"));
+                _is_mem_pool_inited = true;
+            }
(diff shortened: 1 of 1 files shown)


be/src/vec/columns/predicate_column.h
```diff
@@ -410,6 +429,9 @@ class PredicateColumnType final : public COWHelper<IColumn, PredicateColumnType<
 
 private:
     Container data;
+    // manages the memory for slice's data(For string type)
+    std::unique_ptr<MemPool> _pool;
+    bool _is_mem_pool_inited = false;;
```
yiguolei, line 434: just check _pool == nullptr?

be/src/vec/columns/predicate_column.h
```diff
@@ -230,17 +230,36 @@ class PredicateColumnType final : public COWHelper<IColumn, PredicateColumnType<
...
             }
         }
     }
 
     void insert_default() override { data.push_back(T()); }
 
-    void clear() override { data.clear(); }
+    void clear() override { 
+        data.clear(); 
+        if (_is_mem_pool_inited) {
```
yiguolei, line 259: Will mempool release the memory when _pool is desconstructed?

be/src/vec/columns/predicate_column.h
```diff
@@ -230,17 +230,36 @@ class PredicateColumnType final : public COWHelper<IColumn, PredicateColumnType<
...
+            for (size_t i = 0; i < num; i++) {
+                total_mem_size += len_array[i];
+            }
+            
+            char* destination = (char*)_pool->allocate(total_mem_size);
             for (size_t i = 0; i < num; i++) {
                 uint32_t len = len_array[i];
                 uint32_t start_offset = start_offset_array[i];
-                insert_string_value(data_array + start_offset, len);
+                memcpy(destination, data_array + start_offset, len);
```
jacktengg, line 247: Not sure how much performnce will be impacted by allocate and memcpy here.

be/src/vec/columns/predicate_column.h
wangbo: This is just for memory reuse.
```
 mem_pool.h
   /// Makes all allocated chunks available for re-use, but doesn't delete any chunks.
    void clear();
```

be/src/vec/columns/predicate_column.h
wangbo: 👌🏻

morningman:
It seems that we are likely to refer to the memory in the page elsewhere, and the same problem exists?

yiguolei (approved):
LGTM

PR #9527, opened 2022-05-12
doris support query hive bitmap type

lihuigang:
# Proposed changes

## Problem Summary:

Doris support query hive bitmap type

The binary(bitmap) in hive is generated by the bitmap udf function in fe/hive-udf, the bitmap in hive is exactly the same as the bitmap in doris

## Checklist(Required)

## Further comments


fe/fe-core/src/main/java/org/apache/doris/load/Load.java
```diff
@@ -1064,6 +1064,12 @@ private static void initColumns(Table tbl, List<ImportColumnDesc> columnExprs,
                 // which is not nullable.
                 slotDesc.setIsNullable(true);
                 slotDesc.setColumn(new Column(realColName, PrimitiveType.VARCHAR));
+
+                boolean isBitmapType = tbl.getColumn(columnName).getType().isBitmapType();
+                if(isBitmapType){
```
cambyzju, line 1069: I think it is better to change the code logic like: if (isBitmapType) { xxx } else { xxx }.
Then we could sure functions like, setIsMaterialized and setIsNullable and so on, are called as expected.

new-docs/en/ecosystem/external-table/hive-of-doris.md
```diff
@@ -104,16 +105,20 @@ The supported Hive column types correspond to Doris in the following table.
 |   DECIMAL  | DECIMAL |  |
 |   DATE   |  DATE  |                                   |
 |   TIMESTAMP  | DATETIME | Timestamp to Datetime will lose precision |
+|  BINARY   |  BITMAP  |  |
```
cambyzju, line 108: need another white-space before BINARY? It seams there are 4 white-space before 'TIMESTAMP', but 3 white-space before 'BINARY'.

fe/fe-core/src/main/java/org/apache/doris/load/Load.java
lihuigang: Yes, thank you. I have changed the code logic

new-docs/en/ecosystem/external-table/hive-of-doris.md
lihuigang: Thank you. I've added a white-space

fe/fe-core/src/main/java/org/apache/doris/analysis/ColumnDef.java
```diff
@@ -172,7 +172,7 @@ public void analyze(boolean isOlap) throws AnalysisException {
         }
 
         // disable Bitmap Hll type in keys, values without aggregate function.
-        if (type.isBitmapType() || type.isHllType()) {
+        if (isOlap && type.isBitmapType() || type.isHllType()) {
```
BiteTheDDDDt, line 175: Does this affect other types of external tables?

lihuigang: I don't think it will affect. In the olap type table, the bitmap and hll fields need to be restricted to only be used as aggregate types, not as keys. All external tables are non-olap type tables, and non-olap type tables do not need to be restricted

PR #9533, opened 2022-05-12
[refactor] Refactoring Status static methods to format message using …

BePPPower:
…fmt::format

# Proposed changes

Issue Number: close #xxx

## Problem Summary:

```
static Status InvalidArgument(const Slice& msg, int16_t precise_code = 1,
                              const Slice& msg2 = Slice()) {
    return Status(TStatusCode::INVALID_ARGUMENT, msg, precise_code, msg2);
}

Status::InvalidArgument(
    strings::Substitute("task(signature=$0, type=$1) has wrong task type", signature, task_type)
);
```
Probems:
1. Almost all of 'Status' static functions have three arguments, but in practice they were used only the first argument, and the second and third arguments use the default values
2. Strings :Substitute is basically used to construct string in the assignment of the first parameter 'msg' (even string splicing method is used in some places), which is tedious and inconsistent to use

## Checklist(Required)

3. Has unit tests been added: (No Need)

## Further comments

### DIFF
--- a/be/src/agent/agent_server.cpp
+++ b/be/src/agent/agent_server.cpp
@@ -121,14 +121,14 @@ void AgentServer::submit_tasks(TAgentResult& agent_result,
         TTaskType::type task_type = task.task_type;
         int64_t signature = task.signature;
 
-#define HANDLE_TYPE(t_task_type, work_pool, req_member)                                     \
-    case t_task_type:                                                                       \
-        if (task.__isset.req_member) {                                                      \
-            work_pool->submit_task(task);                                                   \
-        } else {                                                                            \
-            ret_st = Status::InvalidArgument("task(signature={}) has wrong request member", \
-                                             signature);                                    \
-        }                                                                                   \
+#define HANDLE_TYPE(t_task_type, work_pool, req_member)                                          \
+    case t_task_type:                                                                            \
+        if (task.__isset.req_member) {                                                           \
+            work_pool->submit_task(task);                                                        \
+        } else {                                                                                 \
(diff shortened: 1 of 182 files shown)


be/src/agent/agent_server.cpp
```diff
@@ -120,14 +120,14 @@ void AgentServer::submit_tasks(TAgentResult& agent_result,
...
-        } else {                                                                \
-            ret_st = Status::InvalidArgument(strings::Substitute(               \
-                    "task(signature=$0) has wrong request member", signature)); \
-        }                                                                       \
+#define HANDLE_TYPE(t_task_type, work_pool, req_member)                                     \
+    case t_task_type:                                                                       \
+        if (task.__isset.req_member) {                                                      \
+            work_pool->submit_task(task);                                                   \
+        } else {                                                                            \
+            ret_st = Status::InvalidArgument("task(signature={}) has wrong request member", \
```
cambyzju, line 128: use #req_member to print the real request member, such as 'create_tablet_req', 'drop_tablet_req', and so on, otherwise this error message is still confusing.

BePPPower: Change it like this? : `Status::InvalidArgument("task(signature={}, request menmber={}) has wrong request member", signature, req_menber);`

cambyzju: #req_member

like: task(signature={}) missing request member={}

BePPPower: ok

yiguolei (approved):
LGTM

PR #9534, opened 2022-05-12
[bugfix]Clean the version.sh file before build, otherwise the version informa…

hello-stephen:
## Problem Summary:

During the testing process, it was found that the recompiled version information of the be was not updated, which made it easy to make mistakes when comparing the version performance during the development and testing process.

It has been modified simply, and there may be a better way to modify it.

## Checklist(Required)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -266,6 +266,7 @@ FE_MODULES=$(IFS=, ; echo "${modules[*]}")
 
 # Clean and build Backend
 if [ ${BUILD_BE} -eq 1 ] ; then
+    if [ -e ${DORIS_HOME}/gensrc/build/gen_cpp/version.h ]; then rm -f ${DORIS_HOME}/gensrc/build/gen_cpp/version.h ; fi
     CMAKE_BUILD_TYPE=${BUILD_TYPE:-Release}
     echo "Build Backend: ${CMAKE_BUILD_TYPE}"
     CMAKE_BUILD_DIR=${DORIS_HOME}/be/build_${CMAKE_BUILD_TYPE}


cambyzju:
If you want to release the version, you could run 'sh build.sh --clean' at first, then run 'sh build.sh' to compile and release the version.

Most of the time, we test and debug frequently, I personally do not think it is necessary to update the version.

yiguolei (approved):
LGTM

hello-stephen:
> If you want to release the version, you could run 'sh build.sh --clean' at first, then run 'sh build.sh' to compile and release the version.
> 
> Most of the time, we test and debug frequently, I personally do not think it is necessary to update the version.

We all agree that the new commit information needs to be packed into the newly compiled package in time. FE is included, but BE is not. In the way of --clean All files will be recompiled, while only 3 files will be recompiled by deleting version.h, which is much faster.

cambyzju:
> > If you want to release the version, you could run 'sh build.sh --clean' at first, then run 'sh build.sh' to compile and release the version.
> > Most of the time, we test and debug frequently, I personally do not think it is necessary to update the version.
> 
> We all agree that the new commit information needs to be packed into the newly compiled package in time. FE is included, but BE is not. In the way of --clean All files will be recompiled, while only 3 files will be recompiled by deleting version.h, which is much faster.

If it is really needed, Does call 'gensrc/script/gen_build_version.sh' better than remove files?

for example:
if [ ${BUILD_BE} -eq 1 ] ; then
    ${DORIS_HOME}/gensrc/script/gen_build_version.sh --be
    ...
fi

PR #8808, opened 2022-04-01
[feature](cold-hot) support s3 resource

qidaye:
# Proposed changes

issue：#8807 
1. Support create s3 resource 
2. Support alter resource
3. Add check whether the resource is in use before deleting it
4. Support create OLAP table with s3 resource
5. Add `remote_storage_resource` and `remote_storage_cooldown_time` in `DataProperty`.
6. Support for modifying a single data property, do not support modify `remote_storage_resource`

## Problem Summary:

Describe the overview of changes.

## Checklist(Required)

1. Does it affect the original behavior: (Yes/No/I Don't know)
7. Has unit tests been added: (Yes/No/No Need)
8. Has document been added or modified: (Yes/No/No Need)
9. Does it need to update dependencies: (Yes/No)
10. Are there any changes that cannot be rolled back: (Yes/No)

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #9554, opened 2022-05-13
Add StoragePolicyResource for Remote Storage

pengxiangyu:
# Proposed changes

Issue Number: close #8808

## Problem Summary:

Add StoragePolicyResource for Remote Storage

## Checklist(Required)

## Further comments

Add StoragePolicyResource for table

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/StoragePolicyResource.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/StoragePolicyResource.java
@@ -77,6 +77,11 @@ public class StoragePolicyResource extends Resource {
         // optional
         checkOptionalProperty(COOLDOWN_DATETIME, DEFAULT_COOLDOWN_DATETIME);
         checkOptionalProperty(COOLDOWN_TTL, DEFAULT_COOLDOWN_TTL);
+        if (properties.containsKey(COOLDOWN_DATETIME) && properties.containsKey(COOLDOWN_TTL) &&
+                !properties.get(COOLDOWN_DATETIME).isEmpty() && !properties.get(COOLDOWN_TTL).isEmpty()) {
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/org/apache/doris/catalog/StoragePolicyResource.java
```diff
@@ -0,0 +1,128 @@
...
+
+    private void checkOptionalProperty(String propertyKey, String defaultValue) {
+        this.properties.putIfAbsent(propertyKey, defaultValue);
+    }
+
+    @Override
+    public void modifyProperties(Map<String, String> properties) throws DdlException {
+        // modify properties
+        replaceIfEffectiveValue(this.properties, STORAGE_RESOURCE, properties.get(STORAGE_RESOURCE));
+        replaceIfEffectiveValue(this.properties, COOLDOWN_DATETIME, properties.get(COOLDOWN_DATETIME));
```
morningman, line 98: We should check that only one of `COOLDOWN_DATETIME` and `COOLDOWN_TTL` can be specified

fe/fe-core/src/main/java/org/apache/doris/catalog/StoragePolicyResource.java
```diff
@@ -0,0 +1,128 @@
...
+
+        if (!copiedProperties.isEmpty()) {
+            throw new AnalysisException("Unknown S3 resource properties: " + copiedProperties);
+        }
+    }
+
+    @Override
+    public Map<String, String> getCopiedProperties() {
+        Map<String, String> copiedProperties = Maps.newHashMap(properties);
+        return copiedProperties;
```
morningman, line 118:
```suggestion
        return Maps.newHashMap(properties);
```

morningman (approved):
LGTM

PR #9560, opened 2022-05-14
fix memory leak

GoGoWen:
# Proposed changes

    /// FunctionContexts for each registered expression. The FunctionContexts are created
    /// and owned by this VExprContext.
    std::vector<FunctionContext*> _fn_contexts;  

    _fn_contexts not be freed which cause memory leak.

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/vec/exprs/vexpr_context.cpp
+++ b/be/src/vec/exprs/vexpr_context.cpp
@@ -30,6 +30,13 @@ VExprContext::VExprContext(VExpr* expr)
           _closed(false),
           _last_result_column_id(-1) {}
 
+VExprContext::~VExprContext() {
+    DCHECK(!_prepared || _closed);
+    for (int i = 0; i < _fn_contexts.size(); ++i) {
+        delete _fn_contexts[i];
+    }
+}
+
 doris::Status VExprContext::execute(doris::vectorized::Block* block, int* result_column_id) {
     Status st = _root->execute(this, block, result_column_id);
     _last_result_column_id = *result_column_id;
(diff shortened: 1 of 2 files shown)


be/src/vec/exprs/vexpr_context.cpp
```diff
@@ -30,6 +30,13 @@ VExprContext::VExprContext(VExpr* expr)
           _closed(false),
           _last_result_column_id(-1) {}
 
+VExprContext::~VExprContext() {
+    DCHECK(!_prepared || _closed);
+    for (int i = 0; i < _fn_contexts.size(); ++i) {
+        delete _fn_contexts[i];
+    }
```
yiguolei, line 37: Although this is a problem. But I do not think it is the root cause for your memory leak. Because 
void VExprContext::close(doris::RuntimeState* state) {
    DCHECK(!_closed);
    FunctionContext::FunctionStateScope scope =
            _is_clone ? FunctionContext::THREAD_LOCAL : FunctionContext::FRAGMENT_LOCAL;
    _root->close(state, this, scope);

    for (int i = 0; i < _fn_contexts.size(); ++i) {
        _fn_contexts[i]->impl()->close();
        delete _fn_contexts[i];
    }
    // _pool can be NULL if Prepare() was never called
    if (_pool != NULL) {
        _pool->free_all();
    }
    _closed = true;
}

the close method will delete every _fn_contexts[i]

yiguolei: Maybe we missed call close some where......

GoGoWen:
as we use v1.0,   after verify with master branch.  the memory leak  should be fixed.  so close this pr.

PR #9566, opened 2022-05-14
optimize lz4 compress and decompress speed by reusing context

xiaokang:
# Proposed changes

optimize lz4 compress and decompress speed by reusing context

## Problem Summary:

Currently, context object is create and free for each lz4 compress and decompress function call. It's not efficient since memory alloc and free is too frequent for once per compression/decompression block, usually a page of less than 64KB. 

In practice, we see hot pot in LZ4_decompress_safe, memcpy, SpinLock::SpinLoop as in the following screenshot.

image

## Checklist(Required)

### DIFF
--- a/be/src/util/block_compression.cpp
+++ b/be/src/util/block_compression.cpp
@@ -399,7 +399,7 @@ Status get_block_compression_codec(segment_v2::CompressionTypePB type,
         ptr = new Lz4BlockCompression();
         break;
     case segment_v2::CompressionTypePB::LZ4F:
-        ptr = new Lz4fBlockCompression(); 
+        ptr = new Lz4fBlockCompression();
         break;
     case segment_v2::CompressionTypePB::ZLIB:
         ptr = new ZlibBlockCompression();
--- a/be/src/util/block_compression.cpp
+++ b/be/src/util/block_compression.cpp
@@ -88,11 +88,6 @@ public:
 // Used for LZ4 frame format, decompress speed is two times faster than LZ4.
 class Lz4fBlockCompression : public BlockCompressionCodec {
 public:
-    static const Lz4fBlockCompression* instance() {
-        static Lz4fBlockCompression s_instance;
-        return &s_instance;
-    }
-
     Status init() override {
         auto ret1 = LZ4F_createCompressionContext(&ctx_c, LZ4F_VERSION);
         if (LZ4F_isError(ret1)) {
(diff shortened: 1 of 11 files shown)


xiaokang:
We see x6.75 speedup in the following comparison test.

test env
- test data: 10GB text log, 44 million rows
- test table: test_table(ts varchar(30), log string)
- test SQL: set enable_vectorized_engine=1; select sum(length(log)) from test_table
- be.conf: disable_storage_page_cache = true   

test result 
- master branch runtime: 5.2 seconds
- lz4_compress_opt branch runtime: 0.77 seconds

be/src/util/block_compression.cpp
```diff
@@ -370,27 +388,39 @@ class ZlibBlockCompression : public BlockCompressionCodec {
...
+    BlockCompressionCodec* ptr = nullptr;
     switch (type) {
     case segment_v2::CompressionTypePB::NO_COMPRESSION:
-        *codec = nullptr;
-        break;
+        codec.reset(nullptr);
+        return Status::OK();
     case segment_v2::CompressionTypePB::SNAPPY:
-        *codec = SnappyBlockCompression::instance();
+        ptr = new SnappyBlockCompression();
```
yiguolei, line 398: could write this way:  codec.reset(new SnappyBlockCompression());

be/src/util/block_compression.cpp
```diff
@@ -370,27 +388,39 @@ class ZlibBlockCompression : public BlockCompressionCodec {
...
+        ptr = new ZlibBlockCompression();
         break;
     default:
         return Status::NotFound(strings::Substitute("unknown compression type($0)", type));
     }
-    return Status::OK();
+
+    if (!ptr)
+        return Status::NotFound("Failed to create compression codec");
+
```
yiguolei, line 415: just write: return codec->init();

be/src/util/block_compression.cpp
xiaokang: Yes, it's an easier way to use codec.reset(new SnappyBlockCompression()). 

But I changed to the current more complex way, new a ptr first and then call ptr->init() and call codec.reset(ptr) finally, to avoid reset codec to a not properly initialized prt.

yiguolei (approved):
LGTM

xiaokang:
Add some information about correctness validation. We import the test data and then export data (select * from test_table into outfile ...) using master and lz4_compress_opt branch. The results of two branches are the same after sort.

btw. add a comment for NOTICE of NOT thread safe.

PR #9582, opened 2022-05-16
[refactor](decimalv3) optimize decimal performance and precision

### LINKED ISSUE #9208: [Enhancement] (decimalv3) optimize decimal performance and precision
### Search before asking

### Description

At present, the decimal type has some problems.

1. The unified use of decimal128 in the vectorization execution layer brings performance and memory overhead. For low precision, decimal32 or decimal64 can be used.

2. Insufficient precision support for decimal. For decimal(p, s),  1 < p < 27 and 0 < s < 9. In some scenarios, we need higher precision and scale.

### Solution

There are several stages to optimize the performance and precision of decimal.

**Step1:**  Fe passed precision and scale of decimal to be, so the vectorization execution layer can use this information.

**Step2:** The vectorization execution layer uses decimal32, decimal64 or decimal128 to calculate according to the precision and scale information from fe.

**Step3:** Add a new decimal type to the storage layer. Decimal of low precision uses decimal 32 (int32_t) or decimal 64 (int64_t) for storage.  Decimal of high precision uses decimal128 (_int128) for storage. It can support the maximum precision of 38. Scale only needs to be less than precision.

### Code of Conduct

liaoxin01:
# Proposed changes

Issue Number: close #9208 

## Problem Summary:

## Checklist(Required)

## Further comments
TODO：
1. Schema change on decimal column between different precision and scale.
2. Cast on decimal between different precision and scale.
3. The division operation requires that the scale of the dividend is not less than the scale of divisor, otherwise be will crash.
4. Overflow in decimal calculations needs to be handled.
5. predicate pushdown for decimalv3.


be/src/common/config.h
```diff
@@ -736,6 +736,12 @@ CONF_Validator(string_type_length_soft_limit_bytes,
 // used for olap scanner to save memory, when the size of unused_object_pool
 // is greater than object_pool_buffer_size, release the object in the unused_object_pool.
 CONF_Int32(object_pool_buffer_size, "100");
+
+// decimalv3: Decimal32/Decimal64/Decimal128
+// decimalv2 is converted to decimalv3 for calculation when enable_engine_decimalv3 is true.
+CONF_Bool(enable_engine_decimalv3, "false");
```
Gabriel39, line 742:
```suggestion
CONF_Bool(enable_execution_decimalv3, "false");
```

be/src/olap/types.h
```diff
@@ -878,6 +893,87 @@ struct FieldTypeTraits<OLAP_FIELD_TYPE_DECIMAL>
...
+    }
+    static std::string to_string(const void* src) {
+        int128_t value = reinterpret_cast<const PackedInt128*>(src)->value;
+        fmt::memory_buffer buffer;
+        fmt::format_to(buffer, "{}", value);
+        return std::string(buffer.data(), buffer.size());
+    }
+    static void set_to_max(void* buf) {
+        *reinterpret_cast<PackedInt128*>(buf) =
+                static_cast<int128_t>(999999999999999999ll) * 100000000000000000ll * 1000ll +
```
Gabriel39, line 967: Better to use hexadecimal numbers

be/src/vec/aggregate_functions/aggregate_function_avg.h
```diff
@@ -46,13 +46,15 @@ struct AggregateFunctionAvgData {
...
             return static_cast<ResultT>(sum);
         }
-        // to keep the same result with row vesion; see AggregateFunctions::decimalv2_avg_get_value
-        if constexpr (std::is_same_v<ResultT, Decimal128> && std::is_same_v<T, Decimal128>) {
-            DecimalV2Value decimal_val_count(count, 0);
-            DecimalV2Value decimal_val_sum(static_cast<Int128>(sum));
-            DecimalV2Value cal_ret = decimal_val_sum / decimal_val_count;
-            Decimal128 ret(cal_ret.value());
-            return ret;
+        if (!config::enable_engine_decimalv3) {
```
Gabriel39, line 49:
```suggestion
        if (!config::enable_execution_decimalv3) {
```

be/src/vec/exprs/vliteral.cpp
```diff
@@ -109,8 +109,33 @@ void VLiteral::init(const TExprNode& node) {
         case TYPE_DECIMALV2: {
             DCHECK_EQ(node.node_type, TExprNodeType::DECIMAL_LITERAL);
             DCHECK(node.__isset.decimal_literal);
-            DecimalV2Value value(node.decimal_literal.value);
-            field = DecimalField<Decimal128>(value.value(), value.scale());
+            if (config::enable_engine_decimalv3) {
+                DataTypePtr type_ptr = create_decimal(node.type.types[0].scalar_type.precision,
+                                                      node.type.types[0].scalar_type.scale);
+                WhichDataType which(type_ptr->get_type_id());
```
cambyzju, line 115: WhichDataType(const DataTypePtr& data_type) support DataTypePtr as input argument.

fe/fe-core/src/main/java/org/apache/doris/analysis/TypeDef.java
```diff
@@ -159,15 +159,28 @@ private void analyzeScalarType(ScalarType scalarType)
...
-          throw new AnalysisException("Precision of decimal must between 1 and 27."
-                  + " Precision was set to: " + precision + ".");
-        }
-        // scale: [0, 9]
-        if (scale < 0 || scale > 9) {
-          throw new AnalysisException("Scale of decimal must between 0 and 9."
-                  + " Scale was set to: " + scale + ".");
+        if (Config.enable_storage_decimalv3) {
+            // precision: [1, 38]
+            if (precision < 1 || precision > 38) {
```
cambyzju, line 164: it is better replace 38 with MAX_DECIMAL128_PRECISION.

fe/fe-core/src/main/java/org/apache/doris/analysis/TypeDef.java
```diff
@@ -159,15 +159,28 @@ private void analyzeScalarType(ScalarType scalarType)
...
+                throw new AnalysisException("Precision of decimal must between 1 and 27."
```
cambyzju, line 165: The error message "between 1 and 27" do not match the if condition(1~38).

be/src/olap/field.h
```diff
@@ -298,6 +298,11 @@ class Field {
     Field* get_sub_field(int i) const { return _sub_fields[i].get(); }
     size_t get_sub_field_count() const { return _sub_fields.size(); }
 
+    void set_precision(int32_t precision) { _precision = precision; }
+    void set_frac(int32_t frac) { _frac = frac; }
+    int32_t precision() const { return _precision; }
+    int32_t frac() const { return _frac; }
```
cambyzju, line 304: Does frac the same as scale? If it is true, why not use scale?

be/src/vec/columns/column.h
```diff
@@ -183,6 +183,11 @@ class IColumn : public COW<IColumn> {
     /// All data will be inserted as single element
     virtual void insert_data(const char* pos, size_t length) = 0;
 
+    // just used for ColumnDecimal to be compatible with decimalv2
+    virtual void insert_many_decimalv2_data(const char* pos, size_t num) {
```
cambyzju, line 187: `insert_many_decimalv2_data` is not a common function, it is better not add it to IColumn.

While using, you could cast IColumn to ColumnDecimal, and call the function.

be/src/vec/data_types/data_type_decimal.h
```diff
@@ -106,8 +107,8 @@ class DataTypeDecimal final : public IDataType {
         }
 
         // Now, Doris only support precision:27, scale: 9
```
cambyzju, line 109: just remove these lines.

be/src/vec/sink/mysql_result_writer.cpp
```diff
@@ -419,9 +447,13 @@ Status VMysqlResultWriter::append_block(Block& input_block) {
         }
         case TYPE_DECIMALV2: {
             if (type_ptr->is_nullable()) {
-                status = _add_one_column<PrimitiveType::TYPE_DECIMALV2, true>(column_ptr, result);
+                auto& nested_type =
+                        assert_cast<const DataTypeNullable&>(*type_ptr).get_nested_type();
+                status = _add_one_column<PrimitiveType::TYPE_DECIMALV2, true>(column_ptr, result,
```
cambyzju, line 452: why we need nested_type here?

Gabriel39:
We should support schema change on decimal column between different precision and scale which is not done in this PR.

be/src/vec/sink/mysql_result_writer.cpp
liaoxin01: used for _add_one_column function to infer the type of decimal column.

be/src/olap/field.h
liaoxin01: yes, frac is the same as scale. All code uses frac in src/olap, so I used frac here.

be/src/vec/columns/column.h
liaoxin01: This function will be called in bitshuffle_page.h. ColumnDecimal<T> is template class，but we don't know the real type of T in bitshuffle_page.h. so we can't  use static_cast here.

There are several solutions：
1. use dynamic_cast
2. change insert_many_decimalv2_data to a template function

I don't think they are better than the current implementation.

liaoxin01:
> We should support schema change on decimal column between different precision and scale which is not done in this PR.

Yes, we should support  it. I plan to write the schema change code after this PR.

be/src/vec/functions/function_binary_arithmetic.h
```diff
@@ -449,6 +469,18 @@ struct DecimalBinaryOperation {
 private:
     /// there's implicit type convertion here
     static NativeResultType apply(NativeResultType a, NativeResultType b) {
+        if (config::enable_execution_decimalv3) {
+            if constexpr (OpTraits::can_overflow && check_overflow) {
+                NativeResultType res;
+                if (Op::template apply<NativeResultType>(a, b, res)) {
+                    LOG(FATAL) << "Decimal math overflow";
```
morningman, line 476: Better not use FATAL log, it will cause BE crash.
Could we return a invalid value or null to indicate this error?
And in what situation can we get here?

be/src/vec/columns/column_decimal.cpp
```diff
@@ -130,6 +131,38 @@ void ColumnDecimal<T>::insert_data(const char* src, size_t /*length*/) {
...
+            if (is_negative) {
+                int_value = std::abs(int_value);
+                frac_value = std::abs(frac_value);
+            }
+            frac_value /= (DecimalV2Value::ONE_BILLION / get_scale_multiplier());
+            T value = T(int_value) * get_scale_multiplier() + T(frac_value);
+            if (is_negative) {
+                value = -value;
+            }
+            this->insert_data(reinterpret_cast<char*>(&value), 0);
```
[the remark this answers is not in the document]
liaoxin01: In this method, we can insert many decimal with same precision and scale. This method converts storage layer decimalv2 to decimalv3 for calculation.

be/src/vec/columns/column_decimal.cpp
```diff
@@ -130,6 +131,38 @@ void ColumnDecimal<T>::insert_data(const char* src, size_t /*length*/) {
...
+        } else {
+            DecimalV2Value decimal_val(int_value, frac_value);
+            this->insert_data(reinterpret_cast<char*>(&decimal_val), 0);
+        }
+    }
+}
+
+template <typename T>
+void ColumnDecimal<T>::insert_many_fix_len_data(const char* data_ptr, size_t num) {
```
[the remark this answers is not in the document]
liaoxin01: New decimalv3 type added on storage layer will use this method, and origin decimalv2 will use insert_many_decimalv2_data.

be/src/vec/data_types/data_type_factory.cpp
```diff
@@ -26,13 +26,75 @@ namespace doris::vectorized {
 
 DataTypePtr DataTypeFactory::create_data_type(const doris::Field& col_desc) {
     DataTypePtr nested = nullptr;
-    if (col_desc.type() == OLAP_FIELD_TYPE_ARRAY) {
+    switch (col_desc.type()) {
```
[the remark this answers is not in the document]
liaoxin01: Decimal need precision and scale information, but primitive data type don't have that information.

be/src/vec/functions/function_binary_arithmetic.h
liaoxin01: Decimal addition and multiplication may overflow.

Gabriel39: how about to return null when overflow?

liaoxin01: The column may not be ColumnNullable, so we can't return null.

Gabriel39:
1. Add UTs in BE
2. rebase master to resolve conflicts

be/src/vec/olap/olap_data_convertor.h
```diff
@@ -164,6 +165,45 @@ class OlapBlockDataConvertor {
         PaddedPODArray<decimal12_t> _values;
```
Gabriel39, line 165: We should use different types to store value for different decimal precision

Gabriel39:
PredicateColumn for DecimalV3 is still not supported, right?

be/src/vec/exec/vaggregation_node.cpp
```diff
@@ -145,9 +145,22 @@ void AggregationNode::_init_hash_method(std::vector<VExprContext*>& probe_exprs)
             _agg_data.init(AggregatedDataVariants::Type::int64_key, is_nullable);
             return;
         case TYPE_LARGEINT:
-        case TYPE_DECIMALV2:
-            _agg_data.init(AggregatedDataVariants::Type::int128_key, is_nullable);
+        case TYPE_DECIMALV2: {
```
Gabriel39, line 148: this code block should also be added to vhashjoinnode

Gabriel39:
also add new decimal types to `_infos_mapping` in `AggregateFuncResolver`

be/src/olap/tablet_meta.cpp
```diff
@@ -202,12 +202,24 @@ void TabletMeta::_init_column_from_tcolumn(uint32_t unique_id, const TColumn& tc
...
+    uint32_t length = TabletColumn::get_field_length_by_type(tcolumn.column_type.type,
+                                                             tcolumn.column_type.len);
     if (tcolumn.column_type.type == TPrimitiveType::DECIMALV2) {
         column->set_precision(tcolumn.column_type.precision);
         column->set_frac(tcolumn.column_type.scale);
+        if (config::enable_storage_decimalv3) {
+            if (tcolumn.column_type.precision <= 9) {
+                length = sizeof(int32_t);
+                column->set_type("DECIMAL32");
+            } else if (tcolumn.column_type.precision <= 18) {
```
HappenLee, line 214: should use `MAX_DECIMAL64_PRECISION` to replace all code use `18`.
same to `9` and `38`

be/src/vec/data_types/data_type_factory.cpp
```diff
@@ -26,13 +26,75 @@ namespace doris::vectorized {
 
 DataTypePtr DataTypeFactory::create_data_type(const doris::Field& col_desc) {
```
HappenLee, line 27: You should simplify this code by making call `DataTypeFactory::create_data_type(const TabletColumn& col_desc, bool is_nullable)`. do not have two big and same `switch`

be/src/vec/functions/function_binary_arithmetic.h
```diff
@@ -529,6 +565,31 @@ struct DecimalBinaryOperation {
             return apply(a, b, is_null);
         }
     }
+
+    template <bool scale_left>
+    static NativeResultType apply_scaled_mod(NativeResultType a, NativeResultType b,
+                                             NativeResultType scale, UInt8& is_null) {
+        if constexpr (OpTraits::is_mod) {
```
HappenLee, line 572: if not `is_mod` return nothing?

be/src/vec/sink/mysql_result_writer.cpp
```diff
@@ -144,6 +145,35 @@ Status VMysqlResultWriter::_add_one_column(const ColumnPtr& column_ptr,
             _buffer.close_dynamic_mode();
             result->result_batch.rows[i].append(_buffer.buf(), _buffer.length());
         }
+    } else if (type == TYPE_DECIMALV2 && config::enable_execution_decimalv3) {
```
HappenLee, line 148: why the code here not 234 line like
```
if (config::enable_execution_decimalv3) {
  .....
} else {
 DecimalV2Value ...
}
```

be/src/vec/olap/olap_data_convertor.h
```diff
@@ -182,12 +182,56 @@ class OlapBlockDataConvertor {
...
+    template <typename T>
     class OlapColumnDataConvertorDecimal
             : public OlapColumnDataConvertorPaddedPODArray<decimal12_t> {
     public:
         Status convert_to_olap() override;
     };
 
+    // decimalv3 don't need to do any convert
+    template <typename T>
+    class OlapColumnDataConvertorDecimalV3 : public OlapColumnDataConvertorBase {
```
HappenLee, line 197: should be a subclass of `OlapColumnDataConvertorSimple` only need override the `convert_to_olap` method

be/src/olap/field.h
```diff
@@ -340,6 +345,8 @@ class Field {
     uint16_t _index_size;
     bool _is_nullable;
     std::vector<std::unique_ptr<Field>> _sub_fields;
+    int32_t _precision;
```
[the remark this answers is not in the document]
liaoxin01: Just to be consistent with other code.

Gabriel39:
DecimalV3 also needs to be processed in runtime filter

liaoxin01:
Will be resubmitted in another pr #10545.

PR #9593, opened 2022-05-16
[Feature]Support create hive table without column info.

Jibing-Li:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

Support create hive external table without giving the schema. Doris would sync the schema from HMS automatically.
And user doesn't need to manually refresh the column info when the hive schema changes. 

Example:
CREATE TABLE `region` 
engine=hive 
properties ("database"="default", "table"="region", "hive.metastore.uris"="thrift://172.21.16.11:7004");

## Checklist(Required)

1. Does it affect the original behavior: (Yes/No/I Don't know) No
2. Has unit tests been added: (Yes/No/No Need) No Need
3. Has document been added or modified: (Yes/No/No Need) No
4. Does it need to update dependencies: (Yes/No) No
5. Are there any changes that cannot be rolled back: (Yes/No) No

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/HiveTable.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/HiveTable.java
@@ -18,13 +18,18 @@
 package org.apache.doris.catalog;
 
 import org.apache.doris.common.DdlException;
+import org.apache.doris.common.FeMetaVersion;
 import org.apache.doris.common.io.Text;
 import org.apache.doris.thrift.THiveTable;
 import org.apache.doris.thrift.TTableDescriptor;
 import org.apache.doris.thrift.TTableType;
 
 import com.google.common.base.Strings;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import org.apache.hadoop.hive.metastore.api.FieldSchema;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import java.io.DataInput;
 import java.io.DataOutput;
@@ -32,12 +37,16 @@ import java.io.IOException;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * External hive table
  * Currently only support loading from hive table
  */
 public class HiveTable extends Table {
+    private static final Logger LOG = LogManager.getLogger(HiveTable.class);
+
     private static final String PROPERTY_MISSING_MSG = "Hive %s is null. Please add properties('%s'='xxx') when create table";
 
     private static final String HIVE_DB = "database";
@@ -49,6 +58,9 @@ public class HiveTable extends Table {
     private String hiveTable;
     private Map<String, String> hiveProperties = Maps.newHashMap();
 
+    private boolean isLocalSchema = true;
+    private Pattern digitPattern = null;
+
     public HiveTable() {
         super(TableType.HIVE);
     }
@@ -56,6 +68,10 @@ public class HiveTable extends Table {
     public HiveTable(long id, String name, List<Column> schema, Map<String, String> properties) throws DdlException {
         super(id, name, TableType.HIVE, schema);
         validate(properties);
+        if (schema == null || schema.size() == 0) {
+            isLocalSchema = false;
+            digitPattern = Pattern.compile("(\\d+)");
+        }
     }
 
     public String getHiveDbTable() {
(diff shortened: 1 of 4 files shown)


fe/fe-core/src/main/java/org/apache/doris/catalog/HiveTable.java
```diff
@@ -74,6 +89,92 @@ public Map<String, String> getHiveProperties() {
...
 
+    @Override
+    public List<Column> getBaseSchema(boolean full) {
+        if (isLocalSchema) {
+            return super.getBaseSchema(full);
+        }
+        List<Column> hiveMetastoreSchema = Lists.newArrayList();
+        try {
+            for (FieldSchema field : HiveMetaStoreClientHelper.getSchema(this)) {
+                field.getType();
```
qidaye, line 100: unused code.

fe/fe-core/src/main/java/org/apache/doris/catalog/HiveTable.java
```diff
@@ -74,6 +89,92 @@ public Map<String, String> getHiveProperties() {
...
+        }
+        for (Column column : fullSchema) {
+            if (column.getName().equals(name)) {
+                return column;
+            }
+        }
+        return col;
+    }
+
+    private Type convertToDorisType(String hiveType) {
```
qidaye, line 129: `boolean`, `timestamp` are missing.

fe/fe-core/src/main/java/org/apache/doris/catalog/HiveTable.java
```diff
@@ -74,6 +89,92 @@ public Map<String, String> getHiveProperties() {
...
+        } catch (DdlException e) {
+            LOG.warn("Failed to get schema of hive table. DB {}, Table {}. {}",
+                    this.hiveDb, this.hiveTable, e.getMessage());
+            return null;
+        }
+        fullSchema = hiveMetastoreSchema;
+        return fullSchema;
+    }
+
+    public Column getColumn(String name) {
```
qidaye, line 113: Unused function.

qidaye (commented):
How many steps are you planning to implement this in? I don't see the sync logic after schema change.

fe/fe-core/src/main/java/org/apache/doris/catalog/HiveTable.java
Jibing-Li: added

fe/fe-core/src/main/java/org/apache/doris/catalog/HiveTable.java
Jibing-Li: This is function is used in describe table and query statement. It is override the function in Table

fe/fe-core/src/main/java/org/apache/doris/catalog/HiveTable.java
Jibing-Li: Removed

Jibing-Li:
> How many steps are you planning to implement this in? I don't see the sync logic after schema change.

This is the only PR to implement this. We don't need to sync the changed schema, it is read the schema from HMS every time. So it is automatically synced.

fe/fe-core/src/main/java/org/apache/doris/catalog/HiveTable.java
qidaye: OK

fe/fe-core/src/main/java/org/apache/doris/catalog/HiveTable.java
```diff
@@ -74,6 +89,98 @@ public Map<String, String> getHiveProperties() {
...
+            }
+            if (match.find()) {
+                scale = Integer.parseInt(match.group(1));
+            }
+            return ScalarType.createDecimalV2Type(precision, scale);
+        }
+        if (lowerCaseType.equals("date")) {
+            return Type.DATE;
+        }
+        if (lowerCaseType.equals("datetime")) {
```
qidaye, line 177:
```suggestion
        if (lowerCaseType.equals("timestamp")) {
```

qidaye:
> In the code, we don't keep the schema in Catalog, the schema is fetched from HMS every time when getBaseSchema is called, so it is always consistent with the Hive schema. getBaseSchema is called when user execute describe or query statement. It was tested and the behavior looks correct.

Is this the right way to go to HMS to get the schema for every query? 
Or can we retry and get the new schema when the query goes wrong, or expose an refresh interface for user to update it manually, like the iceberg external table?

@morningman @yangzhg can you take a look at this?

morningman (commented):
I don't recommend this approach very much.
I propose to unify this work into our subsequent multi catalog solution.

PR #9418, opened 2022-05-06
[Bug] Missing error tablet list when close_wait return error

pengxiangyu:
# Proposed changes

Issue Number: close #9417

## Problem Summary:

In function close_wait() of delta_writer.cpp, if _flush_token->wait() return an error, it will not be returned to front.
Because in PInternalServiceImpl::tablet_writer_add_batch, _exec_env->load_channel_mgr()->add_batch(*request, response); will not return an error, but only return error_tablets. However error_tablets in close_wait() is not returned.

## Checklist(Required)

1. Does it affect the original behavior: (No)
2. Has unit tests been added: (No)
3. Has document been added or modified: (No)
4. Does it need to update dependencies: (No)
5. Are there any changes that cannot be rolled back: (No)

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #9678, opened 2022-05-19
add tablet erros when close_wait return error

pengxiangyu:
# Proposed changes

Issue Number: close #9418 

## Problem Summary:

add tablet erros when close_wait return error

## Checklist(Required)

1. Does it affect the original behavior: (No/

### DIFF
--- a/be/src/olap/delta_writer.h
+++ b/be/src/olap/delta_writer.h
@@ -67,9 +67,6 @@ public:
     Status close();
     // wait for all memtables to be flushed.
     // mem_consumption() should be 0 after this function returns.
-    Status close_wait(google::protobuf::RepeatedPtrField<PTabletInfo>* tablet_vec,
-                      google::protobuf::RepeatedPtrField<PTabletError>* tablet_errors,
-                      bool is_broken);
     Status close_wait();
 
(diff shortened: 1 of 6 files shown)


be/src/olap/delta_writer.h
```diff
@@ -70,6 +70,7 @@ class DeltaWriter {
     Status close_wait(google::protobuf::RepeatedPtrField<PTabletInfo>* tablet_vec,
                       google::protobuf::RepeatedPtrField<PTabletError>* tablet_errors,
                       bool is_broken);
+    Status close_wait();
```
yangzhg, line 73: why add this method？

be/src/runtime/tablets_channel.cpp
```diff
@@ -137,6 +137,23 @@ Status TabletsChannel::close(int sender_id, int64_t backend_id, bool* finished,
     return Status::OK();
 }
 
+void TabletsChannel::_close_wait(DeltaWriter* writer,
```
yangzhg, line 140: where has this method has been used ?

be/src/olap/delta_writer.h
pengxiangyu: it is not ready,

PR #9689, opened 2022-05-19
[fix] NullPredicate should implement evaluate_vec

dataroaring:
select column from table where column is null

# Proposed changes

Issue Number: close #xxx

## Problem Summary:

select from where is null does not work.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/null_predicate.cpp
+++ b/be/src/olap/null_predicate.cpp
@@ -172,9 +172,9 @@ void NullPredicate::evaluate_vec(vectorized::IColumn& column, uint16_t size, boo
         for (uint16_t i = 0; i < size; ++i) {
             flags[i] = (null_map[i] == _is_null);
         }
-        return;
+    } else {
+        if (_is_null) memset(flags, false, size);
     }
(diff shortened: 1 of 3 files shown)


be/src/olap/null_predicate.cpp
```diff
@@ -165,4 +165,16 @@ void NullPredicate::evaluate_and(IColumn& column, uint16_t* sel, uint16_t size,
...
     }
 }
+
+void NullPredicate::evaluate_vec(vectorized::IColumn& column, uint16_t size, bool* flags) const {
+    if (auto* nullable = check_and_get_column<ColumnNullable>(column)) {
+        auto& null_map = nullable->get_null_map_data();
+        for (uint16_t i = 0; i < size; ++i) {
+            flags[i] = (null_map[i] == _is_null);
+        }
+        return;
```
HappenLee, line 175:
```suggestion
  } else if (_is_null) {
     memset
  }
```

HappenLee (approved):
LGTM

PR #9720, opened 2022-05-21
[improvement] Optimize send fragment logic to reduce send fragment timeout error

### LINKED ISSUE #8942: [Bug] too many "send fragment timeout. backend id" error
### Search before asking

### Version

trunk

### What's Wrong?

In some high concurrency scenarios, a large number of "send fragment timeout. backend id xxx" errors may appear in the fe.log. And all subsequent requests keep reporting this error and cannot be recovered.

### What You Expected?

This is because in high load scenarios, the execution thread pool on the BE side is full, and new requests will enter the waiting queue of the thread pool.
However, the timeout period of the plan fragment request sent by FE to BE is only 5 seconds. The waiting time for requests entering the waiting queue may be very long, resulting in a large number of rpc errors `send fragment timeout`.

And subsequent requests will continue to enter the waiting queue, causing all subsequent requests to time out.

### How to Reproduce?

Run a complex queries with high concurrency.

### Anything Else?

_No response_

### Code of Conduct

morningman:
# Proposed changes

Issue Number: close #8942

## Problem Summary:

This CL mainly changes:
1. Reducing the rpc timeout problem caused by rpc waiting for the worker thread of brpc.
    1. Merge multiple fragment instances on the same BE to send requests to reduce the number of send fragment rpcs
    2. If fragments size >= 3, use 2 phase RPC: one is to send all fragments, two is to start these fragments. So that there
         will be at most 2 RPC for each query on one BE.

3. Set the timeout of send fragment rpc to the query timeout to ensure the consistency of users' expectation of query timeout period.

4. Do not close the connection anymore when rpc timeout occurs.
5. Change some log level from info to debug to simplify the fe.log content.

NOTICE:
1. Change the definition of execPlanFragment rpc, must first upgrade BE.
3. Remove FE config `remote_fragment_exec_timeout_ms`

## Checklist(Required)

4. Has unit tests been added: (No Need)
6. Has document been added or modified: (Yes)

## Further comments

### DIFF
--- a/be/src/runtime/fragment_mgr.cpp
+++ b/be/src/runtime/fragment_mgr.cpp
@@ -546,8 +546,8 @@ Status FragmentMgr::start_query_execution(const PExecPlanFragmentStartRequest* r
     if (search == _fragments_ctx_map.end()) {
         return Status::InternalError(
                 strings::Substitute("Failed to get query fragments context. Query may be "
-                    "timeout or be cancelled. host: ",
-                    BackendOptions::get_localhost()));
+                                    "timeout or be cancelled. host: ",
+                                    BackendOptions::get_localhost()));
     }
     search->second->set_ready_to_execute();
     return Status::OK();
@@ -629,8 +629,7 @@ Status FragmentMgr::exec_plan_fragment(const TExecPlanFragmentParams& params, Fi
             std::lock_guard<std::mutex> lock(_lock);
             auto search = _fragments_ctx_map.find(params.params.query_id);
             if (search == _fragments_ctx_map.end()) {
-                _fragments_ctx_map.insert(
-                        std::make_pair(fragments_ctx->query_id, fragments_ctx));
+                _fragments_ctx_map.insert(std::make_pair(fragments_ctx->query_id, fragments_ctx));
             } else {
                 // Already has a query fragmentscontext, use it
                 fragments_ctx = search->second;
@@ -639,8 +638,8 @@ Status FragmentMgr::exec_plan_fragment(const TExecPlanFragmentParams& params, Fi
     }
 
     exec_state.reset(new FragmentExecState(fragments_ctx->query_id,
-                                           params.params.fragment_instance_id,
-                                           params.backend_num, _exec_env, fragments_ctx));
+                                           params.params.fragment_instance_id, params.backend_num,
+                                           _exec_env, fragments_ctx));
     if (params.__isset.need_wait_execution_trigger && params.need_wait_execution_trigger) {
         // set need_wait_execution_trigger means this instance will not actually being executed
         // until the execPlanFragmentStart RPC trigger to start it.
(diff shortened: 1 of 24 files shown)


gensrc/thrift/PaloInternalService.thrift
```diff
@@ -346,6 +346,10 @@ struct TExecPlanFragmentParams {
   19: optional TGlobalDict global_dict  // scan node could use the global dict to encode the string value to an integer
 }
 
+struct TExecPlanFragmentParamsList {
+    1: optional list<TExecPlanFragmentParams> paramsList;
```
xinyiZzz, line 350: Add comment, better

xinyiZzz:
LGTM

morningman:
I have made following tests:
1. Complex query with 47 fragments. Before: 47 RPCs. After: 2 RPCs
2. High concurrency test for query with 3 fragments: there is no impact on QPS.
3. High concurrency test for query with 2 fragments: these is no impact on QPS.
4. The new BE is compatible with request from old FE.

morningman:
@xinyiZzz @yangzhg PTAL

be/src/runtime/fragment_mgr.cpp
```diff
@@ -225,6 +230,11 @@ Status FragmentExecState::prepare(const TExecPlanFragmentParams& params) {
 }
 
 Status FragmentExecState::execute() {
+    if (_need_wait_execution_trigger) {
+        // if _need_wait_execution_trigger is true, which means this instance
+        // is prepared but need to wait for the signal to do the rest execution.
+        _fragments_ctx->wait_for_start();
```
yiguolei, line 236: Add a timeout here, avoid occupy the running thread too much time during exceptions.

yiguolei, line 236: Add a timeout here, avoid occupy the running thread too much time during exceptions.

morningman: No need, there is a "timeout checker" on BE side to check and notify this.

yangzhg:
I think it is better to divide into two rpc interfaces. `exec_plan_fragment` processing only  fragments size < 3. When 2 phase RPC is required, use two new rpc interfaces `exec_plan_fragment_prepare` and `exec_plan_fragment_start`. This may be more clear

yangzhg:
> I have made following tests:
> 
> 1. Complex query with 47 fragments. Before: 47 RPCs. After: 2 RPCs
> 2. High concurrency test for query with 3 fragments: there is no impact on QPS.
> 3. High concurrency test for query with 2 fragments: these is no impact on QPS.
> 4. The new BE is compatible with request from old FE.

In addition to the observed reduction in the number of RPCs, are there any mitigations for the timeout issue?

morningman:
> I think it is better to divide into two rpc interfaces. `exec_plan_fragment` processing only fragments size < 3. When 2 phase RPC is required, use two new rpc interfaces `exec_plan_fragment_prepare` and `exec_plan_fragment_start`. This may be more clear

Let me try

morningman:
> In addition to the observed reduction in the number of RPCs, are there any mitigations for the timeout issue?

I tested a sql with 3 fragments, 1FE, 1BE, jmeter, thread num 100.
before: send fragment timeout after running a few seconds.
after: no error.

dataroaring (approved):
LGTM

PR #9725, opened 2022-05-22
[Enhancement][Step1] Add a ProfileStorage interface to support record all the querylog in persist storage.

SaintBacchus:
# Proposed changes
Related with [9724](https://github.com/apache/incubator-doris/issues/9724)

## Problem Summary:
Abstract an interface named ProfileStorage with write and read method and impl the InMemoryProfileStorage which will keep the same behaviour with now.

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/common/Pair.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/Pair.java
@@ -22,7 +22,6 @@ package org.apache.doris.common;
 
 import com.google.gson.annotations.SerializedName;
 
-import java.io.Serializable;
 import java.util.Comparator;
 
 /**
@@ -30,7 +29,7 @@ import java.util.Comparator;
  *
  * Notice: When using Pair for persistence, users need to guarantee that F and S can be serialized through Gson
  */
-public class Pair<F, S> implements Serializable {
+public class Pair<F, S> {
     public static PairComparator<Pair<?, Comparable>> PAIR_VALUE_COMPARATOR = new PairComparator<>();
 
     @SerializedName(value = "first")
--- a/fe/fe-core/src/main/java/org/apache/doris/common/Pair.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/Pair.java
@@ -22,6 +22,7 @@ package org.apache.doris.common;
 
 import com.google.gson.annotations.SerializedName;
 
+import java.io.Serializable;
 import java.util.Comparator;
 
 /**
@@ -29,7 +30,7 @@ import java.util.Comparator;
  *
  * Notice: When using Pair for persistence, users need to guarantee that F and S can be serialized through Gson
  */
-public class Pair<F, S> {
+public class Pair<F, S> implements Serializable {
     public static PairComparator<Pair<?, Comparable>> PAIR_VALUE_COMPARATOR = new PairComparator<>();
 
     @SerializedName(value = "first")
--- a/fe/fe-core/src/main/java/org/apache/doris/common/profile/InMemoryProfileStorage.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/profile/InMemoryProfileStorage.java
@@ -1,0 +1,138 @@
+package org.apache.doris.common.profile;
+
+import org.apache.doris.common.util.ProfileManager;
+import org.apache.doris.common.util.ProfileManager.ProfileElement;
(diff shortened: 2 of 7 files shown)


fe/fe-core/src/main/java/org/apache/doris/common/profile/InMemoryProfileStorage.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.concurrent.locks.ReentrantReadWriteLock.WriteLock;
+
+/**
+ * An inMemory impl of profile storage. It will only keep 100 records in memory.
+ * If you want to keep more query records, you can use other impl of profile storage.
+ */
+public class InMemoryProfileStorage implements ProfileStorage {
+    private static final Logger LOG = LogManager.getLogger(InMemoryProfileStorage.class);
+
+    private static final int ARRAY_SIZE = 100;
```
stalary, line 47: Maybe we can control the size by variables.

fe/fe-core/src/main/java/org/apache/doris/common/util/ProfileManager.java
```diff
@@ -104,160 +90,60 @@ public static ProfileManager getInstance() {
     }
 
     private ProfileManager() {
-        lock = new ReentrantReadWriteLock(true);
-        readLock = lock.readLock();
-        writeLock = lock.writeLock();
-        queryIdDeque = new LinkedList<>();
-        queryIdToProfileMap = new ConcurrentHashMap<>();
+        storage = new InMemoryProfileStorage();
```
stalary, line 93: Will we use variables to decide which storage to use in the future?

fe/fe-core/src/main/java/org/apache/doris/common/Pair.java
```diff
@@ -22,14 +22,15 @@
...
+import java.io.Serializable;
 import java.util.Comparator;
 
 /**
  * The equivalent of C++'s std::pair<>.
  *
  * Notice: When using Pair for persistence, users need to guarantee that F and S can be serialized through Gson
  */
-public class Pair<F, S> {
+public class Pair<F, S> implements Serializable {
```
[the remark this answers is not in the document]
SaintBacchus: I planned to seriable profile to bytes. And now I will remove this

fe/fe-core/src/main/java/org/apache/doris/common/util/ProfileManager.java
SaintBacchus: @stalary yes, I plan to use  variables in next pr, because there is only one impl of it

fe/fe-core/src/main/java/org/apache/doris/common/profile/InMemoryProfileStorage.java
SaintBacchus: This is a good idea. But this size is also hard code in web ui which plan to change in the last pr

stalary:
Can you supplement the unit tests.

SaintBacchus:
@stalary In the lastest commit I separate the interface `getProfile(queryId)` into two parts of  `getProfileContent` and `getProfileBuilder`. 
Since if we want to store the record in a db system and read the big field to construct the `ProfileBuilder` instance, it maybe a heave thing if we only want to list the record or get the query profile content.

But it maybe take a new disadvantage that if we want to add more filed into query profile, it has to add a new method instead of add a filed in `ProfileElemt`. It may cause the number of methods loss control.

Which one do you think is better?

stalary:
I think it's ok to return enough data at the bottom level for the top level to choose. This will be easier to extend later when we support more storage methods.

stalary:
> @stalary In the lastest commit I separate the interface `getProfile(queryId)` into two parts of `getProfileContent` and `getProfileBuilder`. Since if we want to store the record in a db system and read the big field to construct the `ProfileBuilder` instance, it maybe a heave thing if we only want to list the record or get the query profile content.
> 
> But it maybe take a new disadvantage that if we want to add more filed into query profile, it has to add a new method instead of add a filed in `ProfileElemt`. It may cause the number of methods loss control.
> 
> Which one do you think is better?

I think it's ok to return enough data at the bottom level for the top level to choose. This will be easier to extend later when we support more storage methods. So in my opinion, split is not necessary.

SaintBacchus:
OK, I revert the last commit and do the performence improve in the later worker.

PR #9727, opened 2022-05-22
[opt] avoid memcpy when building page

dataroaring:
PageBuilder finishes a page when the page is full, when the page is
full, its size is greater than page size, so the last row would
trigger an memcpy due to there is not enough space in faststring.

The patch fix the problem by 2 changes:
1. PageBuilder reserves extra bytes
   If the last row is less than extra bytes, then no memory copy is
   triggered.
2. faststring uses malloc/realloc/free rather than new/delete
   realloc can avoid memcopy some times.

The patch is not tested, so please do not merge.

# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/util/faststring.cc
+++ b/be/src/util/faststring.cc
@@ -47,11 +47,11 @@ void faststring::GrowArray(size_t newcapacity) {
         }
     }
     capacity_ = newcapacity;
-    data_ = newdata;
     if (data_ == initial_data_) {
         ASAN_POISON_MEMORY_REGION(initial_data_, arraysize(initial_data_));
     }
-    ASAN_POISON_MEMORY_REGION(data_ + len_, capacity_ - len_);
+    data_ = newdata;
+    ASAN_POISON_MEMORY_REGION(data_, capacity_);
 }
 
 void faststring::ShrinkToFitInternal() {
--- a/be/src/util/faststring.cc
+++ b/be/src/util/faststring.cc
@@ -37,18 +37,20 @@ void faststring::GrowToAtLeast(size_t newcapacity) {
 
 void faststring::GrowArray(size_t newcapacity) {
     DCHECK_GE(newcapacity, capacity_);
-    std::unique_ptr<uint8_t[]> newdata(new uint8_t[newcapacity]);
-    if (len_ > 0) {
-        memcpy(&newdata[0], &data_[0], len_);
+    uint8_t* newdata = NULL;
+    if (data_ != initial_data_) {
+        newdata = (uint8_t*)realloc((void*)data_, newcapacity);
+    } else {
+        newdata = (uint8_t*)malloc(newcapacity);
+        if (len_ > 0) {
+            memcpy(newdata, data_, len_);
+        }
     }
     capacity_ = newcapacity;
-    if (data_ != initial_data_) {
-        delete[] data_;
-    } else {
+    data_ = newdata;
+    if (data_ == initial_data_) {
         ASAN_POISON_MEMORY_REGION(initial_data_, arraysize(initial_data_));
     }
-
-    data_ = newdata.release();
     ASAN_POISON_MEMORY_REGION(data_ + len_, capacity_ - len_);
 }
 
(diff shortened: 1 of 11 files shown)


be/src/util/faststring.cc
```diff
@@ -37,18 +37,20 @@ void faststring::GrowToAtLeast(size_t newcapacity) {
 
 void faststring::GrowArray(size_t newcapacity) {
     DCHECK_GE(newcapacity, capacity_);
-    std::unique_ptr<uint8_t[]> newdata(new uint8_t[newcapacity]);
-    if (len_ > 0) {
-        memcpy(&newdata[0], &data_[0], len_);
-    }
-    capacity_ = newcapacity;
+    uint8_t* newdata = NULL;
```
yiguolei, line 40: Not use NULL, use nullptr please.

be/src/util/faststring.cc
```diff
@@ -37,18 +37,20 @@ void faststring::GrowToAtLeast(size_t newcapacity) {
...
+        newdata = (uint8_t*)realloc((void*)data_, newcapacity);
     } else {
+        newdata = (uint8_t*)malloc(newcapacity);
+        if (len_ > 0) {
+            memcpy(newdata, data_, len_);
+        }
+    }
+    capacity_ = newcapacity;
+    data_ = newdata;
+    if (data_ == initial_data_) {
```
yiguolei, line 51: This branch never happens

be/src/util/faststring.cc
```diff
@@ -57,14 +59,14 @@ void faststring::ShrinkToFitInternal() {
     if (len_ <= kInitialCapacity) {
         ASAN_UNPOISON_MEMORY_REGION(initial_data_, len_);
         memcpy(initial_data_, &data_[0], len_);
-        delete[] data_;
```
yiguolei, line 60: I think we'd better use c++ function delete[] not C function free()

be/src/util/faststring.cc
```diff
@@ -57,14 +59,14 @@ void faststring::ShrinkToFitInternal() {
...
         data_ = initial_data_;
         capacity_ = kInitialCapacity;
     } else {
-        std::unique_ptr<uint8_t[]> newdata(new uint8_t[len_]);
-        memcpy(&newdata[0], &data_[0], len_);
-        delete[] data_;
-        data_ = newdata.release();
+        uint8_t* newdata = (uint8_t*)malloc(len_);
+        memcpy(newdata, data_, len_);
+        free(data_);
```
yiguolei, line 68: And here, use delete[] please.

be/src/util/faststring.cc
```diff
@@ -37,18 +37,20 @@ void faststring::GrowToAtLeast(size_t newcapacity) {
...
     } else {
+        newdata = (uint8_t*)malloc(newcapacity);
+        if (len_ > 0) {
+            memcpy(newdata, data_, len_);
+        }
+    }
+    capacity_ = newcapacity;
+    data_ = newdata;
+    if (data_ == initial_data_) {
         ASAN_POISON_MEMORY_REGION(initial_data_, arraysize(initial_data_));
```
yiguolei, line 52: data_ never equal to initial_data_?

be/src/util/faststring.h
```diff
@@ -82,7 +82,7 @@ class faststring {
     OwnedSlice build() {
         uint8_t* ret = data_;
         if (ret == initial_data_) {
-            ret = new uint8_t[len_];
+            ret = (uint8_t*)malloc(len_);
             memcpy(ret, data_, len_);
```
yiguolei, line 86: ownedslice deconstructor using delete[]  to release data_, if we use malloc here, we need to modify it.

be/src/util/faststring.cc
xinyiZzz: initial_data_ is a reserved memory, which is no longer equal to after the first GrowArray.

dataroaring: yep, there is a bug visible to the naked eye.

PR #9754, opened 2022-05-24
[feature] datetime column type support auto-initialized with default …

### LINKED ISSUE #9570: [Feature] datetime column type support auto-initialized  with default CURRENT_TIMESTAMP
### Search before asking

### Description

An auto-initialized column is set to the current timestamp for inserted rows that specify no value for the column.

so when we define a column like ```sync_at DateTime  DEFAULT CURRENT_TIMESTAMP COMMENT "sync time"```

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

BePPPower:
…CURRENT_TIMESTAMP

# Proposed changes

Issue Number: close #9570 

## Problem Summary:

An auto-initialized column is set to the current timestamp for inserted rows that specify no value for the column.
so when we define a column like sync_at DateTime DEFAULT CURRENT_TIMESTAMP COMMENT "sync time"

## Checklist(Required)

## Further comments


BePPPower:
There is still a problem: when we set default CURRENT_TIMESTAMP for varchar/char column type, it also will success.

fe/fe-core/src/main/java/org/apache/doris/analysis/ColumnDef.java
```diff
@@ -61,10 +61,19 @@ public class ColumnDef {
     public static class DefaultValue {
         public boolean isSet;
         public String value;
+        // only used for datetime column type
+        public boolean isCurrentTimestamp;
 
         public DefaultValue(boolean isSet, String value) {
             this.isSet = isSet;
             this.value = value;
+            this.isCurrentTimestamp = false;
```
yiguolei, line 70: Do not use boolean current_timestamp here. Use a expr to save it. Actually, currentTimestamp is function "now". we may add other default value types. Expr is more normal because it could stand for any functions.

yiguolei: when user use a function as default value. We save the function name to value, and the actual function to the expr field.

fe/fe-core/src/main/java/org/apache/doris/catalog/Column.java
```diff
@@ -75,6 +76,8 @@ public class Column implements Writable {
     private boolean isAllowNull;
     @SerializedName(value = "defaultValue")
     private String defaultValue;
+    @SerializedName(value = "isCurrentTimestamp")
+    private boolean isCurrentTimestamp;
     @SerializedName(value = "comment")
```
yiguolei, line 81: 1. private FunctionCallExpr defaultValueExpr; 更通用一些， current timestamp --> now();
2. 持久化： FunctionCallExpr --> writeTo 字节流， read(bytestream)； hll default expr（持久化）

fe/fe-core/src/main/java/org/apache/doris/planner/LoadScanNode.java
```diff
@@ -159,7 +160,11 @@ protected void finalizeParams(Map<String, SlotDescriptor> slotDescByName,
                 } else {
                     Column column = destSlotDesc.getColumn();
                     if (column.getDefaultValue() != null) {
-                        expr = new StringLiteral(destSlotDesc.getColumn().getDefaultValue());
+                        if(column.getType().isDatetime() && column.isCurrentTimestamp()) {
```
yiguolei, line 163: expr --> functioncallexpr.

PR #9771, opened 2022-05-25
[improvement] Show detail status code string for BE http api

jacktengg:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

Failure message for BE http api does not show detail status code:
```
    "status": "Fail",
    "msg": ""
}
```
Improved to:
```
{
    "status": "Not found",
    "msg": "Tablet not found. tablet_id=1202"
}
```

## Checklist(Required)

## Further comments


be/src/util/json_util.cpp
```diff
@@ -32,11 +32,7 @@ std::string to_json(const Status& status) {
     writer.StartObject();
     // status
     writer.Key("status");
-    if (status.ok()) {
```
yiguolei, line 35: Hi jack, I think we could move to_json(Status& status) method to  common/status.h since there is only one method in this file and it is related with common/status.
And also you  could remove json_util_test.h because we could not ensure that the error message always the same with the expected string.

yiguolei (approved):
LGTM

PR #9779, opened 2022-05-25
Add backend_rpc_timeout_second in AdminConfig

pengxiangyu:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

Add backend_rpc_timeout_second in AdminConfig

## Checklist(Required)


fe/fe-core/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -1651,4 +1651,8 @@ public class Config extends ConfigBase {
 
     @ConfField
     public static boolean enable_vectorized_load = false;
+
+    @ConfField(mutable = false, masterOnly = true)
+    public static int backend_rpc_timeout_ms = 60000; // 1 min
```
morningman, line 1656: Please add comment for this config and also modify the config document
`admin-manual/config/fe-config.md`

morningman (approved):
LGTM

PR #9786, opened 2022-05-26
There is Policy which is not on TABLE, Policy should be a base class, and use TablePolicy to extends it.

### LINKED ISSUE #9793: [Feature] There is Policy which is not on TABLE, Policy should be a base class, and use TablePolicy to extends it.
### Search before asking

### Description

Some policy is not on TABLE, such as policy on storage or database.
It is better to set Policy as a base class and create a TablePolicy to extend it.
So that, I can create a policy on storage and extends Policy.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

pengxiangyu:
# Proposed changes

Issue Number: close #9793

## Problem Summary:

Some policy is not on TABLE, such as policy on storage or database.
It is better to set Policy as a base class and create a TablePolicy to extend it.
So that, I can create a policy on storage and extends Policy.

## Checklist(Required)


stalary:
Can you submit an issue describing what you are going to do？

fe/fe-core/src/main/java/org/apache/doris/policy/Policy.java
```diff
@@ -17,106 +17,97 @@
...
-     * PERMISSIVE | RESTRICTIVE, If multiple types exist, the last type prevails.
-     **/
-    @SerializedName(value = "filterType")
-    private final FilterType filterType;
+    @SerializedName(value = "policyName")
+    protected String policyName;
 
-    private Expr wherePredicate;
+    @SerializedName(value = "dbId")
+    protected long dbId;
```
stalary, line 60: Is it possible that stroage is not for DB, but for be nodes?

fe/fe-core/src/main/java/org/apache/doris/policy/PolicyMgr.java
```diff
@@ -68,7 +68,7 @@ public class PolicyMgr implements Writable {
      * Cache merge policy for match.
      * key：dbId:tableId-type-user
      **/
-    private Map<Long, Map<String, Policy>> dbIdToMergePolicyMap = Maps.newConcurrentMap();
+    private Map<Long, Map<String, TablePolicy>> dbIdToMergePolicyMap = Maps.newConcurrentMap();
```
stalary, line 71:
```suggestion
    private Map<Long, Map<String, TablePolicy>> dbIdToMergeTablePolicyMap = Maps.newConcurrentMap();
```

stalary:
Need to rebase.

stalary:
LGTM

fe/fe-core/src/main/cup/sql_parser.cup
```diff
@@ -1354,7 +1354,7 @@ create_stmt ::=
     {:
         RESULT = new CreateSqlBlockRuleStmt(ruleName, properties);
     :}
-    /* row policy */
+    /* table policy */
```
morningman, line 1357: This is indeed a row policy, why change to table policy in comment?

fe/fe-core/src/main/java/org/apache/doris/policy/Policy.java
```diff
@@ -18,79 +18,61 @@
...
 
 /**
  * Save policy for filtering data.
  **/
 @Data
-@AllArgsConstructor
-public class Policy implements Writable, GsonPostProcessable {
-
-    public static final String ROW_POLICY = "ROW";
+public abstract class Policy implements Writable, GsonPostProcessable {
```
morningman, line 46: Modify the comment of this class

fe/fe-core/src/main/java/org/apache/doris/policy/Policy.java
```diff
@@ -18,79 +18,61 @@
...
     /**
-     * PERMISSIVE | RESTRICTIVE, If multiple types exist, the last type prevails.
+     * Use for Serialization/deserialization.
      **/
-    @SerializedName(value = "filterType")
-    private final FilterType filterType;
-
-    private Expr wherePredicate;
+    @SerializedName(value = "originStmt")
+    protected String originStmt;
```
morningman, line 60: Not all policy need to save origin stmt. I think we can move this field to TablePolicy.
For row policy, we save origin stmt only because we can not serialize the filter expr directly.
But for storage policy, there is only key value properties that can be serialized directly.

fe/fe-core/src/main/java/org/apache/doris/policy/Policy.java
```diff
@@ -101,54 +83,42 @@ public static Policy fromCreateStmt(CreatePolicyStmt stmt) throws AnalysisExcept
...
     }
 
-    /**
-     * Read policy from file.
-     **/
-    public static Policy read(DataInput in) throws IOException {
-        String json = Text.readString(in);
-        return GsonUtils.GSON.fromJson(json, Policy.class);
+    protected boolean checkMatched(PolicyTypeEnum type, String policyName) {
+        return (type == null || type.equals(this.type))
```
morningman, line 109: the `type` and `policyname` should not be null.

fe/fe-core/src/main/java/org/apache/doris/policy/TablePolicy.java
```diff
@@ -0,0 +1,166 @@
...
+        if (!(policy instanceof TablePolicy)) {
+            return false;
+        }
+        TablePolicy tablePolicy = (TablePolicy) policy;
+        return checkMatched(tablePolicy.getDbId(), tablePolicy.getTableId(), tablePolicy.getType(),
+                            tablePolicy.getPolicyName(), tablePolicy.getUser());
+    }
+
+    @Override
+    public boolean matchPolicy(DropPolicyLog dropPolicyLog) {
```
morningman, line 157: This method name is strange

morningman:
Currently, Doris has row policy and storage policy.
But they all apply to the table granularity, actually they can both called table policy.
So I think it is not suitable to use `TablePolicy` for row policy.

We can just use an abstract class `Policy` as base class, and `RowPolicy` and `StoragePolicy` as derived classes.
And in `Policy` class, there is `id`, `name`,  `type` fields. and an abstract `matchPolicy()` method which should be implemented in derived class.

fe/fe-core/src/main/java/org/apache/doris/policy/RowPolicy.java
```diff
@@ -0,0 +1,174 @@
...
+        Database database = Catalog.getCurrentCatalog().getDbOrAnalysisException(this.dbId);
+        Table table = database.getTableOrAnalysisException(this.tableId);
+        return Lists.newArrayList(this.policyName, database.getFullName(), table.getName(), this.type.name(),
+                this.filterType.name(), this.wherePredicate.toSql(), this.user.getQualifiedUser(), this.originStmt);
+    }
+
+    /**
+     * Read Table Policy from file.
+     **/
+    public static RowPolicy read(DataInput in) throws IOException {
```
morningman, line 118: This `read` method should be implement in `Policy` abstract class.
And you need to add a `RuntimeTypeAdapterFactory` in `GsonUtils` for derived class.
You can refer to `Resource.java` and its derived class like `SparkResource` and `S3Resource`.

And please and ut for the read/write mothod of RowPolicy, you can refer to `DropInfoTest.java`

morningman (approved):
LGTM

stalary:
LGTM

PR #9792, opened 2022-05-26
[Improvement] optimize  scannode concurrency query performance in vectorized engine.

yiguolei:
# Proposed changes

Issue Number: close #xxx

ssb 3.3 query:
1 FE, 3BE
concurrency = 1, parallel fragment=8,
before:  latency = 600ms
after:  latency = 510ms

concurrency = 10, parallel fragment=1,
before:  latency = 6000ms
after:  latency = 4300ms

1 FE, 1BE
concurrency = 1, parallel fragment=1,
before:  latency = 3200ms
after:  latency = 2200ms

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -1548,8 +1548,6 @@ void OlapScanNode::scanner_thread(OlapScanner* scanner) {
     // (_scan_cpu_timer, the class member) is not destroyed after `_running_thread==0`.
     ThreadCpuStopWatch cpu_watch;
     cpu_watch.start();
-
-    scanner->start_scan_working_timer();
     Status status = Status::OK();
     bool eos = false;
     RuntimeState* state = scanner->runtime_state();
@@ -1680,8 +1678,7 @@ void OlapScanNode::scanner_thread(OlapScanner* scanner) {
         }
     }
 
-    //_scan_cpu_timer->update(cpu_watch.elapsed_time());
-    _scan_cpu_timer->update(scanner->update_scan_working_timer());
+    _scan_cpu_timer->update(cpu_watch.elapsed_time());
     _scanner_wait_worker_timer->update(wait_time);
 
     // The transfer thead will wait for `_running_thread==0`, to make sure all scanner threads won't access class members.
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -143,6 +143,7 @@ void OlapScanNode::_init_counter(RuntimeState* state) {
     _index_load_timer = ADD_TIMER(_segment_profile, "IndexLoadTime_V1");
 
     _scan_timer = ADD_TIMER(_scanner_profile, "ScanTime");
+    _eval_conjuctx_timer = ADD_TIMER(_scanner_profile, "EvalConjuctxTime");
     _scan_cpu_timer = ADD_TIMER(_scanner_profile, "ScanCpuTime");
 
     _total_pages_num_counter = ADD_COUNTER(_segment_profile, "TotalPagesNum", TUnit::UNIT);
@@ -180,6 +181,8 @@ Status OlapScanNode::prepare(RuntimeState* state) {
     // create scanner profile
     // create timer
     _tablet_counter = ADD_COUNTER(runtime_profile(), "TabletCount ", TUnit::UNIT);
+    _scanner_sched_counter = ADD_COUNTER(runtime_profile(), "ScannerSchedCount ", TUnit::UNIT);
+
     _rows_pushed_cond_filtered_counter =
             ADD_COUNTER(_scanner_profile, "RowsPushedCondFiltered", TUnit::UNIT);
     _init_counter(state);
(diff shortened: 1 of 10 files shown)


be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -705,6 +704,9 @@ void SegmentIterator::_vec_init_lazy_materialization() {
     //  _schema.column_ids() stands for storage layer block schema, so it contains delete columnid
     //  we just regard delete column as common pred column here.
     if (_schema.column_ids().size() > pred_column_ids.size()) {
+        if (pred_column_ids.size() > 0) {
```
Gabriel39, line 707: remove for-loop codes below

be/src/exec/olap_scanner.cpp
```diff
@@ -327,6 +327,7 @@ Status OlapScanner::get_batch(RuntimeState* state, RowBatch* batch, bool* eof) {
                 break;
             }
 
+            SCOPED_TIMER(_parent->_eval_conjuctx_timer);
```
HappenLee, line 330: should not do timer each rows, could cause performance problem

be/src/exec/olap_scan_node.cpp
```diff
@@ -689,11 +692,11 @@ Status OlapScanNode::build_scan_key() {
     return Status::OK();
 }
 
-static Status get_hints(TabletSharedPtr table, const TPaloScanRange& scan_range,
-                        int block_row_count, bool is_begin_include, bool is_end_include,
-                        const std::vector<std::unique_ptr<OlapScanRange>>& scan_key_range,
-                        std::vector<std::unique_ptr<OlapScanRange>>* sub_scan_range,
-                        RuntimeProfile* profile) {
+Status OlapScanNode::get_hints(TabletSharedPtr table, const TPaloScanRange& scan_range,
```
HappenLee, line 695: why change the code to not static

be/src/olap/bloom_filter_predicate.h
```diff
@@ -158,6 +160,16 @@ void BloomFilterColumnPredicate<T>::evaluate(vectorized::IColumn& column, uint16
             new_size += _specific_filter->find_olap_engine(cell_value);
         }
     }
+    // If the pass rate is very high, for example > 50%, then the bloomfilter is useless.
+    // Some bloomfilter is useless, for example ssb 4.3, it consumes a lot of cpu but it is
+    // useless.
+    _evaluated_rows += *size;
+    _passed_rows += new_size;
+    if (_evaluated_rows > 100) {
```
HappenLee, line 168: `100` should be a config of BE

be/src/olap/column_predicate.h
```diff
@@ -91,6 +91,9 @@ class ColumnPredicate {
 protected:
     uint32_t _column_id;
     bool _opposite;
+    mutable uint64_t _evaluated_rows = 1;
```
HappenLee, line 94: the value only use in bloom filter predicate. it should not define here?

be/src/vec/exec/volap_scan_node.cpp
```diff
@@ -554,8 +576,11 @@ Block* VOlapScanNode::_alloc_block(bool& get_free_block) {
 int VOlapScanNode::_start_scanner_thread_task(RuntimeState* state, int block_per_scanner) {
     std::list<VOlapScanner*> olap_scanners;
     int assigned_thread_num = _running_thread;
-    size_t max_thread = std::min(_volap_scanners.size(),
-                                 static_cast<size_t>(config::doris_scanner_thread_pool_thread_num));
+    size_t max_thread = config::doris_scanner_queue_size;
```
HappenLee, line 579: If this code change is not sure to be effective, I think it can not be modified first

yiguolei: The previous logic is wrong, For example, if assiged thread num = 13 and volap_scanner.size = 5.
thread_slot_num = std::min(thread_slot_num, max_thread - assigned_thread_num);
                if (thread_slot_num <= 0) {
                    thread_slot_num = 1;
                }
This logic will set thread_slot_num = 1. Will not put as many as task to queue.

be/src/olap/column_predicate.h
yiguolei: Yes. My mistake.

be/src/exec/olap_scanner.cpp
yiguolei: my mistake, it should be removed.

be/src/exec/olap_scan_node.cpp
yiguolei: I do not like static method. I think we should use member function as much as possible. And I will use it in volapscannode. It is only defined in olap_scan_node.cpp .

be/src/vec/exec/volap_scan_node.cpp
```diff
@@ -554,8 +575,11 @@ Block* VOlapScanNode::_alloc_block(bool& get_free_block) {
...
+    if (config::doris_scanner_row_num > state->batch_size()) {
```
morningman, line 579: This logic need refine. Actually, no one would change this `doris_scanner_row_num` config, and this `if` condition is always true. So that this `if` condition is meaningless.

yiguolei: I think we need this if condition check. If we do not allow anyone modify config:: doris_scanner_row_num, I think it should be a const expr variable not in config file. If it is in conf file, then it maybe modified by someone, so that we should check it.
And also, maybe nobody will modify doris_scanner_row_num, but somebody could modify batch size, the batch size maybe larger than the config::doris_scanner_row_num.

morningman (approved):
LGTM

PR #9805, opened 2022-05-27
[fix](third-party) Fix build issues of third-party in ubuntu docker

compiletheworld:
Fixes:
1. Fully static linkage of protoc may lead to core dump when using it.
2. Building libmyclient may fail due to miss-used option `-v` of `make`
3. Building lzma implicitly requires autopoint
4. Building xml2 implicitly requires pkg-config


dataroaring (approved):
LGTM

morningman (approved):
LGTM

PR #9815, opened 2022-05-27
[chore] fix some compatible problems on old linux distributions

### LINKED ISSUE #9845: Some compatible problems on old linux distributions when build thirdparty
1. On some old linux distributions such as centos 6，when build arrow， cmake may not find boost. If we sepcify the boost root, new   linux distributions will not use that macro and print some warning
2. If protoc is not build  static， the build time glibc may be higher than runtime， cause protoc cannot run this is introduced by #9805 
image

yangzhg:
# Proposed changes

Issue Number: close #9845

## Problem Summary:
 fix some compatible problems on old linux distributions

## Checklist(Required)

## Further comments

### DIFF
--- a/thirdparty/build-thirdparty.sh
+++ b/thirdparty/build-thirdparty.sh
@@ -312,14 +312,17 @@ build_protobuf() {
     LDFLAGS="-L${TP_LIB_DIR} -static-libstdc++ -static-libgcc -Wl,--undefined=pthread_create" \
     ./configure --prefix=${TP_INSTALL_DIR} --disable-shared --enable-static --with-zlib=${TP_INSTALL_DIR}/include
 
-    # ATTN: If protoc is built fully statically with clang-14 on ubuntu,
-    #       it may core dump when parse some protobuf source files.
-    #       The root cause remains unknown, but it is possibly related glibc...
-    #       And, we don't actually need to be that static,
-    #       dyn-linking to glibc and etc. is OK
-    # cd src
-    # sed -i 's/^AM_LDFLAGS\(.*\)$/AM_LDFLAGS\1 -all-static/' Makefile
-    # cd -
+    # ATTN: If protoc is not built fully statically the linktime libc may newer than runtime.
+    #       This will casue protoc cannot run
+    #       If you really need to dynamically link protoc, please set the environment variable DYN_LINK_PROTOC=1
+
+    if [[ "${DYN_LINK_PROTOC}" == "1" ]]; then
+        echo "link protoc dynamiclly"
+    else 
+        cd src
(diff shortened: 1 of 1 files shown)


morningman:
Hi @yangzhg , could you create an issue to describe the problem?

thirdparty/build-thirdparty.sh
```diff
@@ -312,14 +312,17 @@ build_protobuf() {
...
 
-    # ATTN: If protoc is built fully statically with clang-14 on ubuntu,
-    #       it may core dump when parse some protobuf source files.
-    #       The root cause remains unknown, but it is possibly related glibc...
-    #       And, we don't actually need to be that static,
-    #       dyn-linking to glibc and etc. is OK
-    # cd src
-    # sed -i 's/^AM_LDFLAGS\(.*\)$/AM_LDFLAGS\1 -all-static/' Makefile
-    # cd -
+    # ATTN: If protoc is not built fully statically the linktime libc may newer than runtime.
```
compiletheworld, line 315: Would you please create an issue associated with this pull request?
I would like to know the dyn-link info of protoc in your env., info like

1. `which ld`
2. `ldd protoc`
3. `ldd --version`

Thanks.

yangzhg: using the ldb_toochain 0.9

yangzhg: this is not a problem with ldd, it is because of higher  version gcc or clang will link glibc new than 2.12， and for old version linux such as centos 6 only contain glibc  lower than 2.12

HappenLee (approved):
LGTM

thirdparty/build-thirdparty.sh
compiletheworld: > this is not a problem with ldd, it is because of higher version gcc or clang will link glibc new than 2.12， and for old version linux such as centos 6 only contain glibc lower than 2.12

@yangzhg
Sorry, I didn't make it clear, 
1. Command `which ld` is for the linker info, in your env. it may be  `/usr/bin/ld` or `${path_to_ldb_toolchain}/bin/ld` 
2. Command `ldd protoc` is for the dyn-linkage info of protoc, the result of linking
3. Command `ldd --version` is for version checking of glibc of you env. 

With the given info, it seems that protoc was built using the system's default linker `/usr/bin/ld` instead of `${path_to_ldb_toolchain}/bin/ld`? 

To confirm that, you can check out `thirdparty/src/protobuf-${version}/config.log`,  by `grep`ing this line "checking if the linker",  for more info. about how protoc was built and linked.

**Your patch works**, however, we can figure out why that version messed up if you are interested in digging config.log.

BTW, we can also try to fix the wrong version of dyn-linked glibc by modifying protoc binary
```
patchelf --replace-needed libc.so.6 ${path_to_ldb_toolchain}/lib/libc.so.6 protoc
```

PR #9824, opened 2022-05-28
[Bugfix(Vec)] Fix some memory leak issues

amosbird:
## Problem Summary:

Fix some memory leak issues in vectorized engine.

## Checklist(Required)

## Further comments

The following leak issues are catched by running stress tests with tpch queries.

1. `HashJoinNode` doesn't close owned expression contexts.
2. `VExprContext` doesn't free owned function contexts.
3. `AggregationNode` doesn't close owned aggregate_evaluators. Also it might fail to close properly.
4. merging `VExchangeNode` doesn't close owned vsort_exec_exprs.
5. `VSortNode` doesn't return error code if close fails.

There is also a related fix proposed in https://github.com/apache/incubator-doris/pull/9820 whose implementation differs from `ExprContext`. I think deallocation suits better inside d'tor, which is more consistent.

### DIFF
--- a/be/src/vec/exec/join/vhash_join_node.cpp
+++ b/be/src/vec/exec/join/vhash_join_node.cpp
@@ -820,6 +820,8 @@ Status HashJoinNode::close(RuntimeState* state) {
         return Status::OK();
     }
 
+    VExpr::close(_build_expr_ctxs, state);
+    VExpr::close(_probe_expr_ctxs, state);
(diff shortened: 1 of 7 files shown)


HappenLee (dismissed):
LGTM

amosbird:
Fix another case: `VAnalyticEvalNode` doesn't open and close properly.

morningman:
hi @amosbird , please rebase master to fix BE ut bug.

morningman (approved):
LGTM

PR #9842, opened 2022-05-30
[feature] global dict optimization for low cardinality string column

### LINKED ISSUE #9242: [Global Dict]  Using global dict to accelerate low cardinality query performance
### Search before asking

### Description

In Doris, many dimension columns are string type with low cardinality (cardinality < 1024), such as some status codes, provinces, regions, genders, etc. During group by or join operation, it has to make a hash table in the form of strings. The efficiency is relatively low. Through performance testing, it is found that the performance is about 3 times worse than using int type. If we can convert string to int through a global dictionary during a query, then the query efficiency will be greatly improved. The solution is described below.

### Design:

Maintain global dictionary for every column in FE as needed, for example it could be organized as map<(table_id,col_id), dict> in DictManager.
When creating a table, set the attribute of the field to low_cardinality to identify whether the field is a low-cardinality field. For low-cardinality fields, use a global dictionary to speed up the corresponding query. The specific syntax is as follows:

```
// create table
create table table_name (
    col1 int,
    col2 string|varchar(10) low_cardinality
);

// alter table
alter table table_name modify column col2 set low_cardinality=true

```

At the beginning, the dict is empty, so it will not be used during query. FE will start a background task to obtain global dictionary information from BE for every low cardinality column, and the task is sent to BE in the form of SQL, as follows:

`select dict(col) from table[meta]`

So that there will be a new aggregation method dict and new scannode in BE. In the new scannode, BE reads the local dictionary of each field and get all dictinct strings. dict is an aggregation function that deduplicates and merges the dictionaries of multiple BEs and tablets. The final result is saved in FE's DictManager.

During query, FE puts the global dictionary information into the Fragment structure and sends it to each BE. The structure is as follows:

```
struct TStringColumnDict {
list<int16> id,    // the offset or id of the string
list<string> value    // the string value
}
struct TColumnDict {
TStringColumnDict xxx,
TDateColumnDict xxx,
....
}
struct TGlobalDict {
map<id, TColumnDict> columnDicts // map from column id to column dict
}
struct TExecPlanFragmentParams {
...
TGlobalDict globalDict
}
```

The OlapScanNode will encodes the string data according to the value and ID of each string in the dict, and the subsequent calculations (such as group by) are operated according to int, and at the final result sink node, the result is decode into the corresponding string.

### Maintain GlobalDict Changes

During process of loading, user will import new strings that are not in the GlobalDict, so this requires an update mechanism for the global dictionary. Stream load is actually a query, so it will also generate a fragment, and according to the task description of our query, there is a global dict in the fragment. In TabletSink, each value of the column of low cardinality should be evaluated with ColumnDict, if find a new string, the coordinator need to inform FE that the dict of this column is invalid. Specifically, we need to make some modifications in LoadTxnCommitRequest and add a field to indicate which columns are invalid.

```
struct TLoadTxnCommitRequest {
...
list<int> invalid_col_ids
}
```

Invalid col ids should be persisted to the transaction state. Once the transaction is published, we will delete the dict from the dict manager. In this way, the query will not use this dictionary optimization in the future. At the same time, FE generates a task to get the latest Dict. When the latest Dict is obtained, the query can be optimized using Dict again.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

starocean999:
# Proposed changes

Issue Number: close [#9242](https://github.com/apache/incubator-doris/issues/9242), you can also find more detail about the problem to solve and architecture design

## Problem Summary:

Group by operator is kind of time cosuming, especially for string column. However, some string column are low cardinality and can be dict encoded. We can take advantage of the dict encoded integer values in group by operator instead of the original strings to get much better performance.

## Checklist(Required)

## Further comments

add low_cardinality keyword. use "alter table xxx modify column yyy low_cardinality true;" to set the column yyy as low cardinality. The the column would be dict encoded and in group by operator, we can use integer dict code instead of strings.

Use ssb_flat 100G as an example. We add low_cardinality key word for some string columns and run test sqls as bellow:

alter table lineorder_flat modify column lo_orderpriority low_cardinality true;
alter table lineorder_flat modify column lo_shipmode low_cardinality true;
alter table lineorder_flat modify column lo_orderdate low_cardinality true;
alter table lineorder_flat modify column s_city low_cardinality true;
alter table lineorder_flat modify column c_city low_cardinality true;
alter table lineorder_flat modify column s_nation low_cardinality true;
alter table lineorder_flat modify column s_region low_cardinality true;
alter table lineorder_flat modify column c_nation low_cardinality true;
alter table lineorder_flat modify column p_category low_cardinality true;
alter table lineorder_flat modify column p_mfgr low_cardinality true;

alter table lineorder modify column lo_orderpriority low_cardinality true;
alter table lineorder modify column lo_shipmode low_cardinality true;
alter table lineorder modify column lo_orderdate low_cardinality true;
alter table supplier modify column s_city low_cardinality true;
alter table customer modify column c_city low_cardinality true;
alter table supplier modify column s_nation low_cardinality true;
alter table supplier modify column s_region low_cardinality true;
alter table customer modify column c_nation low_cardinality true;
alter table part modify column p_category low_cardinality true;
alter table part modify column p_mfgr low_cardinality true;

--Q1
select count(*),lo_shipmode from lineorder_flat group by lo_shipmode;
--Q2
select count(distinct lo_shipmode) from lineorder_flat;
--Q3
select count(*),lo_shipmode,lo_orderpriority from lineorder_flat group by lo_shipmode,lo_orderpriority;
--Q4
select count(*),lo_shipmode,lo_orderpriority from lineorder_flat group by lo_shipmode,lo_orderpriority,lo_shippriority;
--Q5
select count(*),lo_shipmode,s_city from lineorder_flat group by lo_shipmode,s_city;
--Q6
select count(*) from lineorder_flat group by c_city,s_city;
--Q7
select count(*) from lineorder_flat group by lo_shipmode,lo_orderdate;
--Q8
select count(*) from lineorder_flat group by lo_orderdate,s_nation,s_region;
--Q9
select count(*) from lineorder_flat group by c_city,s_city,c_nation,s_nation;
--Q10
select count(*) from (select count(*) from lineorder_flat group by lo_shipmode,lo_orderpriority,p_category,s_nation,c_nation) t;
--Q11
select count(*) from (select count(*) from lineorder_flat group by lo_shipmode,lo_orderpriority,p_category,s_nation,c_nation,p_mfgr) t;
--Q12
select count(*) from (select count(*) from lineorder_flat group by substr(lo_shipmode,2),lower(lo_orderpriority),p_category,s_nation,c_nation,s_region,p_mfgr) t;

The test results are(in ms):
query | low_cardinality=false | low_cardinality=true
-- | -- | --
q1 | 707 | 367
q2 | 694 | 343
q3 | 1206 | 641
q4 | 1384 | 727
q5 | 1502 | 779
q6 | 7218 | 1993
q7 | 1337 | 753
q8 | 3264 | 1711
q9 | 12868 | 2703
q10 | 17895 | 7611
q11 | 20533 | 8654
q12 | 24342 | 14983


gensrc/thrift/PaloInternalService.thrift
```diff
@@ -278,15 +278,17 @@ struct TTxnParams {
...
+//   1: optional Types.TPrimitiveType type
+//   2: list<string> str_dict  // map one string to a integer, using offset as id
+// }
 
-struct TGlobalDict {
-  1: optional map<i32, TColumnDict> dicts,  // map dict_id to column dict
-  2: optional map<i32, i32> slot_dicts // map from slot id to column dict id, because 2 or more column may share the dict
-}
+// ExecPlanFragment
+
```
yiguolei, line 287: 这种代码，如果删除了就删除好了，不要这种注释了，以后还得有人清理

fe/pom.xml
```diff
@@ -953,6 +954,11 @@ under the License.
                 <artifactId>mariadb-java-client</artifactId>
                 <version>${mariadb-java-client.version}</version>
             </dependency>
+			<dependency>
+	            <groupId>org.apache.arrow</groupId>
```
yiguolei, line 958: This is useless any more. Delete it.

be/src/olap/delta_writer.cpp
```diff
@@ -320,6 +321,41 @@ Status DeltaWriter::close_wait() {
...
+        }
+
+        auto beta_rowset = dynamic_cast<BetaRowset*>(_cur_rowset.get());
+        assert(beta_rowset);
+        std::set<std::string> dict_words;
+        for (const TabletColumn* column : dict_columns) {
+            auto iter = _req.dicts->find(column->name().data());
+            assert(iter != _req.dicts->end());
+            dict_words.clear();
+            Status status = beta_rowset->get_dict_data(dict_words, column->unique_id());
```
yiguolei, line 346: If there are too many new dict words in column, get dict data may failed to get dict and return error status. If you return error status here, the load task is failed.

yiguolei: I think get dict data could not affect normal load job. Maybe you could just put the column to invalid columns if failed to get dict data.

be/src/olap/rowset/beta_rowset.cpp
```diff
@@ -230,4 +230,14 @@ bool BetaRowset::check_file_exist() {
     return true;
 }
 
+Status BetaRowset::get_dict_data(std::set<std::string>& dict_words, int col_id) {
+    std::vector<segment_v2::SegmentSharedPtr> segments;
+    RETURN_NOT_OK(load_segments(&segments));
+    for (auto& seg_ptr : segments) {
+        Status status = seg_ptr->get_dict_data(dict_words, col_id);
+        if (!status.ok()) return status;
```
yiguolei, line 238: use macro RETURN_NOT_OK(status)

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -320,10 +320,9 @@ Status BetaRowsetWriter::_create_segment_writer(
     }
 
     DCHECK(wblock != nullptr);
```
yiguolei, line 322: need not change this code, it does not matter whether it is a local variable or a member variable.

be/src/olap/rowset/beta_rowset_writer.h
```diff
@@ -100,6 +101,8 @@ class BetaRowsetWriter : public RowsetWriter {
 
     bool _is_pending = false;
     bool _already_built = false;
+
+    segment_v2::SegmentWriterOptions _writer_options;
```
yiguolei, line 105: useless member variable.

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -375,6 +375,79 @@ Status ColumnReader::new_iterator(ColumnIterator** iterator) {
...
+    RETURN_IF_ERROR(_ensure_index_loaded());
+    OrdinalPageIndexIterator last_iter(_ordinal_index.get(), _ordinal_index->num_data_pages() - 1);
+    PageHandle handle;
+    Slice page_body;
+    PageFooterPB footer;
+    iter_opts.type = DATA_PAGE;
+    ParsedPage parsed_page;
+    RETURN_IF_ERROR(read_page(iter_opts, last_iter.page(), &handle, &page_body, &footer, codec));
+    // parse data page
+    RETURN_IF_ERROR(ParsedPage::create(std::move(handle), page_body, footer.data_page_footer(),
```
yiguolei, line 396: Are you sure this code could compiled? This line will return error status not boolean.

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -375,6 +375,79 @@ Status ColumnReader::new_iterator(ColumnIterator** iterator) {
...
+        if (all_pages_encoded_by_dict(iter_opts, compress_codec.get())) {
+            PageHandle dict_page_handle;
+            PageFooterPB dict_footer;
+            read_page(iter_opts, get_dict_page_pointer(), &dict_page_handle, &dict_slice,
+                      &dict_footer, compress_codec.get());
+            auto dict_decoder = std::make_unique<BinaryPlainPageDecoder>(dict_slice);
+            RETURN_IF_ERROR(dict_decoder->init());
+
+            auto* pd_decoder = (BinaryPlainPageDecoder*)dict_decoder.get();
+            StringRef* dict_word_info = new StringRef[pd_decoder->_num_elems];
```
yiguolei, line 430: using unique_ptr instead of raw ptr to avoid forgetting delete[]

be/src/olap/tablet.cpp
```diff
@@ -1531,4 +1532,24 @@ std::shared_ptr<MemTracker>& Tablet::get_compaction_mem_tracker(CompactionType c
     }
 }
 
+Status Tablet::get_dict_data(std::set<std::string>& dict_words, int col_id) {
+    std::vector<BetaRowset*> beta_row_sets;
```
yiguolei, line 1536: Do not use raw ptr here. Because if you use raw ptr, the rowset maybe released before you use it. May core!!!!

be/src/olap/tablet.cpp
```diff
@@ -1531,4 +1532,24 @@ std::shared_ptr<MemTracker>& Tablet::get_compaction_mem_tracker(CompactionType c
...
+    {
+        std::shared_lock rdlock(get_header_lock());
+        for (auto entry : _rs_version_map) {
```
yiguolei, line 1539: std::shared_lock rdlock(get_header_lock());
        for (auto entry : _rs_version_map) {
       RETURN_IF_NOT_OK( rs->get_dict_data(dict_words, col_id));
            }

be/src/runtime/descriptors.h
```diff
@@ -140,6 +146,8 @@ class SlotDescriptor {
     int _field_idx;
 
     const bool _is_materialized;
+    bool _is_global_dict_column = false;
```
yiguolei, line 149: rename it to  has_global_dict

be/src/runtime/descriptors.cpp
```diff
@@ -90,7 +90,11 @@ void SlotDescriptor::to_protobuf(PSlotDescriptor* pslot) const {
 vectorized::MutableColumnPtr SlotDescriptor::get_empty_mutable_column() const {
     auto data_type = get_data_type_ptr();
     if (data_type) {
-        return data_type->create_column();
+        auto column_ptr = data_type->create_column();
+        if (is_global_dict_column()) {
```
yiguolei, line 94: rename to has_global_dict()

be/src/runtime/runtime_state.cpp
```diff
@@ -467,4 +467,38 @@ int64_t RuntimeState::get_load_mem_limit() {
...
+        for (auto it = tglobal_dict->slot_dicts.begin(); it != tglobal_dict->slot_dicts.end();
+             it++) {
+            assert(_dict_id_to_global_dict_map.find(it->second) !=
+                   _dict_id_to_global_dict_map.end());
+            _slot_id_to_global_dict_map[it->first] = _dict_id_to_global_dict_map[it->second];
+        }
+    }
+}
+
+vectorized::GlobalDictSPtr RuntimeState::get_global_dict(int slot_id) {
```
yiguolei, line 485: GlobalDictsPtr

be/src/olap/rowset/segment_v2/options.h
```diff
@@ -18,6 +18,7 @@
 #pragma once
 
 #include <cstddef>
+#include <parallel_hashmap/phmap.h>
```
yiguolei, line 21: useless change

be/src/runtime/runtime_state.cpp
```diff
@@ -467,4 +467,38 @@ int64_t RuntimeState::get_load_mem_limit() {
...
+    assert(_slot_id_to_global_dict_map.find(slot_id) != _slot_id_to_global_dict_map.end());
+    return _slot_id_to_global_dict_map[slot_id];
+}
+
+vectorized::GlobalDictSPtr RuntimeState::find_global_dict(int slot_id) {
```
yiguolei, line 490: find global dict is duplicated with get_global_dict please merge these 2 method to 1

be/src/runtime/tablets_channel.cpp
```diff
@@ -234,6 +250,7 @@ Status TabletsChannel::_open_all_writers(const PTabletWriterOpenRequest& request
         wrequest.tuple_desc = _tuple_desc;
         wrequest.slots = index_slots;
```
yiguolei, line 251: _dicts is constructed from slots, but wrequest already has a field slots so that I think the node could parse dict from wrequest's slots field. No need to create dicts here.

be/src/vec/exec/volap_scan_node.cpp
```diff
@@ -34,6 +34,32 @@ VOlapScanNode::VOlapScanNode(ObjectPool* pool, const TPlanNode& tnode, const Des
     _free_blocks.reserve(_max_materialized_blocks);
 }
 
+Status VOlapScanNode::init(const TPlanNode& tnode, RuntimeState* state) {
+    assert(state);
```
yiguolei, line 38: Do not need to check here. If state is not set, then many method will crashed. We do not need to check every condition.

be/src/vec/exec/volap_scan_node.h
```diff
@@ -48,6 +49,7 @@ class VOlapScanNode final : public OlapScanNode {
     Status _add_blocks(std::vector<Block*>& block);
     int _start_scanner_thread_task(RuntimeState* state, int block_per_scanner);
     Block* _alloc_block(bool& get_free_block);
+    Status _do_dict_encode(Block& block) const;
```
yiguolei, line 52: This method is useless?

be/src/vec/exec/volap_scan_node.h
```diff
@@ -69,6 +71,8 @@ class VOlapScanNode final : public OlapScanNode {
     int _max_materialized_blocks;
 
     size_t _block_size = 0;
+
+    std::map<int, GlobalDictSPtr> _dicts;
```
yiguolei, line 75: Is this field useful?

be/src/vec/runtime/dict/dict_mgr.h
```diff
@@ -0,0 +1,80 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
```
yiguolei, line 4: This file is useless, please do not commit it.

be/src/vec/runtime/dict/global_dict.cpp
```diff
@@ -0,0 +1,159 @@
...
+GlobalDict::GlobalDict(const std::vector<std::string>& data) : Dict(data) {
+    _dict_data = data;
+    for (const auto& val : _dict_data) {
+        StringValue v {val.data(), (int)val.size()};
+        insert_value(v);
+    }
+}
+
+bool GlobalDict::encode(ColumnWithTypeAndName& col) {
+    // encode method is only used by unit test
```
yiguolei, line 36: Is encode method ever used?

be/src/vec/columns/column_dict_encoded_string.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include "vec/columns/column_vector.h"
+#include "vec/common/assert_cast.h"
+
+using DictId = int32_t;
+
+namespace doris::vectorized {
+
+template <typename T>
+class ColumnDictEncodedString final : public ColumnVector<T> {
```
yiguolei, line 28: Is this class useful?

PR #9856, opened 2022-05-30
[feature-wip](array-type) Add array type support for vectorized parquet-orc scanner

yinzhijian:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

Only support one level array now.
for example:
- nullable(array(nullable(tinyint))) is **support**.
- nullable(array(nullable(array(xx))) is **not support**.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/vec/data_types/data_type_factory.hpp
+++ b/be/src/vec/data_types/data_type_factory.hpp
@@ -49,30 +49,64 @@ public:
         static std::once_flag oc;
         static DataTypeFactory instance;
         std::call_once(oc, []() {
-            instance.register_data_type("UInt8", std::make_shared<DataTypeUInt8>());
-            instance.register_data_type("UInt16", std::make_shared<DataTypeUInt16>());
-            instance.register_data_type("UInt32", std::make_shared<DataTypeUInt32>());
-            instance.register_data_type("UInt64", std::make_shared<DataTypeUInt64>());
-            instance.register_data_type("Int8", std::make_shared<DataTypeInt8>());
-            instance.register_data_type("Int16", std::make_shared<DataTypeInt16>());
-            instance.register_data_type("Int32", std::make_shared<DataTypeInt32>());
-            instance.register_data_type("Int64", std::make_shared<DataTypeInt64>());
-            instance.register_data_type("Int128", std::make_shared<DataTypeInt128>());
-            instance.register_data_type("Float32", std::make_shared<DataTypeFloat32>());
-            instance.register_data_type("Float64", std::make_shared<DataTypeFloat64>());
-            instance.register_data_type("Date", std::make_shared<DataTypeDate>());
-            instance.register_data_type("DateTime", std::make_shared<DataTypeDateTime>());
-            instance.register_data_type("String", std::make_shared<DataTypeString>());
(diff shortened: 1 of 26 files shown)


be/src/vec/data_types/data_type_factory.hpp
```diff
@@ -49,22 +49,33 @@ class DataTypeFactory {
...
+            };
+            for (auto const& [key, val] : base_type_map) {
+                instance.register_data_type(key, val);
+                instance.register_data_type("Array(" + key + ")",
+                                            std::make_shared<vectorized::DataTypeArray>(val));
+                instance.register_data_type(
+                        "Array(Nullable(" + key + "))",
+                        std::make_shared<vectorized::DataTypeArray>(
+                                std::make_shared<vectorized::DataTypeNullable>(val)));
+            }
```
adonis0147, line 78: This modification may be unnecessary.

yinzhijian: 1. This map is used to convert string structures such as: (Array(Nullable(Varchar))) to a DataType structure, which is not the same as create_data_type

2. Finding and returning pre-created structures from the map is currently the most efficient way to do String <=> DataType

be/src/vec/functions/function_cast.h
```diff
@@ -1040,6 +1073,69 @@ class FunctionCast final : public IFunctionBase {
...
+        const auto* from_type = check_and_get_data_type<DataTypeArray>(from_type_untyped.get());
+
+        if (!from_type) {
+            LOG(FATAL) << "CAST AS Array can only be performed between same-dimensional Array, "
+                          "String types";
+        }
+
+        DataTypePtr from_nested_type = from_type->get_nested_type();
+
+        /// In query SELECT CAST([] AS Array(Array(String))) from type is Array(Nothing)
```
HappenLee, line 1092: here may cast `[Array String]` to `[Array Int]`, but the `string` cast to `int` may lead to `NULL`, how to handle it?

be/src/vec/utils/arrow_column_to_doris_column.cpp
```diff
@@ -236,19 +237,51 @@ static Status convert_column_with_decimal_data(const arrow::Array* array, size_t
...
+    size_t num_of_data = 0;
+    // get start idx and num of values from arrow offsets
+    RETURN_IF_ERROR(convert_offset_from_list_column(array, array_idx, data_column, num_elements,
+                                                    &start_idx_of_data, &num_of_data));
+    auto& data_column_ptr = static_cast<ColumnArray&>(*data_column).get_data_ptr();
+    auto concrete_array = down_cast<const arrow::ListArray*>(array);
+    std::shared_ptr<arrow::Array> arrow_data = concrete_array->values();
+
+    return arrow_column_to_doris_column(arrow_data.get(), start_idx_of_data, data_column_ptr,
+                                        num_of_data, timezone);
```
HappenLee, line 269: do offset valid in this function，do not affects other processing logic which no need concern the arrow offset

HappenLee (dismissed):
LGTM

HappenLee (approved):
LGTM

PR #9864, opened 2022-05-30
[Enhancement] Support select specify tablet id

xinyiZzz:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

select specifies the tablet id for data sampling, for example:
```
SELECT * FROM db1.table1 TABLET(10031,10032,10033)
```

## Checklist(Required)

## Further comments


caiconghui:
better not providing select tablet id to normal user，normal user only need to know about table，which make doris easy and simple. I think admin can select * from tablet id directly for debug，but for normal user，tablet id is invisible

xinyiZzz:
>The tablet id is unique in doris, If we specify tablet id here, why do we need table? since we could get table using tabletid.

> better not providing select tablet id to normal user，normal user only need to know about table，which make doris easy and simple. I think admin can select * from tablet id directly for debug，but for normal user，tablet id is invisible

`select tablet` is normally not exposed to the user, which is a sub-function of data sampling.

I will submit this api in a unity pr.

PR #9887, opened 2022-05-31
[Enhancement] add some metrics for cpu and memory

carlvinhust2012:
# Proposed changes
1. add some metrics for cpu monitor;
2. add metrics for process state monitor;
3. add metrics for memory monitor;

It is convenient for us to use grafana to filter through different conditions.

After the added, we can find the cpu metrics like this：
doris_be_cpu{device="cpu1",mode="guest_nice"} 0
doris_be_cpu{device="cpu1",mode="guest"} 0
doris_be_cpu{device="cpu1",mode="steal"} 0
doris_be_cpu{device="cpu1",mode="soft_irq"} 107168
doris_be_cpu{device="cpu1",mode="irq"} 0
doris_be_cpu{device="cpu1",mode="iowait"} 3726931
doris_be_cpu{device="cpu1",mode="idle"} 2358039214
doris_be_cpu{device="cpu1",mode="system"} 58699464
doris_be_cpu{device="cpu1",mode="nice"} 1700438
doris_be_cpu{device="cpu1",mode="user"} 54974091

we can find the memory metrics as follow：
doris_be_memory_pswpin 167785
doris_be_memory_pswpout 203724
doris_be_memory_pgpgin 22308762092
doris_be_memory_pgpgout 152101956232

we also can find the process metrics as follow:
doris_be_proc{mode="interrupt"} 421721020416
doris_be_proc{mode="ctxt_switch"} 2806640907317
doris_be_proc{mode="procs_running"} 8
doris_be_proc{mode="procs_blocked"} 3

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

5. Has unit tests been added: (No)
6. Has document been added or modified: (No)

## Further comments

### DIFF
--- a/be/src/util/system_metrics.cpp
+++ b/be/src/util/system_metrics.cpp
@@ -88,14 +88,26 @@ struct CpuMetrics {
 #define DEFINE_MEMORY_GAUGE_METRIC(metric, unit) \
     DEFINE_GAUGE_METRIC_PROTOTYPE_2ARG(memory_##metric, unit);
 DEFINE_MEMORY_GAUGE_METRIC(allocated_bytes, MetricUnit::BYTES);
+DEFINE_MEMORY_GAUGE_METRIC(pgpgin, MetricUnit::NOUNIT);
+DEFINE_MEMORY_GAUGE_METRIC(pgpgout, MetricUnit::NOUNIT);
+DEFINE_MEMORY_GAUGE_METRIC(pswpin, MetricUnit::NOUNIT);
+DEFINE_MEMORY_GAUGE_METRIC(pswpout, MetricUnit::NOUNIT);
 
 struct MemoryMetrics {
     MemoryMetrics(MetricEntity* ent) : entity(ent) {
         INT_GAUGE_METRIC_REGISTER(entity, memory_allocated_bytes);
+        INT_GAUGE_METRIC_REGISTER(entity, memory_pgpgin);
+        INT_GAUGE_METRIC_REGISTER(entity, memory_pgpgout);
+        INT_GAUGE_METRIC_REGISTER(entity, memory_pswpin);
+        INT_GAUGE_METRIC_REGISTER(entity, memory_pswpout);
     }
 
     MetricEntity* entity = nullptr;
     IntGauge* memory_allocated_bytes;
+    IntGauge* memory_pgpgin;
+    IntGauge* memory_pgpgout;
+    IntGauge* memory_pswpin;
+    IntGauge* memory_pswpout;
 };
 
 #define DEFINE_DISK_COUNTER_METRIC(metric, unit) \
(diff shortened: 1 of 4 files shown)


be/src/util/system_metrics.cpp
```diff
@@ -718,4 +788,174 @@ void SystemMetrics::get_max_net_traffic(const std::map<std::string, int64_t>& ls
...
+
+void SystemMetrics::_install_proc_metrics(MetricEntity* entity) {
+    _proc_metrics.reset(new ProcMetrics(entity));
+}
+
+void SystemMetrics::_update_proc_metrics() {
+#ifdef BE_TEST
+    FILE* fp = fopen(k_ut_stat_path, "r");
+#else
+    FILE* fp = fopen("/proc/stat", "r");
```
BiteTheDDDDt, line 800: Why don't we use C++ style file read api?

carlvinhust2012: because I want to keep the style with the old code.

be/src/util/system_metrics.cpp
```diff
@@ -718,4 +788,174 @@ void SystemMetrics::get_max_net_traffic(const std::map<std::string, int64_t>& ls
...
+    }
+
+    fclose(fp);
+}
+
+void SystemMetrics::get_metrics_from_proc_meminfo() {
+#ifdef BE_TEST
+    FILE* fp = fopen(k_ut_meminfo_path, "r");
+#else
+    FILE* fp = fopen("/proc/meminfo", "r");
```
xinyiZzz, line 850: `/proc/meminfo` is system memory statistics, which should not be used as metrics for doris BE process.

xinyiZzz: Doris currently uses TCMalloc. Using the following cmd, can get the memory counted by TCMalloc (`http://be_ip:be_httpport/memz`), but this does not include mmap
`MallocExtension::instance()->GetStats`
`MallocExtension::instance()->GetNumericProperty`

Memory statistics including mmap: `http://be_ip:be_httpport/mem_tracker`

carlvinhust2012: > Doris currently uses TCMalloc. Using the following cmd, can get the memory counted by TCMalloc (`http://be_ip:be_httpport/memz`), but this does not include mmap `MallocExtension::instance()->GetStats` `MallocExtension::instance()->GetNumericProperty`
> 
> Memory statistics including mmap: `http://be_ip:be_httpport/mem_tracker`

you means that we no need add the statistics in /proc/meminfo, and just use "http://be_ip:be_httpport/mem_tracker" is enough, right?

xinyiZzz: > you means that we no need add the statistics in /proc/meminfo, and just use "http://be_ip:be_httpport/mem_tracker" is enough, right?

no,
In this pr, use `/proc/vmstat` and `/proc/meminfo` to get system memory or process memory?

carlvinhust2012: after update, just use /proc/vmstat to get system memory.

xinyiZzz: > after update, just use /proc/vmstat to get system memory.

Why use system memory stats as doris metrics. This should use VM monitoring.

carlvinhust2012: just like the cpu metrics, it is also not just belong to doris. so I think the memory stats in /proc/vmstat could be capture d here.

caiconghui: for system metric, I think it is better to use thirdparty tool to do system metric collection instead of integrated into doris, it is a common work for all system manager

xinyiZzz: > for system metric, I think it is better to use thirdparty tool to do system metric collection instead of integrated into doris, it is a common work for all system manager

+1, CPU metrics are the same

yangzhg (approved):
LGTM

PR #9915, opened 2022-06-01
[feature](compaction) add some optimize for compaction

yixiutt:
1. seperate base and cu compaction thread pool
origin thread pool may all be held by base compaction, a case is
be has 8 dirs, each dir can have 4 task at the same time, even
every dir will reserve one slot for cumu compaction, (4-1)*8=24 is
much more bigger than max compaction thread num 10.

2.set base compaction idle schedule to run in low priority

3.remove base compaction for dup key if no delete_predict applied,
  base compaction waste a lot of cpu&io for dup model.

4.add a max_rowset_num limit for base compaction in case base compaction
  run too long

# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/compaction.cpp
+++ b/be/src/olap/compaction.cpp
@@ -158,6 +158,8 @@ Status Compaction::do_compaction_impl(int64_t permits) {
               << ". tablet=" << _tablet->full_name() << ", output_version=" << _output_version
               << ", current_max_version=" << current_max_version
               << ", disk=" << _tablet->data_dir()->path() << ", segments=" << segments_num
+              << ", input_row_num=" << _input_row_num
+              << ", output_row_num=" << _output_rowset->num_rows()
               << ". elapsed time=" << watch.get_elapse_second()
               << "s. cumulative_compaction_policy="
               << _tablet->cumulative_compaction_policy()->name() << ".";
--- a/be/src/olap/olap_server.cpp
+++ b/be/src/olap/olap_server.cpp
@@ -67,11 +67,16 @@ Status StorageEngine::start_bg_threads() {
         data_dirs.push_back(tmp_store.second);
     }
 
-    int32_t max_thread_num = config::max_compaction_threads;
-    ThreadPoolBuilder("CompactionTaskThreadPool")
+    int32_t max_thread_num = config::max_base_compaction_threads;
+    ThreadPoolBuilder("BaseCompactionTaskThreadPool")
             .set_min_threads(max_thread_num)
             .set_max_threads(max_thread_num)
-            .build(&_compaction_thread_pool);
+            .build(&_base_compaction_thread_pool);
+    max_thread_num = config::max_cumu_compaction_threads;
+    ThreadPoolBuilder("CumuCompactionTaskThreadPool")
+            .set_min_threads(max_thread_num)
+            .set_max_threads(max_thread_num)
+            .build(&_cumu_compaction_thread_pool);
 
     int32_t convert_rowset_thread_num = config::convert_rowset_thread_num;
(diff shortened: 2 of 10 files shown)


be/src/olap/tablet_manager.cpp
```diff
@@ -636,6 +636,13 @@ TabletSharedPtr TabletManager::find_best_tablet_to_compaction(
                            << ", tablet_id=" << tablet_ptr->tablet_id();
                 continue;
             }
+            if (compaction_type == CompactionType::BASE_COMPACTION &&
```
morningman, line 639: We should make sure that the rowset is large enough to not perform base comapction. Instead of simply prohibiting base compaction of duplicate tables. Otherwise we might end up with a lot of small files.

be/src/olap/compaction.cpp
```diff
@@ -158,6 +158,8 @@ Status Compaction::do_compaction_impl(int64_t permits) {
               << ". tablet=" << _tablet->full_name() << ", output_version=" << _output_version
               << ", current_max_version=" << current_max_version
               << ", disk=" << _tablet->data_dir()->path() << ", segments=" << segments_num
+              << ", input_row_num=" << _input_row_num
+              << ", output_row_num=" << _output_rowset->num_rows()
```
Gabriel39, line 162: Should we also print filtered rows and merged rows?

be/src/olap/olap_server.cpp
```diff
@@ -304,6 +309,46 @@ void StorageEngine::_tablet_checkpoint_callback(const std::vector<DataDir*>& dat
...
+                      << " to " << config::max_base_compaction_threads;
+        }
+    }
+
+    if (_cumu_compaction_thread_pool->max_threads() != config::max_cumu_compaction_threads) {
+        int old_max_threads = _cumu_compaction_thread_pool->max_threads();
+        Status status =
+                _cumu_compaction_thread_pool->set_max_threads(config::max_cumu_compaction_threads);
+        if (status.ok()) {
+            LOG(INFO) << "update cumu compaction thread pool max_threads from " << old_max_threads
```
Gabriel39, line 337: Change to debug level log?

be/src/olap/olap_server.cpp
```diff
@@ -304,6 +309,46 @@ void StorageEngine::_tablet_checkpoint_callback(const std::vector<DataDir*>& dat
...
+                      << " to " << config::max_cumu_compaction_threads;
+        }
+    }
+    if (_cumu_compaction_thread_pool->min_threads() != config::max_cumu_compaction_threads) {
+        int old_min_threads = _cumu_compaction_thread_pool->min_threads();
+        Status status =
+                _cumu_compaction_thread_pool->set_min_threads(config::max_cumu_compaction_threads);
+        if (status.ok()) {
+            LOG(INFO) << "update cumu compaction thread pool min_threads from " << old_min_threads
```
Gabriel39, line 346: ditto

be/src/olap/olap_server.cpp
yixiutt: ok

be/src/olap/olap_server.cpp
yixiutt: ok

be/src/olap/tablet_manager.cpp
yixiutt: i'll fix it, hanlde it in pick rowset, filter big rowset if no delete exist

be/src/olap/compaction.cpp
yixiutt: we can do it later

PR #9916, opened 2022-06-01
[Feature-wip] support datev2

### LINKED ISSUE #9575: [Feature] Refactor date/datetime type
### Search before asking

### Description

Currently, some problems still exist in Doris date/datetime type. 
1. We have different formats for date/datetime type in storage layer and execution layer and we have to do this transformation to read/write from/to storage.
2. We only support precision to second now. We need to support micro-seconds in the future
3. Timezone is not supported now which is needed for users.

I will do this job in following steps:
1. Add new data type DateV2/DatetimeV2/TimeV2 in FE.
2. Use new format in execution layer in order to more effecient computation.
3. Add new data type DateV2/DatetimeV2 and unify storage layer and execution layer.
4. Support Timestamp.
5. Support timezone for DatetimeV2 and Timestamp.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

Gabriel39:
# Proposed changes

Issue Number: close #9575 

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/types.h
+++ b/be/src/olap/types.h
@@ -500,6 +500,11 @@ struct CppTypeTraits<OLAP_FIELD_TYPE_DATE> {
     using UnsignedCppType = uint24_t;
 };
 template <>
+struct CppTypeTraits<OLAP_FIELD_TYPE_DATEV2> {
+    using CppType = uint32_t;
+    using UnsignedCppType = uint32_t;
+};
+template <>
 struct CppTypeTraits<OLAP_FIELD_TYPE_DATETIME> {
     using CppType = int64_t;
     using UnsignedCppType = uint64_t;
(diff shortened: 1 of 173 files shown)


be/src/olap/types.h
```diff
@@ -974,6 +988,104 @@ struct FieldTypeTraits<OLAP_FIELD_TYPE_DATE> : public BaseFieldtypeTraits<OLAP_F
...
+
+        return Status::OLAPInternalError(OLAP_ERR_INVALID_SCHEMA);
+    }
+    static void set_to_max(void* buf) {
+        // max is 9999 * 16 * 32 + 12 * 32 + 31;
+        *reinterpret_cast<CppType*>(buf) = doris::vectorized::MAX_DATE_V2;
+    }
+    static void set_to_min(void* buf) {
+        // min is 0 * 16 * 32 + 1 * 32 + 1;
+        *reinterpret_cast<CppType*>(buf) = doris::vectorized::MIN_DATE_V2;
```
HappenLee, line 1085: mysql seems support "0000-00-00", rethink whether we need to support the data

be/src/exec/olap_scan_node.cpp
```diff
@@ -1000,7 +1001,8 @@ std::pair<bool, void*> OlapScanNode::should_push_down_eq_predicate(doris::SlotDe
 
 template <typename T, typename ChangeFixedValueRangeFunc>
 Status OlapScanNode::change_fixed_value_range(ColumnValueRange<T>& temp_range, PrimitiveType type,
-                                              void* value, const ChangeFixedValueRangeFunc& func) {
+                                              void* value, const ChangeFixedValueRangeFunc& func,
+                                              PrimitiveType to_type) {
```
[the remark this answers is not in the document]
Gabriel39: This argument is added in order to meet patterns like `cast(datev2 as datetimev2)`. Since datetimev2 has not been  supported, this argument is indeed not used now.

be/src/olap/types.h
Gabriel39: > The DATE type is used for values with a date part but no time part. MySQL retrieves and displays DATE values in 'YYYY-MM-DD' format. The supported range is '1000-01-01' to '9999-12-31'.

> The DATETIME type is used for values that contain both date and time parts. MySQL retrieves and displays DATETIME values in 'YYYY-MM-DD hh:mm:ss' format. The supported range is '1000-01-01 00:00:00' to '9999-12-31 23:59:59'.

This is described in document[1]

[1] https://dev.mysql.com/doc/refman/8.0/en/datetime.html

yiguolei: Maybe we do not support years before 1000. Although we could save it but could not support calculations. And maybe we could support it in the future.

yiguolei (approved):
LGTM

yiguolei:
I will merge it and we can do decimal v3 based on this.

PR #9926, opened 2022-06-01
[feature](storage) rowset meta add tablet schema for future use

yixiutt:
rowset persist tablet schema for future use, such as fast schema change
or compatiable type upgrade. Init rowset meta while be start up so every
rowset will have a tablet schema so we can handle it without too much
code like:
if (rowset_meta.has_schema()) {
    xxxx
} else {
    xxxx
}

# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/rowset/rowset_meta.h
+++ b/be/src/olap/rowset/rowset_meta.h
@@ -28,6 +28,7 @@
 #include "json2pb/json_to_pb.h"
 #include "json2pb/pb_to_json.h"
 #include "olap/olap_common.h"
+#include "olap/tablet_schema.h"
 
 namespace doris {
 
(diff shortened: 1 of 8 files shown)


be/src/olap/rowset/rowset_meta.h
```diff
@@ -276,6 +277,22 @@ class RowsetMeta {
...
+        CHECK(_schema == nullptr);
+        _schema = std::make_shared<TabletSchema>(*tablet_schema);
+    }
+
+    void set_tablet_schema(const TabletSchema& tablet_schema) {
+        TabletSchemaPB* ts_pb = _rowset_meta_pb.mutable_tablet_schema();
+        tablet_schema.to_schema_pb(ts_pb);
+        CHECK(_schema == nullptr);
+        _schema = std::make_shared<TabletSchema>(tablet_schema);
+    }
```
dataroaring, line 292: Two set_tablet_schema methods here is duplicated? Does either one is enough?

yixiutt: fixed

PR #9931, opened 2022-06-02
[fix](regression-test) fix the comparison errors caused by inconsistent accuracy

AshinGau:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

Keep the scale of low precision data to resolve the comparison errors caused by inconsistent accuracy:
```
Check tag 'q09' failed, line 3, DECIMAL result mismatch.
Expect cell is: 0.0395
But real is: 0.039535109
```

## Checklist(Required)

## Further comments

### DIFF
--- a/regression-test/framework/src/main/groovy/org/apache/doris/regression/util/OutputUtils.groovy
+++ b/regression-test/framework/src/main/groovy/org/apache/doris/regression/util/OutputUtils.groovy
@@ -24,6 +24,8 @@ import org.apache.commons.csv.CSVPrinter
 import org.apache.commons.csv.CSVRecord
 import org.apache.commons.io.LineIterator
 
+import java.math.BigDecimal
+import java.math.RoundingMode
 import java.util.function.Function
 import java.sql.ResultSetMetaData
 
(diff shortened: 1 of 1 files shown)


dataroaring (approved):
LGTM

BiteTheDDDDt:
Will this cause the bug to be covered up? The decimal type doesn't seem to have precision errors

AshinGau:
> > Will this cause the bug to be covered up? The decimal type doesn't seem to have precision errors
> 
> Maybe we should change this back after the decimal v3 is completed. Because currently, the decimal type doesn't return real precision.

This fix was introduced for two reasons:
1. The previous fix judges whether two numbers(float & double) are equal by the threshold of relative change. However, this is a fixed value, which may not solve the accuracy problem with larger error, so the rounding method may be better.
2. Other correctness regression tests(tpch 100/1000) are prepared to be added, but the correct result sets come from other frameworks(Trino/Spark). Different frameworks cannot guarantee consistent behavior of decimal.

BiteTheDDDDt:
> > > Will this cause the bug to be covered up? The decimal type doesn't seem to have precision errors
> > 
> > 
> > Maybe we should change this back after the decimal v3 is completed. Because currently, the decimal type doesn't return real precision.
> 
> This fix was introduced for two reasons:
> 
>     1. The previous fix judges whether two numbers(float & double) are equal by the threshold of relative change. However, this is a fixed value, which may not solve the accuracy problem with larger error, so the rounding method may be better.
> 
>     2. Other correctness regression tests(tpch 100/1000) are prepared to be added, but the correct result sets come from other frameworks(Trino/Spark). Different frameworks cannot guarantee consistent behavior of decimal.

OK.
BTW, maybe we should make some modifications to these result sets from other frameworks as appropriate.

PR #9939, opened 2022-06-02
[Enhancement] Refactor functions in int_exp by templates

PragmaTwice:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

The current implementation of functions in `int_exp.h` used some hand-written, hard-coded arrays, which made the code lengthy. 
These functions can be rewritten using templates, which is more generalized and less error-prone.

Supervised by @jackwener. :kissing_heart:

## Checklist(Required)

2. Has unit tests been added: (?)

### DIFF
--- a/be/src/vec/common/int_exp.h
+++ b/be/src/vec/common/int_exp.h
@@ -22,6 +22,30 @@
 
 #include <cstdint>
 #include <limits>
+#include <utility>
+
+namespace exp_details {
+
+template <typename T, T v, std::size_t n>
+constexpr inline const T exp = v * exp<T, v, n - 1>;
+
+template <typename T, T v>
+constexpr inline const T exp<T, v, 0> = 1;
+
+template <typename T, T v, std::size_t ...I>
+constexpr inline const T exp_table[] = { exp<T, v, I>... };
+
+template <typename T, T v, std::size_t ...I>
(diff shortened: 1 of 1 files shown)


BiteTheDDDDt (dismissed):
LGTM

jackwener (dismissed):
LGTM cc @yiguolei

PragmaTwice:
Could you please review it again since I formatted the code? @BiteTheDDDDt

be/src/vec/common/int_exp.h
```diff
@@ -22,6 +22,34 @@
 
 #include <cstdint>
 #include <limits>
+#include <utility>
+
+namespace exp_details {
+
+// compile-time exp(v, n) by linear recursion
+template <typename T, T v, std::size_t n>
+constexpr inline const T exp = v* exp<T, v, n - 1>;
```
BiteTheDDDDt, line 31: Seems here have a bug of clang-format, maybe we can use `T(v) * exp<T, v, n - 1>` to work around.

BiteTheDDDDt (approved):
LGTM

be/src/vec/common/int_exp.h
PragmaTwice: done

PR #9941, opened 2022-06-02
[Enhancement] Column prune for olap scan node

Kikyou1997:
# Proposed changes

Add Column prune for olap scan node based on the project planner framework

## Problem Summary:

Implement the `initOutputSlotIds`  method

## Checklist(Required)

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/OlapScanNode.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/OlapScanNode.java
@@ -26,6 +26,7 @@ import org.apache.doris.analysis.InPredicate;
 import org.apache.doris.analysis.IntLiteral;
 import org.apache.doris.analysis.PartitionNames;
 import org.apache.doris.analysis.SlotDescriptor;
+import org.apache.doris.analysis.SlotId;
 import org.apache.doris.analysis.SlotRef;
 import org.apache.doris.analysis.TupleDescriptor;
 import org.apache.doris.analysis.TupleId;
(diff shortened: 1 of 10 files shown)


yiguolei:
Do not need any changes in BE?

fe/fe-core/src/main/java/org/apache/doris/planner/OlapScanNode.java
```diff
@@ -941,4 +943,17 @@ public DataPartition constructInputPartitionByDistributionInfo() throws UserExce
             return DataPartition.RANDOM;
         }
     }
+
+    @Override
+    public void initOutputSlotIds(Set<SlotId> requiredSlotIdSet, Analyzer analyzer) throws NotImplementedException {
```
EmmyMiao87, line 948: Although fe calculates the output slot of the scan node, if be does not make any changes, the function of project cannot be realized.

fe/fe-core/src/main/java/org/apache/doris/planner/PlanNode.java
```diff
@@ -984,4 +984,17 @@ public String toString() {
         sb.append("\n").append(getNodeExplainString("", TExplainLevel.BRIEF));
         return sb.toString();
     }
+
+    /**
+     * Used to append some common explains to output
+     */
+    protected void appendCommonExplainString(String detailPrefix, StringBuilder output) {
```
EmmyMiao87, line 991: It seems that there is already a getExplainString in the plan node for common string output.

EmmyMiao87 (changes requested):

fe/fe-core/src/main/java/org/apache/doris/planner/OlapScanNode.java
Kikyou1997: OK, I'll add supports in BE later

zenoyang:
Excuse me, why close pr, is there any plan for this?

PR #9951, opened 2022-06-03
[feature] Support grant node_priv to other user.

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

Currently, only the root user has node_priv privileges.
That is, only the root user can operate the addition and deletion of nodes.

In the original design of Doris, there is an Operator role. This role can have node_priv for node operations.

This PR supports assigning node_priv to users other than root.
However, only users who have both grant_priv and node_priv can assign node_priv to other users.
This ensures that only the root user has this permission, and users who are given node_priv
cannot continue to expand this permission outward.

## Checklist(Required)

## Further comments


hf200012 (approved):
LGTM

PR #9964, opened 2022-06-06
add hyperscan and its dependency ragel to thirdparty

xiaokang:
# Proposed changes

We have tested hyperscan for regexp and like. It's 2-3 times faster than re2. Firstly, we need to add hyperscan and its dependency ragel to thirdparty to update doris regression test env.

## Problem Summary:

## Checklist(Required)


xiaokang:
add license and changelog for hyperscan thirdparty

morningman (approved):
LGTM

yangzhg:
hyperscan is not suport arm platform

PR #9976, opened 2022-06-06
Remove the redundant "this."

zengwenjie128:
# Proposed changes

Issue Number: no relative issue

## Problem Summary:

In Java, we can use `this.` to access class member variables .
But, in most scenario, it's redundant. We can remove it from non-constructor functions.

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/task/AgentBatchTask.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/task/AgentBatchTask.java
@@ -64,7 +64,7 @@ public class AgentBatchTask implements Runnable {
     private Map<Long, List<AgentTask>> backendIdToTasks;
 
     public AgentBatchTask() {
-         backendIdToTasks = new HashMap<Long, List<AgentTask>>();
+        backendIdToTasks = new HashMap<Long, List<AgentTask>>();
     }
 
     public AgentBatchTask(AgentTask singleTask) {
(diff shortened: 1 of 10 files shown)


fe/fe-core/src/main/java/org/apache/doris/task/AgentBatchTask.java
```diff
@@ -64,7 +64,7 @@ public class AgentBatchTask implements Runnable {
     private Map<Long, List<AgentTask>> backendIdToTasks;
 
     public AgentBatchTask() {
-        this.backendIdToTasks = new HashMap<Long, List<AgentTask>>();
+         backendIdToTasks = new HashMap<Long, List<AgentTask>>();
     }
```
jackwener, line 68: redundant space.

jackwener (commented):
Please redundant space.

fe/fe-core/src/main/java/org/apache/doris/task/AgentBatchTask.java
zengwenjie128: Thanks!Let me remove these spaces.

dataroaring (changes requested):
Could you explain more on 'removing this'? IMHO, code with explict this is much more readable, and extensive changes would introduce more conflicts for future code maintaince, especially when we pick some commit to dev-1.0.1 branch.

jackwener (commented):
Close it.
As commented above, it's unreasonable PR.

PR #9980, opened 2022-06-07
Fix hive scan node

rebelar:
# Proposed changes

Obviously  a littile mistake.

## Problem Summary:

It should be lower case judge

## Checklist(Required)


shiningbyte:
abandoned

PR #10001, opened 2022-06-07
[Bug] runtime filter is not used as expected

### LINKED ISSUE #10000: [Bug] runtime filter is not used as expected
### Search before asking

### Version

master

### What's Wrong?

OLAP scan node will continue to scan data files when runtime filters are not ready at beginning. As expected, runtime filter should be used once it's ready during scan thread running. But now, runtime filter is not used when it's not ready at beginning.

### What You Expected?

As expected, runtime filter should be used once it's ready during scan thread running. 

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

Gabriel39:
# Proposed changes

Issue Number: close #10000 

## Problem Summary:

OLAP scan node will continue to scan data files when runtime filters are not ready at beginning. As expected, runtime filter should be used once it's ready during scan thread running. But now, runtime filter is not used when it's not ready at beginning.

I tested this PR on SSB using query below:

select * from customer, lineorder where lineorder.lo_custkey = customer.c_custkey and customer.c_name = 'Customer#000200371' limit 1000;

And I got a performance comparison:

Before: 2 min 35.04 sec
After: 1 min 17.50 sec

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -1566,7 +1566,7 @@ void OlapScanNode::scanner_thread(OlapScanner* scanner) {
     auto& scanner_filter_apply_marks = *scanner->mutable_runtime_filter_marks();
     DCHECK(scanner_filter_apply_marks.size() == _runtime_filter_descs.size());
     for (size_t i = 0; i < scanner_filter_apply_marks.size(); i++) {
-        if (!scanner_filter_apply_marks[i]) {
+        if (!scanner_filter_apply_marks[i] && !_runtime_filter_ctxs[i].apply_mark) {
             IRuntimeFilter* runtime_filter = nullptr;
             state->runtime_filter_mgr()->get_consume_filter(_runtime_filter_descs[i].filter_id,
                                                             &runtime_filter);
@@ -1574,7 +1574,6 @@ void OlapScanNode::scanner_thread(OlapScanner* scanner) {
             bool ready = runtime_filter->is_ready();
             if (ready) {
                 runtime_filter->get_prepared_context(&contexts, row_desc(), _expr_mem_tracker);
-                _runtime_filter_ctxs[i].apply_mark = true;
                 scanner_filter_apply_marks[i] = true;
             }
         }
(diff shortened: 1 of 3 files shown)


be/src/exprs/runtime_filter.cpp
```diff
@@ -976,15 +984,14 @@ Status IRuntimeFilter::get_prepared_context(std::vector<ExprContext*>* push_expr
...
     std::lock_guard<std::mutex> guard(_inner_mutex);
 
-    if (!_push_down_ctxs.empty()) {
-        push_expr_ctxs->insert(push_expr_ctxs->end(), _push_down_ctxs.begin(),
-                               _push_down_ctxs.end());
-        return Status::OK();
+    if (_push_down_ctxs.empty()) {
+        RETURN_IF_ERROR(_wrapper->get_push_context(&_push_down_ctxs, _state, _probe_ctx));
+        RETURN_IF_ERROR(Expr::prepare(_push_down_ctxs, _state, desc, tracker));
+        RETURN_IF_ERROR(Expr::open(_push_down_ctxs, _state));
```
BiteTheDDDDt, line 990: If there is no return here, will it cause `_push_down_ctxs` to be inserted twice?

be/src/exec/olap_scan_node.cpp
```diff
@@ -1566,7 +1566,7 @@ void OlapScanNode::scanner_thread(OlapScanner* scanner) {
     auto& scanner_filter_apply_marks = *scanner->mutable_runtime_filter_marks();
     DCHECK(scanner_filter_apply_marks.size() == _runtime_filter_descs.size());
     for (size_t i = 0; i < scanner_filter_apply_marks.size(); i++) {
-        if (!scanner_filter_apply_marks[i] && !_runtime_filter_ctxs[i].apply_mark) {
```
HappenLee, line 1569: should keep the  `!_runtime_filter_ctxs[i].apply_mark`.
because if `runtime filter` come before 1s, the `apply_mark` is `true`, we do not need dispose the `runtime filter` already push down

be/src/exprs/runtime_filter.cpp
Gabriel39: `_push_down_ctxs` will be only inserted once for each scanner.

be/src/exec/olap_scan_node.cpp
Gabriel39: done

HappenLee (approved):
LGTM

PR #10043, opened 2022-06-09
[optimize](compaction) optimize cu compaction policy in pick rowset

yixiutt:
origin _levels is between size_based_compaction_lower_bound_size and
size_based_promotion_size, default conf is 64M and 1024M, which is too
big for cu compaction, and once rowset size bigger than 64M it will be
promoted to base, so this rule seems to be useless.

update _leves to between 32M and 1K, which will reduce a lot of cpu.
In our test, we have 20 concurrent stream load and each rowset in disk
is 3K, after introduce this change, we save about 25% cpu.

# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments


be/src/olap/cumulative_compaction_policy.cpp
```diff
@@ -32,10 +32,11 @@ SizeBasedCumulativeCompactionPolicy::SizeBasedCumulativeCompactionPolicy(
...
           _size_based_promotion_min_size(size_based_promotion_min_size),
           _size_based_compaction_lower_bound_size(size_based_compaction_lower_bound_size) {
-    // init _levels by divide 2 between size_based_promotion_size and size_based_compaction_lower_bound_size
-    int64_t i_size = size_based_promotion_size / 2;
+    // init _levels by divide 2 between size_based_compaction_lower_bound_size and 1K
+    // cu compaction handle file size less then size_based_compaction_lower_bound_size
+    int64_t i_size = size_based_compaction_lower_bound_size / 2;
 
-    while (i_size >= size_based_compaction_lower_bound_size) {
+    while (i_size >= 1024) {
```
zhannngchen, line 39: This value should be configurable?

dataroaring (dismissed):

PR #10060, opened 2022-06-10
[Bug] Fix bug push value predicate of unique table when have sequence column

### LINKED ISSUE #10077: [Bug] Push down value predicate in unique table base tablet with sequence column may cause error result
### Search before asking

### Version

master

### What's Wrong?

unique table have version [0-1] [2-Y]
if rowset version [2-Y] is not compactioned means different segment may have same key like:
segment 1                    segment 2 
k1, v1, seq1                 k1, v2, seq2

seq2 > seq1. so we need output the `k1,v2`

but if the value predicate is `v = v1`, we get wrong result of `k1, v1`. because the `k1,v2,seq2` is filter by the `value predicate`

### What You Expected?

right result

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

HappenLee:
# Proposed changes

Issue Number: close #10077 

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/rowset/beta_rowset_reader.cpp
+++ b/be/src/olap/rowset/beta_rowset_reader.cpp
@@ -71,10 +71,8 @@ Status BetaRowsetReader::init(RowsetReaderContext* read_context) {
                                               read_context->predicates->begin(),
                                               read_context->predicates->end());
     }
-    // if unique table with rowset [0-x] or [0-1] [2-y] [...],
-    // value column predicates can be pushdown on rowset [0-x] or [2-y]
-    if (_rowset->keys_type() == UNIQUE_KEYS &&
-        (_rowset->start_version() == 0 || _rowset->start_version() == 2)) {
+
+    if (_should_push_down_value_predicates()) {
         if (read_context->value_predicates != nullptr) {
             read_options.column_predicates.insert(read_options.column_predicates.end(),
                                                   read_context->value_predicates->begin(),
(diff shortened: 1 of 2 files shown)


be/src/olap/rowset/beta_rowset_reader.h
```diff
@@ -53,6 +53,8 @@ class BetaRowsetReader : public RowsetReader {
     RowsetTypePB type() const override { return RowsetTypePB::BETA_ROWSET; }
 
 private:
+    bool _should_push_down_value_predicates() const;
```
yangzhg, line 56: private function should not  with prefix  `_`

yangzhg (dismissed):
LGTM

be/src/olap/rowset/beta_rowset_reader.cpp
```diff
@@ -237,4 +235,12 @@ Status BetaRowsetReader::next_block(vectorized::Block* block) {
     return Status::OK();
 }
 
+bool BetaRowsetReader::_should_push_down_value_predicates() const {
```
Gabriel39, line 238: should we check if there is a sequence column in tablet schema?

morningman:
please create an issue to describe the problem or error msg

be/src/olap/rowset/beta_rowset_reader.cpp
HappenLee: not need. `not_overlapping()` means the rowset is compactioned, after compactioned the sequence column have same key only keep one row

yangzhg (approved):
LGTM

PR #10070, opened 2022-06-10
[enhancement] Change array offset type from UInt32 to UInt64

cambyzju:
# Proposed changes

Issue Number: close #7570

## Problem Summary:
Now column `Array<T>` contains column `offsets` and `data`, and type of column `offsets` is UInt32 now.
If we call array_union to merge arrays repeatedly, the size of array may overflow.
So we need to extend it before `Array Data Type` release.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/runtime/collection_value.h
+++ b/be/src/runtime/collection_value.h
@@ -64,25 +64,25 @@ class CollectionValue {
 public:
     CollectionValue() = default;
 
-    explicit CollectionValue(uint32_t length)
+    explicit CollectionValue(int64_t length)
             : _data(nullptr), _length(length), _has_null(false), _null_signs(nullptr) {}
 
-    CollectionValue(void* data, uint32_t length)
+    CollectionValue(void* data, int64_t length)
             : _data(data), _length(length), _has_null(false), _null_signs(nullptr) {}
 
-    CollectionValue(void* data, uint32_t length, bool* null_signs)
+    CollectionValue(void* data, int64_t length, bool* null_signs)
             : _data(data), _length(length), _has_null(true), _null_signs(null_signs) {}
 
-    CollectionValue(void* data, uint32_t length, bool has_null, bool* null_signs)
+    CollectionValue(void* data, int64_t length, bool has_null, bool* null_signs)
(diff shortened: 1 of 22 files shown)


be/src/runtime/collection_value.h
```diff
@@ -123,21 +123,21 @@ class CollectionValue {
...
     static Status init_collection(CollectionValue* value, const AllocateMemFunc& allocate,
-                                  uint32_t size, PrimitiveType child_type);
+                                  int64_t size, PrimitiveType child_type);
     ArrayIterator internal_iterator(PrimitiveType child_type) const;
 
 private:
     // child column data
     void* _data;
-    uint32_t _length;
+    int64_t _length;
```
adonis0147, line 140: Using `uint64_t` is better. For some places which need signed value, we use signed value there.

be/src/runtime/collection_value.h
```diff
@@ -160,7 +160,7 @@ class ArrayIterator {
         }
         return false;
     }
-    bool seek(uint32_t n) const {
+    bool seek(int64_t n) const {
```
adonis0147, line 163: If using `int64_t` here, the function `seek` should check whether `n` is negative.

HappenLee:
why int64 not uint64？

cambyzju:
> 

int64 is already enough, and JAVA do not have uint64.

be/src/vec/columns/column_vector.cpp
```diff
@@ -327,7 +327,7 @@ ColumnPtr ColumnVector<T>::replicate(const IColumn::Offsets& offsets) const {
 
     // vectorized this code to speed up
     IColumn::Offset counts[size];
-    for (size_t i = 0; i < size; ++i) {
+    for (ssize_t i = 0; i < size; ++i) {
```
HappenLee, line 330: size is `size_t`, not need change `i` to `sszie_t`

HappenLee (approved):
LGTM

be/src/vec/columns/column_vector.cpp
cambyzju: We use `offsets[i - 1]` inside this for loop, so it is better to use `ssize_t` for `index -1`.

PR #10084, opened 2022-06-13
[feature](vectorized) Support block aggregate in scanner

### LINKED ISSUE #10083: [Enhancement] Improve single-node aggregation parallelism for aggregation tables
### Search before asking

### Description

In our production environment, for 100 million data volumes and cardinality of Bitmaps is ten million, for precise deduplication queries, the bottleneck is always in the Streaming Aggregate stage, which is single-threaded, so we hope to improve the parallelism of aggregation.
For aggregated tables, when the number of key columns by group by is small, it is an easy way to push down aggregates to the Scanner thread because Scanners are executed in parallel.

**Performance Testing**
1. production data:

Test env: 5 BE
tbl has 80 million rows of data, 100 Bucket
```sql
select k1,
   count(distinct device_id)        
from tbl1
group by k1;
```
before: 312.2 s
after aggregation pushdown: 71.48 s

2. data comes from the SSB dataset:

Test environment: 3 BE
lineorder_bitmap_v1 has 10 Bucket
Test sql:
```sql
select LO_ORDERPRIORITY, LO_SHIPMODE, count(distinct LO_ORDERKEY)
from lineorder_bitmap_v1
group by LO_ORDERPRIORITY, LO_SHIPMODE;
```

before: 76.69 s
after: 23.88s

### Solution

_No response_

### Code of Conduct

zenoyang:
# Proposed changes

Issue Number: close #10083 
In scanner thread, the same key of adjacent rows is aggregated, which reduces data transmission and improves the parallelism of aggregation.

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/util/simd/bits.h
+++ b/be/src/util/simd/bits.h
@@ -34,6 +34,8 @@
 
 namespace doris::simd {
 
+static constexpr size_t UINT64_MAX_MASK = 0xFFFFFFFFFFFFFFFF;
+
 inline uint64_t bytes64_mask_to_bits64_not_mask(const int8_t* bytes64) {
 #ifdef __AVX2__
     const auto zero32 = _mm256_setzero_si256();
--- a/be/src/util/simd/bits.h
+++ b/be/src/util/simd/bits.h
@@ -29,58 +29,61 @@
 #include <sse2neon.h>
 #endif
 
-namespace doris {
-namespace simd {
+#include "vec/columns/column.h"
+#include "vec/common/pod_array.h"
 
-/// todo(zeno) Compile add avx512 parameter, modify it to bytes64_mask_to_bits64_mask
-/// Transform 32-byte mask to 32-bit mask
-inline uint32_t bytes32_mask_to_bits32_mask(const uint8_t* data) {
+namespace doris::simd {
+
+inline uint64_t bytes64_mask_to_bits64_not_mask(const int8_t* bytes64) {
 #ifdef __AVX2__
-    auto zero32 = _mm256_setzero_si256();
-    uint32_t mask = static_cast<uint32_t>(_mm256_movemask_epi8(
-            _mm256_cmpgt_epi8(_mm256_loadu_si256(reinterpret_cast<const __m256i*>(data)), zero32)));
-#elif defined(__SSE2__) || defined(__aarch64__)
-    auto zero16 = _mm_setzero_si128();
-    uint32_t mask =
-            (static_cast<uint32_t>(_mm_movemask_epi8(_mm_cmpgt_epi8(
-                    _mm_loadu_si128(reinterpret_cast<const __m128i*>(data)), zero16)))) |
-            ((static_cast<uint32_t>(_mm_movemask_epi8(_mm_cmpgt_epi8(
-                      _mm_loadu_si128(reinterpret_cast<const __m128i*>(data + 16)), zero16)))
-              << 16) &
-             0xffff0000);
+    const auto zero32 = _mm256_setzero_si256();
+    uint64_t res =
+            static_cast<uint64_t>(
+                    _mm256_movemask_epi8(_mm256_cmpeq_epi8(
+                            _mm256_loadu_si256(reinterpret_cast<const __m256i*>(bytes64)),
+                            zero32)) &
+                    0xffffffff) |
+            (static_cast<uint64_t>(_mm256_movemask_epi8(_mm256_cmpeq_epi8(
+                     _mm256_loadu_si256(reinterpret_cast<const __m256i*>(bytes64 + 32)), zero32)))
+             << 32);
+#elif __SSE2__
+    static const __m128i zero16 = _mm_setzero_si128();
+    uint64_t res =
+            static_cast<uint64_t>(_mm_movemask_epi8(_mm_cmpeq_epi8(
+                    _mm_loadu_si128(reinterpret_cast<const __m128i*>(bytes64)), zero16))) |
+            (static_cast<uint64_t>(_mm_movemask_epi8(_mm_cmpeq_epi8(
+                     _mm_loadu_si128(reinterpret_cast<const __m128i*>(bytes64 + 16)), zero16)))
+             << 16) |
+            (static_cast<uint64_t>(_mm_movemask_epi8(_mm_cmpeq_epi8(
+                     _mm_loadu_si128(reinterpret_cast<const __m128i*>(bytes64 + 32)), zero16)))
+             << 32) |
+            (static_cast<uint64_t>(_mm_movemask_epi8(_mm_cmpeq_epi8(
+                     _mm_loadu_si128(reinterpret_cast<const __m128i*>(bytes64 + 48)), zero16)))
+             << 48);
 #else
-    uint32_t mask = 0;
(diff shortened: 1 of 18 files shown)


be/src/vec/olap/block_reader.cpp
```diff
@@ -133,6 +133,29 @@ Status BlockReader::init(const ReaderParams& read_params) {
...
     }
 
+    // only group by key num <= 3 and has bitmap_union col, enable block aggregate
+    if (config::enable_block_aggregate_in_scanner && _aggregation &&
+        (_tablet->keys_type() == AGG_KEYS)) {
+        int agg_key_size = 0;
+        for (int i = 0; i < return_column_size; ++i) {
+            auto cid = read_params.origin_return_columns->at(i);
+            auto& col = _tablet->tablet_schema().column(cid);
+            if (col.is_key() && ++agg_key_size > 3) {
```
Gabriel39, line 143: if `col1` and `col2` are both table keys and agg keys, `_block_aggregation` should also be false because we will get a block  which is already aggregated

Gabriel39 (changes requested):

be/src/vec/olap/block_reader.cpp
```diff
@@ -133,6 +133,29 @@ Status BlockReader::init(const ReaderParams& read_params) {
...
+                _block_aggregation = false;
+                break;
+            }
+            if (!col.is_key() &&
+                col.aggregation() == FieldAggregationMethod::OLAP_FIELD_AGGREGATION_BITMAP_UNION) {
```
Gabriel39, line 148: Maybe all agg will benefit from this? Could we make this optimization  more generic?

be/src/vec/olap/block_reader.cpp
zenoyang: no. If `col1` and `col2` are both table keys and agg keys, the block we get is aggregated based on all keys in the aggregation table, not based on the group by key column. So `_block_aggregation` still needs to be true

be/src/vec/olap/block_reader.cpp
zenoyang: May be. Because the aggregated single-threaded throughput such as `sum` and `count` is already very large, whether it is necessary to use multi-threading requires more testing.
I'll provide more tests soon to decide if it's more general.

BiteTheDDDDt:
Hi, I got a compile fail on clang
```cpp
segment_iterator.cpp:893:43: error: result of comparison of constant 18446744073709551615 with expression of type 'unsigned int' is always false [-Werror,-Wtautological-constant-out-of-range-compare]
            } else if (0xffffffffffffffff == mask) {
```

zenoyang:
> Hi, I got a compile fail on clang
> 
> ```c++
> segment_iterator.cpp:893:43: error: result of comparison of constant 18446744073709551615 with expression of type 'unsigned int' is always false [-Werror,-Wtautological-constant-out-of-range-compare]
>             } else if (0xffffffffffffffff == mask) {
> ```
> 
> maybe we can use `UINT_MAX` here.

Fixed

be/src/util/simd/bits.h
```diff
@@ -89,16 +92,31 @@ inline size_t count_zero_num(const int8_t* __restrict data, size_t size) {
     return num;
 }
 
-inline size_t count_zero_num(const int8_t* __restrict data, const uint8_t* __restrict null_map,
```
Gabriel39, line 92: why remove `__restrict` ?

Gabriel39: This function seems never used. Could we remove it?

be/src/vec/columns/column_decimal.cpp
```diff
@@ -240,13 +240,13 @@ ColumnPtr ColumnDecimal<T>::filter(const IColumn::Filter& filt, ssize_t result_s
...
-    static constexpr size_t SIMD_BYTES = 32;
+    static constexpr size_t SIMD_BYTES = 64;
     const UInt8* filt_end_sse = filt_pos + size / SIMD_BYTES * SIMD_BYTES;
 
     while (filt_pos < filt_end_sse) {
-        uint32_t mask = simd::bytes32_mask_to_bits32_mask(filt_pos);
+        auto mask = simd::bytes64_mask_to_bits64_mask(filt_pos);
 
-        if (0xFFFFFFFF == mask) {
+        if (0xFFFFFFFFFFFFFFFF == mask) {
```
Gabriel39, line 249: Could you make this mask to a common constexpr? I see it is used in multiple places.

be/src/vec/olap/block_reader.cpp
```diff
@@ -168,6 +177,49 @@ Status BlockReader::init(const ReaderParams& read_params) {
     return Status::OK();
 }
 
+Status BlockReader::next_block_with_aggregation(Block* block, MemPool* mem_pool,
+                                                ObjectPool* agg_pool, bool* eof) {
+    return _block_aggregator ? _agg_next_block(block, mem_pool, agg_pool, eof)
```
Gabriel39, line 182: If agg is push down, we'd better to let `_next_block_func` be `BlockReader::_agg_next_block` instead of make this condition judgment each time

be/src/util/simd/bits.h
zenoyang: Because `__restrict` is redundant for `const` pointers. I see this function is called in `VRuntimeFilterWrapper::execute` and cannot be remove.

be/src/vec/columns/column_decimal.cpp
zenoyang: done

be/src/vec/olap/block_reader.cpp
zenoyang: Great suggestion, I've changed it.

be/src/vec/olap/block_aggregator.cpp
```diff
@@ -0,0 +1,234 @@
...
+        return &compare_previous<ColumnFloat32>;
+    case OLAP_FIELD_TYPE_DOUBLE:
+        return &compare_previous<ColumnFloat64>;
+    case OLAP_FIELD_TYPE_DECIMAL:
+        return &compare_previous<ColumnDecimal<Decimal128>>;
+    case OLAP_FIELD_TYPE_CHAR:
+    case OLAP_FIELD_TYPE_VARCHAR:
+    case OLAP_FIELD_TYPE_STRING:
+        return &compare_previous<ColumnString>;
+    case OLAP_FIELD_TYPE_DATE:
```
Gabriel39, line 129: OLAP_FIELD_TYPE_DATEV2,OLAP_FIELD_TYPE_DATETIMEV2,OLAP_FIELD_TYPE_DECIMAL32,OLAP_FIELD_TYPE_DECIMAL64,OLAP_FIELD_TYPE_DECIMAL128 also need to be processed

zenoyang: done

Gabriel39 (approved):
LGTM

yiguolei:
I think we could merge it to 1.1-lts first. 
For master, there are lots of issues related with parallelism. For example, there is only one node for bitmap_union function. we could improve performance by using multiple thread. 
I think we should consider about Pipeline mechanism. 

We could talk about pipeline next week  @liutang123 @wangbo

PR #10136, opened 2022-06-14
[Feature] Lightweight schema change of add/drop column

### LINKED ISSUE #10135: [Feature] Lightweight schema change of add/drop column
### Search before asking

### Description

Background

Add/drop column are heavy operators. They will do linkedSchemaChange or will copy data when data in s3. When add/drop column frequently, a lot of time is wasted waiting. So we need a new way to optimize the process.

Improvement

This improvement involves three aspects, read, writer, compaction. In original impl, BE will hold the tablet schema, set unique id for each column. When read/writer/compaction, BE can get the schema from tablet meta. The core of the modification is
that get the schema from FE when read/writer. And Every rowset will hold its schema. Using the newest schema when doing compaction.

Modification

1. Generate Unique ID by FE.
2. When reading/inserting data, FE will send the newest schema to BE.  
3. When inserting, BE will persistent the schema with rowset meta.
4. When doing compaction, BE will choose newest schema from compation rowsets and make it persistent with new rowset meta after compaction.
5. The improvement is only acting on add/drop value. If add/drop key, it will be done by the old way.
6. It will compatible with old table. Old table is mean that system already has tables before the upgrade. But old table will always do the change by old way although the column is value.

Result

When add/drop value column, they will be lightweight operators. They don't need rewrite the data and complete quickly.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

Lchangliang:
# Proposed changes

Issue Number: close #10135

## Problem Summary:

Optimize ideas for writing in associated issue

## Checklist(Required)

## Further comments


Gabriel39:
What a fancy improvement! And after reading your issue, I'm curious will it increase the pressure on metadata storage because  we persist all schema of all rowsets?

yixiutt:
> What a fancy improvement! And after reading your issue, I'm curious will it increase the pressure on metadata storage because we persist all schema of all rowsets?

schema is really small, and one schema per rowset is really controllable.

yiguolei:
There is already a schema info for every segment. I think you could use it.

gensrc/proto/olap_file.proto
```diff
@@ -93,6 +93,8 @@ message RowsetMetaPB {
     optional int64 num_segments = 22;
     // rowset id definition, it will replace required rowset id 
     optional string rowset_id_v2 = 23;
+    // tablet meta pb, for compaction
+    optional TabletSchemaPB tablet_schema = 24;
```
yiguolei, line 97: No need, there is already a tablet schema for every segment.

be/src/exec/olap_scanner.cpp
```diff
@@ -86,6 +88,14 @@ Status OlapScanner::prepare(
             LOG(WARNING) << ss.str();
             return Status::InternalError(ss.str());
         }
+        _tablet_schema = _tablet->tablet_schema();
```
yiguolei, line 91: OlapScanner is not used in vectorized engine.

be/src/olap/tablet.cpp
```diff
@@ -1636,4 +1641,13 @@ std::shared_ptr<MemTracker>& Tablet::get_compaction_mem_tracker(CompactionType c
     }
 }
 
+const TabletSchema& Tablet::tablet_schema() const {
+    std::shared_lock wrlock(_meta_lock);
+    const RowsetSharedPtr last_rowset = rowset_with_max_version();
```
yiguolei, line 1646: Currently rowset does not have tablet schema. So tablet schema == null

fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
```diff
@@ -4922,4 +4916,190 @@ private static void addTableComment(Table table, StringBuilder sb) {
...
+            LOG.debug("logModifyTableAddOrDropColumns info:{}", info);
+            editLog.logModifyTableAddOrDropColumns(info);
+        }
+
+        //for compatibility, we need create a finished state schema change job v2
+
+        SchemaChangeJobV2 schemaChangeJob = new SchemaChangeJobV2(jobId, db.getId(), olapTable.getId(),
+                olapTable.getName(), 1000);
+        schemaChangeJob.setJobState(AlterJobV2.JobState.FINISHED);
+        schemaChangeJob.setFinishedTimeMs(System.currentTimeMillis());
```
morningman, line 5067: I think we should save the read `createTime` and `finishTime` of this schema change job in `TableAddOrDropColumnsInfo`.(Maybe we can just use one timestamp for both create and finish time).
Otherwise, it will make user confused because after FE restart, the finish time of the schema change job will be changed.

gensrc/proto/olap_file.proto
Lchangliang: There are two reasons that RowsetMeta need the new field. 
1. When compaction, we need to use newset version schema not the schema which newest rowset has. The diff is that when a long load is going, the schema change, and then a short load begin. When the short load was over faster than the long load, the newest schema is the short load rowset has. That's because of that segment don't persist the  schema version too.
2. When we need to use rowset schema，we can read it directly not need to read segment footer.

be/src/exec/olap_scanner.cpp
Lchangliang: VOlapScanner is modified too.

be/src/olap/tablet.cpp
Lchangliang: the continue code process it. 
if (last_rowset == nullptr) {
        return _schema;
 }
that will use the schema from BE saves.

be/src/olap/rowset/beta_rowset.cpp
```diff
@@ -54,11 +54,12 @@ Status BetaRowset::do_load(bool /*use_cache*/) {
     return Status::OK();
 }
 
-Status BetaRowset::load_segments(std::vector<segment_v2::SegmentSharedPtr>* segments) {
+Status BetaRowset::load_segments(std::vector<segment_v2::SegmentSharedPtr>* segments,
+                                 const TabletSchema* read_tablet_schema) {
     for (int seg_id = 0; seg_id < num_segments(); ++seg_id) {
```
yiguolei, line 59: Why segment should related with the input schema? Actually segment or rowset has its own schema.

be/src/olap/rowset/rowset.cpp
```diff
@@ -35,6 +34,12 @@ Rowset::Rowset(const TabletSchema* schema, const FilePathDesc& rowset_path_desc,
         Version version = _rowset_meta->version();
         _is_cumulative = version.first != version.second;
     }
+    // build schema from RowsetMeta.tablet_schema or Tablet.tablet_schema
+    if (_rowset_meta->tablet_schema() != nullptr) {
```
yiguolei, line 38: I find there are some logic that check whether the rowset has tablet schema. Why not add tablet meta to all rowset during data_dir.load()?
Then we do not need deal with these if else logic.

be/src/olap/schema_change.cpp
```diff
@@ -1649,8 +1650,8 @@ Status VSchemaChangeWithSorting::_internal_sorting(
     MultiBlockMerger merger(new_tablet);
 
     std::unique_ptr<RowsetWriter> rowset_writer;
-    RETURN_IF_ERROR(
-            new_tablet->create_rowset_writer(version, VISIBLE, segments_overlap, &rowset_writer));
+    RETURN_IF_ERROR(new_tablet->create_rowset_writer(version, VISIBLE, segments_overlap,
+                                                     &new_tablet->tablet_schema(), &rowset_writer));
```
yiguolei, line 1654: If create rowset writer always bind the tablet's tablet_schema, then its no need as an input parameter. We could get the tablet schema during create_rowset_writer inside.

be/src/olap/compaction.cpp
```diff
@@ -138,8 +138,21 @@ Status Compaction::do_compaction_impl(int64_t permits) {
 
     LOG(INFO) << "start " << merge_type << compaction_name() << ". tablet=" << _tablet->full_name()
               << ", output_version=" << _output_version << ", permits: " << permits;
+    const TabletSchema* cur_tablet_schema = nullptr;
+    if (_input_rowsets.front()->rowset_meta()->tablet_schema() == nullptr) {
+        cur_tablet_schema = &(_tablet->tablet_schema());
+    } else {
+        // get cur schema if rowset schema exist, rowset schema must be newer than tablet schema
+        auto max_version_rowset =
```
yiguolei, line 146: rowset with max version may not have the latest tablet schema.

fe/fe-core/src/main/java/org/apache/doris/catalog/Catalog.java
SWJTU-ZhangLei: AlterJobV2's constructor will calculate createTime.

SWJTU-ZhangLei: > I think we should save the read `createTime` and `finishTime` of this schema change job in `TableAddOrDropColumnsInfo`.(Maybe we can just use one timestamp for both create and finish time). Otherwise, it will make user confused because after FE restart, the finish time of the schema change job will be changed.

> AlterJobV2's constructor will calculate createTime.

dataroaring (changes requested):

be/src/olap/compaction.h
```diff
@@ -67,7 +67,7 @@ class Compaction {
     Status modify_rowsets();
     void gc_output_rowset();
 
-    Status construct_output_rowset_writer();
+    Status construct_output_rowset_writer(const TabletSchema* schama);
```
yiguolei, line 70: typo error

be/src/olap/convert_rowset.cpp
```diff
@@ -48,8 +48,9 @@ Status ConvertRowset::do_convert() {
 
         std::unique_ptr<RowsetWriter> output_rs_writer;
         RETURN_NOT_OK(_tablet->create_rowset_writer(output_version, VISIBLE, NONOVERLAPPING,
-                                                    &output_rs_writer));
-        res = Merger::merge_rowsets(_tablet, ReaderType::READER_BASE_COMPACTION, {input_rs_reader},
+                                                    &_tablet->tablet_schema(), &output_rs_writer));
```
yiguolei, line 51: This file is useless now.

be/src/olap/segment_loader.cpp
```diff
@@ -58,16 +59,17 @@ void SegmentLoader::_insert(const SegmentLoader::CacheKey& key, SegmentLoader::C
 }
 
 Status SegmentLoader::load_segments(const BetaRowsetSharedPtr& rowset,
-                                    SegmentCacheHandle* cache_handle, bool use_cache) {
-    SegmentLoader::CacheKey cache_key(rowset->rowset_id());
-    if (_lookup(cache_key, cache_handle)) {
+                                    SegmentCacheHandle* cache_handle,
+                                    const TabletSchema* read_tablet_schema, bool use_cache) {
+    SegmentLoader::CacheKey cache_key(rowset->rowset_id(), *read_tablet_schema);
```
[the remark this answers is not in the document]
Lchangliang: @yiguolei will rewrite this part soon. It is not necessary to modify it.

be/src/olap/schema_change.cpp
Lchangliang: it is not only use for new_tablet. Such as PushHandler::_convert_v2, Compaction::construct_output_rowset_writer. They should set newer schema from outside.

be/src/olap/convert_rowset.cpp
```diff
@@ -0,0 +1,141 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include "olap/convert_rowset.h"
+
+namespace doris {
+
+Status ConvertRowset::do_convert() {
```
yiguolei, line 22: 这个文件没用了，直接删了吧

be/src/olap/delta_writer.cpp
```diff
@@ -121,10 +121,11 @@ Status DeltaWriter::init() {
         RETURN_NOT_OK(_storage_engine->txn_manager()->prepare_txn(_req.partition_id, _tablet,
                                                                   _req.txn_id, _req.load_id));
     }
+    // build tablet schema in request level
+    _build_current_tablet_schema(_req.index_id, _req.ptable_schema_param, _tablet->tablet_schema());
```
yiguolei, line 125: 在升级过程中可能出现，当前机器是1.2 版本，老机器是1.1 版本，那么导入数据时，1.1 版本的机器转发过来的数据是没schema的，此时这里会有错误。

be/src/olap/delta_writer.h
```diff
@@ -116,7 +122,11 @@ class DeltaWriter {
     // TODO: Recheck the lifetime of _mem_table, Look should use unique_ptr
     std::shared_ptr<MemTable> _mem_table;
     std::unique_ptr<Schema> _schema;
-    const TabletSchema* _tablet_schema;
+    //const TabletSchema* _tablet_schema;
+    // tablet schema owned by delta writer, all write will use this tablet schema
+    // it's build from tablet_schema（stored when create tablet） and OlapTableSchema
+    // every request will have it's own tablet schema so simple schema change can work
+    std::shared_ptr<TabletSchema> _tablet_schema;
```
yiguolei, line 129: why not unique ptr?

be/src/olap/schema_change.cpp
```diff
@@ -2150,9 +2169,9 @@ Status SchemaChangeHandler::_convert_historical_rowsets(const SchemaChangeParams
     bool sc_directly = false;
```
yiguolei, line 2170: schema change 可能会有一个问题。
比如 tablet1 我们做了add column C, 然后就一直没有导入数据； 然后我们做一个schema change 2， 按照C 列排序，此时我们往tablet 2写，从tablet 1 中读取的数据是没有C 列的。
所以schema change 是不是应该，从FE 传递进来原始的和新的schema。

be/src/olap/tablet_schema.cpp
```diff
@@ -460,27 +503,87 @@ void TabletSchema::init_from_pb(const TabletSchemaPB& schema) {
...
+    _schema_version = schema.schema_version();
 }
 
-void TabletSchema::to_schema_pb(TabletSchemaPB* tablet_meta_pb) {
-    tablet_meta_pb->set_keys_type(_keys_type);
+void TabletSchema::build_current_tablet_schema(int64_t index_id,
+                                               const POlapTableSchemaParam& ptable_schema_param,
+                                               const TabletSchema& ori_tablet_schema) {
+    // copy from ori_tablet_schema
+    _keys_type = ori_tablet_schema.keys_type();
```
yiguolei, line 513: 这里的逻辑我看也只有导入用到了，我们可不可以直接导入的时候优FE 生成PB 结构，这个逻辑感觉跟创建表的重合了

be/src/runtime/tablets_channel.h
```diff
@@ -204,6 +204,7 @@ Status TabletsChannel::add_batch(const TabletWriterAddRequest& request,
         if constexpr (std::is_same_v<TabletWriterAddRequest, PTabletWriterAddBatchRequest>) {
             return RowBatch(*_row_desc, request.row_batch());
         } else {
+            LOG(INFO) << "2 block columns: " << vectorized::Block(request.block()).columns();
```
yiguolei, line 207: 调试日志删掉

be/src/olap/delta_writer.cpp
Lchangliang: I don't think that's a problem. In _build_current_tablet_schema, the code will check field 'columns_desc' has value or not. 1.1 send the old proto to 1.2, BE will use tablet_schema directly.

dataroaring (approved):

Issue #10056, opened 2022-06-10
[Enhancement]  compaction strategy enhancement for cu compaction with concurrent stream load


yixiutt:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Description

When concurrent import small files to doris, it seems that cu compaction cost about 80% cpu,
and sometimes be will give -235 version too much error, we should solve this problem

be compaction score

![m4vtXIAwFl](https://user-images.githubusercontent.com/102007456/172988774-fed649be-0a0d-402f-a125-9d9d6f71cd74.png)

### Solution

1.set cumulative_size_based_promotion_min_size_mbytes
2. optimize small file level_size, origin level size is between 64M and 1024M, which is too big for small file import

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #10098, opened 2022-06-13
[improvement](compaction) remove pick rowset skip window strategy to reduce compaction score

yixiutt:
romove the stategy that will not pick rowset generated in 530 seconds.
intruduced in pr #3270, but we now reserve compacted rowset for 30
minutes, so this logic is useless but will lead to a high compaction
score when load frequency is high.

# Proposed changes

Issue Number: close #xxx

## Problem Summary:

Describe the overview of changes.

## Checklist(Required)

1. Does it affect the original behavior: (Yes/No/I Don't know)
2. Has unit tests been added: (Yes/No/No Need)
3. Has document been added or modified: (Yes/No/No Need)
4. Does it need to update dependencies: (Yes/No)
5. Are there any changes that cannot be rolled back: (Yes/No)

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #10153, opened 2022-06-15
[opt](compaction) optimize compaction in concurrent load

yixiutt:
add some logic to opt compaction:
1.seperate base&cumu compaction in case base compaction runs too long and
affect cumu compaction
2.fix level size in cu compaction so that file size below 64M have a right level
size, when choose rowsets to do compaction, the policy will ignore big rowset,
this will reduce about 25% cpu in high frequency concurrent load
3.remove skip window restriction so rowset can do compaction right after
generated, cause we'll not delete rowset after compaction. This will highly
reduce compaction score in concurrent log.
4.remove version consistence check in can_do_compaction, we'll choose a
consecutive rowset to do compaction, so this logic is useless

after add logic above, compaction score and cpu cost will have a substantial
optimize in concurrent load.
compaction score:
m6uuU7V5jZ
00w3cb5xbR

cpu idle:
KKMBQQaJAP

# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/cumulative_compaction_policy.cpp
+++ b/be/src/olap/cumulative_compaction_policy.cpp
@@ -32,10 +32,11 @@ SizeBasedCumulativeCompactionPolicy::SizeBasedCumulativeCompactionPolicy(
           _size_based_promotion_ratio(size_based_promotion_ratio),
           _size_based_promotion_min_size(size_based_promotion_min_size),
           _size_based_compaction_lower_bound_size(size_based_compaction_lower_bound_size) {
-    // init _levels by divide 2 between size_based_promotion_size and size_based_compaction_lower_bound_size
+    // init _levels by divide 2 between size_based_compaction_lower_bound_size and 1K
+    // cu compaction handle file size less then size_based_compaction_lower_bound_size
     int64_t i_size = size_based_promotion_size / 2;
 
-    while (i_size >= size_based_compaction_lower_bound_size) {
+    while (i_size >= 1024) {
         _levels.push_back(i_size);
         i_size /= 2;
     }
@@ -460,16 +461,11 @@ void NumBasedCumulativeCompactionPolicy::calculate_cumulative_point(
 }
 
 void CumulativeCompactionPolicy::pick_candidate_rowsets(
-        int64_t skip_window_sec,
         const std::unordered_map<Version, RowsetSharedPtr, HashOfVersion>& rs_version_map,
         int64_t cumulative_point, std::vector<RowsetSharedPtr>* candidate_rowsets) {
-    int64_t now = UnixSeconds();
     for (auto& it : rs_version_map) {
         // find all rowset version greater than cumulative_point and skip the create time in skip_window_sec
-        if (it.first.first >= cumulative_point &&
-            ((it.second->creation_time() + skip_window_sec < now)
-             // this case means a rowset has been compacted before which is not a new published rowset, so it should participate compaction
-             || (it.first.first != it.first.second))) {
+        if (it.first.first >= cumulative_point) {
             candidate_rowsets->push_back(it.second);
         }
     }
--- a/be/src/olap/storage_engine.h
+++ b/be/src/olap/storage_engine.h
@@ -274,6 +274,8 @@ private:
 
     Status _handle_quick_compaction(TabletSharedPtr);
 
+    void _adjust_compaction_thread_num();
+
 private:
     struct CompactionCandidate {
         CompactionCandidate(uint32_t nicumulative_compaction_, int64_t tablet_id_, uint32_t index_)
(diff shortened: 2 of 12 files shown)


Gabriel39:
#10098 needs to be closed

be/src/olap/cumulative_compaction_policy.cpp
```diff
@@ -32,10 +32,11 @@ SizeBasedCumulativeCompactionPolicy::SizeBasedCumulativeCompactionPolicy(
           _size_based_promotion_ratio(size_based_promotion_ratio),
           _size_based_promotion_min_size(size_based_promotion_min_size),
           _size_based_compaction_lower_bound_size(size_based_compaction_lower_bound_size) {
-    // init _levels by divide 2 between size_based_promotion_size and size_based_compaction_lower_bound_size
+    // init _levels by divide 2 between size_based_compaction_lower_bound_size and 1K
+    // cu compaction handle file size less then size_based_compaction_lower_bound_size
     int64_t i_size = size_based_promotion_size / 2;
 
-    while (i_size >= size_based_compaction_lower_bound_size) {
+    while (i_size >= 1024) {
```
Gabriel39, line 39: 1K is magic number? should we make it a config?

yixiutt: i think 1k is ok whether in big or small file compaction, so i don't use a config here

be/src/olap/cumulative_compaction_policy.cpp
```diff
@@ -460,16 +461,11 @@ void NumBasedCumulativeCompactionPolicy::calculate_cumulative_point(
...
-        int64_t skip_window_sec,
         const std::unordered_map<Version, RowsetSharedPtr, HashOfVersion>& rs_version_map,
         int64_t cumulative_point, std::vector<RowsetSharedPtr>* candidate_rowsets) {
-    int64_t now = UnixSeconds();
     for (auto& it : rs_version_map) {
         // find all rowset version greater than cumulative_point and skip the create time in skip_window_sec
-        if (it.first.first >= cumulative_point &&
-            ((it.second->creation_time() + skip_window_sec < now)
-             // this case means a rowset has been compacted before which is not a new published rowset, so it should participate compaction
-             || (it.first.first != it.first.second))) {
```
zhannngchen, line 472: We should not remove this condition?

be/src/olap/tablet.cpp
```diff
@@ -713,21 +713,6 @@ bool Tablet::can_do_compaction(size_t path_hash, CompactionType compaction_type)
         return false;
     }
 
-    if (tablet_state() == TABLET_RUNNING) {
-        // if tablet state is running, we need to check if it has consistent versions.
-        // tablet in other state such as TABLET_NOTREADY may not have complete versions.
-        std::shared_lock rdlock(_meta_lock);
-        const RowsetSharedPtr lastest_delta = rowset_with_max_version();
```
zhannngchen, line 720: the `rowset_with_max_version` method checked un-published version, we'd better change the implementation here.
e.g. if the published rowset have missing version, we should cancel compaction.

be/src/olap/cumulative_compaction_policy.cpp
yixiutt: ((it.second->creation_time() + skip_window_sec < now)
             // this case means a rowset has been compacted before which is not a new published rowset, so it should participate compaction
             || (it.first.first != it.first.second)))
is useless here, if no skip_window condition, then (it.first.first != it.first.second) is also unnecessary

be/src/olap/tablet.cpp
yixiutt: in concurrent load, this check is not correct cause some publish task is flying, remove this check because we have consective version check in pick input rowset

be/src/olap/cumulative_compaction_policy.cpp
zhannngchen: Ah... you're right.

be/src/olap/cumulative_compaction_policy.cpp
```diff
@@ -32,10 +32,11 @@ SizeBasedCumulativeCompactionPolicy::SizeBasedCumulativeCompactionPolicy(
           _size_based_promotion_ratio(size_based_promotion_ratio),
           _size_based_promotion_min_size(size_based_promotion_min_size),
           _size_based_compaction_lower_bound_size(size_based_compaction_lower_bound_size) {
-    // init _levels by divide 2 between size_based_promotion_size and size_based_compaction_lower_bound_size
+    // init _levels by divide 2 between size_based_compaction_lower_bound_size and 1K
```
zhannngchen, line 35: should be "between size_based_promotion_size and 1K"

zhannngchen (dismissed):

be/src/olap/storage_engine.h
```diff
@@ -272,7 +272,11 @@ class StorageEngine {
 
     Status _submit_compaction_task(TabletSharedPtr tablet, CompactionType compaction_type);
 
+<<<<<<< HEAD
```
dataroaring, line 275: Here is a compilation error.

morningman (approved):
LGTM

dataroaring (approved):
LGTM

Gabriel39:
FYI: If we resolve issue #10056 after merging this PR, you should mention it

PR #10165, opened 2022-06-15
Constrain run-regression-test mem to 2G

pingchunzhang:


dataroaring (approved):
LGTM

HappenLee:
I Think should delete the set in the script. should set any java param by  exporting the `JAVA_OPT`  value

cambyzju:
> I Think should delete the set in the script. should set any java param by exporting the `JAVA_OPT` value

Yes, you are right.

I fix it in PR: https://github.com/apache/doris/pull/10234

PR #10169, opened 2022-06-15
optimize substr performance

xiaokang:
# Proposed changes

optimize substr performance about 1.5~2x speedup.

## Problem Summary:

1. use volnitsky string search algorithm instead of std::search
2. use long run length searching for a big string instead of search for each string

## Checklist(Required)

3. Has unit tests been added: (No Need)

## Further comments

Thanks to clickhouse community, the optimization idea and volnitsky.h string_searcher.h code is from clickhouse.

### DIFF
--- a/be/src/vec/common/string_searcher.h
+++ b/be/src/vec/common/string_searcher.h
@@ -1,0 +1,907 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+// This file is copied from
+// https://github.com/ClickHouse/ClickHouse/blob/master/src/Common/StringSearcher.h
+// and modified by Doris
+
+#pragma once
+
+#include <algorithm>
+#include <vector>
+#include <stdint.h>
+#include <string.h>
+#include <limits>
+
+#include "vec/common/string_utils/string_utils.h"
+
+#ifdef __SSE2__
+    #include <emmintrin.h>
+#endif
+
+#ifdef __SSE4_1__
+    #include <smmintrin.h>
+#endif
+
+namespace doris
+{
+
+// namespace ErrorCodes
+// {
+//     extern const int BAD_ARGUMENTS;
+// }
+
+
+/** Variants for searching a substring in a string.
+  * In most cases, performance is less than Volnitsky (see Volnitsky.h).
+  */
+
+class StringSearcherBase
+{
+public:
+    bool force_fallback = false;
+#ifdef __SSE2__
+protected:
+    static constexpr auto n = sizeof(__m128i);
+    const int page_size = sysconf(_SC_PAGESIZE); //::getPageSize();
+
+    bool page_safe(const void * const ptr) const
+    {
+        return ((page_size - 1) & reinterpret_cast<std::uintptr_t>(ptr)) <= page_size - n;
+    }
+#endif
+};
+
+
+/// Performs case-sensitive and case-insensitive search of UTF-8 strings
+template <bool CaseSensitive, bool ASCII> class StringSearcher;
+
+// comment out since it's not used in doris and UTF8 dependency is not easy to meet
+// /// Case-insensitive UTF-8 searcher
+// template <>
+// class StringSearcher<false, false> : public StringSearcherBase
+// {
+// private:
+//     using UTF8SequenceBuffer = uint8_t[6];
+
+//     /// substring to be searched for
+//     const uint8_t * const needle;
+//     const size_t needle_size;
+//     const uint8_t * const needle_end = needle + needle_size;
+//     /// lower and uppercase variants of the first octet of the first character in `needle`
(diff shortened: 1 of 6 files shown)


BiteTheDDDDt:
Hi, `std::search` has much better performance than gcc in clang, I wonder if `volnitsky` will be better than `std::search` in clang?

xiaokang:
> Hi, `std::search` has much better performance than gcc in clang, I wonder if `volnitsky` will be better than `std::search` in clang?

important information. I will do some comparison test.

xiaokang:
I have done the performance comparison test. The test shows that:

1. volnitsky algorithm with searching across rows is the best, which is 2.0x and 1.5x faster than std::search in two test cases. And gcc
2. gcc-11.1.0 and clang-13.0.1 is almost the same, with little difference and is negligible since tested at different time.

image

xiaokang:
Here are some detailed explanation for the test.

**test env**

- test data: 25GB text log, 110 million rows
- test table: test_table(ts varchar(30), log string)
- test SQL 1: select count() from belog_lz4fix where log like '%minidump%'; the result is 1
- test SQL 2: select count() from belog_lz4fix where log like '%vaggregation\_node.cpp%'; the result is 1201154
- be.conf: disable_storage_page_cache = true
test result


**test result**

- No.1 : the original version, using std::search, std::default_searcher algorithm and calling search once for each single row
- No.2,3,4: change search algorithm to std::boyer_moore_searcher, std::boyer_moore_horspool_searcher, volnitsky respectively.
- No.5,6,7,8: corresponding to 1,2,3,4, change to call search across multi rows. It leverages the continuous memory layout of string column and benefits from long run length miss match.

**analysis**

- compare No.4 with No.1: replacing std::search with volnitsky get 1.51x speedup.
- compare No.5 with No.1: replacing search single row with search across rows for std::search, get 1.25x speedup.
- compare No.8 with No.4: replacing search single row with search across rows for volnitsky, the speedup grows from 1.51x to 2.04x.
- compare No.2,3 with No.1, No.6,7 with 5: there are little difference between std::default_searcher, std::boyer_moore_searcher, std::boyer_moore_horspool_searcher.
- compare the two SQL: the more miss match (or less match), the more speedup

**conclusion**
For the test data and test case:
1. volnitsky algrithm with search across rows is the best.
2. std::boyer_moore_horspool_searcher is better than std::default_searcher

xiaokang:
@BiteTheDDDDt is there more comment?

be/src/vec/functions/like.cpp
```diff
@@ -129,6 +130,42 @@ Status FunctionLikeBase::vector_vector(const ColumnString::Chars& values,
                                        const ColumnString::Offsets& pattern_offsets,
                                        ColumnUInt8::Container& result, const LikeFn& function,
                                        LikeSearchState* search_state) {
+    // for constant_substring_fn, use long run length search for performance
+    if (constant_substring_fn ==
```
BiteTheDDDDt, line 134: Maybe we should move it to a function like named `vector_constant`

be/src/vec/functions/like.cpp
```diff
@@ -129,6 +130,42 @@ Status FunctionLikeBase::vector_vector(const ColumnString::Chars& values,
...
+        while (pos < end) {
+            // search return matched substring start offset
+            pos = (UInt8*)search_state->substring_pattern.search((char*)pos, end - pos);
+            if (pos >= end) break;
+
+            /// Determine which index it refers to.
+            /// begin + value_offsets[i] is the start offset of string at i+1
+            while (begin + value_offsets[i] <= pos) ++i;
+
+            /// We check that the entry does not pass through the boundaries of strings.
```
BiteTheDDDDt, line 156: Whether our `text` and `pattern` may contain `\0`? 
If it is impossible to contain `\0`, then this check will be unnecessary

be/src/vec/common/string_searcher.h
```diff
@@ -0,0 +1,860 @@
...
+        return ((page_size - 1) & reinterpret_cast<std::uintptr_t>(ptr)) <= page_size - n;
+    }
+#endif
+};
+
+/// Performs case-sensitive and case-insensitive search of UTF-8 strings
+template <bool CaseSensitive, bool ASCII>
+class StringSearcher;
+
+// comment out since it's not used in doris and UTF8 dependency is not easy to meet
```
BiteTheDDDDt, line 69: Does we should keep those unused code?

be/src/vec/functions/like.cpp
xiaokang: It's necessary to check the boundary.

Here is an example:

pattern: "cdef"
row1: "abcd"
row2: "efgh"

values.data: "abcdefgh"
value_offsets: [4, 8]

values.data "abcdefgh" will match pattern "cdef" at pos 2. The end of matched substring is pos+needle_size == 6. It's beyond the end offset of row1, which is begin+value_offsets[0]. So due to the check 'if (pos + needle_size < begin + value_offsets[i])', we will not set result[i]=1 improperly.

be/src/vec/functions/like.cpp
xiaokang: There are different functions vectorVector, vectorConstant, constantVector for FunctionsStringSearch in clickhouse.

There are some difference between FunctionsStringSearch in clickhouse and like in doris.

We have many constant functions in like, such as constant_equals_fn, constant_starts_with_fn, constant_ends_with_fn, constant_substring_fn. We still need to add  this special branch for constant_substring_fn, since we need to do special optimization for substring.

be/src/vec/common/string_searcher.h
xiaokang: I keep them in case for possible further usage. If it's better to remove them, I will add a commit.

be/src/vec/functions/like.cpp
BiteTheDDDDt: `values.data` will be "abcd\0efgh" here.
But after my investigation, there may be `\0` in the input string, this check is necessary.

BiteTheDDDDt (approved):

be/src/vec/functions/like.cpp
xiaokang: OK. thanks for you careful review!

PR #10195, opened 2022-06-16
[Improvement] Use IOBufAsZeroCopyInputStream for bloom filter

### LINKED ISSUE #10198: [Enhancement] remove unnecessary memcpy for bloom filter
### Search before asking

### Description

remove unnecessary memcpy for bloom filter

### Solution

_No response_

### Code of Conduct

Gabriel39:
# Proposed changes

Issue Number: close #10198

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/exprs/block_bloom_filter.hpp
+++ b/be/src/exprs/block_bloom_filter.hpp
@@ -21,9 +21,13 @@
 #pragma once
 
 #include "common/status.h"
-#include "gutil/macros.h"
+#include "fmt/format.h"
 #include "util/hash_util.hpp"
 #include "util/slice.h"
+
+namespace butil {
+class IOBufAsZeroCopyInputStream;
+}
 
 namespace doris {
 
@@ -40,11 +44,14 @@ public:
     explicit BlockBloomFilter();
     ~BlockBloomFilter();
 
+    BlockBloomFilter(const BlockBloomFilter&) = delete;
+    BlockBloomFilter& operator=(const BlockBloomFilter&) = delete;
+
     Status init(int log_space_bytes, uint32_t hash_seed);
     // Initialize the BlockBloomFilter from a populated "directory" structure.
     // Useful for initializing the BlockBloomFilter by de-serializing a custom protobuf message.
-    Status init_from_directory(int log_space_bytes, const Slice& directory, bool always_false,
-                               uint32_t hash_seed);
+    Status init_from_directory(int log_space_bytes, butil::IOBufAsZeroCopyInputStream* data,
+                               const size_t data_size, bool always_false, uint32_t hash_seed);
 
     void close();
 
@@ -176,7 +183,7 @@ private:
 
 #endif
     // Size of the internal directory structure in bytes.
-    int64_t directory_size() const { return 1ULL << log_space_bytes(); }
+    size_t directory_size() const { return 1ULL << log_space_bytes(); }
 
     // Some constants used in hashing. #defined for efficiency reasons.
 #define BLOOM_HASH_CONSTANTS                                                                   \
@@ -200,8 +207,6 @@ private:
         // Rehash32to32(hash2) is minimal.
         return (static_cast<uint64_t>(hash) * m + a) >> 32U;
     }
-
-    DISALLOW_COPY_AND_ASSIGN(BlockBloomFilter);
 };
 
 } // namespace doris
--- a/be/src/exprs/block_bloom_filter_impl.cc
+++ b/be/src/exprs/block_bloom_filter_impl.cc
@@ -26,6 +26,8 @@
 #include <mm_malloc.h>
 #endif
 
+#include <butil/iobuf.h>
+
 #include <algorithm>
 #include <climits>
 #include <cmath>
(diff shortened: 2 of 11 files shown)


be/src/exprs/block_bloom_filter.hpp
```diff
@@ -44,8 +47,8 @@ class BlockBloomFilter {
     Status init(int log_space_bytes, uint32_t hash_seed);
     // Initialize the BlockBloomFilter from a populated "directory" structure.
     // Useful for initializing the BlockBloomFilter by de-serializing a custom protobuf message.
-    Status init_from_directory(int log_space_bytes, const Slice& directory, bool always_false,
-                               uint32_t hash_seed);
+    Status init_from_directory(int log_space_bytes, butil::IOBufAsZeroCopyInputStream* data,
+                               const size_t& data_size, bool always_false, uint32_t hash_seed);
```
adonis0147, line 51: Using `const size_t` is enough.
```suggestion
                               const size_t data_size, bool always_false, uint32_t hash_seed);
```

be/src/exprs/block_bloom_filter.hpp
```diff
@@ -177,7 +180,7 @@ class BlockBloomFilter {
 
 #endif
     // Size of the internal directory structure in bytes.
-    int64_t directory_size() const { return 1ULL << log_space_bytes(); }
+    size_t directory_size() const { return 1ULL << log_space_bytes(); }
```
adonis0147, line 183: Why use platform related size here? Using `uint64_t` is better.

```suggestion
    uint64_t directory_size() const { return 1ULL << log_space_bytes(); }
Write Preview
```

be/src/exprs/block_bloom_filter.hpp
```diff
@@ -202,6 +205,13 @@ class BlockBloomFilter {
         return (static_cast<uint64_t>(hash) * m + a) >> 32U;
     }
 
+#ifdef DISALLOW_COPY_AND_ASSIGN
+#undef DISALLOW_COPY_AND_ASSIGN
+#endif
+#define DISALLOW_COPY_AND_ASSIGN(TypeName) \
+    TypeName(const TypeName&) = delete;    \
+    void operator=(const TypeName&) = delete
+
```
adonis0147, line 214: To much similar macros in the codebase, try to unify them rather than create a new one.

be/src/exprs/block_bloom_filter.hpp
```diff
@@ -202,6 +205,13 @@ class BlockBloomFilter {
...
     DISALLOW_COPY_AND_ASSIGN(BlockBloomFilter);
```
cambyzju, line 215:
```suggestion
   BlockBloomFilter(const BlockBloomFilter&) = delete;
   void operator=(const BlockBloomFilter&) = delete;
```

just expand it, it looks like more simpler.

be/src/runtime/runtime_filter_mgr.h
```diff
@@ -68,7 +72,8 @@ class RuntimeFilterMgr {
                          const TQueryOptions& options, int node_id = -1);
 
     // update filter by remote
-    Status update_filter(const PPublishFilterRequest* request, const char* data);
+    Status update_filter(const PPublishFilterRequest* request,
+                         butil::IOBufAsZeroCopyInputStream* data);
```
cambyzju, line 76:
```suggestion
                         const butil::IOBufAsZeroCopyInputStream* data);
```

it is better add const, because we can not modify attachment data.

be/src/exprs/block_bloom_filter_impl.cc
```diff
@@ -83,17 +85,26 @@ Status BlockBloomFilter::init(const int log_space_bytes, uint32_t hash_seed) {
     return Status::OK();
 }
 
-Status BlockBloomFilter::init_from_directory(int log_space_bytes, const Slice& directory,
-                                             bool always_false, uint32_t hash_seed) {
+Status BlockBloomFilter::init_from_directory(int log_space_bytes,
+                                             butil::IOBufAsZeroCopyInputStream* data,
+                                             const size_t& data_size, bool always_false,
+                                             uint32_t hash_seed) {
```
adonis0147, line 91: It may not be a good idea to change the interface like that which increases the number of arguments. `data` and `data_size` are related and users will use them at the same time here. Try to create some `struct` like `Slice`.

be/src/exprs/block_bloom_filter_impl.cc
```diff
@@ -83,17 +85,26 @@ Status BlockBloomFilter::init(const int log_space_bytes, uint32_t hash_seed) {
...
-                directory.size, directory_size()));
+                data_size, directory_size()));
+    }
+    int size = 0;
+    char* tmp;
+    const void** ptr = (const void**)&tmp;
+    char* data_ptr = reinterpret_cast<char*>(_directory);
+    while (data->Next(ptr, &size)) {
+        memcpy(data_ptr, *ptr, size);
+        data_ptr += size;
```
adonis0147, line 106: If you create other `Slice`, you can define a member function to hide this logic which looks a little bit complex.

adonis0147 (changes requested):

be/src/runtime/runtime_filter_mgr.h
Gabriel39: `IOBufAsZeroCopyInputStream.Next` is a non-const method so we can't make it const here

be/src/exprs/block_bloom_filter.hpp
Gabriel39: It's actually s size_t and we use it to compare with other variables with size_t type everywhere. So I just want to keep consistency with other variables here.

be/src/exprs/block_bloom_filter_impl.cc
Gabriel39: Data blocks in IOBufAsZeroCopyInputStream is non-continuous. This is why we use another memcpy before. This PR delete this redundant memcpy and so we can't use Slice here.

adonis0147: I didn't mean to use `Slice`. I think you could create another `struct`.

be/src/exprs/block_bloom_filter_impl.cc
Gabriel39: we only provide users SQL interface. For developers, we can't continue to use Slice here because data blocks in IOBuf are not continuous. Besides, I think changing Slice to  IOBufAsZeroCopyInputStream is not different from adding an argument here.

Gabriel39: I'm not sure if it's necessary. `data_size ` here only means bloom filter's size so it may not be used anywhere. So do we really need to make this abstract? I have some doubts for this.

adonis0147: Two points here:
1. The number of arguments a little more.
2. The following snippet (100-107) looks a little bit complex.

If you introduce another `struct` to wrap them, you can make the code more cleaner.

This comment aimed to enhance the readability other than functionality.

be/src/exprs/block_bloom_filter.hpp
```diff
@@ -40,11 +44,14 @@ class BlockBloomFilter {
     explicit BlockBloomFilter();
     ~BlockBloomFilter();
 
+    BlockBloomFilter(const BlockBloomFilter&) = delete;
+    void operator=(const BlockBloomFilter&) = delete;
```
adonis0147, line 48:
```suggestion
    BlockBloomFilter& operator=(const BlockBloomFilter&) = delete;
```

According to [copy_assignment](https://en.cppreference.com/w/cpp/language/copy_assignment).

adonis0147 (changes requested):

PR #10214, opened 2022-06-17
support tvf

### LINKED ISSUE #10217: [Feature] support table valued function.
### Search before asking

### Description

A table-valued function (TVF) is a function that returns a relation or a set of rows. There are two types of TVFs :
- a TVF that can be specified in a FROM clause, e.g. range;
- a TVF that can be specified in SELECT/LATERAL VIEW clauses, e.g. explode.

Just like the definition of spark and clickhouse：
- spark: https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-tvf.html.
- clickhouse: https://clickhouse.com/docs/en/sql-reference/table-functions.  Currently Clickhouse only support from clause table function

The table valued function can be used to test our vector engines performance.

### Use case

The table valued function works like this：
MySQL [(none)]> select sum(number) from (select number from numbers("1000") where number < 10) as c;
+---------------+
| sum(number) |
+---------------+
| 45 |
+---------------+

That is: numbers("1000") returns a temporary table which only has one column named 'number' and it's row-values are from 0 to 1000.

### Related issues

_No response_

### Code of Conduct

BePPPower:
# Proposed changes

Issue Number: close #10217 

Here implements the first table valued function framework in this PR. The frameworks is as follows:
- add TableValuedFunctionRef as a table ref in SQL definition.
- TableValuedFunctionRef will be parsed and analyzed during analysis process, it will bind to the related TableValuedFunctionInf according to function name.
- A table valued function will generate a dynamic table and scannode during plan stage. So that a TableValuedFunctionInf mainly defined two interface:
- public abstract List getTableColumns()
- public abstract List<TableValuedFunctionTask> getTasks();

Developer should also implement the related ScanNode on BE.

## Problem Summary:

In this PR, there is a demo table valued function——numbers(). User could use select sum(number) from numbers("10000"). And we could use this table function to test our vector engines performance.

MySQL [(none)]> select sum(number) from (select number from numbers("1000") where number < 10) as c;
+---------------+
| sum(number) |
+---------------+
| 45 |
+---------------+
1 row in set (0.01 sec)

Numbers table function is only supported under vectoried engine.

## Checklist(Required)

## Further comments


Gabriel39:
better to create a issue

BePPPower:
> better to create a issue

ok

BePPPower:
> 

Yes, although numbers("5") can be rewritted  to another sql use `table function` , I still think we need `table-valued function` : 
1. `table-valued function` sql is easier to write than rewriting SQL。
2. In addition to `numbers()` , we can implement more `table-valued function` in the future.

BiteTheDDDDt:
> > 
> 
> Yes, although numbers("5") can be rewrote to another sql use `table function` , I still think we need `table-valued function` :
> 
>     1. `table-valued function` sql is easier to write than rewriting SQL。
> 
>     2. In addition to `numbers()` , we can implement more `table-valued function` in the future which may can not be rewrote.

Hi, I think `table function` and `table-valued function` still can reuse a lot of code.

1. We can add `Rewrite-Rule` to FE to automatically rewrite sql, then user can just use `select number(5);` directly.
2. Compared with `TVF`, `TF` has one more input column as argument. But we have some function like `explode_numbers` just ignore the input column. 

We can think of this `TF` that ignores input as `TVF`. Similarly, if we want to implement a new `TVF`, we can implement a `TF` that ignores input and let it participate in `Rewrite-Rule`.

be/src/vec/exec/vtable_valued_function_scannode.cpp
```diff
@@ -0,0 +1,125 @@
...
+VTableValuedFunctionScanNode::VTableValuedFunctionScanNode(ObjectPool* pool, const TPlanNode& tnode,
+                                                           const DescriptorTbl& descs)
+        : ScanNode(pool, tnode, descs),
+          _is_init(false),
+          _tuple_id(tnode.table_valued_func_scan_node.tuple_id),
+          _tuple_desc(nullptr) {
+    // set _table_func here
+    switch (tnode.table_valued_func_scan_node.func_name) {
+    case TTVFunctionName::NUMBERS:
+        _table_func = std::make_shared<VNumbersTBF>(_tuple_id, _tuple_desc);
```
yiguolei, line 42: Is unique ptr ok?

yiguolei:
> ```sql
> explode_numbers
> ```

Yes, TVF numbers could be implemented by using explode_numbers TF.  Numbers is only a simple demo for TVF. We will implement more table valued functions based on this framework. For example, we could implement TVF segments table function. Segments table valued function will have columns (dbid, tableid, rowsetid, segment id, columnname, column type, compressed columnsize, org size).  System admin could get many detail info from this tvf. And we could also add some metric TVF like compactions. We could get current running compactions from every BE.

yiguolei (approved):
LGTM

PR #10278, opened 2022-06-20
[Improvement] remove olap filters when use in key ranges

### LINKED ISSUE #10277: [Enhancement] remove redundant filters in scan node
### Search before asking

### Description

Currently, if a conjuncts is a filter on key columns, we compute it twice while scanning:
1. first, we use it to filter scan keys
2. second, we use it as a push down predicate

### Solution

remove a filter from predicates if it is applicable to key ranges.

### Code of Conduct

Gabriel39:
# Proposed changes

Issue Number: close #10277

### perf test:
table: 
image

records: 59986052

sql:
select count(*) from lineitem where l_orderkey < 60000001;

before this PR:
aAPSIZ3FgK
after this PR:
jjuin9nN59

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/exec/olap_common.h
+++ b/be/src/exec/olap_common.h
@@ -256,9 +256,6 @@ public:
               _is_convertible(true) {}
 
     template <class T>
-    Status extend_scan_key(ColumnValueRange<T>& range, int32_t max_scan_key_num);
-
-    template <class T>
     Status extend_scan_key(ColumnValueRange<T>& range, int32_t max_scan_key_num, bool* exact_value);
 
     Status get_key_range(std::vector<std::unique_ptr<OlapScanRange>>* key_range);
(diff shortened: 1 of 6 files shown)


be/src/exec/olap_common.h
```diff
@@ -258,6 +258,9 @@ class OlapScanKeys {
     template <class T>
     Status extend_scan_key(ColumnValueRange<T>& range, int32_t max_scan_key_num);
 
+    template <class T>
+    Status extend_scan_key(ColumnValueRange<T>& range, int32_t max_scan_key_num, bool* exact_value);
+
```
zenoyang, line 263: The implementations of `Status extend_scan_key(ColumnValueRange<T>& range, int32_t max_scan_key_num);` and `Status extend_scan_key(ColumnValueRange<T>& range, int32_t max_scan_key_num, bool* exact_value);` are very similar, I think they can be combined into one, just give `exact_value` a default value `nullptr` when the function is declared. Before assigning a value to `*exact_value`, first determine whether `exact_value` is `nullptr`.

adonis0147 (changes requested):

be/src/exec/olap_common.h
Gabriel39: To be honest, I considered this way before. Finally I made this because I don't want to mix logics with row-based engine anymore. However, I'm not sure if it's necessary for now and let us wait for more opnions from others :)

BiteTheDDDDt (approved):
LGTM

PR #10298, opened 2022-06-21
[feature-wip](load) Support single replica load

### LINKED ISSUE #9847: [Enhancement] Support single replica load
### Search before asking

### Description

During load process, the same operation are performed on all replicas  such as sort and aggregation, which are resource-intensive. Concurrent data load would consume much CPU and memory resources.
单副本导入

### Solution

It's better to perform write process (writing data into MemTable and then data flush) on single replica and synchronize data files to other replicas before transaction finished.
single_replica_load

### Code of Conduct

weizuo93:
# Proposed changes

Issue Number: close #9847 

## Problem Summary:

During load process, the same operation are performed on all replicas such as sort and aggregation, which are resource-intensive. Concurrent data load would consume much CPU and memory resources. It's better to perform write process (writing data into MemTable and then data flush) on single replica and synchronize data files to other replicas before transaction finished.

## Checklist(Required)

5. Are there any changes that cannot be rolled back:No

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -35,6 +35,11 @@ CONF_Int32(brpc_port, "8060");
 // the number of bthreads for brpc, the default value is set to -1, which means the number of bthreads is #cpu-cores
 CONF_Int32(brpc_num_threads, "-1");
 
+// port to brpc server for single replica load
+CONF_Int32(single_replica_load_brpc_port, "8070");
+// the number of bthreads to brpc server for single replica load
+CONF_Int32(single_replica_load_brpc_num_threads, "64");
+
 // Declare a selection strategy for those servers have many ips.
 // Note that there should at most one ip match this list.
 // this is a list in semicolon-delimited format, in CIDR notation, e.g. 10.10.10.0/24
@@ -360,11 +365,19 @@ CONF_Int32(webserver_num_workers, "48");
 // Period to update rate counters and sampling counters in ms.
 CONF_mInt32(periodic_counter_update_period_ms, "500");
 
+CONF_Bool(enable_single_replica_load, "false");
+
+// Port to download server for single replica load
+CONF_Int32(single_replica_load_download_port, "8050");
+// Number of download workers for single replica load
+CONF_Int32(single_replica_load_download_num_workers, "64");
+
 // Used for mini Load. mini load data file will be removed after this time.
 CONF_Int64(load_data_reserve_hours, "4");
 // log error log will be removed after this time
 CONF_mInt64(load_error_log_reserve_hours, "48");
 CONF_Int32(number_tablet_writer_threads, "16");
+CONF_Int32(number_slave_replica_download_threads, "64");
 
 // The maximum amount of data that can be processed by a stream load
 CONF_mInt64(streaming_load_max_mb, "10240");
@@ -381,6 +394,7 @@ CONF_mInt32(streaming_load_rpc_max_alive_time_sec, "1200");
 CONF_Int32(tablet_writer_open_rpc_timeout_sec, "60");
 // You can ignore brpc error '[E1011]The server is overcrowded' when writing data.
 CONF_mBool(tablet_writer_ignore_eovercrowded, "false");
+CONF_mInt32(slave_replica_writer_rpc_timeout_sec, "60");
 // Whether to enable stream load record function, the default is false.
 // False: disable stream load record
 CONF_mBool(enable_stream_load_record, "false");
--- a/be/src/exec/tablet_sink.h
+++ b/be/src/exec/tablet_sink.h
@@ -174,6 +174,10 @@ public:
 
     virtual Status init(RuntimeState* state);
 
+    void add_slave_tablet_nodes(int64_t tablet_id, const std::vector<int64_t>& slave_nodes) {
+        _slave_tablet_nodes[tablet_id] = slave_nodes;
+    }
+
     // we use open/open_wait to parallel
     void open();
     virtual Status open_wait();
@@ -287,6 +291,8 @@ protected:
     RefCountClosure<PTabletWriterOpenResult>* _open_closure = nullptr;
 
     std::vector<TTabletWithPartition> _all_tablets;
+    // map from tablet_id to node_id where slave replicas locate in
+    std::unordered_map<int64_t, std::vector<int64_t>> _slave_tablet_nodes;
     std::vector<TTabletCommitInfo> _tablet_commit_infos;
 
     AddBatchCounter _add_batch_counter;
@@ -477,6 +483,8 @@ protected:
     // TODO(zc): think about cache this data
     std::shared_ptr<OlapTableSchemaParam> _schema;
     OlapTableLocationParam* _location = nullptr;
+    bool _write_single_replica = false;
+    OlapTableLocationParam* _slave_location = nullptr;
     DorisNodesInfo* _nodes_info = nullptr;
 
     RuntimeProfile* _profile = nullptr;
--- a/be/src/olap/delta_writer.h
+++ b/be/src/olap/delta_writer.h
@@ -72,7 +72,13 @@ public:
     Status close();
     // wait for all memtables to be flushed.
     // mem_consumption() should be 0 after this function returns.
-    Status close_wait();
+    Status close_wait(const PSlaveTabletNodes& slave_tablet_nodes, const bool write_single_replica);
+
+    bool check_slave_replicas_done(google::protobuf::Map<int64_t, PSuccessSlaveTabletNodeIds>*
+                                           success_slave_tablet_node_ids);
+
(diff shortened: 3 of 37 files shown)


caiconghui:
wonderful contribution！

morningman:
https://cwiki.apache.org/confluence/display/DORIS/DSIP-015%3A+Support+single+replica+load+for+load

be/src/common/config.h
```diff
@@ -35,6 +35,11 @@ CONF_Int32(brpc_port, "8060");
 // the number of bthreads for brpc, the default value is set to -1, which means the number of bthreads is #cpu-cores
 CONF_Int32(brpc_num_threads, "-1");
 
+// port to brpc server for single replica load
+CONF_Int32(single_replica_load_brpc_port, "8070");
```
morningman, line 39: The notification rpc is a light weight rpc. Do we really need to create a new brpc port for this?

be/src/exec/tablet_sink.h
```diff
@@ -174,6 +174,10 @@ class NodeChannel {
 
     virtual Status init(RuntimeState* state);
 
+    void add_slave_tablet_nodes(int64_t tablet_id, std::vector<int64_t> slave_nodes) {
```
morningman, line 177:
```suggestion
    void add_slave_tablet_nodes(int64_t tablet_id, const std::vector<int64_t>& slave_nodes) {
```

be/src/olap/delta_writer.h
```diff
@@ -67,7 +67,13 @@ class DeltaWriter {
     Status close();
     // wait for all memtables to be flushed.
     // mem_consumption() should be 0 after this function returns.
-    Status close_wait();
+    Status close_wait(PSlaveTabletNodes slave_tablet_nodes, const bool write_single_replica);
```
morningman, line 70:
```suggestion
    Status close_wait(const PSlaveTabletNodes& slave_tablet_nodes, bool write_single_replica);
```

be/src/runtime/tablets_channel.cpp
```diff
@@ -102,7 +111,7 @@ Status TabletsChannel::close(int sender_id, int64_t backend_id, bool* finished,
         _state = kFinished;
         // All senders are closed
         // 1. close all delta writers
-        std::vector<DeltaWriter*> need_wait_writers;
+        std::set<DeltaWriter*> need_wait_writers;
```
morningman, line 114: Why change to std::set?

be/src/runtime/tablets_channel.cpp
```diff
@@ -124,20 +133,46 @@ Status TabletsChannel::close(int sender_id, int64_t backend_id, bool* finished,
...
+        }
+
+        if (write_single_replica) {
+            CountDownLatch latch(1);
+            while (need_wait_writers.size() > 0 &&
+                   (time(nullptr) - parent->last_updated_time()) < (parent->timeout() * 0.9)) {
+                for (auto writer : need_wait_writers) {
+                    bool is_done = writer->check_slave_replicas_done(success_slave_tablet_node_ids);
+                    if (is_done) {
+                        need_wait_writers.erase(writer);
```
morningman, line 156: Is it safe to erase a map's element when it is being iterated?

be/src/runtime/tablets_channel.cpp
```diff
@@ -124,20 +133,46 @@ Status TabletsChannel::close(int sender_id, int64_t backend_id, bool* finished,
...
             // close may return failed, but no need to handle it here.
             // tablet_vec will only contains success tablet, and then let FE judge it.
-            _close_wait(writer, tablet_vec, tablet_errors);
+            _close_wait(writer, tablet_vec, tablet_errors, slave_nodes, write_single_replica);
+        }
+
+        if (write_single_replica) {
+            CountDownLatch latch(1);
+            while (need_wait_writers.size() > 0 &&
+                   (time(nullptr) - parent->last_updated_time()) < (parent->timeout() * 0.9)) {
```
morningman, line 152: Add comment to explain why it multiply 0.9.

be/src/runtime/tablets_channel.cpp
```diff
@@ -124,20 +133,46 @@ Status TabletsChannel::close(int sender_id, int64_t backend_id, bool* finished,
...
+                for (auto writer : need_wait_writers) {
+                    bool is_done = writer->check_slave_replicas_done(success_slave_tablet_node_ids);
+                    if (is_done) {
+                        need_wait_writers.erase(writer);
+                    }
+                }
+                latch.wait_for(std::chrono::milliseconds(100));
```
morningman, line 159: Why using a CountDownLatch here?
I think we can just call `sleep()`?

be/src/service/internal_service.h
```diff
@@ -175,9 +183,13 @@ class PInternalServiceImpl : public PBackendService {
                                   PTabletWriterAddBlockResult* response,
                                   google::protobuf::Closure* done);
 
+    void _response_pull_slave_rowset(std::string remote_host, int64_t brpc_port, int64_t txn_id,
```
morningman, line 186:
```suggestion
    void _response_pull_slave_rowset(const std::string& remote_host, int64_t brpc_port, int64_t txn_id,
```

fe/fe-core/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -722,6 +722,24 @@ public class Config extends ConfigBase {
...
+     * Whether to enable to write single replica for broker load.
+     */
+    @ConfField(mutable = true, masterOnly = true)
+    public static boolean enable_single_replica_broker_load = false;
+
+    /**
+     * Whether to enable to write single replica for insert.
+     */
+    @ConfField(mutable = true, masterOnly = true)
+    public static boolean enable_single_replica_insert = false;
```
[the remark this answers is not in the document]
weizuo93: > better to use a load property for stream and broker load. and use session variable to control the insert operation.

done

be/src/exec/tablet_sink.h
```diff
@@ -287,6 +291,7 @@ class NodeChannel {
     RefCountClosure<PTabletWriterOpenResult>* _open_closure = nullptr;
 
     std::vector<TTabletWithPartition> _all_tablets;
+    std::unordered_map<int64_t, std::vector<int64_t>> _slave_tablet_nodes;
```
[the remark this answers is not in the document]
weizuo93: > Add comment to explain these `int64_t`

done

be/src/exec/tablet_sink.h
weizuo93: done

be/src/olap/delta_writer.h
weizuo93: done

be/src/runtime/tablets_channel.cpp
weizuo93: > Add comment to explain why it multiply 0.9.

done

be/src/service/internal_service.h
weizuo93: done

be/src/runtime/tablets_channel.cpp
weizuo93: > Why change to std::set?

Element erase operation will follow and  std::set is more efficient than std::vector to execute `erase`.

morningman:
As we discussed earlier, this feature may need to continue to be modified.
So I suggest the following:
1. Add a function switch configuration on the BE side, which is disabled by default. And if it is disabled, all related services and thread pools are not started (because the new port is likely to cause the original service to fail to start due to port conflict)
2. Combine the 3 function switch configurations of FE into one (too many configurations..)

weizuo93:
> As we discussed earlier, this feature may need to continue to be modified. So I suggest the following:
> 
> 1. Add a function switch configuration on the BE side, which is disabled by default. And if it is disabled, all related services and thread pools are not started (because the new port is likely to cause the original service to fail to start due to port conflict)
> 2. Combine the 3 function switch configurations of FE into one (too many configurations..)
> 3. And if BE is disabled, it should return an error to tell user.

done.

fe/fe-core/src/main/java/org/apache/doris/qe/SessionVariable.java
```diff
@@ -511,6 +513,9 @@ public class SessionVariable implements Serializable, Writable {
     @VariableMgr.VarAttr(name = SESSION_CONTEXT, needForward = true)
     public String sessionContext = "";
 
+    @VariableMgr.VarAttr(name = ENABLE_SINGLE_REPLICA_INSERT)
```
morningman, line 516:
```suggestion
    @VariableMgr.VarAttr(name = ENABLE_SINGLE_REPLICA_INSERT, needForward = true)
```

morningman (dismissed):
LGTM

morningman (approved):
LGTM

PR #10311, opened 2022-06-21
[bug] fix window function nullable type bug

### LINKED ISSUE #10308: [Bug] outer join with window function might crash in some case
### Search before asking

### Version

master

### What's Wrong?

SELECT online_detail.game_code, next_login_time
FROM (
select dwd_online_detail.game_code, LEAD(dwd_online_detail.login_time, 1, '9999-12-30 00:00:00') OVER () AS next_login_time
from dim_account_userid_mapping right join dwd_online_detail
on dim_account_userid_mapping.game_code = dwd_online_detail.game_code
) online_detail
LEFT JOIN (
    SELECT day, game AS game_code, dt
    FROM ods_logout
) logout
ON online_detail.game_code = logout.game_code;

crash because of the LEAD function didn't produce nullable type value correctly.

### What You Expected?

get the correct result

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

starocean999:
# Proposed changes

Issue Number: close https://github.com/apache/doris/issues/10308

## Problem Summary:

## Checklist(Required)

## Further comments

the output slot of window function is a nullable type, but the function produce non-nullable value. So before return the block to next node, just convert the column to nullable type if necessary


be/src/vec/exec/vanalytic_eval_node.cpp
```diff
@@ -571,7 +571,16 @@ Status VAnalyticEvalNode::_output_current_block(Block* block) {
     }
 
     for (size_t i = 0; i < _result_window_columns.size(); ++i) {
-        block->insert({std::move(_result_window_columns[i]), _agg_functions[i]->data_type(), ""});
+        SlotDescriptor* output_slot_desc = _output_tuple_desc->slots()[i];
```
Gabriel39, line 574: I think we should fix this bug in another way. Specifically, I think we should figure out why `_agg_functions[i]->data_type()` here is not nullable. Maybe we should change it to nullable in planning phase

HappenLee:
hello, the problem seems a planner problem. I fix the case in #100340

morningman (dismissed):
LGTM

morningman:
This is a bottom-line solution, essentially because the FE planner cannot handle such problems well at present.

morningman:
This PR has been merged into dev-1.0.1 as #10430

PR #10367, opened 2022-06-23
[fix][storage]BetaRowsetReader::next_block does not return 0 rows before eof

### LINKED ISSUE #10366: [Bug] query with predicate returns wrong result when enable storage vectorization
### Search before asking

### Version

dev-1.0.1
master

### What's Wrong?

select count(*) from table where value_column = xxx  returns 0, but expected result is greater than 0.

### What You Expected?

as above.

### How to Reproduce?

The first batch of rows are filtered by predicate of a rowset.

### Anything Else?

_No response_

### Code of Conduct

dataroaring:
Doris uses both Status::Eof and 0 rows as eof in iterators, it is
a little messy. VCollectIterator::Level0Iterator::_refresh_current_row
returns EOF when empty block is returned, then VCollectIterator::build_heap
does not put the iterator to the heap.

So if we have a rowset of which the first batch rows is filtered by
pushed down predicates, then the rowset is not read any more. However,
there may be valid rows after the first batch.

# Proposed changes

Issue Number: close #10366 

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/rowset/beta_rowset_reader.cpp
+++ b/be/src/olap/rowset/beta_rowset_reader.cpp
@@ -184,15 +184,17 @@ Status BetaRowsetReader::next_block(RowBlock** block) {
 Status BetaRowsetReader::next_block(vectorized::Block* block) {
     SCOPED_RAW_TIMER(&_stats->block_fetch_ns);
     if (config::enable_storage_vectorization && _context->is_vec) {
-        auto s = _iterator->next_batch(block);
-        if (!s.ok()) {
-            if (s.is_end_of_file()) {
-                return Status::OLAPInternalError(OLAP_ERR_DATA_EOF);
-            } else {
-                LOG(WARNING) << "failed to read next block: " << s.to_string();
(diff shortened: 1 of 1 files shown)


wangbo (approved):
LGTM

be/src/olap/rowset/beta_rowset_reader.cpp
```diff
@@ -184,15 +184,17 @@ Status BetaRowsetReader::next_block(RowBlock** block) {
...
+            if (!s.ok()) {
+                if (s.is_end_of_file()) {
+                    return Status::OLAPInternalError(OLAP_ERR_DATA_EOF);
+                } else {
+                    LOG(WARNING) << "failed to read next block: " << s.to_string();
+                    return Status::OLAPInternalError(OLAP_ERR_ROWSET_READ_FAILED);
+                }
             }
-        }
+        } while (block->rows() == 0);
```
Gabriel39, line 197: After this PR, rows in block will never be 0 unless EOF. So maybe you also need to update codes in `_refresh_current_row`?

PR #10386, opened 2022-06-24
[improvement]Use std::iota to set values of _block_rowids in SegmentIterator::_read_columns_by_index

mrhhsg:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

`std::iota` is faster than for loop.

ssb-flat 100g q2.3  reduced from 1.5s to 1.2s

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_v2/segment_iterator.cpp
+++ b/be/src/olap/rowset/segment_v2/segment_iterator.cpp
@@ -872,9 +872,11 @@ Status SegmentIterator::_read_columns_by_index(uint32_t nrows_read_limit, uint32
                 _read_columns(_first_read_column_ids, _current_return_columns, rows_to_read));
         _cur_rowid += rows_to_read;
         if (set_block_rowid) {
-            for (uint32_t rid = range_from; rid < range_to; rid++) {
-                _block_rowids[nrows_read++] = rid;
-            }
+            // Here use std::iota is better performance than for-loop, maybe for-loop is not vectorized
+            auto start = _block_rowids.data() + nrows_read;
+            auto end = start + rows_to_read;
+            std::iota(start, end, range_from);
+            nrows_read += rows_to_read;
         } else {
             nrows_read += rows_to_read;
         }


yiguolei (dismissed):
LGTM

adonis0147:
Hi @mrhhsg , the underlying implementation of `std::iota` is also a loop. Why does this modification take effect?

```cpp
template <class _ForwardIterator, class _Tp>
inline _LIBCPP_INLINE_VISIBILITY
void
iota(_ForwardIterator __first, _ForwardIterator __last, _Tp __value_)
{
    for (; __first != __last; ++__first, (void) ++__value_)
        *__first = __value_;
}
```

mrhhsg:
> Hi @mrhhsg , the underlying implementation of `std::iota` is also a loop. Why does this modification take effect?
> 
> ```c++
> template <class _ForwardIterator, class _Tp>
> inline _LIBCPP_INLINE_VISIBILITY
> void
> iota(_ForwardIterator __first, _ForwardIterator __last, _Tp __value_)
> {
>     for (; __first != __last; ++__first, (void) ++__value_)
>         *__first = __value_;
> }
> ```

Maybe the original for-loop cannot be vectorized.

adonis0147:
> > Hi @mrhhsg , the underlying implementation of `std::iota` is also a loop. Why does this modification take effect?
> > ```c++
> > template <class _ForwardIterator, class _Tp>
> > inline _LIBCPP_INLINE_VISIBILITY
> > void
> > iota(_ForwardIterator __first, _ForwardIterator __last, _Tp __value_)
> > {
> >     for (; __first != __last; ++__first, (void) ++__value_)
> >         *__first = __value_;
> > }
> > ```
> 
> Maybe the original for-loop cannot be vectorized.

I think it is better to provide some evidences to prove it.

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -872,9 +872,11 @@ Status SegmentIterator::_read_columns_by_index(uint32_t nrows_read_limit, uint32
                 _read_columns(_first_read_column_ids, _current_return_columns, rows_to_read));
         _cur_rowid += rows_to_read;
         if (set_block_rowid) {
-            for (uint32_t rid = range_from; rid < range_to; rid++) {
-                _block_rowids[nrows_read++] = rid;
-            }
+            // Here use std::iota is better performance than for-loop, maybe for-loop is not vectorized
```
adonis0147, line 875: This comment is wrong, compiler vectorizes both loops.

adonis0147: The improvement here seems the cycle cost consumed by different arithmetical instructions.

mrhhsg: I tested with the code below(compile command: ` g++ test.cc -O3 -fopt-info-vec-all`:
1. `nrows_read` as reference:
```c++
#include <string>
#include <vector>

std::vector<uint32_t> _block_rowids;

void func(uint32_t range_from, uint32_t range_to, uint32_t& nrows_read) {
    uint32_t* data = _block_rowids.data();
    for (uint32_t rid = range_from; rid < range_to; rid++) {
        data[nrows_read++] = rid;
    }
}

int main(int argc, char* argv[]) {
    uint32_t nrows_read = 0;
    _block_rowids.resize(4096);
    func(100, 2000, nrows_read);
    return 0;
}
```
got:
image

2. `nrows_read` as value and signed.
```c++
#include <string>
#include <vector>

std::vector<uint32_t> _block_rowids;

void func(uint32_t range_from, uint32_t range_to, int32_t nrows_read) {
    uint32_t* data = _block_rowids.data();
    for (uint32_t rid = range_from; rid < range_to; rid++) {
        data[nrows_read++] = rid;
    }
}

int main(int argc, char* argv[]) {
    uint32_t nrows_read = 0;
    _block_rowids.resize(4096);
    func(100, 2000, nrows_read);
    return 0;
}
```
image

wangbo:
https://github.com/apache/doris/pull/9111
How about test this pr?  Let's see which is faster.

mrhhsg:
> #9111 How about test this pr? Let's see which is faster.

@wangbo Great!  It is the same with `std::iota`

yiguolei (approved):
LGTM

be/src/olap/rowset/segment_v2/segment_iterator.cpp
adonis0147: I think the way #9111 which @wangbo used to analyze is more straightforward.

adonis0147: Why not just change the interface `SegmentIterator::_read_columns_by_index` to

```cpp
Status SegmentIterator::_read_columns_by_index(uint64_t nrows_read_limit, uint64_t& nrows_read,
                                               bool set_block_rowid) {

```

PR #10390, opened 2022-06-24
Revert "[improvement](function) optimize substr performance"

morningman:
Reverts apache/doris#10169
This cause be UT fail


yiguolei (approved):
LGTM

PR #10420, opened 2022-06-25
[improvement]Do not lazily read dict encoded columns

### LINKED ISSUE #10419: [Enhancement] No need to lazy read dict encoding column
### Search before asking

### Description

If one column is all dict encoding in one segment, it is almost the same as one int32_t column, so we can read it together with predicate colmuns.

### Solution

_No response_

### Code of Conduct

mrhhsg:
# Proposed changes

Issue Number: close #10419

## Problem Summary:
Test with ssb-flat q3.1 100g:
||master|dict_not_lazy_read|
|-|-|-|
|total|10.74 sec|5.73 sec|
|profile|image|image|

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_v2/segment_iterator.cpp
+++ b/be/src/olap/rowset/segment_v2/segment_iterator.cpp
@@ -643,13 +643,18 @@ Status SegmentIterator::next_batch(RowBlockV2* block) {
 
 // todo(wb) need a UT here
 void SegmentIterator::_vec_init_lazy_materialization() {
-    _is_pred_column.resize(_schema.columns().size(), false);
+    _is_first_read_column.resize(_schema.columns().size(), false);
 
     // including short/vec/delete pred
     std::set<ColumnId> pred_column_ids;
+
+    // Other columns(not predicate columns) can read firstly together with predicate columns
+    std::set<ColumnId> other_first_read_column_ids;
+
     _lazy_materialization_read = false;
 
     std::set<ColumnId> del_cond_id_set;
+    std::set<ColumnId> lazy_read_column_ids;
     _opts.delete_condition_predicates->get_all_column_ids(del_cond_id_set);
 
     if (!_col_predicates.empty() || !del_cond_id_set.empty()) {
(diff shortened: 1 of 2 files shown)


be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -704,61 +705,48 @@ void SegmentIterator::_vec_init_lazy_materialization() {
...
+            _non_predicate_columns.emplace_back(cid);
+            switch (type) {
+            case OLAP_FIELD_TYPE_VARCHAR:
+            case OLAP_FIELD_TYPE_CHAR:
+            case OLAP_FIELD_TYPE_STRING: {
+                // if a string column is all dict encoding in one segment, it's almost same as
+                // an int32_t column, it can be read together with predicate columns.
+                if (config::enable_low_cardinality_optimize &&
+                    _column_iterators[cid]->is_all_dict_encoding()) {
+                    pred_column_ids.insert(cid);
```
yiguolei, line 721: A litter confusing, we should distinguish first read columns and predicate columns. Here, the string column (all dict encoding) is not a predicate column but you add it to pred_column_ids. The code is very confusing.

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -704,61 +705,48 @@ void SegmentIterator::_vec_init_lazy_materialization() {
...
                 _is_pred_column[cid] = true;
+                break;
             }
         }
     }
 
+    _lazy_materialization_read = !lazy_read_column_ids.empty();
+    _first_read_column_ids.assign(pred_column_ids.begin(), pred_column_ids.end());
+    if (_lazy_materialization_read && (_is_need_vec_eval || _is_need_short_eval)) {
+        _non_predicate_columns.assign(lazy_read_column_ids.begin(), lazy_read_column_ids.end());
```
yiguolei, line 747: In the past, segment iterator first read first_columns_ids and then non_predicate_colums because predicate columns == first read columns. If non predicate all dict encoding could also be first read columns then we should change current logic.

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -1042,7 +1045,7 @@ Status SegmentIterator::next_batch(vectorized::Block* block) {
     }
 
     if (!_is_need_vec_eval && !_is_need_short_eval) {
-        _output_non_pred_columns(block);
+        _output_columns_by_relace(block, _first_read_column_ids);
```
yiguolei, line 1048: non pred columns != first read columns. If there is a delete condition, the first read columns will contains delete condition columns.

yiguolei (approved):
LGTM

PR #10427, opened 2022-06-26
[build] Fix nested resource path error when importing as maven project from eclipse

compiletheworld:
1. Fix nested resource path error when importing as maven project from eclipse
2. Add instructions for "Eclipse import FE as maven project" in the developer guide

## Checklist(Required)

3. Has the document been added or modified: (Yes)


yiguolei (approved):
LGTM

PR #10442, opened 2022-06-27
optimize substr performance and fix ASAN global buffer overflow 

xiaokang:
# Proposed changes

fix https://github.com/apache/doris/pull/10169


be/src/vec/common/string_searcher.h
```diff
@@ -0,0 +1,860 @@
...
+        return ((page_size - 1) & reinterpret_cast<std::uintptr_t>(ptr)) <= page_size - n;
+    }
+#endif
+};
+
+/// Performs case-sensitive and case-insensitive search of UTF-8 strings
+template <bool CaseSensitive, bool ASCII>
+class StringSearcher;
+
+// comment out since it's not used in doris and UTF8 dependency is not easy to meet
```
HappenLee, line 69: we should delete unless comment code， same of other code

xiaokang: ok. 

btw, can you have a look at the commit to fix global buffer overflow? It's relevant to vectorization. I think you are the one who is very familiar with vectorization code.

kpfly (dismissed):

xiaokang:
@HappenLee @kpfly Is there any more comments before it can be merged?

kpfly (approved):
LGTM

yiguolei (approved):
LGTM

PR #10447, opened 2022-06-27
[Load][Vectorized] load opt code by change `replace` and `replace_if_not_null` do not copy value

HappenLee:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

After do the opt of this

unique_table 1000W data load time

before:45s -> after:35s

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/vec/aggregate_functions/aggregate_function_window.h
+++ b/be/src/vec/aggregate_functions/aggregate_function_window.h
@@ -213,9 +213,9 @@ public:
     void set_null(bool is_null) { _is_null = is_null; }
     StringRef get_value() const { return _ptr->get_data_at(_offset); }
 
-    void set_value(StringRef ref) {
-        _ptr = (IColumn*)ref.data;
-        _offset = ref.size;
+    void set_value(const IColumn* column, size_t row) {
+        _ptr = column;
+        _offset = row;
     }
     void reset() {
         _is_null = false;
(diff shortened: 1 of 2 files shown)


be/src/vec/aggregate_functions/aggregate_function_window.h
```diff
@@ -262,58 +266,31 @@ struct LeadAndLagData {
                 col.insert_default();
```
BiteTheDDDDt, line 266: Maybe we can sympfy this function to two case `is_null` or not.

be/src/vec/aggregate_functions/aggregate_function_window.h
```diff
@@ -210,26 +210,30 @@ class WindowFunctionNTile final
 struct Value {
 public:
     bool is_null() const { return _is_null; }
-    StringRef get_value() const { return _value; }
-
     void set_null(bool is_null) { _is_null = is_null; }
-    void set_value(StringRef value) { _value = value; }
+    StringRef get_value() const { return _ptr->get_data_at(_offset); }
+
+    void set_value(StringRef ref) {
```
BiteTheDDDDt, line 216: Whether use `StringRef` to represent column_ptr and offset is too tricky?

HappenLee: just keep same with `copy_value` do make sure the interface is same

be/src/vec/aggregate_functions/aggregate_function_window.h
HappenLee: just for reduce one time `virtual function call`

BiteTheDDDDt (approved):
LGTM

PR #10448, opened 2022-06-27
[Load][Vectorized] opt the mem use of aggregate function in load to speed up

HappenLee:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

After do the opt of this

unique_table 1000W data load time

before:35s -> after:25s

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/memtable.cpp
+++ b/be/src/olap/memtable.cpp
@@ -183,8 +183,9 @@ void MemTable::_insert_one_row_from_block(RowInBlock* row_in_block) {
                 _offsets_of_aggregate_states.data());
         for (auto cid = _schema->num_key_columns(); cid < _schema->num_columns(); cid++) {
             auto col_ptr = _input_mutable_block.mutable_columns()[cid].get();
-            _agg_functions[cid]->add(row_in_block->agg_places(cid),
-                                     const_cast<const doris::vectorized::IColumn**>(&col_ptr),
+            auto data = row_in_block->agg_places(cid);
+            _agg_functions[cid]->create(data);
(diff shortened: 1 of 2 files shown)


be/src/olap/memtable.cpp
```diff
@@ -161,11 +179,12 @@ void MemTable::_insert_one_row_from_block(RowInBlock* row_in_block) {
     if (is_exist) {
         _aggregate_two_row_in_block(row_in_block, _vec_hint.curr->key);
     } else {
-        row_in_block->init_agg_places(_agg_functions, _schema->num_key_columns());
+        row_in_block->init_agg_places(
+                (char*)_table_mem_pool->allocate(_total_size_of_aggregate_states),
```
k-i-d-d, line 183: How much is `_total_size_of_aggregate_states` usually, thks

xinyiZzz, line 183: How much is _total_size_of_aggregate_states usually, thks

spaces-X: I think `_total_size_of_aggregate_states` will not be large, depending on the number and type of value columns. Each value column needs a buffer to store the agg result, which is independent of the number of lines.

Usually the number of value column will not be very large, may be less than 10. And the size and alignment length of basic data types are not large.

However, the cumulative value of memory allocated by mem pool will be relatively large.

spaces-X:
LGTM

be/src/olap/memtable.cpp
xinyiZzz: This is suitable for MemPool~

xinyiZzz:
LGTM @HappenLee  rebase for p0 and UT

BiteTheDDDDt (approved):

PR #10476, opened 2022-06-28
[fix] fix api of table schema in http v2

sumtumn:
# Proposed changes
schema in this api should be a list, just like in v1

## Problem Summary:

## Checklist(Required)

## Further comments


yinzhijian:
It should be consistent, but does the v2 version already have users using it? changing it maybe makes it unusable for them?

sumtumn:
> It should be consistent, but does the v2 version already have users using it? changing it maybe makes it unusable for them?

I think this api is not usable before this fix, because anyone who query this api expect to get all columns info in the table but actually he would just got the table's last column info. It is absolutely an unexpected behaviour. So I guess nobody  is really using this api right now.

morningman (approved):
LGTM

PR #10547, opened 2022-07-01
[multi-catalog]Fix hive partiton prune in hive and hudi exteranl table.

### LINKED ISSUE #10546: [Bug] Hive and hudi exteranal table get wrong partition prune for some case.
### Search before asking

### Version

master branch

### What's Wrong?

Use tpch table to execute sql like below:
```
select count(*) from lineitem where  l_shipdate <= date '1998-12-01';
```
this only return 41 rows.

### What You Expected?

as the trino do:
image

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

SaintBacchus:
# Proposed changes

Issue Number: close #10546

## Problem Summary:
`ExprBuilder` use stack to build the expr. 
The input order is : col, value and the output is value, col, but the `>=` is not reverse.
Example:
`col >=  1` => `1 >= col`

In this case, it's better use the queue to keeper the input order.

And also the `CompoundPredicate(OR)` have some problems,  it should be `alwaysTrue` whenever it's not a partiton key or it's not a supported op.

## Checklist(Required)

## Further comments


morningman:
It is strange that all exprs should pass the `NormalizeBinaryPredicatesRule`, and expr like `1 < col` should be converted to `col > 1` already.

SaintBacchus:
> It is strange that all exprs should pass the `NormalizeBinaryPredicatesRule`, and expr like `1 < col` should be converted to `col > 1` already.

it get the col,value and op expr one by one, and then compose they by fixed order. so it will cause issue

fe/fe-core/src/main/java/org/apache/doris/planner/HudiScanNode.java
```diff
@@ -229,41 +225,11 @@ private void initParams(Analyzer analyzer) {
...
-                    .val(TypeInfoFactory.intTypeInfo, 1)
-                    .pred("=", 2).build();
-        }
-    }
-
     private InputSplit[] getSplits() throws UserException, IOException {
         String splitsPath = basePath;
         if (partitionKeys.size() > 0) {
-            extractHivePartitionPredicate();
+            hivePartitionPredicate = HiveMetaStoreClientHelper.convertToHivePartitionExpr(
```
morningman, line 231: We call `convertToHivePartitionExpr` twice in `HudiScanNode`.
I think the call in `getFileStatus()` can be removed?

fe/fe-core/src/main/java/org/apache/doris/catalog/HiveMetaStoreClientHelper.java
```diff
@@ -400,61 +401,123 @@ public static List<FieldSchema> getSchema(String dbName, String tableName, Strin
...
+            return eligible;
+        }
+
+        public ExprNodeGenericFuncDesc getFuncDesc() {
+            return funcDesc;
+        }
+    }
+
+    private static ExprNodeGenericFuncDescContext convertToHivePartitionExpr(Expr dorisExpr,
             List<String> partitions, String tblName) throws DdlException {
```
morningman, line 469:
```suggestion
            List<String> partitionKeys, String tblName) throws DdlException {
```

fe/fe-core/src/main/java/org/apache/doris/catalog/HiveMetaStoreClientHelper.java
```diff
@@ -400,61 +401,123 @@ public static List<FieldSchema> getSchema(String dbName, String tableName, Strin
...
+    private static class ExprNodeGenericFuncDescContext {
+        private final ExprNodeGenericFuncDesc funcDesc;
+        private final boolean eligible;
+
+        public ExprNodeGenericFuncDescContext(ExprNodeGenericFuncDesc funcDesc) {
+            this.funcDesc = funcDesc;
+            this.eligible = true;
+        }
+
+        public ExprNodeGenericFuncDescContext() {
```
morningman, line 451: We can use a `static final` field to replace this constructor.

fe/fe-core/src/main/java/org/apache/doris/catalog/HiveMetaStoreClientHelper.java
```diff
@@ -400,61 +401,123 @@ public static List<FieldSchema> getSchema(String dbName, String tableName, Strin
...
                 default:
-                    return null;
+                    return new ExprNodeGenericFuncDescContext();
             }
         }
+        return binaryExprDesc(dorisExpr, partitions, tblName);
+    }
 
+    private static ExprNodeGenericFuncDescContext binaryExprDesc(Expr dorisExpr,
+            List<String> partitions, String tblName) throws DdlException {
```
morningman, line 520:
```suggestion
            List<String> partitionKeys, String tblName) throws DdlException {
```

fe/fe-core/src/main/java/org/apache/doris/catalog/HiveMetaStoreClientHelper.java
```diff
@@ -400,61 +401,123 @@ public static List<FieldSchema> getSchema(String dbName, String tableName, Strin
...
+                    } else {
+                        return new ExprNodeGenericFuncDescContext();
                     }
-                    return null;
                 }
                 case OR: {
-                    ExprNodeGenericFuncDesc left = convertToHivePartitionExpr(
-                            compoundPredicate.getChild(0), partitions, tblName);
-                    ExprNodeGenericFuncDesc right = convertToHivePartitionExpr(
+                    ExprNodeGenericFuncDescContext left = convertToHivePartitionExpr(
```
morningman, line 497:
```
                    ExprNodeGenericFuncDescContext left = convertToHivePartitionExpr(
                            compoundPredicate.getChild(0), partitions, tblName);
                    ExprNodeGenericFuncDescContext right = convertToHivePartitionExpr(
                            compoundPredicate.getChild(1), partitions, tblName);
```
This part can be extracted out of `switch case` block

fe/fe-core/src/main/java/org/apache/doris/catalog/HiveMetaStoreClientHelper.java
```diff
@@ -400,61 +401,123 @@ public static List<FieldSchema> getSchema(String dbName, String tableName, Strin
...
+                            compoundPredicate.getChild(1), partitions, tblName);
+                    if (left.isEligible() && right.isEligible()) {
+                        List<ExprNodeDesc> andArgs = new ArrayList<>();
+                        andArgs.add(left.getFuncDesc());
+                        andArgs.add(right.getFuncDesc());
+                        return new ExprNodeGenericFuncDescContext(getCompoundExpr(andArgs, "or"));
+                    } else {
+                        // If it is not a partition key, this is an always true expr.
+                        // Or if is a partition key and also is a not supportedOp, this is an always true expr.
+                        return new ExprNodeGenericFuncDescContext(genAlwaysTrueExpr(tblName));
```
morningman, line 509: I think it can just return `new ExprNodeGenericFuncDescContext()` ?

morningman (approved):

morningman (changes requested):
need modify

morningman (approved):
LGTM

PR #10548, opened 2022-07-01
[WIP](unique-key-merge-on-write) Add delete bitmap for DSIP-018

compiletheworld:
# Proposed changes

Add delete bitmap for 
[DSIP-018: Support Merge-On-Write implementation for UNIQUE KEY data model](https://cwiki.apache.org/confluence/display/DORIS/DSIP-018%3A+Support+Merge-On-Write+implementation+for+UNIQUE+KEY+data+model)

## Checklist(Required)

### DIFF
--- a/be/src/olap/tablet_meta.cpp
+++ b/be/src/olap/tablet_meta.cpp
@@ -48,8 +48,8 @@ Status TabletMeta::create(const TCreateTabletReq& request, const TabletUid& tabl
     return Status::OK();
 }
 
-TabletMeta::TabletMeta() : _tablet_uid(0, 0), _schema(new TabletSchema),
-                           _delete_bitmap(new DeleteBitmap()) { }
+TabletMeta::TabletMeta()
+        : _tablet_uid(0, 0), _schema(new TabletSchema), _delete_bitmap(new DeleteBitmap()) {}
 
 TabletMeta::TabletMeta(int64_t table_id, int64_t partition_id, int64_t tablet_id,
                        int64_t replica_id, int32_t schema_hash, uint64_t shard_id,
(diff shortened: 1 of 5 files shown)


gensrc/proto/olap_file.proto
```diff
@@ -283,3 +284,10 @@ message OLAPDataHeaderMessage {
 message OLAPRawDeltaHeaderMessage {
     required int32 schema_hash = 2;
 }
+
+message DeleteBitmapPB {
+    repeated string rowset_ids = 1;
+    repeated uint32 segment_ids = 2;
+    repeated int64 versions = 3;
+    repeated bytes segment_delete_bitmaps = 4;
```
kpfly, line 292: add some comments please.

be/src/olap/tablet_meta.h
```diff
@@ -214,9 +217,120 @@ class TabletMeta {
...
+ * to use a single uint32_t.
+ *
+ * e.g.
+ * There is a key "key1" in rowset id 1, version [1,1], segment id 1, row id 1.
+ * A new load also contains "key1", the rowset id 2, version [2,2], segment id 1
+ * the delete bitmap will be `{1,1,2} -> 1`, which means the "row id 1" in
+ * "rowset id 1, segment id 1" is deleted/overitten by some loads at "version 2"
+ */
+class DeleteBitmap {
+    public:
```
kpfly, line 242: indentation

kpfly (commented):
some checker fail.

be/src/olap/tablet_meta.cpp
```diff
@@ -710,4 +742,106 @@ bool operator!=(const TabletMeta& a, const TabletMeta& b) {
...
+    return *this;
+}
+
+DeleteBitmap DeleteBitmap::snapshot() const {
+    std::shared_lock l(lock);
+    return DeleteBitmap(*this);
+}
+
+void DeleteBitmap::add(const BitmapKey& bitmap, uint32_t row_id) {
+    std::lock_guard l(lock);
```
spaces-X, line 772: It looks like the granularity of the lock is tablet level.

I wonder will there be a performance penalty for this part of serial update when loading concurrently that need to update multiple versions of bitmaps?

compiletheworld: Your concern makes sense, however, 
1. This `add` operation is in-memory and fairly fast enough, and contention on this lock is not significant because there won't be thousands of concurrent loads on a single tablet.
2. The read lock is acquired when the tablet meta is being saved (called by `TabletManager::do_tablet_meta_checkpoint`), serialized bitmap will go to RocksDB, more precisely -- page cache of WAL, it's also kind of memory operation.

It is OK that we don't optimize the lock granularity for now.

be/src/olap/tablet_meta.h
compiletheworld: Done.

gensrc/proto/olap_file.proto
compiletheworld: Done.

kpfly (dismissed):
LGTM

morningman (dismissed):
LGTM

zhannngchen:
@compiletheworld  the formatter failed, pls fix it.

zhannngchen (dismissed):
LGTM

morningman (approved):

PR #10555, opened 2022-07-01
[hotfix][feature](dev-1.0.1) support `max_by` and `min_by` on row-based engine

Gabriel39:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

This is because in previous, we only implement `max/min_by` function in vec engine.
But the vec is not working well will outer join operation for now, so use can not use
`max/min_by` with outer join in vec engine.

So I implement the `max/min_by` in non-vec engine so that user can use it in non-vec.
This is just a compromise way and I just push it to dev-1.0.1 branch.

We are still working on supporting outer join in vec engine in master branch.
So maybe I will not push this implement to master branch.

## Checklist(Required)

## Further comments


be/src/exprs/aggregate_functions.cpp
```diff
@@ -363,6 +363,320 @@ struct DecimalV2AvgState {
...
+        return;
+    }
+    DCHECK(dst->ptr != nullptr);
+    if constexpr (std::is_same_v<T, StringVal> || std::is_same_v<KT, StringVal>) {
+        DCHECK_EQ(sizeof(MaxMinByStateWithString<T, KT>), dst->len);
+        auto max_by = reinterpret_cast<MaxMinByStateWithString<T, KT>*>(dst->ptr);
+
+        bool condition = false;
+        if constexpr (std::is_same_v<StringVal, KT>) {
+            condition = !max_by->flag || maxminby_compare<KT, max_by_fn>(slot2, max_by->val2);
```
morningman, line 504: I can't see the difference among these `if else`?

morningman (approved):
LGTM

PR #10565, opened 2022-07-02
[feature](multi-catalog) Support es datasource

stalary:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

Support es datasource, This PR supports query index related metadata.


fe/fe-core/src/main/java/org/apache/doris/datasource/EsExternalDataSource.java
```diff
@@ -18,42 +18,131 @@
...
      * Default constructor for EsExternalDataSource.
      */
-    public EsExternalDataSource(String name, Map<String, String> props) {
+    public EsExternalDataSource(String name, Map<String, String> props) throws DdlException {
         setName(name);
         getDsProperty().setProperties(props);
         setType("es");
+        validate(props);
+        dbNameToId.put("user", nextId.incrementAndGet());
+        dbNameToId.put("system", nextId.incrementAndGet());
```
morningman, line 81: the dbId should be global unique id. So we need to use Catalog.getNextId();
And add comment to explain this 2 databases.

stalary:
Currently, it only runs es7/8 simple types, es6 and array types need to be developed later.

fe/fe-core/pom.xml
```diff
@@ -629,6 +629,7 @@ under the License.
         <dependency>
             <groupId>org.apache.iceberg</groupId>
             <artifactId>iceberg-core</artifactId>
+            <scope>provided</scope>
```
morningman, line 632: Are you sure that this dependency is provided?

fe/fe-core/src/main/java/org/apache/doris/datasource/EsExternalDataSource.java
```diff
@@ -18,43 +18,175 @@
...
+        try {
+            validate(this.dsProperty.getProperties());
+        } catch (DdlException e) {
+            LOG.warn("validate error", e);
+        }
+        dbNameToId = Maps.newConcurrentMap();
+        idToDb = Maps.newConcurrentMap();
+        this.esRestClient = new EsRestClient(this.nodes, this.username, this.password, this.enableSsl);
+        long defaultDbId = Catalog.getCurrentCatalog().getNextId();
+        dbNameToId.put("default", defaultDbId);
```
morningman, line 157: define a `static final` field for this "default" string.

fe/fe-core/src/main/java/org/apache/doris/catalog/external/EsExternalTable.java
```diff
@@ -0,0 +1,119 @@
...
+        return dbName;
+    }
+
+    @Override
+    public TTableDescriptor toThrift() {
+        return super.toThrift();
+    }
+
+    public EsTable toEsTable() {
+        EsTable esTable = new EsTable(this.id, this.name, this.getFullSchema(), TableType.ES_EXTERNAL_TABLE, name, null,
```
morningman, line 114: I think we don't need to new a `EsTable` everytime we call `toEsTable()`.
You can create it in `init`.

fe/fe-core/pom.xml
```diff
@@ -624,6 +624,7 @@ under the License.
...
```
morningman, line 627: remove this

morningman (dismissed):
LGTM
please rebase to resolve the conflict

morningman (approved):
LGTM

PR #10593, opened 2022-07-04
[fix](load) skip empty orc file

### LINKED ISSUE #10594: [Bug] Load with orc file meet error: Invalid column selected xx
### Search before asking

### Version

1.1

### What's Wrong?

Load orc file from hdfs with error:
```
msg:errCode = 2, detailMessage = ParseError : Invalid column selected id
```

where `id` is one of table's columns

### What You Expected?

The orc file is valid but with no schema, we should skip it.

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

morningman:
# Proposed changes

Issue Number: close #10594

## Problem Summary:

Something the upstream system(eg, hive) may create empty orc file
which only has a header and footer, without schema.
And if we call `_reader->createRowReader()` with selected columns,
it will throw ParserError: Invalid column selected xx.
So here we first check its number of rows and skip these kind of files.

This is only a fix for non-vec load, for vec load, it use arrow scanner
to read orc file, which does not have this problem.

## Checklist(Required)

## Further comments


yixiutt:
LGTM

qidaye (approved):
LGTM

PR #10597, opened 2022-07-04
[Doc] Add idea provided error doc.

stalary:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

The following error will occur when use provided dependency.
` java.lang.ClassNotFoundException: org.apache.hadoop.hive.metastore.api.MetaException`

Because part of the dependency is `provided`, idea needs to do a special config.


hf200012 (approved):
LGTM

jackwener (approved):

PR #10615, opened 2022-07-05
[Enhancement]  Add column prune support for VOlapScanNode

Kikyou1997:
# Proposed changes

As we don't prune column on the olap scan, doris always serialize a lot of uneccesary column in multi BE scenes which cause a huge impact on the performance.

## Problem Summary:

Env: 1FE 3BE
Data Size: 1G tpch tools generated data

We use tpch q13 as example:

time cost before :

0.99
0.95
0.98
1.11
1.04

after this commit:

0.74
0.67
0.71
0.74
0.68

## Checklist(Required)

## Further comments


be/src/vec/exec/volap_scan_node.cpp
```diff
@@ -1833,4 +1842,12 @@ Status VOlapScanNode::get_hints(TabletSharedPtr table, const TPaloScanRange& sca
     return Status::OK();
 }
 
+void VOlapScanNode::init_output_slots() {
+    for (const auto& slot_desc : _tuple_desc->slots()) {
+        _output_slot_flags.emplace_back(_output_slot_ids.empty() ||
+                                        std::find(_output_slot_ids.begin(), _output_slot_ids.end(),
+                                                  slot_desc->id()) != _output_slot_ids.end());
```
yiguolei, line 1849: why? !=

be/src/vec/exec/volap_scan_node.cpp
```diff
@@ -1649,6 +1650,14 @@ Status VOlapScanNode::get_next(RuntimeState* state, Block* block, bool* eos) {
             std::lock_guard<std::mutex> l(_free_blocks_lock);
             _free_blocks.emplace_back(materialized_block);
         }
+
+        auto columns = block->get_columns();
+        auto slots = _tuple_desc->slots();
+        for (int i = 0; i < slots.size(); i++) {
+            if (!_output_slot_flags[i]) {
+                (*std::move(columns[i])).assume_mutable()->clear();
```
yiguolei, line 1658: std::move(columns[i])->

be/src/vec/exec/volap_scanner.h
```diff
@@ -112,6 +112,7 @@ class VOlapScanner {
     std::vector<bool> _runtime_filter_marks;
 
     int _id;
```
yiguolei, line 114: is this  field useful?

Kikyou1997: I think it would be helpful when debugging, cause the id of volapscanner always be -1 now, maybe  it should be marked as compile in  debug only...

be/src/vec/exec/volap_scan_node.cpp
Kikyou1997: done

be/src/vec/exec/volap_scanner.h
Kikyou1997: deleted _id_gen

be/src/vec/exec/volap_scan_node.cpp
Kikyou1997: Only when the slot_desc 's against output_slot_ids exists, the flag could be true

yiguolei (approved):
LGTM

PR #10650, opened 2022-07-06
[Bug][Function] pass intermediate argument list to be

### LINKED ISSUE #10641: [Bug] Core dump when aggregate function with no group by
### Search before asking

### Version

master

### What's Wrong?

Some query like `select max_by(k1,k2) from test` will meet core dump. 
The aggregate function should have more than one arguments.

### What You Expected?

fix it.

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

BiteTheDDDDt:
# Proposed changes

Issue Number: close #10641

## Problem Summary:

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/vec/aggregate_functions/aggregate_function_orthogonal_bitmap.cpp
+++ b/be/src/vec/aggregate_functions/aggregate_function_orthogonal_bitmap.cpp
@@ -34,8 +34,6 @@ AggregateFunctionPtr create_aggregate_function_orthogonal(const std::string& nam
         LOG(WARNING) << "Incorrect number of arguments for aggregate function " << name;
         return nullptr;
     } else if (argument_types.size() == 1) {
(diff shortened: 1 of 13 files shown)


Gabriel39 (commented):
Why we need to pass all parameters for agg in both phase? I'm still confused in this. Intuitively, I think the original logics is better understood.

I tried to resolve this problem caused by `min_by` and `max_by` before. I think a better way is to support variadic arguments for aggregate functions just like the simple functions. We can obtain functions by different names by this way. What do you think?

BiteTheDDDDt:
> 

We should get same argument at two diffrent aggregate phase, then we can create same specific function.
Even if we support variadic aggregate functions, we still need the correct parameter list.

Gabriel39 (approved):
LGTM

yiguolei (approved):
lgtm

PR #10670, opened 2022-07-07
[fix](optimizer) join reorder may cause column non-existence problem

### LINKED ISSUE #10674: [Bug] join reorder may cause column non-existence problem
### Search before asking

### Version

master

### What's Wrong?

summary:
Join reorder will make the dependencies out of order, resulting in non-existent columns in reanalyzing.
for example：
```SQL
select * from t1 inner join t2 on t1.a = t2.b inner join t3 on t3.c = t2.b;
```
The table size relationship is t3>t2>t1,  and the problem that t2.b does not exist will occur in reanalyzing.

The reason is that the dependency check in ReorderTable only needs to satisfy one of the predicates involved regardless of its order, for example, when checking t2, which involves t3.c = t2.b and t1.a = t2.b, the order of [t3,t2] can satisfy t3.c = t2.b, so that the order of [t3,t2,t1] can be checked through

### What You Expected?

fix this

### How to Reproduce?

see https://github.com/apache/doris/pull/10670 testDefaultJoinReorderWithView

### Anything Else?

_No response_

### Code of Conduct

yinzhijian:
# Proposed changes

Issue Number: close #10674

## Problem Summary:

for example:
select * from t1 inner join t2 on t1.a = t2.b inner join t3 on t3.c = t2.b;
If t3 is a large table, it will be placed first after the reorderTable,
and the problem that t2.b does not exist will occur in reanalyzing.

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/SelectStmt.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/SelectStmt.java
@@ -819,7 +819,18 @@ public class SelectStmt extends QueryStmt {
                     List<Expr> candidateEqJoinPredicates = analyzer.getEqJoinConjunctsExcludeAuxPredicates(tid);
                     for (Expr candidateEqJoinPredicate : candidateEqJoinPredicates) {
                         List<TupleId> candidateTupleList = Lists.newArrayList();
-                        Expr.getIds(Lists.newArrayList(candidateEqJoinPredicate), candidateTupleList, null);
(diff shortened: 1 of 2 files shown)


yangzhg (dismissed):
LGTM

yangzhg:
you need to create an issue

EmmyMiao87 (dismissed):
LGTM

yangzhg (approved):
LGTM

PR #10684, opened 2022-07-07
[BugFix] Column datas doesn't match nullmap when vectorization load

Lchangliang:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:
In
```
SegmentWriter::append_block(const vectorized::Block* block, size_t row_pos, size_t num_rows)
```
if a block is divided into several parts to append, we need to use row_pos and num_rows to know the range to append. There is a problem when the column is nullable type.
In
```
void OlapBlockDataConvertor::OlapColumnDataConvertorBase::set_source_column(
        const ColumnWithTypeAndName& typed_column, size_t row_pos, size_t num_rows) {
    DCHECK(row_pos + num_rows <= typed_column.column->size())
            << "row_pos=" << row_pos << ", num_rows=" << num_rows
            << ", typed_column.column->size()=" << typed_column.column->size();
    _typed_column = typed_column;
    _row_pos = row_pos;
    _num_rows = num_rows;
    if (_typed_column.column->is_nullable()) {
        auto nullable_column =
                assert_cast<const vectorized::ColumnNullable*>(_typed_column.column.get());
        _nullmap = nullable_column->get_null_map_data().data();
    }
}
```
,OlapColumnDataConvertorBase will get the nullmap which associated with column.
When call convert_to_olap method
```
Status convert_to_olap() override {
            const vectorized::ColumnVector<T>* column_data = nullptr;
            if (_nullmap) {
                auto nullable_column =
                        assert_cast<const vectorized::ColumnNullable*>(_typed_column.column.get());
                column_data = assert_cast<const vectorized::ColumnVector<T>*>(
                        nullable_column->get_nested_column_ptr().get());
            } else {
                column_data =
                        assert_cast<const vectorized::ColumnVector<T>*>(_typed_column.column.get());
            }

            assert(column_data);
            _values = (const T*)(column_data->get_data().data()) + _row_pos;
            return Status::OK();
        }
```
it will move the ptr of values through the _row_pos. But nullmap doesn't do it and it will return directly.
and then when call append_nullable method
```
Status ColumnWriter::append_nullable(const uint8_t* null_map, const uint8_t** ptr,
                                     size_t num_rows) {
    size_t offset = 0;
    auto next_run_step = [&]() {
        size_t step = 1;
        for (auto i = offset + 1; i < num_rows; ++i) {
            if (null_map[offset] == null_map[i])
                step++;
            else
                break;
        }
        return step;
    };

    do {
        auto step = next_run_step();
        if (null_map[offset]) {
            RETURN_IF_ERROR(append_nulls(step));
            *ptr += get_field()->size() * step;
        } else {
            // TODO:
            //  1. `*ptr += get_field()->size() * step;` should do in this function, not append_data;
            //  2. support array vectorized load and ptr offset add
            RETURN_IF_ERROR(append_data(ptr, step));
        }
        offset += step;
    } while (offset < num_rows);

    return Status::OK();
}
```
it will always read the nullmap starting from zero so that the datas in block starting from row_pos doesn't match the nullmap. This can lead to that the position of nullmap is one, but the data exists.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/vec/olap/olap_data_convertor.cpp
+++ b/be/src/vec/olap/olap_data_convertor.cpp
@@ -195,7 +195,7 @@ Status OlapBlockDataConvertor::OlapColumnDataConvertorBitMap::convert_to_olap() 
 
     size_t total_size = 0;
     if (_nullmap) {
-        const UInt8* nullmap_cur = _nullmap + _row_pos;
+        const UInt8* nullmap_cur = _nullmap;
         while (bitmap_value_cur != bitmap_value_end) {
             if (!*nullmap_cur) {
                 total_size += bitmap_value_cur->getSizeInBytes();
@@ -216,7 +216,7 @@ Status OlapBlockDataConvertor::OlapColumnDataConvertorBitMap::convert_to_olap() 
     char* raw_data = _raw_data.data();
     Slice* slice = _slice.data();
     if (_nullmap) {
-        const UInt8* nullmap_cur = _nullmap + _row_pos;
+        const UInt8* nullmap_cur = _nullmap;
         while (bitmap_value_cur != bitmap_value_end) {
             if (!*nullmap_cur) {
                 slice_size = bitmap_value_cur->getSizeInBytes();
@@ -234,7 +234,7 @@ Status OlapBlockDataConvertor::OlapColumnDataConvertorBitMap::convert_to_olap() 
             ++nullmap_cur;
             ++bitmap_value_cur;
         }
-        assert(nullmap_cur == _nullmap + _row_pos + _num_rows && slice == _slice.get_end_ptr());
+        assert(nullmap_cur == _nullmap + _num_rows && slice == _slice.get_end_ptr());
     } else {
         while (bitmap_value_cur != bitmap_value_end) {
             slice_size = bitmap_value_cur->getSizeInBytes();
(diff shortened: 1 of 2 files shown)


dataroaring (dismissed):

be/src/vec/olap/olap_data_convertor.cpp
```diff
@@ -138,6 +138,7 @@ void OlapBlockDataConvertor::OlapColumnDataConvertorBase::set_source_column(
         auto nullable_column =
                 assert_cast<const vectorized::ColumnNullable*>(_typed_column.column.get());
         _nullmap = nullable_column->get_null_map_data().data();
+        _nullmap += row_pos;
```
yiguolei, line 141: If the method is call twice, the nullmap maybe wrong.

Gabriel39, line 141: `row_pos` has been already used to compute offset of _nullmap somewhere (HLL, BitMap, String, etc.). Maybe some basic types `row_pos` is missing to compute `_nullmap` and we should only fix this in those types.

Gabriel39 (changes requested):

dataroaring (approved):

Gabriel39 (approved):
LGTM

yiguolei (approved):
LGTM

PR #10694, opened 2022-07-08
optimize topn query if order by columns is prefix of sort keys of table

### LINKED ISSUE #10646: [Feature] Push sort to the olapscan.
### Search before asking

### Description

Now, table is order by the `Keys` like `Aggregate Key`....

When we use the sql include `orderby` which match `keys` of table. We can push it.

like:

```sql
create

CREATE TABLE IF NOT EXISTS example_db.expamle_tbl
(
    `user_id` LARGEINT NOT NULL COMMENT "用户id",
    `date` DATE NOT NULL COMMENT "数据灌入日期时间",
    `city` VARCHAR(20) COMMENT "用户所在城市",
    `age` SMALLINT COMMENT "用户年龄",
    `sex` TINYINT COMMENT "用户性别",
    `last_visit_date` DATETIME REPLACE DEFAULT "1970-01-01 00:00:00" COMMENT "用户最后一次访问时间",
    `cost` BIGINT SUM DEFAULT "0" COMMENT "用户总消费",
    `max_dwell_time` INT MAX DEFAULT "0" COMMENT "用户最大停留时间",
    `min_dwell_time` INT MIN DEFAULT "99999" COMMENT "用户最小停留时间"
)
AGGREGATE KEY(`user_id`, `date`, `city`, `age`, `sex`)
DISTRIBUTED BY HASH(`user_id`) BUCKETS 1
PROPERTIES (
"replication_allocation" = "tag.location.default: 1"
);

query:
select * from tbl order by user_id, city limit 10;
```

We can push `sort` to `olapscan`.

Limit->Sort->OlapScan  ------> Limit->Merge->OlapScan.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

xiaokang:
# Proposed changes

Issue Number: close #10646

BE part for https://github.com/apache/doris/issues/10646. The FE part is https://github.com/apache/doris/pull/10647.

## Problem Summary:

There is a common query pattern to find latest time serials data.
 eg. SELECT * from t_log WHERE t>t1 AND t<t2 ORDER BY t DESC LIMIT 100

If the ORDER BY columns is the prefix of the sort key of table, it can
be greatly optimized to read much fewer data instead of read all data
between t1 and t2.

By leveraging the same order of ORDER BY columns and sort key of table,
just read the LIMIT N rows for each related segment and merge N rows.

1. set read_orderby_key to true for read_params and _reader_context
   if olap_scan_node's sort info is set.
2. set read_orderby_key_reverse to true for read_params and _reader_context
   if is_asc_order is false.
3. rowset reader force merge read segments if read_orderby_key is true.
4. block reader and tablet reader force merge read rowsets if read_orderby_key is true.

5. for ORDER BY DESC, read and compare in reverse order
5.1 segment iterator read backward using a new BackwardBitmapRangeIterator and
    reverse the result block before return to caller.
5.2 VCollectIterator::LevelIteratorComparator, VMergeIteratorContext return
    opposite result for _is_reverse order in its compare function.

## Checklist(Required)

## Further comments


be/src/vec/olap/vcollect_iterator.h
```diff
@@ -102,12 +102,14 @@ class VCollectIterator {
...
     public:
-        LevelIteratorComparator(int sequence = -1) : _sequence(sequence) {}
+        LevelIteratorComparator(int sequence, bool is_reverse) :
+            _sequence(sequence), _is_reverse(is_reverse) {}
 
         bool operator()(LevelIterator* lhs, LevelIterator* rhs);
 
     private:
         int _sequence;
+        bool _is_reverse = false;
```
jackwener, line 112: Add comment explain its function?

xiaokang:
we test in a log matching senario, runtime of the following SQL drops significantly from **1.8s to 0.1s** 

`SELECT * FROM log_table WHERE timestamp > t1 ORDER BY timestamp DESC LIMIT 100`

be/src/olap/rowset/beta_rowset_reader.cpp
```diff
@@ -111,9 +112,10 @@ Status BetaRowsetReader::init(RowsetReaderContext* read_context) {
     RowwiseIterator* final_iterator;
     if (config::enable_storage_vectorization && read_context->is_vec) {
         if (read_context->need_ordered_result &&
```
HappenLee, line 114: Rethink the `read_orderby_key` and `need_ordered_result`, seems we only need one

be/src/vec/olap/vcollect_iterator.cpp
```diff
@@ -145,7 +150,7 @@ bool VCollectIterator::LevelIteratorComparator::operator()(LevelIterator* lhs, L
     bool lower = (cmp_res != 0) ? (cmp_res < 0) : (lhs->version() < rhs->version());
     lower ? lhs->set_same(true) : rhs->set_same(true);
 
-    return lower;
+    return UNLIKELY(_is_reverse) ? !lower : lower;
```
HappenLee, line 153: do not need reverse in key equals case

be/src/vec/olap/vcollect_iterator.cpp
```diff
@@ -34,14 +34,19 @@ VCollectIterator::~VCollectIterator() {}
...
     // when aggregate is enabled or key_type is DUP_KEYS, we don't merge
     // multiple data to aggregate for better performance
     if (_reader->_reader_type == READER_QUERY &&
         (_reader->_direct_mode || _reader->_tablet->keys_type() == KeysType::DUP_KEYS)) {
         _merge = false;
     }
+
+    if (force_merge) {
+        _merge = true;
+    }
```
HappenLee, line 48: _merge = force_merge

be/src/vec/olap/vgeneric_iterators.cpp
```diff
@@ -177,7 +178,7 @@ class VMergeIteratorContext {
         if (_is_unique) {
             result ? this->set_skip(true) : rhs.set_skip(true);
         }
-        return result;
+        return UNLIKELY(_is_reverse) ? !result : result;
```
HappenLee, line 181: do not need reverse in key equals case

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -104,6 +107,43 @@ class SegmentIterator::BitmapRangeIterator {
     bool _eof = false;
 };
 
+// A backward range iterator for roaring bitmap. Output ranges use closed-open form, like [from, to).
+// Example:
+//   input bitmap:  [0 1 4 5 6 7 10 15 16 17 18 19]
+//   output ranges: , [15,20), [10,11), [4,8), [0,2) (when max_range_size=10)
+//   output ranges: [18,20), [15,18), [10,11), [4,8), [0,2) (when max_range_size=3)
```
HappenLee, line 114: rethink the logic?

be/src/olap/rowset/rowset_reader_context.h
```diff
@@ -34,6 +34,9 @@ struct RowsetReaderContext {
     const TabletSchema* tablet_schema = nullptr;
     // whether rowset should return ordered rows.
     bool need_ordered_result = true;
+    //
```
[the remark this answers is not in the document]
xiaokang: yes, added

be/src/vec/olap/vcollect_iterator.h
xiaokang: added comment

be/src/olap/reader.cpp
```diff
@@ -197,11 +197,17 @@ Status TabletReader::_capture_rs_readers(const ReaderParams& read_params,
             // it's ok for rowset to return unordered result
             need_ordered_result = false;
         }
+
+        if (read_params.read_orderby_key) {
+            need_ordered_result = true;
```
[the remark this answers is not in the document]
xiaokang: I meant to emphasize read_orderby_key

be/src/olap/rowset/beta_rowset_reader.cpp
xiaokang: removed read_context->read_orderby_key and optimized the new_merge_iterator/new_union_iterator as we discussed.

be/src/vec/olap/vcollect_iterator.cpp
xiaokang: fixed

be/src/vec/olap/vcollect_iterator.cpp
xiaokang: can not just set _merge = force_merge, since when force_merge == false there is a chance _merge is already set to true before this statement.

be/src/vec/olap/vgeneric_iterators.cpp
xiaokang: fixed

be/src/olap/rowset/segment_v2/segment_iterator.cpp
xiaokang: changed the example

xiaokang:
Thansk @HappenLee @jackwener for careful review. The problem is fixed as we discuss.

be/src/olap/reader.cpp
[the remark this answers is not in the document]
xiaokang: btw, for normal query on UNIQUE_KEYS, need_ordered_result is true but read_params.read_orderby_key is false.

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -915,7 +959,8 @@ Status SegmentIterator::_read_columns_by_index(uint32_t nrows_read_limit, uint32
         } else {
             nrows_read += rows_to_read;
         }
-    } while (nrows_read < nrows_read_limit);
+        // if _opts.read_orderby_key_reverse is true, only read one range for fast reverse purpose
```
HappenLee, line 962: I don't quite understand the intent here? Can you explain it in detail?

xiaokang: OK. It's related to the code 'reverse block row order' at the end of SegmentIterator::next_batch.

If we read more than one range in reverse order here, such as range10 with rows 9000~10000 and range9 with rows 8000~9000, the block will contain non-continous ordered rows 9000~10000,8000~9000. Then the reverse block at the end of next_batch will get rows 9000~8000,10000~9000. It's not totally in reversed order.

HappenLee (dismissed):
LGTM

jackwener (dismissed):
LGTM

yiguolei (approved):
LGTM

PR #10721, opened 2022-07-09
[Bug](tablet) Fix bug that segments are removed as trash but tablet meta is normal

### LINKED ISSUE #10720: [Bug] Segment files are removed as trash but tablet meta is normal
### Search before asking

### Version

trunk version. commit id : a5efda68829c0873800b62d7e2e2c3b1807d1734

### What's Wrong?

We replaced the original disk on the BE node with a new disk by wrong operation when BE restarted. When we discovered the mistake and added the original disk. There is no unhealthy replica in the cluster after a period of time, and we removed the wrong disk. When query comes, exception information is thrown and the error code is `-3109` which means `failed to open segment`. We found that segment files for some tablets had been removed as trash but tablet meta is normal in original disk. These abnormal tablets can not be detected and repaired by FE.

### What You Expected?

Tablet metadata should be consistent with data files for a tablet. When segment files removed as trash, the tablet should be droped on the BE node so that FE node could detect and repair the error replica.

### How to Reproduce?

Cluster: 1 FE + 3 BE (BE01, BE02 and BE03, there is one disk called `disk-1` on BE01.)

STEP 1: create a table on the cluster and ensure there are 3 replica for each tablet.

STEP 2: insert data into the table.

STEP 3: remove the `disk-1` on BE01 and add a new disk called `disk-2`, then restart BE01.
            When the deamon start, we will find that there is no replica on BE01 because there is only one empty disk which is `disk-2`, and the replica repair task will clone some replica to the `disk-2` on BE01.

STEP 4: When there is no unhealthy replica in the cluster after a period of time, add the `disk-1` and restart BE01(there is two disks which are `disk-1` and `disk-2`).
            When the deamon start, we will find tablets in `disk-2` would be load and tablets which hold the same id with that on `disk-2` will not be load. Data on different disks are loaded in parallel. If the later loaded tablet with same id is on `disk-1`(there is a tablet with same id on `disk-2` has been loaded before successfully), the tablet will not be loaded successfully. Segment files for this tablet would be removed as trash but metadata is normal.

STEP 5: remove the `disk-2` but keep `disk-1`, then restart BE01(there is only one empty disk which is `disk-1`).
           When the deamon start, we will find tablets in `disk-1` would be load successfully. These tablets hold normal metadata but has no segment files.

STEP 6: query the table. If the query falls on these replica on BE01, an exception will occur and the error code is `-3109` which means `failed to open segment`. These abnormal tablets can not be detected and repaired by FE.

### Anything Else?

NO.

### Code of Conduct

weizuo93:
# Proposed changes

Issue Number: close #10720 

## Problem Summary:

We replaced the original disk on the BE node with a new disk by wrong operation when BE restarted. When we discovered the mistake and added the original disk. There is no unhealthy replica in the cluster after a period of time, and we removed the wrong disk. When query comes, exception information is thrown and the error code is `-3109` which means `failed to open segment`. We found that segment files for some tablets had been removed as trash but tablet meta is normal in original disk. These abnormal tablets can not be detected and repaired by FE.

## Checklist(Required)

### DIFF
--- a/be/src/olap/tablet_manager.cpp
+++ b/be/src/olap/tablet_manager.cpp
@@ -161,6 +161,16 @@ Status TabletManager::_add_tablet_unlocked(TTabletId tablet_id, const TabletShar
         res = _add_tablet_to_map_unlocked(tablet_id, tablet, update_meta, keep_files,
                                           true /*drop_old*/);
     } else {
+        tablet->set_tablet_state(TABLET_SHUTDOWN);
+        tablet->save_meta();
+        {
+            std::lock_guard<std::shared_mutex> shutdown_tablets_wrlock(_shutdown_tablets_lock);
+            _shutdown_tablets.push_back(tablet);
+        }
+        LOG(INFO) << "set tablet to shutdown state."
+                  << "tablet_id=" << tablet->tablet_id()
+                  << ", tablet_path=" << tablet->tablet_path();
+
         res = Status::OLAPInternalError(OLAP_ERR_ENGINE_INSERT_OLD_TABLET);
     }
     LOG(WARNING) << "add duplicated tablet. force=" << force << ", res=" << res


be/src/olap/tablet_manager.cpp
```diff
@@ -154,13 +154,24 @@ Status TabletManager::_add_tablet_unlocked(TTabletId tablet_id, const TabletShar
...
-    bool keep_files = force ? true : false;
+    bool keep_files = force;
     if (force ||
         (new_version > old_version || (new_version == old_version && new_time > old_time))) {
         // check if new tablet's meta is in store and add new tablet's meta to meta store
         res = _add_tablet_to_map_unlocked(tablet_id, tablet, update_meta, keep_files,
                                           true /*drop_old*/);
     } else {
+        if (!keep_files) {
+            tablet->set_tablet_state(TABLET_SHUTDOWN);
```
morningman, line 165: I am confused that, this is a "add tablet" method, but why we finally set tablet to `TABLET_SHUTDOWN`?
Should it be `existed_tablet`?

weizuo93: > I am confused that, this is a "add tablet" method, but why we finally set tablet to `TABLET_SHUTDOWN`? Should it be `existed_tablet`?

@morningman 
Yes, it may be `existed_tablet` due to misoperation of the disk. You can refer to this issue #10720

be/src/olap/tablet_manager.cpp
```diff
@@ -154,13 +154,24 @@ Status TabletManager::_add_tablet_unlocked(TTabletId tablet_id, const TabletShar
...
     // here, the new rowset's file will also be dropped, so use keep files here
-    bool keep_files = force ? true : false;
+    bool keep_files = force;
     if (force ||
         (new_version > old_version || (new_version == old_version && new_time > old_time))) {
         // check if new tablet's meta is in store and add new tablet's meta to meta store
         res = _add_tablet_to_map_unlocked(tablet_id, tablet, update_meta, keep_files,
                                           true /*drop_old*/);
     } else {
+        if (!keep_files) {
```
dataroaring, line 164: keep_files must be false here.

weizuo93: > keep_files must be false here.

OK，Thank you.

caiconghui (approved):
LGTM

PR #10738, opened 2022-07-10
[Bug] CTAS varchar length lost

stalary:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:
1. CTAS varchar length lost because `ScalaType` isAssignedStrLenInColDefinition not serialized.
2. Support for decimal type, direct field maps can be used normally, and aggregated fields default to `decimal (27, 9)`.


yiguolei (approved):
lgtm

PR #10761, opened 2022-07-11
[Refactor](Nereids) remove generic type from concrete expressions

924060929:
# Proposed changes

Issue Number: no issue

## Problem Summary:

in the past, we use generic type for plan and expression to support pattern match framework, it can support type inference without unsafely type cast. then, we observed that expression usually traverse or rewrite by visitor pattern, so generic type is useless for expression and introduces complexity. so we remove generic type from concrete expressions.
 
## Checklist(Required)


EmmyMiao87 (dismissed):
LGTM

jackwener (dismissed):
LGTM

morrySnow (dismissed):

EmmyMiao87 (approved):
LGTM

PR #10772, opened 2022-07-11
fix Level1Iterator memory leak

xiaokang:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

VCollectIterator::Level1Iterator::~Level1Iterator() does not delete elements in _heap. Just check and delete like ~VMergeIterator().

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/vec/olap/vcollect_iterator.cpp
+++ b/be/src/vec/olap/vcollect_iterator.cpp
@@ -260,6 +260,14 @@ VCollectIterator::Level1Iterator::~Level1Iterator() {
             child = nullptr;
         }
     }
+
+    if (_heap) {
+        while (!_heap->empty()) {
+            auto child = _heap->top();
+            _heap->pop();
+            if (child) delete child;
+        }
+    }
 }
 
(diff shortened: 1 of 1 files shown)


be/src/vec/olap/vcollect_iterator.cpp
```diff
@@ -260,6 +260,14 @@ VCollectIterator::Level1Iterator::~Level1Iterator() {
             child = nullptr;
         }
     }
+
+    if (_heap) {
+        while (!_heap->empty()) {
+            auto child = _heap->top();
+            _heap->pop();
+            if (child) delete child;
```
yiguolei, line 268: all elements in _heap are also in _children. Already delete in line 257, maybe double delete here.

yiguolei (approved):
LGTM

be/src/vec/olap/vcollect_iterator.cpp
xiaokang: at line 327, _children.clear() is called after child is added to _heap.

PR #10788, opened 2022-07-12
[enhancement](singal) output git commit id when the program coredump

BePPPower:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

output git commit id when the program coredump。
like this:
image

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/common/signal_handler.h
+++ b/be/src/common/signal_handler.h
@@ -40,7 +40,7 @@
 #include <csignal>
 #include <ctime>
 
-#include "util/debug_util.h"
+#include "gen_cpp/version.h"
 #ifdef HAVE_UCONTEXT_H
 #include <ucontext.h>
 #endif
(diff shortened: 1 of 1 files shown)


be/src/common/signal_handler.h
```diff
@@ -254,6 +256,9 @@ void DumpTimeInfo() {
     formatter.AppendString(" try \"date -d @");
     formatter.AppendUint64(static_cast<uint64>(time_in_sec), 10);
     formatter.AppendString("\" if you are using GNU date ***\n");
+    formatter.AppendString("*** Current BE version: ");
```
yiguolei, line 259: Do not use this method. These code will allocate memory and string is very complicated. std::string get_build_version(bool compact) {
    std::stringstream ss;
    ss << DORIS_BUILD_VERSION
#ifdef NDEBUG
       << " RELEASE"
#else
       << " DEBUG"
#endif
       << " (build " << DORIS_BUILD_HASH << ")";

    if (!compact) {
        ss << std::endl << "Built on " << DORIS_BUILD_TIME << " by " << DORIS_BUILD_INFO;
    }

    return ss.str();
}

You could just use the macro

yiguolei (approved):
LGTM

PR #10843, opened 2022-07-14
[feature] (multi-catalog) read parquet file by start/offset

wsjz:
## Further comments
To avoid reading the repeat row group, we should align offsets
rw9PfsZ1MP


Zasek:
Hello, wsjz, excellent job!  My team is also optimizing the parquet file scanner/reader, but using a different way. We implemented this split read by modifying some Arrow source code. I hope we can discuss this topic😁, you may contact me via email.

morningman:
> Hello, wsjz, excellent job! My team is also optimizing the parquet file scanner/reader, but using a different way. We implemented this split read by modifying some Arrow source code. I hope we can discuss this topic😁, you may contact me via email.

Modifying Arrow source code is hard to maintain. But I am still interested in your approach. Could you share some doc about it?

morningman (approved):
LGTM

wsjz:
Test Param Control:
1.  create hms catalog and use it in current version.
2. split size setting at FE：Change the argument of inputFormat.getSplits(jobConf, 0) in ExternalHiveScanProvider. We can change the second argument '0' to 10 ，split will be divided into 100 parts, then each split  have the same size.
4. row group size setting(MB)：set parquet.block.size= N * 1024 * 1024 when load data with hive/spark.

dujl:
@wsjz @morningman  our parquet align strategy is not same as parquet community.
parquet community check whether the rowgroup's midPoint in the scan range.
if the row group's midpoint in the scan range,  will add the rowGroup to scan list.
Suggest that we align with the parquet community

For parquet
```
      long midPoint = startIndex + totalSize / 2;
      if (filter.contains(midPoint)) {
        newRowGroups.add(rowGroup);
      }
```
```
    public boolean contains(long offset) {
      return offset >= this.startOffset && offset < this.endOffset;
    }
```

PR #10901, opened 2022-07-15
[feature](resource-tag) support multi tag for a single Backend

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

Previously, Doris only support set one tag for a Backend, like: "tag.location" = "default".
This CL mainly changes:

Support multi tag for a single Backend

`alter system modify backend "host:port" set ("tag.location" = "default", "tag.compute" = "c1");`

And there are some limitations:

1. The tag of Backend must contain the tag type of location(tag.location).
     This ensures that the original logic will not change.
2. There can only be one tag type on the same BE.
    This ensures that the original logic for replica allocation remains unchanged.
3. And a new FE config `enable_multi_tags` to enable or disable it. Default is `false`.

## Checklist(Required)

3. Has unit tests been added: (Yes)
6. Are there any changes that cannot be rolled back: (Yes)
    The persist meta of Backend is changed.

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/system/Backend.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/system/Backend.java
@@ -119,13 +119,13 @@ public class Backend implements Writable {
     // additional backendStatus information for BE, display in JSON format
     @SerializedName("backendStatus")
     private BackendStatus backendStatus = new BackendStatus();
-    @SerializedName("tag")
     // the locationTag is also saved in tagMap, use a single field here to avoid
     // creating this everytime we get it.
+    @SerializedName(value = "locationTag", alternate = {"tag"})
     private Tag locationTag = Tag.DEFAULT_BACKEND_TAG;
-    @SerializedName("tagMap")
     // tag type -> tag value.
     // A backend can only be assigned to one tag type, and each type can only have one value.
+    @SerializedName("tagMap")
     private Map<String, String> tagMap = Maps.newHashMap();
 
     public Backend() {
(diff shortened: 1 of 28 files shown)


fe/fe-core/src/main/java/org/apache/doris/system/Backend.java
```diff
@@ -116,8 +119,14 @@ public enum BackendState {
     // additional backendStatus information for BE, display in JSON format
     @SerializedName("backendStatus")
     private BackendStatus backendStatus = new BackendStatus();
-    @SerializedName("tag")
-    private Tag tag = Tag.DEFAULT_BACKEND_TAG;
+    // the locationTag is also saved in tagMap, use a single field here to avoid
+    // creating this everytime we get it.
+    @SerializedName(value = "locationTag", alternate = {"tag"})
+    private Tag locationTag = Tag.DEFAULT_BACKEND_TAG;
```
compiletheworld, line 125: Will we delete this field `locationTag` in the future?
It may be confusing if we keep a special `entry` out of the `tagMap`.

compiletheworld:
LGTM

fe/fe-core/src/main/java/org/apache/doris/system/Backend.java
morningman: It is ok to leave this field here. Because the locationTag is used very frequent.
We don't need to create it every time we get it.
And in future, I will remove `@ SerializedName` annotation of it because it has been saved in `tagMap`

yiguolei (approved):
LGTM

PR #10921, opened 2022-07-16
[feature-wip](unique-key-merge-on-write) Add agg cache for delete bitmap DSIP-018

compiletheworld:
# Proposed changes

## Checklist(Required)

### DIFF
--- a/be/src/olap/tablet_meta.cpp
+++ b/be/src/olap/tablet_meta.cpp
@@ -51,7 +51,7 @@ Status TabletMeta::create(const TCreateTabletReq& request, const TabletUid& tabl
 }
 
 TabletMeta::TabletMeta()
-        : _tablet_uid(0, 0), _schema(new TabletSchema), _delete_bitmap(new DeleteBitmap()) {}
+        : _tablet_uid(0, 0), _schema(new TabletSchema), _delete_bitmap(new DeleteBitmap(_tablet_id)) {}
 
 TabletMeta::TabletMeta(int64_t table_id, int64_t partition_id, int64_t tablet_id,
                        int64_t replica_id, int32_t schema_hash, uint64_t shard_id,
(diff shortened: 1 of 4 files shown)


xinyiZzz:
LGTM, good code

be/src/olap/tablet_meta.cpp
```diff
@@ -869,4 +876,55 @@ void DeleteBitmap::merge(const DeleteBitmap& other) {
...
+    AggCache::Value* val = handle == nullptr
+                                   ? nullptr
+                                   : reinterpret_cast<AggCache::Value*>(_agg_cache->value(handle));
+    static auto ts_ms = [] {
+        return std::chrono::duration_cast<std::chrono::milliseconds>(
+                       std::chrono::system_clock::now().time_since_epoch())
+                .count();
+    };
+    // Renew if needed
+    if (val == nullptr || (val->expiration > 0 && val->expiration < ts_ms())) {
```
xinyiZzz, line 895: expiration evict policy. maybe it can be added to LRUCache later.

compiletheworld: Yes, good point.
There is a chance that the expired cache entry can be added back to the cache, however, the precondition is that nothing has changed since it has been added. That is to say, the loop of L907 still needs to be run, which seems almost the same cost?

zhannngchen (dismissed):
LGTM

kpfly (dismissed):
LGTM

dataroaring (dismissed):

be/src/olap/tablet_meta.h
```diff
@@ -353,6 +353,69 @@ class DeleteBitmap {
...
+     *
+     * Note: aggregation cache may be used.
+     *
+     * @return true if marked deleted
+     */
+    bool contains_agg(const BitmapKey& bitmap, uint32_t row_id) const;
+
+    /**
+     * Gets aggregated delete_bitmap on rowset_id and version, the same effect:
+     * `select sum(roarging::Roaring) where RowsetId=rowset_id and SegmentId=seg_id and Version <= version`
```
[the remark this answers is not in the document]
compiletheworld: Fixed

dataroaring (approved):

dataroaring (approved):
LGTM

PR #10924, opened 2022-07-16
[Enhancement] [Memory] Limit memory usage use process actual physical memory

### LINKED ISSUE #10923: [Enhancement] [Memory] Limit memory usage use process actual physical memory
### Search before asking

### Description

Currently, the mem tracker uses the tcmalloc hook and the tcmalloc property `generic.total_physical_bytes` to limit process memory usage.

Both tcmalloc hook and `generic.total_physical_bytes` record the total length of virtual memory of the process malloc, not the physical memory actually used by the process in the OS.

When there is a large amount of memory that is not used after application (this is abnormal code), the virtual memory of the process will be much larger than the physical memory actually used by the process, causing the mem tracker to limit the memory too strictly.

### Solution

_No response_

### Code of Conduct

xinyiZzz:
# Proposed changes

Issue Number: close #10923

### solution

Limit process memory usage using the actual physical memory of the process in `/proc/self/status`.
This is independent of the consumption value of the mem tracker, which counts the virtual memory of the process malloc.

### motivation

Currently, the mem tracker uses the tcmalloc hook and the tcmalloc property `generic.total_physical_bytes` to limit process memory usage.

Both the tcmalloc hook and `generic.total_physical_bytes` record the total length of virtual memory of the process malloc, not the physical memory actually used by the process in the OS.

When there is a large amount of memory that is not used after application (this is abnormal code), the virtual memory of the process will be much larger than the physical memory actually used by the process, causing the mem tracker to limit the memory too strictly.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/http/action/compaction_action.h
+++ b/be/src/http/action/compaction_action.h
@@ -39,12 +39,7 @@ const std::string PARAM_COMPACTION_CUMULATIVE = "cumulative";
 /// See compaction-action.md for details.
 class CompactionAction : public HttpHandler {
 public:
-    CompactionAction(CompactionActionType type) : _type(type) {
-        _compaction_mem_tracker =
-                type == RUN_COMPACTION ? MemTracker::create_tracker(-1, "ManualCompaction", nullptr,
-                                                                    MemTrackerLevel::VERBOSE)
-                                       : nullptr;
-    }
+    CompactionAction(CompactionActionType type) : _type(type) {}
 
     virtual ~CompactionAction() {}
 
@@ -75,8 +70,6 @@ private:
     static std::mutex _compaction_running_mutex;
     /// whether there is manual compaction running
     static bool _is_compaction_running;
-    /// memory tracker
-    std::shared_ptr<MemTracker> _compaction_mem_tracker;
 };
 
 } // end namespace doris
--- a/be/src/olap/rowset/segment_v2/segment.cpp
+++ b/be/src/olap/rowset/segment_v2/segment.cpp
@@ -48,9 +48,10 @@ Status Segment::open(io::FileSystem* fs, const std::string& path, uint32_t segme
 Segment::Segment(uint32_t segment_id, const TabletSchema* tablet_schema)
         : _segment_id(segment_id), _tablet_schema(*tablet_schema) {
 #ifndef BE_TEST
-    _mem_tracker = StorageEngine::instance()->tablet_mem_tracker();
+    _mem_tracker = MemTracker::create_virtual_tracker(
+            -1, "Segment", StorageEngine::instance()->tablet_mem_tracker());
 #else
-    _mem_tracker = MemTracker::get_process_tracker();
+    _mem_tracker = MemTracker::create_virtual_tracker(-1, "Segment");
 #endif
 }
 
--- a/be/src/runtime/exec_env_init.cpp
+++ b/be/src/runtime/exec_env_init.cpp
@@ -183,7 +183,6 @@ Status ExecEnv::_init_mem_tracker() {
                      << ". Using physical memory instead";
         global_memory_limit_bytes = MemInfo::physical_mem();
     }
-    MemTracker::get_process_tracker()->set_limit(global_memory_limit_bytes);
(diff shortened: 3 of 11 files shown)


be/src/util/perf_counters.h
```diff
@@ -95,6 +97,26 @@ class PerfCounters {
...
 
+    // Refactor
+
+    struct ProcStatus {
+        int64_t vm_size = 0;
+        int64_t vm_peak = 0;
+        int64_t vm_rss = 0;
+    };
+
+    static int parser_int(const std::string& state_key);
```
xiaokang, line 108: parse_xxx is better

be/src/http/action/compaction_action.h
```diff
@@ -75,8 +70,6 @@ class CompactionAction : public HttpHandler {
     static std::mutex _compaction_running_mutex;
     /// whether there is manual compaction running
     static bool _is_compaction_running;
-    /// memory tracker
-    std::shared_ptr<MemTracker> _compaction_mem_tracker;
```
xiaokang, line 79: why remove _compaction_mem_tracker?

be/src/olap/rowset/segment_v2/segment.cpp
```diff
@@ -48,9 +48,10 @@ Status Segment::open(io::FileSystem* fs, const std::string& path, uint32_t segme
 Segment::Segment(uint32_t segment_id, const TabletSchema* tablet_schema)
         : _segment_id(segment_id), _tablet_schema(*tablet_schema) {
 #ifndef BE_TEST
-    _mem_tracker = StorageEngine::instance()->tablet_mem_tracker();
+    _mem_tracker = MemTracker::create_virtual_tracker(
+            -1, "Segment", StorageEngine::instance()->tablet_mem_tracker());
 #else
-    _mem_tracker = MemTracker::get_process_tracker();
```
xiaokang, line 53: should process_tracker be used in other places?

be/src/runtime/mem_tracker.h
```diff
@@ -112,7 +113,11 @@ class MemTracker {
     static std::shared_ptr<MemTracker> get_temporary_mem_tracker(const std::string& label);
 
     Status check_sys_mem_info(int64_t bytes) {
```
xiaokang, line 115: is bytes still virtual memory?

be/src/runtime/exec_env_init.cpp
```diff
@@ -183,7 +183,6 @@ Status ExecEnv::_init_mem_tracker() {
                      << ". Using physical memory instead";
         global_memory_limit_bytes = MemInfo::physical_mem();
     }
-    MemTracker::get_process_tracker()->set_limit(global_memory_limit_bytes);
```
xiaokang, line 186: so we do NOT limit total memory by process tracker, and it's just used to track virtual memory of process, right?

be/src/util/perf_counters.h
xinyiZzz: done

be/src/http/action/compaction_action.h
xinyiZzz: Because `_compaction_mem_tracker` is not used,
Compaction memory is tracked by `StorageEngine::instance()->compaction_mem_tracker()`.

be/src/olap/rowset/segment_v2/segment.cpp
xinyiZzz: `process_tracker` should not be consumed by manual consume/release,
Manual consumption in Segment should create a new virtual tracker
(I'm refactoring mem tracker to avoid this)

be/src/runtime/mem_tracker.h
xinyiZzz: yep, so it is still possible to predict OOM too early, when this virtual memory is not used.

be/src/runtime/exec_env_init.cpp
xinyiZzz: yes, I modified the process tracker comment.

yiguolei (approved):
LGTM

PR #10942, opened 2022-07-17
[feature-wip](multi-catalog) Optimize threads and thrift interface of FileScanNode

AshinGau:
# Proposed changes

Optimize threads and thrift interface of FileScanNode

## Problem Summary:

FileScanNode in be will launch as many threads as the number of splits.
The thrift interface of FileScanNode is excessive redundant.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/vec/exec/file_scan_node.cpp
+++ b/be/src/vec/exec/file_scan_node.cpp
@@ -17,12 +17,14 @@
 
 #include "vec/exec/file_scan_node.h"
 
+#include "common/config.h"
 #include "gen_cpp/PlanNodes_types.h"
 #include "runtime/mem_tracker.h"
 #include "runtime/runtime_state.h"
 #include "runtime/string_value.h"
 #include "runtime/tuple.h"
 #include "runtime/tuple_row.h"
+#include "util/priority_thread_pool.hpp"
 #include "util/runtime_profile.h"
 #include "util/thread.h"
 #include "util/types.h"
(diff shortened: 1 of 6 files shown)


be/src/vec/exec/file_scan_node.cpp
```diff
@@ -100,9 +102,33 @@ Status FileScanNode::start_scanners() {
...
+        } else {
+            PriorityThreadPool::WorkFunction task =
+                    std::bind(&FileScanNode::scanner_worker, this, i, _scan_ranges.size(),
+                              std::ref(_scanners_status[i]));
+            if (!thread_pool->offer(task)) {
+                submit_status = Status::Cancelled("Failed to submit scan task");
+            }
+        }
+        if (!submit_status.ok()) {
+            LOG(FATAL) << "Failed to assign file scanner task to thread pool! "
```
morningman, line 122: FATAL log will crash the BE process. Use WARN instead.

be/src/vec/exec/file_scan_node.cpp
```diff
@@ -100,9 +102,33 @@ Status FileScanNode::start_scanners() {
     }
 
     _scanners_status.resize(_scan_ranges.size());
+    ThreadPoolToken* thread_token = _runtime_state->get_query_fragments_ctx()->get_token();
+    PriorityThreadPool* thread_pool = _runtime_state->exec_env()->scan_thread_pool();
```
morningman, line 106: Not a good idea to use scan thread pool for external table scan.
But we can just leave it here. And later we should refactor this part.

be/src/vec/exec/file_scan_node.cpp
morningman: And if we meet error, no need to continue this loop.

be/src/vec/exec/file_scan_node.cpp
```diff
@@ -205,8 +231,9 @@ Status FileScanNode::close(RuntimeState* state) {
     _scan_finished.store(true);
     _queue_writer_cond.notify_all();
     _queue_reader_cond.notify_all();
-    for (int i = 0; i < _scanner_threads.size(); ++i) {
-        _scanner_threads[i].join();
+    {
+        std::unique_lock<std::mutex> l(_batch_queue_lock);
+        _queue_reader_cond.wait(l, [this] { return _num_running_scanners == 0; });
```
morningman, line 236: Why not just using `_scanner_threads[i].join();`?

AshinGau: Threads in the thread pool are reused and will not be terminated or released.

be/src/vec/exec/file_scan_node.cpp
AshinGau: done

morningman (approved):
LGTM

PR #10992, opened 2022-07-19
[regression] Revert Add ssb sf1 test under unique table with zstd

smallhibiscus:
Reverts apache/doris#10957


smallhibiscus:
The pipeline check fails and the pipeline check succeeds. This pr needs to be reverted to perform the pipeline check again.

dataroaring (approved):

PR #11001, opened 2022-07-19
[enhancement](be) remove useless lock usage

zhannngchen:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

Acquiring a write lock on TabletMeta's `_meta_lock` here is meaningless, since `_rs_metas` and `_stale_rs_metas` is protected by the `_meta_lock` of Tablet.
Using of this lock is confusing here, so we'd better to remove it.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/tablet_meta.cpp
+++ b/be/src/olap/tablet_meta.cpp
@@ -636,7 +636,6 @@ void TabletMeta::modify_rs_metas(const std::vector<RowsetMetaSharedPtr>& to_add,
 // an existing tablet before. Add after revise, only the passing "rs_metas"
 // is needed.
 void TabletMeta::revise_rs_metas(std::vector<RowsetMetaSharedPtr>&& rs_metas) {
-    std::lock_guard<std::shared_mutex> wrlock(_meta_lock);
     _rs_metas = std::move(rs_metas);
     _stale_rs_metas.clear();
 }


be/src/olap/tablet_meta.cpp
```diff
@@ -636,7 +636,6 @@ void TabletMeta::modify_rs_metas(const std::vector<RowsetMetaSharedPtr>& to_add,
 // an existing tablet before. Add after revise, only the passing "rs_metas"
 // is needed.
 void TabletMeta::revise_rs_metas(std::vector<RowsetMetaSharedPtr>&& rs_metas) {
-    std::lock_guard<std::shared_mutex> wrlock(_meta_lock);
```
zhannngchen, line 639: Acquiring a write lock on TabletMeta's `_meta_lock` here is meaningless, since `_rs_metas` and `_stale_rs_metas` is protected by the `_meta_lock` of Tablet.
Using of this lock is confusing here, so we'd better to remove it.
@yiguolei @morningman could you help to check it again?

yiguolei: I am not sure, but I think we'd better not merge this PR. 
These code is written by me. I will think about it why I use two locks here.

yiguolei (changes requested):
Pending. I will think about it.

PR #11022, opened 2022-07-20
[fix](config) Fix uninitialized config validator

compiletheworld:
If we don't set any validator, the original implementation will cause
seg-fault due to nullptr of `RegisterConfValidator::_s_field_validator`,
which initial value relies on the first config validator definition.

We need to skip finding validator from
uninitialized `RegisterConfValidator::_s_field_validator`.

# Proposed changes

## Problem Summary:

## Checklist(Required)


kpfly (approved):
LGTM

xinyiZzz (approved):

PR #11026, opened 2022-07-20
[fix] Move s3 fs connect outside the lock critical area

platoneko:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

1. Fix potential null client bug in S3FileSystem.
2. Move s3 fs connect outside the lock critical area, which may cause serious lock contention when create tablet.
3. Use S3Conf instead of map to create S3FileSystem.
4. Remove a useless ut.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/storage_policy_mgr.cpp
+++ b/be/src/olap/storage_policy_mgr.cpp
@@ -35,8 +35,12 @@ void StoragePolicyMgr::update(const std::string& name, StoragePolicyPtr policy) 
         DCHECK(s3_fs);
         s3_fs->set_ak(policy->s3_ak);
         s3_fs->set_sk(policy->s3_sk);
-        s3_fs->connect();
+        auto st = s3_fs->connect();
         it->second = std::move(policy);
+        if (!st.ok()) {
+            LOG(ERROR) << st;
+            return;
+        }
     } else {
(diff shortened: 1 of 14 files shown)


compiletheworld:
LGTM

be/src/olap/storage_policy_mgr.cpp
```diff
@@ -24,43 +24,60 @@
...
-        auto s3_fs = std::dynamic_pointer_cast<io::S3FileSystem>(
-                io::FileSystemMap::instance()->get(name));
-        DCHECK(s3_fs);
-        s3_fs->set_ak(policy->s3_ak);
-        s3_fs->set_sk(policy->s3_sk);
-        s3_fs->connect();
-        it->second = std::move(policy);
-    } else {
-        // can't find name's policy, so do nothing.
+void StoragePolicyMgr::update(const std::string& name, const StoragePolicyPtr& policy) {
```
morningman, line 27: Looks like the `update` and `refresh` policy process may be called concurrently?
If so, it is not safe to update `s3_fs` outside the lock

platoneko: i will fix soon

morningman (approved):
LGTM

PR #11073, opened 2022-07-21
[enhance](nereids): use optional for combine()

jackwener:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

use optional for combine()

If input list is empty, combine will return true. It's strange.

Use optional to correct it.

## Checklist(Required)

## Further comments


wangshuo128:
The current implementation looks reasonable to me. 
A literal true/false predicate for empty input conjunctive/disjunctive predicates is right. It has no effect on the input data of the filter when the input predicate list is empty. 
And you could check if the result is boolean literal, which has not much different from using an optional result.
A boolean literal result is much better because it has the opportunity to be optimized by boolean value rewrite rules further.

jackwener:
I don't think so, because this function is for combine expression by use `And` or `Or`.
Empty List is an abnormal input, it's invalid by using `LiteralTrue` as result.
And `LiteralTrue` defeat the intent of this function. We want to combine but return a `LiteralTrue`, it's strange.

PR #11102, opened 2022-07-22
[dependency](be) Add vectorscan for support hypserscan on ARM

xiaokang:
# Proposed changes

Issue Number: related to #9964

## Problem Summary:

Describe the overview of changes.

hyperscan is a high-performance regular expression matching library, but can not be used on ARM. [vectorscan](https://github.com/VectorCamp/vectorscan) is an ARM port for hyperscan, it can be used as a drop in replacement.

Since hyperscan is original created by Intel and is popular and mature on x86, so we just use vectorscan only for aarch64 when build thirdparty.

## Checklist(Required)

1. Does it affect the original behavior: (No)
2. Has unit tests been added: (No Need)
3. Has document been added or modified: (No Need)
4. Does it need to update dependencies: (Yes)
5. Are there any changes that cannot be rolled back: (No)

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #11116, opened 2022-07-22
Optimize regexp and like using hyperscan

xiaokang:
# Proposed changes

Issue Number: depends on #9964 #11102  

## Problem Summary:

Use hyperscan for high performance regular expression matching, which is used by regexp and like SQL functions.

## Checklist(Required)

## Further comments


xiaokang:
In the comparation test on a be log dataset, with 118547840 and 26GB, we see 2~10x speedup on both x86 and ARM server.

image

kpfly:
> In the comparation test on a be log dataset, with 118547840 rows and 26GB, we see 2~10x speedup on both x86 and ARM server.
> 
> image

excellent

dataroaring (approved):
LGTM

PR #11131, opened 2022-07-22
[Improvement](light-schema-change) support tablet schema cache

Lchangliang:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

The PR is for #10136. In that, every rowset will presist a tablet schema and use in memory by rowsetMetaPB. But When tablets and rowsets in tablet too more,  memory will has a lot of pressure. So we need a way to avoid it.

In this pr, it will implement a global tablet schema cache. Every rowset will hold a tablet schema sptr from cache. That mean even we has  tens of thousands of rowsets, if their schema is same, there will be one schema in memory.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/exec/olap_scanner.cpp
+++ b/be/src/exec/olap_scanner.cpp
@@ -83,7 +83,7 @@ Status OlapScanner::prepare(
             LOG(WARNING) << ss.str();
             return Status::InternalError(ss.str());
         }
-        _tablet_schema = _tablet->tablet_schema();
+        _tablet_schema.copy_from(*_tablet->tablet_schema());
         if (!_parent->_olap_scan_node.columns_desc.empty() &&
             _parent->_olap_scan_node.columns_desc[0].col_unique_id >= 0) {
             _tablet_schema.clear_columns();
@@ -289,7 +289,7 @@ Status OlapScanner::_init_return_columns(bool need_seq_col) {
     }
 
     // expand the sequence column
-    if (_tablet->tablet_schema().has_sequence_col() && need_seq_col) {
+    if (_tablet_schema.has_sequence_col() && need_seq_col) {
         bool has_replace_col = false;
         for (auto col : _return_columns) {
             if (_tablet_schema.column(col).aggregation() ==
--- a/be/src/olap/data_dir.cpp
+++ b/be/src/olap/data_dir.cpp
@@ -18,6 +18,7 @@
 #include "olap/data_dir.h"
 
 #include <ctype.h>
+#include <gen_cpp/olap_file.pb.h>
 #include <mntent.h>
 #include <stdio.h>
 #include <sys/file.h>
@@ -475,8 +476,8 @@ Status DataDir::load() {
         }
         if (rowset_meta->rowset_state() == RowsetStatePB::COMMITTED &&
             rowset_meta->tablet_uid() == tablet->tablet_uid()) {
-            if (!rowset_meta->get_rowset_pb().has_tablet_schema()) {
-                rowset_meta->set_tablet_schema(&tablet->tablet_schema());
+            if (!rowset_meta->tablet_schema()) {
+                rowset_meta->set_tablet_schema(tablet->tablet_schema());
                 RowsetMetaManager::save(_meta, rowset_meta->tablet_uid(), rowset_meta->rowset_id(),
                                         rowset_meta->get_rowset_pb());
             }
@@ -498,8 +499,8 @@ Status DataDir::load() {
             }
         } else if (rowset_meta->rowset_state() == RowsetStatePB::VISIBLE &&
                    rowset_meta->tablet_uid() == tablet->tablet_uid()) {
-            if (!rowset_meta->get_rowset_pb().has_tablet_schema()) {
-                rowset_meta->set_tablet_schema(&tablet->tablet_schema());
+            if (!rowset_meta->tablet_schema()) {
+                rowset_meta->set_tablet_schema(tablet->tablet_schema());
                 RowsetMetaManager::save(_meta, rowset_meta->tablet_uid(), rowset_meta->rowset_id(),
                                         rowset_meta->get_rowset_pb());
             }
--- a/be/src/olap/delta_writer.cpp
+++ b/be/src/olap/delta_writer.cpp
@@ -123,10 +123,11 @@ Status DeltaWriter::init() {
                                                                   _req.txn_id, _req.load_id));
     }
     // build tablet schema in request level
-    _build_current_tablet_schema(_req.index_id, _req.ptable_schema_param, _tablet->tablet_schema());
+    _build_current_tablet_schema(_req.index_id, _req.ptable_schema_param,
+                                 *_tablet->tablet_schema());
 
     RETURN_NOT_OK(_tablet->create_rowset_writer(_req.txn_id, _req.load_id, PREPARED, OVERLAPPING,
-                                                _tablet_schema.get(), &_rowset_writer));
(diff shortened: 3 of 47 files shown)


be/src/olap/tablet_schema_cache.h
```diff
@@ -0,0 +1,62 @@
...
+    static void create_global_schema_cache() {
+        DCHECK(_s_instance == nullptr);
+        static TabletSchemaCache instance;
+        _s_instance = &instance;
+    }
+
+    static TabletSchemaCache* instance() { return _s_instance; }
+
+    std::shared_ptr<TabletSchema> insert(const std::string& key) {
+        std::lock_guard guard(_mtx);
```
xinyiZzz, line 41: Consider using `parallel-hashmap`, which may have better performance.

And, it has `lazy_emplace_l`, `if_contains` to handle concurrency, which can avoid using locks manually. This looks more concise, although less well understood.

Be careful not to use `try_emplace_l`, this behavior is not as expected https://github.com/greg7mdp/parallel-hashmap/issues/161

be/src/exec/olap_scanner.cpp
```diff
@@ -83,7 +83,7 @@ Status OlapScanner::prepare(
             LOG(WARNING) << ss.str();
             return Status::InternalError(ss.str());
         }
-        _tablet_schema = _tablet->tablet_schema();
```
yiguolei, line 86: better not copy tablet_schema, because TableSchema does not define copy constructor.  There is a filed  TabletColumn* _parent = nullptr; in tablet column, need change it to shared ptr?

be/src/olap/rowset/rowset_meta.h
```diff
@@ -324,24 +337,38 @@ class RowsetMeta {
 
     int64_t newest_write_timestamp() const { return _rowset_meta_pb.newest_write_timestamp(); }
     void set_tablet_schema(const TabletSchema* tablet_schema) {
```
yiguolei, line 339: Do not pass in raw ptr here, could pass TabletSchemaSPtr if possible.

be/src/olap/data_dir.cpp
```diff
@@ -474,10 +475,12 @@ Status DataDir::load() {
         }
         if (rowset_meta->rowset_state() == RowsetStatePB::COMMITTED &&
             rowset_meta->tablet_uid() == tablet->tablet_uid()) {
-            if (!rowset_meta->get_rowset_pb().has_tablet_schema()) {
-                rowset_meta->set_tablet_schema(&tablet->tablet_schema());
+            if (!rowset_meta->tablet_schema()) {
```
yiguolei, line 478: There many cases that call get_rowset_pb. For example in txn manager. If you erase schema from pb then you should add the schema to pb every time.
I think it is better to return a new pb  and add tablet schema to the new pb during get rowset pb .

be/src/olap/delta_writer.cpp
```diff
@@ -120,7 +120,8 @@ Status DeltaWriter::init() {
                                                                   _req.txn_id, _req.load_id));
     }
     // build tablet schema in request level
-    _build_current_tablet_schema(_req.index_id, _req.ptable_schema_param, _tablet->tablet_schema());
+    _build_current_tablet_schema(_req.index_id, _req.ptable_schema_param,
```
yiguolei, line 123: _build_current_tablet_schema will also copy tablet schema.

be/src/exec/olap_scanner.cpp
yiguolei: Maybe is better to use serizalie to pb and init_from_pb to build a new tablet schema.

be/src/olap/rowset/rowset.cpp
```diff
@@ -32,7 +32,7 @@ Rowset::Rowset(const TabletSchema* schema, const std::string& tablet_path,
         _is_cumulative = version.first != version.second;
     }
     // build schema from RowsetMeta.tablet_schema or Tablet.tablet_schema
-    _schema = _rowset_meta->tablet_schema() != nullptr ? _rowset_meta->tablet_schema() : schema;
+    _schema = _rowset_meta->tablet_schema() ? _rowset_meta->tablet_schema().get() : schema;
```
yiguolei, line 35: should also change _schema in rowset.h to TabletSchemaSPtr. There maybe segment fault if we add clean logic in tablet schema cache.

be/src/olap/rowset/rowset_meta.h
```diff
@@ -245,7 +251,14 @@ class RowsetMeta {
...
-    void to_rowset_pb(RowsetMetaPB* rs_meta_pb) const { *rs_meta_pb = _rowset_meta_pb; }
+    void to_rowset_pb(RowsetMetaPB* rs_meta_pb) const {
+        *rs_meta_pb = _rowset_meta_pb;
+        if (_schema) {
+            _schema->to_schema_pb(rs_meta_pb->mutable_tablet_schema());
+        }
+    }
+
+    // warning！don't use tablet_schema in rowset_meta_pb
     const RowsetMetaPB& get_rowset_pb() { return _rowset_meta_pb; }
```
yiguolei, line 262: change to  const RowsetMetaPB get_rowset_pb()  and create a new pb here.

be/src/olap/rowset/rowset_writer_context.h
```diff
@@ -55,7 +55,7 @@ struct RowsetWriterContext {
         context.tablet_schema_hash = new_tablet->schema_hash();
         context.rowset_type = new_rowset_type;
         context.tablet_path = new_tablet->tablet_path();
-        context.tablet_schema = &(new_tablet->tablet_schema());
```
yiguolei, line 58: Pls modify RowsetWriterContext and use TabletSchemaSPtr.

be/src/olap/tablet.cpp
```diff
@@ -116,7 +116,7 @@ Status Tablet::_init_once_action() {
     for (const auto& rs_meta : _tablet_meta->all_rs_metas()) {
         Version version = rs_meta->version();
         RowsetSharedPtr rowset;
-        res = RowsetFactory::create_rowset(&_schema, _tablet_path, rs_meta, &rowset);
+        res = RowsetFactory::create_rowset(_schema.get(), _tablet_path, rs_meta, &rowset);
```
yiguolei, line 119: Change RowsetFactory::create_rowset method, using TabletSchemaSPtr as input parameter.

be/src/olap/schema_change.cpp
```diff
@@ -1809,7 +1811,7 @@ Status SchemaChangeHandler::_do_process_alter_tablet_v2(const TAlterTabletReqV2&
     // delete handlers for new tablet
     DeleteHandler delete_handler;
     std::vector<ColumnId> return_columns;
-    auto base_tablet_schema = base_tablet->tablet_schema();
+    auto base_tablet_schema = *base_tablet->tablet_schema();
```
yiguolei, line 1814: use copy from here

be/src/olap/schema_change.cpp
```diff
@@ -2365,7 +2364,7 @@ Status SchemaChangeHandler::_parse_request(
     }
 
     const TabletSchema& ref_tablet_schema = *base_tablet_schema;
-    const TabletSchema& new_tablet_schema = new_tablet->tablet_schema();
+    const TabletSchema& new_tablet_schema = *new_tablet->tablet_schema();
```
yiguolei, line 2367: use copy_from here

yiguolei (approved):
LGTM

PR #11142, opened 2022-07-23
[test] use suffix of directory as group name and use directory as dbname

dataroaring:
We can rename tpcds_sf1 to tpcds_sf1_p1, then tpcds_sf1 will be in group
p1.  We will group cases to p0, p1, p2, p3 in the future.

p0: function cases running in seconds.
p1: cases with expected out running in minutes, like tpcds_sf1
p2: cases with expected out running in hours, like tpcds_sf10 tpcds_sf100
p3: cases without without expected out to test core dump.

# Proposed changes

Issue Number: close #xxx

## Problem Summary:

## Checklist(Required)

## Further comments


regression-test/framework/src/main/groovy/org/apache/doris/regression/suite/SuiteContext.groovy
```diff
@@ -34,6 +34,7 @@ class SuiteContext implements Closeable {
     public final File file
     public final String suiteName
     public final String group
+    public final String dbName
```
924060929, line 37:
```suggestion
    public final String dbNameSuffix
```

924060929 (approved):
LGTM

yiguolei (approved):
LGTM

PR #11144, opened 2022-07-23
[Optimize](page cache): bypass page cache during compaction

compasses:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:
This PR help bypass the page cache during compaction. Before this , the index page cache and segment cache will be polluted during compaction, which would lead to query latency shaking when under high load of compaction. 

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/merger.cpp
+++ b/be/src/olap/merger.cpp
@@ -41,8 +41,9 @@ Status Merger::merge_rowsets(TabletSharedPtr tablet, ReaderType reader_type,
     reader_params.reader_type = reader_type;
     reader_params.rs_readers = src_rowset_readers;
     reader_params.version = dst_rowset_writer->version();
-
     reader_params.tablet_schema = cur_tablet_schema;
+    reader_params.use_page_cache = false; // shouldn't use page cache
+
     RETURN_NOT_OK(reader.init(reader_params));
 
     RowCursor row_cursor;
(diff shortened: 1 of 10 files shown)


be/src/olap/rowset/beta_rowset_reader.cpp
```diff
@@ -88,7 +88,7 @@ Status BetaRowsetReader::init(RowsetReaderContext* read_context) {
     // load segments
     RETURN_NOT_OK(SegmentLoader::instance()->load_segments(
             _rowset, &_segment_cache_handle,
-            read_context->reader_type == ReaderType::READER_QUERY));
```
dataroaring, line 91: for compaction, reader_type is not READER_QUERY, it does not use cache?

dataroaring (changes requested):

be/src/olap/rowset/beta_rowset_reader.cpp
compasses: Yes, here just make sure it follow the option and to prevent misuse. Since the reader type default value is *READER_QUERY*.

be/src/olap/merger.cpp
```diff
@@ -41,8 +41,9 @@ Status Merger::merge_rowsets(TabletSharedPtr tablet, ReaderType reader_type,
     reader_params.reader_type = reader_type;
     reader_params.rs_readers = src_rowset_readers;
     reader_params.version = dst_rowset_writer->version();
-
     reader_params.tablet_schema = cur_tablet_schema;
+    reader_params.use_page_cache = false; // shouldn't use page cache
```
morningman, line 45: No need to add this field. We can just use `reader_type` to judge whether to use page cache.
And for compaction operation, the reader_type is NOT `READER_QUERY`

compasses: Yes, I see. but this field is not a new one, here just make sure it was set properly. BTW, If we have option for page cache switch, I think it's better to treat the option with top priority.

PR #11169, opened 2022-07-25
[feature](FE): add new property to control whether use light schema change or not

ByteYue:
# Proposed changes

Issue Number: 

## Problem Summary:

Add one new property `light_schema_change` to control whether use the light schema change optimization. When it's set true, Doris would try light schema change when alter schema change job. It's turned  by default, which means user should use `light_schema_change = true` if they wan to use the optimization.

## Checklist(Required)

1. Does it affect the original behavior: Yes
4. Does it need to update dependencies: Yes

## Further comments


fe/fe-core/src/main/java/org/apache/doris/common/util/PropertyAnalyzer.java
```diff
@@ -451,8 +453,27 @@ public static long analyzeTimeout(Map<String, String> properties, long defaultTi
         return timeout;
     }
 
+    public static Boolean analyzeUseLightSchemaChange(Map<String, String> properties) throws AnalysisException {
+        String useSchemaChange = "";
```
dataroaring, line 457: Use boolean instead of String directly？

dataroaring (changes requested):

fe/fe-core/src/main/java/org/apache/doris/common/util/PropertyAnalyzer.java
ByteYue: I adapt this function using the similar logic as analyzeUniqueKeyMergeOnWrite below.

fe/fe-core/src/main/java/org/apache/doris/datasource/InternalDataSource.java
```diff
@@ -1718,6 +1713,26 @@ private void createOlapTable(Database db, CreateTableStmt stmt) throws UserExcep
         // this should be done before create partition.
         Map<String, String> properties = stmt.getProperties();
 
+        // get use light schema change
+        Boolean useLightSchemaChange = true;
+        try {
+            useLightSchemaChange = PropertyAnalyzer.analyzeUseLightSchemaChange(properties);
```
yiguolei, line 1719: Does useLightSchemaChange only control whether a column has unique id? What happens when alter table?

ByteYue: Yes, this new property only affects whether the column would be assigned a unique id or not.
Alter table job will try to use the light schema change optimization (like add new value column) if the columns of the tables are assigned unique id.

Lchangliang:
LGTM

stalary (dismissed):
LGTM

jacktengg:
Please fix FE code style check error.

fe/fe-core/src/main/java/org/apache/doris/analysis/ShowAlterStmt.java
```diff
@@ -104,12 +104,13 @@ private void getPredicateValue(Expr subExpr) throws AnalysisException {
         if (!(subExpr instanceof BinaryPredicate)) {
             throw new AnalysisException("The operator =|>=|<=|>|<|!= are supported.");
         }
+        LOG.info("enter getPredicateValue");
```
dataroaring, line 107: remove the log

fe/fe-core/src/main/java/org/apache/doris/common/util/TimeUtils.java
```diff
@@ -149,13 +152,22 @@ public static String longToTimeString(long timeStamp, SimpleDateFormat dateForma
         return dateFormat.format(new Date(timeStamp));
     }
 
-    public static String longToTimeString(long timeStamp) {
-        SimpleDateFormat datetimeFormatTimeZone = datetimeFormatThreadLocal.get();
+    public static String longToTimeStringProc(long timeStamp, SimpleDateFormat datetimeFormatTimeZone) {
```
dataroaring, line 155: longToTimeStringProc, the name is a little strange. Maybe we can rename it to longToTimeStringWithFormat?

fe/fe-core/src/main/java/org/apache/doris/datasource/InternalDataSource.java
```diff
@@ -1731,6 +1726,26 @@ private void createOlapTable(Database db, CreateTableStmt stmt) throws UserExcep
         // this should be done before create partition.
         Map<String, String> properties = stmt.getProperties();
 
+        // get use light schema change
+        Boolean useLightSchemaChange = true;
```
dataroaring, line 1730: default should be false？

fe/fe-core/src/main/java/org/apache/doris/datasource/InternalDataSource.java
```diff
@@ -1731,6 +1726,26 @@ private void createOlapTable(Database db, CreateTableStmt stmt) throws UserExcep
...
+        try {
+            useLightSchemaChange = PropertyAnalyzer.analyzeUseLightSchemaChange(properties);
+        } catch (AnalysisException e) {
+            throw new DdlException(e.getMessage());
+        }
+        // use light schema change optimization
+        olapTable.setUseLightSchemaChange(useLightSchemaChange);
+        if (useLightSchemaChange) {
+            LOG.info("table: {} uses light schema change", olapTable.getName());
```
dataroaring, line 1739: LOG.debug is enough.

dataroaring (changes requested):

fe/fe-core/src/main/java/org/apache/doris/datasource/InternalDataSource.java
ByteYue: Actually it wouldn't have any effect cause useLightSchemaChange would be assigned value by PropertyAnalyzer.analyzeUseLightSchemaChange which is set to false by default. But it's confusing indeed. Will change it soon.

dataroaring (approved):
LGTM

Issue #11115, opened 2022-07-22
[Bug] vectorized load use after free on master 


yixiutt:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Version

master

### What's Wrong?

#0  0x0000558c91016080 in doris::signal::(anonymous namespace)::FailureSignalHandler (signal_number=11, signal_info=0x7f9121eba5b0,
    ucontext=0x7f9121eba480) at /home/zcp/repo_center/doris_master/be/src/common/signal_handler.h:372
#1  <signal handler called>
#2  0x0000558c91016080 in doris::signal::(anonymous namespace)::FailureSignalHandler (signal_number=6, signal_info=0x7f9121ebac70,
    ucontext=0x7f9121ebab40) at /home/zcp/repo_center/doris_master/be/src/common/signal_handler.h:372
#3  <signal handler called>
#4  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#5  0x00007f92146b0859 in __GI_abort () at abort.c:79
#6  0x0000558c90fec42e in __sanitizer::Abort() ()
#7  0x0000558c90ff7d18 in __sanitizer::Die() ()
#8  0x0000558c90fd7ae4 in __asan::ScopedInErrorReport::~ScopedInErrorReport() ()
#9  0x0000558c90fd73a1 in __asan::ReportGenericError(unsigned long, unsigned long, unsigned long, unsigned long, bool, unsigned long, unsigned int, bool) ()
#10 0x0000558c90fd8117 in __asan_report_load8 ()
#11 0x0000558c93ecb11a in doris::vectorized::Value::get_value (this=0x7f90d1cbc048)
    at /home/zcp/repo_center/doris_master/be/src/vec/aggregate_functions/aggregate_function_window.h:214
#12 0x0000558c93f24b92 in doris::vectorized::LeadAndLagData<long, true, false, doris::vectorized::Value>::insert_result_into (
    this=0x7f90d1cbc048, to=...) at /home/zcp/repo_center/doris_master/be/src/vec/aggregate_functions/aggregate_function_window.h:271
[...]

PR #11182, opened 2022-07-25
always copy data to agg status

starocean999:
# Proposed changes

Issue Number: close (https://github.com/apache/doris/issues/11115)

## Problem Summary:

The agg data should always be copided instead of keep a column ptr + row pos. Because the column may be freed later, the agg data would have an invalid ptr.

## Checklist(Required)

## Further comments


BiteTheDDDDt:
It looks like this pr modifies the behavior of the `load`/`window function`/`reader` related functions.
I only see errors about `load` and `window functions` in #11115, Then maybe we shouldn't change the `reader` related function behavior.

@HappenLee @zhangstar333 PTAL

PR #11187, opened 2022-07-25
[Vectorized] Support order by aggregate function

### LINKED ISSUE #10332: [Feature][Function] support `Order by` aggregate function combinator
### Search before asking

- [X] #10341

### Description

Sometime, we want execute aggregation by some order. 
For example: 
`select group_concat(k2 order by k1,k2) from test;`
`select group_concat(k2 order by k2+k1) from test;`

### Design
I think we can implement a function combinator `order_by`, then we have function `order_by_group_concat(is_desc_string,order_by_arg(0)...,order_by_arg(n-1),origin_arg...)`.

`is_desc_string` is a string only contain `0` and `1`, is_desc_string[i] is `1` represent order_by_arg(i) is desc.

When we call `group_concat(k2 order by k2 desc,k1 asc)` when real call will is `order_by_group_concat(2,k2,true,k1,false,k2)`

Specifically, we can collect all data like `distinct` combinator and do merge, then sort data and add row one by one.

### Design On Other Community
ClickHouse is doing similar work now. [#36076](https://github.com/ClickHouse/ClickHouse/pull/36076)
They are designed like this: `SELECT * FROM groupArrayOrderBy('age asc, name desc')(id, age, name)`

Welcome to disscuss it.

related: #10315 #9744 #10004

HappenLee:
# Proposed changes

Issue Number: close #10332

## Problem Summary:

## Checklist(Required)

## Further comments


be/src/vec/aggregate_functions/aggregate_function_sort.h
```diff
@@ -155,13 +143,13 @@ class AggregateFunctionSort
 
     void insert_result_into(ConstAggregateDataPtr targetplace, IColumn& to) const override {
         auto place = const_cast<AggregateDataPtr>(targetplace);
-        if (!this->data(place).block.is_empty_column()) {
+        if (!this->data(place)._block.is_empty_column()) {
```
BiteTheDDDDt, line 146: Maybe we can just use `block` instead `_block` in struct.

be/src/vec/aggregate_functions/aggregate_function_sort.h
```diff
@@ -64,59 +72,34 @@ struct AggregateFunctionSortData {
...
 };
 
-template <int sort_column_size, template <int> typename Data>
+template <typename Data>
 class AggregateFunctionSort
-        : public IAggregateFunctionDataHelper<Data<sort_column_size>,
-                                              AggregateFunctionSort<sort_column_size, Data>> {
-    using DataReal = Data<sort_column_size>;
+        : public IAggregateFunctionDataHelper<Data, AggregateFunctionSort<Data>> {
+    using DataReal = Data;
```
BiteTheDDDDt, line 95: We can remove `DataReal` directly.

fe/fe-core/src/main/java/org/apache/doris/analysis/FunctionCallExpr.java
```diff
@@ -513,13 +538,14 @@ private void analyzeBuiltinAggFunction(Analyzer analyzer) throws AnalysisExcepti
                         "group_concat requires first parameter to be of type STRING: " + this.toSql());
             }
 
-            if (children.size() == 2) {
+            if (children.size() == 2 && orderByElements.isEmpty()) {
```
BiteTheDDDDt, line 541: Maybe `children.size()-orderByElements.size()==2` is better.

BiteTheDDDDt (approved):
LGTM

PR #11190, opened 2022-07-26
[fix] (mem tracker) Fix MemTracker accuracy

xinyiZzz:
# Proposed changes

Issue Number: close #xxx

#### 1 Accuracy Analysis of Tracking Memory Based on Hook
See: https://github.com/gperftools/gperftools/issues/1361
All memory of BE cannot be tracked through TCMalloc hook. There are diffs with `top -p pid VIRT` and `generic.total_physical_bytes`, which affect the observation effect, but do not affect the limit query memory usage. detailed:

1. I found that Hook and `generic.total_physical_bytes` can only track virtual memory usage.
2. Use `tc_nallocx(size, 0)` and `tc_malloc_size(ptr)` in Hook to calculate and record the memory of `new` and `delete`, which is different from `generic.total_physical_bytes`.
3. After a period of time, Hook and `generic.total_physical_bytes` have a big gap with the actual virtual memory `VIRT` of the process seen by `top -p pid`.

In future, based on Jemalloc's mem tracker, expect to be more accurate : )

#### 2 Fix `LoadChannel` memory tracking inaccurate
Restore the mem tracker level of `LoadChannel` `TabletsChannel` `DeltaWriter` so that `LoadChannelMgr` can handle mem exceed limit correctly.

In non-vectorized load, because DeltaWriter Write releases memory that does not belong to it, Hook automatically tracks that the memory is lower than the real value, so manually track memory in `Memtable`.

In vectorized loads, Hooks are used to automatically track memory usage, which is accurate.

#### 3 Other fix
Some bug fix and detail optimizations

#### 4 Performance
parallel_fragment_exec_instance_num=10

|  test  | proportion of tcmalloc hook in flame graph   |
|  ---- | ---- |
| ssb 2.1,  jmeter thread 1 | 1.1%  |
| ssb 3.1,  jmeter thread 1 |  1.3% |
| ssb 4.1,  jmeter thread 1 | 1.3%  |
| ssb 2.1,  jmeter thread 20 |  0.8% |
| ssb 4.1,  jmeter thread 20 |  0.7% |
| select * from T limit 10,  jmeter thread 100 | 1.9%  |
| load ssb lineorder |  1.2% |
| load big json dict | 4.3%  |

## Checklist(Required)

5. Has unit tests been added: (Yes)
6. Has document been added or modified: (No Need)
7. Does it need to update dependencies: (Yes)

## Further comments

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -651,7 +651,7 @@ CONF_Bool(memory_verbose_track, "false");
 // smaller than this value will continue to accumulate. specified as number of bytes.
 // Decreasing this value will increase the frequency of consume/release.
 // Increasing this value will cause MemTracker statistics to be inaccurate.
-CONF_mInt32(mem_tracker_consume_min_size_bytes, "4194304");
+CONF_mInt32(mem_tracker_consume_min_size_bytes, "1048576");
 
 // The version information of the tablet will be stored in the memory
 // in an adjacency graph data structure.
--- a/be/src/exec/broker_scan_node.cpp
+++ b/be/src/exec/broker_scan_node.cpp
@@ -373,6 +373,8 @@ Status BrokerScanNode::scanner_scan(const TBrokerScanRange& scan_range,
 }
 
 void BrokerScanNode::scanner_worker(int start_idx, int length) {
+    SCOPED_ATTACH_TASK(_runtime_state);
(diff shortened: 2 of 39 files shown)


yiguolei (approved):
LGTM

PR #11301, opened 2022-07-28
[bug](scanner) Improve limit query performance on olapScannode and avoid infinite loop

yangzhg:
# Proposed changes
1. fix a bug  that query large column table may cause infinite loop
3. Optimize the query logic with limit, for the case where the limit value is relatively small, reduce the parallelism of the scanner, reduce unnecessary resource consumption, and increase the number of similar queries that the system can carry at the same time, and increase the query speed by more than 60%

before:
image

after:

image

## Checklist(Required)

1. Type of your changes:
    - [x] Improvement
    - [ ] Fix
    - [ ] Feature-WIP
    - [ ] Feature
    - [ ] Doc
    - [ ] Refator
    - [ ] Others: 
2. Does it affect the original behavior: 
    - [ ] Yes
    - [x] No
    - [ ] I don't know
4. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No

## Further comments

### DIFF
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -95,7 +95,8 @@ Status OlapScanNode::init(const TPlanNode& tnode, RuntimeState* state) {
 
         _runtime_filter_ctxs[i].runtimefilter = runtime_filter;
     }
-
+    _batch_size = _limit == -1 ? state->batch_size()
+                               : std::min(static_cast<int64_t>(state->batch_size()), _limit);
     return Status::OK();
 }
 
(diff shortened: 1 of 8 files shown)


be/src/exec/olap_scan_node.cpp
```diff
@@ -95,7 +95,8 @@ Status OlapScanNode::init(const TPlanNode& tnode, RuntimeState* state) {
 
         _runtime_filter_ctxs[i].runtimefilter = runtime_filter;
     }
-
+    _batch_size = _limit == -1 ? state->batch_size()
```
yiguolei, line 98: Is this only effective on non-vectorized engine?

yangzhg: on vec-exec has the same stmt already

xy720 (approved):
LGTM

PR #11308, opened 2022-07-28
[Bug][Function] core dump on sum(distinct)

BiteTheDDDDt:
# Proposed changes

Issue Number: close #11302

## Checklist(Required)

1. Type of your changes:
    - [ ] Improvement
    - [X] Fix
    - [ ] Feature-WIP
    - [ ] Feature
    - [ ] Doc
    - [ ] Refator
    - [ ] Others: 
2. Does it affect the original behavior: 
    - [ ] Yes
    - [X] No
    - [ ] I don't know
3. Has unit tests been added:
    - [X] Yes
    - [ ] No
    - [ ] No Need
4. Has document been added or modified:
    - [ ] Yes
    - [ ] No
    - [X] No Need
5. Does it need to update dependencies:
    - [ ] Yes
    - [X] No
6. Are there any changes that cannot be rolled back:
    - [ ] Yes
    - [X] No

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #11257, opened 2022-07-27
[improvement]Use phmap::flat_hash_set in AggregateFunctionUniq

mrhhsg:
# Proposed changes

Issue Number: close #xxx

## Problem Summary:

Test on clikcbench with SQL:
```sql
SELECT COUNT(DISTINCT SearchPhrase), count(distinct userid) FROM hits;
```
Execution time reduced to 7.60 sec from 9.73 sec.

## Checklist(Required)

1. Type of your changes:
    - [x] Improvement
    - [ ] Fix
    - [ ] Feature-WIP
    - [ ] Feature
    - [ ] Doc
    - [ ] Refator
    - [ ] Others: 
2. Does it affect the original behavior: 
    - [x] Yes
    - [ ] No
    - [ ] I don't know
3. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
4. Has document been added or modified:
    - [ ] Yes
    - [ ] No
    - [x] No Need
5. Does it need to update dependencies:
    - [ ] Yes
    - [x] No
6. Are there any changes that cannot be rolled back:
    - [x] Yes
    - [ ] No

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #11356, opened 2022-07-30
[revert](coredump) "[improvement]Use phmap::flat_hash_set in AggregateFunctionUniq"

morningman:
Reverts apache/doris#11257
This commit will cause the regression test added in #11308


yiguolei (approved):
LGTM

PR #11366, opened 2022-07-31
[chore](thirdparty) env.sh may report error while JAVA_HOME is not set

0xez:
If "JAVA_HOME" is not set, env.sh will set "JAVACMD" as "$(which java)" which may cause error in some cases.

### DIFF
--- a/env.sh
+++ b/env.sh
@@ -122,7 +122,7 @@ if test -z "${BUILD_THIRDPARTY_WIP}"; then
 
     # check java home
     if [ -z "$JAVA_HOME" ]; then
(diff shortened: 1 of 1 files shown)


stalary (approved):

jackwener (approved):
Good catch! Thanks for your contribution❤️.

yangzhg (approved):
LGTM

0xez:
Many thanks to [stalary](https://github.com/stalary), [jackwener](https://github.com/jackwener) and [yangzhg](https://github.com/yangzhg).

PR #11400, opened 2022-08-01
[Improvement](regresstion test) Fix regression test case failure for ASAN build

jacktengg:
Fix mem limit exceed error for nereids test cases for ASAN build and fix compile errors for ASAN build.

# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/CMakeLists.txt
+++ b/be/CMakeLists.txt
@@ -517,7 +517,7 @@ set(CXX_FLAGS_DEBUG "${CXX_GCC_FLAGS} -O0")
 #   -O3: Enable all compiler optimizations
 #   -DNDEBUG: Turn off dchecks/asserts/debug only code.
 set(CXX_FLAGS_RELEASE "${CXX_GCC_FLAGS} -O3 -DNDEBUG")
-SET(CXX_FLAGS_ASAN "${CXX_GCC_FLAGS} -O3 -fsanitize=address -DADDRESS_SANITIZER")
+SET(CXX_FLAGS_ASAN "${CXX_GCC_FLAGS} -O0 -fsanitize=address -DADDRESS_SANITIZER")
 SET(CXX_FLAGS_LSAN "${CXX_GCC_FLAGS} -O0 -fsanitize=leak -DLEAK_SANITIZER")
 
 # Set the flags to the undefined behavior sanitizer, also known as "ubsan"
--- a/be/CMakeLists.txt
+++ b/be/CMakeLists.txt
@@ -517,7 +517,7 @@ set(CXX_FLAGS_DEBUG "${CXX_GCC_FLAGS} -O0")
 #   -O3: Enable all compiler optimizations
 #   -DNDEBUG: Turn off dchecks/asserts/debug only code.
 set(CXX_FLAGS_RELEASE "${CXX_GCC_FLAGS} -O3 -DNDEBUG")
-SET(CXX_FLAGS_ASAN "${CXX_GCC_FLAGS} -O0 -fsanitize=address -DADDRESS_SANITIZER")
+SET(CXX_FLAGS_ASAN "${CXX_GCC_FLAGS} -O3 -fsanitize=address -DADDRESS_SANITIZER")
 SET(CXX_FLAGS_LSAN "${CXX_GCC_FLAGS} -O0 -fsanitize=leak -DLEAK_SANITIZER")
 
 # Set the flags to the undefined behavior sanitizer, also known as "ubsan"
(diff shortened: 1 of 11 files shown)


hello-stephen:
LGTM

by the way, comment out q72 temporarily also?

be/CMakeLists.txt
```diff
@@ -517,7 +517,7 @@ set(CXX_FLAGS_DEBUG "${CXX_GCC_FLAGS} -O0")
 #   -O3: Enable all compiler optimizations
 #   -DNDEBUG: Turn off dchecks/asserts/debug only code.
 set(CXX_FLAGS_RELEASE "${CXX_GCC_FLAGS} -O3 -DNDEBUG")
-SET(CXX_FLAGS_ASAN "${CXX_GCC_FLAGS} -O0 -fsanitize=address -DADDRESS_SANITIZER")
+SET(CXX_FLAGS_ASAN "${CXX_GCC_FLAGS} -O3 -fsanitize=address -DADDRESS_SANITIZER")
```
adonis0147, line 520: `O3` may bring some obstacles to debug. What about `Og`?

jacktengg: O3 is for reducing online P0 test time in maximum for ASAN build, not for debug problems in the PR P0 test.
If some case failed, the PR committer should check and fix locally.

adonis0147: What about making it configurable? (Default: -O0)

yiguolei: @adonis0147 I think it makes no difference to a configuration or in cmakelist. We have too many configurations, it is like code.

adonis0147: > @adonis0147 I think it makes no difference to a configuration or in cmakelist. We have too many configurations, it is like code.

`run-be-ut.sh` use `ASAN`, this modification will affect the debugging experience for BE UT.

yiguolei: The default asan use O0, it is very slow. I think the better method it to use O3, then in most cases, it is very quickly to run both ut and P0. If the developer find the bugs, then he or she could change the default value then run it again.

yiguolei: I don't think we should provide all the configuations that user could modify as a configuration in command line. There are too many configurations that user or developer want to modify.

adonis0147: > The default asan use O0, it is very slow. I think the better method it to use O3, then in most cases, it is very quickly to run both ut and P0. If the developer find the bugs, then he or she could change the default value then run it again.

It is not a big deal, we can add an argument or an environment variable to the build.sh to control it.

adonis0147: > Could you please provide the command line how to use the configuration?

It is simple. If we pass multiple optimization levels to gcc, only the last one takes effect. We can just modify the [L551](https://github.com/apache/doris/pull/11400/files#diff-3507aac2aff9b5fe5f66d28967f3aa848491d4ced2466f6bf201ab3a97531837L551). Introduce a new variable and pass it from `build.sh` and `run-be-ut.sh`.

Reference:

> If you use multiple -O options, with or without level numbers, the last such option is the one that is effective.

[Optimize-Options](https://gcc.gnu.org/onlinedocs/gcc-12.1.0/gcc/Optimize-Options.html#Optimize-Options)

adonis0147: > I don't think we should provide all the configuations that user could modify as a configuration in command line. There are too many configurations that user or developer want to modify.

We can use `custom_env.sh`, these file can affect the environment variables before build, it is possible for user not to use an extra command argument.

adonis0147:
@jacktengg @yiguolei 

#11444 introduces a way to change the default optimization level.

The followings are the reasons why the default optimization level should be `-O0` or `-Og`:
1. For most users who want to build Doris from source, they can build it by `bash build.sh --be`. The build type of this command is `RELEASE` and the optimization level is `-O3`. All works fine as expect.
2. For most developers who develop Doris daily, `-O3` may harm the debugging experience. The standard edit-compile-debug cycle really matters which is the reason why `-Og` was introduced in `GCC`.
3. For other scenarios (workflows and etc.) , we can just modify the environment variables or deploy scripts to satisfy our requirement.

yiguolei (approved):
LGTM

yiguolei:
> 

Yes, great

PR #11404, opened 2022-08-01
[improve](be) Lock times limit

muyizi:
修改ChunkArena里的死等锁机制，为只尝试三次，经过测试，修改后内存分配的耗时比原来降低59%


muyizi:
实验数据：
image

jackwener (changes requested):
Good catch! Thanks for your contribution.

You can correct title with `[improve](be) .....`

jackwener:
BTW, you can use 

```bash
git remote add upstream https://github.com/apache/doris.git
git pull upstream/master --rebase
git push -f
```

to sync the master to follow upstream master.

git merge will make you branch mess.

be/src/runtime/memory/chunk_allocator.cpp
```diff
@@ -86,14 +88,23 @@ class ChunkArena {
         int idx = BitUtil::Log2Ceiling64(size);
         auto& free_list = _chunk_lists[idx];
 
-        std::lock_guard<SpinLock> l(_lock);
-        if (free_list.empty()) {
-            return false;
+        if (free_list.empty()) return false;
```
yiguolei, line 91: Is this thread safe?

be/src/runtime/memory/chunk_allocator.cpp
```diff
@@ -86,14 +88,23 @@ class ChunkArena {
...
+
+        for (int i = 0; i < TRY_LOCK_TIMES; ++i) {
+            if (_lock.try_lock()) {
```
yiguolei, line 94: Have you tested concurrency? For example, 10 threads call pop free chunk concurrently.
Because in spin lock it will try to call try lock 70 times and then wait.
    // Acquires the lock, spins until the lock becomes available
    void lock() {
        for (int spin_count = 0; !try_lock(); ++spin_count) {
            if (spin_count < NUM_SPIN_CYCLES) {
#if (defined(__i386) || defined(__x86_64__))
                asm volatile("pause\n" : : : "memory");
#elif defined(__aarch64__)
                asm volatile("yield\n" ::: "memory");
#endif
            } else {
                sched_yield();
                spin_count = 0;
            }
        }
    }

be/src/runtime/memory/chunk_allocator.cpp
zuochunwei: if free_list is empty just return

zuochunwei: is ok

be/src/runtime/memory/chunk_allocator.cpp
zuochunwei: 这里减少锁的尝试次数，多次尝试不成功就走malloc

muyizi:
> BTW, you can use
> 
> ```shell
> git remote add upstream https://github.com/apache/doris.git
> git pull upstream/master --rebase
> git push -f
> ```
> 
> to sync the master to follow upstream master.
> 
> git merge will make you branch mess.

ok，thank you

be/src/runtime/memory/chunk_allocator.cpp
muyizi: > 

上述结果是我用的100个线程并行进行测试的

yiguolei (approved):
LGTM

PR #11461, opened 2022-08-03
[improvement](agg)limit the output of agg node

starocean999:
# Proposed changes

Issue Number: close #xxx

## Problem summary
The limit operation can be done at agg node instead of last topN node. This can reduce the time cost to 1/3 of the old one. The sql is as bellow:
SELECT UserID, SearchPhrase, COUNT(*) FROM hits GROUP BY UserID, SearchPhrase LIMIT 10;

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/vec/exec/vaggregation_node.cpp
+++ b/be/src/vec/exec/vaggregation_node.cpp
@@ -27,7 +27,6 @@
 #include "vec/data_types/data_type_string.h"
 #include "vec/exprs/vexpr.h"
 #include "vec/exprs/vexpr_context.h"
-#include "vec/exprs/vslot_ref.h"
 #include "vec/utils/util.hpp"
 
 namespace doris::vectorized {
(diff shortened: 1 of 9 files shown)


be/src/vec/exec/vaggregation_node.cpp
```diff
@@ -897,7 +963,40 @@ Status AggregationNode::_pre_agg_with_serialized_key(doris::vectorized::Block* i
     return Status::OK();
 }
 
+Status AggregationNode::_execute_with_serialized_key_limited(Block* block) {
+    SCOPED_TIMER(_build_timer);
```
HappenLee, line 967: We should use templates to abstract the code of this part of the function to reduce the appearance of duplicate code
like
```
<bool limit>
_execute_with_serialized_key(Block* block) 
```

use limit to check whether to do `find` or `emplace`

be/src/vec/exec/vaggregation_node.cpp
```diff
@@ -1112,7 +1215,55 @@ Status AggregationNode::_serialize_with_serialized_key_result(RuntimeState* stat
     return Status::OK();
 }
 
+Status AggregationNode::_merge_with_serialized_key_limited(Block* block) {
+    SCOPED_TIMER(_merge_timer);
```
HappenLee, line 1219: same to the update

HappenLee (approved):
LGTM

PR #11554, opened 2022-08-05
add java format style

ucasfl:
Add java style in clang-format to avoid format error when writing java code:
image

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -14,3 +14,19 @@ ReflowComments: false
 SortUsingDeclarations: false
 SpacesBeforeTrailingComments: 1
 SpaceBeforeCpp11BracedList: true
+---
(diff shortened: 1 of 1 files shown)


jackwener (commented):
Fe use `checkstyle` check the code. This Pr is not compatible with it.

You can see [this](https://doris.apache.org/zh-CN/community/developer-guide/java-format-code).

ucasfl:
> Fe use `checkstyle` check the code. This Pr is not compatible with it.
> 
> You can see [this](https://doris.apache.org/zh-CN/community/developer-guide/java-format-code).

Sorry, I don't descriple it clearlly, its some editor will use this style file and clang-format to format code when writing code, but now it does not include java style in this file.

PR #11598, opened 2022-08-08
[fix](scanner) delete predicates might be inconsistent with rowset readers

zhannngchen:
# Proposed changes

Issue Number: close #xxx

## Problem summary

For now, the scanner capture rowset readers and init delete_handler in different place, although both of these operations acquired the _meta_lock of Tablet class, but they are separate.

So the following situation would happen:
1. scanner will read rowset A, B, C, D, rowset C has delete predicate
2. scanner capture the rowset reader for A, B, C, D
3. a base compaction just completed, merged A, B, C to C1, then removed the delete predicate of rowset C
4. scanner try to init the delete handler, but it can't get the delete predicate of rowset C 
5. scanner will read all data which should be deleted in rowset C, the result is incorrect

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
7. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
8. Has document been added or modified:
    - [ ] Yes
    - [x] No
    - [ ] No Need
9. Does it need to update dependencies:
    - [ ] Yes
    - [x] No
10. Are there any changes that cannot be rolled back:
    - [x] No

## Further comments

### DIFF
--- a/be/src/olap/merger.cpp
+++ b/be/src/olap/merger.cpp
@@ -41,6 +41,7 @@ Status Merger::merge_rowsets(TabletSharedPtr tablet, ReaderType reader_type,
     reader_params.reader_type = reader_type;
     reader_params.rs_readers = src_rowset_readers;
     reader_params.version = dst_rowset_writer->version();
+    reader_params.delete_predicates = tablet->delete_predicates();
 
     reader_params.tablet_schema = cur_tablet_schema;
     RETURN_NOT_OK(reader.init(reader_params));
@@ -103,6 +104,7 @@ Status Merger::vmerge_rowsets(TabletSharedPtr tablet, ReaderType reader_type,
     reader_params.rs_readers = src_rowset_readers;
     reader_params.version = dst_rowset_writer->version();
     reader_params.tablet_schema = cur_tablet_schema;
+    reader_params.delete_predicates = tablet->delete_predicates();
     reader_params.delete_bitmap = &tablet->tablet_meta()->delete_bitmap();
     if (stats_output && stats_output->rowid_conversion) {
         reader_params.record_rowids = true;
--- a/be/src/olap/merger.cpp
+++ b/be/src/olap/merger.cpp
@@ -41,10 +41,12 @@ Status Merger::merge_rowsets(TabletSharedPtr tablet, ReaderType reader_type,
     reader_params.reader_type = reader_type;
     reader_params.rs_readers = src_rowset_readers;
     reader_params.version = dst_rowset_writer->version();
-    reader_params.delete_predicates = tablet->delete_predicates();
-    std::copy(tablet->delete_predicates().cbegin(), tablet->delete_predicates().cend(),
-              std::inserter(reader_params.delete_predicates,
-                            reader_params.delete_predicates.begin()));
+    {
+        std::shared_lock rdlock(tablet->get_header_lock());
+        std::copy(tablet->delete_predicates().cbegin(), tablet->delete_predicates().cend(),
+                  std::inserter(reader_params.delete_predicates,
+                                reader_params.delete_predicates.begin()));
+    }
 
     reader_params.tablet_schema = cur_tablet_schema;
     RETURN_NOT_OK(reader.init(reader_params));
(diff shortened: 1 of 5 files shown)


be/src/vec/exec/volap_scanner.cpp
```diff
@@ -106,13 +106,11 @@ Status VOlapScanner::prepare(
...
             }
-        }
-    }
 
-    {
-        // Initialize tablet_reader_params
-        RETURN_IF_ERROR(
-                _init_tablet_reader_params(key_ranges, filters, bloom_filters, function_filters));
+            // Initialize tablet_reader_params
+            RETURN_IF_ERROR(_init_tablet_reader_params(key_ranges, filters, bloom_filters,
```
zhannngchen, line 111: @Gabriel39 could you help to review this PR? The comment of method `_init_tablet_reader_params` says it requires caller to hold a read lock of tablet, but our code actually not.

be/src/vec/exec/volap_scanner.cpp
```diff
@@ -195,6 +193,10 @@ Status VOlapScanner::_init_tablet_reader_params(
               std::inserter(_tablet_reader_params.function_filters,
                             _tablet_reader_params.function_filters.begin()));
 
+    std::copy(_tablet->delete_predicates().cbegin(), _tablet->delete_predicates().cend(),
```
Gabriel39, line 196: Would this cause a wrong delete predicate working on a new-arrival rowset?

If we start a query with scanning rowset A (with a delete predicate), and just copy its delete predicate here, but a new rowset B is inserted meanwhile, would this delete predicate work on rowset B?

be/src/olap/reader.cpp
```diff
@@ -613,15 +613,14 @@ Status TabletReader::_init_delete_condition(const ReaderParams& read_params) {
...
-        return _delete_handler.init(_tablet_schema, _tablet->delete_predicates(),
+        return _delete_handler.init(_tablet_schema, read_params.delete_predicates,
                                     read_params.version.second, this);
     };
 
     if (read_params.reader_type == READER_ALTER_TABLE) {
         return delete_init();
     }
 
-    std::shared_lock rdlock(_tablet->get_header_lock());
```
Gabriel39, line 624: Did this lock have any impact before?

be/src/vec/exec/volap_scanner.cpp
zhannngchen: The copied delete predicates only used for current scanner, in your case, current scanner can't see rowset
 B?

be/src/olap/reader.cpp
zhannngchen: `delete_init()` access `tablet->delete_predicates()` in previous code, the access should be protected by a read lock, and now we don't need to access `tablet->delete_predicates()` here any more, so the lock should be removed.

Gabriel39 (dismissed):
LGTM

dataroaring (dismissed):
LGTM

be/src/olap/merger.cpp
```diff
@@ -41,6 +41,10 @@ Status Merger::merge_rowsets(TabletSharedPtr tablet, ReaderType reader_type,
     reader_params.reader_type = reader_type;
     reader_params.rs_readers = src_rowset_readers;
     reader_params.version = dst_rowset_writer->version();
+    reader_params.delete_predicates = tablet->delete_predicates();
```
yiguolei, line 44: line 44 set reader_params.delete_predicates = tablet->delete_predicates(); and then add it again? Are there duplicate predicates?

be/src/olap/merger.cpp
```diff
@@ -41,6 +41,10 @@ Status Merger::merge_rowsets(TabletSharedPtr tablet, ReaderType reader_type,
...
+    std::copy(tablet->delete_predicates().cbegin(), tablet->delete_predicates().cend(),
+              std::inserter(reader_params.delete_predicates,
+                            reader_params.delete_predicates.begin()));
```
yiguolei, line 47: Why not acquire tablet lock during init reader_params's delete predicate property?

be/src/olap/reader.cpp
```diff
@@ -613,15 +613,14 @@ Status TabletReader::_init_delete_condition(const ReaderParams& read_params) {
...
 
     auto delete_init = [&]() -> Status {
-        return _delete_handler.init(_tablet_schema, _tablet->delete_predicates(),
+        return _delete_handler.init(_tablet_schema, read_params.delete_predicates,
                                     read_params.version.second, this);
     };
 
     if (read_params.reader_type == READER_ALTER_TABLE) {
         return delete_init();
     }
```
yiguolei, line 623: If not need acquire tablet lock here, then I think we do not need distinguish alter table and others. 620-623 could be deleted.

be/src/olap/merger.cpp
zhannngchen: forgot to delete this line, thanks for remind.

be/src/olap/merger.cpp
zhannngchen: 1. compaction don't need to care new added delete predicates, it's useless for this compaction.
2. only compaction would remove delete predicates, so it also don't need to worry other thread to remove the delete predicates related with input rowset.

be/src/olap/reader.cpp
zhannngchen: fixed

be/src/olap/merger.cpp
zhannngchen: I added the lock finally, it's more easier to understand and cost very low.

yiguolei (approved):
LGTM

PR #11631, opened 2022-08-09
[fix](like-predicate) Add missing functions in LikeColumnPredicate

mrhhsg:
# Proposed changes

Issue Number: close #xxx

## Problem summary

The `LikeColumnPredicater` in #10355 was incomplete.

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/olap/like_column_predicate.cpp
+++ b/be/src/olap/like_column_predicate.cpp
@@ -23,15 +23,30 @@
 
 namespace doris {
 
-LikeColumnPredicate::LikeColumnPredicate(bool opposite, uint32_t column_id,
-                                         doris_udf::FunctionContext* fn_ctx,
-                                         doris_udf::StringVal val)
+template <>
+LikeColumnPredicate<true>::LikeColumnPredicate(bool opposite, uint32_t column_id,
+                                               doris_udf::FunctionContext* fn_ctx,
+                                               doris_udf::StringVal val)
+        : ColumnPredicate(column_id, opposite),
+          _fn_ctx(fn_ctx),
+          pattern(reinterpret_cast<char*>(val.ptr), val.len) {
+    _state = reinterpret_cast<StateType*>(
+            _fn_ctx->get_function_state(doris_udf::FunctionContext::THREAD_LOCAL));
+    _state->search_state.clone(_like_state);
+}
+
+template <>
+LikeColumnPredicate<false>::LikeColumnPredicate(bool opposite, uint32_t column_id,
+                                                doris_udf::FunctionContext* fn_ctx,
+                                                doris_udf::StringVal val)
         : ColumnPredicate(column_id, opposite), _fn_ctx(fn_ctx), pattern(val) {
-    _state = reinterpret_cast<LikePredicateState*>(
+    _state = reinterpret_cast<StateType*>(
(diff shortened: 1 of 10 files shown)


be/src/olap/like_column_predicate.cpp
```diff
@@ -23,74 +23,131 @@
...
+template <>
+LikeColumnPredicate<true>::LikeColumnPredicate(bool opposite, uint32_t column_id,
+                                               doris_udf::FunctionContext* fn_ctx,
+                                               doris_udf::StringVal val)
+        : ColumnPredicate(column_id, opposite),
+          _fn_ctx(fn_ctx),
+          pattern(reinterpret_cast<char*>(val.ptr), val.len) {
+    _state = reinterpret_cast<StateType*>(
+            _fn_ctx->get_function_state(doris_udf::FunctionContext::THREAD_LOCAL));
+    _state->search_state.clone(_like_state);
```
compasses, line 35: So here clone the state because of the Hyperscan? Not sure the overhead of clone is high or not.

be/src/olap/like_column_predicate.h
```diff
@@ -45,33 +47,115 @@ class LikeColumnPredicate : public ColumnPredicate {
...
+                }
+            }
+        }
+    }
+
     std::string _origin;
     // life time controlled by scan node
     doris_udf::FunctionContext* _fn_ctx;
-    doris_udf::StringVal pattern;
+    using PatternType = std::conditional_t<is_vectorized, StringValue, StringVal>;
```
compasses, line 150: Confusing for the type StringValue and StringVal, especially for the starter.  It's better to keep one of them.

compasses:
Can you help add some regression test case for the like query? Need open the configuration ```enable_function_pushdown```, which default value is ```false```.

be/src/olap/like_column_predicate.cpp
mrhhsg: Yes, every concurrent caller of the Hyperscan API should use a separate scratch. Here each LikeColumnPredicate only clones it once, so I think the overload is acceptable.

mrhhsg:
> Can you help add some regression test case for the like query? Need open the configuration `enable_function_pushdown`, which default value is `false`.

@compasses 
I think using one session variable is better than a configuration item. I will create another PR to do this.

be/src/olap/like_column_predicate.h
mrhhsg: This confusion will be resolved after we remove the non-vectorized code.

be/src/olap/like_column_predicate.cpp
compasses: OK

be/src/olap/like_column_predicate.h
compasses: Got it

compasses (approved):
LGTM

yiguolei (approved):
LGTM

PR #11665, opened 2022-08-11
[Enhancement](stream-load-json) using simdjson to parse json

### LINKED ISSUE #11663: [Enhancement] support simdjson to parse json document when load
### Search before asking

### Description

provide simdjson to parse json document in json scanner

### Solution

Currently we use rapidjson to parse json document, It's fast but not fast enough compare to [simdjson](https://github.com/simdjson/simdjson/blob/master/doc/basics.md).And I found that the simdjson has a parsing front-end called simdjson::ondemand which will parse json when accessing fields and could strip the field token from the original document, using this feature we could reduce the cost of string copy(eg. we convert everthing to a string literal in _write_data_to_column by `sprintf`, I saw a hotspot from the flamegrame in this function, using simdjson::to_json_string will strip the token(a string piece) which is std::string_view and this is exactly we need).And second  in `_set_column_value` we could iterate through the json document by ` for (auto field: object_val) {xxx}`, this is much faster than looking up a field by it's field  name like `objectValue.FindMember("k1")`.The third optimization is the `at_pointer` interface[ simdjson provided](https://github.com/simdjson/simdjson/blob/master/doc/basics.md#json-pointer), this could directly get the json field from original document.

bellow is the performance result from my benchmark using stream load:
using `config::enable_simdjson_reader=true` to turn on simdjson reader to parse

image

### Code of Conduct

eldenmoon:
# Proposed changes

Currently we use rapidjson to parse json document, It's fast but not fast enough compare to [simdjson](https://github.com/simdjson/simdjson/blob/master/doc/basics.md).And I found that the simdjson has a parsing front-end called simdjson::ondemand which will parse json when accessing fields and could strip the field token from the original document, using this feature we could reduce the cost of string copy(eg. we convert everthing to a string literal in _write_data_to_column by sprintf, I saw a hotspot from the flamegrame in this function, using simdjson::to_json_string will strip the token(a string piece) which is std::string_view and this is exactly we need).And second in _set_column_value we could iterate through the json document by for (auto field: object_val) {xxx}, this is much faster than looking up a field by it's field name like objectValue.FindMember("k1").The third optimization is the at_pointer interface[ simdjson provided](https://github.com/simdjson/simdjson/blob/master/doc/basics.md#json-pointer), this could directly get the json field from original document.

Issue Number: close #11663

## Problem summary

using simdjson to parse in vjsonscanner.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [x] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments


yiguolei (approved):
LGTM

PR #11691, opened 2022-08-11
Fix file arrow scanner column index out of range core.

Jibing-Li:
# Proposed changes

```FileArrowScanner::_append_batch_to_block``` may core while executing tpch queries. It also may core while executing query like ```select count(1) from xxx```.

## Problem summary
The `get_column_index(column_name)` may return an index value that out of the `_batch` column range. For example, 
```select count(1) from xxx``` has only 1 column, but `get_column_index(column_name)` may return a value >= 1.

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/exec/arrow/arrow_reader.h
+++ b/be/src/exec/arrow/arrow_reader.h
@@ -100,6 +100,7 @@ public:
     int get_cloumn_index(std::string column_name);
 
     void prefetch_batch();
+    bool is_case_sensitive() {return _caseSensitive; }
 
 protected:
     virtual Status column_indices(const std::vector<SlotDescriptor*>& tuple_slot_descs);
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: be/src/exec/arrow/arrow_reader.h [hunk 1]
```diff
@@ -100,6 +100,7 @@ class ArrowReaderWrap {
     int get_cloumn_index(std::string column_name);
 
     void prefetch_batch();
+    bool is_case_sensitive() {return _caseSensitive; }
```
morningman, line 103:
```suggestion
    bool is_case_sensitive() { return _caseSensitive; }
```

Gabriel39, line 103: How about to use `_case_sensitive` instead?

Jibing-Li: I'll fix it.

morningman (approved):
LGTM

PR #11704, opened 2022-08-11
[enhancement](Status) Optimize Status implementation

### LINKED ISSUE #11698: [Enhancement] Separate code and err msg in Status and save err msg in heap
### Search before asking

### Description

The current `Status` implementation is to save all data to a fixed size (2KB) array on stack. This will lead to too large function stack space and may not be friendly to data locality (unnecessary page faults, TLB utilization, cacheline prefetching, etc.).
Another problem is when RVO(return value optimization) fails (there are many if-else branches in current codes, and the compiler may often be unable to do RVO), the move ctor cannot be used, causing the error message in `Status` to be copied layer by layer in the function call stack.
Doris moving the `Status`  data from heap to stack can be traced back to https://github.com/apache/doris/pull/8117. This PR is designed to avoid dynamic memory allocation. However, in practice, we often construct error messages through `fmt::format`(see `Status::ErrorFmt` https://github.com/apache/doris/blob/a6537a90cd4fda01de4a51b7abdc23a06cc8a901/be/src/common/status.h#L284-L289), so we cannot avoid dynamic memory allocation in fact.
If we separate the status code and err msg in `Status`:
1. The error msg constructed by `fmt::format` can be directly moved to `Status`.
2. When RVO fails, the compiler will use move ctor to pass `Status` to function caller to reduce the copy cost.

### Solution

_No response_

### Code of Conduct

platoneko:
# Proposed changes

Issue Number: close #11698 

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/common/status.cpp
+++ b/be/src/common/status.cpp
@@ -4,11 +4,12 @@
 
 #include "common/status.h"
 
-#include <glog/logging.h>
 #include <rapidjson/prettywriter.h>
 #include <rapidjson/stringbuffer.h>
 
 #include <boost/stacktrace.hpp>
+
+#include "gen_cpp/types.pb.h" // for PStatus
 
 namespace doris {
 
@@ -41,13 +42,13 @@ Status::Status(const TStatus& s) {
     if (s.status_code != TStatusCode::OK) {
         // It is ok to set precise code == 1 here, because we do not know the precise code
         // just from thrift's TStatus
-        if (s.error_msgs.empty()) {
-            assemble_state(s.status_code, Slice(), 1, Slice());
-        } else {
-            assemble_state(s.status_code, s.error_msgs[0], 1, Slice());
+        _code = s.status_code;
+        _precise_code = 1;
+        if (!s.error_msgs.empty()) {
+            _err_msg = s.error_msgs[0];
         }
     } else {
-        _length = 0;
+        _code = 0;
     }
 }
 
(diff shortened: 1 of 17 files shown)


be/src/common/status.cpp
```diff
@@ -58,18 +59,18 @@ Status::Status(const PStatus& s) {
...
+            _err_msg = s.error_msgs(0);
         }
     } else {
-        _length = 0;
+        _code = 0;
     }
 }
 
 // Implement it here to remove the boost header file from status.h to reduce precompile time
-Status Status::ConstructErrorStatus(int16_t precise_code, const Slice& msg) {
```
yiguolei, line 72: why remove msg here? I think the error should be  error message + error stack.

be/src/common/status.h
```diff
@@ -241,47 +240,36 @@ enum ErrorCode {
...
-        return *this;
-    }
+    Status& operator=(Status&& rhs) noexcept = default;
 
     // "Copy" c'tor from TStatus.
     Status(const TStatus& status);
 
     Status(const PStatus& pstatus);
 
+    Status(TStatusCode::type code, std::string_view msg, int16_t precise_code = 1)
```
yiguolei, line 263: If these method is public, then user could call this method directly. Error stack is missing.

be/src/common/status.h
```diff
@@ -291,125 +279,125 @@ class Status {
...
 
     template <typename... Args>
-    static Status Aborted(const std::string& fmt, Args&&... args) {
+    static Status Aborted(std::string_view fmt, Args&&... args) {
         return ErrorFmt(TStatusCode::ABORTED, fmt, std::forward<Args>(args)...);
     }
 
     template <typename... Args>
-    static Status DataQualityError(const std::string& fmt, Args&&... args) {
+    static Status DataQualityError(std::string_view fmt, Args&&... args) {
```
yiguolei, line 383: I find only OLAPInternalError has error stack, static Status DataQualityError(std::string_view fmt, Args&&... args) these method does not generate error stack.

yiguolei (approved):
LGTM

PR #11740, opened 2022-08-12
[enhancement](memtracker) Optimize query memory accuracy

### LINKED ISSUE #11738: [Enhancement] Improve query memory tracking accuracy
### Search before asking

### Description

The value of the query mem tracker is larger than the actual physical memory used by the query.

As a result, based on the value of the query mem tracker, it is impossible to estimate how many queries the cluster can parallelize.

### Solution

_No response_

### Code of Conduct

xinyiZzz:
# Proposed changes

Issue Number: close #11738

### motivation
The value of the query mem tracker is consistent with the physical memory actually used by the query.

### problem causes
Currently, only the virtual memory used by the query can be tracked through the tcmalloc hook. When the memory is not fully used after the application, the recorded virtual memory will be larger than the physical memory.

At present, it is mainly because PODArray does not memset 0 when applying for memory, and blocks applied for through PODArray in places such as VOlapScanNode::_free_blocks are usually used for memory reuse and cannot be fully used.

### Fix
The query mem tracker only records the peak memory used by PODArray and MemPool

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/vec/common/pod_array.h
+++ b/be/src/vec/common/pod_array.h
@@ -102,6 +102,7 @@ protected:
     char* c_start = null; /// Does not include pad_left.
     char* c_end = null;
     char* c_end_of_storage = null; /// Does not include pad_right.
+    char* c_end_peak = null;
 
     /// The amount of memory occupied by the num_elements of the elements.
     static size_t byte_size(size_t num_elements) { return num_elements * ELEMENT_SIZE; }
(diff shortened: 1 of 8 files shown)


be/src/vec/common/pod_array.h
```diff
@@ -111,15 +112,26 @@ class PODArrayBase : private boost::noncopyable,
         return byte_size(num_elements) + pad_right + pad_left;
     }
 
+    inline void reset_peak() {
+        if (c_end - c_end_peak > 1024) {
```
yiguolei, line 116: add unlikely here

be/src/runtime/memory/chunk_allocator.cpp
```diff
@@ -147,7 +147,7 @@ ChunkAllocator::ChunkAllocator(size_t reserve_limit)
 }
 
 Status ChunkAllocator::allocate(size_t size, Chunk* chunk) {
-    DCHECK(BitUtil::RoundUpToPowerOfTwo(size) == size);
+    DCHECK((size & (size - 1)) == 0);
```
yiguolei, line 150: CHECK here, not use DCHECK.

xinyiZzz: done

be/src/vec/common/pod_array.h
xinyiZzz: done

yiguolei (approved):
LGTM

PR #11741, opened 2022-08-12
[docs](fix)[fix some error path] 

daikon12:
# Proposed changes
1. thirdparty/include --> ../thirdparty/include
2.thirdparty/lib/libDorisUdf.a --> ../thirdparty/lib/libDorisUdf.a
3.${BUILD_DIR}/src/udf_samples --> src/udf_samples
4.${BUILD_DIR}/src/udf_samples --> src/udf_samples

## Problem summary

When I tried to compile the udf I wrote according to the official doc,  there would be an error when make. After careful inspection, I found that several paths in the doc may need to be adjusted。

## Checklist(Required)

1. Does it affect the original behavior:  NO
2. Has unit tests been added:NO
3. Has document been added or modified:YES
4. Does it need to update dependencies:NO
5. Are there any changes that cannot be rolled back:NO

### DIFF
--- a/docs/en/docs/ecosystem/udf/native-user-defined-function.md
+++ b/docs/en/docs/ecosystem/udf/native-user-defined-function.md
@@ -160,17 +160,17 @@ Running `sh build.sh` in the root directory of Doris will generate a static libr
 
     ```
     # Include udf
-    include_directories(thirdparty/include)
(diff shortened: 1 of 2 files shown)


dataalive (approved):
LGTM

hf200012 (approved):
LGTM

PR #11742, opened 2022-08-12
[enhancement](load) enhance load from parquet or orc file

GoGoWen:
# Proposed changes

enhance loading from parquet or orc file, when given column not exist in file, set to null default instead of fails with "Invalid Column"

## Problem summary
1, create table:
CREATE TABLE `t22` (
  `name` bigint(20) NOT NULL COMMENT "",
  `id` bigint(20) NOT NULL COMMENT "",
  `id2` bigint(20) NULL COMMENT "",
  `impressions` bigint(20) SUM NULL DEFAULT "0" COMMENT "用户总展现",
  `click` double SUM NULL DEFAULT "0" COMMENT "用户总点击",
  `cost` bigint(20) SUM NULL DEFAULT "0" COMMENT "用户总消费"
) ENGINE=OLAP
AGGREGATE KEY(`name`, `id`, `id2`)
COMMENT "OLAP"
PARTITION BY RANGE(`name`)
(PARTITION p201901 VALUES [("1"), ("100")))
DISTRIBUTED BY HASH(`id`) BUCKETS 16
PROPERTIES (
"replication_allocation" = "tag.location.default: 3",
"in_memory" = "false",
"storage_format" = "V2"
)
2, try to load data in parquet or orc  with  broker load 
2.1 data like below:
name,id,click,impressions,cost
1,111,1111,11111,111111
1,11,,1111,11111
22,222,2222,22222,22222
3,33,333,3333,33333
4,44,444,4444,44444
5,55,555,5555,55555

2.2 broker load without columns  like below:
LOAD LABEL label1 (DATA INFILE (hdfs://filepath) into table 't22' format as "parquet" with broker broker_name (....)

3, the result should like below instead of failed with "Invalid Column with ", the column id2 is NULL as it not exist in file.
+------+------+------+-------------+-------+--------+
| name | id   | id2  | impressions | click | cost   |
+------+------+------+-------------+-------+--------+
|   22 |  222 | NULL |       22222 |  2222 |  22222 |
|   11 |  111 | NULL |       11111 |  1111 | 111111 |
|    5 |   55 | NULL |        5555 |   555 |  55555 |
|    4 |   44 | NULL |        4444 |   444 |  44444 |
|    3 |   33 | NULL |        3333 |   333 |  33333 |
|    1 |   11 | NULL |        1111 |  NULL |  11111 |
+------+------+------+-------------+-------+--------+

Describe your changes.

## Checklist(Required)

    - [Y ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/exec/orc_scanner.cpp
+++ b/be/src/exec/orc_scanner.cpp
@@ -124,18 +124,6 @@ ORCScanner::~ORCScanner() {
 
 Status ORCScanner::open() {
     RETURN_IF_ERROR(BaseScanner::open());
-    if (!_ranges.empty()) {
-        std::list<std::string> include_cols;
-        TBrokerRangeDesc range = _ranges[0];
-        _num_of_columns_from_file = range.__isset.num_of_columns_from_file
-                                            ? range.num_of_columns_from_file
-                                            : _src_slot_descs.size();
-        for (int i = 0; i < _num_of_columns_from_file; i++) {
-            auto slot_desc = _src_slot_descs.at(i);
-            include_cols.push_back(slot_desc->col_name());
-        }
-        _row_reader_options.include(include_cols);
-    }
 
     return Status::OK();
 }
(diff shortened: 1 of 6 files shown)


dutyu (changes requested):
I think this is a great idea to avoid loading problems when hive table's fields have changed. And may be use a property likes 'strict mode' or something else to control the loading behavior is better （whether to set value to null or throw an exception when reader can not find the column in orc/parquet file）.

be/src/vec/exec/format/parquet/vparquet_reader.cpp
```diff
@@ -79,10 +79,8 @@ Status ParquetReader::_init_read_columns(const std::vector<SlotDescriptor*>& tup
         if (iter != _map_column.end()) {
             _include_column_ids.emplace_back(parquet_col_id);
         } else {
-            std::stringstream str_error;
-            str_error << "Invalid Column Name:" << slot_desc->col_name();
-            LOG(WARNING) << str_error.str();
-            return Status::InvalidArgument(str_error.str());
+            // just continue, as the invalid column will be set to null by default
```
morningman, line 82: Do we need to set to "default value" if there is a default value when creating table?

be/src/exec/orc_scanner.cpp
```diff
@@ -428,4 +442,38 @@ void ORCScanner::close() {
     _row_reader.reset(nullptr);
 }
 
+void ORCScanner::build_name_id_map() {
+    std::vector<std::string> columns;
```
dutyu, line 446: We should clear _map_column_to_id first by invoke `_map_column_to_id.clear()` here.

dutyu (changes requested):
There is a bug in orc_scanner.cpp, _map_column_to_id should be cleared before build.

dataalive:
is  there any update here ?

PR #11757, opened 2022-08-13
[fix](query) fix orderby keys limit return less or no result

xiaokang:
fix orderby keys limit return less or no result when a WHERE predicte filter most rows

# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

The bug is caused by use _num_rows_read for limit check. _num_rows_read is count of rows read from storage, but may be  filtered by filter_block for WHERE predicate.

Add a _num_rows_return, which is rows after filter_block for WHERE predicate, for count for really returned rows.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/vec/exec/volap_scanner.h
+++ b/be/src/vec/exec/volap_scanner.h
@@ -134,6 +134,7 @@ private:
     // time costed and row returned statistics
     int64_t _num_rows_read = 0;
     int64_t _raw_rows_read = 0;
+    int64_t _num_rows_return = 0;
     int64_t _compressed_bytes_read = 0;
 
     // number rows filtered by pushed condition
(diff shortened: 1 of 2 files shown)


be/src/vec/exec/volap_scanner.h
```diff
@@ -134,6 +134,7 @@ class VOlapScanner {
     // time costed and row returned statistics
     int64_t _num_rows_read = 0;
```
BiteTheDDDDt, line 135: Are `_num_rows_read` and `_raw_rows_read` always equal?

xiaokang: I'm not sure about this. But the code for update _num_rows_read and _raw_rows_read seems to be very different.

BTW, if you mean _num_rows_read and _num_rows_return, they are not the same. _num_rows_return is smaller than _num_rows_read, since _num_rows_return is counted after filter_block.

BiteTheDDDDt (approved):
LGTM

PR #11782, opened 2022-08-15
[fix](agg)disallow group by bitmap or hll data type

starocean999:
# Proposed changes

disallow "select distinct bitmap, hll from xxx", because in some case, the FE will introduce a agg node with group by bitmap or hll exprs to  implement the "distinct" operator. But group by bitmap or hll data type are not supported in BE.

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/AggregateInfo.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/AggregateInfo.java
@@ -21,7 +21,7 @@
 package org.apache.doris.analysis;
 
 import org.apache.doris.catalog.FunctionSet;
-import org.apache.doris.catalog.PrimitiveType;
+import org.apache.doris.catalog.Type;
 import org.apache.doris.common.AnalysisException;
 import org.apache.doris.common.util.VectorizedUtil;
 import org.apache.doris.planner.DataPartition;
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/org/apache/doris/analysis/AggregateInfo.java
```diff
@@ -160,6 +161,15 @@ public void setPartitionExprs(List<Expr> exprs) {
         partitionExprs = exprs;
     }
 
+    private static void validateGroupingExprs(List<Expr> groupingExprs) throws AnalysisException {
```
spaces-X, line 164: I think it is more reasonable to put this checking logic into `GroupByClause`'s analyze phase.

starocean999: Thanks for your time. But there is no group by clause in sql like "select distinct bitmap, hll from xxx", the agg node is created by fe to implement the "distinct" operator, so no analyze function is called in that case. Anyway, I update the error message which is consistent with the error "select count(*) from xxx group by bitmap" now.

spaces-X (approved):
LGTM

PR #11835, opened 2022-08-16
[enhancement](RowDescriptor) enhance tuple_idx check during runtime

ByteYue:
# Proposed changes

Issue Number: close #xxx

## Problem summary

The original get_tuple_idx function would check whether tuple_id is less than tuple map' size using CHECK macro. It might shut down the whole process once the check failed like the image below.

image

Meanwhile, there exists other occasions where the caller of get_tuple_idx never checks if it returns a valid tuple_idx or not.
This pr aims to enhance the robust during runtime by :
1. use DCHECK instead of CHECK to prevent surprisingly shutdown
2. add check logic when calling get_tuple_idx function 
## Checklist(Required)

    - [ ] No
3. Has unit tests been added:
    - [ ] No
    - [ ] No
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/runtime/descriptors.cpp
+++ b/be/src/runtime/descriptors.cpp
@@ -399,7 +399,8 @@ int RowDescriptor::get_row_size() const {
 }
 
 int RowDescriptor::get_tuple_idx(TupleId id) const {
-    DCHECK_LT(id, _tuple_idx_map.size()) << "RowDescriptor: " << debug_string();
+    // comment CHECK temporarily to make fuzzy test run smoothly
+    // DCHECK_LT(id, _tuple_idx_map.size()) << "RowDescriptor: " << debug_string();
     if (_tuple_idx_map.size() <= id) {
         return RowDescriptor::INVALID_IDX;
     }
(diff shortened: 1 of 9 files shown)


be/src/runtime/descriptors.h
```diff
@@ -382,6 +382,13 @@ class DescriptorTbl {
     DescriptorTbl() = default;
 };
 
+#define RETURN_IF_INVALID_IDX(tuple_id, tuple_idx)                                               \
```
dataroaring, line 385: RETURN_IF_INVALID_TUPLE_IDX

be/src/runtime/descriptors.cpp
```diff
@@ -399,7 +399,10 @@ int RowDescriptor::get_row_size() const {
 }
 
 int RowDescriptor::get_tuple_idx(TupleId id) const {
-    CHECK_LT(id, _tuple_idx_map.size()) << "RowDescriptor: " << debug_string();
+    DCHECK_LT(id, _tuple_idx_map.size()) << "RowDescriptor: " << debug_string();
```
dataroaring, line 402: DCHECK also would be core dump in DEBUG MODE, we should comment DCHECK to make sqlsmith run further.

dataroaring (changes requested):

dataroaring (approved):
LGTM

PR #11836, opened 2022-08-16
[Enhancement](meta) add err type/code/msg in audit event

chovy-3012:
# Proposed changes

Print "err msg" "err type"  and "err code" when sql execute failed .

Add "err msg" "err code" and "err type" in audit event .

Help developers diagnose the causes of failures.

For example:
```shell
errCode = 2, detailMessage = failed to initialize storage reader. tablet=76927688.1343432537.bd4a545605c7d580-8c277220384848b0, res=-230, backend=xxxxx
```

## Problem summary

Get more error information to help diagnose when sql execute failed .

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/ConnectProcessor.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/ConnectProcessor.java
@@ -166,10 +166,17 @@ public class ConnectProcessor {
 
         if (ctx.getState().isQuery()) {
             MetricRepo.COUNTER_QUERY_ALL.increase(1L);
-            if (ctx.getState().getStateType() == MysqlStateType.ERR
-                    && ctx.getState().getErrType() != QueryState.ErrType.ANALYSIS_ERR) {
-                // err query
-                MetricRepo.COUNTER_QUERY_ERR.increase(1L);
+            if (ctx.getState().getStateType() == MysqlStateType.ERR) {
+                // audit error msg
+                ctx.getAuditEventBuilder().setErrType(ctx.getState().getErrType().name());
+                ctx.getAuditEventBuilder().setErrCode(ctx.getState().getErrorCode().name());
+                ctx.getAuditEventBuilder().setErrMsg(ctx.getState().getErrorMessage());
+
+                // not analysis error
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/org/apache/doris/qe/ConnectProcessor.java
```diff
@@ -164,6 +164,13 @@ private void auditAfterExec(String origStmt, StatementBase parsedStmt, Data.PQue
                 .setQueryId(ctx.queryId() == null ? "NaN" : DebugUtil.printId(ctx.queryId()))
                 .setTraceId(spanContext.isValid() ? spanContext.getTraceId() : "");
 
+        // audit error msg
+        if (ctx.getState().getStateType() == MysqlStateType.ERR) {
+            ctx.getAuditEventBuilder().setErrType(ctx.getState().getErrType().name())
+                .setErrCode(ctx.getState().getErrorCode().name())
```
cambyzju, line 170: In the following two cases, ErrType, ErrCode and ErrMsg will be empty or other values? 
1. if we got MysqlStateType.ERR, but getState().getErrorCode() not set;
2. the error code returned by getState().getErrorCode() not registered;

chovy-3012: Yes, this case may happed and cause null point exception in `ctx.getState().getErrorXXX().name()`.
Do you suggest adding a empty/null value judgment ?

cambyzju: Yes I think it is necessary.

chovy-3012: Okeydokey

morningman (approved):
LGTM

PR #11862, opened 2022-08-17
[enhancement](VSlotRef) enhance column_id check in execute function during runtime

ByteYue:
# Proposed changes

Issue Number: close #xxx

## Problem summary

The column id check in VSlotRef::execute function before is too strict for fuzzy test to continuously produce random query. Temporarily loosen the check logic.
Moreover, there exists some careless call to VExpr::get_const_col, it might return a nullptr but not every function call checks if it's valid. It's an underlying problem.

## Checklist(Required)

    - [ ] No
2. Has unit tests been added:
    - [ ] No
    - [ ] No
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/vec/exprs/varray_literal.cpp
+++ b/be/src/vec/exprs/varray_literal.cpp
@@ -28,10 +28,7 @@ Status VArrayLiteral::prepare(RuntimeState* state, const RowDescriptor& row_desc
     Field array = is_null ? Field() : Array();
     for (const auto child : _children) {
         Field item;
-        auto col_ptr = child->get_const_col(context);
-        if (nullptr != col_ptr) {
-            col_ptr->column_ptr->get(0, item);
-        }
+        child->get_const_col(context)->column_ptr->get(0, item);
         array.get<Array>().push_back(item);
     }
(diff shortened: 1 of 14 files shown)


be/src/vec/exprs/varray_literal.cpp
```diff
@@ -28,7 +28,10 @@ Status VArrayLiteral::prepare(RuntimeState* state, const RowDescriptor& row_desc
     Field array = is_null ? Field() : Array();
     for (const auto child : _children) {
         Field item;
-        child->get_const_col(context)->column_ptr->get(0, item);
+        auto col_ptr = child->get_const_col(context);
+        if (nullptr != col_ptr) {
+            col_ptr->column_ptr->get(0, item);
+        }
         array.get<Array>().push_back(item);
```
dataroaring, line 35: It seems that this change should be included in this pr?

dataroaring (changes requested):

dataroaring (dismissed):
LGTM

yiguolei (approved):
LGTM

PR #11869, opened 2022-08-17
[bugfix](light weight schema change) support delete condition in schema change

yiguolei:
# Proposed changes

This is part of https://github.com/apache/doris/pull/10136

This PR also decouple delete hander from reader.

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments


be/src/olap/merger.cpp
```diff
@@ -108,14 +114,20 @@ Status Merger::vmerge_rowsets(TabletSharedPtr tablet, ReaderType reader_type,
...
     reader_params.version = dst_rowset_writer->version();
-    reader_params.tablet_schema = cur_tablet_schema;
     {
         std::shared_lock rdlock(tablet->get_header_lock());
         std::copy(tablet->delete_predicates().cbegin(), tablet->delete_predicates().cend(),
                   std::inserter(reader_params.delete_predicates,
                                 reader_params.delete_predicates.begin()));
     }
-
+    TabletSchemaSPtr merge_tablet_schema = std::make_shared<TabletSchema>();
```
HappenLee, line 123: Don't repeat yourself. same as the line 50, maybe we need a function

be/src/olap/reader.cpp
```diff
@@ -498,15 +493,22 @@ ColumnPredicate* TabletReader::_parse_to_predicate(const FunctionFilter& functio
...
+// to parse the conditions.
+ColumnPredicate* TabletReader::_parse_to_predicate(TabletSchemaSPtr tablet_schema,
+                                                   const TCondition& condition,
                                                    bool opposite) const {
-    // TODO: not equal and not in predicate is not pushed down
-    int32_t index = _tablet_schema->field_index(condition.column_name);
-    if (index < 0) {
-        return nullptr;
+    int32_t col_unique_id = -1;
+    // Delete handler will set column unique id to distinguish different column with same name
```
HappenLee, line 503: need more comment to know which case the if should be `true`

be/src/olap/reader.h
```diff
@@ -201,10 +202,10 @@ class TabletReader {
     std::vector<bool> _is_lower_keys_included;
     std::vector<bool> _is_upper_keys_included;
     // contains condition on key columns in agg or unique table or all column in dup tables
-    Conditions _conditions;
+    std::unique_ptr<Conditions> _conditions;
```
HappenLee, line 205: why change unique_ptr， need more `malloc` operator

be/src/olap/rowset/rowset_meta.h
```diff
@@ -233,8 +233,6 @@ class RowsetMeta {
 
     bool delete_flag() const { return _rowset_meta_pb.delete_flag(); }
```
HappenLee, line 234: this code useless also ?

yiguolei: I think, it is useless. Because it is never be set, but collector depend on it and many logic depend on it, I do not know the meaning of the flag.

be/src/olap/reader.h
yiguolei: because I change tablet schema as a parameter in Conditions constructor and remove set tablet schema method because the condtion should always have a tablet schema and should not be changed. But tablet schema is not required in reader's constructor.

Lchangliang:
LGTM

Lchangliang:
LGTM

be/src/olap/olap_cond.cpp
```diff
@@ -537,7 +537,7 @@ bool CondColumn::eval(const segment_v2::BloomFilter* bf) const {
 
 Status Conditions::append_condition(const TCondition& tcond) {
     DCHECK(_schema != nullptr);
-    int32_t index = _schema->field_index(tcond.column_name);
+    int32_t index = _schema->field_index(tcond.column_unique_id);
```
dataroaring, line 540: Should we hide index in the TabletSchema? We can use a function like this _schema->get_clolumn_by_unique_id(tcond.column_unique_id).

dataroaring (approved):

dataroaring (approved):
LGTM

PR #11930, opened 2022-08-19
[enhancement](memory) Add soft memory limit check to reduce the occurrence of OOM 

xinyiZzz:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Soft memory check means that the memory usage of the query has exceeded 90% of the mem limit. It is expected to wait for the memory to be released to continue. Otherwise, the exceeded mem limit may occur, and the query will be canceled.

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/runtime/memory/mem_tracker_limiter.h
+++ b/be/src/runtime/memory/mem_tracker_limiter.h
@@ -87,11 +87,16 @@ public:
         _limit = limit;
     }
     bool limit_exceeded() const { return _limit >= 0 && _limit < consumption(); }
+    // means that the memory usage of the query has exceeded 90% of the mem limit.
+    // It is expected to wait for the memory to be released to continue. Otherwise,
+    // the exceeded mem limit may occur, and the query will be canceled.
+    bool soft_limit_exceeded() const { return _limit >= 0 && _limit * 0.9 < consumption(); }
 
     // Returns true if a valid limit of this tracker limiter or one of its ancestors is exceeded.
-    bool any_limit_exceeded() const {
+    template <bool SOFT>
+    bool any_limit_exceeded() {
         for (const auto& tracker : _limited_ancestors) {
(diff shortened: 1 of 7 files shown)


be/src/runtime/memory/mem_tracker_limiter.h
```diff
@@ -87,11 +87,16 @@ class MemTrackerLimiter final : public MemTracker {
...
+    // the exceeded mem limit may occur, and the query will be canceled.
+    bool soft_limit_exceeded() const { return _limit >= 0 && _limit * 0.9 < consumption(); }
 
     // Returns true if a valid limit of this tracker limiter or one of its ancestors is exceeded.
-    bool any_limit_exceeded() const {
+    template <bool SOFT>
+    bool any_limit_exceeded() {
         for (const auto& tracker : _limited_ancestors) {
-            if (tracker->limit_exceeded()) {
+            if (SOFT ? tracker->soft_limit_exceeded() : tracker->limit_exceeded()) {
```
cambyzju, line 99:
```suggestion
            if constexpr (SOFT) {
                return tracker->soft_limit_exceeded();
            } else {
                return tracker->limit_exceeded();
            }
```

cambyzju:
Which type of nodes we should prefer soft limit? Could you give some description.

be/src/runtime/memory/mem_tracker_limiter.h
xinyiZzz: good code~

xinyiZzz:
`sleep_for` may keep waiting when exceeding mem limit, I will remove the logic here

PR #11939, opened 2022-08-20
[fix](memtracker) Remove useless memory exceed check

xinyiZzz:
# Proposed changes

Issue Number: close #xxx

## Problem summary

When the query mem tracker exceed limit, the query will be quickly canceled, and `sleep_for` and `_writer_cond.wait_for` are meaningless.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [x] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/exec/tablet_sink.cpp
+++ b/be/src/exec/tablet_sink.cpp
@@ -302,8 +302,7 @@ Status NodeChannel::add_row(Tuple* input_tuple, int64_t tablet_id) {
     // _cancelled may be set by rpc callback, and it's possible that _cancelled might be set in any of the steps below.
     // It's fine to do a fake add_row() and return OK, because we will check _cancelled in next add_row() or mark_close().
     while (!_cancelled && _pending_batches_num > 0 &&
-           (_pending_batches_bytes > _max_pending_batches_bytes ||
-            _parent->_mem_tracker->limit_exceeded(_max_pending_batches_bytes))) {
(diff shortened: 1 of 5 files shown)


yiguolei (approved):
LGTM

be/src/exec/tablet_sink.cpp
```diff
@@ -302,8 +302,7 @@ Status NodeChannel::add_row(Tuple* input_tuple, int64_t tablet_id) {
     // _cancelled may be set by rpc callback, and it's possible that _cancelled might be set in any of the steps below.
     // It's fine to do a fake add_row() and return OK, because we will check _cancelled in next add_row() or mark_close().
     while (!_cancelled && _pending_batches_num > 0 &&
-           (_pending_batches_bytes > _max_pending_batches_bytes ||
-            _parent->_mem_tracker->limit_exceeded(_max_pending_batches_bytes))) {
```
freemandealer, line 306: this is not only useless but a performance bug which will cause unnecessary sleep.

kpfly: > this is not only useless but a performance bug which will cause unnecessary sleep.

Is there any performance tests?

PR #12081, opened 2022-08-25
[Fix](BE) update pid file after be launch succeed

GoGoWen:
# Proposed changes

This pr going to write pid to pid file after be is started.   Before the pid will write to pid file even be is start failed,  which will cause pid to be overwrite by run "sh start_be.sh --daemon" when the be is already started.

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/service/doris_main.cpp
+++ b/be/src/service/doris_main.cpp
@@ -279,29 +279,6 @@ int main(int argc, char** argv) {
     using doris::Status;
     using std::string;
 
-    // open pid file, obtain file lock and save pid
-    string pid_file = string(getenv("PID_DIR")) + "/be.pid";
-    int fd = open(pid_file.c_str(), O_RDWR | O_CREAT,
-                  S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH);
-    if (fd < 0) {
-        fprintf(stderr, "fail to create pid file.");
-        exit(-1);
-    }
-
-    string pid = std::to_string((long)getpid());
-    pid += "\n";
-    size_t length = write(fd, pid.c_str(), pid.size());
-    if (length != pid.size()) {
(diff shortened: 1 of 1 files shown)


be/src/service/doris_main.cpp
```diff
@@ -488,6 +465,28 @@ int main(int argc, char** argv) {
     }
 #endif
 
+    // open pid file, obtain file lock and save pid
```
cambyzju, line 468: as we need to move the code location, is it better wrap these codes into a function?

GoGoWen:
I think it is no need to wrap write/read pidfile into a function,  Because
seems no elsewhere in our code to use it.

camby ***@***.***> 于2022年8月25日周四 22:13写道：

jackwener (dismissed):
please rebase to pass CI

GoGoWen:
> 

done

hello-stephen:
when BE is already started, then run "sh start_be.sh --daemon", be.pid file will be overwritten, Do you mean this?
At present, start_be.sh will check if PID process already exists and throw error info if it does.
In some situations, the PID in be.pid file is not the actual PID of doris_be, and if run start_be.sh, the be.pid file will indeed be overwritten， for example， if run this `for i in {1..10};do  bash start_be.sh --daemon & done`, the pid in be.pid is probably wrong，and if run start_be.sh again, the be.pid file will indeed be overwritten. 
So, it looks like this pr didn't really fix the problem.
any other idea?

PR #12082, opened 2022-08-25
[fix](array-type) array can not be distributed key and aggregation key

cambyzju:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Array column should not be `distributed key` or `group by` or `aggregate key`, we should forbid it before create table.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [x] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/HashDistributionDesc.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/HashDistributionDesc.java
@@ -80,6 +80,9 @@ public class HashDistributionDesc extends DistributionDesc {
                     if (columnDef.getType().isScalarType(PrimitiveType.STRING)) {
                         throw new AnalysisException("String Type should not be used in distribution column["
                                 + columnDef.getName() + "].");
+                    } else if (columnDef.getType().isArrayType()) {
+                        throw new AnalysisException("Array Type should not be used in distribution column["
+                                + columnDef.getName() + "].");
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


fe/fe-core/src/main/java/org/apache/doris/analysis/HashDistributionDesc.java
```diff
@@ -80,6 +80,9 @@ public void analyze(Set<String> colSet, List<ColumnDef> columnDefs) throws Analy
                     if (columnDef.getType().isScalarType(PrimitiveType.STRING)) {
                         throw new AnalysisException("String Type should not be used in distribution column["
                                 + columnDef.getName() + "].");
+                    } else if (columnDef.getType().isArrayType()) {
```
morningman, line 83: Maybe we should move all these checks to `toDistributionInfo()` ?

cambyzju: done

xy720 (approved):
LGTM

xy720 (commented):
We should also update the array docs to tell user don't use array type as distribution key.

cambyzju:
> We should also update the array docs to tell user don't use array type as distribution key.
Already exists:
`An array of T-type items, it cannot be used as a key column.`

PR #12117, opened 2022-08-26
[fix](storage) storage layer should aggregate on replace column

dataroaring:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/exec/olap_scanner.cpp
+++ b/be/src/exec/olap_scanner.cpp
@@ -138,7 +138,7 @@ Status OlapScanner::_init_tablet_reader_params(
         const std::vector<std::pair<string, std::shared_ptr<IBloomFilterFuncBase>>>&
                 bloom_filters) {
     // if the table with rowset [0-x] or [0-1] [2-y], and [0-1] is empty
-    bool single_version =
+    bool single_non_overlapping_version =
             (_tablet_reader_params.rs_readers.size() == 1 &&
              _tablet_reader_params.rs_readers[0]->rowset()->start_version() == 0 &&
              !_tablet_reader_params.rs_readers[0]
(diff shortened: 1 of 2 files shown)


dataalive (approved):
LGTM

be/src/exec/olap_scanner.cpp
```diff
@@ -152,14 +152,15 @@ Status OlapScanner::_init_tablet_reader_params(
                       ->rowset()
                       ->rowset_meta()
                       ->is_segments_overlapping());
+    
+    bool has_replace_col = false;
+    RETURN_IF_ERROR(_init_return_columns(!single_non_overlapping_version, has_replace_col));
 
-    _tablet_reader_params.direct_mode = single_version || _aggregation;
```
zhannngchen, line 156: The issue here is:
1. Planner thinks scanner can scan directly, don't need merge and can do some pre-aggregation works.
2. But scanner found it's a unqiue key table with seq column which need to merge first.

I think we'd fix it in planner? Planner should not set preAggregation as true on a unique table with seq column?

zhannngchen, line 156: Is the following change also can fix the issue?
```
bool unique_key_with_seq =  _tablet->keys_type() == UNIQUE_KEYS && _tablet->tablet_schema().has_sequence_col();
_tablet_reader_params.direct_mode = single_version || (_aggregation && ! unique_key_with_seq) ;
```

dataroaring: I am not sure what we need to in planner.  If fe sets aggregation, we ignore it here, then we can ignore aggreagate in planner somtimes?  Could you throw an example sql?

dataroaring: I  suspect that columns aggregated by replace without SEQ does not work either.

zhannngchen: > I suspect that columns aggregated by replace without SEQ does not work either.

root cause is _aggregation should not be true in such situation?

PR #12131, opened 2022-08-28
update data-model document problem

pangzhili:
When creating the table of the Duplicate model, specify the SortKet of error_code as Yes, but when creating the table later, the DUPLICATE KEY does not contain the error_code field, only the timestamp and type fields exist. Please check whether to add this field.


dataalive:
change the 'SortKey' property of error_code in sample to no will be more easy ?

dataalive (approved):
LGTM

jackwener (approved):
LGTM

jackwener:
Thanks for your contribution.

PR #12132, opened 2022-08-28
[improve](Nereids): unify all plan toString().

jackwener:
# Proposed changes

Issue Number: close #xxx

## Problem summary

unify all plan toString().

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [x] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/nereids/util/Utils.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/nereids/util/Utils.java
@@ -20,6 +20,7 @@ package org.apache.doris.nereids.util;
 import org.apache.doris.nereids.trees.expressions.SlotReference;
 import org.apache.doris.nereids.trees.plans.Plan;
 
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import org.apache.commons.lang3.StringUtils;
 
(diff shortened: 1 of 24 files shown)


fe/fe-core/src/main/java/org/apache/doris/nereids/util/Utils.java
```diff
@@ -119,4 +120,30 @@ public static List<SlotReference> getOutputSlotReference(Plan plan) {
...
     }
+
+    /**
+     * Get sql string for plan.
+     *
+     * @param planName name of plan, like LogicalJoin.
+     * @param variables variable needed to add into sqlString.
+     * @return the string of PlanNode.
+     */
+    public static String toSqlString(String planName, String... variables) {
```
924060929, line 131: `Utils.toSqlString('someField', someField.toString()` maybe cause NPE because someField is null.
So I suggest judge variables whether is null at here and convert to 'null' string.
Let's relax to invoke `Util.toSqlString('someField', someField)` and don't care about whether the someField is null.

```suggestion
    public static String toSqlString(String planName, Object... variables) {
```

jackwener: make sense.

morrySnow (approved):

PR #12160, opened 2022-08-29
[feature](light-schema-change) fix light-schema-change and add more cases

Lchangliang:
# Proposed changes

Issue Number: close #xxx

## Problem summary

1. Fix _delete_sign_idx and _seq_col_idx when append_column or build_schema when load.
2. Tablet schema cache support recycle when schema sptr use count equals 1.
3. Add a http interface for flink-connector to sync ddl.
4. Improve tablet->tablet_schema() by max_version_schema.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments


be/src/vec/olap/block_reader.cpp
```diff
@@ -283,8 +283,7 @@ Status BlockReader::_unique_key_next_block(Block* block, MemPool* mem_pool, Obje
 
     // do filter detete row in base compaction, only base compaction need to do the job
     if (_filter_delete) {
-        int delete_sign_idx =
-                (_sequence_col_idx == -1) ? target_columns.size() - 1 : target_columns.size() - 2;
```
yiguolei, line 287: should also modify non-vectorized engine's code.

Lchangliang: Non-vectorized engine don't need to modify. They have different implementations that non-vectorized engine don't use delete_sign_idx.

be/src/common/config.h
```diff
@@ -852,9 +852,11 @@ CONF_Int32(doris_remote_scanner_thread_pool_queue_size, "10240");
 // This config should be removed when the new scan node is ready.
 CONF_Bool(enable_new_scan_node, "false");
 
+CONF_Int32(tablet_schema_cache_recycle_interval, "86400") // s, one day
+
 #ifdef BE_TEST
-// test s3
```
Gabriel39, line 856: Why change here?

Lchangliang: The CI requests change it.

be/src/olap/tablet_schema.cpp
```diff
@@ -465,6 +466,11 @@ void TabletSchema::append_column(TabletColumn column, bool is_dropped_column) {
     if (column.is_nullable()) {
         _num_null_columns++;
     }
+    if (UNLIKELY(column.name() == DELETE_SIGN)) {
+        _delete_sign_idx = _num_columns;
+    } else if (UNLIKELY(column.name() == SEQUENCE_COL)) {
+        _sequence_col_idx = _num_columns;
+    }
```
dataroaring, line 473: I am not sure delte sign or seq col can be enabled by alter statement.  If  a table is created without 2 columns and then they are enabled by a alter statement. delete sign index and schema does not match for read?

be/src/olap/tablet_schema_cache.h
```diff
@@ -52,6 +55,25 @@ class TabletSchemaCache {
...
     }
 
+private:
+    /**
+     * @brief recycle when TabletSchemaSPtr use_count equals 1.
+     */
+    void _recycle() {
+        for (;;) {
+            std::this_thread::sleep_for(
+                    std::chrono::seconds(config::tablet_schema_cache_recycle_interval));
```
dataroaring, line 65: Could we remove this config? There are too many configs in doris? Do we really need it as a config?

be/src/olap/tablet_schema_cache.h
```diff
@@ -52,6 +55,25 @@ class TabletSchemaCache {
...
+            std::lock_guard guard(_mtx);
+            for (auto iter = _cache.begin(), last = _cache.end(); iter != last;) {
+                if (iter->second.use_count() == 1) {
+                    iter = _cache.erase(iter);
+                } else {
+                    ++iter;
+                }
+            }
```
dataroaring, line 73: Recycler does not consider memory consumption and timing. I worry that would lead to performance drop. May be we should have a performance test with a smaller enough recycle_interval.

dataroaring (changes requested):

be/src/olap/tablet_schema_cache.h
Lchangliang: That won't have performance problem，because it just recycle schemas only catched by tablet_schema_cache.

be/src/olap/reader.cpp
```diff
@@ -475,7 +475,7 @@ ColumnPredicate* TabletReader::_parse_to_predicate(
 }
 
 ColumnPredicate* TabletReader::_parse_to_predicate(const FunctionFilter& function_filter) {
-    int32_t index = _tablet->field_index(function_filter._col_name);
+    int32_t index = _tablet_schema->field_index(function_filter._col_name);
```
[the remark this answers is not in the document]
Lchangliang: Yes, tablet->fieled_index should not be need any more. Because we can use _tablet_schema->field_index directly. I remove the method in this pr too in case someone makes a mistake. That can cause some problems when table is enable light schema change.

be/src/olap/tablet.cpp
```diff
@@ -99,6 +99,15 @@ Tablet::Tablet(TabletMetaSharedPtr tablet_meta, DataDir* data_dir,
     // construct _timestamped_versioned_tracker from rs and stale rs meta
     _timestamped_version_tracker.construct_versioned_tracker(_tablet_meta->all_rs_metas(),
                                                              _tablet_meta->all_stale_rs_metas());
+    // if !_tablet_meta->all_rs_metas()[0]->tablet_schema(),
+    // that mean the tablet_meta is still no upgrade to 1.2.
```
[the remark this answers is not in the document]
Lchangliang: 1.2 mean doris 1.2 version. And i already add more comment information.

be/src/olap/tablet_schema.cpp
Lchangliang: I can't find the alter statement in doris doc. But it doesn't match after light schema change, because the hidden columns are always in the end. So when read and load, BE need to set their idx.

dataroaring (dismissed):
LGTM

dataroaring (approved):
LGTM

PR #12188, opened 2022-08-30
[improvement](spark-load)Spark load supports version 3.x, default supports 3.2.2

hf200012:
# Proposed changes
Spark load supports version 3.x, default supports 3.2.2

Spark load currently supports version 2.x, spark 3.x has been released to 3.3.0, we need to abandon support for version 2.x, because version 3.x in the ExpressionEncoder toRow method is no longer supported, so the two versions can not be compatible, can only choose one of them

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [x] Yes
    - [ ] No
    - [x] No

## Further comments


dujl:
update spark version directly will result in some user can not use spark-load because they only have spark2 runtime env.
how about create a new module spark3-dpp to support spark3 env?

fe/spark-dpp/src/main/java/org/apache/doris/load/loadv2/dpp/SparkDpp.java
```diff
@@ -266,7 +266,7 @@ private void writeRepartitionAndSortedRDDToParquet(JavaPairRDD<List<Object>, Obj
                             }
                             lastBucketKey = curBucketKey;
                         }
-                        InternalRow internalRow = encoder.toRow(rowWithoutBucketKey);
+                        InternalRow internalRow = encoder.createDeserializer().apply(rowWithoutBucketKey);
```
spaces-X:
```suggestion
                        InternalRow internalRow = encoder.createSerializer().apply(rowWithoutBucketKey);
```

hf200012: Just wrote a mistake has been corrected

morningman:
I think its time move the Spark Load out of the Doris' core.
The compatibility issue of Spark and other Hadoop related components is terrible and it is not suitable to be in Doris' core code any more.

spaces-X:
Some ideas to supoort the multi-version spark engine in spark load:
1. The spark load module is relatively independent and has the characteristics of low update frequency.  So maybe we can just update and maintain the fat jar package.
2. For each future version, e.g. 1.x.x, we only maintain and update the latest jar package. 
3. For the elder version(such as spark 2.x), we just package it into a jar and upload it to the official website. 
4. Based on 1 and 2, the spark-dpp build phase can skipped by default when building fe. Instead, the corresponding spark-dpp jar package can be download by enviroment variable url in `build.sh` and users can choose not to upgrade the spark-dpp jar package when upgrading fe.

dujl:
> Some ideas to supoort the multi-version spark engine in spark load:
> 
> 1. The spark load module is relatively independent and has the characteristics of low update frequency.  So maybe we can just update and maintain the fat jar package.
> 2. For each future version, e.g. 1.x.x, we only maintain and update the latest jar package.
> 3. For the elder version(such as spark 2.x), we just package it into a jar and upload it to the official website.
> 4. Based on 1 and 2, the spark-dpp build phase can skipped by default when building fe. Instead, the corresponding spark-dpp jar package can be download by enviroment variable url in `build.sh` and users can choose not to upgrade the spark-dpp jar package when upgrading fe.

The spark runtime environment is very complex. Different user maybe has different version. So It would be great if we could provide flexible version dependencie.
For case 1&2,  There may be multiple spark versions in for a foreseeable period of time，Such as current mainstream version of spark includes spark2.x and spark3.x.  So it is very great to provide spark 2.x and spark3.x dependencie with different compile profile. Like hudi/iceberg.
For case 3, some user maybe need compile the spark-dpp with their own version which major version same as our the elder version(such as spark 2.x).  It is very inconvenient if we can only provide binary packages on website.
For case 4, agree with u to move spark-dpp out of the doris' core.  But some compile env can not down package from the extranet

My idea is:

PR #12214, opened 2022-08-31
[Improvement](compaction) copy row in batch in VCollectIterator&VGenericIterator

yixiutt:
In VCollectIterator&VGenericIterator, use insert_range_from to copy rows
in a block which is continuous to save cpu cost.

If rows in rowset and segment are non overlapping, this whill improve 30%
throughput of compaction.If rows are completely overlapping such as load two
same files, the throughput goes nearly same as before.

# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/vec/olap/vcollect_iterator.cpp
+++ b/be/src/vec/olap/vcollect_iterator.cpp
@@ -444,20 +444,29 @@ Status VCollectIterator::Level1Iterator::_merge_next(Block* block) {
     auto target_columns = block->mutate_columns();
     size_t column_count = block->columns();
     IteratorRowRef cur_row = _ref;
+    IteratorRowRef pre_row_ref = _ref;
+
     if (UNLIKELY(_reader->_reader_context.record_rowids)) {
         _block_row_locations.resize(_batch_size);
     }
+    int continuous_row_in_block = 0;
     do {
-        const auto& src_block = cur_row.block;
-        assert(src_block->columns() == column_count);
-        for (size_t i = 0; i < column_count; ++i) {
-            target_columns[i]->insert_from(*(src_block->get_by_position(i).column),
-                                           cur_row.row_pos);
-        }
         if (UNLIKELY(_reader->_reader_context.record_rowids)) {
             _block_row_locations[target_block_row] = _cur_child->current_row_location();
         }
         ++target_block_row;
+        ++continuous_row_in_block;
+        // cur block finished, copy before merge_next cause merge_next will
+        // clear block column data
+        if (pre_row_ref.row_pos + continuous_row_in_block == pre_row_ref.block->rows()) {
+            const auto& src_block = pre_row_ref.block;
+            for (size_t i = 0; i < column_count; ++i) {
+                target_columns[i]->insert_range_from(*(src_block->get_by_position(i).column),
+                                                     pre_row_ref.row_pos, continuous_row_in_block);
+            }
+            continuous_row_in_block = 0;
+            pre_row_ref.block = nullptr;
+        }
         auto res = _merge_next(&cur_row);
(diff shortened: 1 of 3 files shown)


kpfly:
@Gabriel39 @BiteTheDDDDt Could you guys review this please

be/src/vec/olap/vgeneric_iterators.cpp
```diff
@@ -382,21 +399,36 @@ Status VMergeIterator::next_batch(vectorized::Block* block) {
...
     }
     size_t row_idx = 0;
+    VMergeIteratorContext* pre_ctx = nullptr;
     while (block->rows() < block_row_max) {
         if (_merge_heap.empty()) break;
 
         auto ctx = _merge_heap.top();
         _merge_heap.pop();
 
         if (!ctx->need_skip()) {
```
Gabriel39, line 409: Is it possible to have different `need_skip` in a `VMergeIterator `? If all of `ctx` in each `VMergeIterator` is `need_skip` or not `need_skip`, maybe we should move this condition out of this while-loop

be/src/vec/olap/vgeneric_iterators.cpp
```diff
@@ -382,21 +399,36 @@ Status VMergeIterator::next_batch(vectorized::Block* block) {
         _block_row_locations.resize(block_row_max);
     }
     size_t row_idx = 0;
+    VMergeIteratorContext* pre_ctx = nullptr;
     while (block->rows() < block_row_max) {
         if (_merge_heap.empty()) break;
 
         auto ctx = _merge_heap.top();
```
Gabriel39, line 406: We should make this comparison inside one block as much as possible instead of always using a heap to compare rows between different blocks.

be/src/vec/olap/vcollect_iterator.cpp
```diff
@@ -470,7 +479,32 @@ Status VCollectIterator::Level1Iterator::_merge_next(Block* block) {
             LOG(WARNING) << "next failed: " << res;
             return res;
         }
-    } while (target_block_row < _batch_size);
+        if (target_block_row >= _batch_size) {
```
BiteTheDDDDt, line 482: Maybe we should limit a batch not over batch_size

be/src/vec/olap/vgeneric_iterators.cpp
```diff
@@ -382,21 +399,36 @@ Status VMergeIterator::next_batch(vectorized::Block* block) {
...
         _merge_heap.pop();
 
         if (!ctx->need_skip()) {
-            // copy current row to block
-            ctx->copy_row(block);
+            ctx->add_cur_batch();
+            if (pre_ctx != ctx) {
+                if (pre_ctx) {
```
BiteTheDDDDt, line 412: same problem

be/src/vec/olap/vgeneric_iterators.cpp
yixiutt: need_skip is set in Comparator and it's used for every row

be/src/vec/olap/vcollect_iterator.cpp
yixiutt: target_block_row must be less than batch_size

be/src/vec/olap/vgeneric_iterators.cpp
yixiutt: add a row_idx check below

BiteTheDDDDt (approved):
LGTM

PR #12224, opened 2022-08-31
[fix](new-scan) fix potential wrong cancel when sql has limit

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem summary

One scanner maybe stopped by other scanners earlier, which cause data lose.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -852,7 +852,7 @@ CONF_Int32(doris_remote_scanner_thread_pool_queue_size, "10240");
 
 // If set to true, the new scan node framework will be used.
 // This config should be removed when the new scan node is ready.
-CONF_Bool(enable_new_scan_node, "false");
+CONF_Bool(enable_new_scan_node, "true");
 
 #ifdef BE_TEST
 // test s3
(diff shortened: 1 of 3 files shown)


be/src/common/config.h
```diff
@@ -852,7 +852,7 @@ CONF_Int32(doris_remote_scanner_thread_pool_queue_size, "10240");
 
 // If set to true, the new scan node framework will be used.
 // This config should be removed when the new scan node is ready.
-CONF_Bool(enable_new_scan_node, "false");
+CONF_Bool(enable_new_scan_node, "true");
```
Jibing-Li, line 855: Do we need to test/implement old external table before set this to true?

Jibing-Li: Ignore the above comment, I think broker scan node is not affected.

Jibing-Li (approved):
LGTM

yiguolei (approved):
LGTM

PR #12239, opened 2022-08-31
 [enhancement](http) executeSQL rest api support streaming response

BePPPower:
# Proposed changes

Issue Number: close #xxx

## Problem summary

At now, the http interface `executeSQL` is executing in a synchronous manner, it will store all ResultSets read from jdbc.This may cause OOM exception with this method.The current solution is to limit ResultSets rows in fe to avoid `OOM`.

I use `JSONWriter` of gson and `PrintWriter` of HttpServletResponse to implement streaming response. We use jdbc to read ResultSets in streaming manner and response client in streaming manner too.

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/httpv2/rest/StmtExecutionAction.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/httpv2/rest/StmtExecutionAction.java
@@ -100,7 +100,11 @@ public class StmtExecutionAction extends RestBaseController {
 
         ConnectContext.get().changeDefaultCatalog(ns);
         ConnectContext.get().setDatabase(getFullDbName(dbName));
-        return executeQuery(authInfo, stmtRequestBody.is_sync, stmtRequestBody.limit, stmtRequestBody);
+
+        String streamHeader = request.getHeader("X-Doris-Stream");
+        boolean isStream = !("false".equalsIgnoreCase(streamHeader));
+        return executeQuery(authInfo, stmtRequestBody.is_sync, stmtRequestBody.limit, stmtRequestBody,
+                            response, isStream);
     }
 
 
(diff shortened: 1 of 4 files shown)


fe/fe-core/src/main/java/org/apache/doris/httpv2/rest/StmtExecutionAction.java
```diff
@@ -100,7 +100,11 @@ public Object executeSQL(@PathVariable(value = NS_KEY) String ns, @PathVariable(
 
         ConnectContext.get().changeDefaultCatalog(ns);
         ConnectContext.get().setDatabase(getFullDbName(dbName));
-        return executeQuery(authInfo, stmtRequestBody.is_sync, stmtRequestBody.limit, stmtRequestBody);
+
+        String streamHeader = request.getHeader("X-Doris-Stream");
+        boolean isStream = !("false".equalsIgnoreCase(streamHeader));
```
stalary, line 105: Why not use "true".equalsIgnoreCase(streamHeader), it can not use stream when streamHeader is null.

fe/fe-core/src/main/java/org/apache/doris/httpv2/util/streamresponse/StreamResponseInf.java
```diff
@@ -0,0 +1,56 @@
...
+package org.apache.doris.httpv2.util.streamresponse;
+
+import java.io.PrintWriter;
+import java.sql.ResultSet;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * StreamResponseInf use response.getWriter() to response client
+ */
+public abstract class StreamResponseInf {
```
stalary, line 27: Why does it end in Inf

fe/fe-core/src/main/java/org/apache/doris/httpv2/rest/StmtExecutionAction.java
BePPPower: Because I think stream should be default behavior, and can close it by setting `streamHeader: false`

fe/fe-core/src/main/java/org/apache/doris/httpv2/util/streamresponse/StreamResponseInf.java
BePPPower: This class is an interfance of streamResponse, we can extends this abstract class to implement different streamResponse to achieve different outputFormat.
For example, I implenment JSONStreamResonse to output `json` format. If we want to output `csv` format, we can implenment a class called `CSVStreamResponse`.
So, I define this class end in Inf.

yiguolei (approved):
LGTM

Issue #5664, opened 2021-04-16
[Coredump] data sink node close causes coredump


xinghuayu007:
**Describe the bug**

branch: Doris-0.13 
core content:

#0  free (this=<optimized out>, ptr=<optimized out>) at /builds/olap/doris/be/src/runtime/free_pool.hpp:94
#1  doris_udf::FunctionContext::free (this=this@entry=0x64df63328, buffer=<optimized out>) at /builds/olap/doris/be/src/udf/udf.cpp:308
#2  0x0000000000e06595 in doris::CaseExpr::close (this=0x19a01a6a80, state=0x1d5489f500, ctx=0x1beb4b41e0, scope=doris_udf::FunctionContext::FRAGMENT_LOCAL)
    at /builds/olap/doris/be/src/exprs/case_expr.cpp:74
#3  0x0000000000d76583 in doris::ExprContext::close (this=0x1beb4b41e0, state=state@entry=0x1d5489f500) at /builds/olap/doris/be/src/exprs/expr_context.cpp:89
#4  0x0000000000d6dfd0 in doris::Expr::close (ctxs=…, state=state@entry=0x1d5489f500) at /builds/olap/doris/be/src/exprs/expr.cpp:578
#5  0x00000000016c97bc in doris::DataStreamSender::close (this=0x1b2d9e5380, state=0x1d5489f500, exec_status=…) at /builds/olap/doris/be/src/runtime/data_stream_sender.cpp:673
#6  0x000000000109906e in doris::PlanFragmentExecutor::close (this=this@entry=0xe72828670) at /builds/olap/doris/be/src/runtime/plan_fragment_executor.cpp:562
#7  0x000000000101b7e1 in doris::FragmentExecState::execute (this=0xe72828600) at /builds/olap/doris/be/src/runtime/fragment_mgr.cpp:222
#8  0x000000000101df36 in doris::FragmentMgr::exec_actual(std::shared_ptr<doris::FragmentExecState>, std::function<void (doris::PlanFragmentExecutor)>) (this=0x521c600, exec_state=…, cb=…)
    at /builds/olap/doris/be/src/runtime/fragment_mgr.cpp:430
[...]

PR #12257, opened 2022-09-01
[fix](memory): avoid coredump when list pointer is null

levy5307:
# Proposed changes

Issue Number: [5664](https://github.com/apache/doris/issues/5664)

## Problem summary

To avoid the coredump memtioned in above issue, adding the validation of pointer before using it.
BTW, It's a best practise to validate a pointer before using it.

## Checklist(Required)

    - [ ] No
2. Has unit tests been added:
    - [ ] No Need
    - [ ] No Need
    - [ ] No
    - [ ] No

## Further comments


be/src/runtime/free_pool.hpp
```diff
@@ -128,9 +128,14 @@ class FreePool {
 #ifndef NDEBUG
         check_valid_allocation(list);
 #endif
-        // Add node to front of list.
-        node->next = list->next;
-        list->next = node;
+        if (nullptr == list) {
```
weizuo93, line 131: Why  is `list` nullptr?

levy5307: > Why is `list` nullptr?

Maybe it was released somewhere

weizuo93:
Normally, the pointer `list` here should not be `nullptr`. I think we'd better to explore the reason why the `list` is nullptr and fix it instead of skipping nullptr, otherwise it might hide the potential problem.

xinyiZzz:
`FreePool` will be removed in the future. simple fixes to try.

xinyiZzz:
@levy5307 p0 case failed

PR #12262, opened 2022-09-01
[fix](array-type) revert ARRAY<NOT_NULL(T)>

cambyzju:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

Now there are still lots of bugs related to ARRAY<NOT_NULL(T)>.
We decide not to support nested elements NOT NULLABLE for ARRAY type at the first version.
Nested elements of ARRAY will be always nullable.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [x] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments


xy720 (commented):
You should also update the return type of FunctionCallExpr
In some case, the array function's return type have `ContainsNull = false`.

fe/fe-core/src/main/cup/sql_parser.cup
```diff
@@ -4847,8 +4847,6 @@ type ::=
   {: RESULT = ScalarType.createVarcharType(-1); :}
   | KW_ARRAY LESSTHAN type:value_type GREATERTHAN
   {: RESULT = new ArrayType(value_type); :}
-  | KW_ARRAY LESSTHAN KW_NOT_NULL LPAREN type:value_type RPAREN GREATERTHAN
-  {: RESULT = new ArrayType(value_type, false); :}
```
adonis0147, line 4851: I think we can keep this syntax but just throw an exception due to we don't support it temporarily.

adonis0147 (changes requested):

fe/fe-core/src/main/cup/sql_parser.cup
cambyzju: Lots of codes who called `new ArrayType` will also need to catch the exception,  if we throw an exception while new ArrayType. I think just remove these too lines is better.

xy720 (approved):
LGTM

fe/fe-core/src/main/cup/sql_parser.cup
```diff
@@ -4847,8 +4847,6 @@ type ::=
   {: RESULT = ScalarType.createVarcharType(-1); :}
   | KW_ARRAY LESSTHAN type:value_type GREATERTHAN
   {: RESULT = new ArrayType(value_type); :}
```
adonis0147, line 4849: What about using the following modification?

```suggestion
  {:
      RESULT = new ArrayType(value_type);
      if (!((ArrayType) RESULT).getContainsNull()) {
          throw new AnalysisException("ARRAY<NOT_NULL()> is unsupported.")
      }
  :}
```

adonis0147 (changes requested):

adonis0147 (approved):
LGTM

PR #12293, opened 2022-09-02
[Enhancement](array-type) record offsets info to speed up array colum…

### LINKED ISSUE #12246: [Enhancement] array columns's ArrayFileColumnIterator::seek_to_ordinal maybe too slow
### Search before asking

### Description

Currently we record length of array in a seperate column writer called `_length_writer`, but I think recording length of array could lead to bad seek performance.Below is the performance issue i encountered:

I have a table which is

```
CREATE TABLE `nested` (   
`qid` bigint(20) NULL COMMENT "",   
`tag` ARRAY<text> NULL COMMENT "",   
`creationDate` datetime NULL COMMENT "",   
`title` text NULL COMMENT "",   
`user` text NULL COMMENT "",   
`answers.user` ARRAY<text> NULL COMMENT "",   
`answers.date` ARRAY<datetime> NULL COMMENT "",   
INDEX tag_idx (`tag`) USING INVERTED COMMENT '',   
INDEX creation_date_idx (`creationDate`) USING INVERTED COMMENT '',   
INDEX title_idx (`title`) USING INVERTED COMMENT '',   
INDEX user_idx (`user`) USING INVERTED(NONE) COMMENT '',   
INDEX answers_user_idx (`answers.user`) USING INVERTED COMMENT '',   
INDEX answers_date_idx (`answers.date`) USING INVERTED COMMENT '',   ... 
) 
ENGINE=OLAP 
DUPLICATE KEY(`qid`) COMMENT "OLAP" 
DISTRIBUTED BY HASH(`qid`) BUCKETS 18 
PROPERTIES ( 
"in_memory" = "false", 
"storage_format" = "V2", 
"compression" = "ZSTD" 
);
```

As the sql above, I created inverted index(which is similar to bitmap index) on some columns.

this is the data sample

```
{"title": "Display Progress Bar at the Time of Processing", "qid": "1000000", "answers": [{"date": "2009-06-16T09:55:57.320", "user": "Micha\u0142 Niklas (22595)"}, {"date": "2009-06-17T12:34:22.643", "user": "Jack Njiri (77153)"}], "tag": ["vb6", "progress-bar"], "user": "Jash", "creationDate": "2009-06-16T07:28:42.770"}
```

Then I run a benchmark with sql like `select * from {}.{} where  title match '{}' limit 10`(We have a special sql word `match` which will filter rows when  `title` contains this word by inverted index). Then I found that the sql could be slow due to ` ArrayFileColumnIterator::seek_to_ordinal`

image

The reason is that:
1. Filtered row_ids from index is very discrete and we need do lots of column seek_to_ordinal
2. The compute of  ArrayFileColumnIterator::seek_to_ordinal is very slow. We need to caculate ordinal of specified ordinal_id
by read from _length_iterator and needs lots of memory copy and then sum massive items read from _length_iterator to get the  specified ordinal_id.The length_iterator's page could contains thousands of length items(which I saw more than 50000), caculate them could lead to very bad performance

image

### Solution

We could change `_length_writer` to  `_offset_writer`, then seek_to_ordinal could be very fast(could directly get the ordinal of _item_iterator by read only 1 item from _offset_writer)

after I modified related code, the performace is  as below
image

**seek_to_ordinal cpu cost from 74% down to 14%**

### sequence seek performance  test
**query:** select filtered by qid which is primary key
`sql = "select * from nested where qid > 26619949 limit 1000"`

seek_by_length
image

seek_by_offsets
image

image

- [X] Yes I am willing
 to submit a PR!

### Code of Conduct

eldenmoon:
…n seek performance

# Proposed changes

Issue Number: close #12246

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [x] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -851,6 +851,8 @@ CONF_Bool(enable_simdjson_reader, "false");
 CONF_Int32(doris_remote_scanner_thread_pool_thread_num, "16");
 // number of s3 scanner thread pool queue size
 CONF_Int32(doris_remote_scanner_thread_pool_queue_size, "10240");
+// write offsets instead of length to speed up seek performance
+CONF_Bool(enable_write_array_offset, "false");
 
 // If set to true, the new scan node framework will be used.
 // This config should be removed when the new scan node is ready.
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -851,8 +851,6 @@ CONF_Bool(enable_simdjson_reader, "false");
 CONF_Int32(doris_remote_scanner_thread_pool_thread_num, "16");
 // number of s3 scanner thread pool queue size
 CONF_Int32(doris_remote_scanner_thread_pool_queue_size, "10240");
-// write offsets instead of length to speed up seek performance
-CONF_Bool(enable_write_array_offset, "false");
 
 // If set to true, the new scan node framework will be used.
 // This config should be removed when the new scan node is ready.
--- a/be/src/olap/column_vector.cpp
+++ b/be/src/olap/column_vector.cpp
@@ -228,9 +228,8 @@ void ArrayColumnVectorBatch::put_item_ordinal(segment_v2::ordinal_t* ordinals, s
                                               size_t size) {
     DCHECK(size > 0);
     size_t first_offset = *(_offsets->scalar_cell_ptr(start_idx));
-    segment_v2::ordinal_t first_ordinal = ordinals[0];
-    size_t i = 0;
-    while (++i < size) {
+    for (size_t i = 1; i < size; ++i) {
+        segment_v2::ordinal_t first_ordinal = ordinals[0];
         *(_offsets->scalar_cell_ptr(start_idx + i)) = first_offset + (ordinals[i] - first_ordinal);
     }
 }
--- a/be/src/olap/column_vector.cpp
+++ b/be/src/olap/column_vector.cpp
@@ -224,6 +224,17 @@ Status ArrayColumnVectorBatch::resize(size_t new_cap) {
     return Status::OK();
 }
 
+void ArrayColumnVectorBatch::put_item_ordinal(segment_v2::ordinal_t* ordinals, size_t start_idx,
+                                              size_t size) {
+    DCHECK(size > 0);
+    size_t first_offset = *(_offsets->scalar_cell_ptr(start_idx));
+    segment_v2::ordinal_t first_ordinal = ordinals[0];
+    size_t i = 0;
+    while (++i < size) {
+        *(_offsets->scalar_cell_ptr(start_idx + i)) = first_offset + (ordinals[i] - first_ordinal);
+    }
+}
+
 void ArrayColumnVectorBatch::get_offset_by_length(size_t start_idx, size_t size) {
     DCHECK(start_idx >= 0);
     DCHECK(start_idx + size < _offsets->capacity());
--- a/be/src/olap/column_vector.h
+++ b/be/src/olap/column_vector.h
@@ -233,6 +233,13 @@ public:
      */
     void get_offset_by_length(size_t start_idx, size_t size);
 
+    // From `start_idx`, put `size` ordinals to _item_offsets
+    // Ex:
+    // original _item_offsets: 0 3 5 9; ordinals to be added: 100 105 111; size: 3; satart_idx: 3
+    // --> _item_offsets: 0 3 5 (9 + 100 - 100) (9 + 105 - 100) (9 + 111 - 100)
+    // _item_offsets becomes 0 3 5 9 14 20
+    void put_item_ordinal(segment_v2::ordinal_t* ordinals, size_t start_idx, size_t size);
+
     size_t get_item_size(size_t start_idx, size_t size) {
         return *(_offsets->scalar_cell_ptr(start_idx + size)) -
                *(_offsets->scalar_cell_ptr(start_idx));
--- a/be/src/olap/column_vector.h
+++ b/be/src/olap/column_vector.h
@@ -236,7 +236,7 @@ public:
     // From `start_idx`, put `size` ordinals to _item_offsets
     // Ex:
     // original _item_offsets: 0 3 5 9; ordinals to be added: 100 105 111; size: 3; satart_idx: 3
-    // --> _item_offsets: 0 3 5 (9 + 100 - 100) (9 + 105 - 100) (9 + 111 - 100)
+    // --> _item_offsets: 0 3 5 9 (9 + 105 - 100) (9 + 111 - 100)
     // _item_offsets becomes 0 3 5 9 14 20
     void put_item_ordinal(segment_v2::ordinal_t* ordinals, size_t start_idx, size_t size);
 
--- a/be/src/olap/column_vector.h
+++ b/be/src/olap/column_vector.h
@@ -235,7 +235,7 @@ public:
 
     // From `start_idx`, put `size` ordinals to _item_offsets
     // Ex:
-    // original _item_offsets: 0 3 5 9; ordinals to be added: 100 105 111; size: 3; satart_idx: 3
+    // original _item_offsets: 0 3 5 9; ordinals to be added: 100 105 111; size: 3; start_idx: 3
     // --> _item_offsets: 0 3 5 9 (9 + 105 - 100) (9 + 111 - 100)
     // _item_offsets becomes 0 3 5 9 14 20
     void put_item_ordinal(segment_v2::ordinal_t* ordinals, size_t start_idx, size_t size);
--- a/be/src/olap/rowset/segment_v2/column_reader.cpp
+++ b/be/src/olap/rowset/segment_v2/column_reader.cpp
@@ -406,6 +406,21 @@ Status ArrayFileColumnIterator::init(const ColumnIteratorOptions& opts) {
     return Status::OK();
 }
 
+Status ArrayFileColumnIterator::_peek_one_offset(ordinal_t* offset) {
+    if (_offset_iterator->get_current_page()->has_remaining()) {
+        PageDecoder* offset_page_decoder = _offset_iterator->get_current_page()->data_decoder;
+        ColumnBlock ordinal_block(_length_batch.get(), nullptr);
+        ColumnBlockView ordinal_view(&ordinal_block);
+        size_t i = 1;
+        RETURN_IF_ERROR(offset_page_decoder->peek_next_batch(&i, &ordinal_view)); // not null
+        DCHECK(i == 1);
+        *offset = *reinterpret_cast<uint64_t*>(_length_batch->data());
+    } else {
+        *offset = _offset_iterator->get_current_page()->next_array_item_ordinal;
+    }
+    return Status::OK();
+}
+
 Status ArrayFileColumnIterator::next_batch(size_t* n, ColumnBlockView* dst, bool* has_null) {
     ColumnBlock* array_block = dst->column_block();
     auto* array_batch = static_cast<ArrayColumnVectorBatch*>(array_block->vector_batch());
@@ -422,17 +437,7 @@ Status ArrayFileColumnIterator::next_batch(size_t* n, ColumnBlockView* dst, bool
         return Status::OK();
     }
 
-    if (_offset_iterator->get_current_page()->has_remaining()) {
-        // peek read one more to get the last array's length
-        size_t i = 1;
-        bool offsets_has_null = false;
-        ordinal_t save = _offset_iterator->get_current_ordinal();
-        RETURN_IF_ERROR(_offset_iterator->next_batch(&i, &offset_view, &offsets_has_null));
-        RETURN_IF_ERROR(_offset_iterator->seek_to_ordinal(save));
-    } else {
-        *(reinterpret_cast<ordinal_t*>(offset_view.data())) =
-                _offset_iterator->get_current_page()->next_array_item_ordinal;
-    }
+    RETURN_IF_ERROR(_peek_one_offset(reinterpret_cast<ordinal_t*>(offset_view.data())));
 
     size_t start_offset = dst->current_offset();
     auto* ordinals = reinterpret_cast<ordinal_t*>(offset_block.data());
@@ -454,7 +459,6 @@ Status ArrayFileColumnIterator::next_batch(size_t* n, ColumnBlockView* dst, bool
     }
 
     // read item
-    // size_t item_size = array_batch->get_item_size(dst->current_offset(), *n + 1);
     size_t item_size = ordinals[*n] - ordinals[0];
     if (item_size >= 0) {
         bool item_has_null = false;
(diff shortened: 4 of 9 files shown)


eldenmoon:
using the original unit test and regression test cases

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -481,13 +481,38 @@ Status ArrayFileColumnIterator::next_batch(size_t* n, vectorized::MutableColumnP
...
+            offsets_data[i] = offsets_data[i - 1] + (offsets_data[i + 1] - offsets_data[i]);
+        }
+        column_offsets.pop_back(1);
+    } else {
+        // for compability
+        auto& offsets_data = column_offsets.get_data();
+        for (ssize_t i = start; i < offsets_data.size(); ++i) {
+            offsets_data[i] += offsets_data[i - 1]; // -1 is ok
+        }
+        num_items = offsets_data.back() - offsets_data[start - 1];
```
adonis0147, line 512: This modification makes the function more complex than before, I think you could encapsulate them into a new function.

be/src/olap/rowset/segment_v2/column_reader.h
```diff
@@ -394,31 +394,47 @@ class ArrayFileColumnIterator final : public ColumnIterator {
...
-                ColumnBlockView ordinal_view(&ordinal_block);
-                RETURN_IF_ERROR(_length_iterator->next_batch(&this_read, &ordinal_view, &has_null));
-                auto* ordinals = reinterpret_cast<uint64_t*>(_length_batch->data());
-                for (int i = 0; i < this_read; ++i) {
-                    item_ordinal += ordinals[i];
+            RETURN_IF_ERROR(_length_iterator->next_batch(&this_read, &ordinal_view, &has_null));
+            RETURN_IF_ERROR(_length_iterator->seek_to_ordinal(ord));
+
+            auto* ordinals = reinterpret_cast<uint64_t*>(_length_batch->data());
+            RETURN_IF_ERROR(_item_iterator->seek_to_ordinal(ordinals[0]));
```
adonis0147, line 409: ditto

adonis0147 (changes requested):

adonis0147:
What about providing some performance statistics to support the optimization?

be/src/olap/rowset/segment_v2/column_reader.cpp
eldenmoon: that's a good suggestion

eldenmoon:
@adonis0147 I did some performance statistics in this issue https://github.com/apache/doris/issues/12246

be/src/olap/rowset/segment_v2/column_reader.cpp
eldenmoon: that's a good suggestion

be/src/common/config.h
```diff
@@ -849,6 +849,8 @@ CONF_Bool(enable_simdjson_reader, "false");
 CONF_Int32(doris_remote_scanner_thread_pool_thread_num, "16");
 // number of s3 scanner thread pool queue size
 CONF_Int32(doris_remote_scanner_thread_pool_queue_size, "10240");
+// write offsets instead of length to speed up seek performance
+CONF_Bool(enable_write_array_offset, "false");
```
cambyzju, line 853: 1. After the data persist, we can not change this flag any more. If really needed, I think we should add some meta to distinguish it is offset or length inside segment file;
2. How about if we delete one row with ARRAY column?
3. Does clickhouse store ARRAY length or offset?

cambyzju (changes requested):

be/src/common/config.h
eldenmoon: 1. the code provided both backward and forward compability, use `_length_iterator->get_current_page()->next_array_item_ordinal > 0` to distinguish whether we go the offset path else length path
2. I think delete row is not effected whether using offset or length, it's re compacted and written a new column?
3. clickhous store array offset

eldenmoon: I can add a seperate meta to distinguish them

morningman: Maybe we should just turn if on by default?

cambyzju: I think we should better choose one before doris1.2 release.
If we decide to choose one, we need more test cases.
This pr focus on random read of Array column, `offset version` is better than `length version` in this case.
Other cases, such as sequence read, we also need to compare.

Please add my wechat (doris_cambyzju).

eldenmoon: I will provide sequence read case too @cambyzju

eldenmoon: I had a conversation with @cambyzju  and decided to use  `offset version`, so this PR is incompatible with previous

be/src/olap/column_vector.h
```diff
@@ -233,6 +233,13 @@ class ArrayColumnVectorBatch : public ColumnVectorBatch {
      */
     void get_offset_by_length(size_t start_idx, size_t size);
 
+    // From `start_idx`, put `size` ordinals to _item_offsets
+    // Ex:
+    // original _item_offsets: 0 3 5 9; ordinals to be added: 100 105 111; size: 3; satart_idx: 3
+    // --> _item_offsets: 0 3 5 (9 + 100 - 100) (9 + 105 - 100) (9 + 111 - 100)
+    // _item_offsets becomes 0 3 5 9 14 20
+    void put_item_ordinal(segment_v2::ordinal_t* ordinals, size_t start_idx, size_t size);
```
cambyzju, line 241: does push back is enough? it seems we always append ordinals at the end.

eldenmoon: yes, it's enough we just need to caculate the ordinals read from _offset_iterator, and then put them into the back of ArrayColumnVectorBatch's _offsets.

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -410,18 +410,33 @@ Status ArrayFileColumnIterator::next_batch(size_t* n, ColumnBlockView* dst, bool
     ColumnBlock* array_block = dst->column_block();
     auto* array_batch = static_cast<ArrayColumnVectorBatch*>(array_block->vector_batch());
 
-    // 1. read n offsets
+    // 1. read n+1 offsets
+    array_batch->offsets()->resize(*n + 1);
     ColumnBlock offset_block(array_batch->offsets(), nullptr);
-    ColumnBlockView offset_view(&offset_block,
-                                dst->current_offset() + 1); // offset应该比collection的游标多1
+    ColumnBlockView offset_view(&offset_block);
```
cambyzju, line 416: first element of offsets is always zero, should we still need to read?

be/src/olap/column_vector.cpp
```diff
@@ -224,6 +224,16 @@ Status ArrayColumnVectorBatch::resize(size_t new_cap) {
     return Status::OK();
 }
 
+void ArrayColumnVectorBatch::put_item_ordinal(segment_v2::ordinal_t* ordinals, size_t start_idx,
+                                              size_t size) {
+    size_t first_offset = *(_offsets->scalar_cell_ptr(start_idx));
+    segment_v2::ordinal_t first_ordinal = ordinals[0];
```
cambyzju, line 230: need DCHECK(size > 0) or put this line inside while loop, do not worry about performance, the compiler will cover it.

eldenmoon: OK

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -410,18 +410,33 @@ Status ArrayFileColumnIterator::next_batch(size_t* n, ColumnBlockView* dst, bool
...
         return Status::OK();
     }
-    array_batch->get_offset_by_length(dst->current_offset(), *n);
+
+    if (_offset_iterator->get_current_page()->has_remaining()) {
+        // peek read one more to get the last array's length
+        size_t i = 1;
+        bool offsets_has_null = false;
+        ordinal_t save = _offset_iterator->get_current_ordinal();
+        RETURN_IF_ERROR(_offset_iterator->next_batch(&i, &offset_view, &offsets_has_null));
```
cambyzju, line 430: should we use another function to replace next_batch, to avoid twice call of next_batch?

for example, we may has a function:
next_batch_for_offsets() {
    // after normal read, always append more offset at the end;
}

eldenmoon: agree, may be `peek_batch_for_offsets`

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -410,18 +410,33 @@ Status ArrayFileColumnIterator::next_batch(size_t* n, ColumnBlockView* dst, bool
     ColumnBlock* array_block = dst->column_block();
     auto* array_batch = static_cast<ArrayColumnVectorBatch*>(array_block->vector_batch());
 
-    // 1. read n offsets
+    // 1. read n+1 offsets
+    array_batch->offsets()->resize(*n + 1);
```
cambyzju, line 414:
```suggestion
    array_batch->offsets()->reserve(*n + 1);
```

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -466,6 +486,42 @@ Status ArrayFileColumnIterator::next_batch(size_t* n, ColumnBlockView* dst, bool
...
+    auto& column_offsets = static_cast<vectorized::ColumnArray::ColumnOffsets&>(*offsets);
+    ordinal_t last_offset = 0;
+    RETURN_IF_ERROR(_peek_one_offset(&last_offset));
+
+    auto& offsets_data = column_offsets.get_data();
+    *num_items = last_offset - offsets_data[start];
+    // caculate real offsets
+    ordinal_t first_offset = offsets_data[start - 1]; // -1 is valid
+    ordinal_t first_ord = offsets_data[start];
+    for (ssize_t i = start; i < offsets_data.size() - 1; ++i) {
```
cambyzju, line 517:
```suggestion
    for (ssize_t i = start; i + 1 < offsets_data.size(); ++i) {
```

in pod_array.h, size function return type is size_t(uint64). 
There will be problem while size==0. You could test this simple case:
```
    uint64_t size = 0;
    for (int64_t i = 0; i < size -1; ++i) {
        std::cout << i << std::endl;
    }
```

eldenmoon: but offsets_data can't be zero, since previous code, but I coud add a DCHECK here
```
 if (*n == 0) {
        return Status::OK();
    }
``

**Review by cambyzju — DISMISSED** [review]
LGTM

**Comment by eldenmoon** [comment]
@adonis0147 plz review again

File: be/src/olap/rowset/segment_v2/column_reader.cpp [hunk 10]
```diff
@@ -466,6 +486,42 @@ Status ArrayFileColumnIterator::next_batch(size_t* n, ColumnBlockView* dst, bool
...
+    RETURN_IF_ERROR(_offset_iterator->seek_to_ordinal(ord));
+    if (_array_reader->is_nullable()) {
+        RETURN_IF_ERROR(_null_iterator->seek_to_ordinal(ord));
+    }
+    return _seek_by_offsets(ord);
+}
+
+Status ArrayFileColumnIterator::_caculate_offsets(ssize_t start,
+                                                  vectorized::MutableColumnPtr& offsets,
+                                                  size_t* num_items) {
```
adonis0147, line 507:
```suggestion
Status ArrayFileColumnIterator::_calculate_offsets(vectorized::ColumnArray::ColumnOffsets& offsets, ssize_t start)
```

1. Typo: `_caculate_offsets` -> `_calculate_offsets`
2. For function arguments, I think it is better to put the mutable arguments before the immutable ones.

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -466,6 +486,42 @@ Status ArrayFileColumnIterator::next_batch(size_t* n, ColumnBlockView* dst, bool
...
+    auto& column_offsets = static_cast<vectorized::ColumnArray::ColumnOffsets&>(*offsets);
+    ordinal_t last_offset = 0;
+    RETURN_IF_ERROR(_peek_one_offset(&last_offset));
+
+    auto& offsets_data = column_offsets.get_data();
+    *num_items = last_offset - offsets_data[start];
```
adonis0147, line 513: We would better do single job in one function. According to the name of this function, we should calculate the new offsets only and put the assignment of `num_items` **outside** the function.

adonis0147 (changes requested):

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -466,6 +486,42 @@ Status ArrayFileColumnIterator::next_batch(size_t* n, ColumnBlockView* dst, bool
...
+    // caculate real offsets
+    ordinal_t first_offset = offsets_data[start - 1]; // -1 is valid
+    ordinal_t first_ord = offsets_data[start];
+    for (ssize_t i = start; i + 1 < offsets_data.size(); ++i) {
+        offsets_data[i] = first_offset + (offsets_data[i + 1] - first_ord);
+    }
```
adonis0147, line 519: Why should we re-calculate the offsets here? Some questions here:

1. Does `vectorized::MutableColumnPtr& dst` in `ArrayFileColumnIterator::next_batch` share some data with others?
2. Does the `offsets_data` read from different pages cause the re-calculation?

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -466,6 +486,42 @@ Status ArrayFileColumnIterator::next_batch(size_t* n, ColumnBlockView* dst, bool
...
+Status ArrayFileColumnIterator::_caculate_offsets(ssize_t start,
+                                                  vectorized::MutableColumnPtr& offsets,
+                                                  size_t* num_items) {
+    auto& column_offsets = static_cast<vectorized::ColumnArray::ColumnOffsets&>(*offsets);
+    ordinal_t last_offset = 0;
+    RETURN_IF_ERROR(_peek_one_offset(&last_offset));
+
+    auto& offsets_data = column_offsets.get_data();
+    *num_items = last_offset - offsets_data[start];
+    // caculate real offsets
```
adonis0147, line 514: typo

be/src/olap/rowset/segment_v2/column_reader.cpp
eldenmoon: 2. But I saw many doris code which put immutable before mutable arguments

be/src/olap/rowset/segment_v2/column_reader.cpp
adonis0147: Ok, that is the scenario `ArrayFileColumnIterator::read_by_rowids`. What about the scenario in read data sequently?

eldenmoon: 1. 2 both answer is yes

adonis0147: Maybe there is a potential optimization point if we don't read by rowids. It seems that the item_writer doesn't reset the next_rowid in the same segment. So I don't think these answer are yes.

eldenmoon: sequenctly also should be recaculate, eg. (0, 1, 3, ....100, 103, 109, 114) , assume the offset=100 row_id = 20, we only need row_id from [20, 23), so after recaculate the offset became( 0, 3, 9)

adonis0147: Ok.

be/src/olap/rowset/segment_v2/column_reader.cpp
eldenmoon: I think both would be ok， since if we put the assignment of `num_items` outside the function, we need to do one more cast `static_cast<vectorized::ColumnArray::ColumnOffsets&>(*offsets)`, this seems a little duplicated code, num_items as a output of `_calculate_offsets` may be straightforward?

be/src/olap/rowset/segment_v2/column_reader.cpp
eldenmoon: done

be/src/olap/rowset/segment_v2/column_reader.cpp
eldenmoon: done

be/src/olap/rowset/segment_v2/column_reader.cpp
adonis0147: We could just do the cast outside and use the explicit type inside the function. No duplicate casts.

adonis0147 (dismissed):
LGTM

adonis0147 (dismissed):
LGTM

xy720 (approved):
LGTM

Issue #8881, opened 2022-04-07
[Feature] Disable all cast operation of HLL/Bitmap type


HappenLee:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Description

Now many operations will implicitly lead to cast of hll and bitmap types. This operation is useless and dangerous. We should disable all cast operation of hll/bitmap type. Directly express our operations on the corresponding types through functions, rather than relying on cast operation.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #12340, opened 2022-09-05
[Enhancement](spark-load) enable spark-load implicitly cast to bitmap

chenlinzhong:
# Proposed changes

Issue Number: close #xxx

## Problem summary

This PR(#8881) disables the HLL / bitmap implicit conversion, causing spark load to fail when importing bitmap type data
so i want to open back door for spark load ,allow implicit convert to bitmap

 **errCode = 2, detailMessage = type not match, originType=VARCHAR(*), targeType=BITMAP**

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/CastExpr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/CastExpr.java
@@ -62,6 +62,9 @@ public class CastExpr extends Expr {
     // True if this cast does not change the type.
     private boolean noOp = false;
 
+    // open back door for spark load ,ingore check type when used in spark-load
+    private boolean ignoreCheckType = false;
+
     private static final Map<Pair<Type, Type>, Function.NullableMode> TYPE_NULLABLE_MODE;
 
     static {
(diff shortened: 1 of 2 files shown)


spaces-X (dismissed):
LGTM

fe/fe-core/src/main/java/org/apache/doris/analysis/CastExpr.java
```diff
@@ -142,11 +142,6 @@ private static boolean disableRegisterCastingFunction(Type fromType, Type toType
                 || toType.isDateType())) {
             return true;
         }
-
```
spaces-X, line 145: Here is the back door is not just for spark load. All of the castExpr to bitmap are allowed.
I think we'd better figure out why the object store type was forbidden to be converted before, and which type conversions are definitely safe

PR #12341, opened 2022-09-05
[enhancement](array-type) Split Array Offsets and String Offsets

cambyzju:
# Proposed changes

Issue Number: close #xxx

## Problem summary
1. In old Doris version string offsets are 32bit, but it is not enough for Array type.
2. If we change string offsets from 32bit to 64bit, there will be problem if we upgrade BE one by one. Because at the same time 32bit Offsets and 64 bit Offsets String will exist at the same time.
3. As a result, we separate the Codes for Array Offsets.

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
4. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/vec/columns/column.h
+++ b/be/src/vec/columns/column.h
@@ -340,12 +340,18 @@ public:
     virtual void get_permutation(bool reverse, size_t limit, int nan_direction_hint,
                                  Permutation& res) const = 0;
 
+    // 32bit offsets for string
+    using Offset = UInt32;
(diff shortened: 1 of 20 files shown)


yiguolei (approved):
LGTM

PR #12389, opened 2022-09-06
[bugfix](be thread) not call rpc during cancel to hang bthread

yiguolei:
# Proposed changes

Issue Number: close #xxx

## Problem summary

When exec plan fragment failed to submit the executor to thread pool, be will try to cancel the executor and call update status to send FE the error msg. But it is in bthread, it will occupy the thread during call thrift rpc to fe. Other brpc request send to be will hang.

And also the cancel logic will check the exec status variable, but the variable maybe set by reporting thread and cancel will skip the later logic. Should remove check status logic.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/runtime/fragment_mgr.cpp
+++ b/be/src/runtime/fragment_mgr.cpp
@@ -658,9 +658,11 @@ Status FragmentMgr::exec_plan_fragment(const TExecPlanFragmentParams& params, Fi
             std::lock_guard<std::mutex> lock(_lock);
             _fragment_map.erase(params.params.fragment_instance_id);
         }
-        exec_state->cancel();
-        return Status::InternalError("Put planfragment to thread pool failed. err = {}, BE: {}",
-                                     st.get_error_msg(), BackendOptions::get_localhost());
+        exec_state->cancel(PPlanFragmentCancelReason::INTERNAL_ERROR,
+                           "Put planfragment to thread pool failed");
+        return Status::InternalError(
+                strings::Substitute("Put planfragment to thread pool failed. err = {}, BE: {}",
+                                    st.get_error_msg(), BackendOptions::get_localhost()));
     }
 
     return Status::OK();
--- a/be/src/runtime/fragment_mgr.cpp
+++ b/be/src/runtime/fragment_mgr.cpp
@@ -39,11 +39,11 @@
 #include "runtime/descriptors.h"
 #include "runtime/exec_env.h"
 #include "runtime/plan_fragment_executor.h"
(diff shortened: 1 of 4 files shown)


be/src/service/internal_service.cpp
```diff
@@ -296,6 +296,8 @@ void PInternalServiceImpl<T>::tablet_writer_cancel(google::protobuf::RpcControll
 template <typename T>
 Status PInternalServiceImpl<T>::_exec_plan_fragment(const std::string& ser_request,
                                                  PFragmentRequestVersion version, bool compact) {
+    // Could not parse fragment id from string here, just print a log to indicate that BE received a request
+    LOG(INFO) << "received a plan fragment request";
```
morningman, line 300: Remove this

be/src/service/internal_service.cpp
```diff
@@ -344,6 +346,8 @@ void PInternalServiceImpl<T>::cancel_plan_fragment(google::protobuf::RpcControll
     }
     if (!st.ok()) {
         LOG(WARNING) << "cancel plan fragment failed, errmsg=" << st.get_error_msg();
+    } else {
+        LOG(INFO) << "cancel fragment, fragment_instance_id=" << print_id(tid) << " successfully.";
```
morningman, line 350: This is already a cancel log in `fragment_mgr()->cancel`

be/src/runtime/fragment_mgr.cpp
```diff
@@ -248,31 +246,17 @@ Status FragmentExecState::execute() {
...
 
-Status FragmentExecState::cancel_before_execute() {
-    // set status as 'abort', cuz cancel() won't effect the status arg of DataSink::close().
-#ifndef BE_TEST
-    SCOPED_ATTACH_TASK(executor()->runtime_state());
-#endif
-    _executor.set_abort();
-    _executor.cancel();
-    if (_pipe != nullptr) {
-        _pipe->cancel("Execution aborted before start");
```
morningman, line 259: where to cancel this pipe now?

yiguolei: Status FragmentExecState::cancel(const PPlanFragmentCancelReason& reason, const std::string& msg)  this method will cancel pipe.

Issue #12303, opened 2022-09-02
[Feature] Support rollup and materialized view in nereids planner.


wangshuo128:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Description

Currently, OLAP table in nereids planner only supports the duplicated data model. Aggregate and Unique data models are not supported.
This issue is for supporting aggregate and unique data models and support rollup and materialized view selection.

Subtasks for the issue:

- [x] Main framework for selecting rollup index with aggregate plan node, pre-aggregation for OLAP scan is turned on by default. #12464 
- [x] Support turning off pre-aggregation for OLAP scan table, e.g., select data without aggregate plan node, select by non-matching aggregate function type, etc. #12785
- [ ] Support all the aggregate column types in the storage engine, such as bitmap, HLL, etc.
- [ ] Support complex expressions in an aggregate function that could use pre-aggregation.
- [x] Support materialized view selection for the duplicated data model. https://github.com/apache/doris/pull/13416
- [ ] Support merge-on-write of the unique key model in nereids planner. (https://cwiki.apache.org/confluence/display/DORIS/DSIP-018%3A+Support+Merge-On-Write+implementation+for+UNIQUE+KEY+data+model)

### Use case

N/A.

### Related issues

No.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #12464, opened 2022-09-08
[Feature](Nereids) Main framework for selecting rollup index.

wangshuo128:
# Proposed changes
First step of #12303 

## Problem summary

This is the first step for supporting rollup index selection for aggregate/unique key OLAP table.

This PR aims to select rollup index when the aggregate node is present and the aggregate function matches the value type. So pre-aggregation is turned on by default.  Cases that pre-aggregation should be turned off will be addressed in the next PR.

Main steps for rollup index selection: 

1. filter rollup indexes with all the required columns.
2. filter rollup indexes that match the key prefix most.
3. order the rollup indexes by row count, column count, rollup index id.

TODO remaining:
1. address cases that pre-aggregation should be turned off. (next PR)
2. add more test cases. 

Refactor
- Add `Project.getSlotToProducer` to extract a map from the project output slot to its producing expression.
- Add `Filter.getConjuncts` to split the filter condition to conjunctive predicates.
- Move the usage of `ExpressionReplacer` to `ExpressionUtils.replace(expr, replaceMap)` to simplify the code.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
5. Has unit tests been added:
    - [x] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments


Kikyou1997:
ExpressionUtils already has a function to extract conjuncts from an expression.

wangshuo128:
> ExpressionUtils already has a function to extract conjuncts from an expression.

Actually, `Filter. getConjuncts` would delegate to  `ExpressionUtils.extractConjunction`. Adding this method is to simplify the code.

fe/fe-core/src/main/java/org/apache/doris/nereids/rules/mv/SelectRollup.java
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    ///////////////////////////////////////////////////////////////////////////
+    // Split conjuncts into equal-to and non-equal-to.
+    ///////////////////////////////////////////////////////////////////////////
+
+    /**
+     * Split conjuncts input equal-to and non-equal-to.
+     */
+    private Map<Boolean, Set<String>> split(List<Expression> conjunct, Map<ExprId, String> exprIdToColName) {
```
Kikyou1997, line 288: Use Another concreter method name?

fe/fe-core/src/main/java/org/apache/doris/nereids/rules/mv/SelectRollup.java
```diff
@@ -0,0 +1,410 @@
...
+            OlapTable table,
+            MaterializedIndex rollup,
+            Set<String> equalColNames,
+            Set<String> nonEqualColNames) {
+        int matchCount = 0;
+        for (Column column : table.getSchemaByIndexId(rollup.getId())) {
+            if (equalColNames.contains(column.getName())) {
+                matchCount++;
+            } else if (nonEqualColNames.contains(column.getName())) {
+                // Unequivalence predicate's columns can match only first column in rollup.
```
Kikyou1997, line 271: I know this logic is same as the origin selector, but would you like to explain why `Unequivalence predicate's columns can match only first column in rollup.`? thx

Kikyou1997:
> > ExpressionUtils already has a function to extract conjuncts from an expression.
> 
> Actually, `Filter. getConjuncts` would delegate to `ExpressionUtils.extractConjunction`. Adding this method is to simplify the code.

ok

fe/fe-core/src/main/java/org/apache/doris/nereids/rules/mv/SelectRollup.java
```diff
@@ -0,0 +1,410 @@
...
+            this.type = result;
+        }
+
+        public static CheckResult createEqual(String name) {
+            return new CheckResult(name, ResultType.SUCCESS_EQUAL);
+        }
+
+        public static CheckResult createNonEqual(String name) {
+            return new CheckResult(name, ResultType.SUCCESS_NON_EQUAL);
+        }
```
Kikyou1997, line 322: need a verb between `could` and `prefix` here

924060929 (approved):

PR #12595, opened 2022-09-14
[fix](storage) Level1Iterator should release iterators in heap

dataroaring:
When there is a query with limit, e.g. select * from limit 1, scanners are closed by FragmentExecutor, then some scanners do not run to eof and mem leaks happen.

We have to release iterators in the heap.

# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/vec/olap/vcollect_iterator.cpp
+++ b/be/src/vec/olap/vcollect_iterator.cpp
@@ -23,6 +23,13 @@
 namespace doris {
 namespace vectorized {
 
+VCollectIterator::VCollectIterator() {
+}
+
+VCollectIterator::~VCollectIterator() {
+    DCHECK(_children.empty());
+}
+ 
 #define RETURN_IF_NOT_EOF_AND_OK(stmt)                                                  \
     do {                                                                                \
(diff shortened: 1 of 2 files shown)


be/src/vec/olap/vcollect_iterator.cpp
```diff
@@ -304,10 +311,12 @@ VCollectIterator::Level1Iterator::Level1Iterator(
...
-        }
+        delete child;
+    }
+    while (_heap != nullptr && !_heap->empty()) {
+        LevelIterator* it = _heap->top();
+        delete it;
+        _heap->pop();
     }
 
     if (_heap) {
```
BiteTheDDDDt, line 322: It looks like there is already the same logic in line322

PR #12604, opened 2022-09-14
[Improvement](sort) fallback to partial sort small block if topN is small

Gabriel39:
When `limit` is small in topN node and some string columns need to be materialized, `Block::merge` brings high overhead when accumulating big blocks. 

This PR add a new strategy, when `limit` is small and some string columns need to be materialized, we fallback to origin way to sort small blocks to avoid `merge`.

I tested on ckbench and find out 256 is a reasonable threshold to do this fallback.

# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments


be/src/vec/exec/vsort_node.cpp
```diff
@@ -182,6 +199,56 @@ Status VSortNode::sort_input(RuntimeState* state) {
...
+                            new SortCursorImpl(_sorted_blocks.back(), _sort_description)));
+                } else {
+                    SortBlockCursor block_cursor(
+                            _pool->add(new SortCursorImpl(block, _sort_description)));
+                    if (!block_cursor.totally_greater(_block_priority_queue.top())) {
+                        _sorted_blocks.emplace_back(std::move(block));
+                        _block_priority_queue.push(block_cursor);
+                        _total_mem_usage += mem_usage;
+                    } else {
+                        continue;
```
HappenLee, line 234: maybe in the case we should reuse the block mem?

be/src/vec/exec/vsort_node.cpp
```diff
@@ -182,6 +199,56 @@ Status VSortNode::sort_input(RuntimeState* state) {
...
+                    }
+                }
+            } else {
+                // dispose normal sort logic
+                _total_mem_usage += mem_usage;
```
HappenLee, line 239: why row == 0 do the work ?

Gabriel39: `_limit == -1` instead of `row == 0` here

be/src/vec/common/sort/sorter.cpp
```diff
@@ -0,0 +1,252 @@
...
+            *mem_reuse = false;
+        } else {
+            Block sorted_block;
+            sorted_block.swap(*block);
+            SortBlockCursor block_cursor(
+                    _pool->add(new SortCursorImpl(sorted_block, _sort_description)));
+            if (!block_cursor.totally_greater(_block_priority_queue.top())) {
+                _state->sorted_blocks.emplace_back(std::move(sorted_block));
+                _block_priority_queue.push(block_cursor);
+                *mem_reuse = false;
```
HappenLee, line 236: you can set false in the first line of the function. only need set `true` else

be/src/vec/common/sort/sorter.cpp
```diff
@@ -0,0 +1,252 @@
...
+    RETURN_IF_ERROR(_do_sort(block, mem_reuse));
+    return Status::OK();
+}
+
+Status TopNSorter::prepare_for_read() {
+    _state->build_merge_tree(_sort_description);
+    return Status::OK();
+}
+
+Status TopNSorter::get_next(RuntimeState* state, Block* block, bool* eos) {
```
HappenLee, line 197: seems the code is same as `FullSorter` `get_next` may should not virtual function and in `Sorter `

HappenLee (approved):
LGTM

PR #12666, opened 2022-09-16
[fix](memory) Fix BE OOM when load -238 fail

### LINKED ISSUE #12661: [Bug] BE OOM when load fails
### Search before asking

### Version

ead016e0d2567030c297277c1dca63198cc18efb

### What's Wrong?
When the load fails, the memory will increase instantly

After the reason load fails, flush will also fail, `wait flush` will return immediately, and then you can continue to add batches to memtable.

image

### What You Expected?

no oom

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

xinyiZzz:
# Proposed changes

Issue Number: close #12661

## Problem summary

When the flush is triggered when the load channel exceeds the mem limit, if the flush fails, an error message is returned and the load is terminated.

Usually flush failure is -238 error code. Because the memtable is frequently flushed after the load channel exceeds the mem limit, the number of segments exceeds the max value.

Before fix:
image

After fix:
image

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/runtime/tablets_channel.cpp
+++ b/be/src/runtime/tablets_channel.cpp
@@ -194,7 +194,8 @@ void TabletsChannel::_close_wait(DeltaWriter* writer,
     }
 }
 
-Status TabletsChannel::reduce_mem_usage(int64_t mem_limit) {
+template <typename TabletWriterAddResult>
+Status TabletsChannel::reduce_mem_usage(int64_t mem_limit, TabletWriterAddResult* response) {
     std::lock_guard<std::mutex> l(_lock);
     if (_state == kFinished) {
         // TabletsChannel is closed without LoadChannel's lock,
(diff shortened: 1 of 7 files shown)


be/src/runtime/tablets_channel.cpp
```diff
@@ -237,8 +238,22 @@ Status TabletsChannel::reduce_mem_usage(int64_t mem_limit) {
...
+                    "tablet writer failed to reduce mem consumption by flushing memtable, "
+                    "tablet_id=$0, txn_id=$1, err=$2, errcode=$3, msg:$4",
+                    writers[i]->tablet_id(), _txn_id, st.code(), st.precise_code(),
+                    st.get_error_msg());
+            LOG(WARNING) << err_msg;
+            PTabletError* error = tablet_errors->Add();
+            error->set_tablet_id(writers[i]->tablet_id());
+            error->set_msg(err_msg);
+            _broken_tablets.insert(writers[i]->tablet_id());
+        }
```
zhannngchen, line 256: If some error happened while `flush_memtable_and_wait`, we should not call `wait_flush` in L260, in that situation, `wait_flush` might wait infinitely?

xinyiZzz: `wait_flush` should not wait infinitely, and the flush token that fails to submit the flush will not actually wait.

Do not return error directly, pass the error to the sink to decide whether to terminate the load.

zhannngchen (approved):
LGTM

dataroaring (approved):
LGTM

be/src/runtime/tablets_channel.cpp
xinyiZzz: done, `_broken_tablets` does not `wait flush`

PR #12688, opened 2022-09-16
[fix](memtracker) Process physical mem check does not include tc/jemalloc allocator cache

xinyiZzz:
# Proposed changes

Issue Number: close #xxx

## Problem summary

tcmalloc/jemalloc allocator cache does not participate in the mem check as part of the process physical memory.

because `new/malloc` will trigger mem hook when using tcmalloc/jemalloc allocator cache, but it may not actually alloc physical memory, which is not expected in mem hook fail.

in addition:
1. The value of tcmalloc/jemalloc allocator cache is used as a mem tracker, the parent is the process mem tracker, which is updated every 1s.
2. Modify the process default `mem_limit` to 90%. expect mem tracker to effectively limit the memory usage of the process.

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -72,7 +72,7 @@ CONF_Int64(tc_max_total_thread_cache_bytes, "1073741824");
 // defaults to bytes if no unit is given"
 // must larger than 0. and if larger than physical memory size,
 // it will be set to physical memory size.
-CONF_String(mem_limit, "80%");
+CONF_String(mem_limit, "90%");
 
 // the port heartbeat service used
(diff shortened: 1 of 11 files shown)


dataroaring (approved):
LGTM

PR #12708, opened 2022-09-17
[fix](memtracker) Fix memory leak by calling `boost::stacktrace` in mem hook

xinyiZzz:
# Proposed changes

Issue Number: close #xxx

## Problem summary

After the consume mem tracker exceeds the mem limit in the mem hook, the boost stacktrace will be printed. A query/load will only be printed once, and the process tracker will only be printed once per second.

After the process memory reaches the upper limit, the boost stacktrace will be printed every second. The observed phenomena are as follows:
1. After query/load is canceled, the memory increases instantly;
2. tcmalloc profile total physical memory is less than perf process memory;
3. The process mem tracker is smaller than the perf process memory;

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/runtime/memory/mem_tracker_limiter.cpp
+++ b/be/src/runtime/memory/mem_tracker_limiter.cpp
@@ -213,7 +213,8 @@ void MemTrackerLimiter::print_log_usage(const std::string& msg) {
         } else {
             detail += "\n" + log_usage();
         }
-        detail += "\n" + boost::stacktrace::to_string(boost::stacktrace::stacktrace());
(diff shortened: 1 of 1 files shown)


dataroaring (approved):
LGTM

PR #12714, opened 2022-09-18
 [feature-wip](statistics) template for building internal query SQL statements

weizhengte:
# Proposed changes

Template for building internal query SQL statements，it mainly used for statistics module. After the template is defined, the executable statement will be built after the given parameters.

For example, template and parameters:
- template: `SELECT ${col} FROM ${table} WHERE id = ${id};`,
- parameters:  `{col=colName, table=tableName, id=1}`
- result sql:  `SELECT colName FROM tableName WHERE id = 1;`

usage:
```
String template = "SELECT * FROM ${table} WHERE id = ${id};";
Map<String, String> params = new HashMap<>();
params.put("table", "table0");
params.put("id", "123");

// result: SELECT * FROM table0 WHERE id = 123;
String result = InternalSqlTemplate.processTemplate(template, params);
```

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [x] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/statistics/util/InternalSqlTemplate.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/statistics/util/InternalSqlTemplate.java
@@ -1,0 +1,295 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package org.apache.doris.statistics.util;
+
+import org.apache.doris.common.InvalidFormatException;
+
+import com.google.common.collect.Sets;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Template for building internal query SQL statements.
+ * e.g.:
+ *   - "SELECT ${column} FROM ${table}"
+ *   - ${column} and ${table} will be replaced with the actual executed table and column.
+ */
+public class InternalSqlTemplate {
+    /** -------------------------- for statistics begin -------------------------- */
+    public static final String MIN_VALUE_SQL = "SELECT MIN(${column}) AS min_value FROM ${table};";
+    public static final String PARTITION_MIN_VALUE_SQL = "SELECT MIN(${column}) AS min_value"
+            + " FROM ${table} PARTITION (${partition});";
+
+    public static final String MAX_VALUE_SQL = "SELECT MAX(${column}) AS max_value FROM ${table};";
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/org/apache/doris/statistics/util/InternalSqlTemplate.java
```diff
@@ -0,0 +1,282 @@
...
+     *   result sql: 'SELECT colName FROM tableName WHERE id = 1;
+     * <p>
+     *
+     * @param template sql template
+     * @param params   k,v parameter, if without parameter, params should be null
+     * @return SQL statement with parameters concatenated
+     */
+    public static String processTemplate(String template, Map<String, String> params) {
+        Matcher matcher = PATTERN.matcher(template);
+        StringBuffer sb = new StringBuffer();
```
Kikyou1997, line 97: It's a local variable which means there is no race condition, so we don't need `StringBuffer` Here, `StringBuilder` is  much enough

fe/fe-core/src/main/java/org/apache/doris/statistics/util/InternalSqlTemplate.java
```diff
@@ -0,0 +1,282 @@
...
+        if (params != null) {
+            Set<String> paramsSet = params.keySet();
+            return paramsSet.containsAll(requiredParams);
+        } else {
+            return requiredParams == null;
+        }
+    }
+
+    public static String buildStatsMinValueSql(Map<String, String> params) throws InvalidFormatException {
+        Set<String> requiredParams = Sets.newHashSet("table", "column");
```
Kikyou1997, line 121: We'd better to wrap the requiredParams and SQL template string to a individual class rather  than hard code it in every single method.

fe/fe-core/src/main/java/org/apache/doris/statistics/util/InternalSqlTemplate.java
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+        matcher.appendTail(sb);
+        LOG.debug("Template:{}, params: {}, SQL: {}", template, params, sb.toString());
+
+        return sb.toString();
+    }
+
+    private static boolean checkParams(Set<String> requiredParams, Map<String, String> params) {
+        if (params != null) {
```
Kikyou1997, line 112: Pass a empty collections is much better than null I think

Kikyou1997:
review plz @morrySnow

Kikyou1997 (dismissed):

Kikyou1997:
Besides, I'm not sure whether simply use regex to do the mach and replace would lead to the SQL injection or this problem could be tolerant since Doris is normaly used as a internal service

fe/fe-core/src/main/java/org/apache/doris/statistics/util/InternalSqlTemplate.java
weizhengte: It seems that using StringBuilder makes the following appendReplacement() and appendTail() throw exceptions (prompting me to use jdk9++, Usage of API documented as @since 9++), so I used StringBuffer here.

fe/fe-core/src/main/java/org/apache/doris/statistics/util/InternalSqlTemplate.java
weizhengte: sure.

weizhengte:
add label area/statistics and project plz @morrySnow

fe/fe-core/src/main/java/org/apache/doris/statistics/util/InternalSqlTemplate.java
Kikyou1997: Ok

morrySnow (dismissed):

morrySnow (approved):

PR #12792, opened 2022-09-20
[Docs] Add python env for Mac M1

EmmyMiao87:
For Mac M1, the default is python3 instead of python. When FE compiles, there will be an error that python cannot be found. This PR complements this part of the description.


morningman (approved):
LGTM

liuzhuang2017 (approved):
Thanks you contribute,LGTM.

PR #12802, opened 2022-09-21
[feature-wip](multi-catalog) Support flatten catalog

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Doris is now support 3 level schema:
`Catalog -> Database -> Table`

However, when using some MySQL-compatible tools to connect to Doris,
these tools only support two-level (Database -> Table) metadata,
so the database information in the external catalog cannot be displayed.

This PR attempts to make compatibility improvements to this issue.

In normal use, when users need to switch to other catalogs, they can use the switch statement to switch, such as:
`switch hms_catalog;`

and then use the `use` statement to switch the database:
`use db;`

At the same time, we can directly use the following commands to directly switch the catalog and database:
`use __hms_catalog__db;`

Here we use `__` as a special identifier so that the system can identify and parse out the catalog name.
At the same time, we prohibit `__` in the catalog name, so as to ensure that there are no semantic errors.

This change does not require any additional configuration, and Doris will automatically perform the correct
operation based on the name.

I have tested mysql command line tool and dbeaver, which works fine with this feature.

```
$ mysql -h 127.0.0.1 -P 9030 -uroot -D__hive__tpch100;
mysql> show tables;
+-------------------+
| Tables_in_tpch100 |
+-------------------+
| customer          |
| lineitem          |
| nation            |
| orders            |
| part              |
| partsupp          |
| region            |
| supplier          |
+-------------------+
8 rows in set (0.00 sec)
```

The document will be added later with full `multi-catalog` feature.

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [x] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/BaseTableRef.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/BaseTableRef.java
@@ -66,7 +66,6 @@ public class BaseTableRef extends TableRef {
      */
     @Override
     public void analyze(Analyzer analyzer) throws UserException {
-        name = analyzer.getFqTableName(name);
         name.analyze(analyzer);
         desc = analyzer.registerTableRef(this);
         isAnalyzed = true;  // true that we have assigned desc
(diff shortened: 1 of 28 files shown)


stalary:
Show databases can return all database?

morningman:
> 

No, it still only return database in current catalog.
I thought about this feature. This feature requires additional config to tell the system whether to return all databasses or not,
So I'm hesitant to add this feature.

PR #12803, opened 2022-09-21
[Opt](Vectorized) Support push down no grouping agg

HappenLee:
# Proposed changes

Use the segment static to speed up some not grouping agg query：`min`, `max`, `count`

Here need to a new SegmentIter call `StatisticsIter` to do the work.

lineorder : 60kw line num

query             Before             After
count(*)         0.5s                0.02s
min                0.4s                0.01s
max               0.4s                 0.01s

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_v2/column_reader.cpp
+++ b/be/src/olap/rowset/segment_v2/column_reader.cpp
@@ -171,6 +171,45 @@ Status ColumnReader::get_row_ranges_by_zone_map(
     return Status::OK();
 }
 
+Status ColumnReader::next_batch_of_zone_map(size_t* n, vectorized::MutableColumnPtr& dst) const {
+    // TODO: this work to get min/max value seems should only do once
+    FieldType type = _type_info->type();
+    std::unique_ptr<WrapperField> min_value(WrapperField::create_by_type(type, _meta.length()));
+    std::unique_ptr<WrapperField> max_value(WrapperField::create_by_type(type, _meta.length()));
+    _parse_zone_map(_zone_map_index_meta->segment_zone_map(), min_value.get(), max_value.get());
+
+    dst->reserve(*n);
+    bool is_string = is_olap_string_type(type);
+    if (max_value->is_null()) {
+        assert_cast<vectorized::ColumnNullable&>(*dst).insert_default();
+    } else {
+        if (is_string) {
+            auto sv = (StringValue*)max_value->cell_ptr();
+            dst->insert_data(sv->ptr, sv->len);
+        } else {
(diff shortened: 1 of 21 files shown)


be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -171,6 +171,44 @@ Status ColumnReader::get_row_ranges_by_zone_map(
...
+    bool is_string = is_olap_string_type(type);
+    if (max_value->is_null()) {
+        assert_cast<vectorized::ColumnNullable&>(*dst).insert_default();
+    } else {
+        if (is_string) {
+            auto sv = (StringValue*)max_value->cell_ptr();
+            dst->insert_data(sv->ptr, sv->len);
+        } else {
+            dst->insert_many_fix_len_data(static_cast<const char*>(max_value->cell_ptr()), 1);
+        }
```
BiteTheDDDDt, line 191: why not just use `insert_data` here?

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -171,6 +171,44 @@ Status ColumnReader::get_row_ranges_by_zone_map(
...
+        assert_cast<vectorized::ColumnNullable&>(*dst).insert_null_elements(size);
+    } else {
+        if (is_string) {
+            auto sv = (StringValue*)min_value->cell_ptr();
+            dst->insert_many_data(sv->ptr, sv->len, size);
+        } else {
+            // TODO: the work may cause performance problem, opt latter
+            for (int i = 0; i < size; ++i) {
+                dst->insert_many_fix_len_data(static_cast<const char*>(min_value->cell_ptr()), 1);
+            }
```
BiteTheDDDDt, line 205: Maybe we can use `insert_many_fix_len_data( ,size)`

HappenLee: en，just add a todo here to support in the future

be/src/olap/rowset/segment_v2/column_reader.cpp
HappenLee: may have uint24_t use in `Date` , `insert_data` may cause bug.

BiteTheDDDDt (dismissed):
LGTM

BiteTheDDDDt (approved):
LGTM

PR #12866, opened 2022-09-22
[enhancement](compaction) introduce segment compaction (#12609)

### LINKED ISSUE #12609: [Enhancement] segment compaction
### Search before asking

### Description

-238 (OLAP_ERR_TOO_MANY_SEGMENTS) is a common issue when loading large data.
Things get worse when data has low compression ratio or memtables flushed due to exceeding memory limit. Because too mush small segments are created.
see: https://github.com/apache/doris/issues?q=-238

### Solution

Rowsetwise compaction happens when load (transaction) is done.
Maybe we should do compaction DURING load -- segmentwise compaction.

### Code of Conduct

freemandealer:
Implement segmentwise compaction during rowset write to reduce the number of segments produced by load jobs, otherwise may cause OLAP_ERR_TOO_MANY_SEGMENTS (-238).

# Proposed changes

Issue Number: close #12609

## Introduction

The default limit is 200 segment perf rowset. Too many segments may fail the whole load process (OLAP_ERR_TOO_MANY_SEGMENTS -238). If we increase the limit, the load will succeed but the pressure is transferred to the subsequential rowsetwise  compaction. Things get worse when the user issue a query, e.g. insert into select stmt, right after load job but before rowsetwise compaction, he/she will suffer the performance disaster or maybe end up with OOM.

So we are introducing segmentwise compaction which will compact data DURING the write process, instead of waiting for rowsetwise compaction until txn has been committed.

### Trigger

Every time when a rowset writer produces more than N (e.g. 10) segments, we trigger segment compaction. Note that only one segment compaction job for a single rowset at a time to ensure no recursing/queuing nightmare.

### Target Selection

We collect segments during every trigger. We skip big segments whose row num > M (e.g. 10000) coz we get little benefits from compacting them comparing our effort. Hence, we only pick the 'Longest Consecutive Small" segment group to do actual compaction.

### Compaction Process

A new thread pool is introduced to help do the job. We submit the above-mentioned 'Longest Consecutive Small" segment group to the pool. Then the worker thread does the followings:

- build a MergeIterator from the target segments
- create a new segment writer
- for each block readed from MergeIterator, the Writer append it

### SegID handling

SegID must remain consecutive after segment compaction. 

If a rowset has small segments named seg_0, seg_1, seg_2, seg_3 and a big segment seg_4:

- we create a segment named "seg_0-3" to save compacted data for seg_0, seg_1, seg_2 and seg_3
- delete seg_0, seg_1, seg_2 and seg_3
- rename seg_0-3 to seg_0
- rename seg_4 to seg_1

It is worth noticing that we should wait inflight segment compaction tasks to finish before building rowset meta and committing this txn.

### The amount of data can Doris load 

First, we test the data amount that we can successfully load into doris disable/enable segment compaction.Tests are based on TPCH. Table is created as 1 bucket and no parallel. We trigger segment compaction every 10 segments produced by rowset writer.

| cases                 | data amount                |
| --------------------- | -------------------------- |
| Disable SegCompaciton | 1.12 million rows, 18.67GB |
| Enable SegCompaction  | 11 million rows, 183GB     |

The result shows that the amount of data we can load to doris improve 10 times after enabling segment compaction. The ratio is correspond to the triggering segment number.

### Impact on latency

When segment compaction is disabled, a load job will finish in 1260s during the test. And the sequential rowsetwise compaction cost 151s.

We give the test results when enabling segment compaction in different triggering segment number:

| triggering segment number    | Load Latency | RowsetCompaction Latency |
| ---------------------------- | ------------ | ------------------------ |
| 5 (trigger every 5 segments) | 1089s (-13%)  | 242s (+60%)              |
| 10                           | 1053s (-16%) | 166s (+9%)               |
| 20                           | 960s (-23%)  | 172s (+13%)              |
| 40                           | 1320s (+4%)  | 169s (+11%)              |

We load without segment compaction for serveral times and each gives us a different latency range from (-25%, +25%). So we believe that segment compaction has little impact on the latency.

~In addition to the above costs, we wait inflight segment compaction tasks to finish before building rowset meta and publishing the data. The length of the wait time depends on when the build takes the place but there is a theoretical range for it and the range is related to the time each segment compaction task will cost:~

| triggering segment number | Single SegCompaction Task Latency |
| ------------------------- | --------------------------------- |
| 5                         | ~5s~                                |
| 10                        | ~9s~                                |
| 20                        | ~20s~                               |
| 40                        | ~60s~                               |

[UPDATE] no more wait time since we don't do segcompaction for the last segments before build. We only rename them.

### Impact on memory usage

Compaction itself will consume memory. The following test results show the memory footprint when enabling segment compaction.

When segment compaction disabled, a load job will use 4.83% of 128GB memory. And the sequential rowsetwise compaction takes 8.21%.

When enable segment compaction:

| triggering segment number    | Load Memory Usage       | RowsetCompaction Memory Usage |
| ---------------------------- | ----------------------- | ----------------------------- |
| 5 (trigger every 5 segments) | Avg.:6.62%  Peak:7.01%  | 6.9% (-16%)                   |
| 10                           | Avg.:7.63%   Peak:9.61% | 6.56% (-20%)                  |
| 20                           | Avg.:7.6%     Peak:9.8% | 6.9% (-16%)                   |
| 40                           | Avg.:5.09%  Peak:9%     | 6.62% (-19%)                  |

Segment compaction uses more memory because we add another segment writer to write compacted data and multiple segment readers to read source data for each active rowset.

However, since data are more ordered and the number of segment is decreased after segment compaction, RowsetCompaction uses less memory.

## Checklist(Required)

    - [X] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [X] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [X] No Need
    - [ ] Yes
    - [X] No
    - [X] No

## Further comments

### DIFF
--- a/be/src/olap/delta_writer.cpp
+++ b/be/src/olap/delta_writer.cpp
@@ -171,9 +171,6 @@ Status DeltaWriter::write(Tuple* tuple) {
     // if memtable is full, push it to the flush executor,
     // and create a new memtable for incoming data
     if (_mem_table->memory_usage() >= config::write_buffer_size) {
-        if (++_segment_counter > config::max_segment_num_per_rowset) {
-            return Status::OLAPInternalError(OLAP_ERR_TOO_MANY_SEGMENTS);
-        }
         RETURN_NOT_OK(_flush_memtable_async());
         // create a new memtable for new incoming data
         _reset_mem_table();
@@ -230,9 +227,6 @@ Status DeltaWriter::write(const vectorized::Block* block, const std::vector<int>
 }
 
 Status DeltaWriter::_flush_memtable_async() {
-    if (++_segment_counter > config::max_segment_num_per_rowset) {
-        return Status::OLAPInternalError(OLAP_ERR_TOO_MANY_SEGMENTS);
-    }
     return _flush_token->submit(std::move(_mem_table));
 }
 
--- a/be/src/olap/delta_writer.cpp
+++ b/be/src/olap/delta_writer.cpp
@@ -56,8 +56,6 @@ DeltaWriter::~DeltaWriter() {
         _garbage_collection();
     }
 
-    _mem_table.reset();
-
     if (!_is_init) {
         return;
     }
@@ -77,6 +75,8 @@ DeltaWriter::~DeltaWriter() {
         _tablet->data_dir()->remove_pending_ids(ROWSET_ID_PREFIX +
                                                 _rowset_writer->rowset_id().to_string());
     }
+
+    _mem_table.reset();
 }
 
 void DeltaWriter::_garbage_collection() {
@@ -171,9 +171,12 @@ Status DeltaWriter::write(Tuple* tuple) {
     // if memtable is full, push it to the flush executor,
     // and create a new memtable for incoming data
     if (_mem_table->memory_usage() >= config::write_buffer_size) {
-        RETURN_NOT_OK(_flush_memtable_async());
+        auto s = _flush_memtable_async();
         // create a new memtable for new incoming data
         _reset_mem_table();
+        if (OLAP_UNLIKELY(!s.ok())) {
+            return s;
+        }
     }
     return Status::OK();
 }
@@ -193,8 +196,11 @@ Status DeltaWriter::write(const RowBatch* row_batch, const std::vector<int>& row
     }
 
     if (_mem_table->memory_usage() >= config::write_buffer_size) {
-        RETURN_NOT_OK(_flush_memtable_async());
+        auto s = _flush_memtable_async();
         _reset_mem_table();
+        if (OLAP_UNLIKELY(!s.ok())) {
+            return s;
+        }
     }
 
     return Status::OK();
@@ -218,8 +224,11 @@ Status DeltaWriter::write(const vectorized::Block* block, const std::vector<int>
     if (_mem_table->need_to_agg()) {
         _mem_table->shrink_memtable_by_agg();
         if (_mem_table->is_flush()) {
-            RETURN_NOT_OK(_flush_memtable_async());
+            auto s = _flush_memtable_async();
             _reset_mem_table();
+            if (OLAP_UNLIKELY(!s.ok())) {
+                return s;
+            }
         }
     }
 
@@ -247,8 +256,11 @@ Status DeltaWriter::flush_memtable_and_wait(bool need_wait) {
     VLOG_NOTICE << "flush memtable to reduce mem consumption. memtable size: "
                 << _mem_table->memory_usage() << ", tablet: " << _req.tablet_id
                 << ", load id: " << print_id(_req.load_id);
-    RETURN_NOT_OK(_flush_memtable_async());
+    auto s = _flush_memtable_async();
     _reset_mem_table();
+    if (OLAP_UNLIKELY(!s.ok())) {
+        return s;
+    }
 
     if (need_wait) {
         // wait all memtables in flush queue to be flushed.
@@ -295,9 +307,13 @@ Status DeltaWriter::close() {
         return Status::OLAPInternalError(OLAP_ERR_ALREADY_CANCELLED);
     }
 
-    RETURN_NOT_OK(_flush_memtable_async());
+    auto s = _flush_memtable_async();
     _mem_table.reset();
-    return Status::OK();
+    if (OLAP_UNLIKELY(!s.ok())) {
+        return s;
+    } else {
+        return Status::OK();
+    }
 }
 
 Status DeltaWriter::close_wait(const PSlaveTabletNodes& slave_tablet_nodes,
--- a/be/src/olap/rowset/beta_rowset_writer.cpp
+++ b/be/src/olap/rowset/beta_rowset_writer.cpp
@@ -35,6 +35,7 @@
 #include "olap/rowset/segment_v2/segment_writer.h"
 //#include "olap/storage_engine.h"
 #include "runtime/exec_env.h"
+#include "runtime/memory/mem_tracker_limiter.h"
 
 namespace doris {
 
@@ -49,6 +50,7 @@ BetaRowsetWriter::BetaRowsetWriter()
           _num_rows_written(0),
           _total_data_size(0),
           _total_index_size(0),
+          _raw_num_rows_written(0),
           _is_doing_segcompaction(false) {
     _segcompaction_status.store(OLAP_SUCCESS);
 }
@@ -264,6 +266,13 @@ Status BetaRowsetWriter::_check_correctness(std::unique_ptr<OlapReaderStatistics
 }
 
 Status BetaRowsetWriter::_do_segcompaction(SegCompactionCandidatesSharedPtr segments) {
+    SCOPED_ATTACH_TASK(StorageEngine::instance()->segcompaction_mem_tracker(),
+                       ThreadContext::TaskType::COMPACTION);
+    // throttle segcompaction task if memory depleted.
+    if (MemTrackerLimiter::sys_mem_exceed_limit_check(GB_EXCHANGE_BYTE)) {
+        LOG(WARNING) << "skip segcompaction due to memory shortage";
+        return Status::MemoryLimitExceeded("skip segcompaction due to memory shortage");
+    }
     uint64_t begin = (*(segments->begin()))->id();
     uint64_t end = (*(segments->end() - 1))->id();
     LOG(INFO) << "BetaRowsetWriter:" << this << " do segcompaction at " << segments->size()
@@ -280,6 +289,8 @@ Status BetaRowsetWriter::_do_segcompaction(SegCompactionCandidatesSharedPtr segm
         LOG(WARNING) << "failed to get segcompaction reader";
         return Status::OLAPInternalError(OLAP_ERR_SEGCOMPACTION_INIT_READER);
     }
+    std::unique_ptr<vectorized::VMergeIterator> reader_ptr;
+    reader_ptr.reset(reader);
     auto writer = _create_segcompaction_writer(begin, end);
     if (UNLIKELY(writer == nullptr)) {
         LOG(WARNING) << "failed to get segcompaction writer";
@@ -288,10 +299,12 @@ Status BetaRowsetWriter::_do_segcompaction(SegCompactionCandidatesSharedPtr segm
     uint64_t row_count = 0;
     vectorized::Block block = _context.tablet_schema->create_block();
     while (true) {
-        auto status = reader->next_batch(&block);
+        auto status = reader_ptr->next_batch(&block);
         row_count += block.rows();
         if (status != Status::OK()) {
             if (LIKELY(status.is_end_of_file())) {
+                RETURN_NOT_OK_LOG(_add_block_for_segcompaction(&block, &writer),
+                                  "write block failed");
                 break;
             } else {
                 LOG(WARNING) << "read block failed: " << status.to_string();
@@ -536,7 +549,7 @@ Status BetaRowsetWriter::_add_block(const vectorized::Block* block,
         row_offset += input_row_num;
     } while (row_offset < block_row_num);
 
-    //_num_rows_written += block_row_num;
+    _raw_num_rows_written += block_row_num;
     return Status::OK();
 }
 
@@ -566,6 +579,7 @@ Status BetaRowsetWriter::_add_row(const RowType& row) {
                       _segment_writer->num_rows_written() >= _context.max_rows_per_segment)) {
         RETURN_NOT_OK(_flush_segment_writer(&_segment_writer));
     }
+    ++_raw_num_rows_written;
     return Status::OK();
 }
 
@@ -602,7 +616,8 @@ Status BetaRowsetWriter::flush() {
 }
 
 Status BetaRowsetWriter::flush_single_memtable(MemTable* memtable, int64_t* flush_size) {
-    int64_t current_flush_size = _total_data_size + _total_index_size;
+    int64_t size = 0;
+    int64_t sum_size = 0;
     // Create segment writer for each memtable, so that
     // all memtables can be flushed in parallel.
     std::unique_ptr<segment_v2::SegmentWriter> writer;
@@ -622,15 +637,24 @@ Status BetaRowsetWriter::flush_single_memtable(MemTable* memtable, int64_t* flus
 
         if (PREDICT_FALSE(writer->estimate_segment_size() >= MAX_SEGMENT_SIZE ||
                           writer->num_rows_written() >= _context.max_rows_per_segment)) {
-            RETURN_NOT_OK(_flush_segment_writer(&writer));
+            auto s = _flush_segment_writer(&writer, &size);
+            sum_size += size;
+            if (OLAP_UNLIKELY(!s.ok())) {
+                *flush_size = sum_size;
+                return s;
+            }
         }
     }
 
     if (writer != nullptr) {
-        RETURN_NOT_OK(_flush_segment_writer(&writer));
+        auto s = _flush_segment_writer(&writer, &size);
+        sum_size += size;
+        *flush_size = sum_size;
+        if (OLAP_UNLIKELY(!s.ok())) {
+            return s;
+        }
     }
 
-    *flush_size = (_total_data_size + _total_index_size) - current_flush_size;
     return Status::OK();
 }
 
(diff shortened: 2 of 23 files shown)


be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -102,6 +110,284 @@ Status BetaRowsetWriter::add_block(const vectorized::Block* block) {
...
+    std::stringstream ss;
+    for (auto& segment : (*segments.get())) {
+        ss << "[id:" << segment->id() << " num_rows:" << segment->num_rows() << "]";
+    }
+    LOG(INFO) << "candidate segments num:" << s << " list of candidates:" << ss.str();
+}
+
+SegCompactionCandidatesSharedPtr BetaRowsetWriter::get_segcompaction_candidates(bool is_last) {
+    SegCompactionCandidatesSharedPtr segments = std::make_shared<SegCompactionCandidates>();
+    if (is_last) {
```
zhannngchen, line 340: 1. Should we check the `segcompaction_small_threshold` here as well?
e.g. you compacted 10 segments, for some reason it works very slow, and after it finished, all load data have been written, generated 100 segments ,all of them are greater than 1MB, we should not compact them?

2. In the previous case, if there left 100 segments, should we do segment compaction on all of them at once?

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -102,6 +110,284 @@ Status BetaRowsetWriter::add_block(const vectorized::Block* block) {
...
+        find_longest_consecutive_small_segment(segments);
+    }
+    return segments;
+}
+
+void BetaRowsetWriter::segcompaction_if_necessary() {
+    if (!config::enable_segcompaction || !config::enable_storage_vectorization) {
+        return;
+    }
+    if (!_is_doing_segcompaction &&
```
zhannngchen, line 357: Two threads can both see `_is_doing_segcompaction` as false, and enter the following block?

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -102,6 +110,284 @@ Status BetaRowsetWriter::add_block(const vectorized::Block* block) {
...
+
+void BetaRowsetWriter::segcompaction_ramaining_if_necessary() {
+    if (!config::enable_segcompaction || !config::enable_storage_vectorization) {
+        return;
+    }
+    if (!is_segcompacted() || _segcompacted_point == _num_segment) {
+        // no need if never segcompact before or all segcompacted
+        return;
+    }
+    CHECK_EQ(_is_doing_segcompaction, false);
```
zhannngchen, line 379: DCHECK_EQ

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -219,17 +516,35 @@ Status BetaRowsetWriter::flush_single_memtable(const vectorized::Block* block) {
...
     RETURN_NOT_OK(_add_block(block, &writer));
     RETURN_NOT_OK(_flush_segment_writer(&writer));
     return Status::OK();
 }
 
+void BetaRowsetWriter::wait_flying_segcompaction() {
+    uint64_t begin_wait = GetCurrentTimeMicros();
+    while (_is_doing_segcompaction) { // TODO: memory barrier?
+        // change sync wait to async?
+        std::unique_lock<std::mutex> l(_segcompacting_cond_lock);
```
zhannngchen, line 531: the lock is not used anywhere else, maybe thread token is more suitable?

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -309,12 +641,23 @@ Status BetaRowsetWriter::_create_segment_writer(
...
     writer_options.enable_unique_key_merge_on_write = _context.enable_unique_key_merge_on_write;
-    writer->reset(new segment_v2::SegmentWriter(file_writer.get(), _num_segment,
-                                                _context.tablet_schema, _context.data_dir,
-                                                _context.max_rows_per_segment, writer_options));
-    {
-        std::lock_guard<SpinLock> l(_lock);
-        _file_writers.push_back(std::move(file_writer));
+
+    if (is_segcompaction) {
+        writer->reset(new segment_v2::SegmentWriter(file_writer.get(), _num_segcompacted + 1,
```
zhannngchen, line 646: The only difference of these 2 branch is the parameter sgement_id?

be/src/olap/rowset/beta_rowset_writer.h
```diff
@@ -73,42 +77,81 @@ class BetaRowsetWriter : public RowsetWriter {
...
 
+    Status _do_create_segment_writer(std::unique_ptr<segment_v2::SegmentWriter>* writer,
+                                     bool is_segcompaction, int64_t begin, int64_t end);
     Status _create_segment_writer(std::unique_ptr<segment_v2::SegmentWriter>* writer);
+    Status _create_segment_writer_for_segcompaction(
+            std::unique_ptr<segment_v2::SegmentWriter>* writer, uint64_t begin, uint64_t end);
 
     Status _flush_segment_writer(std::unique_ptr<segment_v2::SegmentWriter>* writer);
     void _build_rowset_meta(std::shared_ptr<RowsetMeta> rowset_meta);
+    void segcompaction_if_necessary();
```
zhannngchen, line 98: private method should start with _

be/src/olap/rowset/beta_rowset_writer.h
```diff
@@ -73,42 +77,81 @@ class BetaRowsetWriter : public RowsetWriter {
...
     std::atomic<int64_t> _total_index_size;
     // TODO rowset Zonemap
 
+    struct statistics {
+        int64_t row_num;
+        int64_t data_size;
+        int64_t index_size;
+        KeyBoundsPB key_bounds;
+    };
+    std::map<uint32_t, statistics> _segid_statistics_map;
```
zhannngchen, line 143: map is not thread safe, we should hold a lock when writing it.

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -102,6 +110,284 @@ Status BetaRowsetWriter::add_block(const vectorized::Block* block) {
...
+void BetaRowsetWriter::rename_compacted_segment_plain(uint64_t seg_id) {
+    int ret;
+    auto src_seg_path =
+            BetaRowset::local_segment_path(_context.tablet_path, _context.rowset_id, seg_id);
+    auto dst_seg_path = BetaRowset::local_segment_path(_context.tablet_path, _context.rowset_id,
+                                                       _num_segcompacted++);
+    LOG(INFO) << "segcompaction skip this segment. rename " << src_seg_path << " to "
+              << dst_seg_path;
+    if (src_seg_path.compare(dst_seg_path) != 0) {
+        CHECK_EQ(_segid_statistics_map.find(seg_id + 1) == _segid_statistics_map.end(), false);
```
zhannngchen, line 194: DCHECK_EQ

be/src/olap/rowset/beta_rowset_writer.cpp
zhannngchen: skip last several small segments.

docs/zh-CN/docs/admin-manual/maint-monitor/be-olap-error-code.md
```diff
@@ -6,7 +6,7 @@
 
 ---
 
-<!-- 
+<!--
```
freemandealer, line 9: unecessary

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -304,60 +397,76 @@ void BetaRowsetWriter::find_longest_consecutive_small_segment(
...
         ss << "[id:" << segment->id() << " num_rows:" << segment->num_rows() << "]";
     }
     LOG(INFO) << "candidate segments num:" << s << " list of candidates:" << ss.str();
+    return Status::OK();
 }
 
-SegCompactionCandidatesSharedPtr BetaRowsetWriter::get_segcompaction_candidates(bool is_last) {
-    SegCompactionCandidatesSharedPtr segments = std::make_shared<SegCompactionCandidates>();
+Status BetaRowsetWriter::_get_segcompaction_candidates(SegCompactionCandidatesSharedPtr& segments,
+                                                      bool is_last) {
```
freemandealer, line 429: indent

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -165,52 +171,91 @@ Status BetaRowsetWriter::delete_original_segments(uint32_t begin, uint32_t end)
...
                                                        _num_segcompacted++);
     LOG(INFO) << "segcompaction skip this segment. rename " << src_seg_path << " to "
               << dst_seg_path;
-    if (src_seg_path.compare(dst_seg_path) != 0) {
-        CHECK_EQ(_segid_statistics_map.find(seg_id + 1) == _segid_statistics_map.end(), false);
-        CHECK_EQ(_segid_statistics_map.find(_num_segcompacted) == _segid_statistics_map.end(),
-                 true);
+    {
+        std::lock_guard<std::mutex> lock(_segid_statistics_map_mutex);
+        DCHECK_EQ(_segid_statistics_map.find(seg_id + 1) == _segid_statistics_map.end(), false);
```
zhannngchen, line 208: rebase to master, and use `seg_id` instead of `seg_id + 1`

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -304,60 +397,76 @@ void BetaRowsetWriter::find_longest_consecutive_small_segment(
...
+    }
+}
+
+Status BetaRowsetWriter::_segcompaction_if_necessary() {
+    if (!config::enable_segcompaction || !config::enable_storage_vectorization ||
+        !_check_and_set_is_doing_segcompaction()) {
+        return Status::OK();
+    }
+    if (_segcompaction_status.load() != OLAP_SUCCESS) {
+        _is_doing_segcompaction = false;
```
zhannngchen, line 460: Need a log here? print `_segcompaction_status`

be/src/olap/rowset/beta_rowset_writer.h
```diff
@@ -149,9 +157,15 @@ class BetaRowsetWriter : public RowsetWriter {
     std::vector<KeyBoundsPB> _segments_encoded_key_bounds;
     // record rows number of every segment
     std::vector<uint32_t> _segment_num_rows;
+
+    // ensure only one inflight segcompaction task for each rowset
     std::atomic<bool> _is_doing_segcompaction;
+    // enforce compare-and-swap on _is_doing_segcompaction
+    std::mutex _is_doing_segcompaction_lock;
```
zhannngchen, line 164: `_segcompacting_cond_lock` is enought to help to protect `_is_doing_segcompaction`, no need to import another lock?
I don't think we need two locks here, you can keep 1 of them, and the code can works well and simple?

be/src/olap/rowset/beta_rowset_writer.cpp
freemandealer: Sure, the log is needed but I did it in BetaRowsetWriter::segcompaction which is the first crime scene.

zhannngchen: ok，👍

zhannngchen (dismissed):
LGTM

be/src/olap/delta_writer.cpp
```diff
@@ -295,9 +307,13 @@ Status DeltaWriter::close() {
         return Status::OLAPInternalError(OLAP_ERR_ALREADY_CANCELLED);
     }
 
-    RETURN_NOT_OK(_flush_memtable_async());
+    auto s = _flush_memtable_async();
     _mem_table.reset();
-    return Status::OK();
+    if (OLAP_UNLIKELY(!s.ok())) {
```
zhannngchen, line 312: You can simply use `return s;`?

be/src/olap/delta_writer.cpp
```diff
@@ -56,8 +56,6 @@ DeltaWriter::~DeltaWriter() {
         _garbage_collection();
     }
 
-    _mem_table.reset();
```
zhannngchen, line 59: Why remove it?

hello-stephen:
https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20221014193348_clickbench_pr_29387.html

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -102,6 +115,409 @@ Status BetaRowsetWriter::add_block(const vectorized::Block* block) {
     return _add_block(block, &_segment_writer);
 }
 
+vectorized::VMergeIterator* BetaRowsetWriter::_get_segcompaction_reader(
```
yixiutt, line 118: I recommend create a new file such as seg_compaction.cpp to handle segment compaction and logic in beta_rowset_writer could be much lightweight. In segment compaction, call init、do_impl、finish interface so that this code would be much clear to review

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -31,20 +33,31 @@
...
           _total_data_size(0),
-          _total_index_size(0) {}
+          _total_index_size(0),
+          _raw_num_rows_written(0),
+          _is_doing_segcompaction(false) {
+    _segcompaction_status.store(OLAP_SUCCESS);
+}
 
 BetaRowsetWriter::~BetaRowsetWriter() {
+    OLAP_UNUSED_ARG(_wait_flying_segcompaction());
```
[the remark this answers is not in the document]
freemandealer: No. It is added here because of personal paranoid. It can be deleted in further patch.

be/src/olap/rowset/beta_rowset_writer.cpp
freemandealer: Thanks for your suggestion. It seems to need large modification on the patch and re-test. So I'd rather do it in further patches (maybe when add multi-pass segcompaction). If this commit itself is good, I suggest to merge it as it is -- a step by step methodology.

yixiutt (dismissed):

yixiutt:
LGTM

zhannngchen (dismissed):
LGTM

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -236,6 +692,26 @@ RowsetSharedPtr BetaRowsetWriter::build() {
             return nullptr;
         }
     }
+    Status status;
```
dataroaring, line 695: Not related to this pr, we should close file_writer timely, otherwise we consumes a lot fds.

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -236,6 +692,26 @@ RowsetSharedPtr BetaRowsetWriter::build() {
...
+    status = _wait_flying_segcompaction();
+    if (!status.ok()) {
+        LOG(WARNING) << "segcompaction failed when build new rowset, res=" << status;
+        return nullptr;
+    }
+    status = _segcompaction_ramaining_if_necessary();
+    if (!status.ok()) {
+        LOG(WARNING) << "segcompaction failed when build new rowset, res=" << status;
```
dataroaring, line 703: We should not return error due to segcompaction failure?

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -102,6 +115,409 @@ Status BetaRowsetWriter::add_block(const vectorized::Block* block) {
...
+}
+
+void BetaRowsetWriter::segcompaction(SegCompactionCandidatesSharedPtr segments) {
+    Status status = _do_segcompaction(segments);
+    if (!status.ok()) {
+        int16_t errcode = status.precise_code();
+        switch (errcode) {
+        case OLAP_ERR_SEGCOMPACTION_INIT_READER:
+        case OLAP_ERR_SEGCOMPACTION_INIT_WRITER:
+            LOG(WARNING) << "segcompaction failed, try next time:" << status;
```
dataroaring, line 348: We should skip memory exceed limit too.

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -102,6 +115,409 @@ Status BetaRowsetWriter::add_block(const vectorized::Block* block) {
...
+    }
+    size_t s = segments->size();
+    if (!is_terminated_by_big && s <= (config::segcompaction_threshold_segment_num / 2)) {
+        // start with big segments and end with small, better to do it in next
+        // round to compact more at once
+        segments->clear();
+        return Status::OK();
+    }
+    if (s == 1) { // poor bachelor, let it go
+        LOG(INFO) << "only one candidate segment";
```
dataroaring, line 428: INFO -> DEBUG or VLOG

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -102,6 +115,409 @@ Status BetaRowsetWriter::add_block(const vectorized::Block* block) {
...
+        RETURN_NOT_OK(_rename_compacted_segment_plain(_segcompacted_point++));
+        segments->clear();
+        return Status::OK();
+    }
+    std::stringstream ss;
+    for (auto& segment : (*segments.get())) {
+        ss << "[id:" << segment->id() << " num_rows:" << segment->num_rows() << "]";
+    }
+    LOG(INFO) << "candidate segments num:" << s << " list of candidates:" << ss.str();
```
dataroaring, line 437: maybe DEBUG?

dataroaring (changes requested):

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -102,6 +115,409 @@ Status BetaRowsetWriter::add_block(const vectorized::Block* block) {
...
+            LOG(WARNING) << "segcompaction failed, try next time:" << status;
+            return;
+        default:
+            LOG(WARNING) << "segcompaction fatal, terminating the write job:" << status;
+            // status will be checked by the next trigger of segcompaction or the final wait
+            _segcompaction_status.store(OLAP_ERR_OTHER_ERROR);
+        }
+    }
+    DCHECK_EQ(_is_doing_segcompaction, true);
+    _is_doing_segcompaction = false;
```
dataroaring, line 357: We do not hold mutex here, so the program may run into a case in which there are no thread to signal the waiting threads.

dataroaring (changes requested):
Please look

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -315,12 +823,23 @@ Status BetaRowsetWriter::_create_segment_writer(
...
-                                                _context.max_rows_per_segment, writer_options));
-    {
-        std::lock_guard<SpinLock> l(_lock);
-        _file_writers.push_back(std::move(file_writer));
+
+    if (is_segcompaction) {
+        writer->reset(new segment_v2::SegmentWriter(file_writer.get(), _num_segcompacted,
+                                                    _context.tablet_schema, _context.data_dir,
+                                                    _context.max_rows_per_segment, writer_options));
+        if (_segcompaction_file_writer != nullptr) {
```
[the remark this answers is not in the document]
freemandealer: yes. it is now. I add instant close right after writer finishing writing.

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -332,7 +851,31 @@ Status BetaRowsetWriter::_create_segment_writer(
...
+Status BetaRowsetWriter::_create_segment_writer(
+        std::unique_ptr<segment_v2::SegmentWriter>* writer) {
+    size_t total_segment_num = _num_segment - _segcompacted_point + 1 + _num_segcompacted;
+    if (UNLIKELY(total_segment_num > config::max_segment_num_per_rowset)) {
+        LOG(ERROR) << "too many segments in rowset."
+                   << " max:" << config::max_segment_num_per_rowset
+                   << " _num_segment:" << _num_segment
+                   << " _segcompacted_point:" << _segcompacted_point
+                   << " _num_segcompacted:" << _num_segcompacted;
+        return Status::OLAPInternalError(OLAP_ERR_TOO_MANY_SEGMENTS);
```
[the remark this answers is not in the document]
freemandealer: sure, nice suggestion.

be/src/olap/rowset/beta_rowset_writer.cpp
freemandealer: sure.

be/src/olap/delta_writer.cpp
```diff
@@ -171,12 +171,12 @@ Status DeltaWriter::write(Tuple* tuple) {
     // if memtable is full, push it to the flush executor,
     // and create a new memtable for incoming data
     if (_mem_table->memory_usage() >= config::write_buffer_size) {
-        if (++_segment_counter > config::max_segment_num_per_rowset) {
```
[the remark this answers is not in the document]
freemandealer: yes it is currently.

be/src/olap/rowset/beta_rowset_writer.cpp
freemandealer: Theoretically, it should not. Non-deadly failures are handled as need-retry. Here is the the last-ditch for the remaining failures that is hard to handle by simply retrying. 

Eliminating them is hard but not impossible, as long as more effort had been made. I think we can polish them up in future patches if it causes any real problem.

dataroaring (dismissed):
LGTM

hello-stephen:
 the sum of best hot time: 38.65 seconds
 load time: 560 seconds
 storage size: 17154655347 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20221030170442_clickbench_pr_35874.html

dataroaring (dismissed):
LGTM

dataroaring (dismissed):
LGTM

morningman (approved):
LGTM

PR #12919, opened 2022-09-23
[fix](memory): avoid coredump when list pointer is null

levy5307:
# Proposed changes

Issue Number: [5664](https://github.com/apache/doris/issues/5664)

## Problem summary

- To avoid the coredump memtioned in above issue, adding the validation of pointer before using it.
BTW, It's a best practise to validate a pointer before using it.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ]  I don't know
3. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

If this is a relatively large or complex change, kick off the discussion at [[dev@doris.apache.org](mailto:dev@doris.apache.org)](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...


dataroaring (changes requested):
Could you  add a regression test?

be/src/util/likely_util.h
```diff
@@ -0,0 +1,23 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#pragma once
+
+namespace doris {
+#define likely(pred) (__builtin_expect(!!(pred), 1))
```
Gabriel39, line 21: We already have LIKELY and UNLIKELY in common/compiler_util.h. Maybe you do not need to add a new one

levy5307: OK

levy5307:
> Could you add a regression test?

Sure

levy5307:
> Could you add a regression test?

It's hard to add a regression test because this bug is not a problem bound to arise. So I add a unit test for this function instead.

yiguolei (approved):
LGTM

dataroaring (approved):
LGTM

PR #12936, opened 2022-09-24
[enhancement](string column)copy string continuous

yiguolei:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Some query read string columns continuously, but the string will be copied to ColumnString one by one from page. There are two improvement in this PR:
1. Resize the chars using the avg size of 100 string rows.
2. Calculate the max continuous rows and copy using a single memcpy command.

I also add a padding size in page, so that we could using overflow memcpy in the future.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
4. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_v2/column_reader.cpp
+++ b/be/src/olap/rowset/segment_v2/column_reader.cpp
@@ -715,6 +715,8 @@ Status FileColumnIterator::next_batch(size_t* n, vectorized::MutableColumnPtr& d
     size_t curr_size = dst->byte_size();
     size_t remaining = *n;
     *has_null = false;
+    bool is_string_column = dst->is_column_string();
+    bool has_reserved = false;
     while (remaining > 0) {
         if (!_page.has_remaining()) {
             bool eos = false;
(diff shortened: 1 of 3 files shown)


yiguolei:
In lastes clang-14, our memcpy_overflow will be replaced by memcpy...

void my_memcpy(char * __restrict dst, const char * __restrict src, ssize_t n)
{
    while (n > 0)
    {
        _mm_storeu_si128(reinterpret_cast<__m128i *>(dst),
            _mm_loadu_si128(reinterpret_cast<const __m128i *>(src)));
        dst += 16;
        src += 16;
        n -= 16;
    }
}

You can find detail in this discussion: https://github.com/llvm/llvm-project/issues/56467

be/src/olap/rowset/segment_v2/page_io.cpp
```diff
@@ -34,7 +34,7 @@ namespace doris {
 namespace segment_v2 {
 
 using strings::Substitute;
-
+enum { PAGE_PADDING_SIZE = 16 };
```
HappenLee, line 37: why not use static constexpr ？

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -760,6 +762,27 @@ Status FileColumnIterator::next_batch(size_t* n, vectorized::MutableColumnPtr& d
             _current_ordinal += nrows_to_read;
         }
         remaining -= nrows_in_page;
+        // If the column is string, there are many cost during resize chars array, so that do such optimization:
```
HappenLee, line 765: maybe we should impl the function in column string as a interface ？

be/src/olap/rowset/segment_v2/page_io.cpp
yiguolei: done

Issue #3861, opened 2020-06-15
Incorrect result of CAST AS CHAR


sduzh:
**Describe the bug**
Incorrect result of CAST ... AS CHAR

**To Reproduce**
```SQL
SELECT cast('20190101' AS CHAR(2));
```

**Expected behavior**
```SQL
+-----------------------------+
| cast('20190101' as CHAR(2)) |
+-----------------------------+
| 20                          |
+-----------------------------+
```

**Real behavior**
```SQL
+-------------------------------+
| CAST('20190101' AS CHAR(2)) |
+-------------------------------+
| 20190101                      |
+-------------------------------+
```

PR #12957, opened 2022-09-26
[fix](planner) : Incorrect result of CAST AS CHAR

lsy3993:
# Proposed changes

Issue Number: #3861

## Problem summary

when cast string to char(int), fe should judge the length of target char, rather than return whole string

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments


BiteTheDDDDt:
We are not doing char type length processing in the execution engine, it seems that this will make us not have consistent performance globally

lsy3993:
This bug fix is only for the case of casting constant string in FE. When the cast statement calculates in BE, we may fix it later

zy-kkk (dismissed):
LGTM

PR #13133, opened 2022-10-06
[chore] Format keywords in sql_parser.cup

gavinchou:
The key keyword definition section of `sql_parser.cup` is unordered and messy:
1. It is almost unreadable
2. There are no rules to format it when we make a change to it
3. **It takes unnecessary effort to resolve conflict caused by the unordered keywords**

We can apply some simple rules to format it:
1. Sort in lexicographical order
4. Break into several "sections", keywords in each section have the same prefix `KW_${first_letter}`
5. Every 2 sections are connected with an empty line containing only 4 white spaces

e.g.

```
terminal String
    KW_A...

    KW_B...

    ...

    KW_Z...
```


morningman:
If you do so, I suggest to make one keyword in one line. That could be more clear and easier to resolve conflict.

gavinchou:
> If you do so, I suggest to make one keyword in one line. That could be more clear and easier to resolve conflict.

@morningman Done, it's true that one-keyword-per-line is much easier to resolve conflict.

morningman (approved):
LGTM

PR #13167, opened 2022-10-08
[fix](string) Fix over-allocated memory for string type

gavinchou:
For string/varchar/text type, the length field is fixed to 2GB. (`ColumnMetaPB`)
We don't actually have to allocate 2GB for every string type because we
will reallocate the precise size of memory for the string in
`WrapperField::from_string()`

```
    Status from_string(const std::string& value_string, const int precision = 0,
                       const int scale = 0) {
        if (_is_string_type) {
            if (value_string.size() > _var_length) {
                Slice* slice = reinterpret_cast<Slice*>(cell_ptr());
                slice->size = value_string.size();
                _var_length = slice->size;
                _string_content.reset(new char[slice->size]);
                slice->data = _string_content.get();
            }
        }
        return _rep->from_string(_field_buf + 1, value_string, precision, scale);
    }
```

### DIFF
--- a/be/src/olap/wrapper_field.cpp
+++ b/be/src/olap/wrapper_field.cpp
@@ -84,7 +84,7 @@ WrapperField::WrapperField(Field* rep, size_t variable_len, bool is_string_type)
     char* buf = _field_buf + 1;
 
     if (_is_string_type) {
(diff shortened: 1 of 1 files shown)


dataroaring (approved):
LGTM

PR #13219, opened 2022-10-09
[compaction](http_action) enable be run manual compaction concurrently

yixiutt:
In some case, we need to run manual compaction via http interface concurrently, we remove the mutex and tablet's compaction lock is enough to prevent concurrent compaction in tablet.

# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/http/action/compaction_action.cpp
+++ b/be/src/http/action/compaction_action.cpp
@@ -38,9 +38,6 @@ namespace doris {
 
 const static std::string HEADER_JSON = "application/json";
 
-bool CompactionAction::_is_compaction_running = false;
-std::mutex CompactionAction::_compaction_running_mutex;
-
 Status CompactionAction::_check_param(HttpRequest* req, uint64_t* tablet_id) {
     std::string req_tablet_id = req->param(TABLET_ID_KEY);
     if (req_tablet_id == "") {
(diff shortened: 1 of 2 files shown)


dataroaring (approved):
LGTM

be/src/http/action/compaction_action.cpp
```diff
@@ -94,18 +91,7 @@ Status CompactionAction::_handle_run_compaction(HttpRequest* req, std::string* j
...
-        std::lock_guard<std::mutex> lock(_compaction_running_mutex);
-        if (_is_compaction_running) {
-            return Status::TooManyTasks("Manual compaction task is running");
-        } else {
-            // 3.2 execute the compaction task and set compaction task running
-            _is_compaction_running = true;
-            std::thread(std::move(task)).detach();
-        }
-    }
+    std::thread(std::move(task)).detach();
```
gavinchou, line 94: Should we set a limit to it? e.g. count the number of running compaction, and limit with it.
```
    std::packaged_task<Status()> task([this, tablet, compaction_type]() {
        ++static_cnt;
        std::shared_ptr<void> defer_cnt(nullptr, [](...) { --static_cnt; });
        return _execute_compaction_callback(tablet, compaction_type);
    });
```

yixiutt: Parallel run case may fail if we add a limit here。 And this interface should not expose to users but we use it internal, so, it's controllable。

zhannngchen:
use `_push_tablet_into_submitted_compaction` and `_pop_tablet_into_submitted_compaction` instead? Is there any problem if we allow many compaction task of same type for same tablet running at the same time? Note that `_cumulative_compaction_lock` is hold separately during prepare phase and execute phase

PR #13220, opened 2022-10-09
[Bug](replace function) fix be infinite loop and oom when use replace with an empty old str

ByteYue:
# Proposed changes

Issue Number: close #xxx

## Problem summary

The former logic in replace function didn't check whether the input old str is empty or not. The `string.find` function of cpp stl would return pos 0, which would result in an endless string replace logic loop, and end up with OOM.

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [x] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/vec/functions/function_string.h
+++ b/be/src/vec/functions/function_string.h
@@ -1470,6 +1470,9 @@ public:
 
 private:
     std::string replace(std::string str, std::string_view old_str, std::string_view new_str) {
+        if (old_str.empty()) {
+            return str;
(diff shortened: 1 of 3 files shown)


dataroaring (approved):
LGTM

PR #13224, opened 2022-10-09
[fix](rowset) fix that rowset writer doesn't process the return value which may result in data loss

liaoxin01:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/olap/rowset/beta_rowset_writer.cpp
+++ b/be/src/olap/rowset/beta_rowset_writer.cpp
@@ -173,7 +173,11 @@ OLAPStatus BetaRowsetWriter::flush_single_memtable(MemTable* memtable, int64_t* 
 RowsetSharedPtr BetaRowsetWriter::build() {
     // TODO(lingbin): move to more better place, or in a CreateBlockBatch?
     for (auto& wblock : _wblocks) {
-        wblock->close();
+        Status status = wblock->close();
+        if (!status.ok()) {
+            LOG(WARNING) << "failed to close wblock, res=" << status;
+            return nullptr;
+        }
     }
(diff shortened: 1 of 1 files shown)


dataroaring (dismissed):
LGTM

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -173,7 +173,11 @@ OLAPStatus BetaRowsetWriter::flush_single_memtable(MemTable* memtable, int64_t*
 RowsetSharedPtr BetaRowsetWriter::build() {
     // TODO(lingbin): move to more better place, or in a CreateBlockBatch?
     for (auto& wblock : _wblocks) {
```
morningman, line 175: I think we should fix it in a more complete way, to return Status instead of nullptr, which is very error-prone.
And use `RowsetSharedPtr` as an out parameter.

liaoxin01: Now the caller handle the error according to whether the pointer is nullptr. So it‘s ok  to return nullptr, although it is not a complete way. There will be a lot of changes if return Status.

zhannngchen (dismissed):
LGTM

yiguolei (approved):
LGTM

PR #13273, opened 2022-10-11
[improvement](schema change) avoid using column ptr swap

jacktengg:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments


be/src/olap/schema_change.cpp
```diff
@@ -1290,15 +1289,15 @@ Status VSchemaChangeDirectly::_inner_process(RowsetReaderSharedPtr rowset_reader
             std::make_unique<vectorized::Block>(new_tablet->tablet_schema()->create_block());
```
yiguolei, line 1289: could move new block code to while ?  and using do {} while (true)

be/src/olap/schema_change.cpp
```diff
@@ -834,8 +834,34 @@ Status RowBlockChanger::change_block(vectorized::Block* ref_block,
...
+        if (ref_col_nullable != new_col_nullable) {
+            // not nullable to nullable
+            if (new_col_nullable) {
+                auto* new_nullable_col = assert_cast<vectorized::ColumnNullable*>(
+                        std::move(*new_col.column).mutate().get());
+
+                new_nullable_col->swap_nested_column(ref_col.column);
+                new_nullable_col->get_null_map_data().resize_fill(new_nullable_col->size());
+            } else {
+                // nullable to not nullable:
```
BiteTheDDDDt, line 852: It is best to add a logic to check that the nullmap does not contain null. If there is null data, the schema change will fail to prevent accidental data loss.

BiteTheDDDDt (dismissed):
LGTM

yiguolei (approved):
LGTM

PR #13336, opened 2022-10-12
[fix](jsonreader) teach jsonreader to release memory

dataroaring:
Allocator of rapidjson does not release memory, this fix use allocator with local buffer and call Clear to release memory allocated beyond local buffer.

# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/exec/json_scanner.cpp
+++ b/be/src/exec/json_scanner.cpp
@@ -346,7 +346,7 @@ Status JsonReader::_parse_json_doc(size_t* size, bool* eof) {
         return Status::OK();
     }
 
-    // clear memory here. 
+    // clear memory here.
     _origin_json_doc.GetAllocator().Clear();
     bool has_parse_error = false;
     // parse jsondata to JsonDoc
--- a/be/src/exec/json_scanner.cpp
+++ b/be/src/exec/json_scanner.cpp
@@ -247,6 +247,9 @@ JsonReader::JsonReader(RuntimeState* state, ScannerCounter* counter, RuntimeProf
           _strip_outer_array(strip_outer_array),
           _num_as_string(num_as_string),
           _fuzzy_parse(fuzzy_parse),
+          _value_allocator(_value_buffer, sizeof(_value_buffer)),
+          _parse_allocator(_parse_buffer, sizeof(_parse_buffer)),
+          _origin_json_doc(&_value_allocator, sizeof(_parse_buffer), &_parse_allocator),
           _json_doc(nullptr),
           _scanner_eof(scanner_eof) {
     _bytes_read_counter = ADD_COUNTER(_profile, "BytesRead", TUnit::BYTES);
@@ -343,6 +346,8 @@ Status JsonReader::_parse_json_doc(size_t* size, bool* eof) {
         return Status::OK();
     }
 
+    // clear memory here. 
+    _origin_json_doc.GetAllocator().Clear();
     bool has_parse_error = false;
     // parse jsondata to JsonDoc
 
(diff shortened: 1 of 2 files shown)


be/src/exec/json_scanner.h
```diff
@@ -178,7 +178,14 @@ class JsonReader {
     std::vector<std::vector<JsonPath>> _parsed_jsonpaths;
     std::vector<JsonPath> _parsed_json_root;
 
-    rapidjson::Document _origin_json_doc; // origin json document object from parsed json string
+    char _value_buffer[4 * 1024 * 1024];
```
gavinchou, line 181: Does it seem that this class may/should/must not be declared as a stack variable?

dataroaring: It is not used on stack except ut.

be/src/exec/json_scanner.h
```diff
@@ -178,7 +178,15 @@ class JsonReader {
...
+    char _parse_buffer[512 * 1024];
```
levy5307, line 182: It's better to define a constant for 512*1024

be/src/exec/json_scanner.h
```diff
@@ -178,7 +178,15 @@ class JsonReader {
     std::vector<std::vector<JsonPath>> _parsed_jsonpaths;
     std::vector<JsonPath> _parsed_json_root;
 
-    rapidjson::Document _origin_json_doc; // origin json document object from parsed json string
+    char _value_buffer[4 * 1024 * 1024];
```
eldenmoon, line 181: what if json data size exceeds the buffer size?

xinyiZzz (approved):
LGTM

be/src/exec/json_scanner.cpp
```diff
@@ -343,6 +346,8 @@ Status JsonReader::_parse_json_doc(size_t* size, bool* eof) {
         return Status::OK();
     }
 
+    // clear memory here.
+    _origin_json_doc.GetAllocator().Clear();
```
zhannngchen, line 350: Why not just create a new `Document` each time we need to parse? It's not reasonable to make _origin_json_doc as a class member variable?

be/src/exec/json_scanner.h
```diff
@@ -178,7 +178,15 @@ class JsonReader {
...
+    char _parse_buffer[512 * 1024];
+
+    typedef rapidjson::GenericDocument<rapidjson::UTF8<>, rapidjson::MemoryPoolAllocator<>,
+                                       rapidjson::MemoryPoolAllocator<>>
+            Document;
+    rapidjson::MemoryPoolAllocator<> _value_allocator;
+    rapidjson::MemoryPoolAllocator<> _parse_allocator;
```
zhannngchen, line 188: Dose `crtAllocator` a better choice?

be/src/exec/json_scanner.h
levy5307: > The constant is only used here and once, so it can be directly used?

I don't think so. Because we can hardly get the meaning  `512*1024`, and we can get it from the name of a constant

be/src/exec/json_scanner.cpp
dataroaring: It should work either. Just a pointer variable as a member.

be/src/exec/json_scanner.h
dataroaring: crlAllocator uses malloc and free，it is less efficient.

be/src/exec/json_scanner.h
dataroaring: So, if we define a varible named kParserBufferSize, can you get some meaning? I think it can be got from _parse_buffer?

xinyiZzz (approved):
LGTM

PR #13428, opened 2022-10-17
[feature-wip](array) remove array config and check array nested depth

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem summary

1. remove FE config `enable_array_type`
2. limit the nested depth of array in FE side.
3. Fix bug that when loading array from parquet, the decimal type is treated as bigint
4. Fix loading array from csv(vec-engine), handle null and "null"
5. Change the csv array loading behavior, if the array string format is invalid in csv, it will be converted to null. 
6. Remove `check_array_format()`, because it's logic is wrong and meaningless
7. Add stream load csv test cases and more parquet broker load tests

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
8. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
9. Has document been added or modified:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
10. Does it need to update dependencies:
    - [ ] Yes
    - [ ] No
11. Are there any changes that cannot be rolled back:
    - [ ] No

## Further comments


be/src/vec/data_types/data_type_array.cpp
```diff
@@ -180,10 +181,15 @@ Status DataTypeArray::from_string(ReadBuffer& rb, IColumn* column) const {
     auto& offsets = array_column->get_offsets();
 
     IColumn& nested_column = array_column->get_data();
+    DCHECK(nested_column.is_nullable());
     if (*rb.position() != '[') {
         return Status::InvalidArgument("Array does not start with '[' character, found '{}'",
                                        *rb.position());
     }
+    if (*(rb.end() - 1) != ']') {
```
cambyzju, line 189: will overflow if rb is empty

be/src/vec/functions/function_cast.h
```diff
@@ -332,9 +334,16 @@ struct ConvertImplGenericFromString {
...
-                RETURN_IF_ERROR(data_type_to->from_string(read_buffer, col_to));
+                Status st = data_type_to->from_string(read_buffer, col_to);
+                // if parsing failed, will return null
+                (*vec_null_map_to)[i] = !st.ok();
+                if (!st.ok()) {
+                    col_to->insert_default();
+                }
             }
-            block.replace_by_position(result, std::move(col_to));
+            // block.replace_by_position(result, std::move(col_to));
```
xy720, line 344: remove this

fe/fe-core/src/main/java/org/apache/doris/catalog/Type.java
```diff
@@ -46,12 +46,8 @@
...
  */
 public abstract class Type {
-    // Maximum nesting depth of a type. This limit was determined experimentally by
-    // org.apache.doris.rewrite.FoldConstantsRule.apply generating and scanning
-    // deeply nested Parquet and Avro files. In those experiments, we exceeded
-    // the stack space in the scanner (which uses recursion for dealing with
-    // nested types) at a nesting depth between 200 and 300 (200 worked, 300 crashed).
-    public static int MAX_NESTING_DEPTH = 2;
+    // Currently only support Array type with max 9 depths.
+    public static int MAX_NESTING_DEPTH = 9;
```
xy720, line 50: Why is 9?

be/src/vec/exec/vbroker_scanner.cpp
```diff
@@ -91,9 +91,10 @@ Status VBrokerScanner::_fill_dest_columns(const Slice& line,
         return Status::OK();
     }
 
-    if (!check_array_format(_split_values)) {
-        return Status::OK();
-    }
+    // This check is meaningless, should be removed
```
xy720, line 94: If we remove this, may be we should also covert the invalid array value in this row to null. Keep the same logic as cast function.

fe/fe-core/src/main/java/org/apache/doris/catalog/Type.java
morningman: there is a hard limit on BE side. See `TypeInfo* get_array_type_info` in `be/src/olap/types.cpp`:
```
DCHECK(iterations <= 8) << "the depth of nested array type should not be larger than 8";
```

be/src/vec/functions/function_cast.h
morningman: done

be/src/vec/data_types/data_type_array.cpp
```diff
@@ -210,13 +217,9 @@ Status DataTypeArray::from_string(ReadBuffer& rb, IColumn* column) const {
...
-            if (nested_column.is_nullable()) {
-                auto& nested_null_col = reinterpret_cast<ColumnNullable&>(nested_column);
-                nested_null_col.get_nested_column().insert_default();
-                nested_null_col.get_null_map_data().push_back(0);
-            } else {
-                nested_column.insert_default();
-            }
+            auto& nested_null_col = reinterpret_cast<ColumnNullable&>(nested_column);
+            nested_null_col.get_nested_column().insert_default();
+            nested_null_col.get_null_map_data().push_back(0);
```
adonis0147, line 222: We should check whether the nested column is nullable here. In future, we may enable the `not_null` syntax to create an array with the elements which are all not nullable.

adonis0147 (approved):
LGTM

be/src/vec/data_types/data_type_array.cpp
adonis0147: Ok. [data_type_array.cpp:185](https://github.com/apache/doris/blob/4c6cd60a020e334f4a361fecc7b9101efde83845/be/src/vec/data_types/data_type_array.cpp#L185) has `DCHECK`.

morningman: Currently Doris only support "nullable" element inside array.
I will add a DCHECK here.
After we support "not_null" element, I will rethink this logic

PR #13795, opened 2022-10-30
[enhancement](memtracker) Refactor load channel + memtable mem tracker

xinyiZzz:
# Proposed changes

Issue Number: close #xxx

## Problem summary

In the load test, the problem is analyzed according to the mem tracker, and it is found that:
1. In the memtable insert stage, the memory recorded by the mem hook is less than that recorded manually, and the value recorded manually is closer to the real value.
2. The memory value recorded by the mem hook is relatively accurate during the memtable flushing stage.
3. The current mem tracker recording method has a high understanding threshold and is not easy to analyze.

For 3.
- When the memory value of the mem table is greater than `config::write_buffer_size (default 200M), flush it, which depends on manually recording the memory;
- `LoadChannelMgr` checks the total memory of all load channels before each add batch. If it exceeds the limit, select the load channel to flush, which relies on mem hook to record memory;
It is not easy to understand that two different memory recording methods are mixed;

So, I simplified the use of mem tracker, looking forward to better understanding.

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/runtime/load_channel_mgr.h
+++ b/be/src/runtime/load_channel_mgr.h
@@ -28,6 +28,7 @@
 #include "gen_cpp/Types_types.h"
 #include "gen_cpp/internal_service.pb.h"
 #include "gutil/ref_counted.h"
+#include "gutil/walltime.h"
 #include "olap/lru_cache.h"
 #include "runtime/load_channel.h"
 #include "runtime/tablets_channel.h"
@@ -58,6 +59,14 @@ public:
     // cancel all tablet stream for 'load_id' load
     Status cancel(const PTabletWriterCancelRequest& request);
 
+    void refresh_mem_tracker() {
+        int64_t mem_usage = 0;
+        for (auto& kv : _load_channels) {
+            mem_usage += kv.second->mem_consumption();
+        }
+        _mem_tracker->set_consumption(mem_usage);
+    }
+
 private:
     template <typename Request>
     Status _get_load_channel(std::shared_ptr<LoadChannel>& channel, bool& is_eof,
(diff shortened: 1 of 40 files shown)


hello-stephen:
 the sum of best hot time: 38.58 seconds
 load time: 605 seconds
 storage size: 17182689332 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20221031021453_clickbench_pr_35957.html

be/src/runtime/load_channel_mgr.h
```diff
@@ -58,6 +59,14 @@ class LoadChannelMgr {
     // cancel all tablet stream for 'load_id' load
     Status cancel(const PTabletWriterCancelRequest& request);
 
+    void refresh_mem_tracker() {
```
yiguolei, line 62: why do we need refresh memtracker periodly?

be/src/runtime/load_channel_mgr.h
```diff
@@ -80,7 +89,8 @@ class LoadChannelMgr {
     Cache* _last_success_channel = nullptr;
 
     // check the total load channel mem consumption of this Backend
-    std::shared_ptr<MemTrackerLimiter> _mem_tracker;
+    std::unique_ptr<MemTracker> _mem_tracker;
```
yiguolei, line 92: I think we should use MemTrackerLimiter here. Normal memtracker could be used to control flush memtable. But if it could not flush memtable accurately then limiter memtracker will work. If this is a normal memtracker, OOM will occurred.

be/src/runtime/load_channel_mgr.h
xinyiZzz: Because the normal mem tracker used by LoadChannelMgr has no parent-child relationship, the consumption of the MemTable tracker will not be automatically synchronized to the LoadChannelMgr tracker.

be/src/runtime/load_channel_mgr.h
xinyiZzz: Ditto~

yiguolei (approved):
LGTM

PR #13842, opened 2022-10-31
[doc](prefix-index): add the limit rule on the num of key columns in prefix index

spaces-X:
# Proposed changes

fix doc of prefix index

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments


hf200012:
Create table syntax adds a description

docs/zh-CN/docs/data-table/index/prefix-index.md
```diff
@@ -41,6 +41,23 @@ under the License.
...
+需要注意的是为了防止索引膨胀，**在36字节限制和varchar截断限制的基础上，增加了列数限制（默认：3列）**，默认为**前三个KEY列**参与索引的构建，如果用户期望将更多的key列纳入到前缀索引中，需要在建表时修改PROPERTIES如下：
+
+```sql
+CREATE TABLE xxx
+(
+    ...
+)
+...
+PROPERTIES(
+    "shortkey_max_column_count" = "n" -- 其中 n 为用户设定的参与前缀索引构建的列数
```
yiguolei, line 53: why do we need this limit? I think if user could set column count here, he could just set it to 100 and then he or she could use as many as columns in his alter sql.

spaces-X: From the last discussion in the wechat group, the conclusion is that the default value of this parameter is to prevent the short index from being too large if so the index filtering does not work well. But the default value  is goot at all scenarios, so we make this parameter configurable. 

The key point is that there is a difference between the description in the document and the implementation in the engine, which will cause confusion to users

PR #13858, opened 2022-11-01
[feature](new-scan) support transactional insert in new scan framework

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem summary

1.
Support running transactional insert operation with new scan framework. eg:
```
admin set frontend config("enable_new_load_scan_node" = "true");
begin;
insert into tbl1 values(1,2);
insert into tbl1 values(3,4);
insert into tbl1 values(5,6);
commit;
```

2. Add some limitation to transactional insert
    1. Do not support non-literal value in insert stmt

3. Fix some issue about array type:
    1. Forbid cast other non-array type to NESTED array type, it may cause BE crash.
    2. Add `getStringValueForArray()` method for Expr, to get valid string-formatted array type value. 

4. Add `useLocalSessionState=true` in regression-test jdbc url
    without this config, the jdbc driver will send some init cmd each time it connect to server, such as
    `select @@session.tx_read_only`.
    But when we use transactional insert, after `begin` command, Doris do not support any other type of
    stmt except for `insert`, `commit` or `rollback`.
    So adding this config to let the jdbc NOT send cmd when connecting.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
3. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments


fe/fe-core/src/main/java/org/apache/doris/analysis/NullLiteral.java
```diff
@@ -101,6 +101,11 @@ public String getStringValue() {
         return "NULL";
     }
 
+    @Override
+    public String getStringValueForArray() {
```
pengxiangyu, line 105: null?  not return ""？

hello-stephen:
 the sum of best hot time: 38.08 seconds
 load time: 573 seconds
 storage size: 17154821321 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20221101091724_clickbench_pr_37260.html

fe/fe-core/src/main/java/org/apache/doris/analysis/NullLiteral.java
```diff
@@ -101,6 +101,11 @@ public String getStringValue() {
...
+        return "null";
```
morningman, line 106: the null value inside an array is represented as "null", for exampe:
`[null, null]`.
Not same as other primitive type to represent as `\N`.
I will add some comment to this method.

yiguolei (approved):
LGTM

PR #13865, opened 2022-11-01
[feature](remote)Only query can use local cache when reading remote files.

### LINKED ISSUE #13864: [Feature] Only query can use local cache when reading remote files.
### Search before asking

### Description

Only query can use local cache when reading remote files.

### Use case

1. When calling select on remote files, download cache files to local disk.
2. When calling alter table on remote files, read files directly from remote storage. So if tablet is too large, it will not take up too many local disk when creating local cache file.

### Related issues

1.1.3

### Code of Conduct

pengxiangyu:
# Proposed changes

Issue Number: close #13864

## Problem summary

When calling select on remote files, download cache files to local disk.
When calling alter table on remote files, read files directly from remote storage. So if tablet is too large, it will not take up too many local disk when creating local cache file.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No


hello-stephen:
 the sum of best hot time: 39.89 seconds
 load time: 611 seconds
 storage size: 17154810777 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20221101114536_clickbench_pr_37306.html

be/src/olap/segment_loader.cpp
```diff
@@ -35,45 +36,52 @@ SegmentLoader::SegmentLoader(size_t capacity) {
...
     std::vector<segment_v2::SegmentSharedPtr> segments;
-    RETURN_NOT_OK(rowset->load_segments(&segments));
+    RETURN_NOT_OK(rowset->load_segments(use_local_file_cache, &segments));
 
     if (use_cache) {
         // memory of SegmentLoader::CacheValue will be handled by SegmentLoader
         SegmentLoader::CacheValue* cache_value = new SegmentLoader::CacheValue();
         cache_value->segments = std::move(segments);
-        _insert(cache_key, *cache_value, cache_handle);
+        _insert(key, *cache_value, cache_handle);
```
cambyzju, line 84: here maybe two segments cache key, one cache key is "${rowset_id}", the other cache key is "local_file_cache_${rowset_id}", and both of them point to the same segments?

Is there any problem? for example, open twice or close twice, or some other cases?

pengxiangyu: Segment called is not changed. we only create two instances.

cambyzju (commented):
LGTM

be/src/olap/rowset/beta_rowset.cpp
```diff
@@ -96,7 +96,8 @@ Status BetaRowset::do_load(bool /*use_cache*/) {
     return Status::OK();
 }
 
-Status BetaRowset::load_segments(std::vector<segment_v2::SegmentSharedPtr>* segments) {
+Status BetaRowset::load_segments(bool use_local_file_cache,
```
yiguolei, line 99: not pass in such parameters. There is a read_op_type, you can use it to check use local cache or not.

pengxiangyu: read context是一个read config, not suitable in class Rowset

be/src/olap/reader.h
```diff
@@ -101,6 +101,8 @@ class TabletReader {
         bool read_orderby_key_reverse = false;
         // num of columns for orderby key
         size_t read_orderby_key_num_prefix_columns = 0;
+        // use local cache when reading remote file.
+        bool use_local_file_cache = false;
```
yangzhg, line 105: where is this configuration used

yangzhg (commented):
why not use the config `use_local_file_cache` instead of add "use_local_file_cache"  args everywhere

be/src/olap/reader.h
pengxiangyu: used in beta_rowset_reader.cpp

be/src/io/cache/dummy_file_cache.h
```diff
@@ -36,7 +36,7 @@ class DummyFileCache final : public FileCache {
 
     Status close() override { return Status::OK(); }
 
-    Status read_at(size_t offset, Slice result, size_t* bytes_read) override {
+    Status read_at(size_t offset, Slice result, size_t* bytes_read, const IOContext* ctx) override {
```
yiguolei, line 39: use reference not pointer. 
Pointer is used as output parameter. 
and please add ctx before bytes_read.

be/src/olap/reader.h
yiguolei, line 105: This line is useless.

be/src/olap/rowset/beta_rowset_reader.cpp
```diff
@@ -149,6 +149,7 @@ Status BetaRowsetReader::init(RowsetReaderContext* read_context) {
     read_options.record_rowids = read_context->record_rowids;
     read_options.read_orderby_key_reverse = read_context->read_orderby_key_reverse;
     read_options.read_orderby_key_columns = read_context->read_orderby_key_columns;
+    read_options.io_ctx.use_local_file_cache = read_context->use_local_file_cache;
```
yiguolei, line 152: There is reader type , I think you'd better pass reader type to iO CTX

be/src/olap/rowset/segment_v2/segment.h
```diff
@@ -74,7 +74,8 @@ class Segment : public std::enable_shared_from_this<Segment> {
 
     uint32_t num_rows() const { return _footer.num_rows(); }
 
-    Status new_column_iterator(const TabletColumn& tablet_column, ColumnIterator** iter);
+    Status new_column_iterator(const TabletColumn& tablet_column, const IOContext& io_ctx,
```
yiguolei, line 77: Do not modify this method. could modify column iterator's init method and pass io ctx via init method.

be/src/olap/schema_change.cpp
```diff
@@ -1942,6 +1942,7 @@ Status SchemaChangeHandler::_do_process_alter_tablet_v2(const TAlterTabletReqV2&
             reader_context.is_vec = config::enable_vectorized_alter_table;
             reader_context.delete_bitmap = &base_tablet->tablet_meta()->delete_bitmap();
             reader_context.version = Version(0, end_version);
+            reader_context.use_local_file_cache = false;
```
yiguolei, line 1945: I think reader type is better than local file cache because it contains more info.

be/src/olap/rowset/segment_v2/segment.h
pengxiangyu: init is called just in new_column_iterator()

be/src/olap/rowset/segment_v2/segment.cpp
```diff
@@ -259,9 +263,8 @@ Status Segment::new_column_iterator(const TabletColumn& tablet_column, ColumnIte
                         tablet_column.has_default_value(), tablet_column.default_value(),
                         tablet_column.is_nullable(), std::move(type_info), tablet_column.length(),
                         tablet_column.precision(), tablet_column.frac()));
-        ColumnIteratorOptions iter_opts;
 
-        RETURN_IF_ERROR(default_value_iter->init(iter_opts));
+        RETURN_IF_ERROR(default_value_iter->init(opts));
         *iter = default_value_iter.release();
         return Status::OK();
     }
```
yiguolei, line 270: not only default iterator should call init here, also other column iterators on line 271 should call init here.

yiguolei (dismissed):
LGTM

yiguolei (approved):
LGTM

PR #14100, opened 2022-11-09
[enhancement](plugin) import audit logs for slow queries into a separate table

### LINKED ISSUE #14099: [Enhancement] import audit logs for slow queries into a separate table
### Search before asking

### Description

The current audit log plugin periodically imports FE audit logs to specified Doris cluster for users to view and analyze audit logs using SQL. In most cases, users pay more attention to the slow query of audit logs. Based on the current audit log plugin, this PR imports the audit logs corresponding to slow query into specified table, facilitating users to query and analysis the audit logs through SQL.

### Solution

Set a slow query threshold(using the configuration qe_slow_log_ms in config.java). If the query time is longer than threshold, import the audit logs to the slow table doris_slow_log_tbl__.

### Code of Conduct

yongjinhou:
# Proposed changes

Issue Number: close #14099

## Problem summary

The current audit log plugin periodically imports FE audit logs to specified Doris cluster for users to view and analyze audit logs using SQL. In most cases, users pay more attention to the slow query of audit logs. Based on the current audit log plugin, this PR imports the audit logs corresponding to slow query into specified table, facilitating users to query and analysis the audit logs through SQL.

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [x] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/fe_plugins/auditloader/src/main/java/org/apache/doris/plugin/audit/AuditLoaderPlugin.java
+++ b/fe_plugins/auditloader/src/main/java/org/apache/doris/plugin/audit/AuditLoaderPlugin.java
@@ -24,6 +24,7 @@ import org.apache.doris.plugin.Plugin;
 import org.apache.doris.plugin.PluginContext;
 import org.apache.doris.plugin.PluginException;
 import org.apache.doris.plugin.PluginInfo;
+import org.apache.doris.common.Config;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -56,8 +57,10 @@ public class AuditLoaderPlugin extends Plugin implements AuditPlugin {
     private static final ThreadLocal<SimpleDateFormat> dateFormatContainer = ThreadLocal.withInitial(
             () -> new SimpleDateFormat("yyyy-MM-dd HH:mm:ss"));
 
-    private StringBuilder auditBuffer = new StringBuilder();
-    private long lastLoadTime = 0;
+    private StringBuilder auditLogBuffer = new StringBuilder();
+    private StringBuilder slowLogBuffer = new StringBuilder();
+    private long lastLoadTimeAuditLog = 0;
+    private long lastLoadTimeSlowLog = 0;
 
     private BlockingQueue<AuditEvent> auditEventQueue;
     private DorisStreamLoader streamLoader;
@@ -75,7 +78,8 @@ public class AuditLoaderPlugin extends Plugin implements AuditPlugin {
             if (isInit) {
                 return;
             }
-            this.lastLoadTime = System.currentTimeMillis();
+            this.lastLoadTimeAuditLog = System.currentTimeMillis();
+            this.lastLoadTimeSlowLog = System.currentTimeMillis();
 
             loadConfig(ctx, info.getProperties());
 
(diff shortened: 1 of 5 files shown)


hello-stephen:
 the sum of best hot time: 33.69 seconds
 load time: 435 seconds
 storage size: 17179527557 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20221109025112_clickbench_pr_42373.html

yangzhg (dismissed):
LGTM

morningman:
Why not just query slow queries with `select * from audit_table where time > 5000`?

yangzhg:
> Why not just query slow queries with `select * from audit_table where time > 5000`? That would be more easy to use and no need to maintain to many tables.

Many endpoint users are concerned with slow queries, hoping to analyze such queries in a more convenient way. The audit log table often has a very large amount of data, and the proportion of slow queries in it is very small. Therefore, slow queries are stored as a separate table for It is more reasonable for users with slow query analysis needs

fe_plugins/auditloader/src/main/java/org/apache/doris/plugin/audit/DorisStreamLoader.java
```diff
@@ -123,7 +127,13 @@ public LoadResponse loadBatch(StringBuilder sb) {
         HttpURLConnection beConn = null;
         try {
             // build request and send to fe
-            feConn = getConnection(loadUrlStr, label);
+            if (slowLog) {
+                label = "slow" + label;
```
morningman, line 131:
```suggestion
                label = "slow_" + label;
```

fe_plugins/auditloader/src/main/java/org/apache/doris/plugin/audit/DorisStreamLoader.java
```diff
@@ -123,7 +127,13 @@ public LoadResponse loadBatch(StringBuilder sb) {
...
+                feConn = getConnection(slowLogLoadUrlStr, label);
+            } else {
+                label = "audit" + label;
```
morningman, line 134:
```suggestion
                label = "audit_" + label;
```

fe_plugins/auditloader/src/main/java/org/apache/doris/plugin/audit/AuditLoaderPlugin.java
```diff
@@ -146,28 +150,35 @@ public void exec(AuditEvent event) {
...
-        auditBuffer.append(event.scanRows).append("\t");
-        auditBuffer.append(event.returnRows).append("\t");
-        auditBuffer.append(event.stmtId).append("\t");
-        auditBuffer.append(event.isQuery ? 1 : 0).append("\t");
-        auditBuffer.append(event.feIp).append("\t");
-        auditBuffer.append(event.cpuTimeMs).append("\t");
-        auditBuffer.append(event.sqlHash).append("\t");
-        auditBuffer.append(event.sqlDigest).append("\t");
-        auditBuffer.append(event.peakMemoryBytes).append("\t");
+        if (event.queryTime > Config.qe_slow_log_ms) {
```
morningman, line 153: `qe_slow_log_ms` should be a config of plugin itself.
If you use `Config.qe_slow_log_ms`, than it is a fix value that can not be changed.

yongjinhou: it could be changed by configuring the qe_slow_log_ms in fe.cong, also to be consistent with the slow query threshold in the audit log.

fe_plugins/auditloader/src/main/java/org/apache/doris/plugin/audit/DorisStreamLoader.java
yongjinhou: An underscore is already included in “label".

fe_plugins/auditloader/src/main/java/org/apache/doris/plugin/audit/DorisStreamLoader.java
yongjinhou: An underscore is already included in “label".

dutyu:
> > Why not just query slow queries with `select * from audit_table where time > 5000`? That would be more easy to use and no need to maintain to many tables.
> 
> Many endpoint users are concerned with slow queries, hoping to analyze such queries in a more convenient way. The audit log table often has a very large amount of data, and the proportion of slow queries in it is very small. Therefore, slow queries are stored as a separate table for It is more reasonable for users with slow query analysis needs

Why not just create a view like "create view  slow_sql_view as select ... from audit_table where time > 5000", i think this is a more convenient way.

yiguolei:
I do not think we should create a seperate table for slow query.
"SLOW" has many different definitions and we should provide a single table for all queries and user could analysis it using SQL Query.

yiguolei:
IF  we add a table for slow query, then should we add a table for query with errors for example timeout or analysis error? should we add a table for invalid access because some user concern about security problems?

ChPi:
> > Why not just query slow queries with `select * from audit_table where time > 5000`? That would be more easy to use and no need to maintain to many tables.
> 
> Many endpoint users are concerned with slow queries, hoping to analyze such queries in a more convenient way. The audit log table often has a very large amount of data, and the proportion of slow queries in it is very small. Therefore, slow queries are stored as a separate table for It is more reasonable for users with slow query analysis needs

i think its beteer to set parittion with query_time for audit_table, this also can speed up slow query.

yiguolei:
As discussed, please add a config for this feature and set disabled by default.

dutyu:
> > > Why not just query slow queries with `select * from audit_table where time > 5000`? That would be more easy to use and no need to maintain to many tables.
> > 
> > 
> > Many endpoint users are concerned with slow queries, hoping to analyze such queries in a more convenient way. The audit log table often has a very large amount of data, and the proportion of slow queries in it is very small. Therefore, slow queries are stored as a separate table for It is more reasonable for users with slow query analysis needs
> 
> its beteer to set parittion with query_time for audit_table, this also can speed up slow query. but dynamic partition does not like support this😂.

Maybe we can also create a `ROLLUP` and place `query_time` as the first key column to speed up queries (use `prefix index`).

yangzhg (approved):
LGTM

PR #14132, opened 2022-11-09
[enhancement](array) support read list(Array) type from orc file

cambyzju:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Before this pr, if we try to load ORC file with native list(or array) type data, the be will crash.

Because complex types in ORC file include multi real columns, so we need to filter columns by column names.
Otherwise we could not read all columns we need.

Now arrow release-7.0.0 only support create stripe reader by column index, so we patch it to support create stripe reader by column names.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [x] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [x] Yes
    - [ ] No
    - [x] No

## Further comments


morningman:
Hi @cambyzju , the new orc reader is implemented in `be/src/vec/exec/format/orc/vorc_reader.h`, which already support type.
Do we still need to patch arrow?
In the new implementation, I think we can discard arrow library.

cambyzju:
> Hi @cambyzju , the new orc reader is implemented in `be/src/vec/exec/format/orc/vorc_reader.h`, which already support type. Do we still need to patch arrow? In the new implementation, I think we can discard arrow library.

1. now load orc file with native list(array) type, the be will crash;
2. after new scanner works, we could remove all these codes include lib arrow;

carlvinhust2012:
LGTM

xy720 (approved):
LGTM

PR #14134, opened 2022-11-09
[feature-wip](multi-catalog) Support hive partition cache

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem summary

1. Two Caches
	1. HiveMetaStoreCache: mainly used to cache table partition and file information
	2. ExternalSchemaCache: mainly used to cache the schema information of the table.

		ExternalSchemaCache works for all tables in ExternalCatalog, while HiveMetaStoreCache works only for tables in HMSExternalCatalog.
		
2. ExternalMetaCacheMgr

	Cache management class, responsible for managing the mapping relationship between Catalog and HiveMetaStoreCache and ExternalSchemaCache.
	
3. Cache strategy

	1. ExternalSchemaCache
		
		Both Database and Table in ExternalCatalog are persisted to Doris metadata. However, the schema information of the Table is not persisted, but is obtained directly from the ExternalSchemaCache. If it does not exist in the cache, it will access the HMS to get it and cache it.

		The cache uses Guava's LoadingCache class. The schema of up to 10,000 tables is cached by default, which can be modified through the FE configuration item max_external_schema_cache_num .
  
  		Each FE caches its own required schema independently.

	2. HiveMetaStoreCache

		Three types of information are cached:
		
		1. partitionValuesCache

			Cache table to all information with partition value. This information is mainly used for partition pruning.
			
    		When a table is accessed, Doris will obtain all the partition values ​​(Partition values) information of the table in full, and convert it into a ListPartitionItem object inside Doris and store it in the cache.
    		
    		The ListPartitionItem object can construct a RangeMap for partition clipping. After cropping, a set of partition IDs can be obtained.
    		
    		The default cache is 10W, which can be adjusted by max_hive_partition_cache_num

		2. partitionCache
    
    		Partition cache, the key is <table-partition_values>, and the value is the partition information, including the inputformat, location and other information of the partition. After the partition is cut, we can obtain a set of partitionCache keys through the partition ID, and then obtain the partition.
    		
    		The default cache is 10W, which can be adjusted by max_hive_partition_cache_num
    	
    	3. fileCache
    
    		File cache, the key is the location, and the value is the information of all files under the location. We can use the location of the partition as the key of the fileCache, and finally get all the file information.
    		
    		The default cache is 10W, which can be adjusted by max_external_file_cache_num

4. Cache optimization

	After partition pruning, a set of partitions is obtained. We need to traverse these partitions and get the partition information from the partitionCache in turn. When the cache misses, the HMS will be accessed to obtain partition information. If there are many partitions, a large number of RPCs will be generated. In the actual test, it is found that it takes more than 5 minutes to obtain 1W partition information with a single thread.
	
	Here we use java's parallelStream to acquire partition concurrently, and the time-consuming can be reduced to 1 minute.
	
	At the same time, we set the number of background threads of loadingCache to 10, and this thread pool is shared by all caches.
	
5. Why not get partitions in batch from HMS?

	Because partitionCache is partition-based, not table-based. Batch acquisition can only take the table as the granularity, and each cache entry caches all partition information of a table. In this way, the purpose of "on-demand caching" is achieved, and partition cutting cannot be performed.

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
6. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
7. Has document been added or modified:
    - [x] Yes
    - [ ] No
    - [ ] No Need
8. Does it need to update dependencies:
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/external/HMSExternalTable.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/external/HMSExternalTable.java
@@ -18,15 +18,9 @@
 package org.apache.doris.catalog.external;
 
 import org.apache.doris.catalog.Column;
-import org.apache.doris.catalog.Env;
-import org.apache.doris.catalog.HiveMetaStoreClientHelper;
-import org.apache.doris.common.DdlException;
+import org.apache.doris.catalog.Type;
 import org.apache.doris.common.MetaNotFoundException;
-import org.apache.doris.common.util.Util;
 import org.apache.doris.datasource.HMSExternalCatalog;
-import org.apache.doris.datasource.InitTableLog;
-import org.apache.doris.datasource.PooledHiveMetaStoreClient;
-import org.apache.doris.qe.MasterCatalogExecutor;
 import org.apache.doris.thrift.THiveTable;
 import org.apache.doris.thrift.TTableDescriptor;
 import org.apache.doris.thrift.TTableType;
@@ -34,9 +28,6 @@ import org.apache.doris.thrift.TTableType;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
-import org.apache.hadoop.hive.metastore.api.Partition;
-import org.apache.hadoop.hive.ql.exec.SerializationUtilities;
-import org.apache.hadoop.hive.ql.plan.ExprNodeGenericFuncDesc;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -44,6 +35,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * Hive metastore external table.
@@ -61,6 +53,8 @@ public class HMSExternalTable extends ExternalTable {
     }
 
     private volatile org.apache.hadoop.hive.metastore.api.Table remoteTable = null;
+    private List<Column> partitionColumns;
+
     private DLAType dlaType = DLAType.UNKNOWN;
 
     public enum DLAType {
@@ -106,21 +100,6 @@ public class HMSExternalTable extends ExternalTable {
             }
             objectCreated = true;
         }
-        if (!initialized) {
-            if (!Env.getCurrentEnv().isMaster()) {
-                fullSchema = null;
-                // Forward to master and wait the journal to replay.
-                MasterCatalogExecutor remoteExecutor = new MasterCatalogExecutor();
-                try {
-                    remoteExecutor.forward(catalog.getId(), catalog.getDbNullable(dbName).getId(), id);
-                } catch (Exception e) {
-                    Util.logAndThrowRuntimeException(LOG,
-                            String.format("failed to forward init external table %s operation to master", name), e);
-                }
-                return;
-            }
-            init();
-        }
     }
 
     /**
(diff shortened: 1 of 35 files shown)


hello-stephen:
 the sum of best hot time: 33.75 seconds
 load time: 441 seconds
 storage size: 17179273541 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20221110091544_clickbench_pr_43315.html

fe/fe-core/src/main/java/org/apache/doris/datasource/hive/HiveMetaStoreCache.java
```diff
@@ -0,0 +1,334 @@
...
+// 3. fileCache: cache the files of a location.
+public class HiveMetaStoreCache {
+    private static final Logger LOG = LogManager.getLogger(HiveMetaStoreCache.class);
+
+    private static final int MAX_PARTITION_CACHE_SIZE = 1000 * 1000;
+    private HMSExternalCatalog catalog;
+    private Executor executor;
+
+    // cache from <dbname-tblname> -> <values of partitions>
+    private LoadingCache<PartitionValueCacheKey, ImmutableList<ListPartitionItem>> partitionValuesCache;
```
Jibing-Li, line 65: How do we update the three cache?

fe/fe-core/src/main/java/org/apache/doris/datasource/PooledHiveMetaStoreClient.java
```diff
@@ -62,49 +64,56 @@ public List<String> getAllDatabases() {
...
             return client.client.tableExists(dbName, tblName);
         } catch (Exception e) {
-            throw new RuntimeException(e);
+            throw new HMSClientException("failed to check if table %s in db %s exists", e, tblName, dbName);
+        }
+    }
+
+    public List<String> listPartitionNames(String dbName, String tblName) {
+        try (CachedClient client = getClient()) {
+            return client.client.listPartitionNames(dbName, tblName, MAX_LIST_PARTITION_NUM);
```
Jibing-Li, line 89: Could it cause wrong query result if the partitions in Hive are more than MAX_LIST_PARTITION_NUM?

fe/fe-core/src/main/java/org/apache/doris/catalog/external/HMSExternalTable.java
```diff
@@ -340,12 +352,28 @@ public Map<String, String> getS3Properties() {
...
+    private void initPartitionColumns() {
+        if (partitionColumns != null) {
+            return;
+        }
+        synchronized (this) {
+            if (partitionColumns != null) {
+                return;
+            }
+            List<String> partitionKeys = remoteTable.getPartitionKeys().stream().map(FieldSchema::getName)
+                    .collect(Collectors.toList());
```
AshinGau, line 364: collect as set to avoid traversing `partitionKeys`

morningman: done

fe/fe-core/src/main/java/org/apache/doris/datasource/hive/HiveMetaStoreCache.java
morningman: it will update automatically in `CacheLoader.asyncReloading`

fe/fe-core/src/main/java/org/apache/doris/datasource/PooledHiveMetaStoreClient.java
morningman: I tested and yes, it will only return partitions up to the limit, and the result will be wrong.
I will find a way to handle this later.

AshinGau:
LGTM

yiguolei (approved):
LGTM

PR #13957, opened 2022-11-03
[fix](Nerieds) fix tpch and support trace plan's change event

924060929:
# Proposed changes

This pr fix some bugs for run tpc-h
1. fix the avg(decimal) crash the backend. The fix code in `Avg.getFinalType()` and every child class of `ComputeSinature`
2. fix the ReorderJoin dead loop. The fix code in `ReorderJoin.findInnerJoin()`
3. fix the TimestampArithmetic can not bind the functions in the child. The fix code in `BindFunction.FunctionBinder.visitTimestampArithmetic()`

New feature: support trace the plan's change event, you can `set enable_nereids_trace=true` to open trace log and see some log like this:
```
2022-11-03 21:07:38,391 INFO (mysql-nio-pool-0|208) [Job.printTraceLog():128] ========== RewriteBottomUpJob ANALYZE_FILTER_SUBQUERY ==========
before:
LogicalProject ( projects=[S_ACCTBAL#17, S_NAME#13, N_NAME#4, P_PARTKEY#19, P_MFGR#21, S_ADDRESS#14, S_PHONE#16, S_COMMENT#18] )
+--LogicalFilter ( predicates=((((((((P_PARTKEY#19 = PS_PARTKEY#7) AND (S_SUPPKEY#12 = PS_SUPPKEY#8)) AND (P_SIZE#24 = 15)) AND (P_TYPE#23 like '%BRASS')) AND (S_NATIONKEY#15 = N_NATIONKEY#3)) AND (N_REGIONKEY#5 = R_REGIONKEY#0)) AND (R_NAME#1 = 'EUROPE')) AND (PS_SUPPLYCOST#10 =  (SCALARSUBQUERY) (QueryPlan: LogicalAggregate ( phase=LOCAL, outputExpr=[min(PS_SUPPLYCOST#31) AS `min(PS_SUPPLYCOST)`#33], groupByExpr=[] )), (CorrelatedSlots: [P_PARTKEY#19, S_SUPPKEY#12, S_NATIONKEY#15, N_NATIONKEY#3, N_REGIONKEY#5, R_REGIONKEY#0, R_NAME#1]))) )
   +--LogicalJoin ( type=CROSS_JOIN, hashJoinCondition=[], otherJoinCondition=[] )
      |--LogicalJoin ( type=CROSS_JOIN, hashJoinCondition=[], otherJoinCondition=[] )
      |  |--LogicalJoin ( type=CROSS_JOIN, hashJoinCondition=[], otherJoinCondition=[] )
[...]

PR #14169, opened 2022-11-10
[Enhancement](Nerieds) Support numbers TableValuedFunction and some bitmap/hll aggregate function

924060929:
# Proposed changes

## Problem summary
This pr support
1. `numbers` TableValuedFunction for nereids test, like `select * from numbers(number = 10, backend_num = 1)`
2. bitmap/hll aggregate function
3. support find variable length function in function registry, like `coalesce`
4. fix a bug that print nerieds trace will throw exception because use RewriteRule in ApplyRuleJob, e.g: `AggregateDisassemble`, introduced by #13957

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
5. Has unit tests been added:
    - [x] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No


hello-stephen:
 the sum of best hot time: 33.94 seconds
 load time: 438 seconds
 storage size: 17179844008 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20221110132839_clickbench_pr_43562.html

fe/fe-core/src/main/java/org/apache/doris/nereids/trees/expressions/functions/agg/BitmapUnionInt.java
```diff
@@ -0,0 +1,79 @@
...
+        return new BitmapUnionInt(getAggregateParam(), children.get(0));
+    }
+
+    @Override
+    public List<AbstractDataType> expectedInputTypes() {
+        if (isGlobal() && inputTypesBeforeDissemble().isPresent()) {
+            return ImmutableList.of();
+        } else {
+            return ImmutableList.of(new TypeCollection(
+                    TinyIntType.INSTANCE, SmallIntType.INSTANCE, IntegerType.INSTANCE, BigIntType.INSTANCE));
```
wangshuo128, line 61: `BigIntType.INSTANCE` is not valid for `bitmap_union_int`.

morrySnow (dismissed):

fe/fe-core/src/main/java/org/apache/doris/nereids/trees/expressions/functions/FunctionBuilder.java
```diff
@@ -35,12 +36,43 @@
...
+    public boolean canApply(List<? extends Object> arguments) {
+        if (isVariableLength && arity > arguments.size() + 1) {
+            return false;
+        }
+        if (!isVariableLength && arguments.size() != arity) {
+            return false;
+        }
+        for (int i = 0; i < arguments.size(); i++) {
+            Class constructorArgumentType = getConstructorArgumentType(i);
+            if (!constructorArgumentType.isInstance(arguments.get(i))) {
```
wangshuo128, line 60: We should check if the class is precisely the same or only an instance of (derived class ok) ?

fe/fe-core/src/main/java/org/apache/doris/nereids/trees/expressions/functions/agg/BitmapUnionInt.java
924060929: It can, see `aggregate_function_bitmap.cpp`

```cpp
template <bool nullable, template <bool, typename> class AggregateFunctionTemplate>
static IAggregateFunction* createWithIntDataType(const DataTypes& argument_type) {
    auto type = argument_type[0].get();
    if (type->is_nullable()) {
        type = assert_cast<const DataTypeNullable*>(type)->get_nested_type().get();
    }
    WhichDataType which(type);
    if (which.idx == TypeIndex::Int8) {
        return new AggregateFunctionTemplate<nullable, ColumnVector<Int8>>(argument_type);
    }
    if (which.idx == TypeIndex::Int16) {
        return new AggregateFunctionTemplate<nullable, ColumnVector<Int16>>(argument_type);
    }
    if (which.idx == TypeIndex::Int32) {
        return new AggregateFunctionTemplate<nullable, ColumnVector<Int32>>(argument_type);
    }
    if (which.idx == TypeIndex::Int64) {
        return new AggregateFunctionTemplate<nullable, ColumnVector<Int64>>(argument_type);
    }
    return nullptr;
}

AggregateFunctionPtr create_aggregate_function_bitmap_union_int(const std::string& name,
                                                                const DataTypes& argument_types,
                                                                const Array& parameters,
                                                                const bool result_is_nullable) {
    const bool arg_is_nullable = argument_types[0]->is_nullable();
    if (arg_is_nullable) {
        return std::shared_ptr<IAggregateFunction>(
                createWithIntDataType<true, AggregateFunctionBitmapCount>(argument_types));
    } else {
        return std::shared_ptr<IAggregateFunction>(
                createWithIntDataType<false, AggregateFunctionBitmapCount>(argument_types));
    }
}
```

wangshuo128: The FE and BE are different for the type?
https://github.com/apache/doris/blob/8e17fcef3ff4e7322f8b612e7896877aea604c24/fe/fe-core/src/main/java/org/apache/doris/analysis/FunctionCallExpr.java#L652-L654

fe/fe-core/src/main/java/org/apache/doris/nereids/trees/expressions/functions/FunctionBuilder.java
924060929: This code say it can apply if the argument type is the same as signature type or is the child class of the signature type.
For example, `Substring(StringLiteral, IntegerLiteral, IntegerLiteral)` can apply to `Substring(Expression, Expression, Expression)`

fe/fe-core/src/main/java/org/apache/doris/nereids/trees/expressions/functions/agg/BitmapUnionInt.java
924060929: For my test, `numbers(number=10)` return bigint values, and backend can execute
`select bitmap_union_int(number) from numbers(number=10)`.
I think we should support bigint for `bitmap_union_int`

wangshuo128 (dismissed):

morrySnow (approved):

PR #14382, opened 2022-11-17
[improvement](ctas) use string type for varchar/char/string

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem summary

When executing `create table as select` stmt,
the varchar/char/string type of column in created table will be unified to `string` type.

Because when select from external table (mysql/pg, etc), the length of varchar in external database
is calculated by "char" length, not "byte" length.
So if there is a column with `varchar(10)` in external table, then there will be a same `varchar(10)`
in created table. But the byte length of data in external table may be larger than 10, causing failure of CTAS.

Change to string will not impact performance of the capacity of disk storage.
And notice that if a string type column is the first column, it will be changed to `varchar(65535)`,
because we do not allow string type column as sort key column.

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [x] Yes
    - [ ] No
    - [ ] No Need
    - [x] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments


hello-stephen:
 the sum of best hot time: 35.27 seconds
 load time: 432 seconds
 storage size: 17149899748 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20221117174148_clickbench_pr_48254.html

yiguolei (approved):
LGTM

Issue #14448, opened 2022-11-21
[Enhancement] optimize segment compaction log


freemandealer:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

### Description

tablet_id & rowset_id are missing in current segment compaction log and -238 log.

### Solution

provide tablet_id & rowset_id 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #14449, opened 2022-11-21
[Enhancement](storage) optimize segment compaction log (#14448)

freemandealer:

# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/olap/rowset/beta_rowset_writer.cpp
+++ b/be/src/olap/rowset/beta_rowset_writer.cpp
@@ -281,8 +281,6 @@ Status BetaRowsetWriter::_do_compact_segments(SegCompactionCandidatesSharedPtr s
     }
     uint64_t begin = (*(segments->begin()))->id();
     uint64_t end = (*(segments->end() - 1))->id();
-    LOG(INFO) << "BetaRowsetWriter:" << this << " do segcompaction at " << segments->size()
-              << " segments. Begin:" << begin << " End:" << end;
     uint64_t begin_time = GetCurrentTimeMicros();
 
     auto schema = std::make_shared<Schema>(_context.tablet_schema->columns(),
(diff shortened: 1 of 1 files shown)


dataroaring (approved):
LGTM

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -340,16 +338,19 @@ Status BetaRowsetWriter::_do_compact_segments(SegCompactionCandidatesSharedPtr s
...
                    << " _num_segment:" << _num_segment << " _num_segcompacted:" << _num_segcompacted
                    << " list directory:" << fmt::to_string(vlog_buffer);
     }
 
     _segcompacted_point += (end - begin + 1);
 
     uint64_t elapsed = GetCurrentTimeMicros() - begin_time;
-    LOG(INFO) << "BetaRowsetWriter:" << this << " segcompaction completed. elapsed time:" << elapsed
-              << "us. _segcompacted_point update:" << _segcompacted_point;
+    LOG(INFO) << "segcompaction completed. tablet_id:" << _context.tablet_id
```
luozenglin, line 350: Maybe `fmt::format` is better.

luozenglin: Better performance, you can refer to: https://stackoverflow.com/questions/58157033/why-is-fmt-slower-than-stdstringstream

yiguolei: I do not think we should use fmt in log... It is very ugly. If it is required, I think it is better to add another logger, like LOG_INFO(fmt, args....);

freemandealer: Thanks for explanation. Now I am a little confused at when I should use which ... I will benchmark them in my spare time.

PR #14523, opened 2022-11-23
[test](pipline)Adjust mem limit to 30% for regression

pingchunzhang:
# Proposed changes

Issue Number: close #xxx

## Problem summary

load task of teamcity regression pipline occur cancel because of over memory limit  recently, so adjust mem limit of be.conf to 30%

## Checklist(Required)

    - [ ] No
2. Has unit tests been added:
    - [ ] No Need
    - [ ] No Need
    - [ ] No
    - [ ] No


dataroaring (approved):
LGTM

PR #14531, opened 2022-11-23
[improvement](meta) make database,table,column names to support unicode (replace PR #13467 with this)

meredith620:
# Proposed changes

## Problem summary

Make database, table, column and other names support unicode by changing `LABEL_REGEX` `COMMON_NAME_REGIEX` `COMMON_TABLE_NAME_REGEX` `COLUMN_NAME_REGEX` regular expressions in class FeNameFormat.

P.S. @SharpRay has transfered PR #13467 to me, and I‘m responsible for the task now. There will be some modifications during the review period, so I create a new PR and the original #13467 could be closed. Thanks.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No


hello-stephen:
 the sum of best hot time: 33.56 seconds
 load time: 443 seconds
 storage size: 17180379792 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20221123155251_clickbench_pr_51697.html

yiguolei:
Is there any strong requirement to support chinese as table or column name? If not, I am preferred not add this support because both BE and FE use column names. There may be some potential problems if use chinese as column or table name. And also I find there are some problems in github workflow so that I remove some regression test that use chinese.

yiguolei (changes requested):
We need discuss it.

meredith620:
> Is there any strong requirement to support chinese as table or column name? If not, I am preferred not add this support because both BE and FE use column names. There may be some potential problems if use chinese as column or table name. And also I find there are some problems in github workflow so that I remove some regression test that use chinese.

First, if using doris in an system from scrach, I think it is not hard to avoid using chinese table or column names.
We found many of our customers in different industry want to expand their exists production database's analyis abilitiy by introducing an OLAP system. We recommend them to try doris, then some of them (not too few) found they have some chinese table or column names in their database but not supported by doris, so they can't do the analyis jobs by importing tables. And then they have to choose other databases not as fast as doris but support unicode names.

So here is the situation:
  1. Some companies have chinese table names or column names in their OLTP database for many years, and they want to introduce an OLAP database now
  2. There will be a huge cost and protential risks in bussiness systems to change table or column names, for they do not have as many developers as big companies
  3. Only changing names in doris after imported is not acceptable. Usually there are hundred of columns in one table, transfering the OLTP-OLAP names mapping between technical staffs and data analysts is easy to get wrong but not easy to check.

And it seems not a weird feature since the popular databases also support unicode names, e.g. mysql, postgresql. So I think to make doris to support unicode names (not only chinese) could be a solution for different companies in different industries.

yiguolei:
@meredith620 I see. Could you please add a session variable for this feature? I think we could disable this by default and then you could first try this feature in production and then we may set this to true by default in 1.3 or 1.2.3.

meredith620:
@yiguolei It's a good idea, Let me add a session variable.

meredith620:
I add a session variable 'enable_unicode_name_support', and use it to switch regex like this in class FeNameFormat
```java
  public static String getLabelRegex() {
        if (ConnectContext.get().getSessionVariable().isEnableUnicodeNameSupport()) {
            return UNICODE_LABEL_REGEX;
        } else {
            return LABEL_REGEX;
        }
  }
```

Then I find one section in [TimeUtils.java](https://github.com/apache/doris/blob/1.1.5-rc01/fe/fe-core/src/main/java/org/apache/doris/common/util/TimeUtils.java#L123)
```java
public static TimeZone getTimeZone() {
        String timezone;
        if (ConnectContext.get() != null) {
            timezone = ConnectContext.get().getSessionVariable().getTimeZone();
        } else {
            timezone = VariableMgr.getDefaultSessionVariable().getTimeZone();
        }
        return TimeZone.getTimeZone(ZoneId.of(timezone, timeZoneAliasMap));
    }
```
@yiguolei My question is when should I check if `ConnectContext.get()` returns null and use `VariableMgr.getDefaultSessionVariable()` ?

meredith620:
> I add a session variable 'enable_unicode_name_support', and use it to switch regex like this in class FeNameFormat
> 
> ```java
>   public static String getLabelRegex() {
>         if (ConnectContext.get().getSessionVariable().isEnableUnicodeNameSupport()) {
>             return UNICODE_LABEL_REGEX;
>         } else {
>             return LABEL_REGEX;
>         }
>   }
> ```
> 
> Then I find one section in [TimeUtils.java](https://github.com/apache/doris/blob/1.1.5-rc01/fe/fe-core/src/main/java/org/apache/doris/common/util/TimeUtils.java#L123)
> 
> ```java
> public static TimeZone getTimeZone() {
>         String timezone;
>         if (ConnectContext.get() != null) {
>             timezone = ConnectContext.get().getSessionVariable().getTimeZone();
>         } else {
>             timezone = VariableMgr.getDefaultSessionVariable().getTimeZone();
>         }
>         return TimeZone.getTimeZone(ZoneId.of(timezone, timeZoneAliasMap));
>     }
> ```
> 
> @yiguolei My question is when should I check if `ConnectContext.get()` returns null and use `VariableMgr.getDefaultSessionVariable()` ?

I see it fails unittest, maybe I should use the same method like TimeUtils.java

meredith620:
@yiguolei  I'm confused by FE UT and P0 Regression test. 
1. I ran FE UT locally on my machine and everything was fine. After I pushed branch github shows the FE UT was failed and I didn't find the relation between the failed job and my commit, then I retrigger the github checks, this time the FE UT was successed. Will the FE UT be impact by workload or something else?
2. There were three test suites failed in P0 Regression on github. (FOLDER:SUITE  => statistics:alter_col_stats, query_p0:test_date_function, query_p0:test_query_sys) but I run locally these three Regression tests, they are all passed.

Can you give me some advice or some experience about the above situation, thanks.

meredith620:
@yiguolei  Hi，the FE checks are all passed. A few days ago some tag was added accidently by my rebaes, and I correct the codes. Could the tags be clean manually?
image

meredith620:
@yiguolei How can I keep pushing this PR, the UT and regression test are all passed.

yiguolei (dismissed):
LGTM

yiguolei:
run buildall

meredith620:
run arm

yiguolei (approved):
LGTM

PR #14580, opened 2022-11-25
[enhancement](memory) Refactored process memory statistics periodically refresh, and fix catch bad_alloc

xinyiZzz:
# Proposed changes

Issue Number: close #xxx

## Problem summary

This pr is to reduce the probability of OOM.

1. Previously, process memory statistics were refreshed every 1 second, including `/proc/meminfo`, `/proc/self/status`, `metrics of tcmalloc and jemalloc`, consumption of each type of `mem tracker`, this refresh frequency is very Slow, making it impossible to respond in time before OOM.
Now when the process is about to exceed the limit and the system memory available is insufficient, the refresh frequency will be accelerated, which will not affect performance.

2. Fixed thread catch bad_alloc, which will also help avoid OOM

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [x] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -471,8 +471,8 @@ CONF_String(buffer_pool_limit, "20%");
 // This is the percentage of buffer_pool_limit
 CONF_String(buffer_pool_clean_pages_limit, "50%");
 
-// Sleep time in seconds between memory maintenance iterations
-CONF_mInt64(memory_maintenance_sleep_time_s, "10");
+// Sleep time in milliseconds between memory maintenance iterations
+CONF_mInt64(memory_maintenance_sleep_time_ms, "500");
 
 // Alignment
 CONF_Int32(memory_max_alignment, "16");
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -58,6 +58,16 @@ CONF_String(memory_mode, "performance");
 // it will be set to physical memory size.
 CONF_String(mem_limit, "80%");
 
+// Soft memory limit as a fraction of hard memory limit.
+CONF_Double(soft_mem_limit_frac, "0.9");
+
+// The maximum low water mark of the system `/proc/meminfo/MemAvailable`, Unit byte, default 1.6G,
+// actual low water mark=min(1.6G, MemTotal * 10%), avoid wasting too much memory on machines
+// with large memory larger than 16G.
+// Turn up max. On machines with more than 16G memory, more memory buffers will be reserved for Full GC.
+// Turn down max. will use as much memory as possible.
+CONF_Int64(max_sys_mem_available_low_water_mark_bytes, "1717986918");
+
 // the port heartbeat service used
 CONF_Int32(heartbeat_service_port, "9050");
 // the count of heart beat service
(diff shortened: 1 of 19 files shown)


adonis0147 (dismissed):
Please modify this line([be/src/runtime/thread_context.h#L381]([be/src/runtime/thread_context.h#L381](https://github.com/apache/doris/blob/master/be/src/runtime/thread_context.h#L381))) as following to fix the build on macOS.

```cpp
#define CONSUME_THREAD_MEM_TRACKER(size) (true)
```

hello-stephen:
 the sum of best hot time: 35.24 seconds
 load time: 430 seconds
 storage size: 17123343051 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20221126073322_clickbench_pr_53367.html

be/src/common/config.h
```diff
@@ -58,6 +58,16 @@ CONF_String(memory_mode, "performance");
...
 
+// Soft memory limit as a fraction of hard memory limit.
+CONF_Double(soft_mem_limit_frac, "0.9");
+
+// The maximum low water mark of the system `/proc/meminfo/MemAvailable`, Unit byte, default 1.6G,
+// actual low water mark=min(1.6G, MemTotal * 10%), avoid wasting too much memory on machines
+// with large memory larger than 16G.
+// Turn up max. On machines with more than 16G memory, more memory buffers will be reserved for Full GC.
+// Turn down max. will use as much memory as possible.
+CONF_Int64(max_sys_mem_available_low_water_mark_bytes, "1717986918");
```
yiguolei, line 69: 我觉得这个参数，咱们要不支持到MB 啊。。。 bytes 太小了，参数也不好看

be/src/common/daemon.cpp
```diff
@@ -117,6 +119,53 @@ void Daemon::memory_maintenance_thread() {
     }
 }
 
+void Daemon::memory_proc_refresh_thread() {
+    int64_t interval_milliseconds = 500;
```
yiguolei, line 123: change it to a config. If we find the refresh process has impact on normal workflow we could change this parameter to slow down refresh.

be/src/vec/common/allocator.h
```diff
@@ -134,7 +145,10 @@ class Allocator {
                                 alignment, size),
                         doris::TStatusCode::VEC_BAD_ARGUMENTS);
 
-            CONSUME_THREAD_MEM_TRACKER(size);
+            if (!CONSUME_THREAD_MEM_TRACKER(size)) {
```
yiguolei, line 148: if const expr mmap_populate

be/src/common/daemon.cpp
xinyiZzz: done

yiguolei (approved):
LGTM

PR #15018, opened 2022-12-12
[fix](merge-on-write) calc delete bitmap need all segments which _do_flush in one memtable

nextdreamblue:
when some case(need modify be.conf), a memtable may flush many segments and then calc delete bitmap with new data. but now, it just only load one segment with max sgement id and this bug will not cala delte bitmap with all data of all segment of one memtable, and will get many rows with same key from merge-on-write table.

# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/olap/memtable.h
+++ b/be/src/olap/memtable.h
@@ -146,7 +146,8 @@ private:
     void _insert_one_row_from_block(RowInBlock* row_in_block);
     void _aggregate_two_row_in_block(RowInBlock* new_row, RowInBlock* row_in_skiplist);
 
-    Status _generate_delete_bitmap();
+    Status _generate_delete_bitmap(int64_t atomic_num_segments_before_flush,
+                                   int64_t atomic_num_segments_after_flush);
 
 private:
     TabletSharedPtr _tablet;
--- a/be/src/olap/rowset/rowset_writer.h
+++ b/be/src/olap/rowset/rowset_writer.h
@@ -93,6 +93,8 @@ public:
         return Status::NotSupported("to be implemented");
     }
 
+    virtual int32_t get_atomic_num_segment() const = 0;
+
 private:
     DISALLOW_COPY_AND_ASSIGN(RowsetWriter);
 };
--- a/be/src/olap/memtable.cpp
+++ b/be/src/olap/memtable.cpp
@@ -453,7 +453,8 @@ Status MemTable::flush() {
     int64_t atomic_num_segments_before_flush = _rowset_writer->get_atomic_num_segment();
     RETURN_NOT_OK(_do_flush(duration_ns));
     int64_t atomic_num_segments_after_flush = _rowset_writer->get_atomic_num_segment();
-    RETURN_NOT_OK(_generate_delete_bitmap(atomic_num_segments_before_flush, atomic_num_segments_after_flush));
+    RETURN_NOT_OK(_generate_delete_bitmap(atomic_num_segments_before_flush,
+                                          atomic_num_segments_after_flush));
     DorisMetrics::instance()->memtable_flush_total->increment(1);
     DorisMetrics::instance()->memtable_flush_duration_us->increment(duration_ns / 1000);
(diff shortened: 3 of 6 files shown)


be/src/olap/rowset/rowset_writer.h
```diff
@@ -62,10 +62,12 @@ class RowsetWriter {
     virtual Status flush_columns() { return Status::Error<ErrorCode::NOT_IMPLEMENTED_ERROR>(); }
     virtual Status final_flush() { return Status::Error<ErrorCode::NOT_IMPLEMENTED_ERROR>(); }
 
-    virtual Status flush_single_memtable(MemTable* memtable, int64_t* flush_size) {
+    virtual Status flush_single_memtable(MemTable* memtable, int64_t* flush_size,
+                                         std::vector<int64_t>* seg_ids_for_delete_bitmap) {
```
liaoxin01, line 66: seg_ids_for_delete_bitmap -> seg_ids may be more generic.

nextdreamblue: ok

be/src/olap/memtable.h
```diff
@@ -224,6 +224,8 @@ class MemTable {
     DeleteBitmapPtr _delete_bitmap;
     RowsetIdUnorderedSet _rowset_ids;
     int64_t _cur_max_version;
+    // seg ids in one _do_flush for calc delete bitmap
+    std::vector<int64_t> _seg_ids;
```
zhannngchen, line 228: We don't need to record a vector, for MoW table, memtable is flushed one by one (`should_serial = true` when `create_flush_token`), and `segment_id` is monotonically increasing, so we just need to record the segment id before flushing.
BTW. in this way, we don't need to change the interface definition of method `add_block`, which looks a little wired in current solution.

nextdreamblue: it is better, i will fix code by this way

nextdreamblue: @zhannngchen 
貌似通过segment num来判断不大行，如果打开segment compaction，segment个数其实是一直上下浮动的。
flush_single_memtable(block)这个地方一开始就会判断是否可以进行segment compaction，如果可以就会触发一组segment compaction，会导致num_segments()不准，比如可能10个合并成1个，num_segements()返回1，继续新插入的segement的id是继续递增的，11/12/13...，必须使用这个真实的序号去获取segment的文件去load_segment才能读到正确的data文件，否则可能会读到已经被合并的不存在的文件，或者读到已有的老segment。
所以这个地方必须精确拿到每次执行时新增的，未被compaction的segement的真实id才行

zhannngchen: @nextdreamblue segment compaction 完成之前，_num_segment 这个变量是单调递增的，问题主要出在rowset_meta 里记录的 segment 数量这里。由于更新 delete bitmap时用的是由 build_tmp()构建的临时的 rowset，这时候不可能做完 segment compaction，所以可以加一个特殊的处理来在 build_tmp()时获取准确的 num_segments
```
void BetaRowsetWriter::_build_rowset_meta(std::shared_ptr<RowsetMeta> rowset_meta, bool temporary) {
    int64_t num_seg = _is_segcompacted() && !temporary ? _num_segcompacted : _num_segment;
```
@freemandealer pls also help to review this change.

nextdreamblue: 找到解决办法了，新的segment是依据writer里的_num_segment来单调递增的，加了一个获取这个值的办法，flush前后就是那些需要load的segment

zhannngchen (approved):
LGTM, nice work

yiguolei (approved):
LGTM

PR #15168, opened 2022-12-19
[Performance] Optimize performance of like expr using strstr()

weizuo93:
## Problem summary

Optimize performance of like expr using `std::strstr()` instead of `std::search()`.


yiguolei (approved):
LGTM

PR #15274, opened 2022-12-22
[fix](multi-catalog) fix and optimize iceberg v2 reader

AshinGau:
# Proposed changes

Fix and optimize iceberg v2 reader.

## Problem summary

Fix three bugs when read iceberg v2 tables:
1. The `delete position` in `delete file` represents the position of delete row in the entire file, but the `read range` in `RowGroupReader` represents the position in current row group. Therefore, we need to subtract the position of first row of current row group from `delete position`.
2. When only reading the partition columns, `RowGroupReader` skips processing the `delete position`.
3. If the `delete position` has delete all rows in a row group, the `read range` is empty, but we read the whole row group in such case.

Optimize four performance issues:
1. We change `delete position` to `delete range`, and then merge `delete range` and `read range` into the final read ranges. This process is too tedious and time-consuming. . we can merge `delete position` and `read range` directly.
2. `delete position` is ordered in a `delete file`, so we can use merge-sort, instead of ordered-set.
3. Initialize `RowGroupReader` when reading, instead of initialize all row groups when opening a `ParquetReader`, to save memory usage, and the same as `IcebergReader`.
4. Change the recursive call of `_do_lazy_read` to loop logic.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
5. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/service/internal_service.cpp
+++ b/be/src/service/internal_service.cpp
@@ -477,7 +477,7 @@ void PInternalServiceImpl::fetch_table_schema(google::protobuf::RpcController* c
     }
     std::vector<std::string> col_names;
     std::vector<TypeDescriptor> col_types;
-    st = reader->get_parsered_schema(&col_names, &col_types);
+    st = reader->get_parsed_schema(&col_names, &col_types);
     if (!st.ok()) {
         LOG(WARNING) << "fetch table schema failed, errmsg=" << st;
         st.to_protobuf(result->mutable_status());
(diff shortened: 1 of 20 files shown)


hello-stephen:
 the sum of best hot time: 34.72 seconds
 load time: 631 seconds
 storage size: 17123707946 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20221222163637_clickbench_pr_67492.html

morningman (approved):
LGTM

PR #15509, opened 2022-12-30
[feature](partition)add default list partition

### LINKED ISSUE #15507: [Feature] support default partition for list partition when creating table
### Search before asking

### Description

There is currently no default partition for list partition. Other DBMS system like GreenPlum supports such [feature](https://gpdb.docs.pivotal.io/6-5/admin_guide/ddl/ddl-partition.html#topic80) that user can create one table using stmt as follows.

```sql
CREATE TABLE rank (id int, rank int, year int, gender 
char(1), count int ) 
DISTRIBUTED BY (id)
PARTITION BY LIST (gender)
( PARTITION girls VALUES ('F'), 
  PARTITION boys VALUES ('M'), 
  DEFAULT PARTITION other );
```

I'd like Doris could support one default list partition implementation similar to GreenPlum's.
For data loading: 
> If incoming data does not match a partition's CHECK constraint and there is no default partition, the data is rejected. Default partitions ensure that incoming data that does not match a partition is inserted into the default partition.
 
For query:
> At runtime, the query optimizer scans the entire table inheritance hierarchy and uses the CHECK table constraints to determine which of the child table partitions to scan to satisfy the query's conditions. The DEFAULT partition (if your hierarchy has one) is always scanned. DEFAULT partitions that contain data slow down the overall scan time.

And there exists such occasion where user might at first insert some not satisfying prior partition key data into the table and now create one new partition. It's user's responsibility to migrate the data satisfying new partition's key in default partition to new partition. 

As for the range default partition, i'll leave it as one future work.

### Use case

It would be annoying for users who want to migrate from system supporting default partition like GreenPlum that they have to change their DDL or even worse they can't meet their business requirements which prefer storing all data even not satisfying prior partition key. For those users we can implement such feature which would help a lot. 

### Related issues

_No response_

### Code of Conduct

ByteYue:
# Proposed changes

Issue Number: close #15507
This pr implements the list default partition referred in related [issue](https://github.com/apache/doris/issues/15507). It's similar as GreenPlum's default's partition which would store all data not satisfying prior partition key's constraints and optimizer wouldn't filter default partition which means default partition would be scanned each time you try to select data from one table with default partition.

User could either create one table with default partition or alter add one default partition.

```sql
PARTITION LIST(key) {
PARTITION p1 values in (xx,xx),
PARTITION DEFAULT
}

ALTER TABLE XXX ADD PARTITION DEFAULT
```

We don't support automatically migrate data inside default partition which meets newly added partition key's constraint to newly add partition when alter add new partition. User should select default partition using new constraints as predicate and insert them to new partition.

```sql
insert into tbl select * from tbl partition default where partition_key=xx;
```

## Problem summary

Describe your changes.
For BE: add one default partition pointer field in `VOlapTablePartitionParam` which is used when loading data. If one data doesn't match any other constraints, it would be loaded into default partition.

For FE: 
1. Adapt parser to take empty partition value list as default partition.
2. Add default partition check inside `PartitionItem` and it's child class.
3. Add default partition key field inside `PartitionKey` to display the key's information
4. Adapt `PartitionPruner`. It will add default partition to scan plan by default once the table has default partition

Future work: 
1. Currently only list partition could support default partition, range partition could also support this just like GreenPlum does.
2. For simplicity I use min value of each type to serve as default partition's key value, which would make it impossible for user to take the min value as partition key once default partition exists or if user has taken min value as partition key there is no way to add one default partition. This should be fixed in the future.

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
6. Has unit tests been added:
    - [x] Yes
    - [ ] No
    - [ ] No Need
7. Has document been added or modified:
    - [x] Yes
    - [ ] No
    - [ ] No Need
8. Does it need to update dependencies:
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments


hello-stephen:
 the sum of best hot time: 34.67 seconds
 load time: 629 seconds
 storage size: 17123626678 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20221230172625_clickbench_pr_72140.html

fe/fe-core/src/main/cup/sql_parser.cup
```diff
@@ -3043,7 +3048,13 @@ partition_key_desc ::=
 
 /* list partition PartitionKeyDesc */
 list_partition_key_desc ::=
-    LPAREN list_partition_values_list:keys RPAREN
+    /* empty */
```
morningman, line 3051: How about adding `DEFAULT`?

fe/fe-core/src/main/java/org/apache/doris/catalog/PartitionKey.java
```diff
@@ -92,7 +101,9 @@ public static PartitionKey createListPartitionKeyWithTypes(List<PartitionValue>
         // PARTITION BY LIST(k1, k2)
         // (
         //     PARTITION p1 VALUES IN (("1","beijing"), ("1", "shanghai")),
-        //     PARTITION p2 VALUES IN (("2","shanghai"))
+        //     PARTITION p2 VALUES IN (("2","shanghai")),
+        //     PARTITION p3 VALUES IN,
+        //     PARTITION p4,
```
morningman, line 106:
```suggestion
        //     PARTITION p4 DEFAULT,
```

fe/fe-core/src/main/cup/sql_parser.cup
ByteYue: > How about adding `DEFAULT`?

Sounds great. It'll be more readable for user. I'll change it soon.

morningman: Not changed?

docs/zh-CN/docs/sql-manual/sql-reference/Data-Definition-Statements/Alter/ALTER-TABLE-PARTITION.md
```diff
@@ -64,6 +64,8 @@ partition_desc ["key"="value"]
 - 如指定分桶方式，只能修改分桶数，不可修改分桶方式或分桶列。如果指定了分桶方式，但是没有指定分桶数，则分桶数会使用默认值10，不会使用建表时指定的分桶数。如果要指定分桶数，则必须指定分桶方式。
 - ["key"="value"] 部分可以设置分区的一些属性，具体说明见 [CREATE TABLE](../Create/CREATE-TABLE)
 - 如果建表时用户未显式创建Partition,则不支持通过ALTER的方式增加分区
+- 如果用户使用的是List Partition则可以增加default partition
+  -  ALTER TABLE ADD PARTITION DEFAULT
```
morningman, line 68: Please explain what is default partition, eg, which value will be saved in default partition

ByteYue:
run buildall

ZhangGuoqiang666:
run p0

ZhangGuoqiang666:
run p0

ByteYue:
run buildall

ByteYue:
run buildall

morningman (approved):
LGTM

PR #15745, opened 2023-01-09
[fix](having) revert 15143 and fix having clause with multi-conditions

cambyzju:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

Firstly having clause of Mysql is really very complex, we are hard to follow all rules, so we revert [pr15143](https://github.com/apache/doris/pull/15143) to keep the logic the same as before.

Secondly the origin implementation has problem while having clause has multi-conditions.
For example:

1. case1: here `v2` inside having clause use table column `test_having_alias_tb.v2`

- `SELECT id, v1-2 as v, sum(v2) v2 FROM test_having_alias_tb GROUP BY id,v having(v2>1);`

```
ERROR 1105 (HY000): errCode = 2, detailMessage = HAVING clause not produced by aggregation output (missing from GROUP BY clause?): (`v2` > 1)
```

2. case2: here `v2` inside having clause use alias name `v2` =`sum(test_having_alias_tb.v2)`, another condition make logic of v2 differently.

- `SELECT id, v1-2 as v, sum(v2) v2 FROM test_having_alias_tb GROUP BY id,v having(v>0 AND v2>1) ORDER BY id,v;`

```
+------+------+------+
| id   | v    | v2   |
+------+------+------+
|    2 |    1 |    3 |
+------+------+------+
```

So here we try to make the having clause rules simple:
Rule1: if alias name inside having clause is the same as column name, we use column name not alias name;
Rule2: if alias name inside having clause do not have same name as column name, we use alias name;

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
4. Has unit tests been added:
    - [x] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments


hello-stephen:
 the sum of best hot time: 35.31 seconds
 load time: 474 seconds
 storage size: 17122429658 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20230109143948_clickbench_pr_76246.html

fe/fe-core/src/main/java/org/apache/doris/analysis/SelectStmt.java
```diff
@@ -1022,35 +1022,37 @@ private void analyzeAggregation(Analyzer analyzer) throws AnalysisException {
...
-                for (Expr expr : groupByClause.getGroupingExprs()) {
-                    expr.collect(SlotRef.class, groupBySlots);
-                }
-                for (Expr expr : groupBySlots) {
-                    if (excludeGroupByaliasSMap.get(expr) == null) {
+                ExprSubstitutionMap excludeAliasSMap = aliasSMap.clone();
+                List<Expr> havingSlots = Lists.newArrayList();
+                havingClause.collect(SlotRef.class, havingSlots);
+                for (Expr expr : havingSlots) {
+                    if (excludeAliasSMap.get(expr) == null) {
```
starocean999, line 1043: looks like sql "SELECT id, v1-2 as v, sum(v2) v2 FROM test_having_alias_tb GROUP BY id,v having(v2>1);" still report the same error. Please double check if it's expected behavior of this pr

cambyzju: This behavior is the same as before pr15143, because here we use column name firstly, so it report an error.
1. Before 15143, we use column name firstly; (So this SQL failed)
2. After 15143, we use column name firstly for columns inside group by; for other columns, we use alias name firstly;
(So the SQL success)
3. Now we revert 15143 to keep the logic simple: always use column name firstly. (So the SQL failed)

If we change `having(v2>1)` to `having(sum(v2)>1)`, the SQL will success.

Later we will add a config to choose alias name firstly or column name firstly: https://github.com/apache/doris/pull/15748

xy720 (approved):
LGTM

fe/fe-core/src/main/java/org/apache/doris/analysis/SelectStmt.java
starocean999: LGTM

PR #15769, opened 2023-01-10
[Fix](storage) Fix bug that cooldown time for partition in SSD is wrong

weizuo93:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Cooldown time is wrong for data in SSD, because cooldown time for all `table/partitionis` is only calculated once when class `DataProperty` loaded and that cannot be updated later. This patch is to ensure that cooldown time for each table/partition can be calculated in real time when table/partition is created.


yiguolei (approved):
LGTM

Issue #14174, opened 2022-11-10
[Bug] segcompaction check failed when rename plain


freemandealer:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Version

updated master

### What's Wrong?

Check failed: _segid_statistics_map.find(_num_segcompacted) == _segid_statistics_map.end() == true (0 vs. 1)

Check failure stack trace: ***
    @     0x564fd2da2a3d  google::LogMessage::Fail()
    @     0x564fd2da4f79  google::LogMessage::SendToLog()
    @     0x564fd2da25a6  google::LogMessage::Flush()
    @     0x564fd2da55e9  google::LogMessageFatal::~LogMessageFatal()
    @     0x564fc8cb9dd6  doris::BetaRowsetWriter::_rename_compacted_segment_plain()
    @     0x564fc8cbfd05  doris::BetaRowsetWriter::_find_longest_consecutive_small_segment()
    @     0x564fc8cc139d  doris::BetaRowsetWriter::_get_segcompaction_candidates()
    @     0x564fc8cc1bd6  doris::BetaRowsetWriter::_segcompaction_if_necessary()
    @     0x564fc8cc5c76  doris::BetaRowsetWriter::flush_single_memtable()
    @     0x564fc79c4eaa  doris::MemTable::_do_flush()
    @     0x564fc79c3c37  doris::MemTable::flush()
    @     0x564fc79acbe2  doris::FlushToken::_flush_memtable()
    @     0x564fc79b248f  doris::MemtableFlushTask::run()
    @     0x564fc99158f2  doris::ThreadPool::dispatch_thread()
    @     0x564fc9937274  std::__invoke_impl<>()
    @     0x564fc9936b13  std::__invoke<>()
    @     0x564fc9935eb2  ZNSt5_BindIFMN5doris10ThreadPoolEFvvEPS1_EE6_callIvJEJLm0EEEET_OSt5tupleIJDpT0_EESt12_Index_tupleIJXspT1_EEE
    @     0x564fc99349c3  std::_Bind<>::operator()<>()
    @     0x564fc99315b4  std::__invoke_impl<>()
[...]

PR #15785, opened 2023-01-10
[fix](compaction) segcompaction coredump if the rowset starts with a …

freemandealer:
…big segment (#14174)

# Proposed changes

Issue Number: close #xxx

## Problem summary

Check will fail because _segid_statistics_map.find(_num_segcompacted) == _segid_statistics_map.end().
Here the check is ensuring _segid_statistics_map has no existing entry indexed by _num_segcompacted.

When will _segid_statistics_map add an entry? The answer is:

after a segment is flushed, or
after segcompacting, or
after renaming the big segments which need not segcompact.
We note a segment that has never been played by segcompaction as 'raw_seg'. When these raw segments are compacted, their records will be erased from _segid_statistics_map and 'new_seg' (compacted results) will be added to the map as a replacement.

For example:

For 7 raw segments 'oooOOoo' ('O' for the big segment while 'o' for the small), we break it into four parts: 1) ooo , 2)O, 3)O, 4)oo.
Group 1 will be compacted to form 'new_seg_1-3', and raw_seg_1, raw_seg_2, raw_seg_3 are wiped out.
Group 2 will be renamed from 'raw_seg_4' to 'new_seg_2' and add it to the map.
Group 3 will be renamed from 'raw_seg_5' to 'new_seg_3' and add it to the map.
Group 4 will be compacted to form 'new_seg_6-7', and raw_seg_6, raw_seg_7 are wiped out.
Finally, we rename 'new_seg_1-3' to 'new_seg_1' and 'new_seg_6-7' to 'new_seg_4'. So we end up having new_seg_1, new_seg_2, new_seg_3, and new_seg_4.

But for those who start with one or more big segments, the problem happens.
Take 'OOoooo' as an example. We break them into 3 groups: 1) O, 2) O, 3) oooo.
Group 1 will be renamed from 'raw_seg_1' to 'new_seg_1' and add it to the map. Coz it is the first segment that is big, filenames get lined up -- src filename & dst filename are the same (ignore the raw/new sign that are only used to distinguish in this comment).

The case should be carefully handled. We do not need to actually rename it but we should count it as handled. If we miss counting (increase _num_segcompacted), the following group 2 will still want to be renamed as 'new_seg_1', but 'new_seg_1' is already in the map, causing the check to fail at last.

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/olap/rowset/beta_rowset_writer.cpp
+++ b/be/src/olap/rowset/beta_rowset_writer.cpp
@@ -206,6 +206,7 @@ Status BetaRowsetWriter::_rename_compacted_segments(int64_t begin, int64_t end) 
 
 Status BetaRowsetWriter::_rename_compacted_segment_plain(uint64_t seg_id) {
     if (seg_id == _num_segcompacted) {
+        ++_num_segcompacted;
         return Status::OK();
     }
 
--- a/be/test/olap/segcompaction_test.cpp
+++ b/be/test/olap/segcompaction_test.cpp
@@ -91,6 +91,7 @@ public:
             delete l_engine;
             l_engine = nullptr;
         }
+        config::enable_segcompaction = false;
     }
 
 protected:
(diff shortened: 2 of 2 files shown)


dataroaring (approved):
LGTM

PR #15824, opened 2023-01-11
[Enhancement](load) delete skiplist for duplicate table in memtable

zbtzbtzbt:
# Proposed changes

delete skiplist for dup table when data load.
## Problem summary

when load data, data will be insert into skiplist in memtable, for each row skiplist need to **Find** and **Sort**, which time cost is O(log(n)), I don't think it's a good way.

There are two ways to insert data into memtable:

1. insert into skiplist: **O(log(n))**, when need flush, the data is already sorted
2. insert into a block(append only): **O(1)**, when need flush, sort once.

this pr implement way2 for duplicate table: **append data to a block and sort when need flush.**

I don't know how good or bad it turned out
This pr needs to use the community's testing framework
So I submit first to see the result.

### DIFF
--- a/be/src/olap/memtable.cpp
+++ b/be/src/olap/memtable.cpp
@@ -241,8 +241,9 @@ void MemTable::_collect_memtable_results() {
     // collect dup table results
     if (_keys_type == KeysType::DUP_KEYS) {
         DCHECK(is_final);
-        vectorized::Block _input_block_dup = _input_mutable_block_dup.to_block();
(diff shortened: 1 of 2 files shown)


hello-stephen:
 the sum of best hot time: 36.64 seconds
 load time: 517 seconds
 storage size: 17134295852 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20230112052016_clickbench_pr_78264.html

Issue #11640, opened 2022-08-10
[Feature] New BulkLoad, which support build segment file isolate from Doris cluster


compasses:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-doris/issues?q=is%3Aissue) and found no similar issues.

### Description

Currently, there are several ways to load data into Doris, like broker load, stream load etc. But all these ways have some kind of shortcomings from our perspective, for example:

1. High resource cost, because each tablet has multiple replica, and each replica will do the data load separately.
2. Hard to ensure cluster stable and performance, load jobs may lead to high load of resource competition.
3. Cause query latency climbing up and down, this kind of issue may came across again and again.

So we may want a new way to do data load, like lightweight read / write splitting, which can extremely keep high-throughput write and read.

Here we just have a very rough design, and many details need clarify. 

The overall flow：

![image](https://user-images.githubusercontent.com/10161171/211504237-5020526f-cbb2-4732-93b4-8ca7d676c380.png)

The new bulk load may have some connect with function like backup / restore, broker load etc. 
1. FE issue the bulk load command, and BE will write tablet meta to HDFS.
2. Then FE will schedule a spark / flink job to run segment builder, which will read HDFS data file and build segment file to local, and upload these segment file to HDFS when build finish.
3. Then FE will start to load these segment from HDFS, mainly each BE do the real job.
4. Last the FE need publish this transaction like the broker load.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #15839, opened 2023-01-11
[Feature] support segment builder tool

compasses:
# Proposed changes

Issue Number: just one part of #11640

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ✓] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ✓] No Need
    - [ ] Yes
    - [ ✓] No
    - [ ] No Need
    - [ ] Yes
    - [ ✓] No
    - [ ✓] No

## Further comments
This PR is one part of our bulk load implementation, which provide the tool to build the segment file of a tablet in an external way.
It's support build local and HDFS, which means you need provide the meta file and the data file like this:

```
./segment_builder --meta_file=/path/to/hdr/88409.hdr --data_path=/path/to/data/file --format=parquet --is_remote=false

ll /path/to/data/file
xxx1..gz.parquet
xxx2..gz.parquet
...
```
If the file all from the HDFS, the path should be the HDFS path. Currently only support parquet.

Since from internal we use the privately-owned HDFS lib, **so this PR HDFS related code may not work**.  I don't have such open source HDFS environment to test it.

image

From above picture you can see the final work flow:
1. Read the hdr file from the meta path, do some validation and system initialization.
2. Build the HDFS scanner, and read the parquet file from HDFS directly, and generate the segment file on local disk.
3. At last upload the segment file to HDFS, same path with the hdr file, and all these files will be used by the load segment statement.

(@morningman edit)
I sorted out the main process for easy review:

1. `BuilderHelper::initial_build_env()`

	Initialize some objects necessary to start a BE instance

2. `BuilderHelper::open()`

	Open the `StorageEngine` object. And check that the relevant directories are correct, etc.

3. `BuilderHelper::build()`

	1. Read header file from HDFS or local file
	2. Generate `TabletMeta` and `Tablet` object according to header
	3. Collect the list of data files that need to be read
	4. Create `BuilderScannerMemtable` and call `doSegmentBuild(files)`

		1. Generate `BrokerScanNode` and initialize it.
		2. Generate `DeltaWriter` and initialize it
		3. Loop through the `BrokerScanNode` to read the file and write to the `DeltaWriter`
		4. When the reading is complete, close the `DeltaWriter`
		5. Call `revise_rs_metas` to modify `tablet_meta`

	5. After the writing is completed, a new header file is generated
	6. Upload segment file and header file to HDFS if needed

### DIFF
--- a/be/src/olap/tablet_manager.h
+++ b/be/src/olap/tablet_manager.h
@@ -142,7 +142,6 @@ public:
 
     std::set<int64_t> check_all_tablet_segment(bool repair);
 
-private:
     // Add a tablet pointer to StorageEngine
     // If force, drop the existing tablet add this new one
     //
@@ -151,7 +150,7 @@ private:
     //        Status::Error<UNINITIALIZED>(), if not inited
     Status _add_tablet_unlocked(TTabletId tablet_id, const TabletSharedPtr& tablet,
                                 bool update_meta, bool force);
-
+private:
     Status _add_tablet_to_map_unlocked(TTabletId tablet_id, const TabletSharedPtr& tablet,
                                        bool update_meta, bool keep_files, bool drop_old);
 
--- a/be/src/tools/builder_helper.cpp
+++ b/be/src/tools/builder_helper.cpp
@@ -1,0 +1,295 @@
+#include "tools/builder_helper.h"
+
+#include <cstdlib>
+#include <filesystem>
+#include <fstream>
+#include <iostream>
+#include <set>
+#include <sstream>
+#include <string>
+
+#include "common/object_pool.h"
+#include "common/status.h"
+#include "env/env.h"
+#include "exec/parquet_scanner.h"
+#include "exprs/cast_functions.h"
+#include "gen_cpp/Descriptors_types.h"
+#include "gen_cpp/PaloInternalService_types.h"
+#include "gen_cpp/PlanNodes_types.h"
+#include "gen_cpp/olap_file.pb.h"
+#include "gen_cpp/segment_v2.pb.h"
+#include "gutil/strings/numbers.h"
+#include "gutil/strings/split.h"
+#include "gutil/strings/substitute.h"
+#include "io/buffered_reader.h"
+#include "io/file_factory.h"
+#include "io/file_reader.h"
+#include "io/local_file_reader.h"
+#include "json2pb/pb_to_json.h"
+#include "olap/data_dir.h"
+#include "olap/file_helper.h"
+#include "olap/olap_define.h"
+#include "olap/options.h"
+#include "olap/row.h"
+#include "olap/rowset/beta_rowset.h"
+#include "olap/rowset/rowset_id_generator.h"
+#include "olap/rowset/rowset_meta_manager.h"
+#include "olap/rowset/segment_v2/binary_plain_page.h"
+#include "olap/rowset/segment_v2/column_reader.h"
+#include "olap/schema_change.h"
+#include "olap/storage_engine.h"
+#include "olap/storage_policy_mgr.h"
+#include "olap/tablet.h"
+#include "olap/tablet_meta.h"
+#include "olap/tablet_meta_manager.h"
+#include "olap/tablet_schema.h"
+#include "olap/tablet_schema_cache.h"
+#include "olap/utils.h"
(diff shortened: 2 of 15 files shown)


hello-stephen:
 the sum of best hot time: 35.77 seconds
 load time: 496 seconds
 storage size: 17122683656 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20230111160607_clickbench_pr_77941.html

be/src/tools/builder_helper.cpp
```diff
@@ -0,0 +1,295 @@
...
+    if (!doris::config::init(nullptr, true, false, true)) {
+        LOG(FATAL) << "init config fail";
+        exit(-1);
+    }
+    CpuInfo::init();
+    DiskInfo::init();
+    MemInfo::init();
+
+    // write buffer size before flush
+    config::write_buffer_size = 14097152000;
```
spaces-X, line 90: why need to redefine these config variable here?

be/src/tools/builder_helper.cpp
```diff
@@ -0,0 +1,295 @@
...
+    if (_isHDFS) {
+        auto* dfs = hdfsFileSystem.get();
+        std::vector<io::Path> names;
+        dfs->list(_data_path, &names);
+        for (const auto& name : names) {
+            std::string filename = name.filename().c_str();
+            if (filename.substr(filename.size() - _file_type.size()) == _file_type) {
+                LOG(INFO) << "get file:" << name;
+                files.emplace_back(filename);
+            }
```
spaces-X, line 224: suggest to add some warning log when the file type are not matched both in hdfs and local dir.

be/src/tools/builder_scanner_memtable.cpp
```diff
@@ -0,0 +1,346 @@
...
+                              tuple_desc,
+                              &(tuple_desc->slots())};
+
+    DeltaWriter* delta_writer = nullptr;
+    DeltaWriter::open(&write_req, &delta_writer, load_id);
+    status = delta_writer->init();
+    if (!status.ok()) LOG(FATAL) << "delta_writer init fail:" << status.to_string();
+
+    std::filesystem::path segment_path(std::filesystem::path(_build_dir + "/segment"));
+    std::filesystem::remove_all(segment_path);
```
spaces-X, line 284: Nit::
Can we guarantee that the `segment_path` is always a new empty directory ? 
If so it had better to failed rather than `remove all` when there are files under this path.

morningman:
Some questions and suggestions:

1. `builder_scanner` doesn't seem to be used? Only `builder_scanner_memtable` is used?
2. Need to unify the inputs and outputs:
	
	1. Inputs:
	
		* header file in json and data file in parquet(can be orc or other supported file format)
		* In the code, the reading methods of different file systems can be unified, and there is no need to use `isHDFS` to judge.
	
	2. Outputs
	
		* new header file in json and Doris segment data file

3. The final upload logic can be encapsulated without being limited to HDFS
4. I think we can generate a manifest file to save all output file list. So that the downstream system can read it.

be/src/olap/tablet_manager.h
```diff
@@ -151,7 +150,7 @@ class TabletManager {
     //        Status::Error<UNINITIALIZED>(), if not inited
     Status _add_tablet_unlocked(TTabletId tablet_id, const TabletSharedPtr& tablet,
```
freemandealer, line 151: As a gentle reminder, strip the '_' for the public method

be/src/tools/segment_builder.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include "util/crc32c.h"
+#include "util/file_utils.h"
+#include "util/runtime_profile.h"
+#include "util/time.h"
+#include "vec/exec/format/parquet/vparquet_file_metadata.h"
+#include "vec/exec/vbroker_scan_node.h"
+
+DEFINE_string(meta_file, "", "tablet header meta file");
+DEFINE_string(data_path, "", "this tablet's data to be build");
+DEFINE_bool(is_remote, true, "input data's format, just support parquet");
```
freemandealer, line 82: misdescription for is_remote

PR #15873, opened 2023-01-12
[feature](cooldown) Decouple storage policy and resource

platoneko:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Currently, when users want to use the cold hot separation feature, they need to create resource first and then use created resource to create storage policies. However, for some historical reasons, the BE side will save the properties of resource in the `StoragePolicy`, and construct a `FileSystem` object for each storage policy, store the `FileSystem` object with storage policy name as key in global `FileSystemMap`. For the cooldowned remote rowsets, the storage policy name is recorded to find the location of the data in remote storage. The disadvantages of this design are:
1. Multiple storage policies are likely to use the same resource, but BE will construct `FileSystem` objects for each storage policy, resulting in a waste of memory resources.
2. FE drop storage policy will cause the cooldowned rowset to be unable to find the remote path through the storage policy name, resulting in data loss. It is more appropriate for rowset to index the remote path through the resource id.
3. The inconsistency between BE's implementation and FE semantics of storage policy may lead to difficulties in future development and maintenance.

Note that because the value stored by rowset to locate the remote storage is changed from storage policy name to resource id, the data cooldowned by the storage policy in the previous versions cannot be accessed, which means this PR does not consider compatibility of cold hot feature with previous versions.

In addition, the mechanism for FE to synchronize storage policies to BE has been updated to reduce network overhead. BE periodic report tablets with resources and storage policies in BE's memory, and FE receives the tablet report from BE, then calculates the difference of resources and storage policies between BE and FE, and sends these diffs to BE for synchronization.

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [ ] No

## Further comments


hello-stephen:
 the sum of best hot time: 36.18 seconds
 load time: 487 seconds
 storage size: 17122510321 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20230113185350_clickbench_pr_80131.html

dataroaring (approved):
LGTM

PR #16161, opened 2023-01-28
[fix](inverted index) fulltext match query should not prune by zone map

Tanya-W:
# Proposed changes
zone map prune data by min/max value, should not prune by zone map when fulltext match term is substring in a string type column's value, otherwise, some data meeting the conditions will be pruned.

such like:
- test data:

|  col_1   |
|  ----  |
| hello world |
| say hello world |
| zero |

- match query:
```
select * from tb where col_1 match 'world';
```
`hello world` and `say hello world` should be returned to the user instead of being pruned by zone map.

## Problem summary

Describe your changes.

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_v2/segment_iterator.cpp
+++ b/be/src/olap/rowset/segment_v2/segment_iterator.cpp
@@ -367,6 +367,12 @@ Status SegmentIterator::_get_row_ranges_from_conditions(RowRanges* condition_row
     RowRanges zone_map_row_ranges = RowRanges::create_single(num_rows());
     // second filter data by zone map
     for (auto& cid : cids) {
+        if (_inverted_index_iterators[_schema.unique_id(cid)] != nullptr &&
+            _inverted_index_iterators[_schema.unique_id(cid)]->get_inverted_index_reader_type() ==
+                    InvertedIndexReaderType::FULLTEXT) {
+            // fulltext match query should not prune by zone map
+            continue;
+        }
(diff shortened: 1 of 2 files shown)


hello-stephen:
 the sum of best hot time: 35.29 seconds
 load time: 539 seconds
 storage size: 17122430817 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20230128111713_clickbench_pr_85584.html

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -367,6 +367,12 @@ Status SegmentIterator::_get_row_ranges_from_conditions(RowRanges* condition_row
     RowRanges zone_map_row_ranges = RowRanges::create_single(num_rows());
     // second filter data by zone map
     for (auto& cid : cids) {
+        if (_inverted_index_iterators[_schema.unique_id(cid)] != nullptr &&
```
airborne12, line 370: may be we can use the same logic above to tell whether it's a fulltext match query or not?

PR #16326, opened 2023-02-01
[improvement](multi-catalog) increase default batch_size to 4064

AshinGau:
# Proposed changes

Issue Number: close #xxx

## Problem summary
The performance of ClickBench Q30 is affected by batch_size:
| batch_size | 1024 | 4096 | 20480 |
| -- | -- | -- | -- |
| Q30 query time | 2.27 | 1.08 | 0.62 |

Because aggregation operator will create a new result block for each batch block, and Q30 has 90 columns, which is time-consuming. Larger batch_size will decrease the number of aggregation blocks, so the larger batch_size will improve performance.

Doris internal reader will read at least 4064 rows even if batch_size < 4064, so this PR keep the process of reading external table the same  as internal table.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/vec/exec/format/csv/csv_reader.cpp
+++ b/be/src/vec/exec/format/csv/csv_reader.cpp
@@ -207,7 +207,7 @@ Status CsvReader::get_next_block(Block* block, size_t* read_rows, bool* eof) {
         return Status::OK();
     }
 
-    const int batch_size = _state->batch_size();
+    const int batch_size = std::max(_state->batch_size(), (int)_MIN_BATCH_SIZE);
     size_t rows = 0;
     auto columns = block->mutate_columns();
     while (rows < batch_size && !_line_reader_eof) {
--- a/be/src/vec/exec/format/generic_reader.h
+++ b/be/src/vec/exec/format/generic_reader.h
@@ -60,6 +60,8 @@ public:
     }
 
 protected:
+    const size_t _MIN_BATCH_SIZE = 4064; // 4094 - 32(padding)
+
     /// Whether the underlying FileReader has filled the partition&missing columns
     bool _fill_all_columns = false;
 };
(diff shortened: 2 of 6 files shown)


SaintBacchus:
We usually set the batch_size to be `4096` in session, is (n*1024 -32) better than (n*1024)?

BiteTheDDDDt:
> We usually set the batch_size to be `4096` in session, is `n*1024 -32` better than `n*1024`?

Like the comment said: "4096 minus 16 + 16 bytes padding that in padding pod array"
If we set batch size to 4096 directly, we will waste many memory.

SaintBacchus:
@BiteTheDDDDt if we I set it to 4096, it only will waster 32 bytes? or waster total 4096 bytes?

And if we waster a lot of memroy, should we add a doc to point out it in doc or just pass the num which is batch_size session value minus 32 to the be node.

AshinGau:
> @BiteTheDDDDt if we I set it to 4096, it only will waster 32 bytes? or waster total 4096 bytes?
> 
> And if we waster a lot of memroy, should we add a doc to point out it in doc or just pass the num which is batch_size session value minus 32 to the be node.

The batch_size is not aligned with n*1024 -32, 1024 - (batch_size + 32) % 1024 will be wasted.

hello-stephen:
 the sum of best hot time: 33.37 seconds
 load time: 493 seconds
 storage size: 17171141953 Bytes
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20230201154239_clickbench_pr_88746.html

morningman (approved):
LGTM

PR #16574, opened 2023-02-09
[improvement](filecache) Use consistent hash to assign the same scan range into the same backend among different queries

AshinGau:
# Proposed changes

When file cache enabled, running the same query for the second time may be still slow, for `FE` will assign the same scan range into different backends among different queries, and the former cached data in `BE` will be useless if the scan range is changed.

So, this PR introduce consistent hash to assign the same scan range into the same backend among different queries.

## Checklist(Required)

    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/Coordinator.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/Coordinator.java
@@ -31,6 +31,7 @@ import org.apache.doris.common.Status;
 import org.apache.doris.common.UserException;
 import org.apache.doris.common.telemetry.ScopedSpan;
 import org.apache.doris.common.telemetry.Telemetry;
+import org.apache.doris.common.util.ConsistentHash;
 import org.apache.doris.common.util.DebugUtil;
 import org.apache.doris.common.util.ListUtil;
 import org.apache.doris.common.util.ProfileWriter;
@@ -59,6 +60,7 @@ import org.apache.doris.planner.RuntimeFilterId;
 import org.apache.doris.planner.ScanNode;
 import org.apache.doris.planner.SetOperationNode;
 import org.apache.doris.planner.UnionNode;
+import org.apache.doris.planner.external.ExternalScanNode;
 import org.apache.doris.proto.InternalService;
 import org.apache.doris.proto.InternalService.PExecPlanFragmentResult;
 import org.apache.doris.proto.InternalService.PExecPlanFragmentStartRequest;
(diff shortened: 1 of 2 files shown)


morningman (approved):
LGTM

fe/fe-core/src/main/java/org/apache/doris/qe/Coordinator.java
```diff
@@ -1768,12 +1797,56 @@ public TScanRangeLocation selectBackendsByRoundRobin(List<TScanRangeLocation> lo
...
+        }
+        int virtualNumber = Math.max(Math.min(512 / aliveBEs.size(), 32), 2);
+        ConsistentHash<TScanRangeLocations, Backend> consistentHash = new ConsistentHash<>(
+                Hashing.murmur3_128(), new ScanRangeHash(), new BackendHash(), aliveBEs, virtualNumber);
+        for (TScanRangeLocations scanRangeLocations : locations) {
+            TScanRangeLocation minLocation = scanRangeLocations.locations.get(0);
+            Backend backend = consistentHash.getNode(scanRangeLocations);
+            TNetworkAddress execHostPort = new TNetworkAddress(backend.getHost(), backend.getBePort());
+            this.addressToBackendID.put(execHostPort, backend.getId());
+            // Why only increase 1 in other implementations ?
```
morningman, line 1819: TODO: because the file size of each scan range is same, so we can just use 1

fe/fe-core/src/main/java/org/apache/doris/qe/Coordinator.java
```diff
@@ -1768,12 +1797,56 @@ public TScanRangeLocation selectBackendsByRoundRobin(List<TScanRangeLocation> lo
...
+            FragmentScanRangeAssignment assignment,
+            Map<TNetworkAddress, Long> assignedBytesPerHost,
+            Map<TNetworkAddress, Long> replicaNumPerHost) throws Exception {
+        Collection<Backend> aliveBEs = idToBackend.values().stream().filter(SimpleScheduler::isAvailable)
+                .collect(Collectors.toList());
+        if (aliveBEs.isEmpty()) {
+            throw new UserException("No available backends");
+        }
+        int virtualNumber = Math.max(Math.min(512 / aliveBEs.size(), 32), 2);
+        ConsistentHash<TScanRangeLocations, Backend> consistentHash = new ConsistentHash<>(
```
morningman, line 1812: TODO: I think we can move this ConsistentHash to `SystemInfoService`.
No need to create one for every query

PR #16605, opened 2023-02-10
[bugfix](column_reader) index_page should not be pre-decoded

yixiutt:
In our current logic, index page will be pre-decoded but it will return OK as index page use BinaryPlainPageBuilder and first 4 bytes of the page is a offset so it's high probablility not equal to EncodingTypePB::DICT_ENCODING which is 5.
Code in bitshuffle_page_pre_decode.h
```
   	if constexpr (USED_IN_DICT_ENCODING) {
            auto type = decode_fixed32_le((const uint8_t*)&data.data[0]);
            if (static_cast<EncodingTypePB>(type) != EncodingTypePB::DICT_ENCODING) {
                return Status::OK();
            }
            size_of_dict_header = BINARY_DICT_PAGE_HEADER_SIZE;
            data.remove_prefix(4);
        }
```
But if type just equal to EncodingTypePB::DICT_ENCODING and then it will use BitShuffle to decode BinaryPlainPage, which will leads to an fatal error.

# Proposed changes

Issue Number: close #xxx

## Problem summary

image

## Checklist(Required)

    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No Need
    - [ ] Yes
    - [ ] No
    - [ ] No

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_v2/column_reader.cpp
+++ b/be/src/olap/rowset/segment_v2/column_reader.cpp
@@ -210,6 +210,10 @@ Status ColumnReader::read_page(const ColumnIteratorOptions& iter_opts, const Pag
     opts.type = iter_opts.type;
     opts.encoding_info = _encoding_info;
     opts.io_ctx = iter_opts.io_ctx;
+    // index page should not pre decode
+    if (iter_opts.type == INDEX_PAGE) {
+        opts.pre_decode = false;
+    }
 
     return PageIO::read_and_decompress_page(opts, handle, page_body, footer);
 }


be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -210,6 +210,10 @@ Status ColumnReader::read_page(const ColumnIteratorOptions& iter_opts, const Pag
     opts.type = iter_opts.type;
     opts.encoding_info = _encoding_info;
     opts.io_ctx = iter_opts.io_ctx;
+    // index page should not pre decode
+    if (iter_opts.type == INDEX_PAGE) {
```
carlvinhust2012, line 214: maybe add some regression test will more better ?

zhannngchen:
Add a DCHECK for the type variable to make sure we used the right decoder? 
```
if constexpr (USED_IN_DICT_ENCODING) {
            auto type = decode_fixed32_le((const uint8_t*)&data.data[0]);
            if (static_cast<EncodingTypePB>(type) != EncodingTypePB::DICT_ENCODING) {
                return Status::OK();
            }
            size_of_dict_header = BINARY_DICT_PAGE_HEADER_SIZE;
            data.remove_prefix(4);
        }
```

hello-stephen:
 the sum of best hot time: 34.17 seconds
 stream load tsv:          465 seconds loaded 74807831229 Bytes, about 153 MB/s
 stream load orc:          68 seconds loaded 1101869774 Bytes, about 15 MB/s
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20230210143935_clickbench_pr_93863.html

be/src/olap/rowset/segment_v2/column_reader.cpp
yixiutt: it's a inner bug and logic here is two mistakes leads to running correnct, we already have case.

yixiutt:
> Add a DCHECK for the type variable to make sure we used the right decoder?
> 
> ```
> if constexpr (USED_IN_DICT_ENCODING) {
>             auto type = decode_fixed32_le((const uint8_t*)&data.data[0]);
>             if (static_cast<EncodingTypePB>(type) != EncodingTypePB::DICT_ENCODING) {
>                 return Status::OK();
>             }
>             size_of_dict_header = BINARY_DICT_PAGE_HEADER_SIZE;
>             data.remove_prefix(4);
>         }
> ```

only USED_IN_DICT_ENCODING have a type, other encoder may encode offset or nums in the first bytes

dataroaring (approved):
LGTM

PR #16619, opened 2023-02-10
Revert "[Performance] Optimize performance of like expr using strstr()"

Henry2SS:
Reverts apache/doris#15168

in our case,  there are one table named `table_A`, having a column `col_1`. It has value "三星" and "红包". 

a sql like 
```
select `col_1` from table_A where ... and `col_1` like '%红包%' ... group by xxx  LEFT JOIN (select `col_1` from table_A where ...  and `col_1` like '%红包%' ... group by xxx）on xxxx = xxxx
```
returns a row that `col_1` == "三星"。

### DIFF
--- a/be/src/runtime/string_search.hpp
+++ b/be/src/runtime/string_search.hpp
@@ -41,11 +41,13 @@ public:
             return -1;
         }
 
-        char* occurence = std::strstr(str->ptr, _pattern->ptr);
-        if (occurence == nullptr) {
+        auto it = std::search(str->ptr, str->ptr + str->len,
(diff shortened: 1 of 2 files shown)


morningman:
Do we have same issue on master?

Henry2SS:
> Do we have same issue on master?

tested locally, master doesn't have the same problem.

Henry2SS:
> Do we have same issue on master?

and also no problem with 1.2-lts

caiconghui:
cc @weizuo93

weizuo93 (approved):

PR #16639, opened 2023-02-11
[Improvement](brpc) Using a thread pool for RPC service avoiding std::mutex block brpc::bthread

### LINKED ISSUE #16634: [Enhancement] replace brpc bthread with pthead
### Search before asking

### Description

The bthread of brpc cannot be scheduled under high load, resulting in stuck interface callback. A common problem is:
`timeout when waiting for send fragments RPC`
This interface takes a short time. It should return in hundreds of microseconds, but it will be stuck for a long time under high load. We are waiting for the lock through the perf top tool under high load

image

The reason is that bthreads need to find free ptreads to schedule successfully. If all pthreads are blocking, bthreads will wait all the time. What is the reason for the long blocking of pthreads?

*The main reason is that the pthread blocking function is called, such as waiting for a lock (std:: mutex)

The doris be interface uses a lot of std:: mutex. It is easy to exhaust pthread threads under high load. A simple query can generate tens of thousands or tens of thousands of qps between be, affecting the processing of all requests of brpc

How to solve

There are usually three ways
- 1、brpc_ num_ Thread: increase the number of threads
- 2、bthread:: mutex replaces std:: mutex
- 3、brpc+thread pool mode: brpc will no longer process requests, but only send and receive network requests, which will be sent to the thread pool after being received

Method 1： addresses the symptoms rather than the root cause, and pthread depletion will still affect all brpc interfaces
Method 2：is difficult to implement. It is not easy to operate when it is developed by many people. No way, everyone knows which functions need bthread mutex and which need pthread mutex
Method 3： can completely solve the impact of pthread depletion

be在高负载下brpc的bthread存在无法调度的问题，导致卡住接口的回调，一个常见的问题就是:
timeout when waiting for send fragments RPC
这个接口的耗时很短，通常数百微秒就应该返回，但是在高负载下会卡住很久，我们在高负载下通过perf top工具都在等待锁
image

原因是：bthread需要找到空闲的ptread才能成功调度，如果所有的pthread都在阻塞那么bthread会一直等待，那么导致pthread阻塞长时间阻塞的原因是啥？
* 主要是调用了pthread阻塞函数，比如在等待一把锁(std::mutex)

doris be接口中大量使用了std::mutex，在高负载下很容易把pthread的线程耗尽，一个简单的查询在be间产生的qps就能达到上万或者数万，影响brpc的所有请求的处理

如何解决
通常有3种方式
- 1.brpc_num_thread: 调大线程的数量
- 2.bthread::mutex替换std::mutex
- 3.brpc+线程池方式：brpc不再处理请求，只做网络请求收发，请求接到到丢到线程池中

方式1 治标不治本，pthread耗尽仍会影响所有brpc接口
方案2  落地较难，这个多人开发的时候不太好操作了。没办法每个人都知道哪些函数需要bthread的mutex,哪些是需要pthread mutex
方案3  可以彻底解决pthread耗尽影响

### Solution

add thread pool to handle the be service logic, do not use brpc any more

### Code of Conduct

chenlinzhong:
# Proposed changes

Issue Number: close #16634 

## Problem summary

mainly include:
1. brpc service adds two types of thread pools. The number of "light" and "heavy" thread pools is different
2. Classify the interfaces of be. Those related to data transmission are classified as heavy interfaces and others as light interfaces
3. Add some monitoring to the thread pool, including the queue size and the number of active threads. Use these indicators to guide the configuration of the number of threads

主要包括
- 1.brpc服务增加2类线程池，"轻","重"  不同线程池的数量配置不一样
- 2.对be的接口进行分类，跟数据传输有关的归类为重接口，其他为轻接口
- 3.增加一些监控，包括be每个接口qps和线程池队列大小、活跃线程个数，根据这些指标来指导线程数的配置
image

## Checklist(Required)

    - [x] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
    - [ ] Yes
    - [x] No
    - [ ] No Need
    - [ ] Yes
    - [x] No
    - [x] No

## Further comments

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -35,8 +35,10 @@ CONF_Int32(be_port, "9060");
 // port for brpc
 CONF_Int32(brpc_port, "8060");
 
-// the number of bthreads for brpc, the default value is set to -1, which means the number of bthreads is #cpu-cores
-CONF_Int32(brpc_num_threads, "-1");
+// the number of bthreads for brpc, the default value is set to 32
+// brpc only for network service send or accept request
+// no more  process any logic
+CONF_Int32(brpc_num_threads, "32");
 
 // port to brpc server for single replica load
 CONF_Int32(single_replica_load_brpc_port, "8070");
@@ -385,8 +387,15 @@ CONF_Int32(single_replica_load_download_num_workers, "64");
 CONF_Int64(load_data_reserve_hours, "4");
 // log error log will be removed after this time
 CONF_mInt64(load_error_log_reserve_hours, "48");
-CONF_Int32(number_tablet_writer_threads, "16");
-CONF_Int32(number_slave_replica_download_threads, "64");
+
+// be brpc interface is classified into two categories: light and heavy
+// each category has diffrent thread number
+// threads to handle heavy api interface, such as transmit_data/transmit_block etc
+CONF_Int32(brpc_heavy_work_pool_threads, "192");
+// threads to handle light api interface, such as exec_plan_fragment_prepare/exec_plan_fragment_start
+CONF_Int32(brpc_light_work_pool_threads, "32");
+CONF_Int32(brpc_heavy_work_pool_max_queue_size, "10240");
+CONF_Int32(brpc_light_work_pool_max_queue_size, "10240");
 
 // The maximum amount of data that can be processed by a stream load
 CONF_mInt64(streaming_load_max_mb, "10240");
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -35,10 +35,9 @@ CONF_Int32(be_port, "9060");
 // port for brpc
 CONF_Int32(brpc_port, "8060");
 
-// the number of bthreads for brpc, the default value is set to 32
-// brpc only for network service send or accept request
-// no more  process any logic
-CONF_Int32(brpc_num_threads, "32");
+// the number of bthreads for brpc, the default value is set to -1,
+// which means the number of bthreads is #cpu-cores
+CONF_Int32(brpc_num_threads, "-1");
 
 // port to brpc server for single replica load
 CONF_Int32(single_replica_load_brpc_port, "8070");
--- a/be/src/service/internal_service.cpp
+++ b/be/src/service/internal_service.cpp
@@ -173,7 +173,7 @@ void PInternalServiceImpl::tablet_writer_open(google::protobuf::RpcController* c
                                               const PTabletWriterOpenRequest* request,
                                               PTabletWriterOpenResult* response,
                                               google::protobuf::Closure* done) {
-    _light_work_pool.offer([this, request, response, done]() {
+    bool ret = _light_work_pool.offer([this, request, response, done]() {
         VLOG_RPC << "tablet writer open, id=" << request->id()
                  << ", index_id=" << request->index_id() << ", txn_id=" << request->txn_id();
         brpc::ClosureGuard closure_guard(done);
@@ -185,6 +185,12 @@ void PInternalServiceImpl::tablet_writer_open(google::protobuf::RpcController* c
         }
         st.to_protobuf(response->mutable_status());
     });
+    if (!ret) {
+        LOG(WARNING) << "fail to offer request to the work pool";
+        brpc::ClosureGuard closure_guard(done);
+        response->mutable_status()->set_status_code(TStatusCode::CANCELLED);
+        response->mutable_status()->add_error_msgs("fail to offer request to the work pool");
+    }
 }
 
 void PInternalServiceImpl::exec_plan_fragment(google::protobuf::RpcController* controller,
@@ -209,29 +215,41 @@ void PInternalServiceImpl::exec_plan_fragment_prepare(google::protobuf::RpcContr
                                                       const PExecPlanFragmentRequest* request,
                                                       PExecPlanFragmentResult* response,
                                                       google::protobuf::Closure* done) {
-    _light_work_pool.offer([this, controller, request, response, done]() {
+    bool ret = _light_work_pool.offer([this, controller, request, response, done]() {
         exec_plan_fragment(controller, request, response, done);
     });
+    if (!ret) {
+        LOG(WARNING) << "fail to offer request to the work pool";
+        brpc::ClosureGuard closure_guard(done);
+        response->mutable_status()->set_status_code(TStatusCode::CANCELLED);
+        response->mutable_status()->add_error_msgs("fail to offer request to the work pool");
+    }
 }
 
 void PInternalServiceImpl::exec_plan_fragment_start(google::protobuf::RpcController* controller,
                                                     const PExecPlanFragmentStartRequest* request,
                                                     PExecPlanFragmentResult* result,
                                                     google::protobuf::Closure* done) {
-    _light_work_pool.offer([this, controller, request, result, done]() {
+    bool ret = _light_work_pool.offer([this, controller, request, result, done]() {
         auto span = telemetry::start_rpc_server_span("exec_plan_fragment_start", controller);
         auto scope = OpentelemetryScope {span};
         brpc::ClosureGuard closure_guard(done);
         auto st = _exec_env->fragment_mgr()->start_query_execution(request);
         st.to_protobuf(result->mutable_status());
     });
+    if (!ret) {
+        LOG(WARNING) << "fail to offer request to the work pool";
+        brpc::ClosureGuard closure_guard(done);
+        result->mutable_status()->set_status_code(TStatusCode::CANCELLED);
+        result->mutable_status()->add_error_msgs("fail to offer request to the work pool");
+    }
 }
 
 void PInternalServiceImpl::tablet_writer_add_block(google::protobuf::RpcController* controller,
                                                    const PTabletWriterAddBlockRequest* request,
                                                    PTabletWriterAddBlockResult* response,
                                                    google::protobuf::Closure* done) {
-    _heavy_work_pool.offer([this, controller, request, response, done]() {
+    bool ret = _heavy_work_pool.offer([this, controller, request, response, done]() {
         // TODO(zxy) delete in 1.2 version
         google::protobuf::Closure* new_done = new NewHttpClosure<PTransmitDataParams>(done);
         brpc::Controller* cntl = static_cast<brpc::Controller*>(controller);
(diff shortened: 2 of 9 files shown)


hello-stephen:
 the sum of best hot time: 34.68 seconds
 stream load tsv:          469 seconds loaded 74807831229 Bytes, about 152 MB/s
 stream load orc:          68 seconds loaded 1101869774 Bytes, about 15 MB/s
 stream load parquet:          28 seconds loaded 861443392 Bytes, about 29 MB/s
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20230211162649_clickbench_pr_94553.html

be/src/service/internal_service.cpp
```diff
@@ -286,125 +334,134 @@ Status PInternalServiceImpl::_exec_plan_fragment(const std::string& ser_request,
...
-    VLOG_RPC << "fetch table schema";
-    brpc::ClosureGuard closure_guard(done);
-    TFileScanRange file_scan_range;
-    Status st = Status::OK();
-    {
-        const uint8_t* buf = (const uint8_t*)(request->file_scan_range().data());
-        uint32_t len = request->file_scan_range().size();
-        st = deserialize_thrift_msg(buf, &len, false, &file_scan_range);
+    DorisMetrics::instance()->fetch_table_schema->increment(1);
+    _light_work_pool.offer([request, result, done]() {
```
morningman, line 380: This is a heavy work, it may read the remote file

be/src/service/internal_service.cpp
```diff
@@ -423,200 +480,233 @@ void PInternalServiceImpl::tablet_fetch_data(google::protobuf::RpcController* co
...
-                                ->get_kafka_latest_offsets_for_partitions(
-                                        request->kafka_meta_request(), &partition_offsets);
-            if (st.ok()) {
-                PKafkaPartitionOffsets* part_offsets = response->mutable_partition_offsets();
-                for (const auto& entry : partition_offsets) {
-                    PIntegerPair* res = part_offsets->add_offset_times();
-                    res->set_key(entry.key());
-                    res->set_val(entry.val());
+    DorisMetrics::instance()->get_info->increment(1);
+    _light_work_pool.offer([this, request, response, done]() {
```
morningman, line 496: This is a heavy work, it will visit kafka broker, which is a remote service.

be/src/service/internal_service.cpp
```diff
@@ -423,200 +480,233 @@ void PInternalServiceImpl::tablet_fetch_data(google::protobuf::RpcController* co
...
-    }
-    if (!st.ok()) {
-        LOG(WARNING) << "exec fold constant expr failed, errmsg=" << st;
-    }
-    st.to_protobuf(response->mutable_status());
+        Status st = Status::OK();
+        if (request->has_request()) {
+            st = _fold_constant_expr(request->request(), response);
+        } else {
+            // TODO(yangzhengguo) this is just for compatible with old version, this should be removed in the release 0.15
```
morningman, line 702: Remove this TODO?

be/src/service/internal_service.cpp
```diff
@@ -1097,9 +1208,8 @@ void PInternalServiceImpl::multiget_data(google::protobuf::RpcController* contro
                                          const PMultiGetRequest* request,
                                          PMultiGetResponse* response,
                                          google::protobuf::Closure* done) {
-    // Submit task to seperate ThreadPool for avoiding block bthread working pthread
-    ThreadPool* task_pool = StorageEngine::instance()->get_bg_multiget_threadpool();
-    Status submit_st = task_pool->submit_func([request, response, done, this]() {
+    DorisMetrics::instance()->multiget_data->increment(1);
+    _light_work_pool.offer([request, response, done, this]() {
```
morningman, line 1212: This is a heavy work, because it will open and read segment

morningman:
Please also update this document:
https://doris.apache.org/zh-CN/docs/dev/admin-manual/maint-monitor/monitor-metrics/metrics

be/src/util/doris_metrics.cpp
```diff
@@ -197,6 +230,37 @@ DorisMetrics::DorisMetrics() : _metric_registry(_s_registry_name) {
...
+    INT_COUNTER_METRIC_REGISTER(_server_metric_entity, tablet_writer_add_block_by_http);
+    INT_COUNTER_METRIC_REGISTER(_server_metric_entity, tablet_writer_cancel);
+    INT_COUNTER_METRIC_REGISTER(_server_metric_entity, get_info);
+    INT_COUNTER_METRIC_REGISTER(_server_metric_entity, update_cache);
+    INT_COUNTER_METRIC_REGISTER(_server_metric_entity, fetch_cache);
+    INT_COUNTER_METRIC_REGISTER(_server_metric_entity, clear_cache);
+    INT_COUNTER_METRIC_REGISTER(_server_metric_entity, merge_filter);
+    INT_COUNTER_METRIC_REGISTER(_server_metric_entity, apply_filter);
+    INT_COUNTER_METRIC_REGISTER(_server_metric_entity, transmit_block);
+    INT_COUNTER_METRIC_REGISTER(_server_metric_entity, transmit_block_by_http);
```
dataroaring, line 252: actually, there are these metrics in bvar in brpc already.

be/src/common/config.h
```diff
@@ -35,8 +35,10 @@ CONF_Int32(be_port, "9060");
 // port for brpc
 CONF_Int32(brpc_port, "8060");
 
-// the number of bthreads for brpc, the default value is set to -1, which means the number of bthreads is #cpu-cores
-CONF_Int32(brpc_num_threads, "-1");
+// the number of bthreads for brpc, the default value is set to 32
+// brpc only for network service send or accept request
+// no more  process any logic
+CONF_Int32(brpc_num_threads, "32");
```
dataroaring, line 41: We do not need to change default value here, e.g. if a user runs server on a machine with 128 cores, it would break his/her config used.

be/src/service/internal_service.cpp
```diff
@@ -423,200 +480,233 @@ void PInternalServiceImpl::tablet_fetch_data(google::protobuf::RpcController* co
...
                                        const PFetchCacheRequest* request, PFetchCacheResult* result,
                                        google::protobuf::Closure* done) {
-    brpc::ClosureGuard closure_guard(done);
-    _exec_env->result_cache()->fetch(request, result);
+    DorisMetrics::instance()->fetch_cache->increment(1);
+    _heavy_work_pool.offer([this, request, result, done]() {
+        brpc::ClosureGuard closure_guard(done);
+        _exec_env->result_cache()->fetch(request, result);
+    });
 }
```
dataroaring, line 573: If offer fails, what's the result of brpc in result ?

dataroaring (changes requested):

be/src/common/config.h
chenlinzhong: brpc now only for network  i/o, i think no need to configure so many threads

be/src/service/internal_service.cpp
chenlinzhong: ok

be/src/service/internal_service.cpp
chenlinzhong: ok

be/src/service/internal_service.cpp
chenlinzhong: ok

be/src/service/internal_service.cpp
chenlinzhong: ok

be/src/service/internal_service.cpp
```diff
@@ -423,200 +480,233 @@ void PInternalServiceImpl::tablet_fetch_data(google::protobuf::RpcController* co
                                              const PTabletKeyLookupRequest* request,
                                              PTabletKeyLookupResponse* response,
                                              google::protobuf::Closure* done) {
-    [[maybe_unused]] brpc::Controller* cntl = static_cast<brpc::Controller*>(controller);
-    brpc::ClosureGuard guard(done);
-    Status st = _tablet_fetch_data(request, response);
-    st.to_protobuf(response->mutable_status());
+    DorisMetrics::instance()->tablet_fetch_data->increment(1);
+    _heavy_work_pool.offer([this, controller, request, response, done]() {
```
eldenmoon, line 484: please Do not put this RPC interface into thread pool, it's very light weight and can't be mixed with other rpc interface, since it's latency sensitive

be/src/service/internal_service.cpp
chenlinzhong: if fails, just means rpc service is shoutdown otherwise will block until the request handled

be/src/service/internal_service.cpp
chenlinzhong: before this pr ,this rpc interface also  mixed with each other 
i don't think put into thread pool will effect latency

eldenmoon: if the thread pool in hight CPU work load, this interface will be queued up and the latency is infected

eldenmoon: Other heavy tasks could block all pthread and no worker thread for this light weight rpc and infect latency

chenlinzhong: Even if I don't add this pr, will have the same problem

be/src/common/config.h
dataroaring: brpc creates a pthread for each cpu, this is a default behavior, and we do not care it on any machine if we use it in a right way. I can not see any benefits on changing it.

be/src/service/internal_service.cpp
dataroaring: Do you  test this case like mock a offer failure?  In the case, we do not set status, so i do not how brpc handle it.

be/src/util/doris_metrics.cpp
chenlinzhong: Provide two metrics interfaces for the same node
- http://ip:http_port/metrics
- http://ip:brpc_port/brpc_metrics

It is not very reasonable from the perspective of use. I think it is reasonable and acceptable to have redundant data here, and it does not affect performance

chenlinzhong:
> Please also update this document: https://doris.apache.org/zh-CN/docs/dev/admin-manual/maint-monitor/monitor-metrics/metrics

resolved

yinzhijian (dismissed):
LGTM

chenlinzhong:
run feut

chenlinzhong:
run beut

chenlinzhong:
run p0

chenlinzhong:
run beut

chenlinzhong:
run p0

chenlinzhong:
run p0

chenlinzhong:
run p0

chenlinzhong:
run p0

chenlinzhong:
run p0

dataroaring (approved):
LGTM

chenlinzhong:
run p0

chenlinzhong:
run p0

chenlinzhong:
run p0

chenlinzhong:
run p0

chenlinzhong:
run p0

PR #16644, opened 2023-02-12
[Bug](cooldown) set new replica id when early exit in doing clone when no missed versions

ByteYue:
# Proposed changes

Issue Number: close #xxx
When doing clone task, the logic would early exit if the missed version is empty. But the tablet meta is not set with the newly replica id inside the clone request. It would result inconsistent cooldown replica id when doing cooldown.

## Problem summary

Describe your changes.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/task/engine_clone_task.cpp
+++ b/be/src/olap/task/engine_clone_task.cpp
@@ -98,6 +98,12 @@ Status EngineCloneTask::_do_clone() {
         if (missed_versions.empty()) {
             LOG(INFO) << "missed version size = 0, skip clone and return success. tablet_id="
                       << _clone_req.tablet_id << " req replica=" << _clone_req.replica_id;
+            // update replica id to meet cooldown replica
+            {
+                std::lock_guard<std::shared_mutex> wrlock(tablet->get_header_lock());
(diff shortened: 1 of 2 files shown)


dataroaring (dismissed):
LGTM

be/src/olap/task/engine_clone_task.cpp
```diff
@@ -98,6 +98,12 @@ Status EngineCloneTask::_do_clone() {
         if (missed_versions.empty()) {
             LOG(INFO) << "missed version size = 0, skip clone and return success. tablet_id="
                       << _clone_req.tablet_id << " req replica=" << _clone_req.replica_id;
+            // update replica id to meet cooldown replica
```
platoneko, line 101: Seems only `if (_clone_req.replica_id != tablet->replica_id())` should reset replica id.

ByteYue: I thought each clone task had one new replica id. It seems VERSION_INCOMPLETE task would has the same replica id. I'll go fix it.

dataroaring (approved):
LGTM

platoneko:
LGTM

PR #16670, opened 2023-02-13
[Fix](cpp17) fix gutil unary_function/binary_function for cpp17

JackDrogon:
# Proposed changes

be gutil contains unary_function/binary_function, but they are removed in C++17. Fix these for gcc12 to compile be.

image


BiteTheDDDDt (approved):
LGTM

PR #16679, opened 2023-02-13
[Refactor](auth) Add AccessController to support customized authorization

morningman:
# Proposed changes

Issue Number: #16762

## Problem summary

In current implementation, the class `Auth` is used for:
1. Manager all authentication and authorization info such as user, role, password, privileges.
2. Provide an interface for privilege checking

Some user may want to integrate external access management system such as Apache Ranger.
So we should provide a way to let user set their own access controller.

This PR mainly changes:

1. A new class `SystemAccessController`
    This access controller is used to check the global level privileges and resource privileges.

2. A new interface `CatalogAccessController`
    This interface is used to check catalog/database/tbl level privileges.
    It has a default implements `InternalCatalogAccessController`.

    All privilege checking methods are moved from `Auth` to either `SystemAccessController` or 
    `InternalCatalogAccessController`

3. A new class `AccessControllerManager`
    This is the entry point of privilege authentication. All methods previously called from `Auth`
    now are called from `AccessControllerManager`

Now, user can implement the interface `CatalogAccessController` to use their own access controller.
And when creating external catalog, user can specified the access controller class name, so that
different external catalog can use different access controller.

## Checklist(Required)

## Further comments


hello-stephen:
 the sum of best hot time: 34.36 seconds
 stream load tsv:          501 seconds loaded 74807831229 Bytes, about 142 MB/s
 stream load orc:          69 seconds loaded 1101869774 Bytes, about 15 MB/s
 stream load parquet:          28 seconds loaded 861443392 Bytes, about 29 MB/s
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20230213173121_clickbench_pr_95660.html

fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/AccessControllerManager.java
```diff
@@ -0,0 +1,161 @@
...
+
+    // ==== Catalog ====
+    public boolean checkCtlPriv(ConnectContext ctx, String ctl, PrivPredicate wanted) {
+        return checkCtlPriv(ctx.getCurrentUserIdentity(), ctl, wanted);
+    }
+
+    public boolean checkCtlPriv(UserIdentity currentUser, String ctl, PrivPredicate wanted) {
+        boolean hasGlobal = sysAccessController.checkGlobalPriv(currentUser, wanted);
+        boolean hasCtl = getAccessControllerOrDefault(ctl).checkCtlPriv(hasGlobal, currentUser, ctl, wanted);
+        return hasGlobal || hasCtl;
```
zddr, line 79: logic is repeated with 'CatalogAccessController.checkCtlPriv(boolean hasGlobal, UserIdentity currentUser, String ctl, PrivPredicate wanted)',here,can we use 'CatalogAccessController.checkCtlPriv(UserIdentity currentUser, String ctl, PrivPredicate wanted)'? and delete 'CatalogAccessController.checkCtlPriv(boolean hasGlobal, UserIdentity currentUser, String ctl, PrivPredicate wanted)'?

fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/AccessControllerManager.java
```diff
@@ -0,0 +1,161 @@
...
+        return checkTblPriv(ctx, tableName.getCtl(), tableName.getDb(), tableName.getTbl(), wanted);
+    }
+
+    public boolean checkTblPriv(UserIdentity currentUser, String db, String tbl, PrivPredicate wanted) {
+        return checkTblPriv(currentUser, Auth.DEFAULT_CATALOG, db, tbl, wanted);
+    }
+
+    public boolean checkTblPriv(UserIdentity currentUser, String ctl, String db, String tbl, PrivPredicate wanted) {
+        boolean hasGlobal = sysAccessController.checkGlobalPriv(currentUser, wanted);
+        boolean hasTbl = getAccessControllerOrDefault(ctl).checkTblPriv(hasGlobal, currentUser, ctl, db, tbl, wanted);
```
zddr, line 122: same as 'checkCtlPriv'

fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/AccessControllerManager.java
morningman: No exactly,
`CatalogAccessController` can be customized, and some implementation need to audit all request.
So we need to make sure that `CatalogAccessController` is called for each request.
And `sysAccessController` is used to check global level privilege(which is not included in `CatalogAccessController`)

zddr:
If a catalog uses range authentication, what should we do when the user uses' grant 'to authorize the table under the catalog?

morningman:
> If a catalog uses range authentication, what should we do when the user uses' grant 'to authorize the table under the catalog?


When checking the privilege of this catalog, it will skip InternalAccessController, only use user specified RangeAccessController.

yiguolei (approved):
LGTM

PR #16758, opened 2023-02-14
[improve](inverted index) Add element count limit for inverted index searcher cache

Tanya-W:
# Proposed changes

Issue Number: close #xxx

## Problem summary
The element in InvertedIndexSearcherCache is inverted index searcher, which is a file descriptor of inverted index file, so `InvertedIndexSearcherCache` is actually cache file descriptor of inverted index file.

If open file descriptor limit of the Linux system is set too small and config `inverted_index_searcher_cache_limit`  is too big, during high pressure load maybe cause "Too many open files".

So, when insert inverted index searcher into `InvertedIndexSearcherCache`, need also check whether reach file_descriptor_number limit for inverted index file.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/lru_cache.h
+++ b/be/src/olap/lru_cache.h
@@ -289,6 +289,8 @@ public:
     // Return whether h is found and removed.
     bool remove(const LRUHandle* h);
 
+    uint32_t element_count() const;
+
 private:
     FRIEND_TEST(CacheTest, HandleTableTest);
 
@@ -322,6 +324,9 @@ public:
 
     // Separate from constructor so caller can easily make an array of LRUCache
     void set_capacity(size_t capacity) { _capacity = capacity; }
+    void set_element_count_limit(uint32_t element_count_limit) {
+        _element_count_limit = element_count_limit;
+    }
 
     // Like Cache methods, but with an extra "hash" parameter.
     Cache::Handle* insert(const CacheKey& key, uint32_t hash, void* value, size_t charge,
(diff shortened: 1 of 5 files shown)


be/src/olap/rowset/segment_v2/inverted_index_cache.cpp
```diff
@@ -49,6 +49,19 @@ void InvertedIndexSearcherCache::create_global_instance(size_t capacity, uint32_
...
+    struct rlimit l;
+    int ret = getrlimit(RLIMIT_NOFILE, &l);
+    if (ret != 0) {
+        LOG(WARNING) << "call getrlimit() failed. errno=" << strerror(errno)
+                     << ", use default configuration instead.";
+    } else {
+        fd_number = static_cast<uint64_t>(l.rlim_cur);
+    }
+
+    uint64_t open_searcher_limit = fd_number / 3 * 2;
```
xiaokang, line 62: It's better to use a dedicated config, and check it's less than min_file_descriptor_number.

be/src/olap/lru_cache.h
```diff
@@ -322,6 +324,9 @@ class LRUCache {
 
     // Separate from constructor so caller can easily make an array of LRUCache
     void set_capacity(size_t capacity) { _capacity = capacity; }
+    void set_element_count_limit(uint32_t element_count_limit) {
+        _element_count_limit = element_count_limit;
```
xiaokang, line 328: suggest a name _count_capacity

xiaokang (commented):
add some unit test

be/src/olap/lru_cache.h
Tanya-W: use _element_count_capacity instead of it

be/src/olap/rowset/segment_v2/inverted_index_cache.cpp
Tanya-W: The actual file descriptor number maybe much larger than config:: min_file_descriptor_number, so it should be limited by the actual file descriptor number as much as possible.

xiaokang (approved):
LGTM

qidaye (approved):
LGTM

PR #16928, opened 2023-02-20
[fix](tvf) fix bug that failed to get schema of tvf when file is empty

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem summary

In previous implementation, when querying tvf, FE will get schema from BE.
And BE will try to open the first file to get its schema info, but for orc or parquet format,
if the file is empty, it will return error.
But even for an empty file, we can still get schema info from file's footer.
So we should handle the empty file to get schema info correctly.

Also modify the catalog doc to add some FAQ.

## Checklist(Required)

## Further comments


morningman:
run buildall

yiguolei (approved):
LGTM

PR #16680, opened 2023-02-13
[improvement](rowset reader) fix possible memleak

jacktengg:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Fix possible memory leaks when constructing betarowset reader, by using unique_ptrs.

## Checklist(Required)

* [ ] Does it affect the original behavior
* [ ] Has unit tests been added
* [ ] Has document been added or modified
* [ ] Does it need to update dependencies
* [ ] Is this PR support rollback (If NO, please explain WHY)

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #16938, opened 2023-02-20
[fix](union iterator) fix bug that result data order of VUnionIterator is different

jacktengg:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Fix bug of #16680, data order of VUnionIterator outout block is changed, which will impact compaction.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/vec/olap/vgeneric_iterators.cpp
+++ b/be/src/vec/olap/vgeneric_iterators.cpp
@@ -374,6 +374,11 @@ Status VUnionIterator::init(const StorageReadOptions& opts) {
         return Status::OK();
     }
 
+    // we use back() and pop_back() of std::vector to handle each iterator,
+    // so reverse the vector here to keep result block of next_batch to be
+    // in the same order as the original segments.
+    std::reverse(_origin_iters.begin(), _origin_iters.end());
+
     for (auto& iter : _origin_iters) {
         RETURN_IF_ERROR(iter->init(opts));
     }
--- a/be/test/vec/exec/vgeneric_iterators_test.cpp
+++ b/be/test/vec/exec/vgeneric_iterators_test.cpp
@@ -115,17 +115,19 @@ TEST(VGenericIteratorsTest, Union) {
     auto c1 = block.get_by_position(1).column;
     auto c2 = block.get_by_position(2).column;
 
+    size_t row_count = 0;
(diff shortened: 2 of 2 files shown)


jacktengg:
run buildall

liaoxin01 (dismissed):
LGTM

jacktengg:
run p0

jacktengg:
run beut

jacktengg:
run p0

jacktengg:
run p0

be/src/vec/olap/vgeneric_iterators.cpp
```diff
@@ -374,6 +374,11 @@ Status VUnionIterator::init(const StorageReadOptions& opts) {
         return Status::OK();
     }
 
+    // we use back() and pop_back() of std::vector to handle each iterator,
```
zhannngchen, line 377: we'd better to make iterator's order right outside of the UnionIterator?
It's tricky to do this kind of operation inside UnionIterator, if some other developers changed the iterator's order outside of UnionIterator for some reason, the result will not be wrong.
Consider to do this operation at `beta_rowset_reader.cpp:L218`, just before `  _iterator = vectorized::new_union_iterator(std::move(iterators));`, and we should add a DCHECK on `iter.data_id()` to check that the iterators order right.

jacktengg: As I understand, the VUnionIterator's behaviour is to output data from the list of segment iterators, from begin to end.

The user of VUnionIterator is reponsible of the corrrect order of the list of segment iterators, and VUnionIterator should keep the order of output data is the same as the input order of list of segment iterators. 

As in the code comment, we use back() and pop_back() of std::vector to handle each iterator, so reverse the vector here to keep result block of next_batch to be in the same order as the original segments.

zhannngchen: My bad, I understand the context wrong at the beginning. I thought why you need to reverse the iterator here because the input is reversed, so I suggested that we should adjust the input.

jacktengg:
run beut

jacktengg:
run feut

yiguolei (approved):
LGTM

zhannngchen (approved):
LGTM

PR #16957, opened 2023-02-20
[fix](auth) fix losing global priv bug and refactor default role name

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem summary

This PR mainly changes:

1. When upgrading from old version to master, the ADMIN_PRIV for normal user may be lost.
    This may only happen if:
    1. Create a user with ADMIN_PRIV privilege.
    2. Upgrade Doris to v1.2.x or master before the meta image which contains the edit log in step 1 is generate.
    And the ADMIN_PRIV will be lost in Global Privileges

   This PR will rectify this bug and set ADMIN_PRIV to the right place

2. Refactor the user's implicit role name

    In #16091, we refactor the Doris auth model by introducing RBAC. And each user will have an implicit role,
    named with prefix `default_role_rbac_`. But it has wrong format like:
    `default_role_rbac_'default_cluster:user1'@'%'`

    This PR change the role name's format, like:
    ```
    default_role_rbac_user1@%
    default_role_rbac_user2@[domain]
    ```

    NOTICE: this change may cause incompatible metadata, but since #16091 is not released, we should fix it soon.

3. Add a new session variable `show_user_default_role`

    When set to true, it will show implicit role of user in the result of `show roles` stmt. Default is false

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/nereids/glue/translator/PhysicalPlanTranslator.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/nereids/glue/translator/PhysicalPlanTranslator.java
@@ -531,8 +531,12 @@ public class PhysicalPlanTranslator extends DefaultPlanVisitor<PlanFragment, Pla
                 .build();
         TupleDescriptor tupleDescriptor = generateTupleDesc(slotList, table, context);
         tupleDescriptor.setTable(table);
-
         SchemaScanNode scanNode = new SchemaScanNode(context.nextPlanNodeId(), tupleDescriptor);
+        context.getRuntimeTranslator().ifPresent(
+                runtimeFilterGenerator -> runtimeFilterGenerator.getTargetOnScanNode(schemaScan.getId()).forEach(
+                    expr -> runtimeFilterGenerator.translateRuntimeFilterTarget(expr, scanNode, context)
+            )
+        );
         scanNode.finalizeForNereids();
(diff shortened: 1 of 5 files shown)


hello-stephen:
 the sum of best hot time: 34.91 seconds
 stream load tsv:          472 seconds loaded 74807831229 Bytes, about 151 MB/s
 stream load json:         38 seconds loaded 2358488459 Bytes, about 59 MB/s
 stream load orc:          67 seconds loaded 1101869774 Bytes, about 15 MB/s
 stream load parquet:          29 seconds loaded 861443392 Bytes, about 28 MB/s
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20230220155946_clickbench_pr_100446.html

fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/RoleManager.java
```diff
@@ -181,17 +183,27 @@ public Role createDefaultRole(UserIdentity userIdent) throws DdlException {
     }
 
     public Role removeDefaultRole(UserIdentity userIdent) {
-        return roles.remove(getUserDefaultRoleName(userIdent));
+        Role role = roles.remove(getUserDefaultRoleName(userIdent));
+        if (role == null) {
```
zddr, line 187: Repetitive logic

PR #16979, opened 2023-02-21
[fix](decimal) change log fatal to log warning to avoid code dump on decimal type

liaoxin01:
# Proposed changes

Issue Number: close #xxx

## Problem summary

cherry-pick from: #14150

## Checklist(Required)

## Further comments


dataroaring (approved):
LGTM

yiguolei:
buildall

yiguolei:
I think we should use log fatal here, because it will return a wrong result when decimal overflow. If user is using insert into select to do ETL jobs, then data is wrong.

dataroaring (changes requested):
add a config to enable LOG(FATAL) and FATAL by default.

PR #17149, opened 2023-02-26
[fix](sort) fix bug of sort

jacktengg:
# Proposed changes

Issue Number: close #xxx

## Problem summary

The logic of topn and full sort is wrong when there are both offsets and limits, the offset is not considered when doing the max heap optimization, which will lead to wrong result. 

cherry-pick #17151

## Checklist(Required)

## Further comments


Gabriel39 (approved):
LGTM

PR #17182, opened 2023-02-27
[bugfix](TabletSchedule) remove unalive replica first when FORCE_REDUNDANT

yixiutt:
Case like this:
we have 4 be, decommission 1 and than drop one, then this tablet will set FORCE_REDUNDANT in as no enough backend for us to create a new replica, and we will drop replica unScheduleAble, so the decommisioned replica may be dropped and then lead to only one replica alive in tablet. We should drop unalive replica first and wait clone enough replica and then drop decommission replica.

# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

## Further comments


dataroaring:
run buildall

PR #17331, opened 2023-03-02
[Fix](Lightweight schema Change) query error caused by array default type is unsupported

ZhangYu0123:
We have supportted array type default [], but when using lightweight schema Change to add column array type, query failed as follows:

image

1. Fix "array default type is unsupported" error.
2. Fix the default value filling assignment digit problem.

# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_v2/column_reader.cpp
+++ b/be/src/olap/rowset/segment_v2/column_reader.cpp
@@ -1117,8 +1117,16 @@ Status DefaultValueColumnIterator::init(const ColumnIteratorOptions& opts) {
                 ((Slice*)_mem_value.data())->size = _default_value.length();
                 ((Slice*)_mem_value.data())->data = _default_value.data();
             } else if (_type_info->type() == OLAP_FIELD_TYPE_ARRAY) {
-                // TODO llj for Array default value
-                return Status::NotSupported("Array default type is unsupported");
+                if (_default_value != "[]") {
+                    return Status::NotSupported("Array default {} is unsupported", _default_value);
+                } else {
+                    ((Slice*)_mem_value.data())->size = _default_value.length();
+                    ((Slice*)_mem_value.data())->data = _default_value.data();
+                }
+            } else if (_type_info->type() == OLAP_FIELD_TYPE_STRUCT) {
+                return Status::NotSupported("STRUCT default type is unsupported");
+            } else if (_type_info->type() == OLAP_FIELD_TYPE_MAP) {
+                return Status::NotSupported("MAP default type is unsupported");
             } else {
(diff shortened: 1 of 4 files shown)


ZhangYu0123:
run buildall

yiguolei:
Please add some regression test.

ZhangYu0123:
run buildall

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -1186,14 +1204,20 @@ void DefaultValueColumnIterator::insert_default_data(const TypeInfo* type_info,
...
-        data_ptr = ((Slice*)mem_value)->data;
-        data_len = ((Slice*)mem_value)->size;
+        char* data_ptr = ((Slice*)mem_value)->data;
+        size_t data_len = ((Slice*)mem_value)->size;
         dst->insert_many_data(data_ptr, data_len, n);
         break;
     }
+    case OLAP_FIELD_TYPE_ARRAY: {
+        for (int i = 0; i < n; i++) {
+            dst->insert(doris::vectorized::Array(0));
```
eldenmoon, line 1214: why not just use dst->insert_many_default?

ZhangYu0123: Because  columnNullable will make null value as default.   So  in this function is used to process the default  value  situation.   
image

ZhangYu0123: > why not just use dst->insert_many_default?

Because columnNullable will make null value as default. So in this function is used to process the default value situation.
image

ZhangYu0123:
run buildall

ZhangYu0123:
run buildall

be/src/olap/rowset/segment_v2/column_reader.cpp
eldenmoon: you could get the nested type of ColumnNullable and call `insert_many_default` and assign the null map of ColumnNullable with may `0`, the could be much faster

ZhangYu0123: > you could get the nested type of ColumnNullable and call `insert_many_default` and assign the null map of ColumnNullable with may `0`, the could be much faster

ok

ZhangYu0123:
run buildall

ZhangYu0123:
run p1

eldenmoon (approved):
LGTM

yiguolei (approved):
LGTM

PR #14995, opened 2022-12-11
[fix](merge-on-write) delete all rows with same key in all pre segments

nextdreamblue:
# Proposed changes

Issue Number: close #14994

## Problem summary

Describe your changes.

## Checklist(Required)

1. Does it affect the original behavior: 
    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [x] No Need
3. Has document been added or modified:
    - [ ] Yes
    - [ ] No
    - [x] No Need
4. Does it need to update dependencies:
    - [ ] Yes
    - [x] No
5. Are there any changes that cannot be rolled back:
    - [ ] Yes (If Yes, please explain WHY)
    - [x] No

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #17334, opened 2023-03-02
[fix](merge-on-write) fix that delete bitmap is not calculated correctly when clone tablet

liaoxin01:
# Proposed changes

Issue Number: close #xxx

## Problem summary

1. Because of specified_rowset_ids is nullptr, the delete bitmap is not be recalculated when clone, introduced by #14995.  
2. Optimize delete bitmap calculation of full clone, by clone the delete bitmap of remote tablet.
3. Change the set operation of delete bitmap into a merge operation.
4. Delete some useless code.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/tablet.cpp
+++ b/be/src/olap/tablet.cpp
@@ -195,17 +195,55 @@ void Tablet::save_meta() {
 }
 
 Status Tablet::revise_tablet_meta(const std::vector<RowsetSharedPtr>& to_add,
-                                  const std::vector<RowsetSharedPtr>& to_delete) {
+                                  const std::vector<RowsetSharedPtr>& to_delete,
+                                  bool is_incremental_clone) {
     LOG(INFO) << "begin to revise tablet. tablet_id=" << tablet_id();
     delete_rowsets(to_delete, false);
     add_rowsets(to_add);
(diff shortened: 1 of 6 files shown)


liaoxin01:
run buildall

dataroaring (approved):
LGTM

be/src/olap/tablet.cpp
```diff
@@ -195,17 +195,55 @@ void Tablet::save_meta() {
...
+            calc_delete_bitmap_ver = Version(to_add_max_version + 1, max_version().second);
+        }
+        Status res =
+                capture_consistent_rowsets(calc_delete_bitmap_ver, &calc_delete_bitmap_rowsets);
+        // Because the data in memory has been changed, can't return an error.
+        CHECK(res.ok()) << "fail to capture_consistent_rowsets, res: " << res;
+
+        for (auto rs : calc_delete_bitmap_rowsets) {
+            res = update_delete_bitmap_without_lock(rs);
+            CHECK(res.ok()) << "fail to update_delete_bitmap_without_lock, res: " << res;
```
zhannngchen, line 246: Clone is not the critical path, we'd better use DCHECK here?

liaoxin01: Because the data in memory has changed, if do not use CHECK here, it may cause unexpected errors.

PR #17551, opened 2023-03-08
[fix](insert) Session varaiables dont work for transaction insert

1330571:
# Proposed changes

Issue Number: close #xxx

Add more fields defined in thrift but not used in `beginTxn` function

## Problem summary

We can set variables such as query_timeout, mem_exec_limit. But it did not post those parameters from FE to BE properly, which may lead to an unexpected result. After setting those fields in functino `beginTxn` , BE can receive these parameters and work under provided configuration.

## Checklist(Required)

* [ ] Have unit tests been added
* [X] Is this PR support rollback (If NO, please explain WHY)

### MySQL

Execute those commands in mysql CLI to set the exec_mem_limit variable

```
mysql> SET exec_mem_limit = 10000000;

mysql> begin;
mysql> insert into example_tbl values (1,1);
mysql> insert into example_tbl values (2,2);
mysql> commit;
```

### Be Log 

and check BE log to see if it does set the mem limit

```
fragment_mgr.cpp:719] Register query/load memory tracker, query/load id: b2c4bfe09f0f4c03-925271683c000d53 limit: 9.54 MB
```

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
@@ -1491,7 +1491,7 @@ public class StmtExecutor implements ProfileWriter {
         TransactionEntry txnEntry = context.getTxnEntry();
         TTxnParams txnConf = txnEntry.getTxnConf();
         SessionVariable sessionVariable = ConnectContext.get().getSessionVariable();
-        int timeoutSecond = sessionVariable.getQueryTimeoutS();
+        long timeoutSecond = ConnectContext.get().getExecTimeout();
 
         TransactionState.LoadJobSourceType sourceType = TransactionState.LoadJobSourceType.INSERT_STREAMING;
         Database dbObj = Env.getCurrentInternalCatalog()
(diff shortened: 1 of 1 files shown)


1330571:
run buildall

1330571:
run buildall

mymeiyi:
Is it better to rename this PR to [fix](insert) Session varaiables donot work for transaction insert?

1330571:
> Is it better to rename this PR to
> 
> ```
> [fix](insert) Session varaiables donot work for transaction insert?
> ```

done

fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
```diff
@@ -1486,7 +1486,9 @@ private void beginTxn(String dbName, String tblName) throws UserException, TExce
             InterruptedException, ExecutionException, TimeoutException {
         TransactionEntry txnEntry = context.getTxnEntry();
         TTxnParams txnConf = txnEntry.getTxnConf();
-        long timeoutSecond = ConnectContext.get().getExecTimeout();
+        SessionVariable sessionVariable = ConnectContext.get().getSessionVariable();
+        int timeoutSecond = sessionVariable.getQueryTimeoutS();
```
mymeiyi, line 1490: timeout should keep the original logic, because getExecTimeout may be query_timeout or insert_timeout

1330571: resolved with a new commit, now I use `ConnectContext.get().getExecTimeout()` to handle timeout

1330571:
run buildall

1330571:
run buildall

1330571:
run buildall

1330571:
run compile

dataroaring (approved):
LGTM

PR #17578, opened 2023-03-08
[improvement](config)add task pool config

tarepanda1024:
# Proposed changes

add corePoolSize and keepAliveTime config for task pool.

## Checklist(Required)

* [x] Does it affect the original behavior. 
       The default task pool coreSize is 0 , current is 8.

## Further comments

### DIFF
--- a/docs/en/docs/admin-manual/config/fe-config.md
+++ b/docs/en/docs/admin-manual/config/fe-config.md
@@ -469,11 +469,23 @@ Default：1G
 
 Used to set the initial flow window size of the GRPC client channel, and also used to max message size.  When the result set is large, you may need to increase this value.
 
+#### `core_mysql_service_task_threads_num`
+
+Default：8
+
+the number of threads to keep in the pool, even if they are idle, unless allowCoreThreadTimeOut is set.
+
 #### `max_mysql_service_task_threads_num`
 
 Default：4096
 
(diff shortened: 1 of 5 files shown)


morningman:
Why need this?

tarepanda1024:
> Why need this?

When there are small querys：each query cost about 10ms ,create task thead may cost too long, and the gap  time not in audit.log .After thead pool warm up, the  99% latency become better.

```sql
select count(1) from (select xx as user_id) as M left join dm_assets.xxx as Q_0 on M.user_id=Q_0.user_id and Q_0.user_id = xxx left join dm_assets.xxx  as Q_1 on M.user_id=Q_1.user_id and Q_1.user_id = xxx where (not((Q_0.order_confirm_time <= xx and Q_0.order_confirm_time >= xxx AND  array_str_op(Q_0.item_id, 'in', 'xxx'))) AND ( Q_1.xx not in ('1') OR Q_1.xx IS NULL ) )
```

Trace from mysql jdbc client.The gap about 200ms+，but in audit log，the real exec time is about 10ms.

Trace from client:
image

The business log:
image

The audit log：
image
query result is ERR and cost 5ms, because query timeout and canceled.

From client：215ms，but in doris，cost 5ms and canceled by timeout.the gap in task thread pool.
I changed xnio IO thread num，affect little.And big flow,the task pool warm up and latency become better.

Trace from doris fe(not match above query)：
image
image
image
image
the audit log not calc : readEvent to query dispatch cost.

bench mark :

image
the blank is being benchmark

image

image

flow slow down, the latency up.

PR #17623, opened 2023-03-09
[config](balance) set schedule_slot_num_per_path to 8

yixiutt:
When upgrade doris be, we found that recover is too slow and disk io is low, so we change schedule_slot_num_per_path to 8 and our HDD disk io max to 50% and it's acceptable

# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

## Further comments


morningman (changes requested):
Better change config file of user itself

yixiutt:
> Better change config file of user itself

I think it's better to make default value bigger so balance&recovery can run faster, most user don't know real reason and this would makes some replica version incomplete and then writes failed with -235

PR #17720, opened 2023-03-13
[ehancement](profile) Disable profiling for the internal query

Kikyou1997:
# Proposed changes

Once user set  `enable_profile` globally, the internal queries's profile would be cached and display on the web page,such like for sqls for statistics collection which would be annoying for users. 

Add some condition for write/update profile to avoid this.

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/QeProcessorImpl.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/QeProcessorImpl.java
@@ -187,7 +187,8 @@ public final class QeProcessorImpl implements QeProcessor {
         }
         final TReportExecStatusResult result = new TReportExecStatusResult();
         final QueryInfo info = coordinatorMap.get(params.query_id);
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
```diff
@@ -702,6 +702,10 @@ private void forwardToMaster() throws Exception {
 
     @Override
     public void writeProfile(boolean isLastWriteProfile) {
+        if (context.getSessionVariable() != null
```
yiguolei, line 705: better to set enable_profile = false in internal query context.

Kikyou1997: done

yiguolei (approved):
LGTM

yiguolei:
run buildall

morrySnow (approved):

PR #17750, opened 2023-03-13
[fix](join) fix wrong result of null aware left anti join

jacktengg:
# Proposed changes

Issue Number: close #xxx

## Problem summary

When enable shared hash table for broadcast, if null value of probe side appears in fragment instance that DID NOT build hash table, `_probe_ignore_null` is not set to true and null value of probe side is treated as normal data, which result in wrong result. (regression test case test_null_aware_left_anti_join.groovy)

pick from https://github.com/apache/doris/pull/17752

## Checklist(Required)

## Further comments


Gabriel39 (approved):
LGTM

PR #17829, opened 2023-03-15
[Fix](Planner)Add call once logic to analyze of function aes_decrypt

LiBinfeng-01:
# Proposed changes

Add call once to adding third parameter of aes_decrypt type function.

## Problem summary

The problem is an exception when doing analyze:
java.lang.IllegalStateException: exceptions :
errCode = 2, detailMessage = select list expression not produced by aggregation output (missing from GROUP BY clause?): xxx

The scenario is: 
select aes_decrypt(xxx,xxx) as c0 from table group by c0;

Analyze of problem:
The direct problem is mismatched of slotref, and this mismatched due to the mismatched of parameter number of aes_decrypt function. When debuging, we can see the slotref of group column is added to ExprSubstitutionMap, but can not matching with select result columns. And this is because when substiting expr it will analyze again, so the parameter would be added twice. This will cause the mismatching of function, so it would not be substitute as a slotref, the exception would be throw.

Fix:
Add call once to adding third parameter of aes_decrypt type function. Compare the child we want to add to the last child of function. If they are the same, do not add it.

Describe your changes.

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/FunctionCallExpr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/FunctionCallExpr.java
@@ -64,6 +64,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 // TODO: for aggregations, we need to unify the code paths for builtins and UDAs.
 public class FunctionCallExpr extends Expr {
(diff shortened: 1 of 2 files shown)


Bingandbing:
run buildall

xinyiZzz (approved):
LGTM

hello-stephen:
 the sum of best hot time: 34.91 seconds
 stream load tsv:          470 seconds loaded 74807831229 Bytes, about 151 MB/s
 stream load orc:          75 seconds loaded 1101869774 Bytes, about 14 MB/s
 stream load parquet:          32 seconds loaded 861443392 Bytes, about 25 MB/s
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20230315125805_clickbench_pr_114738.html

PR #17830, opened 2023-03-15
[Optimize](functions)Reconstruct basic process for default function implementation.

zclllyybb:
# Proposed changes

## Problem summary

The constant/nullable deal logic relies on weird recursion. And to avoid infinite recursion, we need an unequal logic for const columns. Now we don't anymore.
After, all const-constraint **could and must** be dealt with in `get_arguments_that_are_always_constant`.

## Checklist(Required)

## Further comments

need to fix scattered checks in functions.

### DIFF
--- a/be/src/vec/functions/function.cpp
+++ b/be/src/vec/functions/function.cpp
@@ -128,8 +128,8 @@ NullPresence get_null_presence(const Block& block, const ColumnNumbers& args) {
     return res;
 }
 
-bool allArgumentsAreConstants(const Block& block, const ColumnNumbers& args) {
-    for (auto arg : args) {
+bool all_arguments_are_constant(const Block& block, const ColumnNumbers& args) {
+    for (const auto& arg : args) {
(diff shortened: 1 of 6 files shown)


HappenLee:
run buildall

be/src/vec/functions/function.cpp
```diff
@@ -138,53 +138,51 @@ bool allArgumentsAreConstants(const Block& block, const ColumnNumbers& args) {
...
+                                           arg_num, get_name());
         }
     }
 
     if (args.empty() || !use_default_implementation_for_constants() ||
         !allArgumentsAreConstants(block, args)) {
         return Status::OK();
     }
-
+    // now all columns is const.
```
HappenLee, line 177: change function `allArgumentsAreConstants` to `all_arguments_are_const`

HappenLee (approved):
LGTM

HappenLee (approved):
LGTM

Issue #17413, opened 2023-03-04
[Feature] Refresh catalog' metadata at fix interval.


morningman:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

## Background

Doris supports [multi catalog ](https://doris.apache.org/docs/dev/lakehouse/multi-catalog/) to connect to external datasource.
For any catalog, Doris will sync the metadata, such as database, table, schema, partition, file from external datasource
into FE's memory, so that when accessing the catalog, we don' t need to access the remote data source' meta service.

But it is a one-time operation, after sync, if the external data source change its metadata, such as adding new table, partition,
Doris can't be aware of it and the newly created metadata can not be accessed from Doris.

So we need a way refresh the metadata of the catalog, to sync the metadata at fix interval.

## How to

First, we should set new properties of catalog when creating it, eg:
```
create catalog hive properties (
...
"metadata_refresh_interval_sec" = "7200"
...
);
```

which means for this `hive` catalog, its metadata will be refreshed every 7200s.
And there will be a daemon thread on Master FE to do this for all catalogs.

## Tips

1. Related classes: See `makeSureInitialized()` of `ExternalCatalog` class to learn how to refresh catalog.
2. See [Refresh](https://doris.apache.org/docs/dev/sql-manual/sql-reference/Utility-Statements/REFRESH) command to learn how to refresh catalog manually.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #17884, opened 2023-03-17
[Feature](multi-catalog) add refresh for catalog if it needs

q763562998:
# Proposed changes

Issue Number: close[[#17413](https://github.com/apache/doris/issues/17413)](https://github.com/apache/doris/issues/17413)

## Problem summary

Currently if a catalog is modified externally in doris, doris is not dynamically aware of it.So if a catalog is created with a refresh time configuration, I added a timer for it to refresh the catalog regularly.

## Checklist(Required)

   -  Yes
2. Has unit tests been added:
   -  No
   -  Yes
   -  No
   -  No

## Further comments

If this is a relatively large or complex change, kick off the discussion at [[dev@doris.apache.org](mailto:dev@doris.apache.org)](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...


fe/fe-core/src/main/java/org/apache/doris/PaloFe.java
```diff
@@ -173,6 +176,12 @@ public static void start(String dorisHomeDir, String pidDir, String[] args, Star
                 qeService.start();
             }
 
+            ScheduledThreadPoolExecutor refreshTimer = ThreadPoolManager.newDaemonScheduledThreadPool(1,
```
morningman, line 179: This `ScheduledThreadPoolExecutor` should be in `RefreshMgr`.
And there is already a `CatalogMgr` is `Env.java`, why new a `CatalogMgr` here?

Currently, there are 2 main manager:
- RefreshMgr
- CatalogMgr

All refresh method(refresh catalog/db/table/partition) should be in RefreshMgr.
But currently, there is a mistake that "refresh catalog method" is in CatalogMgr.

I think you can do a refactor to move the "refresh catalog method" from CatalogMgr to RefreshMgr.
And this newly added refresh timer thread should be in RefreshMgr too.

morningman: Please change your IDEA formatter to only format the code you modified. Not changing other code line.

fe/fe-core/src/main/java/org/apache/doris/PaloFe.java
```diff
@@ -173,6 +176,12 @@ public static void start(String dorisHomeDir, String pidDir, String[] args, Star
...
+                "refresh-timer-pool", true);
+            Integer refreshTime = 20;
```
morningman, line 181: What does `20` mean? It is a magic number to me

fe/fe-core/src/main/java/org/apache/doris/datasource/CatalogMgr.java
```diff
@@ -88,17 +91,25 @@ public class CatalogMgr implements Writable, GsonPostProcessable {
     private final Map<Long, CatalogIf> idToCatalog = Maps.newConcurrentMap();
     // this map will be regenerated from idToCatalog, so not need to persist.
     private final Map<String, CatalogIf> nameToCatalog = Maps.newConcurrentMap();
+
+    private static final Map<String, Integer[]> refreshMap = Maps.newConcurrentMap();
```
morningman, line 95: Add some comment in code to explain the key value in refreshMap

fe/fe-core/src/main/java/org/apache/doris/PaloFe.java
q763562998: 1.The reason why I want to create a new `CatalogMgr` is because the `CatalogMgr` needs a timing refresh time, because when the `CatalogMgr` of `Env` is initialized, there is no such timing refresh time, in order to reuse I will subsequently use the `CatalogMgr` of `Env` and assign this refresh time to the` CatalogMg`r of `Env` when it is initialized and constructed
2.Ok,I will  move the "refresh catalog method" from `CatalogMgr` to `RefreshMgr` and putthe refresh timer thread in `RefreshMgr`
3.Ok, I will not change other lines of code

fe/fe-core/src/main/java/org/apache/doris/PaloFe.java
q763562998: Indeed it pops up, it is a timer refresh time that will go after every 20 seconds to refresh the catalog that needs to be refreshed

fe/fe-core/src/main/java/org/apache/doris/datasource/CatalogMgr.java
q763562998: ok,I will add some comment

fe/fe-core/src/main/java/org/apache/doris/catalog/RefreshManager.java
```diff
@@ -19,26 +19,48 @@
...
 import java.util.Map;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
 // Manager for refresh database and table action
-public class RefreshManager {
+public class RefreshManager implements Runnable {
     private static final Logger LOG = LogManager.getLogger(RefreshManager.class);
 
+    private static final ScheduledThreadPoolExecutor REFRESH_Timer = ThreadPoolManager.newDaemonScheduledThreadPool(1,
```
morningman, line 48:
```suggestion
    private static final ScheduledThreadPoolExecutor REFRESH_TIMER = ThreadPoolManager.newDaemonScheduledThreadPool(1,
```

fe/fe-core/src/main/java/org/apache/doris/catalog/RefreshManager.java
```diff
@@ -19,26 +19,48 @@
...
+            "refresh-timer-pool", true);
```
morningman, line 49:
```suggestion
            "catalog-refresh-timer-pool", true);
```

fe/fe-core/src/main/java/org/apache/doris/catalog/RefreshManager.java
```diff
@@ -19,26 +19,48 @@
...
+    // Unit:SECONDS
+    private static final int REFRESH_TIME = 20;
+    // key is the name of a catalog, value is an array of length 2, used to store
+    // the original refresh time and the current remaining time of the catalog
+    private static Map<String, Integer[]> refreshMap = Maps.newConcurrentMap();
```
morningman, line 54:
```suggestion
    private Map<String, Integer[]> refreshMap = Maps.newConcurrentMap();
```

fe/fe-core/src/main/java/org/apache/doris/datasource/CatalogMgr.java
```diff
@@ -465,6 +470,20 @@ public CatalogIf replayCreateCatalog(CatalogLog log) throws DdlException {
...
+                // need refresh
+                String catalogName = log.getCatalogName();
+                Integer metadataRefreshIntervalSec = null;
+                try {
+                    metadataRefreshIntervalSec = Integer.valueOf(props.get(METADATA_REFRESH_INTERVAL_SEC));
+                } catch (NumberFormatException e) {
+                    throw new DdlException("Invalid properties: " + METADATA_REFRESH_INTERVAL_SEC);
+                }
+                Integer[] sec = {metadataRefreshIntervalSec, metadataRefreshIntervalSec};
+                Env.getCurrentEnv().getRefreshManager().getRefreshMap().put(catalogName, sec);
```
morningman, line 484: wrap the `getRefreshMap().put(catalogName, sec)` into `RefreshManger()`.
Better not expose raw map interface outside

morningman:
Please add FE ut for this.
You can refer to `ColumnPrivTest`.
And there are 2 classes: `TestExternalCatalog` and `TestCatalogProvider`.
These classes are for unit test, you can use `TestCatalogProvider` to add any schema you want to `TestExternalCatalog`,
for mocking some "metadata update" cases

morningman:
And also need to rebase to solve the conflict

fe/fe-core/src/main/java/org/apache/doris/datasource/CatalogMgr.java
q763562998: Ok ，I will modify the above code and encapsulation `getRefreshMap().put(catalogName, sec)` into `RefreshManger()`

q763562998:
> Please add FE ut for this. You can refer to `ColumnPrivTest`. And there are 2 classes: `TestExternalCatalog` and `TestCatalogProvider`. These classes are for unit test, you can use `TestCatalogProvider` to add any schema you want to `TestExternalCatalog`, for mocking some "metadata update" cases

Ok I have added FE ut

fe/fe-core/src/main/java/org/apache/doris/catalog/RefreshManager.java
```diff
@@ -146,4 +164,36 @@ private void refreshInternalCtlIcebergTable(RefreshTableStmt stmt, Env env) thro
                 stmt.getTableName(), "ICEBERG", icebergProperties, "");
         env.createTable(createTableStmt);
     }
+
+    public void registerIntoRefreshMap(String catalogName, Integer[] sec) {
```
morningman, line 168:
```suggestion
    public void addToRefreshMap(String catalogName, Integer[] sec) {
```

fe/fe-core/src/main/java/org/apache/doris/datasource/CatalogMgr.java
```diff
@@ -123,6 +124,7 @@ private CatalogIf removeCatalog(long catalogId) {
         if (catalog != null) {
             catalog.onClose();
             nameToCatalog.remove(catalog.getName());
+            Env.getCurrentEnv().getRefreshManager().logOutOfRefreshMap(catalog.getName());
```
morningman, line 127: Move this to `catalog.onClose()`

fe/fe-core/src/main/java/org/apache/doris/datasource/CatalogMgr.java
```diff
@@ -462,6 +464,20 @@ public CatalogIf replayCreateCatalog(CatalogLog log, boolean isReplay) throws Dd
         writeLock();
         try {
             CatalogIf catalog = CatalogFactory.constructorFromLog(log);
+            Map<String, String> props = log.getProps();
+            if (props.containsKey(METADATA_REFRESH_INTERVAL_SEC)) {
```
morningman, line 468: This should be done after `((ExternalCatalog) catalog).checkProperties()`.
And all properties, including `METADATA_REFRESH_INTERVAL_SEC`, should be checked in `checkProperties()`.
After `checkProperties()`, all properties should be valid.

fe/fe-core/src/main/java/org/apache/doris/datasource/CatalogMgr.java
```diff
@@ -462,6 +464,20 @@ public CatalogIf replayCreateCatalog(CatalogLog log, boolean isReplay) throws Dd
...
+                // need refresh
+                String catalogName = log.getCatalogName();
+                Integer metadataRefreshIntervalSec = null;
+                try {
+                    metadataRefreshIntervalSec = Integer.valueOf(props.get(METADATA_REFRESH_INTERVAL_SEC));
+                } catch (NumberFormatException e) {
+                    throw new DdlException("Invalid properties: " + METADATA_REFRESH_INTERVAL_SEC);
+                }
+                Integer[] sec = {metadataRefreshIntervalSec, metadataRefreshIntervalSec};
+                Env.getCurrentEnv().getRefreshManager().registerIntoRefreshMap(catalogName, sec);
```
morningman, line 478: You also need to modify the method `gsonPostProcess()` in this class.
To add all catalogs which need to refresh to the RefreshManger, or these refresh info will be lost.

You can test this bug with following steps:
1. create a catalog with refresh property.
2. Restart FE, and wait for 2 minutes.
3. Restart FE again, you will see that the catalog won't refresh anymore.

fe/fe-core/src/main/java/org/apache/doris/catalog/RefreshManager.java
```diff
@@ -19,26 +19,44 @@
...
+    private static final ScheduledThreadPoolExecutor REFRESH_TIMER = ThreadPoolManager.newDaemonScheduledThreadPool(1,
+        "catalog-refresh-timer-pool", true);
+    // Unit:SECONDS
+    private static final int REFRESH_TIME = 5;
+    // key is the name of a catalog, value is an array of length 2, used to store
+    // the original refresh time and the current remaining time of the catalog
+    private Map<String, Integer[]> refreshMap = Maps.newConcurrentMap();
+
+    public RefreshManager() {
+        REFRESH_TIMER.scheduleAtFixedRate(this::run, 0, REFRESH_TIME, TimeUnit.SECONDS);
```
morningman, line 57: Should not start the schedule thread in constructor.
Because if do so, every time a `RefreshManager` is created, the schedule thread will start.
But the schedule thread should only start on Master FE.
So you need to add a new method `start()` to start schedule thread.
But this `start()` method should be called in `Env.startMasterOnlyDaemonThreads()`

q763562998: Ok, I thought `RefreshManager` was a singleton object before such an bug occurred, I will correct it. And add `start()` in `Env.startMasterOnlyDaemonThreads()`

q763562998:
Thanks for your suggestions.
- I have moved the `RereshManager.removeFromRefreshMap()` to `catalog.onClose()`.
- In `ExternalCatalog.checkProperties()` I have added a judgment on the refresh parameter METADATA_REFRESH_INTERVAL_SEC. 
- I have  added registration to`RefreshManger` in `CatalogMgr.gsonPostProcess()`.
- And changed the timing mechanism of `RefreshManger` 
- Removed `start()` of `RefreshManager` to `Env`

fe/fe-core/src/main/java/org/apache/doris/catalog/RefreshManager.java
```diff
@@ -19,25 +19,38 @@
...
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
 // Manager for refresh database and table action
 public class RefreshManager {
     private static final Logger LOG = LogManager.getLogger(RefreshManager.class);
+    private ScheduledThreadPoolExecutor refreshScheduler = ThreadPoolManager.newDaemonScheduledThreadPool(1,
+            "catalog-refresh-timer-pool", true);
+    // Unit:SECONDS
+    private static final int REFRESH_TIME = 5;
```
morningman, line 50:
```suggestion
    private static final int REFRESH_TIME_SEC = 5;
```

fe/fe-core/src/main/java/org/apache/doris/catalog/RefreshManager.java
```diff
@@ -19,25 +19,38 @@
...
+    // key is the name of a catalog, value is an array of length 2, used to store
```
morningman, line 51:
```suggestion
    // key is the id of a catalog, value is an array of length 2, used to store
```

fe/fe-core/src/main/java/org/apache/doris/catalog/RefreshManager.java
```diff
@@ -146,4 +159,45 @@ private void refreshInternalCtlIcebergTable(RefreshTableStmt stmt, Env env) thro
...
+        refreshMap.remove(catalogId);
+    }
+
+    public void start() {
+        TaskRefresh taskRefresh = new TaskRefresh();
+        this.refreshScheduler.scheduleAtFixedRate(taskRefresh, 0, REFRESH_TIME,
+                TimeUnit.SECONDS);
+    }
+
+    private class TaskRefresh implements Runnable {
```
morningman, line 177:
```suggestion
    private class RefreshTask implements Runnable {
```

fe/fe-core/src/main/java/org/apache/doris/catalog/RefreshManager.java
```diff
@@ -146,4 +159,45 @@ private void refreshInternalCtlIcebergTable(RefreshTableStmt stmt, Env env) thro
...
+            for (Map.Entry<Long, Integer[]> entry : refreshMap.entrySet()) {
+                Long catalogId = entry.getKey();
+                Integer[] timeGroup = entry.getValue();
+                Integer original = timeGroup[0];
+                Integer current = timeGroup[1];
+                if (current - REFRESH_TIME > 0) {
+                    timeGroup[1] = current - REFRESH_TIME;
+                    refreshMap.put(catalogId, timeGroup);
+                } else {
+                    String catalogName = Env.getCurrentEnv().getCatalogMgr().getCatalog(catalogId).getName();
```
morningman, line 189: Catalog may be dropped while you are doing this refresh task.
So need to check if the return value of `getCatalog` is null, or just using `getCatalogOrException` and catch the exception

q763562998: Ok,I will check if the return value of `getCatalog` is null

morningman:
run buildall

morningman (approved):
LGTM

PR #17894, opened 2023-03-17
[fe](test)Add MicroBench module to make it easier for developers to write JMH Test

CalvinKirs:
Writing the benchmark itself is not hard, but getting it right is. Thankfully, the JMH suite provides helpful annotations and features to mitigate most of them. To get started, you need to make your benchmark extend the AbstractMicrobenchmark, which makes sure the test gets run through JUnit and configures some defaults:
```
@BenchmarkMode({Mode.Throughput})
public class LinkedArrayBenchmark extends AbstractMicrobenchmark {
}
```
For most benchmark tests, you can write it as a normal JUnit.
We could also choose Customizing Runtime Conditions.
At present, you can run all JMH Tests by running the jar directly after packaging.

How to used:

  All JMH tests need to extend this class to make it easier for you to complete JMHTest, you can also choose to
 customize runtime conditions (Measurement, Fork, Warmup, etc.)
 <p>
 You can run any of the JMH tests as a normal UT, or you can package it and get all the reported results via `java
 -jar doris-benchmark.jar`, or get the results of a particular Test via `java -jar dorisbenchmarks.jar exampleClassName`.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -14,6 +14,12 @@ Describe your changes.
 * [ ] Does it need to update dependencies
 * [ ] Is this PR support rollback (If NO, please explain WHY)
 
+
+* 📈 Remove this line WHEN AND ONLY WHEN you're improving the performance, follow the checklist 👇 *
+### Improve the performance of <class or module or ...>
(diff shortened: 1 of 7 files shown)


CalvinKirs:
We don't need to run other tests PTAL, thx @morningman

hf200012 (dismissed):
LGTM

hf200012 (approved):
LGTM

morningman:
I suggest to hold this PR until we meet the real need.

Issue #17947, opened 2023-03-20
[Feature] Support hive meta cache TTL


morningman:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

### Description

Currently, if user modify the file on hdfs directly, no through hive. The changes of file will not be noticed by Doris and user
will get wrong data.

One approach is to refresh the catalog at fixed rate. But this will have to refresh all catalog and invalidate all caches.

So I suggest to set TTL(Time-to-Live) config of File Cache, so that the stale file info will be invalidated automatically
after expiring.

### Use case

1. create catalog with new property
```
create catalog hive(
...,
"file_meta_cache_ttl_second" = "60"
);

And file cache of all tables in this catalog will be expired after 60 seconds.
You can also set `file_meta_cache_ttl_second` to 0 to disable file cache.

### Related issues

_No response_

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

lexluo09:
I'm interested in this feature. Could you assign it to me? Thank you

dutyu:
> I'm interested in this feature. Could you assign it to me? Thank you

Hello, wanna get to know is there any progress about this feature?

lexluo09:
> 

Hi, I'm ready to submit it now.

PR #18102, opened 2023-03-24
[Enhancement](multi catalog) Support hive meta cache TTL

lexluo09:
# Proposed changes

Issue Number: #17947

Currently, if user modify the file on hdfs directly, no through hive. The changes of file will not be noticed by Doris and user
will get wrong data. Support the TTL(Time-to-Live) config of File Cache, so that the stale file info will be invalidated automatically after expiring.

## Problem summary
1.Add a parameter configuration to set file cache ttl. "file.meta.cache.ttl-second".
2.Set the value corresponding to guava expireAfterAccess to the configuration value.

## Checklist(Required)

### DIFF
--- a/docs/zh-CN/docs/lakehouse/multi-catalog/hive.md
+++ b/docs/zh-CN/docs/lakehouse/multi-catalog/hive.md
@@ -151,6 +151,22 @@ CREATE CATALOG hive WITH RESOURCE hms_resource PROPERTIES(
 	'key' = 'value'
 );
 ```
+<version since="dev"></version>
+创建 Catalog 时可以采用参数 `file.meta.cache.ttl-second` 来设置 File Cache 自动失效时间，也可以将该值设置为 0 来禁用 File Cache。时间单位为：秒。示例如下：
+```sql
+CREATE CATALOG hive PROPERTIES (
+    'type'='hms',
+    'hive.metastore.uris' = 'thrift://172.21.0.1:7004',
+    'hadoop.username' = 'hive',
+    'dfs.nameservices'='your-nameservice',
+    'dfs.ha.namenodes.your-nameservice'='nn1,nn2',
+    'dfs.namenode.rpc-address.your-nameservice.nn1'='172.21.0.2:4007',
+    'dfs.namenode.rpc-address.your-nameservice.nn2'='172.21.0.3:4007',
+    'dfs.client.failover.proxy.provider.your-nameservice'='org.apache.hadoop.hdfs.server.namenode.ha.ConfiguredFailoverProxyProvider',
+    'file.meta.cache.ttl-second' = '60'
+);
+```
+
 
 我们也可以直接将 hive-site.xml 放到 FE 和 BE 的 conf 目录下，系统也会自动读取 hive-site.xml 中的信息。信息覆盖的规则如下：
 	
--- a/fe/fe-core/src/main/java/org/apache/doris/datasource/HMSExternalCatalog.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/datasource/HMSExternalCatalog.java
@@ -31,6 +31,7 @@ import org.apache.doris.datasource.hive.event.MetastoreNotificationFetchExceptio
 import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import org.apache.commons.lang.math.NumberUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.metastore.api.CurrentNotificationEventId;
@@ -42,6 +43,7 @@ import org.apache.logging.log4j.Logger;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 /**
  * External catalog for hive metastore compatible data sources.
@@ -54,6 +56,12 @@ public class HMSExternalCatalog extends ExternalCatalog {
     // Record the latest synced event id when processing hive events
     private long lastSyncedEventId;
     public static final String ENABLE_SELF_SPLITTER = "enable.self.splitter";
+    public static final String FILE_META_CACHE_TTL_SECOND = "file.meta.cache.ttl-second";
+
+    // -1 means file cache no ttl set
+    public static final int FILE_META_CACHE_NO_TTL = -1;
+    // 0 means file cache is disabled; >0 means file cache with ttl;
+    public static final int FILE_META_CACHE_TTL_DISABLE_CACHE = 0;
 
     /**
      * Default constructor for HMSExternalCatalog.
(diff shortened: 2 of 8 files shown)


docs/zh-CN/docs/lakehouse/multi-catalog/hive.md
```diff
@@ -152,6 +152,22 @@ CREATE CATALOG hive WITH RESOURCE hms_resource PROPERTIES(
 );
 ```
 
+在 2.0 版本之后，我们可以在创建 Catalog 时可以采用参数 `file.meta.cache.ttl-second` 来设置 File Cache 自动失效时间，也可以将该值设置为 0 来禁用 File Cache。时间单位为：秒。示例如下：
```
morningman, line 155:
```
<version since="dev"></version>
创建 Catalog 时可以采用参数 `file.meta.cache.ttl-second` 来设置 File Cache 自动失效时间，也可以将该值设置为 0 来禁用 File Cache。时间单位为：秒。示例如下：
```

fe/fe-core/src/main/java/org/apache/doris/datasource/hive/HiveMetaStoreCache.java
```diff
@@ -120,14 +121,22 @@ public HivePartition load(PartitionCacheKey key) throws Exception {
...
-                    }
-                }, executor));
+        // if the file.meta.cache.ttl-second is equal or greater than 0, the cache expired will be set to that value
+        int fileMetaCacheTtlSecond = NumberUtils.toInt(
+            (catalog.getProperties().getOrDefault(HMSExternalCatalog.FILE_META_CACHE_TTL_SECOND, "")), -1);
+
+        CacheBuilder<Object, Object> fileCacheBuilder = CacheBuilder.newBuilder().maximumSize(Config.max_external_file_cache_num);
+        if (fileMetaCacheTtlSecond >= 0) {
+            fileCacheBuilder.expireAfterWrite(fileMetaCacheTtlSecond, TimeUnit.SECONDS);
+        }
```
morningman, line 131: We also need to let user disable the file cache.
Maybe we can use `0` as a special value?
`0` means file cache is disabled.
`>0` means file cache with ttl
`-1` means file cache no ttl

dutyu: > We also need to let user disable the file cache. Maybe we can use `0` as a special value? `0` means file cache is disabled. `>0` means file cache with ttl `-1` means file cache no ttl

I think maybe we can make this logic easier, `file.meta.cache.ttl-second` should just be greater or equal to 0, 0 means disable the file cache, >0 means enable ttl with file cache, other values are illegal，if user does not want to enable the ttl with file cache, just does not set `file.meta.cache.ttl-second`.

fe/fe-core/src/main/java/org/apache/doris/datasource/HMSExternalCatalog.java
```diff
@@ -69,6 +77,13 @@ public HMSExternalCatalog(long catalogId, String name, String resource, Map<Stri
     @Override
     public void checkProperties() throws DdlException {
         super.checkProperties();
+        // check file.meta.cache.ttl-second parameter
+        String fileMetaCacheTtlSecond = catalogProperty.getOrDefault(FILE_META_CACHE_TTL_SECOND, "");
```
dutyu, line 81: User can set `file.meta.cache.ttl-second` = "" and it will pass the check here, although we will not enable ttl with file cache finally, but i think we should let user follow the rules(the value of `file.meta.cache.ttl-second` only can be set as a non-negative number).

lexluo09: 1. In this case, if user set a negative number , the check will not pass.
2. If user set `file.meta.cache.ttl-second` = "", the check will pass and not enable ttl.

lexluo09: Thank you for your advice.
If user can set `file.meta.cache.ttl-second` = "", it is necessary not to pass the check. I will fix it.

dutyu: Yeah, i know it works, but we can distinguish this scene which user set `file.meta.cache.ttl-second` = "" and let user just do not set `file.meta.cache.ttl-second`, well it's depond on you, current implementation is acceptable.

lexluo09: Yeah, well, thanks for your advice. I'll think about it.

lexluo09: Yeah, well, thanks for your advice. I'll think about it.

fe/fe-core/src/main/java/org/apache/doris/datasource/hive/HiveMetaStoreCache.java
```diff
@@ -120,14 +121,22 @@ public HivePartition load(PartitionCacheKey key) throws Exception {
...
-                    @Override
-                    public ImmutableList<InputSplit> load(FileCacheKey key) throws Exception {
-                        return loadFiles(key);
-                    }
-                }, executor));
+        // if the file.meta.cache.ttl-second is equal or greater than 0, the cache expired will be set to that value
+        int fileMetaCacheTtlSecond = NumberUtils.toInt(
+            (catalog.getProperties().getOrDefault(HMSExternalCatalog.FILE_META_CACHE_TTL_SECOND, "")), HMSExternalCatalog.FILE_META_CACHE_NO_TTL);
+
+        CacheBuilder<Object, Object> fileCacheBuilder = CacheBuilder.newBuilder().maximumSize(Config.max_external_file_cache_num);
```
dutyu, line 128: If `file.meta.cache.ttl-second` = '0', maybe we should not initialize the loading cache, just set `fileCache` = null and let the current thread to load the file splits. Because loading file splits of a hive table with many partitions is a time-consuming task, it can easily drain the cache loader thread pool(pool size = 10 now),  it will block all querys with external tables if any cache (partitionValuesCache/partitionCache/fileCache/schemaCache) is expired.

lexluo09:
run buildall

fe/fe-core/src/main/java/org/apache/doris/datasource/HMSExternalCatalog.java
lexluo09: Thank you for your advice.
If user set `file.meta.cache.ttl-second` = "", the check will not pass check now.

fe/fe-core/src/main/java/org/apache/doris/datasource/hive/HiveMetaStoreCache.java
lexluo09: When duration is zero, this method hands off to maximumSize(0), ignoring any otherwise-specified maximum size or weight. This can be useful in testing, or to disable caching temporarily without a code change.

lexluo09: 1.

lexluo09: Yeah, well, thanks for your advice. I agree with you. To handle the code more gracefully, I recommend that the guava CacheLoader be set to synchronous.

lexluo09:
run buildall

lexluo09:
run buildall

fe/fe-core/src/main/java/org/apache/doris/datasource/hive/HiveMetaStoreCache.java
dutyu: I think this is a good idea~

dutyu (commented):
LGTM

fe/fe-core/src/main/java/org/apache/doris/datasource/hive/HiveMetaStoreCache.java
dutyu: This pr is lgtm, but my perspective is limited, maybe you can request `morinngman` to review it again.

PasunuriSrinidhi:
To implement this solution, the system would set the value corresponding to Guava's expireAfterAccess to the configuration value. Guava is a widely used open-source library for Java, which provides various utilities including caching mechanisms. The expireAfterAccess configuration in Guava's cache mechanism specifies the duration after which a cache entry will be automatically removed from the cache if it has not been accessed during that time.

morningman (dismissed):
LGTM

morningman:
run buildall

lexluo09:
run buildall

lexluo09:
run buildall

morningman (approved):
LGTM

morningman:
run buildall

Issue #18125, opened 2023-03-27
[feature](executor) Add memory limit for pip_scanner_context


wangbo:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

### Description

pip_scanner_context's memory usage should be limited. Because:
1 It can reduce memory usage and reduce allocating new blocks when executing.
2 When the memory is limited, the schedule times of scanner_ctx and scanner can be limited.

# High concurrency scenario testing

## Test Description
Using mysqlslap start query client 1 to simulate high concurrency scenarios
```
 -c 50 -i 100000 -q "SELECT count(distinct LO_SHIPMODE) , count(distinct C_NAME), count(distinct C_ADDRESS), count(distinct C_CITY), count(distinct C_NATION)  FROM lineorder_flat WHERE LO_ORDERDATE >= 19930101     AND LO_ORDERDATE <= 19931231     AND LO_DISCOUNT BETWEEN 1 AND 3     AND LO_QUANTITY < 25"
```

Starting client 2 to simulate short query, mainly using client 2's query latency to observe scheduler's performance.
```
-c 1 -i 20 -q "set enable_profile=true;SELECT    sum(LO_EXTENDEDPRICE  *  LO_DISCOUNT)       FROM  ssb100.lineorder_flat  WHERE  LO_ORDERDATE  >=  19930101  AND  LO_ORDERDATE  <=  19931231"
```

### query time

before:
```
round 1
	Average number of seconds to run all queries: 2.044 seconds
	Minimum number of seconds to run all queries: 0.279 seconds
	Maximum number of seconds to run all queries: 5.529 seconds

round 2
	Average number of seconds to run all queries: 1.884 seconds
	Minimum number of seconds to run all queries: 0.261 seconds
	Maximum number of seconds to run all queries: 5.212 seconds

round 3
	Average number of seconds to run all queries: 1.879 seconds
	Minimum number of seconds to run all queries: 0.264 seconds
	Maximum number of seconds to run all queries: 5.452 seconds

round 4
	Average number of seconds to run all queries: 1.884 seconds
	Minimum number of seconds to run all queries: 0.286 seconds
	Maximum number of seconds to run all queries: 5.716 seconds

round 5
	Average number of seconds to run all queries: 1.809 seconds
	Minimum number of seconds to run all queries: 0.269 seconds
	Maximum number of seconds to run all queries: 5.225 seconds

```

after:
```
round 1
[...]

PR #18126, opened 2023-03-27
[feature](executor) Add memory limit for pip_scanner_context

wangbo:
# Proposed changes
#18125

pip_scanner_context's memory usage should be limited. Because:
1 It can reduce memory usage and reduce allocating new blocks when executing.
2 When the memory is limited, the schedule times of scanner_ctx and scanner can be limited.

## Problem summary

I'll add performance test later

## Checklist(Required)

* [x] Does it affect the original behavior

## Further comments

### DIFF
--- a/be/src/vec/exec/scan/pip_scanner_context.h
+++ b/be/src/vec/exec/scan/pip_scanner_context.h
@@ -31,7 +31,9 @@ public:
                       const std::list<vectorized::VScanner*>& scanners, int64_t limit,
                       int64_t max_bytes_in_blocks_queue)
             : vectorized::ScannerContext(state, parent, input_tuple_desc, output_tuple_desc,
-                                         scanners, limit, max_bytes_in_blocks_queue) {}
+                                         scanners, limit, max_bytes_in_blocks_queue) {
+        _should_resche_after_scanner_finished = false;
+    }
 
     Status get_block_from_queue(RuntimeState* state, vectorized::BlockUPtr* block, bool* eos,
                                 int id, bool wait = false) override {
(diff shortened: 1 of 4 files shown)


Gabriel39:
run buildall

be/src/vec/exec/scan/scanner_context.cpp
```diff
@@ -55,8 +55,7 @@ Status ScannerContext::init() {
     // 1. Calculate max concurrency
     // TODO: now the max thread num <= config::doris_scanner_thread_pool_thread_num / 4
     // should find a more reasonable value.
-    _max_thread_num = _state->shared_scan_opt() ? config::doris_scanner_thread_pool_thread_num
-                                                : config::doris_scanner_thread_pool_thread_num / 4;
+    _max_thread_num = config::doris_scanner_thread_pool_thread_num / 4;
```
Gabriel39, line 58: I think it's better to provide a tuning method for users (maybe add a configuration).

be/src/vec/exec/scan/pip_scanner_context.h
```diff
@@ -89,10 +95,15 @@ class PipScannerContext : public vectorized::ScannerContext {
...
 
+    bool has_enough_space_in_blocks_queue() const override {
+        return _current_used_bytes < _max_bytes_in_queue / 2;
+    }
+
 private:
     int _next_queue_to_feed = 0;
     std::vector<std::unique_ptr<std::mutex>> _queue_mutexs;
     std::vector<std::list<vectorized::BlockUPtr>> _blocks_queues;
+    std::atomic_int64_t _current_used_bytes = 0;
```
Gabriel39, line 106: I suspect if it's necessary to make `_current_used_bytes` atomic since this variable is accessed almost guarded by a mutex

PR #18152, opened 2023-03-27
[feature](colocate) support cross database colocate join

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem summary

In previous, Doris does not support cross-db colocate join, because the a colocate group belongs
to a certain database.
This PR supports cross-db colocate join.
User can create table with `colocate_with` properties with `__global__` prefix group name, eg:

```
CREATE TABLE tbl (k1 int, v1 int sum)
DISTRIBUTED BY HASH(k1)
BUCKETS 8
PROPERTIES(
     "colocate_with" = "__global__group1"
);
```

And the group `__global__group1` no longer belongs to a certain database, but is global unique group.

### Notice
Upgrade the FE meta version to 119

### Known issue
The implementation is not every elegant.
In previous implementation, the colocate group id is assembled with `dbId` + `groupname`.
In order to support cross-db colocate group, I set `dbId` in group id to 0, but this will lost the db id info.
So I have to add another field `tblId2DbId` in `GroupId`, to save the db id info
for the global colocate group(which db id is 0).

And when adding/removing/modify table's colocate group, we must be careful to handle this `tblId2DbId` too.

## Checklist(Required)

## Further comments


morningman:
run buildall

yiguolei (approved):
LGTM

PR #18397, opened 2023-04-05
[bug](GC)the issue of incorrect disk usage

chenlinzhong:
# Proposed changes

disk used capacity is incorrect：
DataUsedCapacity+AvailCapacity !=TotalCapacity
image

## Problem summary

some invalid rowset rowset not be  deleted for ever due to the bug:
image
when key_may_exists() return true not means it is real exists

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/io/fs/file_system.cpp
+++ b/be/src/io/fs/file_system.cpp
@@ -17,6 +17,8 @@
 
 #include "io/fs/file_system.h"
 
+#include <sys/stat.h>
+
 #include "olap/olap_define.h"
 #include "util/async_io.h"
 
@@ -42,6 +44,24 @@ Status FileSystem::create_directory(const Path& dir, bool failed_if_exists) {
 Status FileSystem::delete_file(const Path& file) {
     auto path = absolute_path(file);
     FILESYSTEM_M(delete_file_impl(path));
+}
+
+Status FileSystem::delete_directory_or_file(const Path& path) {
+    auto real_path = absolute_path(path);
+    if (is_dir(path)) {
+        FILESYSTEM_M(delete_directory_impl(path));
+    } else {
+        FILESYSTEM_M(delete_file_impl(path));
+    }
+}
+
+bool FileSystem::is_dir(const Path& path) {
+    struct stat path_stat;
+    if (stat(path.c_str(), &path_stat) != 0) {
+        return false;
+    } else {
+        return S_ISDIR(path_stat.st_mode);
+    }
 }
 
 Status FileSystem::delete_directory(const Path& dir) {
--- a/be/src/io/fs/file_system.cpp
+++ b/be/src/io/fs/file_system.cpp
@@ -46,15 +46,6 @@ Status FileSystem::delete_file(const Path& file) {
     FILESYSTEM_M(delete_file_impl(path));
 }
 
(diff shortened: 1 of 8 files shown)


chenlinzhong:
run p0

chenlinzhong:
run p0

chenlinzhong:
run beut

chenlinzhong:
run feut

chenlinzhong:
run p0

chenlinzhong:
ru beut

chenlinzhong:
run clickbench

be/src/io/fs/file_system.cpp
```diff
@@ -44,6 +46,24 @@ Status FileSystem::delete_file(const Path& file) {
     FILESYSTEM_M(delete_file_impl(path));
 }
 
+Status FileSystem::delete_directory_or_file(const Path& path) {
+    auto real_path = absolute_path(path);
```
morningman, line 50: I think we should implement this only in `LocalFileSystem`.
Because you use `stat` to check if it is dir, which is only available for local fs.

And there is already a method `is_directory()` in local fs, better to use that.

be/src/olap/rowset/rowset_meta_manager.cpp
```diff
@@ -42,6 +42,17 @@ bool RowsetMetaManager::check_rowset_meta(OlapMeta* meta, TabletUid tablet_uid,
     return meta->key_may_exist(META_COLUMN_FAMILY_INDEX, key, &value);
 }
 
+bool RowsetMetaManager::exists(OlapMeta* meta, TabletUid tablet_uid, const RowsetId& rowset_id) {
+    std::string key = ROWSET_PREFIX + tablet_uid.to_string() + "_" + rowset_id.to_string();
+    std::string value;
+    Status s = meta->get(META_COLUMN_FAMILY_INDEX, key, &value);
```
morningman, line 48: Better to return `s` directly instead of a `bool`.
So that if there is other error, we can handle it.
Otherwise, if the `s` is InternalError(for example), this will still return `true`, which is wrong

be/src/olap/tablet.cpp
```diff
@@ -1160,7 +1160,7 @@ bool Tablet::check_rowset_id(const RowsetId& rowset_id) {
             return true;
         }
     }
-    if (RowsetMetaManager::check_rowset_meta(_data_dir->get_meta(), tablet_uid(), rowset_id)) {
+    if (RowsetMetaManager::exists(_data_dir->get_meta(), tablet_uid(), rowset_id)) {
```
morningman, line 1163: Handle the `Status` here

hello-stephen:
 the sum of best hot time: 33.23 seconds
 stream load tsv:          463 seconds loaded 74807831229 Bytes, about 154 MB/s
 stream load json:         21 seconds loaded 2358488459 Bytes, about 107 MB/s
 stream load orc:          73 seconds loaded 1101869774 Bytes, about 14 MB/s
 stream load parquet:          30 seconds loaded 861443392 Bytes, about 27 MB/s
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20230406075000_clickbench_pr_125369.html

chenlinzhong:
run buildall

morningman (approved):
LGTM

PR #18382, opened 2023-04-04
[Bug](ES): Es object mapping error

stalary:
# Proposed changes

Issue Number: close https://github.com/apache/doris/issues/18379

## Problem summary

Es object mapping error

## Checklist(Required)

* [x] Does it affect the original behavior
* [x] Has unit tests been added
* [ ] Has document been added or modified
* [ ] Does it need to update dependencies
* [ ] Is this PR support rollback (If NO, please explain WHY)

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #18438, opened 2023-04-06
[fix](es) ClassCastException when getting root schema

gnehil:
# Proposed changes

Issue Number: close #xxx

## Problem summary

When the dynamic parameter in the mapping and mapping type is set to null, getting the root schema will cause ClassCastException

example:
```json
{
  "test_index": {
    "mappings": {
      "dynamic": "false",
      "properties": {
        "time": {
          "type": "long"
        },
        "type": {
          "type": "keyword"
        },
        "userId": {
          "type": "text",
          "fields": {
            "keyword": {
              "type": "keyword"
            }
          }
        }
      }
    }
  }
}
```

When the mapping structure is as above

The exception that occurs in the latest version is `java.lang.ClassCastException: com.fasterxml.jackson.databind.node.TextNode cannot be cast to com.fasterxml.jackson.databind.node.ObjectNode`,
and for elder version is `java.lang.ClassCastException: java.lang.String cannot be cast to org.json.simple.JSONObject`

Describe your changes.

## Checklist(Required)

* [x] Does it affect the original behavior

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/external/elasticsearch/EsUtil.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/external/elasticsearch/EsUtil.java
@@ -152,6 +152,8 @@ public class EsUtil {
         }
         // remove `dynamic_templates` field
         mappings.remove("dynamic_templates");
+        // remove `dynamic` field
+        // mappings.remove("dynamic");
         // check explicit mapping
         if (mappings.isEmpty()) {
             throw new DorisEsException("Do not support index without explicit mapping.");
--- a/fe/fe-core/src/main/java/org/apache/doris/external/elasticsearch/EsUtil.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/external/elasticsearch/EsUtil.java
@@ -153,7 +153,7 @@ public class EsUtil {
         // remove `dynamic_templates` field
         mappings.remove("dynamic_templates");
         // remove `dynamic` field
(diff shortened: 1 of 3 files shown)


stalary:
It can resolve this problem in this pr https://github.com/apache/doris/pull/18382

gnehil:
I think we are not having the same problem.
The problem I have encountered is that when calling the `getRootSchema` method, when the `mappingType` is null and there is `dynamic` parameter in the mapping json, since the first element obtained by traversing the mapping json through the iterator is `dynamic`, and its type is `TextNode`, a `ClassCastException` will appear when running the following code：
```java
ObjectNode firstData = (ObjectNode) mappings.get(firstType);
```
All this happens before calling the `replaceFieldAlias` method.

stalary:
> I think we are not having the same problem. The problem I have encountered is that when calling the `getRootSchema` method, when the `mappingType` is null and there is `dynamic` parameter in the mapping json, since the first element obtained by traversing the mapping json through the iterator is `dynamic`, and its type is `TextNode`, a `ClassCastException` will appear when running the following code：
> 
> ```java
> ObjectNode firstData = (ObjectNode) mappings.get(firstType);
> ```
> 
> All this happens before calling the `replaceFieldAlias` method.

Okey, I get it.

stalary:
run buildall

stalary (approved):
LGTM

PR #18528, opened 2023-04-10
[wip][fix](storage) free dict_page immediately after usage to save memory

freemandealer:
The dictionary may be used for first few data pages but after a certain point, it will no longer be needed. Free it immediately to save memory.

# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_v2/column_reader.cpp
+++ b/be/src/olap/rowset/segment_v2/column_reader.cpp
@@ -1107,9 +1107,10 @@ Status FileColumnIterator::_read_data_page(const OrdinalPageIndexIterator& iter)
                 Slice dict_data;
                 PageFooterPB dict_footer;
                 _opts.type = INDEX_PAGE;
+                _dict_page_handle = std::make_unique<PageHandle>();
                 RETURN_IF_ERROR(_reader->read_page(_opts, _reader->get_dict_page_pointer(),
-                                                   &_dict_page_handle, &dict_data, &dict_footer,
-                                                   _compress_codec));
+                                                   _dict_page_handle.get(), &dict_data,
+                                                   &dict_footer, _compress_codec));
                 // ignore dict_footer.dict_page_footer().encoding() due to only
                 // PLAIN_ENCODING is supported for dict page right now
                 _dict_decoder = std::make_unique<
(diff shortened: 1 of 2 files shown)


be/src/olap/rowset/segment_v2/column_reader.h
```diff
@@ -359,7 +359,7 @@ class FileColumnIterator final : public ColumnIterator {
     std::unique_ptr<PageDecoder> _dict_decoder;
 
     // keep dict page handle to avoid released
```
zhannngchen, line 361: also update this comment

be/src/olap/rowset/segment_v2/column_reader.h
```diff
@@ -359,7 +359,7 @@ class FileColumnIterator final : public ColumnIterator {
...
-    PageHandle _dict_page_handle;
+    PageHandle* _dict_page_handle = nullptr;
```
zhannngchen, line 362: We'd better to avoid use pointer directly, consider unique_ptr?

be/src/olap/rowset/segment_v2/column_reader.h
freemandealer: > also update this comment

seems no problem? we do need to keep it for some time :)

zhannngchen (dismissed):
LGTM

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -1107,9 +1107,10 @@ Status FileColumnIterator::_read_data_page(const OrdinalPageIndexIterator& iter)
                 Slice dict_data;
                 PageFooterPB dict_footer;
                 _opts.type = INDEX_PAGE;
```
yiguolei, line 1109: If index is used, is there such case that we read page number 10 first and then read page number 1? Or in other words, could we make sure that the data page is read in descending order?

freemandealer: > If index is used, is there such case that we read page number 10 first and then read page number 1? Or in other words, could we make sure that the data page is read in descending order?

no sure, I will figure it out somehow

freemandealer: > If index is used, is there such case that we read page number 10 first and then read page number 1? Or in other words, could we make sure that the data page is read in descending order?

@yiguolei Yes, we are sure the data page is read in ascending order: after locating the first target page, we will read the following pages sequentially, for a single query of course.

freemandealer:
build all

freemandealer:
run p0

freemandealer:
run p0

freemandealer:
run buildall

freemandealer:
not compatible with 'in list'

PR #18609, opened 2023-04-12
[Fix](variables) fix session variable does not take effect immediately when set global variable in follower FE

Cai-Yao:
# Proposed changes

Issue Number: close #xxx

## Problem summary

For the non-master node, the `set global xxx = xxx` will not change the variables of the current session, because of `forward()`  to master node execute.
So the `set global xxx = xxx` needs to be executed again to change the current session after `forward()`.

This PR will change behavior, but I think the old behavior is a bug.

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/MasterOpExecutor.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/MasterOpExecutor.java
@@ -18,8 +18,12 @@
 package org.apache.doris.qe;
 
 import org.apache.doris.analysis.RedirectStatus;
+import org.apache.doris.analysis.SetStmt;
+import org.apache.doris.analysis.StatementBase;
 import org.apache.doris.common.ClientPool;
+import org.apache.doris.common.DdlException;
 import org.apache.doris.common.telemetry.Telemetry;
+import org.apache.doris.qe.QueryState.MysqlStateType;
 import org.apache.doris.thrift.FrontendService;
 import org.apache.doris.thrift.TMasterOpRequest;
 import org.apache.doris.thrift.TMasterOpResult;
@@ -45,6 +49,7 @@ public class MasterOpExecutor {
     private static final float RPC_TIMEOUT_COEFFICIENT = 1.2f;
 
     private final OriginStatement originStmt;
+    private StatementBase parsedStmt;
     private final ConnectContext ctx;
     private TMasterOpResult result;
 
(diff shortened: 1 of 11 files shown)


yiguolei (dismissed):
LGTM

yiguolei:
run buildall

fe/fe-core/src/main/java/org/apache/doris/qe/MasterOpExecutor.java
```diff
@@ -163,6 +179,24 @@ private void forward() throws Exception {
...
     }
 
+    // The SetStmt needs to be executed in follower again
+    // for the current session after master success
+    private void afterForwardExecute() throws DdlException {
+        if (parsedStmt != null) {
+            if (parsedStmt instanceof SetStmt) {
+                SetExecutor executor = new SetExecutor(ctx, (SetStmt) parsedStmt);
+                try {
+                    executor.setSessionVars();
```
morningman, line 189: I think we can set remove the `global` property in `SetStmt` and than call `executor.execute()` method. No need to create a new method `setSessionVars()`.

fe/fe-core/src/main/java/org/apache/doris/qe/SetExecutor.java
```diff
@@ -59,9 +59,24 @@ private void setVariable(SetVar var) throws DdlException {
         }
     }
 
+    private boolean isSessionVar(SetVar var) {
+        return !(var instanceof SetPassVar
```
morningman, line 63: This is error-prone, if other people at a new subclass of `SetVar`, this method may be missed.
How about add a `SetVarType` or something to specify the stmt's type directly?
And write like `return var.setType() == Type.SET_SESSION_VAR`;

Cai-Yao: > This is error-prone, if other people at a new subclass of `SetVar`, this method may be missed. How about add a `SetVarType` or something to specify the stmt's type directly? And write like `return var.setType() == Type.SET_SESSION_VAR`;

Yes, I will modify it.

fe/fe-core/src/main/java/org/apache/doris/qe/MasterOpExecutor.java
Cai-Yao: > I think we can set remove the `global` property in `SetStmt` and than call `executor.execute()` method. No need to create a new method `setSessionVars()`.

Ok, I see what you mean.

yiguolei:
run buildall

morningman (approved):
LGTM

PR #18620, opened 2023-04-13
[improve](jdbc)Support bytea type in postgres jdbc catalog

LemonLiTree:
Support bytea type in postgres jdbc catalog.


LemonLiTree:
run buildall

PR #18642, opened 2023-04-13
[bugfix](GEO)fix precision problem

LemonLiTree:
image
image
This is because S2LatLng and S2Point have different representations and the conversion is lossy.


LemonLiTree:
run buildall

Gabriel39 (approved):
LGTM

PR #18666, opened 2023-04-14
[Improve](fe)Use commons-lang3 uniformly and refactor PatternGenerator#generateTypePattern

CalvinKirs:
# Proposed changes

Issue Number: close #xxx

## Problem summary
`commons-lang `(1and2) is no longer maintained since 2011, and the official recommendation is `commons-lang3`, which can be smoothly upgraded to be compatible with `commons-lang`.
We use both dependencies in `fe`, which can be completely unified.

 `PatternGenerator#generateTypePattern `has many meaningless loops, and IntegerRange is introduced for, which is unnecessary. So I refactored it.

## Describe your changes.

- Use commons-lang3 uniformly
- Add the check that prohibits commons-lang in checkstyle 
- Refactor PatternGenerator#generateTypePattern to reduce meaningless method loops
## Checklist(Required)

## Further comments


CalvinKirs:
todo add UT

924060929:
run buildall

924060929:
The multiple loop in the PatternGenerator#generateTypePattern is for readability, this function will invoke in the compile time so is not the hot point.

Maybe we should use `java.util.stream.IntStream.range()` to replace it

CalvinKirs:
> The multiple loop in the PatternGenerator#generateTypePattern is for readability, this function will invoke in the compile time so is not the hot point.↳
I know, I just replace `IntegerRange`,but even with the current approach, I don't think it has increased the complexity of the code. right?
> 
> Maybe we should use `java.util.stream.IntStream.rangeClosed()` to replace the `IntRange`↳

924060929:
> I know, I just replace `IntegerRange`,but even with the current approach, I don't think it has increased the complexity of the code. right?

ok

CalvinKirs:
run buildall

morningman (approved):
LGTM

PR #18688, opened 2023-04-14
[fix](replica) Fix inconsistent replica id between FE and BE

platoneko:
The replica id is only sent to BE after Doris 1.2, while FE always has a replica id for each replica. However, due to some bugs, replica id of some tablets on BE was forgotten to be set or updated, and the inventory BE tablets before upgrading to 1.2 did not have a replica id (which means `replica_id` is 0), so they became inconsistent.
This PR fixes all known bugs that can cause inconsistent replica id, and provides a backup mechanism that automatically detects and synchronizes inconsistent replica id to cope with inventory data.

### DIFF
--- a/be/src/agent/task_worker_pool.cpp
+++ b/be/src/agent/task_worker_pool.cpp
@@ -295,6 +295,9 @@ bool TaskWorkerPool::_register_task_info(const TTaskType::type task_type, int64_
         // no need to report task of these types
         return true;
     }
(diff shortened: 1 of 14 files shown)


platoneko:
run buildall

platoneko:
run buildall

platoneko:
run buildall

platoneko:
run buildall

dataroaring (approved):
LGTM

PR #18696, opened 2023-04-15
[fix](publish) publish go ahead even if quorum is not met

dataroaring:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
@@ -941,7 +941,20 @@ public class DatabaseTransactionMgr {
                                 }
                             }
 
-                            if (healthReplicaNum < quorumReplicaNum) {
+                            if (transactionState.isPublishTimeout() && healthReplicaNum > 0) {
+                                // We can not do any thing except retrying,
+                                // because publish task is assigned a version,
+                                // and doris does not permit discontinuous
+                                // versions.
+                                //
+                                // If a timeout happens, it means that the rowset
+                                // that are being publised exists on a few replicas we should go
+                                // ahead, otherwise data may be lost and thre
+                                // publish task hangs forever.
+                                LOG.info("publish version timeout for transaction {} on tablet {},"
+                                                + " with {} replicas",
+                                        transactionState, tablet, healthReplicaNum, quorumReplicaNum);
+                            } else if (healthReplicaNum < quorumReplicaNum) {
                                 LOG.info("publish version failed for transaction {} on tablet {},"
                                                 + " with only {} replicas less than quorum {}",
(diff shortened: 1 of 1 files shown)


fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
```diff
@@ -941,7 +941,20 @@ public void finishTransaction(long transactionId, Set<Long> errorReplicaIds) thr
...
 
-                            if (healthReplicaNum < quorumReplicaNum) {
+                            if (transactionState.isPublishTimeout() && healthReplicaNum > 0) {
+                                // We can not do any thing except retrying,
+                                // because publish task is assigned a version,
+                                // and doris does not permit discontinuous
+                                // versions.
+                                //
+                                // If a timeout happens, it means that the rowset
+                                // that are being publised exists on a few replicas we should go
```
morningman, line 951:
```suggestion
                                // that are being published exists on a few replicas we should go
```

fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
```diff
@@ -941,7 +941,20 @@ public void finishTransaction(long transactionId, Set<Long> errorReplicaIds) thr
...
+                                // ahead, otherwise data may be lost and thre
```
morningman, line 952:
```suggestion
                                // ahead, otherwise data may be lost and the
```

morningman: And why data will be lost?

dataroaring:
run buildall

fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
dataroaring: e.g. a be crashes and create a new replica on another be.

fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
```diff
@@ -941,7 +941,20 @@ public void finishTransaction(long transactionId, Set<Long> errorReplicaIds) thr
                                 }
                             }
 
-                            if (healthReplicaNum < quorumReplicaNum) {
+                            if (transactionState.isPublishTimeout() && healthReplicaNum > 0) {
```
morningman, line 944: I suggest to add a config to enable this feature, and default is disabled.
Because the clone speed may not catch up the load speed, causing data with single replica for a long time.

PR #18816, opened 2023-04-19
[Enchancement](auth) Forbid to login doris from 127.0.0.1 without password

Cai-Yao:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Handle the case that user forgot the password

### Before

root and admin user is allowed to login from 127.0.0.1, in case user forget password

### After

add a new FE config `skip_auth_check`, the default is false.
If true, the auth check will be skipped.so that login to Doris without a password.

## Checklist(Required)

## Further comments


fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/Auth.java
```diff
@@ -171,8 +172,8 @@ public void mergeRolesNoCheckName(List<String> roles, Role savedRole) throws Ddl
      */
     public void checkPassword(String remoteUser, String remoteHost, byte[] remotePasswd, byte[] randomString,
             List<UserIdentity> currentUser) throws AuthenticationException {
-        if ((remoteUser.equals(ROOT_USER) || remoteUser.equals(ADMIN_USER)) && remoteHost.equals("127.0.0.1")) {
-            // root and admin user is allowed to login from 127.0.0.1, in case user forget password.
+        if ((remoteUser.equals(ROOT_USER) || remoteUser.equals(ADMIN_USER)) && Config.skip_auth_check) {
```
CalvinKirs, line 175: I haven't looked at the calling process, but generally speaking, we recommend constant.equals to prevent NPE.
it's difficult for us to check whether the caller has done the corresponding null value detection.

```
eg: ROOT_USER.equals(remoteUser)
```

Cai-Yao: OK, I will modify this.

docs/en/docs/admin-manual/privilege-ldap/user-privilege.md
```diff
@@ -236,9 +236,9 @@ ADMIN_PRIV and GRANT_PRIV have the authority of **"grant authority"** at the sam
 
 5. Forget passwords
 
-	If you forget your password and cannot log in to Doris, you can log in to Doris without a password using the following command on the machine where the Doris FE node is located:
+	If you forget your password and cannot log in to Doris, you can add `skip_auth_check` in fe config so that logging to Doris without a password in localhost.
 
-	`mysql-client -h 127.0.0.1 -P query_port -uroot`
+	`skip_auth_check = true`
```
yiguolei, line 241: change to  skip_localhost_auth_check

yiguolei:
run buildall

yiguolei (approved):
LGTM

docs/en/docs/admin-manual/privilege-ldap/user-privilege.md
```diff
@@ -236,9 +236,9 @@ ADMIN_PRIV and GRANT_PRIV have the authority of **"grant authority"** at the sam
 
 5. Forget passwords
 
-	If you forget your password and cannot log in to Doris, you can log in to Doris without a password using the following command on the machine where the Doris FE node is located:
+	If you forget your password and cannot log in to Doris, you can add `skip_localhost_auth_check` in fe config so that logging to Doris without a password in localhost.
```
morningman, line 239: Do we need to restart FE? If yes, please add it in doc

fe/fe-common/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -2116,5 +2116,12 @@ public class Config extends ConfigBase {
      */
     @ConfField(mutable = true)
     public static boolean infodb_support_ext_catalog = false;
+
+    /**
+     * If true, auth check will be disabled. The default value is false.
+     * This is to solve the case that user forgot the password.
+     */
+    @ConfField(mutable = true)
```
morningman, line 2124: The mutable should be `false`, otherwise user can set it via http api

docs/en/docs/admin-manual/privilege-ldap/user-privilege.md
Cai-Yao: Need restart FE

Cai-Yao:
https://github.com/apache/doris/pull/18996

PR #18822, opened 2023-04-19
[fix](schema_change) remove shadow prefix of schema for tablesink

dataroaring:
LSC updates tablet's schema in writing. Be optimized adding columns via linked schema change and it distinguishes adding by comparing column name. e.g. if new column's name is not found in old schema, then it is a newly-add column.

When a table is under schema-changing, it adds __doris_shadow_ prefix in name of columns in shadow index. Then  writes during schema-changing would bring schema with __doris_shadow_ to be. If schema change request arrives at be after writes, then be do it as a add-column schema change due to __doris_shadow_ is not in base tablet.

# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/OlapTableSink.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/OlapTableSink.java
@@ -212,7 +212,7 @@ public class OlapTableSink extends DataSink {
             List<String> columns = Lists.newArrayList();
             List<TColumn> columnsDesc = Lists.newArrayList();
             List<TOlapTableIndex> indexDesc = Lists.newArrayList();
-            columns.addAll(indexMeta.getSchema().stream().map(Column::getName).collect(Collectors.toList()));
+            columns.addAll(indexMeta.getSchema().stream().map(Column::getNonShadowName).collect(Collectors.toList()));
             for (Column column : indexMeta.getSchema()) {
                 TColumn tColumn = column.toThrift();
                 column.setIndexFlag(tColumn, table);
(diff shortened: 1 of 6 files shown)


dataroaring:
run buildall

dataroaring:
run buildall

dataroaring:
run buildall

dataroaring:
run buildall

dataroaring:
run buildall

fe/fe-core/src/main/java/org/apache/doris/planner/OlapTableSink.java
```diff
@@ -212,7 +212,7 @@ private TOlapTableSchemaParam createSchema(long dbId, OlapTable table) {
             List<String> columns = Lists.newArrayList();
             List<TColumn> columnsDesc = Lists.newArrayList();
             List<TOlapTableIndex> indexDesc = Lists.newArrayList();
-            columns.addAll(indexMeta.getSchema().stream().map(Column::getName).collect(Collectors.toList()));
+            columns.addAll(indexMeta.getSchema().stream().map(Column::getNonShadowName).collect(Collectors.toList()));
```
morningman, line 215: The column with `__doris_shadow_` prefix can only be created when doing `modify column`.
And most of `modify column` operation is not a "light schema change" operation.
But there is an exception, which is doing "modify varchar column's length", which is a `modify column` operation
and will do "light schema change".

So I think we should handle it in `SchemaChangeHandler.processModifyColumn()`, to avoid creating column with
shadow prefix if it is a light schema change, instead of modifying here.

And on BE side, there will still be column with shadow prefix, eg, modify column without light schema change

morningman: Oh, I see.
The `columns` here is for loading task, and the schema will be saved in rowset's meta(not tablet's meta).
And for rowset meta, there should be no column with `shadow` prefix.

morningman (approved):
LGTM

PR #18856, opened 2023-04-20
[enhance](Nereids): When requiredSlots is empty, prune child previously in ColumnPruning

jackwener:
# Proposed changes

Issue Number: close #xxx

## Problem summary

When requiredSlots is empty, prune child previously in ColumnPruning.

A example:

```
Agg count(*)
- Filter age > 3
-- Scan id, age
```

We should get 

```
Agg count(*)
- Project age
-- Filter age > 3
--- Scan id, age
```

instead of

```
Agg count(*)
- Project age, id
-- Filter age > 3
--- Scan id, age
```

## Checklist(Required)

## Further comments


924060929:
column pruning has lots of special cases, and can not image all of the cases, so I can not review this pr.
but in my experience, this code more likely make the regression test, because I remember that I've write the code like you and failed.
when you run into the bug, you can unify abstraction and modification based on all known cases.

Issue #16547, opened 2023-02-09
[Roadmap] Support Struct+Map Type


xy720:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

## Struct + Map todo list

- [x] 语法支持 **(2.0)**
    - [x] 建表+desc语句
    - [x] StructLiteral
	- [x] MapLiteral
- [x] Insert **(2.0)**
    - [x] Map
	- [x] Struct
- [x] Select **(2.0)**
    - [x] Map
	- [x] Struct
- [ ] Delete **(2.0)**
    - [ ] Map
    - [ ] Struct
- [ ] 子类型支持 **(2.0)**
    - [ ] struct field 支持 scala type
	- [ ] struct field 支持 nullable
	- [ ] map key/value 支持 scala type
- [ ] Struct嵌套类型支持
	- [ ] Array 嵌套 Struct
	- [ ] Struct 嵌套 Struct
	- [ ] Struct 嵌套 Array
- [ ] Map嵌套类型支持
	- [ ] value 嵌套 Array
	- [ ] value 嵌套 Struct
	- [ ] value 嵌套 Map
- [ ] Cast **(2.0)**
    - [ ] Struct 隐式 Cast 函数
	- [ ] Map 隐式 Cast 函数
	- [ ] 显式 Struct CastTo Struct
	- [ ] 显式 Map CastTo Map
	- [ ] 显式 StructLiteral CastTo StructLiteral
	- [ ] 显式 MapLiteral CastTo MapLiteral
- [ ] 导入 **(2.0)**
- Struct
	- [x] Json格式
	- [ ] csv格式
	- [ ] parquet格式
	- [ ] orc格式
- Map
	- [ ] Json格式
	- [ ] csv格式
	- [ ] parquet格式
	- [ ] orc格式
- [ ] 导出
- Struct
	- [ ] select into outfile
	- [ ] export
- Map
	- [ ] select into outfile
	- [ ] export
- [ ] 函数
- Struct
	- [ ] struct(val1, val2, val3, ...)
	- [ ] named_struct(name1, val1, name2, val2, ...)
	- [ ] s.x
- Map
	- [ ] map(key1, val1, key2, val2, ...)
	- [ ] M[key]/element_at
	- [ ] size/cardinality(Map<K,V>)
	- [ ] map_keys(Map<K,V>)
	- [ ] map_values(Map<K,V>)
[...]

PR #19045, opened 2023-04-25
[feature](struct-type) support struct_element function

xy720:
# Proposed changes

This commit support a function allows return a field column in struct column.
Since the function can return any type, this commit also supports `ANY_STRUCT_TYPE` 
and `ANY_ELEMENT_TYPE`. 

#16547 

example:

```
mysql> select struct_element(named_struct('f1', 1, 'f2', 'a'), 'f2');
+--------------------------------------------------------+
| struct_element(named_struct('f1', 1, 'f2', 'a'), 'f2') |
+--------------------------------------------------------+
| a                                                      |
+--------------------------------------------------------+
1 row in set (0.03 sec)

mysql> select struct_element(named_struct('f1', 1, 'f2', 'a'), 1);
+-----------------------------------------------------+
| struct_element(named_struct('f1', 1, 'f2', 'a'), 1) |
+-----------------------------------------------------+
|                                                   1 |
+-----------------------------------------------------+
1 row in set (0.02 sec)

mysql> select struct_col, struct_element(struct_col, 'f1') from test_struct;
+-------------------------------------------------+-------------------------------------+
| struct_col                                      | struct_element(`struct_col `, 'f1') |
+-------------------------------------------------+-------------------------------------+
| {1, 2, 3, 4, 5}                                 |                                   1 |
| {1, 1000, 10000000, 100000000000, 100000000000} |                                   1 |
| {5, 4, 3, 2, 1}                                 |                                   5 |
| NULL                                            |                                NULL |
| {1, NULL, 3, NULL, 5}                           |                                   1 |
+-------------------------------------------------+-------------------------------------+
9 rows in set (0.01 sec)
```

## Problem summary

Describe your changes.

## Checklist(Required)

## Further comments


xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run beut

eldenmoon:
It seems that `struct_element` will read the whole struct and discard other none related columns，will we implement a function that only read the target column.This could save lots of scan io.

eldenmoon (approved):
LGTM

fe/fe-common/src/main/java/org/apache/doris/catalog/AnyStructType.java
```diff
@@ -0,0 +1,35 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+package org.apache.doris.catalog;
+
+/**
+ *  Describes a AnyStructType type, used for SQL function return type,
+ *   NOT used for table column type.
+ */
+public class AnyStructType extends AnyType {
```
amorynan, line 24: is this describe whole struct type and nested element type ? why not use AnyType?

xy720:
> It seems that `struct_element` will read the whole struct and discard other none related columns，will we implement a function that only read the target column.This could save lots of scan io.

That's a very good idea! How to scan target field is important. But I think what we need is not only a function.
What we should do is that we prune sub fields in struct/map type while generating query plan in Fe.
And then pass the pruned fields into column_reader.new_iterator() function in be, so that we read only the target field in Block, which will be passed to function.
This is very important when we support nested complex type.

fe/fe-common/src/main/java/org/apache/doris/catalog/AnyStructType.java
xy720: We need specific types to describe the return field type and return struct type.
So it is the ANY_ELEMENT_TYPE and ANY_STRUCT_TYPE.
If there is only ANY_TYPE, We cannot distinguish between return field type and return struct type.

cambyzju (approved):
LGTM

PR #19069, opened 2023-04-25
[Enhancement](compaction) support parallel compaction for single tablet

### LINKED ISSUE #18742: [Enhancement] (compaction) support parallel compaction for single replica
### Search before asking

### Description

Currently only one thread can do compaction for a tablet, but some case compaction may cost
a lot of time and we may able to start other thread do compaction.

### Solution

multi-thread compaction for one tablet, modify meta in lock

### Code of Conduct

gitccl:
# Proposed changes

Issue Number: close #18742

### Basic ideas
In this pr, we add support for parallel cumulative compaction for single tablet, and base compaction still runs in single thread.
Firstly, we save all current running cumulative compaction tasks in an array named `cumulative_compactions`, and save current running base compaction task in `base_compaction`.
In the case of multiple cumulative compaction tasks running at the same time, the time and order of completion of the tasks is indeterminate. Therefore, the rowsets that can be compacted by next thread are split into multiple contiguous segments. Every time we want to choose rowsets to compact, we will choose a contiguous segments with maximum score. And in this choose process, we will skip rowsets with large size to avoid their participation in cumulative compaction.
  We also change the behavior of `update_cumulative_point`. In `update_cumulative_point`, we will forward `cumulative_point` as far as possible.

### The use of lock
Since the clone task and the compression task cannot be performed at the same time, we use `shared_mutex` to ensure that the execution of the parallel compression task and the clone task is serialized. We also add a mutex `compaction_meta_lock` to protect the meta data of the compaction such as `cumulative_compactions`, `base_compaction` and `cumulative_point`. 
The following describes the detailed use of locks in each function
1. In `Tablet::calc_compaction_score`, we hold `compaction_meta_lock` and `meta_lock`, thus we can safely access `cumulative_compactions`, `cumulative_point`, etc.
2. In `Tablet::prepare_compaction_and_calculate_permits`, before call `prepare_compact`, we will hold `cumulative_compact_meta_lock`. Thus in `prepare_compact`, we can safely access `cumulative_compactions`, `cumulative_point`, etc. **The reason we hold `cumulative_compact_meta_lock` before calling `prepare_compact` because we need to ensure that choosing the rowsets to compact and adding the compaction task to `cumulative_compactions` must be an atomic operation.** It is the same for base compaction, we also need to get `cumulative_compact_meta_lock` first, and then check whether `base_compaction` is null, if it is not null, it means that base compaction is already running (e.g. triggered by http request).
3. In `CumulativeCompaction::execute_compact_impl`, we will try to get reader lock of `cumulative_compaction_lock` first, and if we can't get it, that means the tablet is under clone. After `do_compaction`, we will hold `compaction_meta_lock` and call `update_cumulative_point` to safely forward `cumulative_point`.
4. In `EngineCloneTask::_finish_clone`, in addition to acquiring `base_compaction_lock`, `cumulative_compaction_lock`(write lock), we also acquire the `cumulative_compact_meta_lock` to access current running compaction tasks and update their `is_clone_occurred` field.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/http/action/compaction_action.cpp
+++ b/be/src/http/action/compaction_action.cpp
@@ -208,7 +208,7 @@ Status CompactionAction::_execute_compaction_callback(TabletSharedPtr tablet,
                 res = Status::Error<TRY_LOCK_FAILED>();
             } else {
                 StorageEngine::instance()->create_base_compaction(tablet, base_compaction);
-                Status res = base_compaction->prepare_compact();
+                res = base_compaction->prepare_compact();
                 if (res.ok()) {
                     tablet->set_base_compaction(base_compaction);
                     need_reset = true;
@@ -240,7 +240,7 @@ Status CompactionAction::_execute_compaction_callback(TabletSharedPtr tablet,
         bool need_reset = false;
         {
             std::unique_lock compaction_meta_lock(tablet->get_compaction_meta_lock());
-            Status res = cumu_compaction->prepare_compact();
+            res = cumu_compaction->prepare_compact();
             if (res.ok()) {
                 tablet->add_cumulative_compaction_unlocked(cumu_compaction);
                 need_reset = true;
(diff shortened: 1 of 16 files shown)


gitccl:
run buildall

gitccl:
run buildall

gitccl:
run buildall

PR #19167, opened 2023-04-27
[feature](table) implement the round robin selection be when create tablet

luwei16:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

## Further comments


fe/fe-core/src/main/java/org/apache/doris/datasource/InternalCatalog.java
```diff
@@ -2538,6 +2541,13 @@ private void createTablets(String clusterName, MaterializedIndex index, ReplicaS
         if (chooseBackendsArbitrary) {
             backendsPerBucketSeq = Maps.newHashMap();
         }
+
+        Map<Tag, Integer> nextIndexs = new HashMap<>();
+        for (Map.Entry<Tag, Short> entry : replicaAlloc.getAllocMap().entrySet()) {
+            // TODO: random roundrobin starting position
+            nextIndexs.put(entry.getKey(), 0);
```
morningman, line 2548: If always starts with 0, it will causing the first node always has more tablets than others.
If intro random start position, than it will be the same as previous implementation

morningman: At very early time, Doris use RR to distribute tablets.
There was a field `nextBEIdx` in `SystemInfoService`, to save the start position.

But it is hard to maintain that value, because we have to consider `tag`, `storage_medium`, `cluster`, etc. So you have to maintain the `idx` for each combination.

So we changed the distribution method to totally random.

morningman: I think of an implementation, for reference:

1. Select all candidate BE list according to the conditions (`tag`, `storage_medium`, etc.).
2. In the list, select the BE with the least number of tablets as the starting node to start executing RoundRobin.

This method should be logically clearer, which is to separate the step of "filtering BE by condition" from the "execution of RR".

dataroaring:
run buildall

hello-stephen:
 the sum of best hot time: 33.93 seconds
 stream load tsv:          426 seconds loaded 74807831229 Bytes, about 167 MB/s
 stream load json:         24 seconds loaded 2358488459 Bytes, about 93 MB/s
 stream load orc:          59 seconds loaded 1101869774 Bytes, about 17 MB/s
 https://doris-community-test-1308700295.cos.ap-hongkong.myqcloud.com/tmp/20230429151038_clickbench_pr_137382.html

fe/fe-core/src/main/java/org/apache/doris/datasource/InternalCatalog.java
luwei16: > I think of an implementation, for reference:
> 
> 1. Select all candidate BE list according to the conditions (`tag`, `storage_medium`, etc.).
> 2. In the list, select the BE with the least number of tablets as the starting node to start executing RoundRobin.
> 
> This method should be logically clearer, which is to separate the step of "filtering BE by condition" from the "execution of RR".

done

morningman (dismissed):
LGTM

morningman:
run buildall

morningman (approved):
LGTM

PR #19317, opened 2023-05-05
[fix](tpch case) fix tpch case failure caused by casting decimal to double

jacktengg:
# Proposed changes

Issue Number: close #xxx

## Problem summary

e.g. for test case tpch_sf100_unique_sql_p2/sql/q15.sql, it fails randomly because  result of `SELECT max(total_revenue) FROM revenue1;`  is not stable, caused by casting decimal to double when doing agg.

This PR change decimal to decimalv3 to avoid this problem.

## Checklist(Required)

## Further comments


jacktengg:
run buildall

PR #19392, opened 2023-05-08
[Enhancement](broker-load) broker load show stmt support display cluster name if specified

### LINKED ISSUE #19222: [Enhancement] Make the 'cluster' in task info for LoadStmt not empty
### Search before asking

### Description

Produce a S3 load, and run show load stmt. As seen, `cluster` is always N/A:
```
         JobId: 10131
         Label: lineitem_load
         State: LOADING
      Progress: ETL:100%; LOAD:77%
          Type: BROKER
       EtlInfo: NULL
      TaskInfo: cluster:N/A; timeout(s):14400; max_filter_ratio:0.0
      ErrorMsg: NULL
    CreateTime: 2023-04-29 12:23:29
  EtlStartTime: 2023-04-29 12:23:30
 EtlFinishTime: 2023-04-29 12:23:30
 LoadStartTime: 2023-04-29 12:23:30
LoadFinishTime: NULL
           URL: NULL
    JobDetails: {"Unfinished backends":{"de827555cbb2413d-a720be8937596b35":[10021]},"ScannedRows":36165536,"TaskNumber":1,"LoadBytes":6140395389,"All backends":{"de827555cbb2413d-a720be8937596b35":[10021]},"FileNumber":1,"FileSize":7895697364}
 TransactionId: 3
  ErrorTablets: {}
          User: root
       Comment: 
```

### Solution

Actually, the `getCluster` method in LoadStmt, as the only access to the passed in `cluster`, is never called.

### Code of Conduct

lvshaokang:
# Proposed changes

Issue Number: close #19222 

## Problem summary

Describe your changes.

## Checklist(Required)

## Further comments


lvshaokang:
run buildall

fe/fe-core/src/main/java/org/apache/doris/load/loadv2/BrokerLoadJob.java
```diff
@@ -77,17 +77,19 @@ public class BrokerLoadJob extends BulkLoadJob {
...
+    private String cluster;
 
     // for log replay and unit test
     public BrokerLoadJob() {
         super(EtlJobType.BROKER);
     }
 
     public BrokerLoadJob(long dbId, String label, BrokerDesc brokerDesc,
-                         OriginStatement originStmt, UserIdentity userInfo)
+                         String cluster, OriginStatement originStmt, UserIdentity userInfo)
```
TangSiyang2001, line 88: Maybe we could merge the field `cluster` into BrokerDesc temporarily, and in our Load refactor task, field `cluster` can be included under the semantic of `resource_desc`, what's ur opinion?

If agreement reached, then the access to cluster can become getBrokerDesc().getCluster() (or getResourceDesc().getCluster() after our load refactor task).

fe/fe-core/src/main/java/org/apache/doris/load/loadv2/BrokerLoadJob.java
```diff
@@ -387,4 +389,9 @@ public void afterVisible(TransactionState txnState, boolean txnOperated) {
         super.afterVisible(txnState, txnOperated);
         writeProfile();
     }
+
+    @Override
+    protected String getResourceName() {
+        return cluster != null ? cluster : "N/A";
+    }
```
TangSiyang2001, line 396: And another way that I think may be tricky is to imitate SparkLoadJob that we just use the broker name. What's ur opinion?

TangSiyang2001: As for the original `cluster` field in LoadStmt, do u it seems a little redundant?

lvshaokang:
@TangSiyang2001 As I  can see, SparkLoadJob's cluster uses `resource.getName`, which declares a required field for a resource.
But the `BY cluster` for BrokerLoad is an optional item, which only serves as a flag, and is not called anywhere else.
Also, I looked at the syntax commit record for `BY cluster` and I think maybe it is a historical reserved field and there should be no compatibility issues if removing it.
Another to discuss it, if we use the second way, the broker load has a name attribute(`WITH Broker name`) , while the s3 load does not have a name attribute, and when it doesn't, we use the default value`default_cluster` as the cluster.
So, there are 2 point need discussed:
1.remove `by cluster` syntax support
2.use broker name as cluster
Looking forward your reply~

TangSiyang2001:
@lvshaokang Sry for late reply. 

1. As far as I can see, since the original implementation did not use the `cluster` field at all, it means that `BY cluster` clause maybe deprecated. Personally, I think maybe it does no harm to rm this syntax. And the [related docs](https://doris.apache.org/zh-CN/docs/dev/sql-manual/sql-reference/Data-Manipulation-Statements/Load/BROKER-LOAD) seems not mention it as well. If worried about compatibility, maybe we could keep the syntax but do not analyze it as it does originally, and show cluter info with other helpful info.

2. As for S3, maybe we could show sth unique to a S3 cluster like uri? Looking forward to ur opinion.

morningman:
We can simple remove this field.
Because `cluster` is only for hadoop dpp load, which is used in Badiu internally and never open source.
So I suggest to just remove this "cluster" field.

lvshaokang:
> We can simple remove this field. Because `cluster` is only for hadoop dpp load, which is used in Badiu internally and never open source. So I suggest to just remove this "cluster" field.

Thanks for you advice, I get it.

lvshaokang:
run buildall

morningman (approved):
LGTM

lvshaokang:
It's ready. PTAL. Thanks. cc @morningman @TangSiyang2001

Hastyshell (approved):
LGTM

PR #19464, opened 2023-05-09
[typo](doc)optimize description of bitmap in materialized view document

figurant:
original description misleading users，some user asked that bitmap acceleration in materialized view only support positive number column type，but it's not true。so update it。


luzhijing (approved):
LGTM

luzhijing:
run buildall

PR #19472, opened 2023-05-09
[Improvement](memory) new page use Allocator

xinyiZzz:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Allocator will throw exception when the memory exceeds the limit.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/olap/page_cache.cpp
+++ b/be/src/olap/page_cache.cpp
@@ -64,9 +64,12 @@ bool StoragePageCache::lookup(const CacheKey& key, PageCacheHandle* handle,
     return true;
 }
 
-void StoragePageCache::insert(const CacheKey& key, const Slice& data, PageCacheHandle* handle,
+void StoragePageCache::insert(const CacheKey& key, DataPage<>* data, PageCacheHandle* handle,
                               segment_v2::PageTypePB page_type, bool in_memory) {
-    auto deleter = [](const doris::CacheKey& key, void* value) { delete[] (uint8_t*)value; };
+    auto deleter = [](const doris::CacheKey& key, void* value) {
+        DataPage<>* cache_value = (DataPage<>*)value;
+        delete cache_value;
+    };
 
     CachePriority priority = CachePriority::NORMAL;
     if (in_memory) {
@@ -74,7 +77,7 @@ void StoragePageCache::insert(const CacheKey& key, const Slice& data, PageCacheH
     }
 
     auto cache = _get_page_cache(page_type);
-    auto lru_handle = cache->insert(key.encode(), data.data, data.size, deleter, priority);
+    auto lru_handle = cache->insert(key.encode(), data, data->bytes, deleter, priority);
     *handle = PageCacheHandle(cache, lru_handle);
 }
 
--- a/be/src/olap/page_cache.cpp
+++ b/be/src/olap/page_cache.cpp
@@ -64,10 +64,10 @@ bool StoragePageCache::lookup(const CacheKey& key, PageCacheHandle* handle,
     return true;
 }
 
-void StoragePageCache::insert(const CacheKey& key, DataPage<>* data, PageCacheHandle* handle,
+void StoragePageCache::insert(const CacheKey& key, DataPage* data, PageCacheHandle* handle,
                               segment_v2::PageTypePB page_type, bool in_memory) {
     auto deleter = [](const doris::CacheKey& key, void* value) {
-        DataPage<>* cache_value = (DataPage<>*)value;
+        DataPage* cache_value = (DataPage*)value;
         delete cache_value;
     };
 
--- a/be/src/olap/page_cache.cpp
+++ b/be/src/olap/page_cache.cpp
@@ -77,7 +77,7 @@ void StoragePageCache::insert(const CacheKey& key, DataPage* data, PageCacheHand
     }
 
     auto cache = _get_page_cache(page_type);
-    auto lru_handle = cache->insert(key.encode(), data, data->bytes, deleter, priority);
+    auto lru_handle = cache->insert(key.encode(), data, data->capacity(), deleter, priority);
     *handle = PageCacheHandle(cache, lru_handle);
 }
 
--- a/be/src/olap/page_cache.h
+++ b/be/src/olap/page_cache.h
@@ -28,10 +28,54 @@
 
 #include "olap/lru_cache.h"
 #include "util/slice.h"
+#include "vec/common/allocator.h"
+#include "vec/common/allocator_fwd.h"
 
 namespace doris {
 
 class PageCacheHandle;
+
+template <typename TAllocator>
+class PageBase : private TAllocator {
+public:
+    PageBase() : data(nullptr), size(0), bytes(0) {}
+
+    PageBase(size_t b) : size(b), bytes(b) {
+        data = reinterpret_cast<char*>(TAllocator::alloc(bytes, ALLOCATOR_ALIGNMENT_16));
+        ExecEnv::GetInstance()->page_no_cache_mem_tracker()->consume(bytes);
+    }
+
+    PageBase(PageBase&& other) : data(other.data), size(other.size) {
+        std::swap(bytes, other.bytes);
+    }
+
+    ~PageBase() {
+        if (data != nullptr) {
+            DCHECK(bytes != 0 && size != 0);
+            TAllocator::free(data, bytes);
+            ExecEnv::GetInstance()->page_no_cache_mem_tracker()->release(bytes);
+        }
+    }
+
+    void reset_size(size_t n) {
+        DCHECK(n <= bytes);
+        size = n;
+    }
+
+    void release() {
+        data = nullptr;
+        size = 0;
+        bytes = 0;
+    }
+
+    char* data;
+    // Effective size, smaller than bytes, such as data page remove checksum suffix.
+    size_t size;
+    size_t bytes = 0;
+};
+
+template <typename TAllocator = Allocator<false>>
+using DataPage = PageBase<TAllocator>;
 
 // Wrapper around Cache, and used for cache page of column data
 // in Segment.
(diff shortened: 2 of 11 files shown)


xinyiZzz:
run buildall

xinyiZzz:
run buildall

be/src/olap/page_cache.cpp
```diff
@@ -64,17 +64,22 @@ bool StoragePageCache::lookup(const CacheKey& key, PageCacheHandle* handle,
...
+    };
 
     CachePriority priority = CachePriority::NORMAL;
     if (in_memory) {
         priority = CachePriority::DURABLE;
     }
 
     auto cache = _get_page_cache(page_type);
-    auto lru_handle = cache->insert(key.encode(), data.data, data.size, deleter, priority);
+    DataPage<>* cache_value = new DataPage<>(std::move(data));
```
yiguolei, line 80: 如果这里使用了move的话，那么67行李的datapage 参数应该显示的为 &&，告诉调用者，我这里会move；否则在80行我们move了，外部可能不知道，结果出现core。

be/src/olap/page_cache.h
```diff
@@ -28,11 +28,62 @@
...
+
+    void release() {
+        data = nullptr;
+        size = 0;
+        bytes = 0;
+    }
+
+    char* data;
+    // Effective size, smaller than bytes, such as data page remove checksum suffix.
+    size_t size;
```
yiguolei, line 80: change name to  charge and size, same as LRUCache's definition.

be/src/olap/page_cache.h
```diff
@@ -85,7 +136,7 @@ class StoragePageCache {
     // This function is thread-safe, and when two clients insert two same key
     // concurrently, this function can assure that only one page is cached.
     // The in_memory page will have higher priority.
-    void insert(const CacheKey& key, const Slice& data, PageCacheHandle* handle,
+    void insert(const CacheKey& key, DataPage<>& data, PageCacheHandle* handle,
```
yiguolei, line 139: why using DataPage<> ? Is DataPage ok?

be/src/olap/page_cache.h
```diff
@@ -28,11 +28,62 @@
...
+#include "vec/common/allocator.h"
+#include "vec/common/allocator_fwd.h"
 namespace doris {
 
 class PageCacheHandle;
 
+template <typename TAllocator>
+class PageBase : private TAllocator {
+public:
+    PageBase() : size(0), bytes(0) {}
```
yiguolei, line 40: data is not initialized

be/src/olap/rowset/segment_v2/page_handle.h
```diff
@@ -46,42 +46,49 @@ class PageHandle {
             : _is_data_owner(false), _cache_data(std::move(cache_data)) {}
 
     // Move constructor
-    PageHandle(PageHandle&& other) noexcept
-            : _is_data_owner(false),
-              _data(std::move(other._data)),
-              _cache_data(std::move(other._cache_data)) {
+    PageHandle(PageHandle&& other) noexcept : _cache_data(std::move(other._cache_data)) {
         // we can use std::exchange if we switch c++14 on
```
yiguolei, line 50: _is_data_owner is not inited, do not know it is true or false, so what is swap means?

be/src/olap/rowset/segment_v2/page_handle.h
```diff
@@ -46,42 +46,49 @@ class PageHandle {
...
-            delete[] _data.data;
             ExecEnv::GetInstance()->page_no_cache_mem_tracker()->consume(-_data.size);
+        } else {
+            DCHECK(_data.size == 0);
         }
     }
 
     // the return slice contains uncompressed page body, page footer, and footer size
-    Slice data() const {
+    char* data() const {
```
yiguolei, line 71: why not return slice ?

be/src/olap/rowset/segment_v2/page_handle.h
```diff
@@ -46,42 +46,49 @@ class PageHandle {
...
-        _data = std::move(other._data);
+        _data.swap(other._data);
         _cache_data = std::move(other._cache_data);
         return *this;
     }
 
     ~PageHandle() {
         if (_is_data_owner && _data.size > 0) {
-            delete[] _data.data;
             ExecEnv::GetInstance()->page_no_cache_mem_tracker()->consume(-_data.size);
```
yiguolei, line 64: if this page handle is data owner, why not call delete to release the memory?

be/src/olap/page_cache.cpp
xinyiZzz: move之后`datapage`依然可以正常访问，只是析构时不会delete data[]
不过参数改成&&确实更优雅

be/src/olap/page_cache.h
xinyiZzz: The `size` here is different from LRUCache's `size`. LRUCache's `size = charge + handle_size` is not the difference between the effective size and the actual memory size.

And charge seems to be a concept in lru cache, and page bytes seems to be easier to understand when describing actual memory?

be/src/olap/page_cache.h
xinyiZzz: No, `DataPage` is a template class, `<>` must be specified, even if the template parameter is the default

be/src/olap/page_cache.h
xinyiZzz: done, grace

be/src/olap/rowset/segment_v2/page_handle.h
xinyiZzz: `_is_data_owner` is initialized to false at definition time before the body of the copy constructor is executed

swap has no special meaning, it has the same meaning as a = b, b = false

be/src/olap/rowset/segment_v2/page_handle.h
xinyiZzz: It seems more graceful that DataPage and Slick are independent?

image

be/src/olap/rowset/segment_v2/page_handle.h
xinyiZzz: Because DataPage will automatically call `free` when it is destructed

The memory life cycle is the same as DataPage, and does not rely on manual release, which seems to be more exception-safe

xinyiZzz:
run buildall

xinyiZzz:
run buildall

xinyiZzz:
run buildall

be/src/olap/page_cache.h
```diff
@@ -28,11 +28,55 @@
...
+class PageBase : private TAllocator {
+public:
+    PageBase() : data(nullptr), size(0), bytes(0) {}
+
+    PageBase(size_t b) : size(b), bytes(b) {
+        data = reinterpret_cast<char*>(TAllocator::alloc(bytes, ALLOCATOR_ALIGNMENT_16));
+        ExecEnv::GetInstance()->page_no_cache_mem_tracker()->consume(bytes);
+    }
+
+    PageBase(PageBase&& other) : data(other.data), size(other.size) {
```
yiguolei, line 48: remove this method

be/src/olap/page_cache.h
```diff
@@ -28,11 +28,55 @@
...
+            ExecEnv::GetInstance()->page_no_cache_mem_tracker()->release(bytes);
+        }
+    }
+
+    void reset_size(size_t n) {
+        DCHECK(n <= bytes);
+        size = n;
+    }
+
+    void release() {
```
yiguolei, line 65: is this method useful?

xinyiZzz: fix

be/src/olap/rowset/segment_v2/page_io.cpp
```diff
@@ -143,8 +143,8 @@ Status PageIO::read_and_decompress_page(const PageReadOptions& opts, PageHandle*
     }
 
     // hold compressed page at first, reset to decompressed page later
-    std::unique_ptr<char[]> page(new char[page_size]);
-    Slice page_slice(page.get(), page_size);
+    std::unique_ptr<DataPage<>> page = std::make_unique<DataPage<>>(page_size);
```
yiguolei, line 146: Here

be/src/olap/rowset/segment_v2/page_io.cpp
```diff
@@ -177,12 +177,12 @@ Status PageIO::read_and_decompress_page(const PageReadOptions& opts, PageHandle*
             return Status::Corruption("Bad page: page is compressed but codec is NO_COMPRESSION");
         }
         SCOPED_RAW_TIMER(&opts.stats->decompress_ns);
-        std::unique_ptr<char[]> decompressed_page(
-                new char[footer->uncompressed_size() + footer_size + 4]);
+        std::unique_ptr<DataPage<>> decompressed_page =
```
yiguolei, line 180: Here

be/src/olap/page_cache.h
```diff
@@ -28,11 +28,50 @@
...
+        data = nullptr;
+        size = 0;
+        bytes = 0;
+    }
+
+    char* data;
```
yiguolei, line 67: 可能你得看看类似java 中buffer的一些概念，比如pos，capacity 这些，size和bytes还是太相似了， 这个东东在buffer 结构里有比较通用的定义

xinyiZzz: A buffer's capacity is the number of elements it contains. The capacity of a buffer is never negative and never changes.

A buffer's limit is the index of the first element that should not be read or written. A buffer's limit is never negative and is never greater than its capacity.

A buffer's position is the index of the next element to be read or written. A buffer's position is never negative and is never greater than its limit.

那就像 PodArray 一样，叫 size 和 capacity 好了
image

be/src/olap/page_cache.h
xinyiZzz: fix

be/src/olap/page_cache.h
xinyiZzz: for page_cache_test.cpp, but it can be optimized

xinyiZzz:
run buildall

yiguolei (approved):
LGTM

PR #19492, opened 2023-05-10
[Enhancement] show total transactions in show proc "/transactions"

whutpencil:
# Proposed changes

Issue Number: close #xxx

## Problem summary

In a scenario where multiple DBs are simultaneously imported with high concurrency, a significant number of transactions will be generated. Without a summary field, we cannot clearly see how many transactions there are in the current cluster. Therefore, I have enhanced this point.

```
mysql> show proc "/transactions";
+-------+-----------------------------------+-----------------------+
| DbId     | DbName                                             | RunningTransactionNum |
+-------+-----------------------------------+-----------------------+
| 10002   | default_cluster:xxxx                         | 0                                     |
| 14005   | default_cluster:__internal_schema | 0                                     |
| Total     |  2                                                         | 0                                     |
+-------+-----------------------------------+-----------------------+
3 rows in set (0.02 sec)
```

image

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
@@ -55,6 +55,8 @@ import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -507,12 +511,15 @@ public List<List<Comparable>> getDbInfo() {
...
                 runningNum = dbMgr.getRunningTxnNums() + dbMgr.getRunningRoutineLoadTxnNums();
+                totalRunningNum += runningNum;
             } catch (AnalysisException e) {
                 LOG.warn("get database running transaction num failed", e);
             }
-            info.add(runningNum);
+            info.add(String.valueOf(runningNum));
             infos.add(info);
         }
+        List<String> info = Arrays.asList("Total", String.valueOf(dbIds.size()), String.valueOf(totalRunningNum));
```
morningman, line 521: For the last line, better use same schema for each column:
I suggest：
```
0, "Total", runningnum
```
So that the first column is always integer

whutpencil: @morningman  This is because in the show proc "/cluster_health/tablet_health" command, it was generated in my current format to maintain consistency. Do you think it still needs to be modified?

whutpencil: @morningman done.

morningman (approved):
LGTM

morningman:
run buildall

morningman:
run buildall

PR #19561, opened 2023-05-11
[fix](pipeline) fix query returns empty result occasionally instead of an error after being cancelled

### LINKED ISSUE #19560: [Bug] query returns empty results occasionally after being cancelled
### Search before asking

### Version

master

### What's Wrong?

截屏2023-05-11 20 54 40

The query is cancelled and occasionally returns null results instead of errors.

### What You Expected?

The query returns an error after being cancelled

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

luozenglin:
# Proposed changes

Issue Number: close #19560

## Problem summary

There are two ways for fe to get the cancel msg after be cancels a query:
1.  be's report thread reports to fe;
2. fe updates the status when it gets the query data from be vresult_sink via receiver.

The reason for this problem is that the cancel status is not passed in vresult_sink when it closed after the cancel, causing fe to think that the query has ended normally.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/pipeline/exec/operator.h
+++ b/be/src/pipeline/exec/operator.h
@@ -32,6 +32,7 @@
 #include "common/status.h"
 #include "exec/exec_node.h"
 #include "runtime/memory/mem_tracker.h"
+#include "runtime/runtime_state.h"
 #include "util/runtime_profile.h"
 #include "vec/core/block.h"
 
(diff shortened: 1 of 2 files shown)


Gabriel39 (approved):
LGTM

PR #19723, opened 2023-05-16
fix demo, make it runnable in flink 1.15.3

caisirius:
# Proposed changes

## Problem summary

make the demo DatabaseFullSync work on Flink 1.15.3

## Checklist(Required)

* [ n] Does it affect the original behavior
* [ y] Has unit tests been added
* [ n] Has document been added or modified
* [ n] Does it need to update dependencies
* [ y] Is this PR support rollback (If NO, please explain WHY)

## Further comments

original demo does not really work, I fix :
1. mvn package the fat jar which could directly upload to a clean Flink 
2. bug of SYNC_TBLS

### DIFF
--- a/samples/doris-demo/flink-demo-v1.1/pom.xml
+++ b/samples/doris-demo/flink-demo-v1.1/pom.xml
@@ -37,34 +37,45 @@ under the License.
     </properties>
     <dependencies>
         <dependency>
+            <groupId>junit</groupId>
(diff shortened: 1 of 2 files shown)


JNSimba:
Thank you for your contribution, this is just a code sample reference, not a packaged executable jar.
Synchronize the entire mysql database, adding a more convenient tool, using the side output stream instead of filter, you can refer to here.

PR #19795, opened 2023-05-18
[fix](memory) Remind log if `vm/overcommit_memory`=2 when be start

xinyiZzz:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Expect vm overcommit memory value to be 1,  system will no longer throw bad_alloc, memory alloc are always accepted,
memory limit check is handed over to Doris Allocator, make sure throw exception position is controllable,
otherwise bad_alloc can be thrown anywhere and it will be difficult to achieve exception safety.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/util/mem_info.cpp
+++ b/be/src/util/mem_info.cpp
@@ -300,13 +300,28 @@ void MemInfo::init() {
         _s_sys_mem_available_warning_water_mark = _s_sys_mem_available_low_water_mark + p1;
     }
 
+    // Expect vm overcommit memory value to be 1, system will no longer throw bad_alloc, memory alloc are always accepted,
+    // memory limit check is handed over to Doris Allocator, make sure throw exception position is controllable,
+    // otherwise bad_alloc can be thrown anywhere and it will be difficult to achieve exception safety.
+    std::ifstream sys_vm("/proc/sys/vm/overcommit_memory", std::ios::in);
+    std::string vm_overcommit;
+    getline(sys_vm, vm_overcommit);
+    if (sys_vm.is_open()) sys_vm.close();
+    if (std::stoi(vm_overcommit) == 2) {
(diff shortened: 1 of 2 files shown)


bin/start_be.sh
```diff
@@ -20,6 +20,13 @@ set -eo pipefail
 
 curdir="$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
 
+if [ $(id -u) -eq 0 ]; then # current user has root privileges
+    # system will no longer throw bad_alloc, memory alloc are always accepted,
```
yiguolei, line 24: maybe we should not do that. it's better to print a warning? or add the message to be.out?
but i am not sure.

xinyiZzz: You are right, should not directly modify sys conf
Changed to print log

xinyiZzz:
run buildall

be/src/util/mem_info.cpp
```diff
@@ -300,13 +300,28 @@ void MemInfo::init() {
         _s_sys_mem_available_warning_water_mark = _s_sys_mem_available_low_water_mark + p1;
     }
 
+    // Expect vm overcommit memory value to be 1, system will no longer throw bad_alloc, memory alloc are always accepted,
+    // memory limit check is handed over to Doris Allocator, make sure throw exception position is controllable,
+    // otherwise bad_alloc can be thrown anywhere and it will be difficult to achieve exception safety.
+    std::ifstream sys_vm("/proc/sys/vm/overcommit_memory", std::ios::in);
```
yiguolei, line 306: Add #if os_linux macro to do this check only in linux not macos.

xinyiZzz: Now in #ifndef __APPLE__

yiguolei (approved):
LGTM

PR #19838, opened 2023-05-18
[fix](schema change) Change table state to NORMAL by SchemaChangeJob instead of SchemaChangeHandler

Tanya-W:
# Proposed changes

Issue Number: close #xxx

## Problem summary
fix problem: 
If there is an unfinished schema change job (job-2), and before this time, another schema change job (job-1) of the same table has been finished.
Then restart fe, will replay edit log (pending log and waiting_txn log) for job-2, and the table's state is set to SHCEMA_CHANGE, but when loadAlterJob after replayJournal, will add job-1 to schema change handler, and then run the job-1 will set the table to NORMAL because of job-1 is done, but at this point, the job-2 is doing runWaitingTxnJob, in this function will check table's state, if not normal will throw exception, not change the job's state, and cannot cancel the job because the table is not under schema change.

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/alter/SchemaChangeHandler.java
@@ -1601,33 +1601,7 @@ public class SchemaChangeHandler extends AlterHandler {
                     });
                 }
             }
-
-            if (alterJobsV2.isDone()) {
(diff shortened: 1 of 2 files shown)


yiguolei:
run buildall

Tanya-W:
run buildall

morningman (approved):
LGTM

PR #19912, opened 2023-05-22
[fix](multi-catalog) JDBC Catalog Unknown UNSIGNED type of mysql, type: [DOUBLE]

AshinGau:
# Proposed changes

Fix error like `JDBC Catalog Unknown UNSIGNED type of mysql, type: [DOUBLE]`
As of MySQL 8.0.17, the UNSIGNED attribute is deprecated for columns of type FLOAT, DOUBLE, and DECIMAL (and any synonyms)(https://dev.mysql.com/doc/refman/8.0/en/numeric-type-syntax.html). 
Unsigned floating&double numbers do not increase the precision range. MySQL only checks whether the number is non negative, so reading data with signed float&double numbers will not cause errors.

## Checklist(Required)

## Further comments


morningman (approved):
LGTM

PR #20004, opened 2023-05-24
[opt](Nereids) restore the set_var hint after finish the execution

Reminiscent:
# Proposed changes
Before the change:
```
mysql> SET enable_nereids_planner=true;

mysql> explain select /*+ SET_var(enable_nereids_planner = false) */ year_floor(cast('2023-04-28' as date));
-- omit the result here
10 rows in set (0.01 sec)

mysql> select @@enable_nereids_planner;
+--------------------------+
| @@enable_nereids_planner |
+--------------------------+
|                        0 |
+--------------------------+
1 row in set (0.00 sec)
```

After the change:
```
mysql> SET enable_nereids_planner=true;

mysql> explain select /*+ SET_var(enable_nereids_planner = false) */ year_floor(cast('2023-04-28' as date));
-- omit the result here
10 rows in set (0.14 sec)

mysql> select @@enable_nereids_planner;
+------+
| TRUE |
+------+
|    1 |
+------+
1 row in set (0.25 sec)
```

## Problem summary
We have already recorded the old session vars when we use the `Nereids` to handle the `set_var` hint.
But after we change the optimizer to the old one, it will handle the `set_var` hint again. But it has already taken effect before. So the old value has already changed. But we will use the changed value to overwrite again.

Describe your changes.
We will check the old session var value when we want to record it first. If there exists the value, just skip it.

## Checklist(Required)

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/SessionVariable.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/SessionVariable.java
@@ -1585,6 +1585,10 @@ public class SessionVariable implements Serializable, Writable {
     }
 
     public void addSessionOriginValue(Field key, String value) {
+        if (sessionOriginValue.containsKey(key)) {
+            // If we already set the origin value, just skip the reset.
+            return;
+        }
         sessionOriginValue.put(key, value);
     }
 


Reminiscent:
run buildall

morrySnow (approved):

PR #20022, opened 2023-05-24
[improvement](jdbc catalog)better handling of postresql bit(1) types with bool type

zy-kkk:
# Proposed changes

Issue Number: close #xxx

## Problem summary

When the postgresql bit type size is 1, it reads as a java.lang.boolean via jdbc, and if we match against string, it will display true or false. But the normal display should be a number, so when I detect that the size of bit is 1, I will match it with boolean

## Checklist(Required)

## Further comments


zy-kkk:
run buildall

zy-kkk:
run buildall

morningman (approved):
LGTM

PR #20064, opened 2023-05-25
[feature](load) add test property 'dry_run_load' for olap table

### LINKED ISSUE #19259: [Enhancement] support creating a table with 0 replicas
### Search before asking

### Description

We need a table without data in two scenarios.

1. when investigating performance issues of loads. e.g. I want to know performance except sink.
2. keep materialized views while do not keep data in a duplicate table.

### Solution

- [ ] supporting a table with 0 replica in FE.
- [ ] do not sink to a table with 0 replica in vtablet_sink, e.g. just returns OK in VtabletSink::send
- [ ] add regression test case for 2 cases in description above.

### Code of Conduct

nextdreamblue:
add table property 'dry_run_load', and this table will not save data really, and will return ok when sink table.

for some test case.

Issue Number: close #19259

## Problem summary

Describe your changes.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/vec/sink/vtablet_sink.cpp
+++ b/be/src/vec/sink/vtablet_sink.cpp
@@ -551,6 +551,10 @@ Status VNodeChannel::add_block(vectorized::Block* block, const Payload* payload,
     if (payload->second.empty()) {
         return Status::OK();
     }
+    if (_parent->_dry_run_load) {
+        VLOG_DEBUG << "_dry_run_load == true, do not send data really";
+        return Status::OK();
+    }
     // If add_block() when _eos_is_produced==true, there must be sth wrong, we can only mark this channel as failed.
     auto st = none_of({_cancelled, _eos_is_produced});
     if (!st.ok()) {
@@ -671,6 +675,10 @@ int VNodeChannel::try_send_and_fetch_status(RuntimeState* state,
                                             std::unique_ptr<ThreadPoolToken>& thread_pool_token) {
     auto st = none_of({_cancelled, _send_finished});
     if (!st.ok()) {
+        return 0;
+    }
+    if (_parent->_dry_run_load) {
+        VLOG_DEBUG << "_dry_run_load == true, do not send data really";
         return 0;
     }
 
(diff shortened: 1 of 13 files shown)


nextdreamblue:
run buildall

morningman:
The name `return_load_in_sink` is related to internal implementation, not suitable for a user-oriented config' name.

There are 2 issues you want to solve in #19259

1. Test load performance
    I think it is better to use a session variable or load properties for this case. You can see there a session variable `dry_run_query`, when set to true, it will run the query without return the result. It resolve the similar issue as you want to resolve.

2. For materialized view.
    This is indeed a table property. I suggest to think about new property name instead of exposing the internal implements.
    Eg: `"storage_policy" = "no_storage"`. (There is already a "storage_policy" used for cold-hot data separation, maybe we can expend it)

All in all, my opinion is that the 2 things you want to resolve are 2 different cases. Better design it carefully to avoid adding confusing properties.

nextdreamblue:
> The name `return_load_in_sink` is related to internal implementation, not suitable for a user-oriented config' name.
> 
> There are 2 issues you want to solve in #19259
> 
> 1. Test load performance
>    I think it is better to use a session variable or load properties for this case. You can see there a session variable `dry_run_query`, when set to true, it will run the query without return the result. It resolve the similar issue as you want to resolve.
> 2. For materialized view.
>    This is indeed a table property. I suggest to think about new property name instead of exposing the internal implements.
>    Eg: `"storage_policy" = "no_storage"`. (There is already a "storage_policy" used for cold-hot data separation, maybe we can expend it)
> 
> All in all, my opinion is that the 2 things you want to resolve are 2 different cases. Better design it carefully to avoid adding confusing properties.

1. I discussed with dataroaring before that session variable may affect all users and tables, and users may not be aware of the data not being loaded in a timely manner, which may result in unexpected data loss and high risks. It is recommended to use table properties, and I will modify a user-friendly name.

2. For materialized view, i will implement it in another branch and may introduce students from Doris-future to participate. and will follow your suggestion that add a table property

nextdreamblue:
run buildall

nextdreamblue:
run buildall

nextdreamblue:
run COMPILE

nextdreamblue:
run buildall

nextdreamblue:
run buildall

nextdreamblue:
run buildall

nextdreamblue:
run buildall

PR #20066, opened 2023-05-25
[fix](resource-group) Fix resource group memory isolation may release too much memory

luozenglin:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Suppose three queries are executed in a resource group with a memory_limit of 8G, and they consume memory of query_a = 3G, query_b = 3G, and query_c = 3G. The total memory used is counted as 9G when the resource group GC is executed, which exceeds the resource group limit and cancels query_a. 

When the resource group is next GC, the memory of query_a may not be freed yet, and it will be counted again in the total memory consumed by that resource group, which again exceeds the resource group limit and cancels query_b. 

From the user's perspective, it is fine to execute query_a and query_b at the same time, but executing query_ a, query_b and query_c will be cancelled for two queries, which is not as expected.

This pr skips the queries that are cancelled when counting the memory used by the resource group. If this causes the process memory to grow, the process gc will handle it.

## Checklist(Required)

## Further comments

### DIFF
--- a/be/src/runtime/query_context.h
+++ b/be/src/runtime/query_context.h
@@ -109,6 +109,9 @@ public:
             _is_cancelled = is_cancelled;
             _ready_to_execute = true;
         }
+        if (query_mem_tracker && is_cancelled) {
+            query_mem_tracker->set_is_query_cancelled(is_cancelled);
+        }
         _start_cond.notify_all();
     }
     void set_ready_to_execute_only() {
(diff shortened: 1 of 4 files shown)


luozenglin:
run buildall

be/src/runtime/query_context.h
```diff
@@ -109,6 +109,9 @@ class QueryContext {
             _is_cancelled = is_cancelled;
             _ready_to_execute = true;
         }
+        if (query_mem_tracker && is_cancelled) {
+            query_mem_tracker->set_is_query_cancelled(is_cancelled);
```
xinyiZzz, line 113: Why here, not in `PlanFragmentExecutor::cancel` and `PipelineFragmentContext::cancel`

seems to break the meaning of `set_ready_to_execute`

luozenglin:
run buildall

yiguolei:
do not use name resource pool. resource pool is used to be resource isolation in olap system. if it is a schema cache , then just call schema_cache

luozenglin:
run buildall

xinyiZzz (approved):
LGTM

PR #18949, opened 2023-04-23
[fix](session-var) ignore exception when setting global only var in non master FE

morningman:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Introduced from #18609.

When setting global variables from Non Master FE, there will be error like:

`Variable 'password_history' is a GLOBAL variable and should be set with SET GLOBAL`

Because when setting global variables from Non Master FE, Doris will do following step:

1. forward this SetStmt to Master FE to execute.
2. Change this SetStmt to "SESSION" level, and execute it again on this Non Master FE.

But for "GLOBAL only" variable, such ash "password_history", it doesn't allow to set on SESSION level.
So when doing step 2, "set password_history=xxx" without "GLOBAL" keywords will throw exception.
So in this case, we should just ignore this exception and return.

## Checklist(Required)

* [ ] Does it affect the original behavior
* [ ] Has unit tests been added
* [ ] Has document been added or modified
* [ ] Does it need to update dependencies
* [ ] Is this PR support rollback (If NO, please explain WHY)

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #20179, opened 2023-05-29
[fix](session-variable) fix set global var on non-master FE return error

morningman:
## Proposed changes

When executing `set global password_history=0` on non-master FE, error return:

`Variable 'password_history' is a GLOBAL variable and should be set with SET GLOBAL`

Introduced from #18949.
In #18949, I caught the exception when setting session var on non-master FE, to ignore those exception.
And the exception is thrown by calling `ErrorReport.reportDdlException()`, this method will also set the error msg
in query state of connection context, which causing error result finally like.

In this PR, I changed the method to use `throw new DdlException` directly, to avoid such error.

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/VariableMgr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/VariableMgr.java
@@ -243,15 +243,18 @@ public class VariableMgr {
     }
 
     // Check if this setVar can be set correctly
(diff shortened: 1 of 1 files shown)


morningman:
run buildall

morningman:
run buildall

morningman:
run buildall

dataroaring (approved):
LGTM

PR #20228, opened 2023-05-30
[fix](regression-test) add jdbc timeout

shuke987:
## Proposed changes
In some cases ( or bugs), doris may returned query to jdbc, but jdbc can not recognized what doris sent back, so hanged. To fix this, add a timeout of 30 minutes to jdbc connection.


shuke987:
run buildall

yiguolei (approved):
lgtm

PR #20307, opened 2023-06-01
[Feature](load)RoutineLoad support multi table load

CalvinKirs:
### Backend:

Synchronize tens of thousands of meter-level data to doris through routine. Currently, routine load only supports the import of one table, and 1w import tasks need to be created for 1w tables, which will put enormous pressure on task management and final txn submission. At the same time, users also need to maintain 1w kafka topics. If the data of multiple tables can be transferred to doris on a topic and then distributed to multiple target tables, the above problems can be solved.

#### Data Driven Planning
The routineload of a single table does execution planning first, and then imports data, while multi-table information is generated dynamically. For example, we define a multi-table JOB, define a topic named order, and set `order_address`,
`order_item`, `order_bill`, `order_pay`, the data information of the three tables are sent to the `order` topic, then in the scheduling process of a single task, a consumption interval may only have part of the information, for example, in most cases there is no
The data of `order_pay`, after all, you need to think twice when you actually pay. Of course, wealthy people don’t consider this aspect. Back to the topic, when we actually consume news and import data, not every task includes all the tables.
If we use the method of planning first, then when planning, we need to know the information of all tables, and there are many meaningless plans.
Therefore, when FE submits a task, if it is multi-table, it will not generate an execution plan. After BE receives the request from Task, it starts to save a batch of data. When the amount of data reaches a certain threshold, it starts to request execution plan from FE, and then Start data import.

image

#### Grammatical semantic modification
```
CREATE ROUTINE LOAD [db.] job_name [ON tbl_name]
[merge_type]
[load_properties]
[job_properties]
```

ON tbl_name is an optional parameter. When not specified, it means multi-table import. When specified, it means single-table import.

The syntax and semantics are 100% compatible, and will not affect the existing single-table import and the upgraded single-table import.

#### Task Management
There is little change from before, only the execution planning process and the way of submitting tasks have been changed.

#### Implementation plan

New execution plan for dynamic multi-table

#### Metadata modification
There is no metadata modification, and the original data logic is used to judge whether it is a multi-table import.

#### Client modification

`SHOW ROUTINE LOAD`
- Added `IS_MULTI_TABLE` column, which is `true` for multi-table import, and `false` for single-table import.
- When importing multiple tables, since the table is a dynamic table, the table name is displayed as `multi-table`, indicating that this is a multi-table import. When importing a single table, the table name is displayed as the actual table name.

### Eg

Suppose we have a user-related library to store the basic information of the user, such as user_address, user_info and other tables,
We need to synchronize these data to Doris, and we use the message middleware Kafka to realize real-time data synchronization.

Create a topic named user, and then write the data of user_address, user_info and other tables into this topic.

We can write table information in Key or Value, and then tell Doris through configuration,

By default, we write in the value in Kafka, and use | as a separator, and the data information can be in json or csv format. Let's take json as an example.
You have a table named user_address, which has three fields, namely user_id, address, and timestamp. We can write the information of these three fields into Kafka.

eg:
user_address|{"timestamp": "2020-01-01", "type": 1, "user_id": 6395}

We create a RoutineLoad task, the label is user_info, and the target library is user library
```sql

CREATE ROUTINE LOAD user.user
COLUMNS TERMINATED BY ","
PROPERTIES
(
"max_batch_interval" = "20",
"max_batch_rows" = "300000",
"max_batch_size" = "209715200",
"format" = "json"
)
FROM KAFKA
(
"kafka_broker_list" = "localhost:9092",
"kafka_topic" = "user",
"property.kafka_default_offsets" = "OFFSET_END"
);
    
```
Suppose your table name needs to be transformed, such as user-address|busi_type=1|timestamp=2020-01-01 into user_address, you can configure mapping to achieve it (TODO new feature)
```sql
FROM KAFKA
(
"kafka_broker_list" = "localhost:9092",
"kafka_topic" = "user",
"property.kafka_default_offsets" = "OFFSET_END"
"table_name_location" = "key" // key or value
"table_name_format" = "TEXT" // TEXT or JSON
"table_name_delimiter" = "|" // delimiter
"table_name_index" = "0" // index
);
```


CalvinKirs:
todo: design and document

freemandealer:
LGTM

CalvinKirs:
run buildall

CalvinKirs:
run buildall

dataroaring:
run buildall

CalvinKirs:
run buildall

CalvinKirs:
re-run P0 (due to timeout)

CalvinKirs:
run buildall

CalvinKirs:
run buildall

CalvinKirs:
run buildall

CalvinKirs:
run buildall

CalvinKirs:
run p0

CalvinKirs:
run clickbench (benchmark)

CalvinKirs:
run buildall

CalvinKirs:
run clickbench (benchmark)

CalvinKirs:
run p0

CalvinKirs:
run clickbench (benchmark)

CalvinKirs:
run buildall

CalvinKirs:
run buildall

zy-kkk (approved):
LGTM

PR #20349, opened 2023-06-01
[fix](olap) deletion statement with space conditions did not take effect

mrhhsg:
## Proposed changes

Deletion statement like this:
```sql
delete from tb where k1 = '  ';
```

The rows whose `k1`'s value is ' ' will not be deleted.

## Further comments


mrhhsg:
run buildall

mrhhsg:
run buildall

be/src/olap/delete_handler.cpp
```diff
@@ -110,7 +110,7 @@ std::string DeleteHandler::construct_sub_predicates(const TCondition& condition)
         } else if (op == "!*=") {
             op = "!=";
         }
-        condition_str = condition.column_name + op + condition.condition_values[0];
+        condition_str = condition.column_name + op + "'" + condition.condition_values[0] + "'";
```
yiguolei, line 113: If we upgrade from doris 1.2 and there is a delete condition. It does not contains '', is there error?

yiguolei: because it already save the previous condition does not have '' in data file.

yiguolei (approved):
LGTM

PR #20361, opened 2023-06-02
[fix](match) fix match query with compound predicates return -6003

Tanya-W:
## Proposed changes

fix match query with compound predicates return -6003

## Further comments


be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -717,10 +717,15 @@ Status SegmentIterator::_apply_index_except_leafnode_of_andnode() {
             continue;
         }
```
eldenmoon, line 718: maybe we should add some test case?

Tanya-W:
run buildall

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -717,10 +717,15 @@ Status SegmentIterator::_apply_index_except_leafnode_of_andnode() {
...
+                                             PredicateTypeTraits::is_equal_or_list(pred_type);
         if (!res.ok()) {
-            if ((res.code() == ErrorCode::INVERTED_INDEX_FILE_NOT_FOUND &&
-                 pred->type() != PredicateType::MATCH) ||
-                res.code() == ErrorCode::INVERTED_INDEX_FILE_HIT_LIMIT) {
+            if (res.code() == ErrorCode::INVERTED_INDEX_FILE_NOT_FOUND ||
+                res.code() == ErrorCode::INVERTED_INDEX_FILE_HIT_LIMIT ||
+                res.code() == ErrorCode::INVERTED_INDEX_EVALUATE_SKIPPED ||
+                (res.code() == ErrorCode::INVERTED_INDEX_NO_TERMS &&
+                 need_remaining_after_evaluate)) {
```
airborne12, line 728: wrap as a function, so _apply_inverted_index_on_column_predicate can also use it, and next time you change your condition, you will not forget that

be/src/olap/rowset/segment_v2/segment_iterator.cpp
Tanya-W: yes, I will add some regression case

be/src/olap/rowset/segment_v2/segment_iterator.cpp
Tanya-W: you are right, I will update it

Tanya-W:
run buildall

eldenmoon (dismissed):
LGTM

Tanya-W:
run buildall

qidaye (approved):
LGTM

PR #20404, opened 2023-06-04
[Fix]fix lazy open commit info lose

sollhui:
## Proposed changes

**problem**
when open enable_lazy_open_partition config, there is a possibility that the open partition rpc will  perform the open operation after the txn commit, which will cause the commit to fail and the load fail.

**What is this pr task**
- Therefore, to solve this problem,When sending the last RPC to obtain the commit information, it is necessary to wait for all lazy opens to end.
- a more comprehensive case has been added for lazy open testing.

## Further comments

### DIFF
--- a/be/src/vec/sink/vtablet_sink.cpp
+++ b/be/src/vec/sink/vtablet_sink.cpp
@@ -320,6 +320,9 @@ Status VNodeChannel::init(RuntimeState* state) {
 
     _rpc_timeout_ms = state->execution_timeout() * 1000;
     _timeout_watch.start();
+    if (config::enable_lazy_open_partition) {
+        _lazy_open_timeout_watch.start();
+    }
 
     // Initialize _cur_add_block_request
     _cur_add_block_request.set_allocated_id(&_parent->_load_id);
@@ -516,7 +519,7 @@ Status VNodeChannel::open_wait() {
 }
 
 void VNodeChannel::open_partition(int64_t partition_id) {
-    _timeout_watch.reset();
+    _lazy_open_timeout_watch.reset();
     SCOPED_CONSUME_MEM_TRACKER(_node_channel_tracker.get());
     OpenPartitionRequest request;
     auto load_id = std::make_shared<PUniqueId>(_parent->_load_id);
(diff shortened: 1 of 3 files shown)


HHoflittlefish777:
run buildall

regression-test/suites/rollup_p0/test_materialized_view_lazy_open.groovy
```diff
@@ -70,6 +75,58 @@ suite("test_materialized_view_lazy_open", "rollup") {
...
+            PARTITION BY RANGE(k1)
+            (
+               PARTITION p1 VALUES LESS THAN ("2000-01-01"),
+               PARTITION p2 VALUES LESS THAN ("2010-01-01"),
+               PARTITION p3 VALUES LESS THAN ("2020-01-01")
+            )
+            DISTRIBUTED BY HASH(k1) BUCKETS 32 
+
+            properties(
+                "light_schema_change" = "true",
```
freemandealer, line 101: Even though we enable schema change here, we have not triggered schema change yet.

be/src/vec/sink/vtablet_sink.cpp
```diff
@@ -320,6 +320,9 @@ Status VNodeChannel::init(RuntimeState* state) {
 
     _rpc_timeout_ms = state->execution_timeout() * 1000;
     _timeout_watch.start();
+    if (config::enable_lazy_open_partition) {
+        _lazy_open_timeout_watch.start();
+    }
```
freemandealer, line 325: no need to start stop watch here.

be/src/vec/sink/vtablet_sink.cpp
```diff
@@ -516,7 +519,7 @@ Status VNodeChannel::open_wait() {
 }
 
 void VNodeChannel::open_partition(int64_t partition_id) {
-    _timeout_watch.reset();
+    _lazy_open_timeout_watch.reset();
```
freemandealer, line 522: use stop watch as local var and start here.

be/src/vec/sink/vtablet_sink.h
```diff
@@ -302,6 +304,7 @@ class VNodeChannel {
     int _rpc_timeout_ms = 60000;
     int64_t _next_packet_seq = 0;
     MonotonicStopWatch _timeout_watch;
+    MonotonicStopWatch _lazy_open_timeout_watch;
```
freemandealer, line 307: use local var instead of member?

HHoflittlefish777:
run buildall

freemandealer (approved):
LGTM

dataroaring (approved):
LGTM

PR #20418, opened 2023-06-05
[Chore](docs)Remove the Committer-related description in the contribution guide

CalvinKirs:
Only PMC members can subscribe to the mailing list, including this in the contributor guide would create ambiguity and unnecessary subscriptions.

Gitbox is also limited to Committer members.

We'd better create new Committer/PMC wizard docs if we can, but anyway, the description here is always inappropriate.


CalvinKirs:
run buildall

luzhijing (approved):
LGTM

PR #20434, opened 2023-06-05
[fix](planner)Fix query hang when using queue

wangbo:
## Proposed changes
Report by @liutang123 
When enable query queue, query may hang because query queue is not release correctly.
So we need to call queue.offer and release it in ```handleQueryWithRetry```


wangbo:
run buildall

liutang123 (approved):
LGTM

PR #20448, opened 2023-06-05
[fix](executor)Fix duplicate timer and add open timer

wangbo:
## Proposed changes

1 Currently, Node's total timer couter has timed twice(in ```Open``` and ```alloc_resource```), this may cause timer in profile is not correct.
2 Add more timer to find more code which may cost much time.


yiguolei (approved):
LGTM

yiguolei:
run buildall

PR #20480, opened 2023-06-06
[fix](alter) fix potential concurrent issue for alter when check olap table state normal outside write lock scope is not atomic

caiconghui:
## Proposed changes

Issue Number: close #xxx

now, we check some olap table state normal outside write lock scope, the table state may be changed to unnormal when we do alter operation

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/alter/SchemaChangeHandler.java
@@ -1173,13 +1173,7 @@ public class SchemaChangeHandler extends AlterHandler {
 
     private void createJob(long dbId, OlapTable olapTable, Map<Long, LinkedList<Column>> indexSchemaMap,
                            Map<String, String> propertyMap, List<Index> indexes) throws UserException {
-        if (olapTable.getState() == OlapTableState.ROLLUP) {
-            throw new DdlException("Table[" + olapTable.getName() + "]'s is doing ROLLUP job");
-        }
-
         checkReplicaCount(olapTable);
-        // for now table's state can only be NORMAL
-        Preconditions.checkState(olapTable.getState() == OlapTableState.NORMAL, olapTable.getState().name());
 
         // process properties first
         // for now. properties has 3 options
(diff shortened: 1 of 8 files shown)


caiconghui:
run buildall

caiconghui:
run buildall

caiconghui:
run buildall

caiconghui:
run buildall

fe/fe-core/src/main/java/org/apache/doris/alter/SchemaChangeHandler.java
```diff
@@ -2471,13 +2466,6 @@ public void modifyTableLightSchemaChange(Database db, OlapTable olapTable,
             throws DdlException {
 
         LOG.debug("indexSchemaMap:{}, indexes:{}", indexSchemaMap, indexes);
-        if (olapTable.getState() == OlapTableState.ROLLUP) {
-            throw new DdlException("Table[" + olapTable.getName() + "]'s is doing ROLLUP job");
-        }
-
-        // for now table's state can only be NORMAL
-        Preconditions.checkState(olapTable.getState() == OlapTableState.NORMAL, olapTable.getState().name());
-
```
Tanya-W, line 2480: Is ok removing this check here? I think here also need `olapTable.checkNormalStateForAlter();`

fe/fe-core/src/main/java/org/apache/doris/catalog/OlapTable.java
```diff
@@ -1489,21 +1489,9 @@ public long getReplicaCount() {
...
     }
 
-    public void checkStableAndNormal(String clusterName) throws DdlException {
+    public void checkNormalStateForAlter() throws DdlException {
         if (state != OlapTableState.NORMAL) {
-            throw new DdlException("Table[" + name + "]'s state is not NORMAL. "
-                    + "Do not allow doing materialized view");
-        }
-        // check if all tablets are healthy, and no tablet is in tablet scheduler
-        boolean isStable = isStable(Env.getCurrentSystemInfo(),
```
Tanya-W, line 1498: why remove this logic which check all tablets healthy?

fe/fe-core/src/main/java/org/apache/doris/alter/SchemaChangeHandler.java
caiconghui: we already check table state inside the table write lock no need double check here

fe/fe-core/src/main/java/org/apache/doris/catalog/OlapTable.java
caiconghui: actually checkTableStable function will check table stable, no need to double check here

weizuo93 (approved):
LGTM

PR #20521, opened 2023-06-06
[improvement](column reader) improve column reader options and delete not necessary options

jacktengg:
## Proposed changes

Issue Number: close #xxx

Minor improvement of ColumnReader options;

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_v2/column_reader.h
+++ b/be/src/olap/rowset/segment_v2/column_reader.h
@@ -74,10 +74,10 @@ class RowRanges;
 class ZoneMapIndexReader;
 
 struct ColumnReaderOptions {
-    // whether verify checksum when read page
-    bool verify_checksum = true;
     // for in memory olap table, use DURABLE CachePriority in page cache
     bool kept_in_memory = false;
+
+    bool use_page_cache = false;
 };
 
 struct ColumnIteratorOptions {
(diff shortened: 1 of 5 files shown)


jacktengg:
run buildall

be/src/olap/rowset/segment_v2/page_io.cpp
```diff
@@ -154,13 +154,11 @@ Status PageIO::read_and_decompress_page(const PageReadOptions& opts, PageHandle*
         opts.stats->compressed_bytes_read += page_size;
     }
 
-    if (opts.verify_checksum) {
```
yiguolei, line 157: Why remove this if? Because it is always true?

be/src/olap/rowset/segment_v2/column_reader.h
```diff
@@ -74,10 +74,10 @@ class RowRanges;
 class ZoneMapIndexReader;
 
 struct ColumnReaderOptions {
-    // whether verify checksum when read page
-    bool verify_checksum = true;
     // for in memory olap table, use DURABLE CachePriority in page cache
     bool kept_in_memory = false;
```
yiguolei, line 78: Is kept_in_memory useful?

be/src/olap/rowset/segment_v2/page_io.cpp
jacktengg: yes

PR #20525, opened 2023-06-06
[fix](executor)Fix emptyoperator may cause node not close

wangbo:
## Proposed changes
When PipelineTask generate an EmptyOperator, it doesn't transfer ExecNode to EmptyOperator, this may cause ExecNode's resource may not released correctly.
So we need to close the node which is replaced of EmptyOperator.


HappenLee (approved):
LGTM

PR #20621, opened 2023-06-09
[Fix](profile)remove DCHECK for rpc time

wangbo:
## Proposed changes

Remove DCHECK for rpc Timer, because rpc client and rpc server may not has the synchronized time


wangbo:
run buildall

yiguolei (approved):
lgtm

PR #20704, opened 2023-06-12
[fix](routine-load) fix stackoverflow bug in routine load

morningman:
## Proposed changes

When executing routine load job, there may encounter StackOverflowException.
This is because the expr in column setting list will be analyze for each routine load sub task,
and there is a self-reference bug that may cause endless loop when analyzing expr.

The following columns expr list may trigger this bug:

```
columns(col1, col2,
col2=null_or_empty(col2),
col1=null_or_empty(col2))
```

This fix is verified by user, but I can't add regression test for this case, because I can't submit a routine load job
in our regression test, and this bug can only be triggered in routine load.

## Further comments


morningman:
run buildall

caiconghui (approved):
LGTM

PR #20732, opened 2023-06-13
[Bug](topn opt) Fix Two-Phase read when some rowset swept

eldenmoon:
If this is a Two-Phase read query, and we need to delay the release of Rowset by row->update_delayed_expired_timestamp() to expand the lifespan of rowsets. This is necessary to avoid data loss during the second phase reading, where some stale rowsets may be swept and result in missing data.
    
## Proposed changes

Issue Number: close #xxx

## Further comments

### DIFF
--- a/be/src/olap/storage_engine.cpp
+++ b/be/src/olap/storage_engine.cpp
@@ -1004,9 +1004,7 @@ void StorageEngine::start_delete_unused_rowset() {
     {
         std::lock_guard<std::mutex> lock(_gc_mutex);
         for (auto it = _unused_rowsets.begin(); it != _unused_rowsets.end();) {
-            uint64_t now = UnixSeconds();
-            if (it->second.use_count() == 1 && it->second->need_delete_file() &&
-                now > it->second->delayed_expired_timestamp()) {
+            if (it->second.use_count() == 1 && it->second->need_delete_file()) {
                 if (it->second->is_local()) {
                     unused_rowsets_copy[it->first] = it->second;
                 }
@@ -1157,15 +1155,10 @@ Status StorageEngine::execute_task(EngineTask* task) {
 }
 
 // check whether any unused rowsets's id equal to rowset_id
-std::optional<RowsetSharedPtr> StorageEngine::get_rowset_in_unused_rowsets(
-        const RowsetId& rowset_id) {
+bool StorageEngine::check_rowset_id_in_unused_rowsets(const RowsetId& rowset_id) {
     std::lock_guard<std::mutex> lock(_gc_mutex);
     auto search = _unused_rowsets.find(rowset_id.to_string());
-    if (search != _unused_rowsets.end()) {
-        return search->second;
-    }
-    // Not found
-    return {};
+    return search != _unused_rowsets.end();
 }
 
 void StorageEngine::create_cumulative_compaction(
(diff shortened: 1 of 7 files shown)


eldenmoon:
run buildall

be/src/runtime/query_context.h
```diff
@@ -222,6 +228,14 @@ class QueryContext {
     taskgroup::TaskGroupPtr _task_group;
     std::unique_ptr<RuntimeFilterMgr> _runtime_filter_mgr;
     const TQueryOptions _query_options;
+
+    // When querying, certain structures need to be delayed in their release.
+    // Placing these structures in _reference_holder can extend their lifespan until the query is completed,
+    // and they are automatically released. For example, in two-phase delayed materialization queries,
+    // Rowsets need to be released after the query is finished to avoid premature release during the query process,
+    // which could lead to query exceptions.
+    std::vector<std::shared_ptr<void>> _reference_holder;
```
yiguolei, line 237: Do not use a very abstract structure. I think you could use std::shared_ptr<Rowset> and then the maintainer could understand very clear that the query context hold some rowset. Or there maybe some memory leak for example, if some other people add olap scan node to query context. The olap scan node maybe a shared ptr, it owns query context and query context owns olap scannode .

be/src/service/internal_service.cpp
```diff
@@ -1512,11 +1512,18 @@ Status PInternalServiceImpl::_multi_get(const PMultiGetRequest& request,
...
             continue;
         }
+        // Get Rowset from either tablet or unused rowsets, since this rowset maybe expired and swept.
+        // But we ensured it's rowset is not released when init Tablet reader param, by QueryContext::hold_reference
         BetaRowsetSharedPtr rowset =
                 std::static_pointer_cast<BetaRowset>(tablet->get_rowset(rowset_id));
         if (!rowset) {
-            LOG(INFO) << "no such rowset " << rowset_id;
-            continue;
+            std::optional<RowsetSharedPtr> from_unused_rowset =
```
yiguolei, line 1520: If it is hold in query context, why not get it from query context?

be/src/olap/storage_engine.cpp
```diff
@@ -1155,10 +1155,15 @@ Status StorageEngine::execute_task(EngineTask* task) {
 }
 
 // check whether any unused rowsets's id equal to rowset_id
-bool StorageEngine::check_rowset_id_in_unused_rowsets(const RowsetId& rowset_id) {
+std::optional<RowsetSharedPtr> StorageEngine::get_rowset_in_unused_rowsets(
```
yiguolei, line 1158: I think its very bad to get a rowset from gc space. GC is a standalone service. Other logic should not depend on it. The rowset is already hold in query context, I think you should get it from query context.

be/src/runtime/query_context.h
eldenmoon: done

be/src/service/internal_service.cpp
eldenmoon: done

be/src/olap/storage_engine.cpp
eldenmoon: done

eldenmoon:
run buildall

eldenmoon:
run buildall

eldenmoon:
run buildall

yiguolei (approved):
LGTM

qidaye (approved):
LGTM

PR #20947, opened 2023-06-18
[fix](leftjoin) fix bug of left and full join with other conjuncts

jacktengg:
## Proposed changes

Issue Number: close #xxx

Fix bug of left and full outer join with other conjuncts. When equal matched row count of a probe row exceed batch_size, some times the _join_node->_is_any_probe_match_row_output flag is not set correcty, which result in outputing extra rows for the probe row.

## Further comments


regression-test/data/query_p0/join/test_full_join_batch_size.out
```diff
@@ -0,0 +1,9 @@
+-- This file is automatically generated. You should know what you did if you want to edit this
+-- !sql1 --
```
yiguolei, line 2: Is this problem exist on master？ and could we add these test cases to master?

PR #21032, opened 2023-06-20
[Fix](mysql proto) avoid send duplicated `OK` packet

eldenmoon:
## Proposed changes

1. The Mysql Go driver has a logic that terminates when it reads an EOF (end-of-file) and expects no data in the buffer. However, the front-end (FE) mistakenly returns an additional OK packet, which causes an exception to be thrown when reading the buffer.

2. Refactor some logic to support full prepared not just in where clause, like 
```
select ?, ? from tbl
```

Issue Number: close #xxx

## Further comments

### DIFF
--- a/fe/fe-core/src/main/cup/sql_parser.cup
+++ b/fe/fe-core/src/main/cup/sql_parser.cup
@@ -67,6 +67,7 @@ parser code {:
     public boolean isVerbose = false;
     public String wild;
     public Expr where;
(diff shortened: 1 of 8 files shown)


xiaokang (approved):
LGTM

SaintBacchus:
LGTM

SaintBacchus (approved):

PR #21042, opened 2023-06-20
[FIX](array-type) fix array string output with fe const expr

amorynan:
## Proposed changes
fe foldconstRule make array() function expr with const literal , and would not pass this array literal to be . but we should make fe array string output format is same with be array string output
Issue Number: close #xxx

## Further comments


amorynan:
run buildall

xiaokang (approved):
LGTM

eldenmoon (approved):
LGTM

PR #21077, opened 2023-06-21
[Fix](Nereids) fix cast compare of varchar with double

LiBinfeng-01:
## Proposed changes

Problem: when using nereids planner, some constant folding does not work. like: 
explain select case 'a' when 1 then 'a' when 'a' then 'b' else 'other' end as col212;
which will add cast to 'a' to double, but Cast is not treated as constant, so this folding can not be done

Solve: Treat Cast(varchar as double) as NullLiteral if can not cast, which cast condition will return false

## Further comments


LiBinfeng-01:
run buildall

PR #21102, opened 2023-06-23
[fix](pipeline) fix MultiCoreTaskQueue steal take core_id assgin error

ZhangYu0123:
## Proposed changes
1.  fix MultiCoreTaskQueue steal take core_id assgin error
stealed pipeline_task will run on core_id thread not previous next_id thread, so the pipeline_task will use core_id queue runtime  and should be assginned core_id. 

Issue Number: close #xxx

## Further comments

### DIFF
--- a/be/src/pipeline/task_queue.cpp
+++ b/be/src/pipeline/task_queue.cpp
@@ -182,7 +182,7 @@ PipelineTask* MultiCoreTaskQueue::_steal_take(size_t core_id) {
         DCHECK(next_id < _core_size);
         auto task = _prio_task_queue_list[next_id].try_take(true);
         if (task) {
-            task->set_core_id(next_id);
+            task->set_core_id(core_id);
             return task;
         }
     }


ZhangYu0123:
run buildall

ZhangYu0123:
run p1

be/src/pipeline/task_queue.cpp
```diff
@@ -182,7 +182,7 @@ PipelineTask* MultiCoreTaskQueue::_steal_take(size_t core_id) {
         DCHECK(next_id < _core_size);
         auto task = _prio_task_queue_list[next_id].try_take(true);
         if (task) {
-            task->set_core_id(next_id);
```
HappenLee, line 185: need change the origin core id seems right？

ZhangYu0123: > need change the origin core id seems right？

I think core_id of pipeline task means which core thread the task belong to and which multi core task queue the task belong to.  On the other hand,  previous_core_id of pipeline task means the origin core id.
So the task is stealed from next_id to the core_id, it should be the core_id.

PR #21110, opened 2023-06-24
[improvement] Add data type validation when create table stmt

ZashJie:
## Proposed changes

Issue Number: close #xxx

Added corresponding type validation in the create table statement to ensure that the data types used during table creation are available.

## Further comments


conf/be.conf
```diff
@@ -52,7 +52,7 @@ enable_auth = false
 # use CIDR format, e.g. 10.10.10.0/24 or IP format, e.g. 10.10.10.1
 # Default value is empty.
 # priority_networks = 10.10.10.0/24;192.168.0.0/16
-
+priority_networks = 127.0.0.1/16
```
yiguolei, line 55: do not modify this config. If the user not modify conf and alreay running and copy this conf to its env, then it will not work correctly.

PR #21186, opened 2023-06-26
[Fix](Planner) Set inline view output as non constant after analyze 

LiBinfeng-01:
## Proposed changes

Problem:
Select list should be non const when from list have tables or multiple tuples. Or upper query will regard wrong of isConstant 
And make wrong constant folding
For example： when using nullif funtion with subquery which result in two alternative constant, planner would treat it as constant expr. So analyzer would report an error of order by clause can not be constant

Solusion:
Change inline view output to non constant, because (select 1 a from table) as view , a in output is no constant when we see 
view.a outside

## Further comments


LiBinfeng-01:
#21212

morrySnow (approved):

starocean999 (approved):

PR #21198, opened 2023-06-26
[fix](regression) Add get master token into regression framework

deadlinefen:
Bug: 
Because the old version does not support getMasterToken, the token is fixed in the old version of the test, which causes the P1 regression test_backup_restore case to fail.

Bug fix:
Add get master token into regression framework.


deadlinefen:
run buildall

deadlinefen:
run P1

dataroaring (approved):
LGTM

JackDrogon (approved):
LGTM

PR #21272, opened 2023-06-28
[Fix](storage)read page cache when seek

wangbo:
## Proposed changes

Currently, when a columnIter is used for seek, then page cache is not set;
When this colunIter is used for later read data, then page cache could not be used.


yiguolei:
run buildall

yiguolei (approved):
LGTM

xinyiZzz (approved):
LGTM

PR #21390, opened 2023-06-30
[Fix](JSON LOAD)fix json load issue when string conform with RFC 4627

### LINKED ISSUE #21389: [Bug] (JSON load) fix json load get string value wrong when the string contain "\u0000"
### Search before asking

### Version

1.2 and master

### What's Wrong?

1,  create table like:
CREATE TABLE `test` (
  `pin_id` bigint(20) NOT NULL,
  `pin` varchar(10) NOT NULL,
  `date` date NOT NULL COMMENT '天级别',
  `ecpc_exp_2_cost` bigint(20) SUM NULL DEFAULT "0" COMMENT 'ecpc'
) ENGINE=OLAP
AGGREGATE KEY(`pin_id`, `pin`, `date`)

2, load below json with stream load
{"pin_id":1233,"pin":"aaaaaaaa\u0000aaaa","date":"2023-05-30","ecpc_exp_2_cost":10}

3, this load will succeed with pin:"aaaaaaaa",  this should be wrong as "aaaaaaaa\u0000aaaa" is a valid string and length exceed 10. the load should failed as "the length of data exceed that defined in schema". 

### What You Expected?

load failed
as the data convert from parquet format, and loading with parquet format will fail. 

### How to Reproduce?

1,  create table like:
CREATE TABLE `test` (
  `pin_id` bigint(20) NOT NULL,
  `pin` varchar(10) NOT NULL,
  `date` date NOT NULL COMMENT '天级别',
  `ecpc_exp_2_cost` bigint(20) SUM NULL DEFAULT "0" COMMENT 'ecpc'
) ENGINE=OLAP
AGGREGATE KEY(`pin_id`, `pin`, `date`)

2, load below json with stream load
{"pin_id":1233,"pin":"aaaaaaaa\u0000aaaa","date":"2023-05-30","ecpc_exp_2_cost":10}

3, this load will succeed with pin:"aaaaaaaa",  this should be wrong as "aaaaaaaa\u0000aaaa" is a valid string and length exceed 10. the load should failed as "the length of data exceed that defined in schema". 

### Anything Else?

None

### Code of Conduct

GoGoWen:
## Proposed changes

should set: enable_simdjson_reader=false in master as master enable_simdjson_reader=true by default.

Issue Number: close #21389

from rapidjson:

Query String
In addition to GetString(), the Value class also contains GetStringLength(). Here explains why:

According to RFC 4627, JSON strings can contain Unicode character U+0000, which must be escaped as "\u0000". The problem is that, C/C++ often uses null-terminated string, which treats \0 as the terminator symbol.

To conform with RFC 4627, RapidJSON supports string containing U+0000 character. If you need to handle this, you can use GetStringLength() to obtain the correct string length.

For example, after parsing the following JSON to Document d:

{ "s" :  "a\u0000b" }
The correct length of the string "a\u0000b" is 3, as returned by GetStringLength(). But strlen() returns 1.

GetStringLength() can also improve performance, as user may often need to call strlen() for allocating buffer.

Besides, std::string also support a constructor:

string(const char* s, size_t count);
which accepts the length of string as parameter. This constructor supports storing null character within the string, and should also provide better performance.

## Further comments

### DIFF
--- a/be/src/vec/exec/format/json/new_json_reader.cpp
+++ b/be/src/vec/exec/format/json/new_json_reader.cpp
@@ -890,7 +890,7 @@ Status NewJsonReader::_write_data_to_column(rapidjson::Value::ConstValueIterator
     switch (value->GetType()) {
     case rapidjson::Type::kStringType:
         str_value = value->GetString();
-        wbytes = strlen(str_value);
+        wbytes = value->GetStringLength();
         break;
     case rapidjson::Type::kNumberType:
         if (value->IsUint()) {


GoGoWen:
run buildall

eldenmoon:
could you add regreesion test?

GoGoWen:
> could you add regreesion test?

the regression is hard to add as need to set be config enable_simdjson_reader=false,  seems master code is use simdjson by default.

eldenmoon (approved):
LGTM

yiguolei (approved):
LGTM

PR #21466, opened 2023-07-04
[Fix](CCR) Use tableId as the credential for CCR syncer instead of tableName

deadlinefen:
Bug:
In the current version, CCR’s getBinlog and BeginTxn operations both depend on tableName, which will cause syncer to fail when the source cluster changes the table name.

Bugfix:
- Add tableId in TGetBinlogRequest.
Add a tableId field in TGetBinlogRequest, and when the tableId field is set, the table field will be invalid. If the tableId field is unset, the source cluster will find the corresponding tableId based on this tableName, and this operation should only be used when initializing syncer.
- change tableName to tableId in TBeginTxnRequest.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/service/FrontendServiceImpl.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/service/FrontendServiceImpl.java
@@ -914,6 +914,24 @@ public class FrontendServiceImpl implements FrontendService.Iface {
         return result;
     }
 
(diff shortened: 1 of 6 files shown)


deadlinefen:
run buildall

JackDrogon (approved):
LGTM

dataroaring (approved):
LGTM

JackDrogon (approved):
LGTM

PR #21508, opened 2023-07-05
[fix](rpc) shutdown removed backend client without wait

w41ter:
1. `ManagedChannel` needs to shutdown manually, otherwise, the channel will leak.
2. If a connection reset occurs, then `removeProxy` will be called, and shutdown will be called further. But the original implementation of `shutdown` will wait until the channel becomes terminated, this process may consume some time, but it is not necessary.


xiaokang:
add description pls

dataroaring:
run buildall

PR #21556, opened 2023-07-06
[fix](gc) infinit loop when handle exceed limit memory

shuke987:
## Proposed changes

Issue Number: close #xxx

In some situation, _handle_mem_exceed_limit will alloc a large memory block, more than 5G. After add some log, we found that:
1. alloc memory was made in vector::insert_realloc
2. writers_to_reduce_mem's size is more than 8 million.
which indicated that an infinite loop was met in while (!tablets_mem_heap.empty()).
By reviewing codes, """ if (std::get<0>(tablet_mem_item)++ != std::get<1>(tablet_mem_item))  """ is wrong, 
which must be  """ if (++std::get<0>(tablet_mem_item) != std::get<1>(tablet_mem_item)) """.
In the original code, we will made ++ on end iterator, and then compare to end iterator, the behavior is undefined.

## Further comments

### DIFF
--- a/be/src/runtime/load_channel_mgr.cpp
+++ b/be/src/runtime/load_channel_mgr.cpp
@@ -394,7 +394,7 @@ void LoadChannelMgr::_handle_mem_exceed_limit() {
                 break;
             }
             tablets_mem_heap.pop();
-            if (std::get<0>(tablet_mem_item)++ != std::get<1>(tablet_mem_item)) {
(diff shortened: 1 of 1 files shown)


liaoxin01 (approved):
LGTM

shuke987:
run buildall

xinyiZzz (approved):
LGTM

PR #21574, opened 2023-07-06
[fix](inverted index) use index id instead of column uid to determine whether a hard link is required when build index

Tanya-W:
## Proposed changes

Fix problem:
For the same column, there are concurrent drop index request and build index request, if build index obtain lock before drop index, build a new index file, but when drop index request execute, link file not contains all index files for the column, that lead to new index file is missed.

Based on the above questions, use index id instead of column unique id to determine whether a hard link is required when do build index

## Further comments

### DIFF
--- a/be/src/olap/task/index_builder.cpp
+++ b/be/src/olap/task/index_builder.cpp
@@ -74,6 +74,18 @@ Status IndexBuilder::update_inverted_index_info() {
             for (auto t_inverted_index : _alter_inverted_indexes) {
                 TabletIndex index;
                 index.init_from_thrift(t_inverted_index, *input_rs_tablet_schema);
+                auto column_uid = index.col_unique_ids()[0];
+                const TabletIndex* exist_index =
+                        output_rs_tablet_schema->get_inverted_index(column_uid);
+                if (exist_index && exist_index->index_id() != index.index_id()) {
+                    // maybe there are concurrent drop index request did not obtain the lock,
+                    // so return error, to wait the drop index request finished.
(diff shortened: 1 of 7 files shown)


Tanya-W:
run buildall

be/src/olap/task/index_builder.cpp
```diff
@@ -74,6 +74,18 @@ Status IndexBuilder::update_inverted_index_info() {
...
+                const TabletIndex* exist_index =
+                        output_rs_tablet_schema->get_inverted_index(column_uid);
+                if (exist_index && exist_index->index_id() != index.index_id()) {
+                    // maybe there are concurrent drop index request did not obtain the lock,
+                    // so return error, to wait the drop index request finished.
+                    LOG(WARNING) << "column: " << column_uid << " has a exist inverted index"
+                                 << ", but the index id not equal request's index id, "
+                                 << ", exist index id: " << exist_index->index_id()
+                                 << ", request's index id: " << index.index_id();
+                    return Status::Error<ErrorCode::INVERTED_INDEX_BUILD_WAITTING>();
```
xiaokang, line 87: What's the effect of return INVERTED_INDEX_BUILD_WAITTING?

xiaokang (approved):
LGTM

xiaokang:
add some testcase

be/src/olap/task/index_builder.cpp
Tanya-W: the task will retry by fe

yiguolei (approved):
LGTM

PR #21614, opened 2023-07-07
[fix](Materialized) fix return error when create mv as select * from

nextdreamblue:
## Proposed changes

Issue Number: close #xxx

when create mv as select * from tbl, will return error like : `ERROR 1105 (HY000): errCode = 2, detailMessage = Unexpected exception: null`, and it is not good for user.

## Further comments


nextdreamblue:
run buildall

nextdreamblue:
run COMPILE

PR #21666, opened 2023-07-10
[type](doc) Add a document of ninedata sql window

SeranaCai:
## Proposed changes

ADD a document of ninedata sql window.  This document describes the feature and usage of ninedata sql window. The developer can connect and query doris with sql window.


SeranaCai:
The sidebar transformation has been completed, and a new SQL window has been added.

PR #21676, opened 2023-07-10
[improvement](planner) Enable preagg when count distinct key

liutang123:
## Proposed changes
A table like this:
```
create table xxx (key1, key2, key3)  AGGREGATE KEY(key, key1, key2, key3)
with rollup r_test (key1, key2,key3)

```
Now, ```select key1, count(distinct key2), count(disinct key3) from table xxx group by key1``` can not use rollup `r_test`.

Issue Number: close #xxx


liutang123:
run buildall

PR #21768, opened 2023-07-12
[Fix](Export) check properties for export

BePPPower:
## Proposed changes

Issue Number: close #xxx

When the user uses the export statement, check if the properties meets the requirements

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/load/ExportJob.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/ExportJob.java
@@ -300,7 +300,7 @@ public class ExportJob implements Writable {
                     header += genTypes();
                     break;
                 default:
-                    throw new DdlException("Unknown format for export.");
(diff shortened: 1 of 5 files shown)


fe/fe-core/src/main/java/org/apache/doris/load/ExportJob.java
```diff
@@ -287,8 +287,8 @@ private String genTypes() {
 
     private String genHeader(Map<String, String> properties) {
         String header = "";
-        if (properties.containsKey("format")) {
-            String headerType = properties.get("format");
+        if (properties.containsKey("csv_format")) {
```
yiguolei, line 290: do not change this key, should compatible with 2.0. 
We should check its value, if it is not valid then throw exception.

BePPPower: done

yiguolei (approved):
LGTM

PR #21811, opened 2023-07-14
[typo](docs) Update dev-prepare.md

wyx123654:
I edited the dev-prepar.md file, removed the Spaces at both ends of the equal sign when setting the PATH variable, and added the missing double quotation marks.


luzhijing (approved):
LGTM

PR #21895, opened 2023-07-17
[Fix](multi catalog)Fix hive partition value contains special character bug.

Jibing-Li:
Hive partition value may contain special characters, such as '/'. This may cause the partition value splitter to get wrong partition values because it use '/' to split partition value. So we need to split the encoded partition name and then decode it to the original value.


morningman (approved):
LGTM

PR #21953, opened 2023-07-18
[Fix](multi catalog)Fix nereids context table always use internal catalog bug.

Jibing-Li:
The getTable function in CascadesContext only handles the internal catalog case (try to find table only in internal catalog and dbs). However, it should take all the external catalogs into consideration, otherwise, it will failed to find a table or get the wrong table while querying external table. This pr is to fix this bug.


Jibing-Li:
run buildall

fe/fe-core/src/main/java/org/apache/doris/nereids/CascadesContext.java
```diff
@@ -388,19 +389,24 @@ private Set<UnboundRelation> extractUnboundRelationFromCTE(LogicalCTE<?> cte) {
         return relations;
     }
 
-    private Table getTable(UnboundRelation unboundRelation) {
+    private TableIf getTable(UnboundRelation unboundRelation) {
         List<String> nameParts = unboundRelation.getNameParts();
         switch (nameParts.size()) {
             case 1: { // table
+                String ctlName = getConnectContext().getEnv().getCurrentCatalog().getName();
```
morningman, line 396: Do we need to check if `current catalog` is set in `connection context`? Otherwise we may got NPE here

morningman: Same issue for dbName

hello-stephen:
(From new machine)TeamCity pipeline, clickbench performance test result:
 the sum of best hot time: 51.16 seconds
 stream load tsv:          506 seconds loaded 74807831229 Bytes, about 140 MB/s
 stream load json:         20 seconds loaded 2358488459 Bytes, about 112 MB/s
 stream load orc:          66 seconds loaded 1101869774 Bytes, about 15 MB/s
 insert into select:          29.3 seconds inserted 10000000 Rows, about 341K ops/s
 storage size: 17168222671 Bytes

morningman (approved):
LGTM

fe/fe-core/src/main/java/org/apache/doris/nereids/CascadesContext.java
LiBinfeng-01, line 396: This statement can get out out this switch case

LiBinfeng-01: ctlName should be changed to catalogName to avoid misunderstanding of "controlName" or some thing else

LiBinfeng-01 (approved):

PR #15113, opened 2022-12-15
[Bug](case function) do not crash if prepare failed

Gabriel39:
# Proposed changes

Issue Number: close #15112

## Problem summary

Describe your changes.

## Checklist(Required)

1. Does it affect the original behavior: 
    - [ ] Yes
    - [ ] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
3. Has document been added or modified:
    - [ ] Yes
    - [ ] No
    - [ ] No Need
4. Does it need to update dependencies:
    - [ ] Yes
    - [ ] No
5. Are there any changes that cannot be rolled back:
    - [ ] Yes (If Yes, please explain WHY)
    - [ ] No

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #21977, opened 2023-07-19
[Bug](case function) do not crash if prepare failed (#15113)

htyoung:
(cherry picked from commit b2438b076dd9e3705e446c9839c67a735134b909)

reproduce case:
- DDL:
create table if not exists test
(
    stat_date varchar(4096) null comment ''
    ,all_charge_num int(11) null comment ''
    ,station_id varchar(255) NULL COMMENT ''
    ,station_type_desc varchar(4096) NULL COMMENT ''
)ENGINE = OLAP
DUPLICATE KEY(`stat_date`)
COMMENT ""
DISTRIBUTED BY HASH(`stat_date`) BUCKETS 1
PROPERTIES (
	"in_memory" = "false",
	"storage_format" = "V2"
);

- SQL:
select
	count(DISTINCT CASE WHEN max(all_charge_num) = 0 THEN station_id END)
from
	test
where
	CASE
		WHEN station_type_desc LIKE '%集1%' THEN '小区集站'
		WHEN station_type_desc LIKE '%集2%' THEN '超充集站'
		ELSE station_type_desc
	END = '小区集享站'
GROUP BY
	stat_date
;

The be node will crash when execute this sql
image

### DIFF
--- a/be/src/vec/exprs/vcase_expr.cpp
+++ b/be/src/vec/exprs/vcase_expr.cpp
@@ -69,23 +69,11 @@ Status VCaseExpr::open(RuntimeState* state, VExprContext* context,
                        FunctionContext::FunctionStateScope scope) {
     RETURN_IF_ERROR(VExpr::open(state, context, scope));
     RETURN_IF_ERROR(VExpr::init_function_context(context, scope, _function));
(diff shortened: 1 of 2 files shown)


Gabriel39 (approved):
LGTM

PR #21996, opened 2023-07-19
[Enhance](Publish) improve publish version performance

qzsee:
## Proposed changes

Issue Number: close #xxx

The current publish daemon executes serial, if a transaction publishes for a long time, it will cause other things to be blocked, resulting in slow transaction release, and some load task will fail due to txn num overload.

We use multi threads to publish and speed up the finish of the transaction

The publish time for each percentile is significantly reduced

image

image

image

The number of load txn also decreased significantly

image## Further comments

### DIFF
--- a/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
@@ -2050,12 +2050,12 @@ public class Config extends ConfigBase {
     public static long statistics_sql_mem_limit_in_bytes = 2L * 1024 * 1024 * 1024;
 
     @ConfField(mutable = true, masterOnly = true)
-    public static boolean mt_enable_parallel_publish = false;
+    public static boolean enable_parallel_publish = false;
 
     @ConfField(mutable = true, masterOnly = true)
-    public static int mt_parallel_publish_thread_num = 3;
+    public static int parallel_publish_thread_num = 3;
 
     @ConfField(mutable = true, masterOnly = true)
-    public static int mt_publish_table_idle_time_secs = 172800;
+    public static int publish_table_idle_time_secs = 172800;
 
 }
--- a/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
@@ -2049,4 +2049,13 @@ public class Config extends ConfigBase {
     @ConfField
     public static long statistics_sql_mem_limit_in_bytes = 2L * 1024 * 1024 * 1024;
 
+    @ConfField(mutable = true, masterOnly = true)
+    public static boolean mt_enable_parallel_publish = false;
+
+    @ConfField(mutable = true, masterOnly = true)
+    public static int mt_parallel_publish_thread_num = 3;
+
+    @ConfField(mutable = true, masterOnly = true)
+    public static int mt_publish_table_idle_time_secs = 172800;
+
 }
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/Table.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/Table.java
@@ -162,11 +162,15 @@ public abstract class Table extends MetaObject implements Writable, TableIf {
     }
 
     public void readUnlock() {
+        long nanoTime = System.nanoTime();
         this.rwLock.readLock().unlock();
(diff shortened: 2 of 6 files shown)


qzsee:
run buildall

fe/fe-common/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -2049,4 +2049,13 @@ public class Config extends ConfigBase {
     @ConfField
     public static long statistics_sql_mem_limit_in_bytes = 2L * 1024 * 1024 * 1024;
 
+    @ConfField(mutable = true, masterOnly = true)
+    public static boolean mt_enable_parallel_publish = false;
```
yiguolei, line 2053: what does mt mean？

fe/fe-core/src/main/java/org/apache/doris/catalog/Table.java
```diff
@@ -162,11 +162,15 @@ public boolean tryReadLock(long timeout, TimeUnit unit) {
     }
 
     public void readUnlock() {
+        long nanoTime = System.nanoTime();
         this.rwLock.readLock().unlock();
+        LOG.debug("table {} get read lock spent {} ms", getId(), (nanoTime - System.nanoTime()) / 1000000L);
```
yiguolei, line 167: add  if(LOG.isDebugEnabled)  because this line will print many many times. It may have performance problem

caiconghui: should not  add time cost caculate here? especial for doris release version?

fe/fe-core/src/main/java/org/apache/doris/common/util/Daemon.java
```diff
@@ -98,10 +98,18 @@ public void setInterval(long intervalMs) {
...
+
+    public boolean isStart() {
+        return isStart.get();
+    }
+
     /**
      * implement in child
      */
-    protected void runOneCycle() {
+    protected void runOneCycle() throws Exception {
```
caiconghui, line 112: why add thorw exception declaration here

fe/fe-core/src/main/java/org/apache/doris/common/util/Daemon.java
```diff
@@ -98,10 +98,18 @@ public void setInterval(long intervalMs) {
         this.intervalMs = intervalMs;
     }
 
+    public boolean isStop() {
+        return isStop.get();
+    }
+
```
caiconghui, line 104: where to use it?

fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
```diff
@@ -1088,7 +1099,12 @@ protected void unprotectedCommitTransaction2PC(TransactionState transactionState
             return;
         }
         // update transaction state version
-        transactionState.setCommitTime(System.currentTimeMillis());
+        long commitTime = System.currentTimeMillis();
+        if (preCommitTime == commitTime || preCommitTime == 0) {
+            commitTime += 1;
+        }
+        preCommitTime = commitTime;
+        transactionState.setCommitTime(commitTime);
```
caiconghui, line 1107: why add preCommitTime logic?

qzsee: Multiple transactions commit time may be consistent, this logic is to ensure that the transaction commit time is strictly incremental, to prevent handler thread stuck

fe/fe-core/src/main/java/org/apache/doris/common/util/Daemon.java
qzsee: This is used when the handler thread is closed

fe/fe-core/src/main/java/org/apache/doris/catalog/Table.java
qzsee: statistics table lock time, we have a lot of table lock time is very serious

fe/fe-core/src/main/java/org/apache/doris/common/util/Daemon.java
qzsee: Unified handling exception

yujun777:
Give a explanation about the pictures.  Is blue curve is  old code,  yellow curve is the PR ?  Why yellow curve of the latency after "07:18 12:00:00" drop sharply ?

PR #22063, opened 2023-07-21
[Fix](multi catalog)Fix nereids context table always use internal catalog bug (#21953)

Jibing-Li:
The getTable function in CascadesContext only handles the internal catalog case (try to find table only in internal catalog and dbs). However, it should take all the external catalogs into consideration, otherwise, it will failed to find a table or get the wrong table while querying external table. This pr is to fix this bug.

Master pr link:
https://github.com/apache/doris/pull/21953

## Proposed changes

Issue Number: close #xxx

## Further comments


morningman:
Please link the master PR

PR #22301, opened 2023-07-27
[fix](scanner) Use batch size when predicate exists, otherwise, use limit

yulongfufu:
## Proposed changes

Issue Number: close #xxx

## Further comments

### DIFF
--- a/be/src/vec/exec/scan/new_olap_scanner.cpp
+++ b/be/src/vec/exec/scan/new_olap_scanner.cpp
@@ -130,8 +130,11 @@ Status NewOlapScanner::init() {
     // batch size is passed down to segment iterator, use _state->batch_size()
     // instead of _parent->limit(), because if _parent->limit() is a very small
     // value (e.g. select a from t where a .. and b ... limit 1),
-    // it will be very slow when reading data in segment iterator
-    _tablet_reader->set_batch_size(_state->batch_size());
+    // it will be very slow when reading data with predicate in segment iterator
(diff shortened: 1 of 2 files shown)


be/src/vec/exec/scan/new_olap_scanner.cpp
```diff
@@ -130,8 +130,11 @@ Status NewOlapScanner::init() {
     // batch size is passed down to segment iterator, use _state->batch_size()
     // instead of _parent->limit(), because if _parent->limit() is a very small
     // value (e.g. select a from t where a .. and b ... limit 1),
-    // it will be very slow when reading data in segment iterator
-    _tablet_reader->set_batch_size(_state->batch_size());
+    // it will be very slow when reading data with predicate in segment iterator
+    _tablet_reader->set_batch_size(
```
yiguolei, line 134: Maybe not.
conjunts maybe empty when the expr is convert to predicate.
If you do not meet an online problem or any benchmark, I think we do not need to add this check. Because users may refactor code and we may forget this condition. And the query maybe very slow.

yulongfufu: Sorry, I'm not very familiar with Doris and the relationship between conjuncts and predicates. What I mean to say is that when there is no predicate, it may be better to rely on the limit, as that's how SR is written. I will find a better way to rewrite this code and do some testing. Thank you for your response :)

PR #22497, opened 2023-08-02
[Fix](test) Not run workload group by default

wangbo:
Even exclude this test in conf, but it may be still called.
So not run workload group by default in code.


wangbo:
run buildall

yiguolei (approved):
LGTM

BePPPower (approved):

PR #22635, opened 2023-08-05
[fix](planner) should not change SlotRef to NullLiteral

mrhhsg:
## Proposed changes

When executing the NullLiteral expression, a new column will be created, whereas executing the SlotRef expression does not create a new column. This difference can result in one unexpected column in the block.

## Further comments


fe/fe-core/src/main/java/org/apache/doris/analysis/Expr.java
```diff
@@ -934,7 +934,7 @@ public TExpr treeToThrift() {
             // being cast to a non-NULL type, the type doesn't matter and we can cast it
             // arbitrarily.
             Preconditions.checkState(this instanceof NullLiteral || this instanceof SlotRef);
-            return NullLiteral.create(ScalarType.BOOLEAN).treeToThrift();
+            type = ScalarType.BOOLEAN;
```
yiguolei, line 937: Could add some regression tests?

BiteTheDDDDt:
need to add some description

starocean999 (approved):

PR #21754, opened 2023-07-12
[Fix](planner)fix ctas incorrect string types of the target table.

sohardforaname:
## Proposed changes

Issue Number: close #xxx

string types from src table will be replaced to text type in ctas table, we change it to be corresponding to the src table.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #22755, opened 2023-08-08
Revert "[Fix](planner)fix ctas incorrect string types of the target table."

morningman:
Reverts apache/doris#21754
This will cause some ctas stmt return `varchar(1)`


morningman:
run buildall

PR #23012, opened 2023-08-15
[fix](regression) Improve the robustness when close target connection

deadlinefen:
bugfix:  
Improper use of syncer in the regression test framework may cause syncer to not initialize targetConnect when closing, resulting in a segment-fault. This PR adds a check for whether targetConnect exists


deadlinefen:
run buildall

JackDrogon (approved):

dataroaring (approved):
LGTM

PR #23240, opened 2023-08-21
[fix](test) case bug caused by groovy variable scope

shuke987:
## Proposed changes

in Groovy, variable without "dev" are global variables, which caused concurrent bugs.


shuke987:
out of date

PR #21236, opened 2023-06-27
[fix](function) to_bitmap parameter parsing failure returns null instead of bitmap_empty

zenoyang:
## Proposed changes

Issue Number: close #xxx
image
Null will be return when the input value is not in bigint range. But now more than this range will return `bitmap_empty` instead of `null`.

image

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #23258, opened 2023-08-21
[fix](function) Fix to_bitmap int parameter causing BE crash

zenoyang:
## Proposed changes

Issue Number: close #xxx
fix #21236 

## Further comments


zenoyang:
run buildall

zenoyang:
run buildall

zenoyang:
run clickbench

BiteTheDDDDt:
The storage layer of doris does not support nullable bitmap columns, so our bitmap should never be null. If there is something inconsistent with the document, we'd better modify the document instead of modifying the behavior

PR #23410, opened 2023-08-23
[fix](export)Export does not support viewing clear error messages

hf200012:
## Proposed changes
export does not support exporting views, but the submission can be successful. The information given to the user here is wrong. It should be clearly indicated that it is not supported, instead of allowing the user to view it through the show export command, and the information seen by the show export command is also inaccurate. Yes, but the prompt was canceled without knowing the specific reason。

Previous assignment submissions：
cf48106e-13f6-4477-8caa-2f212dc8d4f6

current job submission：
image

Issue Number: close #xxx

## Further comments


morningman:
@BePPPower

morningman:
Do not merge this, we will handle it in other PR

dataroaring:
run buildall

PR #23546, opened 2023-08-28
[feature](cache) 2q LRU cache

Hastyshell:
## Proposed changes

This PR aims to support 2q LRU cache, which is more adaptive to the query scenario, that big query should not phase out the hot data in cache.

The implementation is based on the [paper](https://www.vldb.org/conf/1994/P439.PDF) of 2q LRU algorithm.

Keypoints: 
1. 2q LRU cache, with the same interface of the original LRU cache. To be different, our A1out will simply record the key, because data pointer has been stored in meta and used during page IO. Therefore, data recorded in A1out will shared the reading path of non-cached data, but insert to Am queue. 
2. 2q LRU policy for page cache memory management. Page cache now implements LRU policy to process GC during.

## Further comments


dataroaring (changes requested):
Please add ut and add a config to control if use 2q in page_cache.

PR #23667, opened 2023-08-30
[fix](deploy) fix deploy manager can't drop node

zxealous:
## Proposed changes

Issue Number: close #xxx

We use  countermap to count the continuous detected down times, if the continuous down time is more then MAX_MISSING_TIME, we considered this node as down permanently. But the key of countermap is host, backend and broker may be deployed on the same host,  when checking the broker, it is possible to mistake the host of the be to be deleted and remove it from the countermap.

## Further comments


zxealous:
run buildall

luozenglin (approved):
LGTM

zddr (approved):
LGTM

PR #23682, opened 2023-08-30
[Feature](Nereids)support insert overwrite for Nereids

zddr:
## Proposed changes

Issue Number: close #xxx

- support insert overwrite for Nereids
- Solving the problem of table ID changing（When no partition is specified，automatically specify all partitions for table ）

todo: 
- Under the current processing logic, if the master restarts midway, it will generate dirty data (temp partition)
- for old planner ,insert overwrite table ,table id will change

## Further comments


zddr:
run buildall

fe/fe-core/src/main/java/org/apache/doris/catalog/OlapTable.java
```diff
@@ -936,7 +933,11 @@ public Partition dropPartition(long dbId, String partitionName, boolean isForceD
     // get partition by name, not including temp partitions
     @Override
     public Partition getPartition(String partitionName) {
```
morningman, line 935: Why change the behavior of this method?
The comment of this method show that it should not include temp partition

morningman (commented):
Need more regression test cases

fe/fe-core/src/main/java/org/apache/doris/catalog/OlapTable.java
zddr: BindSink. bindPartitionIds will convert partition name to partition id by calling this method，`OlapTable.getPartition(long partitionId)` has the same change , now I change the comment of two method

zddr:
run buildall

zddr:
run p0

fe/fe-core/src/main/java/org/apache/doris/nereids/trees/plans/commands/InsertIntoTableCommand.java
```diff
@@ -151,6 +179,153 @@ public void run(ConnectContext ctx, StmtExecutor executor) throws Exception {
...
+    public void dealOverwrite(ConnectContext ctx, StmtExecutor executor, PhysicalOlapTableSink<?> physicalOlapTableSink)
+            throws Exception {
+        OlapTable targetTable = physicalOlapTableSink.getTargetTable();
+        TableName tableName = new TableName(InternalCatalog.INTERNAL_CATALOG_NAME, targetTable.getQualifiedDbName(),
+                targetTable.getName());
+        List partitionNames = ((UnboundOlapTableSink) logicalQuery).getPartitions();
+        if (CollectionUtils.isEmpty(partitionNames)) {
+            partitionNames = Lists.newArrayList(targetTable.getPartitionNames());
+        }
+        List<String> tempPartitionNames = addTempPartition(ctx, tableName, partitionNames);
```
morningman, line 199: We should not contain any Class of old planner here, such as `TableName`, `ReplacePartitionClause`, etc.
But we need to write a lot more code to implement them in new planner.
So I suggest to add TODO here, and implement it later

morningman (approved):
LGTM

zy-kkk (approved):
LGTM

PR #23914, opened 2023-09-05
[Enhancement]The page cache can be parameterized by the session variable of fe.

Tech-Circle-48:
## Proposed changes

The page cache can be parameterized by the session variable of fe

## Further comments

Modified the original read configuration file to read enable page cache, currently modified to be able to configure through the session variable.


Tech-Circle-48:
Modified the original read enble page cache can be configured by session variable.

PR #23931, opened 2023-09-05
[improve](jsonb)json_extract is compatible with jsonb and string

LemonLiTree:
Three problems are solved here：
1. The function json_extract can perform different calculations depending on whether the parameter is jsonb or string.
2. Function jsonb_extract new optimizer supports variable parameters
3. If one parameter of the function is null, the result will directly return null.
4. fix non simd int128 overflow


LemonLiTree:
run buildall

be/src/vec/functions/simple_function_factory.h
```diff
@@ -164,12 +164,22 @@ class SimpleFunctionFactory {
 
         // if function is variadic, added types_str as key
         if (function_variadic_set.count(key_str)) {
-            for (auto& arg : arguments) {
-                key_str.append(arg.type->is_nullable()
-                                       ? reinterpret_cast<const DataTypeNullable*>(arg.type.get())
-                                                 ->get_nested_type()
-                                                 ->get_family_name()
-                                       : arg.type->get_family_name());
+            if (key_str == "json_extract") {
```
xiaokang, line 167: We should avoid special branch for specific function.

PR #23943, opened 2023-09-05
[chore](thirdparty)We need to issue an error when starting FE without setting the Java home environment

xyfsjq:
## Proposed changes

Issue Number: close #xxx

## Further comments


bin/start_fe.sh
```diff
@@ -108,7 +108,12 @@ if [[ -e "${DORIS_HOME}/bin/palo_env.sh" ]]; then
 fi
 
 if [[ -z "${JAVA_HOME}" ]]; then
-    JAVA="$(command -v java)"
+    if ! command -v java &>/dev/null; then
```
yiguolei, line 111: please add some comment here， why do we need write code like this.
Or the other developer may refactor code and rewrite this code.

xyfsjq: Due to the customer's machine not being configured with Java home, starting FE will result in startup failure, no error prompts, and logs. In this case, when the FE cannot start, it is necessary to prompt an error message indicating that it has not yet configured Java home.

adonis0147: Hi @xyfsjq , would you show the errors here if we don't set `JAVA_HOME` before starting FE?

By the way, if it was caused by the environment variable `JAVA_HOME` which is unset, we should check the variable `JAVA_HOME` straightforwardly.

xiaokang:
run buildall

adonis0147 (changes requested):

adonis0147 (approved):
LGTM

yiguolei:
run buildall

xiaokang (approved):
LGTM

PR #23960, opened 2023-09-06
[Fix](regression) Fix `test_partial_update_schema_change`

bobhan1:
## Proposed changes
When the regression test `test_partial_update_schema_change` try to wait for the current schema change to finish. It use `SHOW ALTER TABLE COLUMN WHERE TableName = '${tableName}' ORDER BY CreateTime DESC LIMIT 1;` to get the status of the current schema change job. Although the field `CreateTime` has precision up to millisecond, before sorting the results, it will first serialize the fields to string and the default string presentation of `CreateTime` only has precision up to seconds. So in order to ensure that the waiting job is the newest schema change job, the time interval between two alter jobs should have at least one second.

## Further comments


bobhan1:
run buildall

zhannngchen:
please add the description for this PR

zhannngchen (approved):
LGTM

Yukang-Lian (approved):

Issue #24068, opened 2023-09-08
[Feature](function) add function 'year_week_format'


nsnhuang:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

### Description

Calculates the year and week number of a date and converts it to a string of the format type.

The date param is a valid date. 
The last three params are described as follows, using the default values if not specified (first_day_of_week=0, first_week_of_year_policy=0, format='%YWK%W') 
Currently, strings of up to 16 bytes are supported, and NULL is returned if the return value is longer than 16 bytes.

1. first_day_of_week:The starting week of the week, where 0 is Monday, 6 is Sunday, and so on.
2. first_week_of_year_policy:How to determine the first week of the year, there are three strategies.
strategies 0：The week of January 1 is the first week of the year.
strategies 1：After January 1, the week in which the first week begins is the first week of the year.
strategies 2：The first week longer than four days is the first week of the year.
3. format: Specifies the output format for the date/datetime.

format The following formats are available:

%W | numbering of weeks, double-digit.

%w | numbering of weeks, minimum one-digit, maximum double-digit.

%Y | year, four figures.       
                           
%y | year, Last two digits.

%% | express %

### Use case

mysql> select year_week_format(20240101);
+----------------------------+
| year_week_format(20240101) |
+----------------------------+
| 2024WK01                   |
+----------------------------+

mysql> select year_week_format(20240101,0,0,'%YWK%W');
+--------------------------------------------+
| year_week_format(20240101, 0, 0, '%YWK%W') |
+--------------------------------------------+
| 2024WK01                                   |
+--------------------------------------------+

mysql> select year_week_format(20240101,1,1,'%YWK%W');
+--------------------------------------------+
| year_week_format(20240101, 1, 1, '%YWK%W') |
+--------------------------------------------+
| 2023WK52                                   |
+--------------------------------------------+

mysql> select year_week_format(20240101,3,2,'%YWK%W');
+--------------------------------------------+
| year_week_format(20240101, 3, 2, '%YWK%W') |
+--------------------------------------------+
| 2023WK53                                   |
+--------------------------------------------+

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #24069, opened 2023-09-08
[feature](function) add function 'year_week_format'

nsnhuang:
## Proposed changes

Issue Number: #24068 

add Func year_week_format
Calculates the year and week number of a date and converts it to a string of the format type.

The date param is a valid date. 
The last three params are described as follows, using the default values if not specified (first_day_of_week=0, first_week_of_year_policy=0, format='%YWK%W') 
Currently, strings of up to 16 bytes are supported, and NULL is returned if the return value is longer than 16 bytes.

1. first_day_of_week:The starting week of the week, where 0 is Monday, 6 is Sunday, and so on.
2. first_week_of_year_policy:How to determine the first week of the year, there are three strategies.
strategies 0：The week of January 1 is the first week of the year.
strategies 1：After January 1, the week in which the first week begins is the first week of the year.
strategies 2：The first week longer than four days is the first week of the year.
3. format: Specifies the output format for the date/datetime.

format The following formats are available:

%W | numbering of weeks, double-digit.

%w | numbering of weeks, minimum one-digit, maximum double-digit.

%Y | year, four figures.       
                           
%y | year, Last two digits.

%% | express %


be/src/vec/functions/function_datetime_string_to_string.h
```diff
@@ -41,20 +41,81 @@
 #include "vec/functions/date_time_transforms.h"
 #include "vec/functions/function.h"
 
+#define GET_COLUMN_CONST_RETURN_IF_ERROR(Index, ColumnType)                                                         \
```
zclllyybb, line 44: it seems no need to use Marco here. please use a function to replace it.

be/src/vec/functions/function_datetime_string_to_string.h
```diff
@@ -41,20 +41,81 @@
...
+public:
+    static constexpr auto name = Transform::name;
+    static constexpr bool has_variadic_argument =
+            !std::is_void_v<decltype(has_variadic_argument_types(std::declval<Transform>()))>;
+    static FunctionPtr create() { return std::make_shared<FunctionDateTimeStringToStringOneArg>(); }
+    String get_name() const override { return name; }
+    size_t get_number_of_arguments() const override { return 0; }
+    bool is_variadic() const override { return true; }
+    DataTypes get_variadic_argument_types_impl() const override {
+        if constexpr (has_variadic_argument) return Transform::get_variadic_argument_types();
```
zclllyybb, line 71: this will break format rule. please use builtin clang-format file to format your pr.

be/src/vec/functions/function_datetime_string_to_string.h
```diff
@@ -41,20 +41,81 @@
...
+                        size_t result, size_t input_rows_count) override {
+        const ColumnPtr source_col = block.get_by_position(arguments[0]).column;
+        const auto* nullable_column = check_and_get_column<ColumnNullable>(source_col.get());
+        const auto* sources = check_and_get_column<ColumnVector<typename Transform::FromType>>(
+                nullable_column ? nullable_column->get_nested_column_ptr().get()
+                                : source_col.get());
+        if (sources) {
+            auto col_res = ColumnString::create();
+            ColumnUInt8::MutablePtr col_null_map_to;
+            col_null_map_to = ColumnUInt8::create();
```
zclllyybb, line 90: why divide it to two statements?

be/src/vec/runtime/vdatetime_value.cpp
```diff
@@ -3812,4 +3812,32 @@ template bool DateV2Value<DateTimeV2ValueType>::datetime_trunc<TimeUnit::YEAR>()
 template bool DateV2Value<DateTimeV2ValueType>::datetime_trunc<TimeUnit::QUARTER>();
 template bool DateV2Value<DateTimeV2ValueType>::datetime_trunc<TimeUnit::WEEK>();
 
+CalcFirstWeekPolicy* CalcFirstWeekPolicy::get_policy_by_code(uint8_t code) {
+    static std::unordered_map<uint8_t, CalcFirstWeekPolicy*> instances;
```
zclllyybb, line 3816: when multithreads hits the if branch below, there's data racing problem. this function could be refactor. consider preload it or remove this cache?

regression-test/suites/query_p0/sql_functions/datetime_functions/test_date_function.groovy
```diff
@@ -719,4 +719,16 @@ suite("test_date_function") {
     assertFalse(res.contains("date_trunc"))
 
     qt_sql """ select date_add("2023-08-17T01:41:18Z", interval 8 hour) """
+
+    // YEAR_WEEK_FORMAT
```
zclllyybb, line 723: please add some testcases which data is from table.

be/src/vec/functions/function_datetime_string_to_string.h
```diff
@@ -117,4 +178,85 @@ class FunctionDateTimeStringToString : public IFunction {
...
+        if constexpr (has_variadic_argument) return Transform::get_variadic_argument_types();
+        return {};
+    }
+
+    DataTypePtr get_return_type_impl(const ColumnsWithTypeAndName& arguments) const override {
+        return make_nullable(std::make_shared<DataTypeString>());
+    }
+
+    bool use_default_implementation_for_nulls() const override { return false; }
+    bool use_default_implementation_for_constants() const override { return true; }
```
zclllyybb, line 204: when use its' default value, no need to override it again.

be/src/vec/functions/function_datetime_string_to_string.h
```diff
@@ -41,20 +41,81 @@
 #include "vec/functions/date_time_transforms.h"
 #include "vec/functions/function.h"
 
+#define GET_COLUMN_CONST_RETURN_IF_ERROR(Index, ColumnType)                                                         \
+    const IColumn& source_col##Index = *block.get_by_position(arguments[Index]).column;                 \
+    const auto* delta_const_column##Index = typeid_cast<const ColumnType*>(&source_col##Index);        \
```
zclllyybb, line 46:
```suggestion
    const auto& delta_const_column##Index = assert_cast<const ColumnType&>(assert_cast<const ColumnConst*>(source_col##Index.get())->get_data_column_ptr())\
```

be/src/vec/functions/function_datetime_string_to_string.h
```diff
@@ -117,4 +178,85 @@ class FunctionDateTimeStringToString : public IFunction {
...
+            ColumnUInt8::MutablePtr col_null_map_to;
+            col_null_map_to = ColumnUInt8::create();
+            auto& vec_null_map_to = col_null_map_to->get_data();
+
+            if(arguments.size() == 4) {
+                GET_COLUMN_CONST_RETURN_IF_ERROR(1, ColumnConst);
+                GET_COLUMN_CONST_RETURN_IF_ERROR(2, ColumnConst);
+                GET_COLUMN_CONST_RETURN_IF_ERROR(3, ColumnConst);
+
+                Int32 first_day_of_week = delta_const_column1->get_field().get<Int32>();
```
zclllyybb, line 227: `delta_const_column##x` changed. then change this to elimate virtual function call.

zclllyybb (changes requested):

PR #24077, opened 2023-09-08
[supplement](regression-test) Pass CCR cases when Fe enable_feature_binlog=false

deadlinefen:
## Proposed changes

Before starting the CCR case test, use syncer.checkEnableFeatureBinlog() to determine whether enable_feature_binlog is true. If true, skip this case.


deadlinefen:
run buildall

JackDrogon (approved):
LGTM

dataroaring (approved):
LGTM

PR #24084, opened 2023-09-08
[fix]Add logging before _builtin_unreachable()

Tech-Circle-48:
## Proposed changes

Issue Number: 

## Further comments
Add FATAL logging before _builtin_unreachable()


be/src/pipeline/exec/hashjoin_build_sink.cpp
```diff
@@ -173,6 +173,7 @@ Status HashJoinBuildSinkLocalState::process_build_block(RuntimeState* state,
                     [&](std::monostate& arg, auto has_null_value,
                         auto short_circuit_for_null_in_build_side) -> Status {
                         LOG(FATAL) << "FATAL: uninited hash table";
+                        LOG(FATAL) << "__builtin_unreachable";
```
yiguolei, line 176: duplicate, there is already a LOG FATAL before it.

PR #24089, opened 2023-09-08
[fix](nereids) avoid throw analysis ex for unsupported type to make ext table goes nereids

xzj7019:
## Proposed changes

avoid throw analysis ex for unsupported type to make ext table goes nereids.
this will improve the nereids' availability for external table if unsupported type is in the basic table schema but not referenced in the real sql.

tested in external table env.

Consider the following case:
select pu.pk_ct_pu as id
  from fms_rd_nc65_zb.NC65P.CT_PU pu
  left join fms_rd_nc65_zb.NC65P.PUB_WF_INSTANCE pwi
    on pu.pk_ct_pu = pwi.billid 
    and pu.vtrantypecode=pwi.billtype
 left join fms_rd_nc65_zb.NC65P.SM_USER su
    on pu.creator = su.cuserid
 where pu.pk_ct_pu='1001A110000000K8XPVN'; 

PUB_WF_INSTANCE table has a BLOB type column and currently it will throw analysis exception and fallback to old optimizer, although this column is not referenced in the real sql. The old optimizer doesn't have the outer join -> inner join rule and the "pu.pk_ct_pu='1001A110000000K8XPVN'; " is not pushed down and the performance will drop down. After the pr, we add the unsupported type instead of throw exception directly, it will decide the unused case and can continue goes nereids and use all advanced optimization for the sql.

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/nereids/rules/rewrite/CheckDataTypes.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/nereids/rules/rewrite/CheckDataTypes.java
@@ -30,6 +30,7 @@ import org.apache.doris.nereids.types.DataType;
 import org.apache.doris.nereids.types.JsonType;
 import org.apache.doris.nereids.types.MapType;
 import org.apache.doris.nereids.types.StructType;
+import org.apache.doris.nereids.types.UnsupportedType;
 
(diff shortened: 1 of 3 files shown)


xzj7019:
run buildall

morrySnow (approved):

starocean999 (approved):

PR #21746, opened 2023-07-12
[feature](partial update) add failure tolerance for strict mode partial update stream load

bobhan1:
## Proposed changes

1. add the number of newly inserted rows to num_filtered_rows in strict mode partial update stream load
2. use delete_bitmap to delete the newly inserted rows in strict mode partial update
3. If num_rows_load_filtered/num_selected_rows > max_filter_ratio, the strict mode partial update stream load will fail.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #24191, opened 2023-09-11
[Fix](merge-on-write) Check the returned filtered rows from different replicas

bobhan1:
## Proposed changes

In https://github.com/apache/doris/pull/21746, we just add up all filtered rows reported up by the executor BEs. When the table has replicas, it will get the wrong number of rows filtered. This PR adds consistency check for reported filtered rows and correct the calculation of the number of filtered rows in partial update.
## Further comments

### DIFF
--- a/be/src/vec/sink/vtablet_sink.cpp
+++ b/be/src/vec/sink/vtablet_sink.cpp
@@ -47,6 +47,7 @@
 #include <iterator>
 #include <memory>
 #include <mutex>
+#include <numeric>
 #include <sstream>
 #include <string>
 #include <type_traits>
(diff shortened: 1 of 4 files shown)


bobhan1:
run buildall

fe/fe-common/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -1453,7 +1453,7 @@ public class Config extends ConfigBase {
      * DO NOT use it for production env.
      */
     @ConfField
-    public static boolean allow_replica_on_same_host = false;
+    public static boolean allow_replica_on_same_host = true;
```
cambyzju, line 1456: this config is `"DO NOT use it for production env"`, we should not open it by default.

bobhan1:
run buildall

bobhan1:
run buildall

bobhan1:
run clickbench

zhannngchen (approved):
LGTM

regression-test/suites/unique_with_mow_p0/partial_update/test_partial_update_strict_mode.groovy
```diff
@@ -30,7 +30,7 @@ suite("test_partial_update_strict_mode", "p0") {
                 `last_access_time` datetime NULL 
             ) ENGINE = OLAP UNIQUE KEY(`id`) 
             COMMENT 'OLAP' DISTRIBUTED BY HASH(`id`) 
-            BUCKETS AUTO PROPERTIES ( 
+            BUCKETS 4 PROPERTIES (
```
cambyzju, line 33: Why we need to change auto bucket to static bucket? It looks this pr have no connected with auto bucket?

bobhan1: just to ensure that the table has more than one buckets to check if the calculation is correct if the table has more than one bucket.

be/src/vec/sink/vtablet_sink.cpp
```diff
@@ -165,9 +166,20 @@ class IndexChannel {
 
     void set_tablets_received_rows(
             const std::vector<std::pair<int64_t, int64_t>>& tablets_received_rows, int64_t node_id);
+    void set_tablets_filtered_rows(
+            const std::vector<std::pair<int64_t, int64_t>>& tablets_filtered_rows, int64_t node_id);
+    int64_t num_rows_filtered() {
```
cambyzju, line 171: it seems not a complete function define. how to pass the compile?

bobhan1: which one

cambyzju (approved):
LGTM

PR #24212, opened 2023-09-12
[fix](forward master op)Set default catalog and db only when they exist in master FE while executing forwarded stmt.

Jibing-Li:
In this case, forward to master will throw catalog or db not found exception:
Connect to a follower:
1. create database test
2. use test
3. drop database test
4. create database test

This is because after step 2, the default db in follower has been set to `test`, drop database will not change the default db. In step 4, the default db `test` is set and forwarded to master, and master will fail to find it because it is already dropped.

This pr is to set the default catalog and db only when they exist.

The actual reason is that, when Follower handle the `drop db` stmt, it will forward to master to execute it, but can not
unset its own "current db"

## Proposed changes

Issue Number: close #xxx

## Further comments


Jibing-Li:
run buildall

Jibing-Li:
run buildall

morningman (approved):
LGTM

wuwenchi (approved):

PR #24297, opened 2023-09-13
[fix](io): use try with resource make io stream close automatically to avoid resource leak

xutaoustc:
Some code are scanned to have potential security and performance problems:
This pr fix the problem that some iostream are not closed after using.
Use try with resource make io stream close automatically to avoid resource leak

### DIFF
--- a/fe/fe-common/src/main/java/org/apache/doris/common/io/DeepCopy.java
+++ b/fe/fe-common/src/main/java/org/apache/doris/common/io/DeepCopy.java
@@ -48,7 +48,6 @@ public class DeepCopy {
                 DataOutputStream out = new DataOutputStream(byteArrayOutputStream)) {
             orig.write(out);
             out.flush();
-            out.close();
 
             try (DataInputStream in = new DataInputStream(byteArrayOutputStream.getInputStream())) {
                 Method readMethod = c.getDeclaredMethod(READ_FIELDS_METHOD_NAME, DataInput.class);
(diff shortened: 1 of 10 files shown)


fe/fe-common/src/main/java/org/apache/doris/common/io/DeepCopy.java
```diff
@@ -44,18 +44,16 @@ public static boolean copy(Writable orig, Writable dest, Class c, int metaVersio
...
         metaContext.setThreadLocalInfo();
 
-        FastByteArrayOutputStream byteArrayOutputStream = new FastByteArrayOutputStream();
-        DataOutputStream out = new DataOutputStream(byteArrayOutputStream);
-        try {
+        try (FastByteArrayOutputStream byteArrayOutputStream = new FastByteArrayOutputStream();
+                DataOutputStream out = new DataOutputStream(byteArrayOutputStream)) {
             orig.write(out);
             out.flush();
             out.close();
```
yiguolei, line 51: Is this line useful?

fe/fe-common/src/main/java/org/apache/doris/common/io/DeepCopy.java
```diff
@@ -74,19 +72,17 @@ public static <T> T copy(Writable orig, Class<T> c, int metaVersion) {
...
```
yiguolei, line 79: and this line

fe/fe-common/src/main/java/org/apache/doris/common/io/DeepCopy.java
CalvinKirs: don't need to explicitly invoke close

yiguolei (approved):
LGTM

yiguolei:
run buildall

CalvinKirs (approved):

PR #24416, opened 2023-09-15
[FIX](decimalv3) fix decimalv3 value with leading zeros

amorynan:
## Proposed changes

Issue Number: close #xxx
now we make error if we deal with leading zeros in decimal value , type_precision >= precision will make value overflow and DCHECK will fail , so if here has leading zero we should only make type_precision > precision to make value right

## Further comments


amorynan:
run buildall

Gabriel39 (approved):
LGTM

BiteTheDDDDt (approved):

PR #24454, opened 2023-09-15
[fix](security): non-static inner class should not implement serialized interface, or when it is serialized it will contain outer class info, which is not safe

xutaoustc:
[fix](*): non-static inner class should not implement serialized interface, or when it is serialized it will contain outer class info, which is not safe

And in this scenario, the class does not use info of outer class, which should use static class instead


yiguolei:
run  buildall

yiguolei (approved):
LGTM

xiaokang (approved):
LGTM

PR #24483, opened 2023-09-15
[fix](security): Use SecureRandom instead of Random, because it provides better security

xutaoustc:
Use SecureRandom instead of Random, because it provides better security

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/cooldown/CooldownConfHandler.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/cooldown/CooldownConfHandler.java
@@ -28,6 +28,7 @@ import com.google.common.collect.Maps;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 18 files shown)


fe/fe-core/src/main/java/org/apache/doris/cooldown/CooldownConfHandler.java
```diff
@@ -78,7 +79,7 @@ private void updateCooldownConf(List<CooldownConf> confToUpdate) {
             if (replicas.isEmpty()) {
                 continue;
             }
-            Random rand = new Random(System.currentTimeMillis());
+            Random rand = new SecureRandom();
```
yiguolei, line 82: Not need a seed any more???

xutaoustc: yes, securerandom can generate a more safe seed

bigben0204:
LGTM

bigben0204 (approved):

yiguolei:
run buildall

bigben0204 (approved):

yiguolei (approved):
LGTM

yiguolei:
run feut

yiguolei (approved):
LGTM

liaoxin01:
@xutaoustc , SecureRandom seems to cause performance issues. What's the main purpose of switching to using SecureRandom?

PR #24489, opened 2023-09-16
[fix](nereids) fix cte filter pushdown if the filters can be aggregated

xzj7019:
## Proposed changes

Current cte common filter extraction doesn't work if the filters can be aggregated, which will lead the common filter can't be pushed down inside cte. Consider the following case:
with main as (select c1 from t1) select * from (select m1.* from main m1, main m2 where m1.c1 = m2.c1) abc where c1 = 1;
The common c1=1 filter can't be pushed down.

This pr fixed the original extraction logic from set to list to make the logic works, and this will also resolve the tpcds query4/11's pattern works well also.

## Further comments


xzj7019:
run buildall

morrySnow (approved):

yiguolei (approved):
LGTM

PR #24589, opened 2023-09-19
[fix](Nereids) fix row count unconsistent when join ordering

keanji-x:
## Proposed changes

In the context of reorder join, when a new plan is generated, it may include a project operation. In this case, the newly generated join root and the original join root will no longer be in the same group. To avoid inconsistencies in the statistics between these two groups, we keep the child group's row count unchanged when the parent group expression is a project operation.

## Further comments


XieJiann:
run buildall

XieJiann:
run buildall

englefly (approved):

jackwener (approved):

PR #24605, opened 2023-09-19
[fix](io): use try with resource make io stream close automatically to avoid resource leak

xutaoustc:
Some code are scanned to have potential security and performance problems:
This pr fix the problem that some iostream are not closed after using.
Use try with resource make io stream close automatically to avoid resource leak


yiguolei (approved):
LGTM

yiguolei:
run buildall

morningman (approved):
LGTM

PR #24635, opened 2023-09-19
[fix](broker) fix broker read issue

morningman:
## Proposed changes

The given "length" of broker's pread() method is the buffer length, not the length required from file.
So it may larger than the file length.
So we should return all read data, instead of return EOF when `read()` method return -1

I will add regression test case later when the framework support broker process.

## Further comments


morningman:
run buildall

kaka11chen (approved):
LGTM.

wsjz (approved):
LGTM

xiaokang (approved):
LGTM

PR #24648, opened 2023-09-20
[improvement](docker) Refine runtime docker image related files

murong00:
## Proposed changes

The new version of the download package bundles `fe/be/extensions` into one tar file, so we should refine the docker image related files to correct the building process. This pull request modifies `Dockerfile` and related document about building docker images.


murong00:
run buildall

PR #24795, opened 2023-09-22
[opt](status) add INVALID_DATA_FORMAT status code for serde

morningman:
## Proposed changes

Add new status code on BE: `INVALID_DATA_FORMAT`,
mainly use for indicating the error when parsing data format failed, such as parsing `xxx` to integer.

In previous, we use `INVALID_ARGUMENT` in these cases,
but it will print a lot error stack log in be.INFO when error happens.

One option is that we can add `INVALID_ARGUMENT` to `capture_stacktrace` to suppress its stack log,
but it may effect other error cases.
So I decide to add a new status code for parsing data format error

## Further comments


amorynan (approved):
LGTM

PR #24834, opened 2023-09-24
[bugfix](restore) Add partition id into convert_rowset_ids()

deadlinefen:
## BUG
The partition_id to be converted is not specified in the convert_rowset_ids() function of the RESTORE operation.This can result in the partition_id in the new tablet meta not being converted, but remaining the same as the partition_id in the snapshot file, which in turn resulting in a mismatch between the tablet and the partition, causing metadata errors.
## FIX
Add the parameter partition_id to the convert_rowset_ids() function and add the partition_id to the Rpc structure CloneReq to ensure proper convert rowset ids

### DIFF
--- a/be/src/olap/snapshot_manager.cpp
+++ b/be/src/olap/snapshot_manager.cpp
@@ -125,7 +125,8 @@ Status SnapshotManager::release_snapshot(const string& snapshot_path) {
 }
 
 Status SnapshotManager::convert_rowset_ids(const std::string& clone_dir, int64_t tablet_id,
-                                           int64_t replica_id, const int32_t& schema_hash) {
+                                           int64_t replica_id, int64_t partition_id,
+                                           const int32_t& schema_hash) {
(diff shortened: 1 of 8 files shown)


deadlinefen:
run buildall

deadlinefen:
run buildall

be/src/olap/snapshot_manager.cpp
```diff
@@ -160,6 +161,9 @@ Status SnapshotManager::convert_rowset_ids(const std::string& clone_dir, int64_t
     new_tablet_meta_pb.set_tablet_id(tablet_id);
     *new_tablet_meta_pb.mutable_tablet_uid() = TabletUid::gen_uid().to_proto();
     new_tablet_meta_pb.set_replica_id(replica_id);
+    if (partition_id != -1) {
```
JackDrogon, line 164: 这个我突然感觉用optional<int64_t>更好些

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #24845, opened 2023-09-25
[improvement](tablet clone) fix balanced new replica will be removed when load txn continuously 

yujun777:
## Proposed changes

For a new clone replica, its init state is CLONE. During CLONE state, the replica can not load data.

When cloning finishing,  check this replica is catchup with the partition.  Check new replica's version is not enough.  Because replica not write txn during CLONE,   when these txns change from prepare to commit or visible, the new replica will be fallbehind (last failed version > 0).  Later check this tablet is reduant and the new replica will be removed cause its last failed version > 0.

To fix this,  when clone a new replica finished, we should check the txns during CLONE should be finished.

TEST:  Let write txn continuously and none-stop,  and make be clone a full snapshot cost 20s.  Then let balance occur,   later the new replica can success cloning after two or three further repair,   and the old replica will be removed later.

## Further comments

### DIFF
--- a/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
@@ -930,6 +930,18 @@ public class Config extends ConfigBase {
     public static long tablet_repair_delay_factor_second = 60;
 
     /**
+     * clone a tablet, further repair timeout.
+     */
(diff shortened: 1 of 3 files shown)


yujun777:
run buildall

Yukang-Lian (approved):

dataroaring (approved):
LGTM

PR #24011, opened 2023-09-07
[Enhancement](merge-on-write) use delete bitmap to mark delete for rows with delete sign when sequence column doesn't exist

bobhan1:
## Proposed changes
For Unique table with merge-on-write enabled, we can directly use delete bitmap to mark the rows with delete sign as deleted when sequence column doesn't exist to eliminate reading delete sign columns in later reads.
This PR can also improve the performance of https://github.com/apache/doris/pull/21773 if we disable `insert ignore` statement on mow Unique table with sequence column.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #24877, opened 2023-09-25
[Fix](merge-on-write) Correct the alignment process when the existing rows with same key has marked delete sign

bobhan1:
## Proposed changes
When we insert a row with delete sign marked, it will not mark itself on the delete bitmap if the table has sequence column(see https://github.com/apache/doris/pull/24011). So if we delete a row using delete sign and then insert a row with same keys using partial update, it will wrongly fill the values in in the unmentioned columns in which the delete sign column will be filled with 1 and the row that should have been inserted will be deleted. So in this occasion we must check the delete sign column to check that if a row really exists in the table.

## Further comments


bobhan1:
run buildall

bobhan1:
run buildall

regression-test/data/unique_with_mow_p0/partial_update/delete_sign2.csv
```diff
@@ -0,0 +1 @@
+1,2
```
zhannngchen, line 1: 这个文件最好改一下名字，delete_sign.csv是用来写带delete sign的数据的，这个文件我最开始也以为是要写delete sign的数据，最后才发现是用来更新v1的。这个名字有误导性

bobhan1:
run buildall

zhannngchen (approved):
LGTM

Yukang-Lian (approved):

Issue #24858, opened 2023-09-25
[Bug] query profile api of fe cann't get result if non-root user query on the other fe


bigben0204:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

### Version

2.0.1

### What's Wrong?

In queryProfileText api, if IS_ALL_NODE_PARA is true, profile might be queried from other fe. 
While **checkAuthByUserAndQueryId** function stopped non-root user to query from other fe, and an error "query with id xxx is not found" is returned.

### What You Expected?

Querying profile from any fe for non-root user will return correct result.

### How to Reproduce?

1. Execute a query with non-root user.
2. Use queryProfileText api to query profile on the other fe.

### Anything Else?

To solve the problem, when IS_ALL_NODE_PARA is true, **checkAuthByUserAndQueryId** function shoud not be executed. When IS_ALL_NODE_PARA is false, **checkAuthByUserAndQueryId** function shoud be executed.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #24914, opened 2023-09-26
 [bug](profile) query profile api of fe cann't get result if non-root user query on the other fe 

bigben0204:
## Proposed changes

Issue Number: #24858

If isAllNode is true, the api should only distribute the query to all fe and do not run checkAuthByUserAndQueryId.
If isAllNode is false, the api queries profile on the fe, at this time the api should  run checkAuthByUserAndQueryId.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/httpv2/rest/manager/QueryProfileAction.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/httpv2/rest/manager/QueryProfileAction.java
@@ -268,10 +268,12 @@ public class QueryProfileAction extends RestBaseController {
             @RequestParam(value = IS_ALL_NODE_PARA, required = false, defaultValue = "true") boolean isAllNode) {
         executeCheckPassword(request, response);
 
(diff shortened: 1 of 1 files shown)


yiguolei:
If isAllNode is true, the api should only distribute the query to all fe and do not run checkAuthByUserAndQueryId.

I  think we should always check auth??? Why not check auth???

@bigben0204

bigben0204:
> If isAllNode is true, the api should only distribute the query to all fe and do not run checkAuthByUserAndQueryId.
> 
> I think we should always check auth??? Why not check auth???
> 
> @bigben0204

When isAllNode is true, the API only distributes the request to all fes with parameter isAllNode as false. Checking auth is deferred when isAllNode is false on each fe.

yiguolei (approved):
LGTM

yiguolei:
run buildall

xinyiZzz (approved):
LGTM

HappenLee (approved):
LGTM

PR #24964, opened 2023-09-27
[fix](compaction) block reader should check column string offset

hust-hhb:
## Proposed changes
there is some bug like cats will make data wrong, compaction will core when reading wrong data, so need to add a check on column string.

Issue Number: close #xxx

## Further comments

### DIFF
--- a/be/src/vec/olap/block_reader.cpp
+++ b/be/src/vec/olap/block_reader.cpp
@@ -419,11 +419,10 @@ Status BlockReader::_unique_key_next_block(Block* block, bool* eof) {
 
 Status BlockReader::_insert_data_normal(MutableColumns& columns) {
     auto block = _next_row.block.get();
-    RETURN_IF_CATCH_EXCEPTION(for (auto idx
-                                   : _normal_columns_idx) {
-        columns[_return_columns_loc[idx]]->insert_from(*block->get_by_position(idx).column,
-                                                       _next_row.row_pos);
-    });
+    for (auto idx : _normal_columns_idx) {
+        RETURN_IF_CATCH_EXCEPTION(columns[_return_columns_loc[idx]]->insert_from(
+                *block->get_by_position(idx).column, _next_row.row_pos));
+    }
     return Status::OK();
 }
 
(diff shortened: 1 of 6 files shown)


hust-hhb:
run buildall

be/src/vec/olap/vertical_block_reader.cpp
```diff
@@ -524,10 +524,10 @@ Status VerticalBlockReader::_unique_key_next_block(Block* block, bool* eof) {
         }
         const auto& src_block = _next_row.block;
         assert(src_block->columns() == column_count);
-        for (size_t i = 0; i < column_count; ++i) {
+        RETURN_IF_CATCH_EXCEPTION(for (size_t i = 0; i < column_count; ++i) {
```
yiguolei, line 527: not catch the loop.
just catch the internal  insert_from is ok

be/src/vec/olap/block_reader.cpp
```diff
@@ -407,12 +407,14 @@ Status BlockReader::_unique_key_next_block(Block* block, bool* eof) {
     return Status::OK();
 }
 
-void BlockReader::_insert_data_normal(MutableColumns& columns) {
+Status BlockReader::_insert_data_normal(MutableColumns& columns) {
     auto block = _next_row.block.get();
-    for (auto idx : _normal_columns_idx) {
+    RETURN_IF_CATCH_EXCEPTION(for (auto idx
```
yiguolei, line 412: not catch the for loop
catch the insert from method

be/src/vec/olap/block_reader.h
```diff
@@ -80,7 +80,7 @@ class BlockReader final : public TabletReader {
 
     void _init_agg_state(const ReaderParams& read_params);
 
-    void _insert_data_normal(MutableColumns& columns);
+    Status _insert_data_normal(MutableColumns& columns);
```
yiguolei, line 83: add [no discard] attribute

hust-hhb:
run buildall

hust-hhb:
run buildall

PR #25013, opened 2023-09-28
[fix](single-replica-load) fix duplicated done run in request_slave_tablet_pull_rowset

freemandealer:
BE will crash because done run twice when try_offer() failed in request_slave_tablet_pull_rowset.

## Proposed changes

Issue Number: close #xxx

## Further comments

### DIFF
--- a/be/src/service/internal_service.cpp
+++ b/be/src/service/internal_service.cpp
@@ -1498,7 +1498,7 @@ void PInternalServiceImpl::request_slave_tablet_pull_rowset(
                                     rowset_meta->tablet_id(), node_id, true);
     });
     if (!ret) {
-        offer_failed(response, done, _heavy_work_pool);
+        offer_failed(response, closure_guard.release(), _heavy_work_pool);
         return;
     }
     Status::OK().to_protobuf(response->mutable_status());


freemandealer:
run buildall

xiaokang (approved):
LGTM

be/src/service/internal_service.cpp
```diff
@@ -1498,7 +1498,7 @@ void PInternalServiceImpl::request_slave_tablet_pull_rowset(
                                     rowset_meta->tablet_id(), node_id, true);
     });
     if (!ret) {
-        offer_failed(response, done, _heavy_work_pool);
```
yiguolei, line 1501: not modify like this.

move  brpc::ClosureGuard closure_guard(done); to try offer code.

freemandealer: Unlike other RPC impl, true response of pull rowset operation is sent to the caller by a new RPC called `response_slave_tablet_pull_rowset`. So here, we need to acknowledge the current `request_slave_tablet_pull_rowset` ASAP.  We do need more discussion here to decide how we can fix it decently (and more on the thread pool single replica load is using). Let's freeze this PR for a while.

dataroaring (approved):
LGTM

PR #25112, opened 2023-10-08
[fix](fe rest api)api gets execution plan, table name case problem

hf200012:
## Proposed changes
The user has configured the parameter lower_case_table_names, which ignores the case of the table name. When executed on the SQL client, the table name can be queried in both case.
But when using Connector to read doris data, the table names must be in the same case, otherwise an error will be reported.

image

image

connector reading error：
image

Issue Number: close #xxx

## Further comments


liugddx:
The current configuration is easily confusing.We need to unify the configuration.

image
image

https://doris.apache.org/docs/dev/lakehouse/multi-catalog/jdbc?_highlight=lower_case_table_names#lowercase-table-name-synchronization

https://doris.apache.org/docs/dev/admin-manual/config/fe-config?_highlight=lower_case_table_names#lower_case_table_names

zy-kkk (approved):
LGTM

zy-kkk:
run buildall

yiguolei (approved):
LGTM

PR #24412, opened 2023-09-15
(enhancement)[fe] Add `isMaster()` check for `FrontendService`

swjtu-zhanglei:
* In `FrontendServiceImpl` service, the api which need to write editlog need add `isMaster()` check

## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #25158, opened 2023-10-09
Pick "(enhancement)[fe] Add `isMaster()` check for `FrontendService` (#24412)"

swjtu-zhanglei:
* In `FrontendServiceImpl` service, the api which need to write editlog need add `isMaster()` check


SWJTU-ZhangLei:
run buildall

PR #25200, opened 2023-10-09
[Fix](func numbers) Remove backend_nums argument of numbers function

zhiqiang-hhhh:
We have some bugs in function numbers:

1. When we have real backends less than requested backend_nums, we do not throw exception.
2. Current thrift definition of function numbers do not support number partition between multiple backends or multiple threads, since we just have total_numbers without range parameter.
3. Computation of execution parallelism is wrong.  

And we do not have requests for parallel execute of function numbers. So we remove backend_num argument.


zhiqiang-hhhh:
run buildall

yiguolei (dismissed):
LGTM

zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

yiguolei:
run buildall

zhiqiang-hhhh:
run BE UT

yiguolei (approved):
LGTM

wangbo (approved):
LGTM

PR #24029, opened 2023-09-07
[Refactor & Bugfix](static variables) move some static vairables to exec_env

zhiqiang-hhhh:
## Proposed changes

First pr to fix issue #24082, aiming to fix be process core during exiting.

PR #25252, opened 2023-10-10
[opt](log) Port conflict message improvement

zhiqiang-hhhh:
Doris main of branch 2.0 can not pick https://github.com/apache/doris/pull/24029. 
As a result, when port conflict happens, doris main will do gracefully exit. On this situation, be.out will print something like `invalid id` instead of port conflict. 

This pr aims to correct error message of above situation and no need to be merge to master, since it will do gracefully exit correctly.


yiguolei (approved):
LGTM

PR #25352, opened 2023-10-12
[enhancement](merge-on-write) consider version count on size-based cu compaction policy

zhannngchen:
## Proposed changes

Issue Number: close #xxx

For MoW table, if there's too many versions, the delete bitmap will grow to a very big size, which may cause the tablet meta too big and the `save_meta` operation too slow.
If the rowset should not promotion according to it's disk size, we should also consider it's version count.

## Further comments


zhannngchen:
run buildall

hust-hhb (approved):
LGTM

liaoxin01 (approved):
LGTM

dataroaring (approved):
LGTM

PR #25366, opened 2023-10-12
[improvement](scanner) Remove the predicate that is always true for the segment

mrhhsg:
## Proposed changes

By utilizing the zonemap index of the segment, we can ascertain if a predicate is always true. For example, if the segment’s maximum value is 100 and the predicate is col < 101, then this predicate is always true for this segment.

## Further comments


mrhhsg:
run buildall

be/src/olap/rowset/segment_v2/segment.cpp
```diff
@@ -157,6 +156,19 @@ Status Segment::new_iterator(SchemaSPtr schema, const StorageReadOptions& read_o
         iter->reset(new SegmentIterator(this->shared_from_this(), schema));
     }
 
+    auto pruned_predicates = read_options.column_predicates;
+    auto pruned = false;
```
yiguolei, line 160: add some regression test

be/src/olap/comparison_predicate.h
```diff
@@ -174,6 +174,29 @@ class ComparisonPredicateBase : public ColumnPredicate {
         }
     }
 
+    bool is_always_true(const std::pair<WrapperField*, WrapperField*>& statistic) const override {
+        if (statistic.first->is_null()) {
```
yiguolei, line 178: add  || statistic.second ->is null??

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -338,6 +339,27 @@ bool ColumnReader::match_condition(const AndBlockColumnPredicate* col_predicates
                                      col_predicates);
 }
 
+bool ColumnReader::prune_predicates_by_zone_map(std::vector<ColumnPredicate*>& predicates,
```
yiguolei, line 342: add a config to disable this
and the prune should only be used in query, not in compaction.

be/src/olap/rowset/segment_v2/segment.cpp
```diff
@@ -157,6 +156,19 @@ Status Segment::new_iterator(SchemaSPtr schema, const StorageReadOptions& read_o
...
+    for (auto& it : _column_readers) {
+        if (it.second->prune_predicates_by_zone_map(pruned_predicates, it.first)) {
+            pruned = true;
+        }
+    }
+
+    if (pruned) {
```
yiguolei, line 167: add a config to disable this
and the prune should only be used in query, not in compaction.

mrhhsg:
run buildall

yiguolei (approved):
LGTM

jacktengg (approved):
LGTM

PR #25436, opened 2023-10-13
[Bugfix](regression-test) Fix a naming conflict in ccr regression test

deadlinefen:
## BUG
1. Operating the same DB in the `ccr_mow_syncer_p0.test_create_table_with_binlog_config` test as in the `ccr_syncer_p0.test_create_table_with_binlog_config` test will result in the same DBs in both test to get results that are not expected when the two tests are executed at the same time
2. Created and manipulated the same username in the `ccr_mow_syncer_p0.test_get_binlog` test as in the `ccr_syncer_p0.test_get_binlog` test, which would cause the test results to not be as expected

## FIX
- Rename DB and user in `ccr_mow_syncer_p0` cases
- Replacing Global Variables with Local Variables


dataroaring (approved):
LGTM

JackDrogon (approved):
LGTM

PR #25486, opened 2023-10-16
[opt](index compaction) optimize checks before index compaction

qidaye:
## Proposed changes

If inverted index file is empty or corrupted, the index compaction will crash.
So we add more checks before apply index compaction.

### index file count tests
```sql
CREATE TABLE t1
(
    k1 int ,
    k2 string,
    k3 char(50),
    k4 varchar(200),
    k5 int,
    index index_int (k1) using inverted,
    index index_str_k2 (k2) using inverted properties("parser"="english"),
    index index_str_k4 (k4) using inverted,
    index index_k5 (k5) using inverted
)
DISTRIBUTED BY RANDOM BUCKETS 1
-- q1
insert into t1 values(1, "json love anny", "json", "anny",1);
-- q2
insert into t1 values(1, "", "", "",null);
-- q3
insert into t1 values(1, null, null, null,null);
```
#### index file list
image

1. q1
image

2. q2
image

3. q3
image

## Further comments

### DIFF
--- a/be/src/olap/compaction.cpp
+++ b/be/src/olap/compaction.cpp
@@ -556,7 +556,8 @@ Status Compaction::construct_output_rowset_writer(RowsetWriterContext& ctx, bool
                             auto index_meta =
                                     rowset->tablet_schema()->get_inverted_index(unique_id);
                             if (index_meta == nullptr) {
-                                LOG(WARNING) << "tablet[" << _tablet->tablet_id() << "] index_unique_id[" << unique_id
+                                LOG(WARNING) << "tablet[" << _tablet->tablet_id()
+                                             << "] index_unique_id[" << unique_id
                                              << "] index meta is null, will skip index compaction";
                                 return false;
                             }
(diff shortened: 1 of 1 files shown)


qidaye:
run buildall

be/src/olap/compaction.cpp
```diff
@@ -567,10 +573,48 @@ Status Compaction::construct_output_rowset_writer(RowsetWriterContext& ctx, bool
...
+                                std::string file_str = p.filename().string();
+                                lucene::store::Directory* dir =
+                                        DorisCompoundDirectory::getDirectory(fs, dir_str.c_str());
+                                auto reader = new DorisCompoundReader(dir, file_str.c_str());
+                                std::vector<std::string> files;
+                                reader->list(&files);
+
+                                // why is 3?
+                                // bkd index will write at least 3 files
+                                if (files.size() < 3) {
```
xiaokang, line 611: It seems that the only new added check in this pr is `files.size() < 3`. Is it the same case that user encountered?

airborne12: need to check the actual index file to confirm that.

qidaye: > It seems that the only new added check in this pr is `files.size() < 3`. Is it the same case that user encountered?

Also add file length check.

xiaokang (approved):
LGTM

airborne12 (approved):
LGTM

PR #24813, opened 2023-09-22
[fix](transaction) fix profile null exception cause txn stuck

yujun777:
Fix runtime profile null exception @morningman

When finishing a txn,  state transfer cb may raise an exception. It will cause the replicas and partitions' version couldn't update. Then later txn will be stucked. Fix this. 

test case will be covered by cases from Baidu, which is under dev.

## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #25575, opened 2023-10-18
[fix](transaction) fix profile null exception cause txn stuck

yujun777:
## Proposed changes

pick: #24813

When finishing a txn, state transfer cb may raise an exception. It will cause the replicas and partitions' version couldn't update. Then later txn will be stucked. Fix this.

test case will be covered by cases from Baidu, which is under dev.

## Further comments


yujun777:
run buildall

PR #25625, opened 2023-10-19
[fix](pipelineX) add double check in Coordinator

Mryange:
## Proposed changes

Because some of the properties on pipelineX are not properly configured, a nullptr exception may occur. This, in turn, will affect the call to markOneInstanceDone.

wait for 2 seconds
```java
        if (isFinished && profileDoneSignal != null) {
            try {
                profileDoneSignal.await(2, TimeUnit.SECONDS);
            } catch (InterruptedException e1) {
                LOG.warn("signal await error", e1);
            }
        }
```

In the past, it could be observed that enabling pipelineX and profiling would consistently result in a 2-second increase in time.

## Further comments


Mryange:
run buildall

PR #18986, opened 2023-04-24
[chore](recover) add a config to recover remaining data in emergency

yangzhg:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

* [ ] Does it affect the original behavior
* [ ] Has unit tests been added
* [ ] Has document been added or modified
* [ ] Does it need to update dependencies
* [ ] Is this PR support rollback (If NO, please explain WHY)

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #25654, opened 2023-10-19
[enhancement](recover) support skipping missing version in select by session variable

xy720:
## Proposed changes

Recently, in version 1.2.7, we encountered a situation where all replicas of tablets in table are having missing versions. In this case, FE is unable to recover these tablets, and the table is unable to select.

In pr #18986, I noticed that we have support config `recover_with_skip_missing_version` to ignore the visible version in FE partition.

This config only work when the replica version is behind partition version. But if the replica on be has a version missing, it actually does not skip any missing versions on be.

For example, 

1、if the replica versions in BE is {1-100, 101, 102, 103}, the partition visible version is 100, `recover_with_skip_missing_version` = `ignore_version` is working well.

2、if the replica versions in BE is {1-100, 103}, the partition visible version is 100, `recover_with_skip_missing_version` = `ignore_version` is not working.

This commit support session variable `skip_missing_version` to control the query behavior.

If `skip_missing_version` is set to true, the query will always try to select the one with the highest lastSuccessVersion among all surviving BE replicas.

If `skip_missing_version` is set to true, the query will always skip the missing rowsets in BE and only return the data from existing rowsets.

## Further comments


xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run p0

dataroaring:
We'd better use a session variable to control behavior of query. Recovery is different than query, recovery is a action without lost, however, if query ignores some missing versions on a replica and another replica has all versions, then query returns a wrong result. If query may return a wrong result, it should be controlled per query and users should know that.

dataroaring (changes requested):
We'd better use a session variable to control behavior of query. Recovery is different than query, recovery is a action without lost, however, if query ignores some missing versions on a replica and another replica has all versions, then query returns a wrong result. If query may return a wrong result, it should be controlled per query and users should know that.

xy720:
Ok, I understand. I will use a session variable to make this behavior only effect one session.

> We'd better use a session variable to control behavior of query. Recovery is different than query, recovery is a action without lost, however, if query ignores some missing versions on a replica and another replica has all versions, then query returns a wrong result. If query may return a wrong result, it should be controlled per query and users should know that.

xy720:
run buildall

xy720:
run p0

xy720:
run buildall

xy720:
run p0

xy720:
run p0

xy720:
> We'd better use a session variable to control behavior of query. Recovery is different than query, recovery is a action without lost, however, if query ignores some missing versions on a replica and another replica has all versions, then query returns a wrong result. If query may return a wrong result, it should be controlled per query and users should know that.

Hi, I already changed to use session variable to control query behavior, besides,  when the skip_missing_version variable is set to true, the query will choose the alive replica with highest last success version to select and skip missing version.

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run buildall

xy720:
run p0

xy720:
run pipelinex_p0

xy720:
run buildall

dataroaring (approved):
LGTM

xy720:
run feut

xy720:
run pipelinex_p0

lide-reed (approved):
LGTM

PR #25665, opened 2023-10-20
[fix](nereids) Set to use nereids planner in select aggregation expr without from scene when regression-test

JingDas:
…without from scene when regression-test

## Proposed changes
In old optimizer, we don't support to select aggregation function without from. So we execute
`SET enable_nereids_planner = TRUE;` before query test to avoid fail when test old optimizer.

## Further comments


JingDas:
run buildall

starocean999 (approved):

keanji-x (approved):

PR #25669, opened 2023-10-20
[FIX](resize) fix array and map offsets resize with default value

amorynan:
## Proposed changes

Issue Number: close #xxx
fix array and map offsets resize with default value
if we create_column() then resize() , offsets in array\map may has wrong default value to push

## Further comments

### DIFF
--- a/be/src/vec/columns/column_array.cpp
+++ b/be/src/vec/columns/column_array.cpp
@@ -422,7 +422,8 @@ void ColumnArray::reserve(size_t n) {
 
 //please check you real need size in data column, because it's maybe need greater size when data is string column
 void ColumnArray::resize(size_t n) {
-    get_offsets().resize(n);
+    auto last_off = get_offsets().back();
+    get_offsets().resize_fill(n, last_off);
     get_data().resize(n);
 }
 
--- a/be/src/vec/columns/column_array.cpp
+++ b/be/src/vec/columns/column_array.cpp
@@ -424,7 +424,6 @@ void ColumnArray::reserve(size_t n) {
 void ColumnArray::resize(size_t n) {
     auto last_off = get_offsets().back();
     get_offsets().resize_fill(n, last_off);
-    get_data().resize(n);
 }
 
 size_t ColumnArray::byte_size() const {
--- a/be/src/vec/columns/column_map.cpp
+++ b/be/src/vec/columns/column_map.cpp
@@ -450,7 +450,8 @@ void ColumnMap::reserve(size_t n) {
 }
 
 void ColumnMap::resize(size_t n) {
-    get_offsets().resize(n);
+    auto last_off = get_offsets().back();
+    get_offsets().resize_fill(n, last_off);
     keys_column->resize(n);
     values_column->resize(n);
 }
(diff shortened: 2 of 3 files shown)


amorynan:
run buildall

be/src/vec/columns/column_array.cpp
```diff
@@ -422,7 +422,8 @@ void ColumnArray::reserve(size_t n) {
 
 //please check you real need size in data column, because it's maybe need greater size when data is string column
 void ColumnArray::resize(size_t n) {
-    get_offsets().resize(n);
+    auto last_off = get_offsets().back();
+    get_offsets().resize_fill(n, last_off);
     get_data().resize(n);
```
eldenmoon, line 427: Is this `get_data().resize(n)` correct? for example if this array contains 3 element s`[1, 2]` `[1, 2]` `[1, 2]`, so the inner data size is 6 instead of 3. so after doing `::resize(3)`, the inner data size became 3 and lost 3 nested element.Is this semantic right?

xiaokang: I think it should not call get_data().resize(n). Just append offsets with the same value.

be/src/vec/columns/column_array.cpp
```diff
@@ -422,7 +422,8 @@ void ColumnArray::reserve(size_t n) {
 
 //please check you real need size in data column, because it's maybe need greater size when data is string column
 void ColumnArray::resize(size_t n) {
-    get_offsets().resize(n);
+    auto last_off = get_offsets().back();
+    get_offsets().resize_fill(n, last_off);
```
xiaokang, line 426: We need to check if resize_fill will modify old offsets value.

be/src/vec/columns/column_map.cpp
```diff
@@ -450,7 +450,8 @@ void ColumnMap::reserve(size_t n) {
 }
 
 void ColumnMap::resize(size_t n) {
-    get_offsets().resize(n);
+    auto last_off = get_offsets().back();
+    get_offsets().resize_fill(n, last_off);
     keys_column->resize(n);
```
xiaokang, line 455: should not call resize on keys and values column

be/src/vec/columns/column_array.cpp
```diff
@@ -422,7 +422,8 @@ void ColumnArray::reserve(size_t n) {
 
 //please check you real need size in data column, because it's maybe need greater size when data is string column
 void ColumnArray::resize(size_t n) {
```
xiaokang, line 424: Is there the same problem for ColumnStruct?

hello-stephen:
run p1

be/src/vec/columns/column_array.cpp
amorynan: struct has no offsets array so do not has this problem

be/src/vec/columns/column_array.cpp
amorynan: done!

be/src/vec/columns/column_map.cpp
amorynan: done!

amorynan:
run buildall

amorynan:
run compile

amorynan:
run beut

amorynan:
run clickbench

amorynan:
run p0

amorynan:
run p1

amorynan:
run external

airborne12 (approved):
LGTM

eldenmoon (approved):
LGTM

xiaokang (approved):
LGTM

PR #25688, opened 2023-10-20
[fix](rpc) Rebuild failed channel to avoid connection refused 

w41ter:
We cannot judge the progress of reconnecting the underlying channel of grpc.
In the worst case, it may take two minutes. But we couldn't stand the connection
refused for two minutes, so we rebuilt the channel directly.


yujun777 (approved):
LGTM

w41ter:
run buildall

dataroaring (approved):

PR #25302, opened 2023-10-11
[refactor](jni) unified jni framework for java udf

AshinGau:
## Proposed changes

Use the unified jni framework to refactor java udf.
The unified jni framework takes `VectorTable` as the container to transform data between c++ and java, and hide the details of data format conversion.
In addition, the unified framework supports complex and nested types. 
The performance of basic types remains consistent, with a 30% improvement in string types and an order of magnitude improvement in complex types.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #25758, opened 2023-10-23
[fix](java udf) make executor class thread local

AshinGau:
## Proposed changes

Java udf will be failed when enable `experimental_enable_pipeline_x_engine`, because `pipeline_x` makes `Expr` shared among fragments. However, PR(https://github.com/apache/doris/pull/25302) has deleted the global reference of executor class when calling `JniContext::close` in thread local, so another fragment can't call `JavaFunctionCall::open` in its local thread.
We have to deleted the global reference of executor class, before PR(https://github.com/apache/doris/pull/25302) exists memory leakage, and there's no need to make executor class fragment local, so we make all the context thread local.

## Further comments

### DIFF
--- a/be/src/vec/functions/function_java_udf.cpp
+++ b/be/src/vec/functions/function_java_udf.cpp
@@ -17,31 +17,16 @@
 
 #include "vec/functions/function_java_udf.h"
 
-#include <glog/logging.h>
-
-#include <cstdint>
 #include <memory>
 #include <string>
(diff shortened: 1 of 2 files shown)


AshinGau:
run buildall

Gabriel39 (approved):
LGTM

HappenLee (approved):
LGTM

PR #25798, opened 2023-10-23
[feature-wip](executor)Fe send topic info to be

wangbo:
## Proposed changes

Fe send workload group info to be periodically, then be update workload group info.
Currently only support delete unused workload group in be.


HappenLee:
run buildall

be/src/agent/topic_subscriber.cpp
```diff
@@ -29,37 +29,41 @@ namespace doris {
...
-    // Delete all listeners in the register
-    std::map<TTopicType::type, std::vector<TopicListener*>>::iterator it =
-            _registered_listeners.begin();
-    for (; it != _registered_listeners.end(); ++it) {
-        std::vector<TopicListener*>& listeners = it->second;
-        std::vector<TopicListener*>::iterator listener_it = listeners.begin();
-        for (; listener_it != listeners.end(); ++listener_it) {
-            delete *listener_it;
-        }
+    auto iter = _registered_listeners.begin();
```
HappenLee, line 32: why out the for loop ？

wangbo:
run buildall

be/src/agent/topic_subscriber.cpp
```diff
@@ -29,37 +29,41 @@ namespace doris {
...
-            delete *listener_it;
-        }
+    for (auto iter = _registered_listeners.begin(); iter != _registered_listeners.end(); iter++) {
+        delete iter->second;
     }
 }
 
-void TopicSubscriber::register_listener(TTopicType::type topic_type, TopicListener* listener) {
+void TopicSubscriber::register_listener(TTopicInfoType::type topic_type,
+                                        TopicListener* topic_listener) {
```
yiguolei, line 38: using shared ptr please， to avoid potential memory leak in the future

be/src/agent/topic_subscriber.cpp
```diff
@@ -29,37 +29,41 @@ namespace doris {
...
-    std::map<TTopicType::type, std::vector<TopicListener*>>::iterator it =
-            _registered_listeners.begin();
-    for (; it != _registered_listeners.end(); ++it) {
-        std::vector<TopicListener*>& listeners = it->second;
-        std::vector<TopicListener*>::iterator listener_it = listeners.begin();
-        for (; listener_it != listeners.end(); ++listener_it) {
-            delete *listener_it;
-        }
+    for (auto iter = _registered_listeners.begin(); iter != _registered_listeners.end(); iter++) {
+        delete iter->second;
```
yiguolei, line 33: delete these code

be/src/agent/topic_subscriber.cpp
```diff
@@ -29,37 +29,41 @@ namespace doris {
...
-    for (; topic_update_it != topic_updates.end(); ++topic_update_it) {
-        std::vector<TopicListener*>& listeners = this->_registered_listeners[topic_update_it->type];
-        std::vector<TopicListener*>::iterator listener_it = listeners.begin();
-        // Send the update to all listeners with protocol version.
-        for (; listener_it != listeners.end(); ++listener_it) {
-            (*listener_it)->handle_update(agent_publish_request.protocol_version, *topic_update_it);
+    pthread_t tids[_registered_listeners.size()];
+    int i = 0;
+    for (auto& listener_pair : _registered_listeners) {
+        ThreadArgsPair* arg_pair = new ThreadArgsPair();
```
yiguolei, line 50: using smart ptr

be/src/agent/topic_subscriber.cpp
```diff
@@ -29,37 +29,41 @@ namespace doris {
...
+        arg_pair->first = listener_pair.second;
+        arg_pair->second = new TPublishTopicRequest(topic_request);
+        int ret = pthread_create(
```
yiguolei, line 53: not use pthread directly. using doris::thread please

fe/fe-core/src/main/java/org/apache/doris/common/publish/TopicPublisherThread.java
```diff
@@ -0,0 +1,95 @@
...
+                BackendService.Client client = null;
+                try {
+                    client = ClientPool.backendPool.borrowObject(addr);
+                } catch (Exception e) {
+                    LOG.warn("Fetch a agent client failed. backend=[{}] reason=[{}]", addr, e);
+                    continue;
+                }
+                try {
+                    client.publishTopicInfo(request);
+                } catch (TException e) {
```
yiguolei, line 83: maybe we should start multi thread to do publishing for every backend?? If one backend hang then the process is hang and we could not publish to other backends?

be/src/agent/topic_subscriber.cpp
```diff
@@ -29,37 +29,41 @@ namespace doris {
...
+                &tids[i], NULL,
+                [](void* arg) {
+                    auto* local_pair = (ThreadArgsPair*)(arg);
```
yiguolei, line 56: I think maybe we need not fork a thread to deal with the logic, just in the handle_topic_info thread is ok

wangbo: We maybe add BEConfigListener later, so using thread can avoid mutual influence.

fe/fe-core/src/main/java/org/apache/doris/common/publish/TopicPublisherThread.java
wangbo: Using a thread pool can alleviate the impact of BE hang, because if a BE hangs frequently, the thread pool may be fullfilled with the hang request. Maybe need a more complex mechanism to avoid it.

wangbo:
run buildall

wangbo:
run buildall

yiguolei (approved):
LGTM

HappenLee (approved):
LGTM

PR #25965, opened 2023-10-26
[fix](udf) avoid exception when fail to find udf in replay logic

morningman:
## Proposed changes

```
ERROR (stateListener|103) [EditLog.loadJournal():984] Operation Type 131 java.lang.RuntimeException: org.apache.doris.common.UserException: errCode = 2, detailMessage = function name does not exist: ase_encryp  at org.apache.doris.catalog.Database.replayDropFunction(Database.java:727) ~[doris-fe.jar:1.0-SNAPSHOT]  at
```

When replaying `drop function` edit log, the function may not be found, causing runtime exception and
FE will fail to start.
The function SHOULD be exist, but the reason is still unknown.
I change the logic to NOT throw exception if function is not found.
This is a workaround to make sure FE can start, and add some log for later debug.

## Further comments


morningman:
run buildall

zhangstar333 (approved):

xiaokang (approved):
LGTM

PR #26080, opened 2023-10-29
[Improvement](inverted index) delay inverted index col read to reduce IO

xiaokang:
## Proposed changes

Issue Number: close #xxx

Currently, match predicate eg.`msg MATCH 'error'` will be added to common expr `_common_expr_ctxs_push_down` to downgrade to text scan if  index is missing.  It causes a side effect that index column `msg` will be read at firstRead stage, which can be delayed to lazyRead stage to reduce IO. 

Noticeable performance degree is observed in the following query, if the selectivity of match predicate `msg MATCH 'error' ` is low but the selectivity of expr predicate `abs(c_bigint)` is high.
`SELECT * FROM test WHERE msg MATCH 'error' AND abs(c_bigint) < 100000`

This PR delete match predicate from common expr `_common_expr_ctxs_push_down` if the index is applied successfully.

## Further comments


xiaokang:
run buildall

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -1109,8 +1161,9 @@ Status SegmentIterator::_init_inverted_index_iterators() {
         int32_t unique_id = _opts.tablet_schema->column(cid).unique_id();
         if (_inverted_index_iterators.count(unique_id) < 1) {
             RETURN_IF_ERROR(_segment->new_inverted_index_iterator(
-                    _opts.tablet_schema->column(cid), _opts.tablet_schema->get_inverted_index(cid),
-                    _opts, &_inverted_index_iterators[unique_id]));
```
airborne12, line 1113: so here is a bug fixed here?

xiaokang: yes, it's an old bug unrelated to this pr.

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -1040,6 +1043,55 @@ Status SegmentIterator::_apply_inverted_index() {
             }
         }
     }
+
+    // delete from _common_expr_ctxs_push_down if a MATCH predicate will be removed from _col_predicates
+    // since it's not necessary to eval it any more to avoid index miss, which is added in _normalize_predicate
+    for (auto pred : _col_predicates) {
+        if (pred->type() == PredicateType::MATCH &&
+            std::find(remaining_predicates.begin(), remaining_predicates.end(), pred) ==
```
eldenmoon, line 1051: If predicate found in `remaining_predicates` means that, it is not applied by inverted index and then continue expr to filter?

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -1040,6 +1043,55 @@ Status SegmentIterator::_apply_inverted_index() {
...
+                    remaining_predicates.end()) {
+            MatchPredicate* match_pred = dynamic_cast<MatchPredicate*>(pred);
```
airborne12, line 1053: if we confirmed that pred type is match, we can use static_cast to avoid dynamic_cast overhead.

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -1040,6 +1043,55 @@ Status SegmentIterator::_apply_inverted_index() {
...
+            for (auto it = _common_expr_ctxs_push_down.begin();
+                 it != _common_expr_ctxs_push_down.end(); it++) {
+                auto expr = (*it)->root().get();
+                // check expr type and child is the same as match predicate
+                if (expr->node_type() == TExprNodeType::MATCH_PRED &&
+                    expr->children().size() == 2 && expr->get_child(0)->is_slot_ref() &&
+                    expr->get_child(1)->is_constant()) {
+                    auto slot_ref = dynamic_cast<vectorized::VSlotRef*>(expr->get_child(0).get());
```
airborne12, line 1061: Also as above comment

be/src/olap/rowset/segment_v2/segment_iterator.cpp
xiaokang: yes

eldenmoon (approved):
LGTM

be/src/olap/rowset/segment_v2/segment_iterator.cpp
xiaokang: add check for safety at present and change it to static_cast in the future.

be/src/olap/rowset/segment_v2/segment_iterator.cpp
xiaokang: add check for safety at present and change it to static_cast in the future.

PR #26111, opened 2023-10-30
[Fix](thrift api) column should be converted if const before serialize to arrow format

eldenmoon:
## Proposed changes

```
    select null;
```
    
The above sql from thrift api will cause crash in `write_column_to_arrow` since it's column const,
we should convert to full column

Issue Number: close #xxx

## Further comments


eldenmoon:
run buildall

xiaokang (approved):
LGTM

eldenmoon:
> can you add a testcase

such case is difficult to add， and i cant find any demo

qidaye (approved):
LGTM

PR #26119, opened 2023-10-30
[Fix](thrift api) column should be converted if const before serializ…

eldenmoon:
…e to arrow format

```
select null;
```

The above sql from thrift api will cause crash in `write_column_to_arrow` since it's column const, we should convert to full column

## Proposed changes

Issue Number: close #xxx

## Further comments


xiaokang (approved):
LGTM

PR #26179, opened 2023-10-31
[fix](thrift)limit be and fe thrift server max pkg size,avoid accepti…

ryanzryu:
…ng error or too large package causing OOM

## Proposed changes
Thrift server reads the first four bytes of a packet as the packet size, and without setting the string_limit parameter, it will directly request the memory corresponding to the packet size represented by the first four bytes. When receiving an error or large packet, it is easy to encounter OOM.
image
Doris currently does not set the string_limit parameter, so when receiving a corrupted packet, it is easy to encounter OOM.

We  can send multiple HTTP packets to the Thrift port simultaneously to reproduce：
#/bin/bash
for (( i=0; i< 20 ;i++))
do
curl -X POST -d'{test}' "http://127.0.0.1:9020" &
done

This PR add string_limit parameter for be and fe thrift server

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -929,6 +929,9 @@ CONF_mBool(allow_invalid_decimalv2_literal, "false");
 // data page size for primary key index.
 CONF_Int32(primary_key_data_page_size, "32768");
 
+// the max package size be thrift server can receive,avoid accepting error or too large package causing OOM,default 20M
+CONF_Int32(be_thrift_max_pkg_bytes, "20000000");
+
 #ifdef BE_TEST
 // test s3
 CONF_String(test_s3_resource, "resource");
--- a/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
@@ -2053,5 +2053,11 @@ public class Config extends ConfigBase {
 
     @ConfField(mutable = true)
     public static boolean use_mysql_bigint_for_largeint = false;
+
+    /** 
+    * the max package size fe thrift server can receive,avoid accepting error or too large package causing OOM,default 20M
+    */
+    @ConfField
+    public static int fe_thrift_max_pkg_bytes = 20000000;
 }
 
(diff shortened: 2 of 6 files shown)


lide-reed (approved):
LGTM

ryanzryu:
您的邮件我已收到。

yiguolei:
@ryanzryu please submit the PR to master， and then pick to 1.2 and 2.0 branch. Because master branch will check regression test and check compile ....

be/src/common/config.h
```diff
@@ -929,6 +929,9 @@ CONF_mBool(allow_invalid_decimalv2_literal, "false");
 // data page size for primary key index.
 CONF_Int32(primary_key_data_page_size, "32768");
 
+// the max package size be thrift server can receive,avoid accepting error or too large package causing OOM,default 20M
+CONF_Int32(be_thrift_max_pkg_size, "20000000");
```
yiguolei, line 933: rename to be_thrift_max_pkg_bytes   so that the user will know that it is bytes or Mb

fe/fe-common/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -2053,5 +2053,11 @@ public class Config extends ConfigBase {
 
     @ConfField(mutable = true)
     public static boolean use_mysql_bigint_for_largeint = false;
+
+    /** 
+    * the max package size fe thrift server can receive,avoid accepting error or too large package causing OOM,default 20M
+    */
+    @ConfField
+    public static int fe_thrift_max_pkg_size = 20000000;
```
yiguolei, line 2061: rename to fe_thrift_max_pkg_bytes

ryanzryu:
> @ryanzryu please submit the PR to master， and then pick to 1.2 and 2.0 branch. Because master branch will check regression test and check compile ....
I have submitted the code to the master branch with this pr https://github.com/apache/doris/pull/26329

be/src/common/config.h
ryanzryu: I have renamed be_thrift_max_pkg_size to be_thrift_max_pkg_bytes

fe/fe-common/src/main/java/org/apache/doris/common/Config.java
ryanzryu: I have renamed fe_thrift_max_pkg_size to fe_thrift_max_pkg_bytes

xiaokang:
@ryanzryu can you submit a pr for branch-2.0

PR #26200, opened 2023-11-01
[Fix](profile) Reload profile content if profile is unfinished

dutyu:
## Proposed changes

Users can get the profile content of a query when the profile is not finished, so we should cache profile content only when the profile is finished.

## Further comments


dutyu:
run buildall

dutyu:
No need to reload, when update the profile, a new `org.apache.doris.common.util.ProfileManager.ProfileElement` object will be built.

PR #26199, opened 2023-11-01
[bug](shared scan) Fix use-after-free when enable pipeline shared scanning

Gabriel39:
## Proposed changes

When enable shared scan, all scanners will be created by one instance. When the main instance reach eos and quit, all states of it will be released. But other instances are still possible to get block from those scanners. So we must assure scanners will not be dependent on any states of the main instance after it quit.

core stack:
==2495324==ERROR: AddressSanitizer: heap-use-after-free on address 0x60d00023c478 at pc 0x5602f11666ec bp 0x7f294a6deed0 sp 0x7f294a6deec8
READ of size 8 at 0x60d00023c478 thread T272 (WithoutGroupTas)
    #0 0x5602f11666eb in __gnu_cxx::__normal_iterator>>::__normal_iterator(doris::SlotDescriptor* const* const&) /var/local/ldb_toolchain/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/stl_iterator.h:1008:20
    #1 0x5602f113ceb4 in std::vector>::begin() const /var/local/ldb_toolchain/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/stl_vector.h:821:16
    #2 0x56030ef15350 in doris::vectorized::ScannerContext::validate_block_schema(doris::vectorized::Block*) /home/zcp/repo_center/doris_master/doris/be/src/vec/exec/scan/scanner_context.cpp:264:21
    #3 0x56030f30113e in doris::pipeline::PipScannerContext::get_block_from_queue(doris::RuntimeState*, std::unique_ptr>*, bool*, int, bool) /home/zcp/repo_center/doris_master/doris/be/src/vec/exec/scan/pip_scanner_context.h:87:13
    #4 0x56030f03e279 in doris::vectorized::VScanNode::get_next(doris::RuntimeState*, doris::vectorized::Block*, bool*) /home/zcp/repo_center/doris_master/doris/be/src/vec/exec/scan/vscan_node.cpp:259:5
    #5 0x5602f4385d15 in doris::ExecNode::pull(doris::RuntimeState*, doris::vectorized::Block*, bool*) /home/zcp/repo_center/doris_master/doris/be/src/exec/exec_node.h:127:16
[...]

PR #26269, opened 2023-11-02
[bug](shared scan) Fix use-after-free when enable pipeline shared sca…

Gabriel39:
…nning (#26199)

When enable shared scan, all scanners will be created by one instance. When the main instance reach eos and quit, all states of it will be released. But other instances are still possible to get block from those scanners. So we must assure scanners will not be dependent on any states of the main instance after it quit.

## Proposed changes

pick #26199

## Further comments


Gabriel39:
run buildall

PR #26329, opened 2023-11-02
[fix](thrift)limit be and fe thrift server max pkg size,avoid accepting error or too large package causing OOM

ryanzryu:
…

## Proposed changes

Thrift server reads the first four bytes of a packet as the packet size, and without setting the string_limit parameter, it will directly request the memory corresponding to the packet size represented by the first four bytes. When receiving an error or large packet, it is easy to encounter OOM.
image

Doris currently does not set the string_limit parameter, so when receiving a corrupted packet, it is easy to encounter OOM.

We can send multiple HTTP packets to the Thrift port simultaneously to reproduce：
#/bin/bash
for (( i=0; i< 20 ;i++))
do
curl -X POST -d'{test}' "http://127.0.0.1:9020/" &
done

This PR add string_limit parameter for be and fe thrift server

I have submitted another identical PR to the 1.2 branch.


morningman:
run buildall

morningman (approved):
LGTM

PR #26339, opened 2023-11-02
[enhance](sql-cache) lazy create scan range locations if sql cache is missing

morningman:
## Proposed changes

Previously, if sql cache is enabled, Doris will first do the query plan, than check is cache is hit.
But query plan may be a time-consuming operation, and some of plan results are useless for sql cache.
This will result in very low QPS even if sql cache is enabled and hit.

This PR mainly changes:

1. Cache the ConsistentHash in FederationBackendPolicy
    Creation of ConsistentHash is also time-consuming because it will create 2048 virtual node by default.

2. Lazy creation scan range locations for FileScanNode
    When there are many splits in FileScanNode, the creation of scan range locations may be very slow.
    But these info is not used for sql cache.
    So we can do a lazy creation, to check the sql cache first, and if cache is missed, create the scan range locations.

In my case, a sql on hive table(which has 2000 files), the QPS with sql cache enable can increase from 100+ to 1000+

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/external/FederationBackendPolicy.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/external/FederationBackendPolicy.java
@@ -30,6 +30,9 @@ import org.apache.doris.thrift.TFileRangeDesc;
 import org.apache.doris.thrift.TScanRangeLocations;
 
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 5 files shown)


morningman:
run buildall

morningman:
run buildall

morningman:
run buildall

morningman:
run buildall

morningman:
run buildall

PR #26428, opened 2023-11-04
[fix](transaction) Fix concurrent schema change and txn cause dead lock

yujun777:
Concurrent schema change and txn may cause dead lock. An example:
1.  Txn T commit but not publish;
2.  Run schema change or rollup on T's related partition,  add alter replica R;
3.  sc/rollup add a sched txn watermark M; 
4.  Restart fe;
5.  After fe restart,  T's loadedTblIndexes will clear  because it's not save to disk;
6.  T will publish version to all tablet,  including sc/rollup's new alter replica R;
7.  Since R not contains txn data,  so the T will fail. It will then always waitting for R's data;
8.  sc/rollup wait for txn before M to finish,  only after that it will let R copy history data;
9.  Since T's not finished, so sc/rollup will always wait,  so R will nerver copy history data;
10. Txn T and sc/rollup will wait each other forever,   cause dead lock;

Fix:  because sc/rollup will ensure double write after the sched watermark M, so for finish transaction, when checking a alter replica:
1. if txn id is bigger than M,  check it just like a normal replica;
2. otherwise skip check this replica,   the BE will modify history data later.

## Proposed changes

Issue Number: close #xxx

## Further comments


yujun777:
run buildall

yujun777:
run buildall

be/src/olap/task/engine_publish_version_task.cpp
```diff
@@ -235,7 +236,19 @@ Status EnginePublishVersionTask::finish() {
             } else {
                 // check if the version exist, if not exist, then set publish failed
                 if (_error_tablet_ids->find(tablet_id) == _error_tablet_ids->end()) {
-                    if (tablet->check_version_exist(version)) {
+                    bool exist_version = tablet->check_version_exist(version);
+                    bool is_converting = false;
+                    if (!exist_version && tablet->tablet_state() == TabletState::TABLET_NOTREADY) {
+                        bool is_mow = tablet->keys_type() == KeysType::UNIQUE_KEYS &&
```
deardeng, line 242: add some log?

luwei16:
LGTM

fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
```diff
@@ -1111,8 +1114,28 @@ public void finishTransaction(long transactionId) throws UserException {
         LOG.info("finish transaction {} successfully, publish result: {}", transactionState, publishResult.name());
     }
 
-    private void checkReplicaContinuousVersionSucc(long tabletId, Replica replica, long version,
-            PublishVersionTask backendPublishTask, Set<Long> errorReplicaIds, List<Replica> tabletSuccReplicas,
+    private boolean isCheckTxnIgnoreAlterReplica(long transactionId, OlapTable table) {
```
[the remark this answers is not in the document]
yujun777: updated

be/src/olap/task/engine_publish_version_task.cpp
yujun777: remove be code

yujun777:
run buildall

yujun777:
run buildall

dataroaring (approved):
LGTM

luwei16 (approved):

PR #26278, opened 2023-11-02
[Feature](auditloader) Plugin auditloader use auth token to avoid using cleartext passwords in config

zhiqiang-hhhh:
1. Doris FE will check if stream load http request has auth token after checking password failed;
2. Plugin audit-log loader can use auth token if plugin config set use_auth_token to true

PR #26532, opened 2023-11-07
[branch-2.0](pick#26278)(auditloader) Plugin auditloader use auth token to avoid using cleartext passwords in config 

zhiqiang-hhhh:
1. Doris FE will check if stream load http request has auth token after checking password failed;
2. Plugin audit-log loader can use auth token if plugin config set use_auth_token to true
pick https://github.com/apache/doris/pull/26278


zhiqiang-hhhh:
run buildall

PR #26379, opened 2023-11-03
[opt](regression test) Add string-like column order by test

zhiqiang-hhhh:
Make sure order by string follows its ascii value

PR #26533, opened 2023-11-07
[branch-2.0](pick#26379) Add string-like column order by test

zhiqiang-hhhh:
pick https://github.com/apache/doris/pull/26379

1. Make sure order by string follows its ascii value


zhiqiang-hhhh:
run buildall

PR #24720, opened 2023-09-21
[feature](Nereids): push down topN through join

jackwener:
## Proposed changes

Push TopN through Join.

Join Type just can be left/right outer join or cross join, because data of their one child can't be filtered.

new TopN is (original limit + original offset, 0) as limit and offset.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #26634, opened 2023-11-09
[feature](Nereids): push down topN through join into Branch-2.0(#24720)

jackwener:
Push TopN through Join.

JoinType just can be left/right outer join or cross join, because data of their one child can't be filtered.

new TopN is (original limit + original offset, 0) as limit and offset.

(cherry picked from commit 3c9ff7af399eacf2ac77e7fd314ded740a385aae)


jackwener:
run buildall

PR #26424, opened 2023-11-04
[Enhancement](sql-cache) Use update time of hive to avoid cache miss through multi fe nodes.

dutyu:
## Proposed changes

Now the update time of hms table is generated by every FE node (Use `System.currentTimestamp()` separately), so the update time of a hms table may be different between FE nodes, always the same query can not hit the sql-cache if we submit it more than one times through different FE nodes. This pr mainly do following changes to avoid this problem.

- Use the `transient_lastDdlTime` instead of `System.currentTimestamp` as the `schemaUpdateTime` of hms tables
- Use the `eventTime` in hms event instead of `System.currentTimestamp` as the update time when processing hms events

### Test case

We can see that the `lastUpdateTime` (2023-10-27 17:52:10) of `call_center` is equals to the `transient_lastDdlTime` (1698400330) at hive side.

```sql
MySQL [(none)]> show proc '/catalogs/19924824/19924895';
+----------+------------------------------+----------+---------------------+--------------+-------+--------------------+--------------------------+--------------+---------------------+
| TableId  | TableName                    | IndexNum | PartitionColumnName | PartitionNum | State | Type               | LastConsistencyCheckTime | ReplicaCount | LastUpdateTime      |
+----------+------------------------------+----------+---------------------+--------------+-------+--------------------+--------------------------+--------------+---------------------+
| 19943647 | call_center                  | NULL     | NULL                | 1            | NULL  | HMS_EXTERNAL_TABLE | NULL                     | 0            | 2023-10-27 17:52:10 |
| 19943648 | call_center_test             | NULL     | NULL                | 1            | NULL  | HMS_EXTERNAL_TABLE | NULL                     | 0            | NULL                |
| 19943649 | call_center_test2            | NULL     | NULL                | 1            | NULL  | HMS_EXTERNAL_TABLE | NULL                     | 0            | NULL                |
| 19943650 | call_center_test3            | NULL     | NULL                | 1            | NULL  | HMS_EXTERNAL_TABLE | NULL                     | 0            | NULL                |
| 19943651 | test_compress                | NULL     | NULL                | 1            | NULL  | HMS_EXTERNAL_TABLE | NULL                     | 0            | NULL                |
| 19943652 | lx_dx_cdx_calendar_loan_dtl  | NULL     | NULL                | 1            | NULL  | HMS_EXTERNAL_TABLE | NULL                     | 0            | NULL                |
[...]

PR #26762, opened 2023-11-10
[Enhancement](sql-cache) Use update time of hive to avoid cache miss through multi fe nodes. (#26424)

morningman:
backport #26424


morningman:
run buildall

PR #26254, opened 2023-11-01
[fix](show_proc) fix show statistic proc dir to ensure that result only contains dbs in internal catalog

caiconghui:
## Proposed changes
```
before:
+-------+----------------------------------------+----------+--------------+----------+-----------+------------+
| DbId  | DbName                                 | TableNum | PartitionNum | IndexNum | TabletNum | ReplicaNum |
+-------+----------------------------------------+----------+--------------+----------+-----------+------------+
| 12003 | default_cluster:test_statistic_proc_db | 0        | 0            | 0        | 0         | 0          |
| 12005 | information_schema                     | 24       | 0            | 0        | 0         | 0          |
| 12006 | test_statistic_proc_db                 | 0        | 0            | 0        | 0         | 0          |
| Total | 3                                      | 24       | 0            | 0        | 0         | 0          |
+-------+----------------------------------------+----------+--------------+----------+-----------+------------+

now:
+-------+----------------------------------------+----------+--------------+----------+-----------+------------+
| DbId  | DbName                                 | TableNum | PartitionNum | IndexNum | TabletNum | ReplicaNum |
+-------+----------------------------------------+----------+--------------+----------+-----------+------------+
| 12003 | default_cluster:test_statistic_proc_db | 0        | 0            | 0        | 0         | 0          |
| Total | 1                                      | 0        | 0            | 0        | 0         | 0          |
+-------+----------------------------------------+----------+--------------+----------+-----------+------------+
```

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #26763, opened 2023-11-10
[fix](show_proc) fix show statistic proc dir to ensure that result only contains dbs in internal catalog (#26254)

morningman:
backport #26254


morningman:
run buildall

PR #26789, opened 2023-11-10
[fix](hudi) fix wrong schema when query hudi table on obs

morningman:
## Proposed changes

When query hudi table which file stored on huawei obs, the path's scheme is wrong, expected  `s3://`, but real is `obs://`.
This PR fix this issue.

## Further comments

If this not a complete fix for this issue, same issue may happen on other kind of catalog such as iceberg, paimon, etc.
I will submit another PR to fix them all.

## TODO
test case is missing, we need to find a way to cover it


morningman:
run buildall

wsjz (approved):
LGTM

xiaokang (approved):
LGTM

PR #26793, opened 2023-11-10
[fix](refresh) fix priv issue of refresh database and table operation

morningman:
## Proposed changes

The `refresh table` and `refresh database` operation are missing catalog info when checking privilege.
This PR fix it

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #26794, opened 2023-11-10
[fix](refresh) fix priv issue of refresh database and table operation (#26793)

morningman:
backport #26793


morningman:
run buildall

PR #26753, opened 2023-11-10
[fix](information_schema) fix test_query_sys_tables schema_privileges regression case. 

hubgeter:
## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #26800, opened 2023-11-11
[fix](information_schema) fix test_query_sys_tables schema_privileges regression case.

hubgeter:
backport #26753


morningman:
run buildall

PR #26801, opened 2023-11-11
[branch-2.0] fix test result

morningman:
fix output error from #26743
On master branch, the value in struct field is wrapped by quota,
but on branch 2.0, the value in struct field is NOT wrapped by quota


morningman:
run buildall

PR #26814, opened 2023-11-11
[chore](fs) Don't print the stack for file system and it's derived class

ByteYue:
## Proposed changes

Issue Number: close #xxx

It would print tons of logs when fs returns error and it's too disk-consuming. This pr tries to disable all the stackprint logic for fs class and it's derivation. 

This implementation is not the most optimal in terms of performance because the default-constructed Status carries a stack, which consumes a certain amount of memory. However, considering that error conditions are expected to be rare and this implementation requires minimal code modifications, I chose this approach. I would appreciate the reviewer's input: Should I specialize the template to false when constructing Status at each occurrence, or should I keep the current implementation?

## Further comments

### DIFF
--- a/be/src/common/status.h
+++ b/be/src/common/status.h
@@ -504,6 +504,12 @@ public:
 
     std::string msg() const { return _err_msg ? _err_msg->_msg : ""; }
 
+    void reset_stack() const {
+        if (nullptr != _err_msg) {
+            _err_msg->_msg.clear();
+        }
+    }
+
 private:
     int _code;
     struct ErrMsg {
--- a/be/src/common/status.h
+++ b/be/src/common/status.h
@@ -504,12 +504,6 @@ public:
 
     std::string msg() const { return _err_msg ? _err_msg->_msg : ""; }
 
-    void reset_stack() const {
-        if (nullptr != _err_msg) {
-            _err_msg->_msg.clear();
-        }
-    }
-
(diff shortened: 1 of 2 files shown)


ByteYue:
run buildall

be/src/common/status.h
```diff
@@ -504,6 +504,12 @@ class [[nodiscard]] Status {
 
     std::string msg() const { return _err_msg ? _err_msg->_msg : ""; }
 
+    void reset_stack() const {
```
yiguolei, line 507: do not add this method. Deal with the stack in other APIs.

be/src/io/fs/file_system.h
```diff
@@ -45,6 +45,7 @@ namespace io {
         if (!_s) {                            \
             LOG(WARNING) << _s;               \
         }                                     \
+        _s.reset_stack();                     \
```
yiguolei, line 48: If you do not want the stack any more, then you could create a new status object and not add stack strace.
For example,  Error<error_code,false>

yiguolei (dismissed):
LGTM

yiguolei:
run buildall

airborne12 (approved):
LGTM

ByteYue:
run buildall

dataroaring (approved):
LGTM

PR #26889, opened 2023-11-13
[enhancement](jdk) support doris fe running in jvm with jdk16+

caiconghui:
## Proposed changes

Issue Number: close #xxx

## Further comments

### DIFF
--- a/bin/start_fe.sh
+++ b/bin/start_fe.sh
@@ -181,7 +181,13 @@ java_version="$(
     jdk_version "${JAVA}"
 )"
 final_java_opt="${JAVA_OPTS}"
-if [[ "${java_version}" -gt 8 ]]; then
+if [[ "${java_version}" -ge 16 ]]; then
+    if [[ -z "${JAVA_OPTS_FOR_JDK_16}" ]]; then
+        echo "JAVA_OPTS_FOR_JDK_16 is not set in fe.conf" >>"${LOG_DIR}/fe.out"
+        exit 1
+    fi
+    final_java_opt="${JAVA_OPTS_FOR_JDK_16}"
+elif [[ "${java_version}" -gt 8 ]]; then
     if [[ -z "${JAVA_OPTS_FOR_JDK_9}" ]]; then
         echo "JAVA_OPTS_FOR_JDK_9 is not set in fe.conf" >>"${LOG_DIR}/fe.out"
         exit 1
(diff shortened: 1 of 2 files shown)


caiconghui:
run buildall

yiguolei (approved):
LGTM

zy-kkk (commented):
👋 Thank you for submitting this PR!

I see that you've added the `--add-opens` options to the JVM parameters to address the reflection limitations with Gson serialization in JDK 16+. This is indeed a practical solution that helps our project maintain compatibility with the latest Java versions.

However, I don’t recommend doing this for the following reasons:

1. **Security and Maintenance Considerations**: Utilizing `--add-opens` certainly resolves the reflection access issue, but it might introduce security vulnerabilities and increase future maintenance work. The access policies for these internal APIs might change with Java version updates.

2. **Long-Term Solution**: I suggest that we also consider a long-term solution. For instance, is it possible to update the Gson library or find an alternative serialization library that relies less on reflection? Or can we refactor our code to reduce dependency on internal API reflection?

3. **Testing and Documentation**: If we decide to proceed with this approach, I recommend thorough testing to ensure no new issues are introduced, and clearly documenting these changes and their rationale in our project documentation, for better understanding and future maintenance by team members.

I would love to hear your thoughts on these ideas. Looking forward to your response!

caiconghui:
@zy-kkk Thanks for you advice. it is a good idea to give a long term solution.  1. I think security vulnerabilities is not introduced by jdk16+?  my idea is first to let doris fe run in mac well. 2. I try to update gson version but no work. if we try another library to make this problem sloved we need a lots of test. (but if you have idea, I think it is the best solution)
3.jdk17 now is experimental and optional for doris fe, we still test it. so we can declare it in conf or document. it is the minimum cost for supporting fe running in jdk17 without introducing potential bugs.

as for java version updates, doris fe have already modified a lot codes to adapt to different JDK versions, I think it cannot be avoided. Doris fe has many library dependencies，once we find new problem in new jdk version，we also need fix it. there is no a one-time solution.

zy-kkk:
@caiconghui Thank you for your thoughtful response and the effort you've put into this PR. After considering the points you've raised, I agree that the current approach of using `--add-opens` Is a practical interim solution to get Doris FE running smoothly on JDK 16+ environments.

I appreciate your proactive stance on adapting to different JDK versions and acknowledge the challen Ges you've outlined in finding an immediate solution. Given this context, I am in favour of accepting your changes as they stand.

Furthermore, your PR sparks an important conversation about future-proofing our project. I believe this is an opportune moment to start planning and working towards full compatibility with JDK 17. As you rightly pointed out, JDK updates will continue, and staying ahead of these changes is crucial .

Let's use this PR as a starting point for our gradual adaptation to JDK 17. This could involve more comprehensive testing, updates in our documentation, and exploring potential Changes in our codebase to ensure compatibility and stability with JDK 17 and beyond.

I look forward to collaborating on this exciting journey towards embracing newer Java versions. Your contributions are invaluable, and I'm excited to see how we can further evolve our project toge Ther.

zy-kkk (approved):
LGTM

PR #26904, opened 2023-11-13
[fix](tls) Manually track memory in Allocator instead of mem hook and ThreadContext life cycle to manual control

xinyiZzz:
## Proposed changes

1. Manually track query/load/compaction/etc. memory in Allocator instead of mem hook.
- Can still use Mem Hook when cannot manually track memory code segments and find memory locations during debugging.
- This will cause memory tracking loss for Query, loss less than 10% compared to the past, but this is expected to be more controllable.
- Similarly, Mem Hook will no longer track unowned memory to the orphan mem tracker by default, so the total memory of all MemTrackers will be less than before. 
- Not need to get memory size from jemalloc in Mem Hook each memory alloc and free, which would lose performance in the past.
- Not require caching bthread local in pthread local for memory hook, in the past this has caused core dumps inside bthread, seems to be a bug in bthread.

2. ThreadContext life cycle to manual control
- In the past, ThreadContext was automatically created when it was used for the first time (this was usually in the Jemalloc Hook when the first malloc memory), and was automatically destroyed when the thread exited.
- Now instead of manually controlling the create and destroy of ThreadContext, it is mainly created manually when the task thread start and destroyed before the task thread end.

Run 43 clickbench query tests.
Use MemHook in the past:
image
image

Manual tracking in Allocator:
image
image

## Further comments


be/src/runtime/memory/thread_mem_tracker_mgr.h
```diff
@@ -161,9 +163,9 @@ inline void ThreadMemTrackerMgr::pop_consumer_tracker() {
     _consumer_tracker_stack.pop_back();
 }
 
-inline void ThreadMemTrackerMgr::consume(int64_t size, bool large_memory_check) {
+inline void ThreadMemTrackerMgr::consume(int64_t size, int skip_large_memory_check) {
```
yiguolei, line 166: why change the input parameter from bool to int?

xinyiZzz: Because `int` can support nesting, Maybe there might be a better way?

be/src/runtime/thread_context.cpp
```diff
@@ -23,34 +23,27 @@
...
-
-ThreadContextPtr::ThreadContextPtr() {
-    INIT_STATIC_THREAD_LOCAL(ThreadContext, _ptr);
-    init = true;
-}
-
 AttachTask::AttachTask(const std::shared_ptr<MemTrackerLimiter>& mem_tracker,
                        const TUniqueId& task_id, const TUniqueId& fragment_instance_id) {
-    SwitchBthreadLocal::switch_to_bthread_local();
+    ThreadLocalHandle::handle_thread_local();
```
yiguolei, line 28: rename  create_thread_local_if_not_exits()?

xinyiZzz:
run buildall

xinyiZzz:
run buildall

xinyiZzz:
run buildall

yiguolei (approved):
LGTM

Gabriel39 (approved):

PR #26978, opened 2023-11-14
[fix](move-memtable) wait stream close before free stream handler

kaijchen:
## Proposed changes

After brpc StreamClose is called, both side will receive `on_closed`.
We should wait for close to prevent memory use-after-free.

## Further comments

### DIFF
--- a/be/src/vec/sink/load_stream_stub.cpp
+++ b/be/src/vec/sink/load_stream_stub.cpp
@@ -108,6 +108,10 @@ LoadStreamStub::LoadStreamStub(LoadStreamStub& stub)
 LoadStreamStub::~LoadStreamStub() {
     if (_is_init.load() && !_handler.is_closed()) {
         brpc::StreamClose(_stream_id);
+        auto st = close_wait(60000); // timeout 60s
+        if (!st.ok()) {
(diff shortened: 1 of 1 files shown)


kaijchen:
run buildall

dataroaring (approved):
LGTM

be/src/vec/sink/load_stream_stub.cpp
```diff
@@ -108,6 +108,10 @@ LoadStreamStub::LoadStreamStub(LoadStreamStub& stub)
 LoadStreamStub::~LoadStreamStub() {
     if (_is_init.load() && !_handler.is_closed()) {
         brpc::StreamClose(_stream_id);
+        auto st = close_wait(60000); // timeout 60s
```
yiguolei, line 111: why do we need wait in an deconstructor method??
I think we should call wait using a explicit method call and then call dector

yiguolei: and 60000 is a magic number

kaijchen:
superseded by #27444

kaijchen:
superseded by #27791

PR #27010, opened 2023-11-15
[test](executor)Code coverage decimal multi count distinct

wangbo:
## Proposed changes

 Code coverage decimal multi count distinct


wangbo:
run buildall

PR #27050, opened 2023-11-15
[fix](tablet sched) fix delete stale remain replica

yujun777:
we have met a case  a tablet's  multiple replicas  have the same replica id,   but only one of them's  backend id is in current backend ids,  the other's backends are unavailable.  maybe they are stale backend which has be decommissioned.

An example: a tablet contains two replicas:
replica 1: (replica id = 100,  backend id = 1000) ,  
replica 2: (replica id = 100,  backend id = 2000).

But the system not contains backend 1000 now.  Then if this tablet is REDUNDANT,  it will pick replica 1 to delete because its backend is not available. Then deleteReplicaInternal will count other replica which replica id is not equals with replica 1, it got 0, then deleteReplicaInternal will fail.

## Proposed changes

Issue Number: close #xxx

## Further comments


yujun777:
run buildall

dataroaring (approved):
LGTM

deardeng (approved):
LGTM

PR #27064, opened 2023-11-15
[fix](nereids)count in correlated subquery shoud not output null value

starocean999:
consider sql: SELECT * FROM t1 WHERE t1.a <= (SELECT COUNT(t2.a) FROM t2 WHERE (t1.b = t2.b)); 
when unnest correlated subquery, we create a left join node. Assume outer query is left table and subquery is right one.
If there is no match, the row from right table is filled with nulls.
But COUNT function is always not nullable. So wrap COUNT with Nvl to ensure it's result is 0 instead of null to get the correct result

## Further comments


starocean999:
run buildall

morrySnow (approved):

keanji-x (approved):

PR #27085, opened 2023-11-16
[enhancement](backup) add property for backup to ignore table error

nextdreamblue:
## Proposed changes

Issue Number: close #xxx

in some case, user want backup data periodically, but sometimes only a little few table has error and make whole backup job error and cancel. when user know job cancel, and need retry backup all tables. 
add a property for backup to ignore some tables error, and make job can continue backup other tables.

## Further comments

### DIFF
--- a/docs/en/docs/sql-manual/sql-reference/Data-Definition-Statements/Backup-and-Restore/BACKUP.md
+++ b/docs/en/docs/sql-manual/sql-reference/Data-Definition-Statements/Backup-and-Restore/BACKUP.md
@@ -56,6 +56,7 @@ illustrate:
 - PROPERTIES currently supports the following properties:
   - "type" = "full": indicates that this is a full update (default)
   - "timeout" = "3600": The task timeout period, the default is one day. in seconds.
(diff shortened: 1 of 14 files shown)


nextdreamblue:
run buildall

nextdreamblue:
run buildall

lide-reed (approved):
LGTM

JackDrogon:
It isn't a good idea to do this. Always make sure to backup/restore once. It's difficult to deal with partial success. I suspect you may be trying to reduce the retry interval. If that's the case, you could implement a weak restriction that only one backup can be in progress for the entire library, and relax that restriction to only one for a table. This way concurrent backups of multiple tables, and whichever one fails, just retry that table.

JackDrogon (changes requested):
Not a good idea

lide-reed:
This comes from real customer needs, especially for users who rely on regular backups. Because the failure of some tables leads to the failure of the entire backup, it is often difficult for users to detect it. It is unacceptable to find that the backup failed when recovery is needed. It is not always feasible to back up multiple tables concurrently. In many cases, users cannot accept the successful backup of multiple tables at different times. The backup needs to be completed in one snapshot at once.

PR #27106, opened 2023-11-16
[fix](stream-load)select coordinator node from user's tag when exec streaming load

wangbo:
## Proposed changes

select coordinator node from user's tag when exec streaming load.
Stream load may failed when the user has no tag or its tag has no BE


wangbo:
run buildall

dataroaring (approved):
LGTM

HappenLee (approved):
LGTM

PR #27108, opened 2023-11-16
[fix](iceberg) iceberg use custom method to encode special characters in column name

AshinGau:
## Proposed changes
Fix two bugs:
1. Missing column is case sensitive, change the column name to lower case in FE for hive/iceberg/hudi
2. Iceberg use custom method to encode special characters in column name. Decode the column name to match the right column in parquet reader.

## Further comments


AshinGau:
run buildall

be/src/vec/exec/scan/vfile_scanner.cpp
```diff
@@ -969,16 +968,11 @@ Status VFileScanner::_generate_fill_columns() {
                 return Status::InternalError("failed to find default value expr for slot: {}",
                                              slot_desc->col_name());
             }
-            _missing_columns->emplace(slot_desc->col_name(), it->second);
+            _missing_col_descs.emplace(slot_desc->col_name(), it->second);
         }
     }
 
-    RETURN_IF_ERROR(_cur_reader->set_fill_columns(*_partition_columns, *_missing_columns));
-    if (_cur_reader->fill_all_columns()) {
```
morningman, line 977: Why remove this?

AshinGau: Because adding this will make the logic even more confusing. There's no need for cleaning the `_partition_columns` and `_partition_columns `.

morningman (approved):
LGTM

wuwenchi (approved):

PR #26078, opened 2023-10-29
[feature](pipelineX) Use dependency instead of block queue in the runtime filter

Mryange:
## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #27163, opened 2023-11-17
Revert "[feature](pipelineX) Use dependency instead of block queue in…

Mryange:
… the runtime filter (#26078)"

This reverts commit 9d8394898722953284ffe69e6b575bea09940815.

In PipelineX, we use Dependency to schedule tasks instead of using a block queue. 
However, for runtime filters, it is necessary to notify if there is a timeout. 
Previously, this pull request was only for test. 
Now,  both the block queue and the filter timeout queue coexist, which is not conducive to performance testing. 
This will be addressed later in conjunction with Dependency scheduling.

## Proposed changes

Issue Number: close #xxx

## Further comments


Mryange:
run buildall

Mryange:
run buildall

PR #27193, opened 2023-11-17
[test](regression) Add more alter stmt regression case (#26988)

swjtu-zhanglei:
## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #26988, opened 2023-11-14
[test](regression) Add more alter stmt regression case

swjtu-zhanglei:
## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #27216, opened 2023-11-17
Revert "[test](regression) Add more alter stmt regression case (#26988)"

xiaokang:
Reverts apache/doris#27193, since it caused test_alter_table_drop_column and test_alter_table_modify_column testcases fail.


xiaokang:
run buildall

PR #27160, opened 2023-11-17
[Branch2.0](Export) add show export regression testes

BePPPower:
## Proposed changes

Issue Number: close #xxx

backport: #27140

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #27217, opened 2023-11-17
Revert "[Branch2.0](Export) add show export regression testes"

xiaokang:
Reverts apache/doris#27160, since it caused test_show_export testcase fail.


xiaokang:
run buildall

PR #27221, opened 2023-11-17
Revert "[Bug](pipeline) try fix the exchange sink buffer result error"

yiguolei:
Reverts apache/doris#27052


yiguolei:
run buildall

PR #27229, opened 2023-11-18
[feature](executor)Make workload group property not required

wangbo:
## Proposed changes

Make workload group property not required.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/resource/workloadgroup/WorkloadGroup.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/resource/workloadgroup/WorkloadGroup.java
@@ -61,9 +61,9 @@ public class WorkloadGroup implements Writable, GsonPostProcessable {
 
     public static final String QUEUE_TIMEOUT = "queue_timeout";
 
-    private static final ImmutableSet<String> REQUIRED_PROPERTIES_NAME = new ImmutableSet.Builder<String>().add(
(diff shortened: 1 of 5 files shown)


fe/fe-core/src/main/java/org/apache/doris/resource/workloadgroup/WorkloadGroup.java
```diff
@@ -61,9 +61,9 @@ public class WorkloadGroup implements Writable, GsonPostProcessable {
 
     public static final String QUEUE_TIMEOUT = "queue_timeout";
 
-    private static final ImmutableSet<String> REQUIRED_PROPERTIES_NAME = new ImmutableSet.Builder<String>().add(
-            CPU_SHARE).add(MEMORY_LIMIT).build();
-
+    // NOTE(wb): all property is not required, some properties default value is set in be
+    // default value is as followed
+    // cpu_share=1024, memory_limit=0%(0 means not limit), enable_memory_overcommit=true
```
yiguolei, line 66: If you add comment here, so why not set the default value in FE not in BE. So that the logic in BE is very simple, it only need follow FE's value

yiguolei (dismissed):
LGTM

yiguolei:
run buildall

wangbo:
run buildall

yiguolei (approved):
LGTM

xinyiZzz (approved):
LGTM

PR #27334, opened 2023-11-21
[fix](stats) Fix thread leaks when doing checkpoint

Kikyou1997:
## Proposed changes

Issue Number: close #xxx

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/statistics/AnalysisTaskExecutor.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/statistics/AnalysisTaskExecutor.java
@@ -32,7 +32,7 @@ import java.util.concurrent.PriorityBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
-public class AnalysisTaskExecutor extends Thread {
+public class AnalysisTaskExecutor {
 
     private static final Logger LOG = LogManager.getLogger(AnalysisTaskExecutor.class);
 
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/org/apache/doris/statistics/AnalysisTaskExecutor.java
```diff
@@ -50,19 +50,12 @@ public AnalysisTaskExecutor(int simultaneouslyRunningTaskNum) {
                     TimeUnit.DAYS, new LinkedBlockingQueue<>(),
                     new BlockedPolicy("Analysis Job Executor", Integer.MAX_VALUE),
                     "Analysis Job Executor", true);
+            cancelExpiredTask();
```
Jibing-Li, line 53: This cancelExpiredTask will only executed once?

Kikyou1997: It would submit a thread to do this job

Jibing-Li: I see, cancelExpiredTask this function will create a thread.
I think we need to sleep for a while between two iterations, or this thread may use too much cpu resource.

Kikyou1997: It try to take elements from a blocking queue and then block until timeout, so won't cost too much CPU

Jibing-Li (approved):
LGTM

Kikyou1997:
run buildall

englefly (dismissed):

Kikyou1997:
run buildall

morrySnow (approved):

PR #27386, opened 2023-11-22
[improvement](routine load) fe metadata add routine load error reason

kobe6th:
## Proposed changes
Add routine load pause reason and cancel reason to fe metadata to prevent the loss of these messages after restarting the FE node. which will make auto resume routine load job faild, and users may be confused about the error reason of the routine load job.

## Further comments


liugddx:
you should add some regression test case

lide-reed (dismissed):
LGTM

PR #27559, opened 2023-11-24
[fix](statistics)Fix auto analyze bugs.

Jibing-Li:
1. Fix auto analyze doesn't filter unsupported type bug.
2. Catch throwable in auto analyze thread for each database, otherwise the thread will quit when one database failed to create jobs and all other databases will not get analyzed.
3. change FE config item full_auto_analyze_simultaneously_running_task_num to auto_analyze_simultaneously_running_task_num

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/statistics/StatisticsAutoCollector.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/statistics/StatisticsAutoCollector.java
@@ -24,6 +24,7 @@ import org.apache.doris.catalog.OlapTable;
 import org.apache.doris.catalog.TableIf;
 import org.apache.doris.catalog.external.ExternalTable;
 import org.apache.doris.common.Config;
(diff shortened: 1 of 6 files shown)


Kikyou1997 (approved):

Jibing-Li:
run buildall

fe/fe-core/src/main/java/org/apache/doris/statistics/StatisticsAutoCollector.java
```diff
@@ -92,7 +98,8 @@ public void analyzeDb(DatabaseIf<TableIf> databaseIf) {
             try {
                 createSystemAnalysisJob(analysisInfo);
             } catch (Exception e) {
```
morningman, line 100:
```suggestion
            } catch (Throwable e) {
```

morningman:
run buildall

morningman (approved):
LGTM

PR #27571, opened 2023-11-24
[nereids] fix stats error when using dateTime type filter

xzj7019:
## Proposed changes

Currently doris doesn't support datetime type filter stats estimation, but only for date type.
It will cause the filter using datetime type column with the same date and different time computing out a inaccurate selectivity and estimate a wrong row count, such as :

```
where o.book_time >= '2020-03-01 00:00:00.0' and o.book_time <= '2020-03-01 23:59:59.0';
```

This pr adds the datetime type(only support hh:mm:ss scale) filter estimation and improve the row count estimation for the above case.

## Further comments


xzj7019:
run buildall

englefly (approved):

morrySnow (approved):

PR #27634, opened 2023-11-27
[enhancement](stats) Stats auto collector execute in serial 

Kikyou1997:
## Proposed changes

Issue Number: close #xxx

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/statistics/StatisticsCollector.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/statistics/StatisticsCollector.java
@@ -33,13 +33,8 @@ public abstract class StatisticsCollector extends MasterDaemon {
 
     private static final Logger LOG = LogManager.getLogger(StatisticsCollector.class);
 
-    protected final AnalysisTaskExecutor analysisTaskExecutor;
-
-    protected int submittedThisRound = StatisticConstants.SUBMIT_JOB_LIMIT;
-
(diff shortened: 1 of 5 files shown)


fe/fe-core/src/main/java/org/apache/doris/statistics/StatisticsCollector.java
```diff
@@ -83,7 +74,14 @@ protected void createSystemAnalysisJob(AnalysisInfo jobInfo)
             analysisManager.createTableLevelTaskForExternalTable(jobInfo, analysisTasks, false);
         }
         Env.getCurrentEnv().getAnalysisManager().registerSysJob(jobInfo, analysisTasks);
-        analysisTasks.values().forEach(analysisTaskExecutor::submitTask);
+        for (BaseAnalysisTask task : analysisTasks.values()) {
+            try {
+                task.execute();
```
Jibing-Li, line 79: Not a block. But it's better to write the successfully executed tasks' result to column statistics table, instead of fail the whole job when one task failed.

Jibing-Li (approved):

fe/fe-core/src/main/java/org/apache/doris/statistics/StatisticsCollector.java
Kikyou1997: Yes, I agree with you and the logic you mentioned has been implemented in `AnalysisJob`

PR #27637, opened 2023-11-27
[fix](insert) txn insert and group commit should write \N string corr…

mymeiyi:
…ectly

## Proposed changes

For insert into values, we can write `\N` string to a column.
```
insert into t2 values(1, '"b"', 100);
insert into t2 values(2, '\N', 100);
insert into t2 values(3, '\\N', 100);

mysql> select * from t2 where name is null;
Empty set (0.03 sec)
```

But txn insert and group commit insert use stream load as the inner implementation, the `\N` means `null`, so the write result is not correct.
```
begin;
insert into t2 values(1, '"b"', 100);
insert into t2 values(2, '\N', 100);
insert into t2 values(3, '\\N', 100);
commit;

mysql> select * from t2 where name is null;
+------+------+-------+
| id   | name | score |
+------+------+-------+
|    3 | NULL |   100 |
+------+------+-------+
1 row in set (0.02 sec)
```

## Further comments


mymeiyi:
run buildall

dataroaring (dismissed):
LGTM

zzzzzzzs (approved):
LGTM

dataroaring (approved):
LGTM

PR #27645, opened 2023-11-27
[Fix](executor)Make blockscheduler first stop then delete

wangbo:
## Proposed changes
Whether ```first stop BlockScheduler then stop TaskScheduler``` or ```first stop TaskScheduler then stop BlockScheduler``` does not matter, because this just stop active thread.
But the ```stop of a scheduler```  must precede the ```resources deletion```.
Currently BlockScheduler's stop is after resource delete, this may cause NPE.
Because the PipelineTask maybe is deleted when BlockScheduler try to visit PipelineTask.

### DIFF
--- a/be/src/runtime/exec_env.h
+++ b/be/src/runtime/exec_env.h
@@ -143,10 +143,8 @@ public:
     ClientCache<FrontendServiceClient>* frontend_client_cache() { return _frontend_client_cache; }
     ClientCache<TPaloBrokerServiceClient>* broker_client_cache() { return _broker_client_cache; }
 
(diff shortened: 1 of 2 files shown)


wangbo:
run buildall

wangbo:
run buildall

yiguolei (approved):

HappenLee (approved):
LGTM

zhiqiang-hhhh:
This PR is unfinished. Lifecycle of `_global_block_scheduler` is not controlled explicitly. We need another refactor.

PR #27677, opened 2023-11-28
select coordinator node from user's tag when exec streaming load (#27106)

wangbo:
select coordinator node from user's tag when exec streaming load
cherry-pick #27106


wangbo:
run buildall

PR #27690, opened 2023-11-28
[improvement](grace stop)BE grace stop improvement

kobe6th:
## Proposed changes

1. Add new be config parameter grace_stop_limit_sec, which enables the control of the grace stop time. The reason is that we found it difficult to estimate and control the time required for graceful stop of the BE when there are many queries and import tasks running in the backend, resulting in a long time for graceful stop.
2. Fix the core issue in grace stop.

## Further comments


lide-reed (approved):
LGTM

PR #27698, opened 2023-11-28
[Fix](show-load)Show load npe(userinfo is null)

CalvinKirs:
### What happend
In some cases, the user information of the job may be lost, so this will cause NPE to occur.
**version 2.0.2**
**execute sql : show load**

```
java.lang.NullPointerException: null
    at org.apache.doris.load.loadv2.LoadJob.getShowInfo(LoadJob.java:823) ~[doris-fe.jar:1.2-SNAPSHOT]
    at org.apache.doris.load.loadv2.LoadManager.getLoadJobInfosByDb(LoadManager.java:565) ~[doris-fe.jar:1.2-SNAPSHOT]
```
### changes
Check whether the user information is null, and add peripheral exception capture and logs to avoid exceptions in a single piece of data, causing the entire show load to fail.


CalvinKirs:
run buildall

morningman (approved):
LGTM

wm1581066 (approved):
LGTM

PR #27762, opened 2023-11-29
[enhane](S3) Use pool executor for s3 operation

ByteYue:
## Proposed changes

Issue Number: close #xxx
The default executor of Aws S3 SDK is so cpu-consuming because it uses one atomic bool to do cas operation to do thread allocation and detach logic.

image

There would be a lot of CPU consumption when concurrent request reaches.

So this pr makes all the S3 operation invoked in one thread pool.

## Further comments


ByteYue:
run buildall

ByteYue:
run buildall

ByteYue:
We don't need pool executor if we had already used the sync interface.

PR #27777, opened 2023-11-29
[Fix](parquet-reader) Fix null map issue in parquet reader.

kaka11chen:
## Proposed changes

Fix null map issue in parquet reader which cause result incorrect such as `min()`, `max()`.

In order to share null map between parquet converted src column and dst column to avoid copying. It is very tricky that will call mutable function `doris_nullable_column->get_null_map_column_ptr()` which will set `_need_update_has_null = true`. Because some operations such as agg will call `has_null()` to set `_need_update_has_null = false`.

## Further comments


kaka11chen:
run buildall

hubgeter (approved):

morningman (approved):
LGTM

PR #27831, opened 2023-11-30
[Fix](exectuor)Fix Follower Fe query queue may not work when exec alter

wangbo:
## Proposed changes
When exec Alter query queue stmt in master FE, the alter stmt can be replayed in follower FE.
The follower FE replayed workload group meta and create a new QueryQueue, this means there maybe multipe QueryQueue at the same time in follower FE.
This could cause follower FE can not limit query as Workload group's property.

## Solution
Move QueryQueue from WorkloadGroup to WorkloadGroupManager.
WorkloadGroup only keeps queryQueue's property.
But QueryQueue's lifecycle is managed by WorkloadGroupManager.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/resource/workloadgroup/QueryQueue.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/resource/workloadgroup/QueryQueue.java
@@ -44,6 +44,10 @@ public class QueryQueue {
     public static final String RUNNING_QUERY_NUM = "running_query_num";
     public static final String WAITING_QUERY_NUM = "waiting_query_num";
 
+    private long wgId;
+
+    private long propVersion;
+
     int getCurrentRunningQueryNum() {
         return currentRunningQueryNum;
     }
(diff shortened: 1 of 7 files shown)


wangbo:
run buildall

fe/fe-core/src/main/java/org/apache/doris/resource/workloadgroup/QueryQueue.java
```diff
@@ -44,6 +47,12 @@ public class QueryQueue {
     public static final String RUNNING_QUERY_NUM = "running_query_num";
     public static final String WAITING_QUERY_NUM = "waiting_query_num";
 
+    private Thread updatePropThread;
```
yiguolei, line 50: 我们写一个thread
1. 遍历所有的workloadgroup
2. 检查每个workload group 里的query queue的配置和workloadgroup 元数据对比，如果不同，那么就reset queue 的设置。

wangbo:
run buildall

fe/fe-core/src/main/java/org/apache/doris/resource/workloadgroup/WorkloadGroupMgr.java
```diff
@@ -75,9 +76,65 @@ public class WorkloadGroupMgr implements Writable, GsonPostProcessable {
...
+
+    public void startUpdateThread() {
+        WorkloadGroupMgr wgMgr = this;
+        updatePropThread = new Thread(new Runnable() {
+            public void run() {
+                while (!wgMgr.stopped) {
+                    try {
+                        wgMgr.resetQueryQueueProp();
+                        Thread.sleep(Config.query_queue_update_interval_ms);
+                    } catch (Exception e) {
```
yiguolei, line 98: catch Throwable

fe/fe-core/src/main/java/org/apache/doris/resource/workloadgroup/WorkloadGroupMgr.java
```diff
@@ -75,9 +76,65 @@ public class WorkloadGroupMgr implements Writable, GsonPostProcessable {
...
     private final Map<Long, WorkloadGroup> idToWorkloadGroup = Maps.newHashMap();
     private final Map<String, WorkloadGroup> nameToWorkloadGroup = Maps.newHashMap();
+    private final Map<Long, QueryQueue> idToQueryQueue = Maps.newHashMap();
     private final ResourceProcNode procNode = new ResourceProcNode();
     private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
 
+    private Thread updatePropThread;
+    private volatile boolean stopped = false;
+
+    public void stopUpdateThread() {
```
yiguolei, line 86: stop never called？

yiguolei (approved):
LGTM

Gabriel39 (approved):

PR #27870, opened 2023-12-01
[fix](nereids)set operation's result type is wrong if decimal overflows

starocean999:
## Proposed changes
select cast(1.0 as decimal(27,9))  union all select cast(2.0 as decimal(34,4));
the old result type is decimal(43, 9) which overflows. This pr limit the result type to decimal(38, 4).
Operation | Result precision | Result scale
-- | -- | --
e1 { UNION \| EXCEPT \| INTERSECT } e2 | max(s1, s2) + max(p1 - s1, p2 - s2) | max(s1, s2)

The result precision and scale have an absolute maximum of 38. When a result precision is greater than 38, it's reduced to 38, and the corresponding scale is reduced to try to prevent truncating the integral part of a result

fix https://github.com/apache/doris/issues/27841

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #27872, opened 2023-12-01
[fix](nereids)set operation's result type is wrong if decimal overflows

starocean999:
## Proposed changes

pick from master https://github.com/apache/doris/pull/27870

because master's pr is just a temporary solution for set operation and the behavior may change very soon, so the test case is not picked.

## Further comments


starocean999:
run buildall

PR #27881, opened 2023-12-01
[fix](env) state listener avoid endless waiting

w41ter:
The follower is not set ready to read because the synchronization interval exceeds meta delay toleration seconds. But the state listener will wait until it becomes ready to read so that once the underlying BEBJE node is selected as the master, an endless waiting is reached.


w41ter:
run buildall

swjtu-zhanglei (approved):
LGTM

dataroaring (approved):
LGTM

PR #27887, opened 2023-12-01
[optimize](invert index) inverted indexes optimize the file cache

zzzxl1993:
## Proposed changes

1. Arrange idx files in the order of segment, fnm, tii, tis to ensure that data opened during the open index phase is added to the idx file cache.

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_v2/inverted_index_compound_directory.cpp
+++ b/be/src/olap/rowset/segment_v2/inverted_index_compound_directory.cpp
@@ -97,6 +97,22 @@ CL_NS(store)::Directory* DorisCompoundFileWriter::getDirectory() {
     return directory;
 }
 
+void DorisCompoundFileWriter::sort_files(std::vector<FileInfo>& file_infos) {
+    auto file_priority = [](const std::string& filename) {
+        if (filename.find("segments") != std::string::npos) return 1;
+        if (filename.find("fnm") != std::string::npos) return 2;
+        if (filename.find("tii") != std::string::npos) return 3;
+        return 4; // Other files
+    };
+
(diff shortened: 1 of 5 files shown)


be/src/olap/rowset/segment_v2/inverted_index_compound_directory.cpp
```diff
@@ -97,6 +97,36 @@ CL_NS(store)::Directory* DorisCompoundFileWriter::getDirectory() {
...
+    }
+
+    file_infos.clear();
+    for (const auto& map : file_maps) {
+        for (const auto& file : map.second) {
+            file_infos.emplace_back(file);
+        }
+    }
+}
+
```
airborne12, line 129: if we suppose to order file_infos by segments, fnm, tii, and other files by size, we could simplify the code, like following
`void DorisCompoundFileWriter::sort_files(std::vector<FileInfo>& file_infos) const {
    auto file_priority = [](const std::string& filename) {
        if (filename.find("segments") != std::string::npos) return 1;
        if (filename.find("fnm") != std::string::npos) return 2;
        if (filename.find("tii") != std::string::npos) return 3;
        return 4; // Other files
    };

    std::sort(file_infos.begin(), file_infos.end(),
        [&](const FileInfo& a, const FileInfo& b) {
            int priority_a = file_priority(a.filename);
            int priority_b = file_priority(b.filename);
            if (priority_a != priority_b) return priority_a < priority_b;
            return a.filesize < b.filesize;
        });
}
`

airborne12 (approved):
LGTM

qidaye (approved):
LGTM

PR #28030, opened 2023-12-05
[fix](split_by_string) Fix split by string core on column string

zhiqiang-hhhh:
wait for another PR to merge to 2.0 branch and then merge this PR.

Applying `split_by_string` on string column, such as `select split_by_string ("abc", string) from tbl`, will trigger LOG_FATAL on master and make branch-2.0 core dump on mysql result serde.

Since we will have an invalid block like below:
```
Int         Array
         Data  Offset
1.       a         3
2.       b
3.       c
4
```


zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

be/src/vec/functions/function_string.h
```diff
@@ -1911,6 +1927,58 @@ class FunctionSplitByString : public IFunction {
...
+                }
+            } else {
+                for (size_t str_pos = 0; str_pos <= str_ref.size;) {
+                    const size_t str_offset = str_pos;
+                    const size_t old_size = column_string_chars.size();
+                    const size_t split_part_size = split_str(str_pos, str_ref, delimiter_ref);
+                    str_pos += delimiter_ref.size;
+                    const size_t new_size = old_size + split_part_size;
+                    column_string_chars.resize(new_size);
+                    if (split_part_size > 0) {
```
BiteTheDDDDt, line 1967: what if str_ref.length>16

zhiqiang-hhhh: I think we are safe here, since ColumnString uses PaddedPODArray as its data type.
`using PaddedPODArray = PODArray<T, initial_bytes, TAllocator, 15, 16>;`  means str_ref will always safe to reading/writing 16 bytes after its end.

BiteTheDDDDt (approved):
LGTM

yiguolei (approved):
LGTM

PR #28069, opened 2023-12-06
[opt](mysql serde) Avoid core dump when converting invalid block to mysql result

zhiqiang-hhhh:
BE will core dump if result block is invalid when we doing result serialization.
An existing bug case is described in https://github.com/apache/doris/pull/28030, so we add check branch to avoid BE core dump due to out of range related problem.

### DIFF
--- a/be/src/vec/data_types/serde/data_type_array_serde.cpp
+++ b/be/src/vec/data_types/serde/data_type_array_serde.cpp
@@ -305,14 +305,8 @@ Status DataTypeArraySerDe::_write_column_to_mysql(const IColumn& column,
     auto& data = column_array.get_data();
     bool is_nested_string = data.is_column_string();
     const auto row_idx_of_col_arr = index_check_const(row_idx_of_mysql, col_const);
-
-    if (column_array.size() <= row_idx_of_col_arr) {
-        return Status::InternalError(
-                "Logical error, trying to fetch {}-th element of column array, whose size is {}",
-                row_idx_of_col_arr, column_array.size());
-    }
-
     result.open_dynamic_mode();
+
     if (0 != result.push_string("[", 1)) {
         return Status::InternalError("pack mysql buffer failed.");
     }
--- a/be/src/vec/data_types/serde/data_type_array_serde.cpp
+++ b/be/src/vec/data_types/serde/data_type_array_serde.cpp
@@ -299,18 +299,28 @@ void DataTypeArraySerDe::read_column_from_arrow(IColumn& column, const arrow::Ar
 template <bool is_binary_format>
 Status DataTypeArraySerDe::_write_column_to_mysql(const IColumn& column,
                                                   MysqlRowBuffer<is_binary_format>& result,
-                                                  int row_idx, bool col_const) const {
+                                                  int row_idx_of_mysql, bool col_const) const {
     auto& column_array = assert_cast<const ColumnArray&>(column);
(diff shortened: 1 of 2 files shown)


zhiqiang-hhhh:
run buildall

yiguolei (dismissed):
LGTM

be/src/vec/sink/vmysql_result_writer.cpp
```diff
@@ -156,16 +161,24 @@ Status VMysqlResultWriter<is_binary_format>::append_block(Block& input_block) {
...
             }
             serde->set_return_object_as_string(output_object_data());
             arguments.emplace_back(column_ptr.get(), col_const, serde);
         }
 
-        for (size_t row_idx = 0; row_idx != num_rows; ++row_idx) {
-            for (int i = 0; i < _output_vexpr_ctxs.size(); ++i) {
-                RETURN_IF_ERROR(arguments[i].serde->write_column_to_mysql(
-                        *(arguments[i].column), row_buffer, row_idx, arguments[i].is_const));
+        for (size_t row_idx = 0; row_idx < num_rows; ++row_idx) {
```
BiteTheDDDDt, line 170: Don’t check every row

zhiqiang-hhhh: My fault.
Check loop is divided from for loop and placed in the front.

zhiqiang-hhhh:
run buildall

BiteTheDDDDt (dismissed):

zhiqiang-hhhh:
run buildall

be/src/vec/data_types/serde/data_type_array_serde.cpp
```diff
@@ -299,18 +299,28 @@ void DataTypeArraySerDe::read_column_from_arrow(IColumn& column, const arrow::Ar
...
-                                                  int row_idx, bool col_const) const {
+                                                  int row_idx_of_mysql, bool col_const) const {
     auto& column_array = assert_cast<const ColumnArray&>(column);
     auto& offsets = column_array.get_offsets();
     auto& data = column_array.get_data();
     bool is_nested_string = data.is_column_string();
-    const auto col_index = index_check_const(row_idx, col_const);
+    const auto row_idx_of_col_arr = index_check_const(row_idx_of_mysql, col_const);
+
+    if (column_array.size() <= row_idx_of_col_arr) {
```
amorynan, line 309: I do not think here should add this check, if here add , string serde should add same check too.

zhiqiang-hhhh: Yes, we should add check for all serde if we can.

amorynan: if u add array , do same with map too.

zhiqiang-hhhh:
run buildall

be/src/vec/data_types/serde/data_type_array_serde.cpp
zhiqiang-hhhh: check is removed since it is checked per row.
this pr just fixed mysql result writer, and we have other writer types, maybe they are subjected to the same problem

yiguolei (approved):
LGTM

wangbo (approved):
LGTM

wm1581066 (approved):
LGTM

PR #28070, opened 2023-12-06
[FIX](bool) fix bool literal insert into different from streamload

amorynan:
## Proposed changes

Issue Number: close #xxx
when we insert into table with bool and stream load for bool value with numbers 
insert into will select out as 1 , but stream load select out as NULL
this pr just make it same behavior

## Further comments


dataroaring (approved):
LGTM

PR #25331, opened 2023-10-11
[improvement](resource-tag) limit the default user's resource tag to 'default'

morningman:
## Proposed changes

In previous, if user property `'resource_tags.location'` is not set, the can use Backends with any resource tag.
It may confuse that when the DBA set part of Backends to resource group A, then the current existing user
should not be able to use this group A util it's `'resource_tags.location'` is set.

So in this PR, I change the behavior, that if user property `'resource_tags.location'` is not set, it can only use the
Backends with `default` tag.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #28088, opened 2023-12-06
[opt](resource-tag) root and admin user can use any resource tag by default

morningman:
## Proposed changes

In #25331, I change the behavior of user's default resource tag, that is, if a user does not set resource tag,
it can only use default resource tag.
This PR change this logic. The normal user can only use default resource tag if resource tag is not set,
but root and admin user can use any resource tag if resource tag is not set.

## Further comments


morningman:
run buildall

zddr (approved):

xiaokang (approved):
LGTM

PR #28141, opened 2023-12-07
[enhance](session)check invalid value when set parallel instance variables

zhangstar333:
in some case, if set incorrectly, will be cause BE core dump
```
10:18:19   *** SIGFPE integer divide by zero (@0x564853c204c8) received by PID 2132555 
    int max_scanners =
            config::doris_scanner_thread_pool_thread_num / state->query_parallel_instance_num();
```

## Proposed changes

Issue Number: close #xxx

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/SessionVariable.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/SessionVariable.java
@@ -679,10 +679,12 @@ public class SessionVariable implements Serializable, Writable {
      * the parallel exec instance num for one Fragment in one BE
      * 1 means disable this feature
      */
-    @VariableMgr.VarAttr(name = PARALLEL_FRAGMENT_EXEC_INSTANCE_NUM, needForward = true, fuzzy = true)
+    @VariableMgr.VarAttr(name = PARALLEL_FRAGMENT_EXEC_INSTANCE_NUM, needForward = true, fuzzy = true,
+                        setter = "checkFragmentInstanceNum")
     public int parallelExecInstanceNum = 1;
 
-    @VariableMgr.VarAttr(name = PARALLEL_PIPELINE_TASK_NUM, fuzzy = true, needForward = true)
+    @VariableMgr.VarAttr(name = PARALLEL_PIPELINE_TASK_NUM, fuzzy = true, needForward = true,
+                        setter = "checkPipelineTaskNum")
     public int parallelPipelineTaskNum = 0;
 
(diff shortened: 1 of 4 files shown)


fe/fe-core/src/main/java/org/apache/doris/qe/VariableMgr.java
```diff
@@ -150,9 +150,24 @@ public static SessionVariable getDefaultSessionVariable() {
...
+            throw new DdlException(
+                    variableName + " value should greater than or equal " + String.valueOf(minValue)
+                            + ", you set value is: " + value);
+        }
+    }
+
     // Set value to a variable
     private static boolean setValue(Object obj, Field field, String value) throws DdlException {
         VarAttr attr = field.getAnnotation(VarAttr.class);
+        if ("parallel_pipeline_task_num".equalsIgnoreCase(attr.name())) {
```
yiguolei, line 165: It is not a good way to do this.
you could refer this code:
    // The global max_execution_time value provides the default for the session value for new connections.
    // The session value applies to SELECT executions executed within the session that include
    // no MAX_EXECUTION_TIME(N) optimizer hint or for which N is 0.
    // https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html
    // So that it is == query timeout in doris
    @VariableMgr.VarAttr(name = MAX_EXECUTION_TIME, fuzzy = true, setter = "setMaxExecutionTimeMS")
    public int maxExecutionTimeMS = -1;

every variable could have a specific setter attribute, then you could add check in specific setter method.

fe/fe-core/src/main/java/org/apache/doris/qe/SessionVariable.java
```diff
@@ -1842,6 +1844,23 @@ public void setMaxExecutionTimeMS(String maxExecutionTimeMS) {
         this.queryTimeoutS = this.maxExecutionTimeMS / 1000;
     }
 
+    public void checkPipelineTaskNum(String value) throws Exception {
+        checkFieldValue("parallel_pipeline_task_num", 0, value);
```
yiguolei, line 1848: It is a setter method.
You should do :
1. check if the value is valid
2. parse and set the value

zhangstar333:
run buildall

zhangstar333:
run buildall

HappenLee (dismissed):
LGTM

yiguolei (dismissed):
LGTM

zhangstar333:
run buildall

yiguolei (approved):
LGTM

yiguolei (approved):
LGTM

jacktengg (approved):

PR #28200, opened 2023-12-09
[fix](config) Enable libdivide sse2 for int128

julic20s:
libdivide use macro to enable simd registers. sse2 will speed up for int128.


julic20s:
run buildall

yiguolei:
Could you please give some performance numbers?
For example, you could insert into table a select number from numbers(100000000); and do some performance test.
@julic20s

julic20s:
@yiguolei My mistake. Add the macro is meaningless now. When the number to be divided is Int128, the libdivide will not be used. (libdivide supports for (i128 / x) but (x / int128) is not supported. Maybe `DivideIntegralImpl` should check it more carefully?)

PR #28223, opened 2023-12-11
[opt](vfilescanner) interrupt running parquet/orc readers when scannode is finished

AshinGau:
## Proposed changes

`VScanNode::get_next` will check whether the `ScanNode` has reached `limit` condition, and send `eos` to `TaskScheduler`, and `TaskScheduler` will try to close `ScanNode`.
However, `ScanNode` must wait all running scanners finished, so even if `ScanNode` has reached `limit` condition, it can't be closed immediately.
This PR try to interrupt the running readers, and make `ScanNode` to end as soon as possible.

## Further comments

### DIFF
--- a/be/src/vec/exec/scan/scanner_context.h
+++ b/be/src/vec/exec/scan/scanner_context.h
@@ -272,6 +272,8 @@ protected:
     // Not need to protect by lock, because only one scheduler thread will access to it.
     std::mutex _scanners_lock;
     std::list<VScannerSPtr> _scanners;
+    // weak pointer for _scanners, used in stop function
+    std::vector<VScannerWPtr> _scanners_ref;
     std::vector<int64_t> _finished_scanner_runtime;
     std::vector<int64_t> _finished_scanner_rows_read;
     std::vector<int64_t> _finished_scanner_wait_worker_time;
--- a/be/src/vec/exec/scan/vfile_scanner.cpp
+++ b/be/src/vec/exec/scan/vfile_scanner.cpp
@@ -718,7 +718,7 @@ Status VFileScanner::_get_next_reader() {
         }
         _cur_reader.reset(nullptr);
         _src_block_init = false;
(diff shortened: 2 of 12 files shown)


be/src/vec/exec/scan/vfile_scanner.cpp
```diff
@@ -1110,4 +1110,11 @@ Status VFileScanner::close(RuntimeState* state) {
     return Status::OK();
 }
 
+void VFileScanner::try_stop() {
+    VScanner::try_stop();
+    if (_io_ctx) {
+        _io_ctx->should_stop = true;
```
morningman, line 1116: Need `should_stop` be atomic?

AshinGau:
run buildall

be/src/vec/exec/scan/vfile_scanner.cpp
AshinGau: `should_stop` only needs the final consistency.

morningman (approved):
LGTM

be/src/vec/exec/scan/scanner_context.h
```diff
@@ -272,6 +272,8 @@ class ScannerContext {
     // Not need to protect by lock, because only one scheduler thread will access to it.
     std::mutex _scanners_lock;
     std::list<VScannerSPtr> _scanners;
+    // weak pointer for _scanners, used in stop function
+    std::vector<VScannerWPtr> _scanners_ref;
```
yiguolei, line 276: There is already a     std::list<VScannerSPtr> _scanners;
why do we need a weak ptr here?

AshinGau: Because `_scanners` will be popped/pushed/released when running, and `try_close` has no need for synchronization and consistency, only do the best effort, so there's no need to use `_scanners` with some lock operations.

yiguolei (approved):
LGTM

PR #28302, opened 2023-12-12
[Improve][COLOCATE] improve colocate balance and colocate join

qzsee:
## Proposed changes

Issue Number: close #xxx

The current colocate group has the following problems:

1. If there is an unrecoverable be or decommission some be, or a tablet is faulty and the tablet is slowly repaired, if the group is very large, the whole group is in a unstable state for a long time, and the colocate join is unavailable for a long time. One obvious problem here is that the control granularity of colocate balance is too coarse, and it is not reasonable to mark the group unstable once a tablet in the whole group is unavailable.

2. colocate balance generates a large number of replica repair tasks, which affect other normal repair tasks

Based on the above problems, the following optimization is done:
1. If any be is unavailable, immediately replace all the be nodes in the unavailable buckets. If we decommission some be nodes, then we replace them one by one. Then, when we query, we take the intersection of query locations, and try not to degrade the performance of join.
2. Perform traffic limiting for the colocate tablet repair

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/Tablet.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/Tablet.java
@@ -691,6 +691,23 @@ public class Tablet extends MetaObject implements Writable {
         return TabletStatus.HEALTHY;
     }
 
+    public Set<Long> getUnavailableBeIdsForColocate(Set<Long> backendsSet, long visibleVersion) {
+        Set<Long> replicaBackendIds = getBackendIds();
+        Set<Long> unavailableBeIds = Sets.newHashSet(Sets.difference(backendsSet, replicaBackendIds));
+        for (Replica replica : replicas) {
+            long backendId = replica.getBackendId();
+            if (backendsSet.contains(backendId)) {
+                if (!replica.isAlive()) {
+                    unavailableBeIds.add(backendId);
+                }
+                if (replica.getLastFailedVersion() > 0 || replica.getVersion() < visibleVersion) {
+                    unavailableBeIds.add(backendId);
+                }
+            }
+        }
+        return unavailableBeIds;
+    }
+
     /**
      * check if this tablet is ready to be repaired, based on priority.
      * VERY_HIGH: repair immediately
--- a/fe/fe-core/src/main/java/org/apache/doris/clone/ColocateTableCheckerAndBalancer.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/clone/ColocateTableCheckerAndBalancer.java
@@ -34,6 +34,7 @@ import org.apache.doris.clone.TabletSchedCtx.Priority;
 import org.apache.doris.clone.TabletScheduler.AddResult;
 import org.apache.doris.common.Config;
 import org.apache.doris.common.DdlException;
+import org.apache.doris.common.Pair;
 import org.apache.doris.common.UserException;
 import org.apache.doris.common.util.MasterDaemon;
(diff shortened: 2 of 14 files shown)


fe/fe-core/src/main/java/org/apache/doris/catalog/Tablet.java
```diff
@@ -691,6 +691,23 @@ public TabletStatus getColocateHealthStatus(long visibleVersion,
...
+    public Set<Long> getUnavailableBeIdsForColocate(Set<Long> backendsSet, long visibleVersion) {
+        Set<Long> replicaBackendIds = getBackendIds();
+        Set<Long> unavailableBeIds = Sets.newHashSet(Sets.difference(backendsSet, replicaBackendIds));
+        for (Replica replica : replicas) {
+            long backendId = replica.getBackendId();
+            if (backendsSet.contains(backendId)) {
+                if (!replica.isAlive()) {
+                    unavailableBeIds.add(backendId);
+                }
+                if (replica.getLastFailedVersion() > 0 || replica.getVersion() < visibleVersion) {
```
yujun777, line 703: replica missing version is very common,  it can clone version from other BE.
do not check their version.

fe/fe-core/src/main/java/org/apache/doris/clone/ColocateTableCheckerAndBalancer.java
```diff
@@ -437,25 +454,35 @@ private Set<GroupId> relocateAndBalanceGroup(Set<GroupId> groupIds, boolean bala
                 Set<Long> beIdsInOtherTag = colocateIndex.getBackendIdsExceptForTag(groupId, tag);
                 List<Long> availableBeIds = getAvailableBeIds(SystemInfoService.DEFAULT_CLUSTER, tag, beIdsInOtherTag,
                         infoService);
+
+                Set<Long> nonAliveBeIds = getNonAliveBeIds(unavailableBeIdsInGroup, infoService);
+                if ((CollectionUtils.isEmpty(unavailableBeIdsInGroup) || nonAliveBeIds.isEmpty())
```
yujun777, line 459: if a be is decommission, should move its tablets to other BE as soon as possible.

fe/fe-core/src/main/java/org/apache/doris/clone/ColocateTableCheckerAndBalancer.java
```diff
@@ -931,6 +1002,54 @@ private boolean relocateAndBalance(GroupId groupId, Tag tag, Set<Long> unavailab
         return isChanged;
     }
 
+    private boolean checkGroupStable(GroupId groupId, int idx, Set<Long> unAvailableBeIds, Map<Integer, Map<Tag,
```
yujun777, line 1005: pls explain this function

fe/fe-core/src/main/java/org/apache/doris/clone/ColocateTableCheckerAndBalancer.java
```diff
@@ -738,10 +819,6 @@ private boolean relocateAndBalance(GroupId groupId, Tag tag, Set<Long> unavailab
             // update backends and hosts at each round
             backendsPerBucketSeq = Lists.partition(flatBackendsPerBucketSeq, replicaNum);
             List<List<String>> hostsPerBucketSeq = getHostsPerBucketSeq(backendsPerBucketSeq, infoService);
-            if (hostsPerBucketSeq == null) {
```
yujun777, line 741: why remove

yujun777 (commented):
need more reviewed

PR #28304, opened 2023-12-12
[fix](statistics)Fix column stats trigger info bug.

Jibing-Li:
Before, we didn't update the jobType info in ColStatsMeta. This will case the jobType always be the type when it first be set. For example, if we manually analyzed table, the jobType will always be MANUAL, even if this table is auto analyzed again later.

bp https://github.com/apache/doris/pull/28303

## Further comments


Jibing-Li:
run buildall

PR #28325, opened 2023-12-13
[fix](user) Avoid throw unknown UserProperty after FE rollback version

xinyiZzz:
## Proposed changes

After using SET PROPERTY to modify the user property, if FE rolls back to a version without this property, `Unknown user property` error will be reported when replay EditLog, just ignore it.
img_v3_0262_90aa299f-d6d7-481f-9f17-f0012e55d91g

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/Auth.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/Auth.java
@@ -960,6 +960,12 @@ public class Auth implements Writable {
                 Env.getCurrentEnv().getEditLog().logUpdateUserProperty(propertyInfo);
             }
             LOG.info("finished to set properties for user: {}", user);
+        } catch (DdlException e) {
+            if (isReplay && e.getMessage().contains("Unknown user property")) {
+                LOG.warn("ReplayUpdateUserProperty failed, maybe FE rolled back version, " + e.getMessage());
+            } else {
+                throw e;
+            }
         } finally {
             writeUnlock();
         }
--- a/fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/Auth.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/Auth.java
@@ -954,18 +954,12 @@ public class Auth implements Writable {
             throws UserException {
         writeLock();
         try {
-            propertyMgr.updateUserProperty(user, properties);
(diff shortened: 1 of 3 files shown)


xinyiZzz:
run buildall

fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/Auth.java
```diff
@@ -960,6 +960,12 @@ public void updateUserPropertyInternal(String user, List<Pair<String, String>> p
                 Env.getCurrentEnv().getEditLog().logUpdateUserProperty(propertyInfo);
             }
             LOG.info("finished to set properties for user: {}", user);
+        } catch (DdlException e) {
+            if (isReplay && e.getMessage().contains("Unknown user property")) {
+                LOG.warn("ReplayUpdateUserProperty failed, maybe FE rolled back version, " + e.getMessage());
```
morningman, line 965:
```suggestion
                LOG.warn("ReplayUpdateUserProperty failed, maybe FE rolled back version. Ignore it" + e.getMessage());
```

fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/Auth.java
```diff
@@ -960,6 +960,12 @@ public void updateUserPropertyInternal(String user, List<Pair<String, String>> p
                 Env.getCurrentEnv().getEditLog().logUpdateUserProperty(propertyInfo);
             }
             LOG.info("finished to set properties for user: {}", user);
+        } catch (DdlException e) {
+            if (isReplay && e.getMessage().contains("Unknown user property")) {
```
morningman, line 964: How about define a new Exception, not using `error message` to check?
And some some comment about how can it happen and why it can be ignored.

fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/Auth.java
xinyiZzz: done

fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/Auth.java
xinyiZzz: done, thx

yiguolei:
run buildall

yiguolei (approved):
LGTM

morningman (approved):
LGTM

PR #19126, opened 2023-04-26
[fix](memory) Fix AggFunc memory leak due to incorrect destroy

xinyiZzz:
# Proposed changes

Issue Number: close #xxx

## Problem summary

Asan memory leak:
```
  #0 0x5598a4363ded in operator new(unsigned long) (/mnt/hdd01/dorisTestEnv/VEC_ASAN/be/lib/doris_be+0x14d0bded) (BuildId: 0e35cb6207c3ab7c)
    #1 0x5598ae1606a1 in COW<doris::vectorized::IColumn>::mutable_ptr<doris::vectorized::ColumnVector<long>> COWHelper<doris::vectorized::ColumnVectorHelper,
doris::vectorized::ColumnVector<long>>::create<>() /doris/be/src/vec/common/cow.h:412:27
    #2 0x5598ae160268 in doris::vectorized::ColumnVector<long>::clone_resized(unsigned long) const /doris/be/src/
vec/columns/column_vector.cpp:319:16
    #3 0x5598a477c0f9 in doris::vectorized::IColumn::clone_empty() const /doris/be/src/vec/columns/column.h:117:5
3
    #4 0x5598ae632232 in doris::vectorized::ColumnWithTypeAndName::clone_empty() const /doris/be/src/vec/core/col
umn_with_type_and_name.cpp:34:30
    #5 0x5598ae5e1d41 in doris::vectorized::Block::clone_empty() const /doris/be/src/vec/core/block.cpp:508:25
    #6 0x5598b32ac19e in doris::vectorized::AggregateFunctionSortData::AggregateFunctionSortData(std::vector<doris::vectorized::SortColumnDescription, std::al
locator<doris::vectorized::SortColumnDescription>>, doris::vectorized::Block const&) /doris/be/src/vec/aggregate_
functions/aggregate_function_sort.h:39:60
    #7 0x5598b32a895f in doris::vectorized::AggregateFunctionSort<doris::vectorized::AggregateFunctionSortData>::create(char*) const /mnt/hdd01/repo_center/do
ris_branch-2.0-alpha/doris/be/src/vec/aggregate_functions/aggregate_function_sort.h:160:21
    #8 0x5598b32a9d3f in doris::vectorized::IAggregateFunctionHelper<doris::vectorized::AggregateFunctionSort<doris::vectorized::AggregateFunctionSortData>>::
deserialize_vec(char*, doris::vectorized::ColumnString const*, doris::vectorized::Arena*, unsigned long) const /d
[...]

PR #28485, opened 2023-12-15
[branch-1.2](memory) Fix AggFunc memory leak due to incorrect destroy

xy720:
## Proposed changes

cherry-pick #19126  into branch-1.2

## Further comments


yiguolei:
1.2's code is not exception safe.... It is hard to deal with memory allocate failed.
I think you'd better maintain this code in a sepearate branch. It is very dangerous

xy720:
> 1.2's code is not exception safe.... It is hard to deal with memory allocate failed.
> I think you'd better maintain this code in a sepearate branch. It is very dangerous

This commit did not add or remove or change any exceptions, it only added additional destroy operation in the catch

xy720:
Not merge until we confirm that it will not cause new problem.

PR #18880, opened 2023-04-21
[Enchancement](statistics) Support sampling collection of statistics

weizhengte:
# Proposed changes

1. Supports sampling to collect statistics
2. Improved syntax for collecting statistics
3. Support histogram specifies the number of buckets
4. Tweaked some code structure

---

The syntax supports WITH and PROPERTIES, using the same syntax as before.

Column Statistics Collection Syntax:
```SQL
ANALYZE [ SYNC ] TABLE table_name
     [ (column_name [, ...]) ]
     [ [WITH SYNC] | [WITH INCREMENTAL] | [WITH SAMPLE PERCENT | ROWS ] ]
     [ PROPERTIES ('key' = 'value', ...) ];
```

Column histogram collection syntax:
```SQL
ANALYZE [ SYNC ] TABLE table_name
     [ (column_name [, ...]) ]
     UPDATE HISTOGRAM
     [ [ WITH SYNC ][ WITH INCREMENTAL ][ WITH SAMPLE PERCENT | ROWS ][ WITH BUCKETS ] ]
     [ PROPERTIES ('key' = 'value', ...) ];
```

Illustrate：
- sync：Collect statistics synchronously. Return after collecting.
- incremental：Collect statistics incrementally. Incremental collection of histogram statistics is not supported.
- sample percent | rows：Collect statistics by sampling. Scale and number of rows can be sampled.
- buckets：Specifies the maximum number of buckets generated when collecting histogram statistics.
- table_name: The purpose table for collecting statistics. Can be of the form `db_name.table_name`.
- column_name: The specified destination column must be a column that exists in `table_name`, and multiple column names are separated by commas.
- properties：Properties used to set statistics tasks. Currently only the following configurations are supported (equivalent to the with statement)
   - 'sync' = 'true'
   - 'incremental' = 'true'
   - 'sample.percent' = '50'
   - 'sample.rows' = '1000'
   - 'num.buckets' = 10

--- 

TODO: 
- Supplement the complete p0 test
- `Incremental` statistics see pr18653

Issue Number: close #xxx

## Problem summary

Describe your changes.

## Checklist(Required)

* [ ] Does it affect the original behavior
* [ ] Has unit tests been added
* [ ] Has document been added or modified
* [ ] Does it need to update dependencies
* [ ] Is this PR support rollback (If NO, please explain WHY)

## Further comments

[...]

PR #28516, opened 2023-12-16
[fix](doc) Updated Analyze syntax as per code

Nitin-Kashyap:
## Proposed changes

Updated the Anlyze page syntax as per code

changes refered from :#18880   #20264 #23978


yiguolei:
run buildall

Nitin-Kashyap:
run buildall

Jibing-Li (commented):
These functions are either not supported yet or deprecated, so we don't need to update the doc right now.

Jibing-Li (changes requested):
These functions are either not supported yet or deprecated, so we don't need to update the doc right now.

Nitin-Kashyap:
> These functions are either not supported yet or deprecated, so we don't need to update the doc right now.

agreed, closing this PR

PR #24853, opened 2023-09-25
[Improvement](statistics)Collect stats for hive partition column using metadata.

Jibing-Li:
Hive partition columns' stats could be calculated from hive metastore data. Doesn't need to execute sql to get the stats.
This PR is using hive partition metadata to collect partition column stats.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #28525, opened 2023-12-16
[feat](stats) Reuse HMS statistics analyzed by Spark engine for Analyze

Nitin-Kashyap:
## Proposed changes

Taking the Idea further from PR #24853 (https://github.com/apache/doris/pull/24853)
Column statistics already analyzed and available in HMS from spark, this PR proposes to reuse the analyzed stats from external source, when executed `WITH SQL` clause of analyze cooamd.

Spark analyzes and stores the statistics in Table properties instead of HiveColumnStatistics. In this PR, we try to get the statistics from these properties and make it available to Doris.

## Further comments
Histogram is also available from spark analyze: it can be captured for Doris usage.
- As of now doris doesnot capture Histogram for HMS External table.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/external/HMSExternalTable.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/external/HMSExternalTable.java
@@ -34,6 +34,7 @@ import org.apache.doris.statistics.BaseAnalysisTask;
 import org.apache.doris.statistics.ColumnStatistic;
 import org.apache.doris.statistics.ColumnStatisticBuilder;
 import org.apache.doris.statistics.HMSAnalysisTask;
+import org.apache.doris.statistics.StatsType;
 import org.apache.doris.statistics.TableStatsMeta;
 import org.apache.doris.statistics.util.StatisticsUtil;
 import org.apache.doris.thrift.THiveTable;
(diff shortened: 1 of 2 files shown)


Nitin-Kashyap:
run buildall

Jibing-Li (approved):
LGTM

fe/fe-core/src/main/java/org/apache/doris/catalog/external/HMSExternalTable.java
```diff
@@ -531,6 +553,31 @@ private void initPartitionColumns(List<Column> schema) {
...
+        }
+
+        Map<String, String> parameters = remoteTable.getParameters();
+        for (StatsType type : statsTypes.keySet()) {
+            String key = SPARK_COL_STATS + colName + MAP_SPARK_STATS_TO_DORIS.getOrDefault(type, "-");
+            if (parameters.containsKey(key)) {
+                stats.put(statsTypes.get(type), parameters.get(key));
+            } else {
+                // should not happen, spark would have all type (except histogram)
+                return false;
```
morningman, line 575: How about put a default value(eg, `-`) in `stats` instead of `return false`?
In case that some properties are missing accidentally.

Nitin-Kashyap:
run buildall

morningman (approved):
LGTM

PR #28561, opened 2023-12-18
[Bug](cooldown) Fix problem that followers may never completely cooldown

platoneko:
## Proposed changes

If an empty rowset be cooldowned in the original code , it will not write the cooldown meta to S3, which can lead to the following scenario:
follower replica version: [0-10][11-12]
cooldown replica version: [0-10][11][12]
Assuming [12] is an empty rowset, the cooldown replica has already cooled down to version 12, but the cooldown version in the S3 meta is 11. As a result, the follower cannot synchronize due to the lack of version [12]. Only after the cooldown replica performs a ColdDataCompaction to ensure that the data size of [x-12] is > 0, then the follower will be able to synchronize successfully.

## Further comments

### DIFF
--- a/be/src/olap/tablet.cpp
+++ b/be/src/olap/tablet.cpp
@@ -2122,16 +2122,7 @@ Status Tablet::_cooldown_data() {
         LOG(INFO) << "cannot pick cooldown rowset in tablet " << tablet_id();
         return Status::OK();
     }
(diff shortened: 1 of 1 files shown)


dataroaring (approved):
LGTM

xiaokang (approved):
LGTM

PR #28635, opened 2023-12-19
[fix](statistics)Fix drop stats fail silently bug.

Jibing-Li:
Drop stats use IN predicate to filter the column stats to delete. The default length of IN predicate is 1024, drop table stats with more than 1024 columns may fail.
This pr is to split the delete sql based on the IN predicate length.
 

## Further comments


Jibing-Li:
run buildall

morrySnow (approved):

fe/fe-core/src/main/java/org/apache/doris/statistics/StatisticsRepository.java
```diff
@@ -192,10 +193,34 @@ public static void dropStatistics(long tblId, Set<String> colNames) throws DdlEx
...
+        Iterator<String> iterator = colNames.iterator();
+        int columnCount = 0;
+        StringBuilder inPredicate = new StringBuilder();
+        while (iterator.hasNext()) {
+            inPredicate.append("'");
+            inPredicate.append(iterator.next());
+            inPredicate.append("'");
+            inPredicate.append(",");
+            columnCount++;
+            if (columnCount == Config.max_allowed_in_element_num_of_delete) {
```
morningman, line 206: if `max_allowed_in_element_num_of_delete` is set to `0`, the check will never succeed.
You can modify it in next PR

morningman (approved):
LGTM

PR #28669, opened 2023-12-19
[fix](pipeline) Fix query is not cancelled when be report error status

zhiqiang-hhhh:
FE will return empty set instead of error, if:

1. BE report error status
2. An instance has multiple pipelines
3. FE process report slowly

This pr fix above problem on branch-2.0 by:
1. Exchange sender will spread Error status to receiver so that
2. Receiver of last result sink fragment will make itself cancelled instead of send an EOS to FE.

PS: Maybe the fix is not perfect ... but it works with almost least modification. A complete fix needs almost a refactor on DataExchangeNode.


zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

yiguolei (approved):
LGTM

PR #28688, opened 2023-12-20
[fix](load) use average of  totalTaskExcutionTimeMs to taskExecuteTimeMs to avoid double-counting

YenchangChan:
## Proposed changes

Issue Number: open #28654 

`taskExectuteTimeMs` double-counting `current_concurrent_number` times,  That cause incorrect when caculate `loadRowsRate` and `receivedBytesRate`, so I give averate value of `totalTaskExcutionTimeMs` to it.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/load/routineload/RoutineLoadJob.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/routineload/RoutineLoadJob.java
@@ -789,6 +789,12 @@ public abstract class RoutineLoadJob extends AbstractTxnStateChangeCallback impl
         this.jobStatistic.unselectedRows += unselectedRows;
         this.jobStatistic.receivedBytes += receivedBytes;
         this.jobStatistic.totalTaskExcutionTimeMs += taskExecutionTime;
+        if (currentTaskConcurrentNum > 0) {
+            this.jobStatistic.taskExecuteTimeMs = this.jobStatistic.totalTaskExcutionTimeMs / currentTaskConcurrentNum;
+        }
+        if (this.jobStatistic.taskExecuteTimeMs == 0) {
+            this.jobStatistic.taskExecuteTimeMs = 1;
+        }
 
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/org/apache/doris/load/routineload/RoutineLoadJob.java
```diff
@@ -789,6 +789,12 @@ private void updateNumOfData(long numOfTotalRows, long numOfErrorRows, long unse
         this.jobStatistic.unselectedRows += unselectedRows;
         this.jobStatistic.receivedBytes += receivedBytes;
         this.jobStatistic.totalTaskExcutionTimeMs += taskExecutionTime;
+        if (currentTaskConcurrentNum > 0) {
+            this.jobStatistic.taskExecuteTimeMs = this.jobStatistic.totalTaskExcutionTimeMs / currentTaskConcurrentNum;
```
HHoflittlefish777, line 793: Can this.jobStatistic.totalTaskExcutionTimeMs / currentTaskConcurrentNum calculate accurately？

HHoflittlefish777:
run buildall

fe/fe-core/src/main/java/org/apache/doris/load/routineload/RoutineLoadJob.java
YenchangChan: Not absolutely accurate, but relatively close. The best way to calculate this is to add up the maximum execution times of multiple concurrent instances each time, but I haven't found a specific way to do this. I'm willing to close my PR if someone has a better way to implement it.

HHoflittlefish777: How about add a timer to the job to calculate the time?

YenchangChan: Theoretically a good idea， I'll try it. Thanks.

PR #28720, opened 2023-12-20
docs: Fix typo about checking the progress of building index

JaySon-Huang:
## Proposed changes

It should be `show build index` but not `show builde index`


JaySon-Huang:
/close

https://github.com/apache/doris/pull/32384 has removed all docs in apache/doris repo

PR #28766, opened 2023-12-21
Update get-starting.md

HSunboy:
Remove incorrect tar command parameters

## Proposed changes

Version 1.2 packages are XZ compressed files, not gzip compressed files. This change modified an incorrect parameter

image


httpshirley (approved):
LGTM

PR #28781, opened 2023-12-21
[Fix](executor)Fix scan queue query hang

wangbo:
## Proposed changes

Fix query may hang when set enable_workload_group_for_scan = true, this because there bug in BlockingPriorityQueue.blocking_get.
When set timeout_ms > 0  and calling method BlockingPriorityQueue.blocking_get, if queue is not empty,  then wait_successful could never be called, then caller can never get a task even queue is not empty.


wangbo:
run buildall

yiguolei:
This PR is useless any more.

PR #28823, opened 2023-12-21
[fix](hash join) fix stack overflow caused by evaluate case expr on huge build block

jacktengg:
## Proposed changes

Issue Number: close #xxx

Hash join operator will merge build blocks to a big block with a lot rows, and using variable length array in expr execution may cause stack overflow.

Also add `-Werror=vla` compiler option to report compile error by default and 
```
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wvla"
#pragma clang diagnostic pop
```
to ignore VLA usages where it's sure not to cause stack overflow.

## Further comments

### DIFF
--- a/be/CMakeLists.txt
+++ b/be/CMakeLists.txt
@@ -285,6 +285,8 @@ if (COMPILER_CLANG)
         endif()
         add_definitions(-DUSE_LIBCPP)
     endif()
+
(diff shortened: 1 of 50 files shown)


jacktengg:
run buildall

PR #28059, opened 2023-12-06
[minor](s3client) use LOG(INFO) to print s3 client debug log if enabled

morningman:
## Proposed changes

So that user only need to set the BE config `aws_log_level` to enable/disable s3 client's debug log,
no need to set `sys_log_verbose_modules` at same time.
It is useful when we only want to debug s3 client.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #28907, opened 2023-12-22
[minor](s3client) use LOG(INFO) to print s3 client debug log if enabled(#28059)

morningman:
bp (#28059)


morningman:
run buildall

PR #28522, opened 2023-12-16
[opt](task-assignment) use consistent hash as default task assigner and cache the consistent hash ring

morningman:
## Proposed changes

1. Use consistent hash algo as the default assigner for file query scan node
    A consistent assignment can better utilize the page cache of BE node.

2. Cache the consistent hash ring
    Init a consistent hash ring is time-consuming because there a thousands of virtual node need to be added.
    So cache it for better performance

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #28908, opened 2023-12-22
[opt](task-assignment) use consistent hash as default task assigner and cache the consistent hash ring (#28522)

morningman:
bp (#28522)


morningman:
run buildall

PR #28991, opened 2023-12-25
[fix](hash join) fix column ref DCHECK failure of hash join node block mem reuse

jacktengg:
## Proposed changes

Introduced by #28851, after evaluating build side expr, some columns in resulting block may be referenced more than once in the same block.

e.g. `coalesce(col_a, 'string')` if `col_a` is nullable but actually contains no null values, in this case funcition `coalesce` will insert a new nullable column which references the original `col_a`.  

## Further comments


jacktengg:
run buildall

HappenLee (approved):
LGTM

yiguolei (approved):
lgtm

PR #29068, opened 2023-12-26
[fix](outfile) add outfile and export write json column with string to csv

nextdreamblue:
bp #29067

PR #29069, opened 2023-12-26
[fix](outfile) add outfile write json column with string to csv

nextdreamblue:
bp https://github.com/apache/doris/pull/29068


cambyzju (approved):
LGTM

PR #27752, opened 2023-11-29
[fix](stream_load)fix bug when stream without content-length or chunk…

xingyingone:
## Proposed changes

Issue Number: close https://github.com/apache/doris/issues/27671

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #29169, opened 2023-12-27
[fix](stream_load)fix bug when stream without content-length or chunk… #27752

xingyingone:
1. forbid thed  stream_load without content-length or chunked Transfer Encoding
2. forbid thed  stream_load both with content-length and chunked Transfer Encoding

## Proposed changes

Issue Number: close #xxx

## Further comments

### DIFF
--- a/be/src/http/action/stream_load.cpp
+++ b/be/src/http/action/stream_load.cpp
@@ -285,6 +285,20 @@ Status StreamLoadAction::_on_header(HttpRequest* http_req, std::shared_ptr<Strea
         }
     }
 
+    if (UNLIKELY((http_req->header(HttpHeaders::CONTENT_LENGTH).empty() &&
+                  !ctx->is_chunked_transfer))) {
+        LOG(WARNING) << "content_length is empty and transfer-encoding!=chunked, please set "
+                        "content_length or transfer-encoding=chunked";
+        return Status::InvalidArgument(
+                "content_length is empty and transfer-encoding!=chunked, please set content_length "
+                "or transfer-encoding=chunked");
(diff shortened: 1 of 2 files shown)


xingyingone:
run   buildall

be/src/http/action/stream_load.cpp
```diff
@@ -285,6 +285,19 @@ Status StreamLoadAction::_on_header(HttpRequest* http_req, std::shared_ptr<Strea
         }
     }
 
+    if (UNLIKELY((http_req->header(HttpHeaders::CONTENT_LENGTH).empty() &&
```
yiguolei, line 288: Does master branch also has this problem?

be/src/http/action/stream_load.cpp
```diff
@@ -285,6 +285,19 @@ Status StreamLoadAction::_on_header(HttpRequest* http_req, std::shared_ptr<Strea
...
+                  !ctx->is_chunked_transfer))) {
+        LOG(WARNING) << "content_length is empty and transfer-encoding!=chunked, please set "
+                        "content_length or transfer-encoding=chunked";
+        return Status::InternalError(
```
yiguolei, line 292: DO NOT use internal error, because it is not an internal error. Maybe invalid input or invalid parameters?

xiaokang:
#29169

be/src/http/action/stream_load.cpp
xingyingone: yes

xingyingone:
run buildall

xingyingone:
run buildall

xingyingone:
run buildall

PR #29186, opened 2023-12-27
[fix][fe]Fixed show proc tablet inaccuracies

felixwluo:
## Proposed changes

Issue Number: close #xxx

Both tables simply need to change the tablet id to get the same result from the show proc statement.

before：
image
Just change the tablet id to get the result, 10139 does not belong to the tablet in the table id of 10127

later:
image

## Further comments


felixwluo:
run buildall

freemandealer:
Would you please make your commit-msg more clear? Especially the description of the problem you have solved. Also use plain-text instead of images in commit-msgs. @felixwluo

freemandealer:
Personally I think the fix is not necessary. The original code works as some a kind of typo tolerance. I understand that tabletid is globally unique in Doris. So when users forget to change the front, if the operation is still successful, it will make their workflow smoother, like some kind of syntactic sugar?

Any other suggestions are welcome.

nextdreamblue:
run external

nextdreamblue (approved):
LGTM

yiguolei (approved):
LGTM

PR #29306, opened 2023-12-29
[Fix](exectutor)Fix topicPublisher thread may NPE when no topic exists

wangbo:
1 How NPE happens?
```
 public int getTopicMapSize() {
    return (this.topic_map == null) ? 0 : this.topic_map.size();
  }

  public void putToTopicMap(TTopicInfoType key, java.util.List<TopicInfo> val) {
    if (this.topic_map == null) {
      this.topic_map = new java.util.EnumMap<TTopicInfoType,java.util.List<TopicInfo>>(TTopicInfoType.class);
    }
    this.topic_map.put(key, val);
  }

  @org.apache.thrift.annotation.Nullable
  public java.util.Map<TTopicInfoType,java.util.List<TopicInfo>> getTopicMap() {
    return this.topic_map;
  }
```

When visiting A thrift Class's collection size, should use getXXXsize(), should not first getXXX then get size, because it may be a null.

2 how to fix?
Currently even request has no topicinfo list,  it still needs send an empty rpc to be, because empty may means delete.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/common/publish/TopicPublisherThread.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/publish/TopicPublisherThread.java
@@ -66,9 +66,8 @@ public class TopicPublisherThread extends MasterDaemon {
             topicPublisher.getTopicInfo(request);
         }
 
-        if (request.getTopicMap().size() == 0) {
-            return;
-        }
+        // even request contains no group and schedule policy, we still need to send an empty rpc.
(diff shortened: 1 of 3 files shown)


wangbo:
run buildall

yiguolei (dismissed):
LGTM

CalvinKirs (approved):

wangbo:
run buildall

yiguolei (approved):
ltm

PR #29373, opened 2024-01-02
Revert "[fix](journal) Fix infinite block due to initial BDB journal failed"

xiaokang:
Reverts apache/doris#29310, since it may cause p0 case fail.


xiaokang:
run buildall

xiaokang:
revert the pr for test

PR #29397, opened 2024-01-02
Revert "Revert "[fix](journal) Fix infinite block due to initial BDB journal failed""

xiaokang:
Reverts apache/doris#29373 to add back #29310


xiaokang:
run buildall

PR #29420, opened 2024-01-02
[opt](ES catalog)Add more description for limitations of docvalue_scan

qidaye:
## Proposed changes

Issue Number: close #29417
> [ignore_above](https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html)
Do not index any string longer than this value. Defaults to 2147483647 so that all values would be accepted. Please however note that default dynamic mapping rules create a sub keyword field that overrides this default by setting ignore_above: 256.

When the field value is longer than the limitation, we will get an empty result.
The solution is turn off `enable_docvalue_scan`, to get data from `_source` field.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #29459, opened 2024-01-03
[opt](ES catalog)Add more description for limitations of docvalue_scan

qidaye:
pick from master: #29420

> [ignore_above](https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html)
Do not index any string longer than this value. Defaults to 2147483647 so that all values would be accepted. Please however note that default dynamic mapping rules create a sub keyword field that overrides this default by setting ignore_above: 256.

When the field value is longer than the limitation, we will get an empty result. The solution is turn off `enable_docvalue_scan`, to get data from `_source` field.

## Proposed changes

Issue Number: close #xxx

## Further comments


qidaye:
run buildall

PR #29491, opened 2024-01-03
[refactor_close](VTabletWriter) plan_fragment_executor should call DataSink::try_close before close

zhiqiang-hhhh:
1. fix
VTabletWriter::close called  VTabletWriter::try_close before refactor
now VTabletWriter::close is called by be/src/runtime/plan_fragment_executor.cpp
2. some formatting


zhiqiang-hhhh:
Refactor of VTabletWriter should be the next target ...

PR #29593, opened 2024-01-05
[improvement](broker-load) Improve broker building process

murong00:
## Proposed changes

Currently the property `outputDirectory` in plugin `maven-thrift-plugin` is not conventional, users need to mark this directory as `Generated Sources Root` manually, this may confuse some new users. This pull request uses the preferred way as below:
- Thrift files are generated into `target/generated-sources` automatically when compiled, then the generated source files can be loaded automatically when reload the maven project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,7 +61,6 @@ fe_plugins/output
 fe_plugins/**/.factorypath
 
 fs_brokers/apache_hdfs_broker/src/main/resources/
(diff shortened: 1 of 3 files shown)


murong00:
run buildall

murong00:
run buildall

murong00:
@morningman Please help to take a review when you are available, thanks.

PR #29631, opened 2024-01-07
[improvement](statistics)Force to use zonemap for collecting string type min max.

Jibing-Li:
Force to use zonemap for collecting string type min max.
String type is not using zonemap for min max, because zonemap value at BE side is truncated at 512 bytes which may cause the value not accurate. But it's OK for statisitcs min max, and this could also avoid scan whole table while sampling.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #29662, opened 2024-01-08
[improvement](statistics) Force to use zonemap for collecting string type min max. (#29631)

Jibing-Li:
Force to use zonemap for collecting string type min max. String type is not using zonemap for min max, because zonemap value at BE side is truncated at 512 bytes which may cause the value not accurate. But it's OK for statisitcs min max, and this could also avoid scan whole table while sampling.

backport https://github.com/apache/doris/pull/29631

## Proposed changes

Issue Number: close #xxx

## Further comments


Jibing-Li:
run buildall

PR #29723, opened 2024-01-09
[feature](iceberg) iceberg table support equality delete

mongo360:
Support iceberg v2 feature:
support iceberg v2 table with equality delete file; three part to modify:
1. FE add identify columns that use for filter the data in equality delete files
2. PlanNodes.thrift add comment in TIcebergDeleteFileDesc to set the delete file type
3. BE iceberg_reader and parquet_reader to read data in equality delete file and do equality delete filter

iceberg v2
https://iceberg.apache.org/spec/#appendix-e-format-version-changes


mongo360:
run buildall

PR #29733, opened 2024-01-09
[fix] (schema change) ignore the error of repeatedly writing rowset when schema change

luwei16:
When load and schema change are concurrent, the same rowset may be written repeatedly. This error needs to be ignored.


luwei16:
run buildall

liaoxin01 (approved):
LGTM

dataroaring (approved):
LGTM

PR #29421, opened 2024-01-02
[Enhancement](inverted index) strictly checkout inverted index properties

airborne12:
## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #29748, opened 2024-01-09
[Enhancement](inverted index) strictly checkout inverted index properties (#29421)

xiaokang:
cherry pick from #29421


xiaokang:
run buildall

PR #29503, opened 2024-01-04
[opt](Nereids) add float type signature for sum aggregate function

zhangstar333:
## Proposed changes
cast_k9

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #29755, opened 2024-01-09
[opt](Nereids) add float type signature for sum aggregate function #29503

xiaokang:
cherry pick from #29503


xiaokang:
run buildall

PR #29665, opened 2024-01-08
[Bug](stream-load) fix stream load failed on table with rollup

BiteTheDDDDt:
## Proposed changes

[ANALYSIS_ERROR]TStatus: errCode = 2, detailMessage = can not cast from origin type VARCHAR(65533) to target type=INVALID_TYPE\n\n\t0# 

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #29757, opened 2024-01-09
[Bug](stream-load) fix stream load failed on table with rollup #29665

xiaokang:
cherry pick from #29665


xiaokang:
run buildall

PR #29766, opened 2024-01-10
[Improve] Use number of columns to count segment cache

Lchangliang:
## Proposed changes

Issue Number: close #xxx

If a segment has many columns, it will use a lot more memory than a segment with few columns, but it is not observable at present. So we use number of columns to control the memory.

## Further comments

### DIFF
--- a/be/src/common/config.cpp
+++ b/be/src/common/config.cpp
@@ -1054,6 +1054,7 @@ DEFINE_mInt32(schema_cache_sweep_time_sec, "100");
 
 // max number of segment cache, default -1 for backward compatibility fd_number*2/5
 DEFINE_mInt32(segment_cache_capacity, "-1");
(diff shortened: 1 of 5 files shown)


Lchangliang:
run buildall

gavinchou:
LGTM

be/src/common/config.cpp
```diff
@@ -1054,6 +1054,7 @@ DEFINE_mInt32(schema_cache_sweep_time_sec, "100");
 
 // max number of segment cache, default -1 for backward compatibility fd_number*2/5
 DEFINE_mInt32(segment_cache_capacity, "-1");
+DEFINE_mInt32(each_segment_have_columns, "30");
```
gavinchou, line 1057: estimated_num_columns_per_segment

PR #29850, opened 2024-01-11
[Fix](schema change) fix the bug that non light schema change tables can rename column

luwei16:
## How to reproduce the problem
1 crate table in 1.1
2 Upgrade to version 1.2
3 rename column
4 select * from table，Report an error.

## reason
Tables disable light schema change are not allowed to perform column rename. However, when upgrading from 1.1 to 1.2, maxColUniqueId will be initialized to 0, resulting in invalid judgment


dataroaring (approved):
LGTM

dataroaring:
run buildall

wm1581066 (approved):
LGTM

liaoxin01 (approved):
LGTM

PR #30047, opened 2024-01-17
[Feature](executor)Add scanbytes/scanrows condition, add move action in be

wangbo:
## Proposed changes

Add scanbytes/scanrows condition, add move action in be.


wangbo:
run buildall

PR #30095, opened 2024-01-18
[fix](chore) path gc should consider tablet migration

zhannngchen:
## Proposed changes

Issue Number: close #xxx

Background:
1. Migration will create new tablet in different DataDir, the old tablet will be moved to TabletManager::_shutdown_tablets.
2. The migration task won't copy data in stale rowsets to new tablet, so after migration, the new tablet don't contains stale rowsets of old tablet
3. The path GC process will check every path, to make sure if it's an useless tablet, or an useless rowset. If it is, will remove data of these tablets/rowsets

The issue:
1. When path GC got a stale rowset path from the data dir of old tablet, it extract the tablet id and rowset id
2. Then it check if the tablet id exists in TabletManager, and the answer is YES!
3. It got the tablet instance, which is the new tablet, then it check if the stale rowset id from the old tablet path exists in the new tablet instance, and got the answer NO.
4. The path GC process treat the rowset as an useless rowset, since it can't find anyone holds reference to it, then delete  the data of this stale rowset.
5. But some query may still holds reference to this stale rowset, the deletion will cause query failure.

Solution:
1. The lifecycle of all rowsets in a shutdown tablet, should be related with the lifecycle of this tablet
2. We need to differentiate the old tablet and the new one created by migration task, while performing path GC.

## Further comments

### DIFF
--- a/be/src/olap/data_dir.cpp
+++ b/be/src/olap/data_dir.cpp
@@ -670,7 +670,13 @@ void DataDir::_perform_path_gc_by_tablet(std::vector<std::string>& tablet_paths)
             std::swap(*forward, *backward);
             continue;
         }
-        if (auto tablet = _engine.tablet_manager()->get_tablet(tablet_id); !tablet) {
+        auto tablet = _engine.tablet_manager()->get_tablet(tablet_id);
+        if (!tablet || tablet->data_dir() != this) {
+            if (tablet) {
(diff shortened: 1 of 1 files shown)


zhannngchen:
run buildall

dataroaring (dismissed):
LGTM

yujun777 (approved):
LGTM

yiguolei (approved):

PR #30234, opened 2024-01-22
[fix](decimalv2) Fix cast to decimalv2 overflow

zhiqiang-hhhh:
before
```sql
select cast(12 as decimalv2(2,1))
--------------

+---------------------------+
| cast(12 as DECIMAL(2, 1)) |
+---------------------------+
|                      12.0 |
+---------------------------+
1 row in set (0.04 sec)
```
after
```sql
select cast(12 as decimalv2(2,1))
--------------

ERROR 1105 (HY000): errCode = 2, detailMessage = (xx)[CANCELLED][E-124] Arithmetic overflow, convert failed from 120, expected data is [-99, 99]
```


zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
decimalv2 will be removed in the future, so closed.

PR #30303, opened 2024-01-24
[feature](replica) Add drop replica safely on backend

yujun777:
Sometimes we want to drop specific replicas on the given backend.  We may use sql cmd 'admin set replica status = bad' to set the replica bad, and drop it later.  But it has two bad things:
1.  If the replica num is 1 or 2,  set one replica bad will make both read and write failed;
2.  Even replica num is 3,  set bad command is not always working. If the tablet is big, clone a new tablet may take long. During that time, when be report tablet is good, fe will auto change this replica from bad to ok.

So we add cmd to set replica status as drop. It's very like set replica as bad. But it solve the two bad things above. When set a replica as drop,  this replica is still both readable and writable.  After clone a new health tablet on another backend,  this replica can safely drop then.

usage:

```
ADMIN SET REPLICA STATUS PROPERTIES("tablet_id" = "10003", "backend_id" = "10001", "status" = "drop");
```

If set a replica as drop,  show tablet xx => detail cmd => this replica's field IsUserDrop = true

## Proposed changes

Issue Number: close #xxx

## Further comments


yujun777:
run buildall

yujun777:
run buildall

dataroaring (approved):
LGTM

deardeng (approved):
LGTM

PR #30529, opened 2024-01-29
[Improvement](executor)Remove ThreadPoolToken from MemTableFlushExecutor

wangbo:
## Proposed changes
ThreadPoolToken not support call wait and run in one threadpool, while we want to put wait/run in one threadpool later to using workload group to limit cpu.
So we can implement ThreadPoolToken's function in FlushToken and remove ThreadPoolToken.


wangbo:
run buildall

kaijchen (approved):

yiguolei (approved):
LGTM

PR #30536, opened 2024-01-29
[enhancement](stmt-forward) record query result for proxy query to avoid EOF

Hastyshell:
## Proposed changes

Originally, if the forwarded statement is a query statement, the query result will be sent to the Master MySQL channel directly, rather than replied to Follower. But actually, these packets should be sent to the Follower MySQL channel.

Therefore, we should record these packets and response them to Follower, so that they could be written to the right channel to avoid weird EOF result when forwarding a query statement.

Most of queries won't be forwarded to master, however, this case does be triggered if follower is far behind master or it is transfering to master..., anyway, making this case assignable.


TangSiyang2001:
run buildall

TangSiyang2001:
run buildall

TangSiyang2001:
run buildall

TangSiyang2001:
run buildall

morningman (approved):
LGTM

dataroaring (approved):
LGTM

PR #30573, opened 2024-01-30
[improvement](load) Enable lzo & Remove dependency on Markus F.X.J. Oberhumer's lzo library

### LINKED ISSUE #29406: [Enhancement] broke load enhanced
### Search before asking

### Description

2.0.4
lzo  and snappy is not support by  broke load
image

### Solution

it is better to support  it

### Code of Conduct

HowardQin:
Issue Number: close #29406

1. increase lzop version to 0x1040, 
    I set to 0x1040 only for decompressing lzo files compressed by higher version of lzop,
	no change of decompressing logic,
	actully, 0x1040 should have "F_H_FILTER" feature, 
	but it mainly for audio and image data, so we do not support it.
2. use orc::lzoDecompress() instead of lzo1x_decompress_safe() to decompress lzo data
3. use crc32c::Extend() instead of lzo_crc32() 
4. use olap_adler32() instead of lzo_adler32()
5. thus, remove dependency of Markus F.X.J. Oberhumer's lzo library
6. remove DORIS_WITH_LZO, so lzo file are supported by stream and broker load by default
7. add some regression test


be/src/exec/lzo_decompressor.cpp
```diff
@@ -24,7 +24,7 @@ namespace doris {
 const uint8_t LzopDecompressor::LZOP_MAGIC[9] = {0x89, 0x4c, 0x5a, 0x4f, 0x00,
                                                  0x0d, 0x0a, 0x1a, 0x0a};
 
-const uint64_t LzopDecompressor::LZOP_VERSION = 0x1030;
+const uint64_t LzopDecompressor::LZOP_VERSION = 0x1040;
```
yiguolei, line 27: Does this modification affect existing data?

HowardQin:
run buildall

yiguolei:
@HowardQin  Hi, if the user is using the lzo before doris 2.0, and then upgrade to this version, there already some data , does it could upgrade successfully?

HowardQin:
> @HowardQin Hi, if the user is using the lzo before doris 2.0, and then upgrade to this version, there already some data , does it could upgrade successfully?

orc::lzoDecompress() can only decompress files compressed by lzop with level 1 to 6, so it's not 100% compatible with liblzo2.a.
If old data in segment compressed using lzo with level above 6, this can be a problem, but I do not have enough info.
however, currently, doris does not have any code about decompressing lzo data in segment, there's only lzo-decompress logic for loading lzop compressed files. it seems we have already unsupported reading lzo data in segment.

HowardQin:
run buildall

yiguolei:
@HowardQin You are right. lzo is only used during read parquet data.

yiguolei (approved):

morningman (approved):
LGTM

PR #30633, opened 2024-01-31
[enhancement](segcompaction) refine segcompaction policy for wide-table

zhannngchen:
## Proposed changes

Issue Number: close #xxx

1. For a wide table, we need to strictly control the number of segments, because the compaction and select* queries of a wide table are very memory intensive. Specifically, each column iterator will preload a page (64KB), so for a table 5000 columns wide, a segment with all columns open at the same time will consume at least 320MB of memory.
2. at the same time for a wide table, because a memtable can store less rows, wide table load tasks tends to generate more segments.
3. segcompaction can effectively reduce the number of segments, but it will not process the last few segments that are less than `segcompaction_batch_size`. this is not a problem for normal tables, but for wide tables, the wider the table, the more we should minimize the number of segments.

**Optimization effect**:
- For a table with 5000 columns and 18 segment generated in a rowset, the preloaded page will consume 5.7GB of memory for a select * of that rowset.
- If segment compaction is turned on, under the previous default policy, 9 segments will eventually be generated and select * will consume 2.8GB of memory
- With this patch, only 2 segments will be generated and select* will consume 640MB of memory

I'll be submitting an optimization to adaptively adjust the batch size based on the table width later on

## Further comments

### DIFF
--- a/be/src/olap/rowset/beta_rowset_writer.cpp
+++ b/be/src/olap/rowset/beta_rowset_writer.cpp
@@ -321,7 +321,7 @@ Status BetaRowsetWriter::_load_noncompacted_segment(segment_v2::SegmentSharedPtr
  *     length is beyond (config::segcompaction_batch_size / 2)
  */
 Status BetaRowsetWriter::_find_longest_consecutive_small_segment(
-        SegCompactionCandidatesSharedPtr& segments) {
+        SegCompactionCandidatesSharedPtr& segments, bool last_batch) {
     segments = std::make_shared<SegCompactionCandidates>();
     // skip last (maybe active) segment
     int32_t last_segment = _num_segment - 1;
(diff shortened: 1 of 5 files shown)


zhannngchen:
run buildall

dataroaring (dismissed):
LGTM

be/src/olap/rowset/beta_rowset_writer.cpp
```diff
@@ -552,6 +564,13 @@ Status BetaRowsetWriter::_close_file_writers() {
         } else {
             RETURN_NOT_OK_STATUS_WITH_WARN(_wait_flying_segcompaction(),
                                            "segcompaction failed when build new rowset");
+            // compact all remaining segments, and since we can't submit a segcompaction
+            // when `_is_doing_segcompaction` is true, wait for the flying segcompaction
+            // to finish first, before doing the check
+            RETURN_NOT_OK_STATUS_WITH_WARN(_segcompaction_if_necessary(true),
```
freemandealer, line 570: As the 'compact remaining' operation is critical for the close wait latency, queueing in the segcompaction task queue is the last thing we want. So maybe it should have a higher priority than those normal & best-effort segcompaction task?

freemandealer (commented):
Also add some ut tests in be/test/olap/segcompaction_test.cpp?

dataroaring:
run buildall

PR #30662, opened 2024-01-31
[test](ut) fix fe ut failed as get the same port

zhangstar333:
## Proposed changes
some case of fe ut failed, as get the same port.
Issue Number: close #xxx

## Further comments


zhangstar333:
run buildall

fe/fe-core/src/test/java/org/apache/doris/utframe/AnotherDemoTest.java
```diff
@@ -80,17 +80,17 @@ public static void tearDown() {
 
     // generate all port from valid ports
     private static void getPorts() {
-        fe_http_port = UtFrameUtils.findValidPort();
```
SWJTU-ZhangLei, line 83: Does it will get the same port when call `UtPortUtils.findValidPort()` multi times

zhangstar333:
run buildall

fe/fe-core/src/test/java/org/apache/doris/utframe/AnotherDemoTest.java
zhangstar333: some case failed as get the same port, but system have a lots of free port.
 I guess as  call findValidPort without lock and have multi same implementation of findValidPort in different files.
so I add synchronized and all call findValidPort from UtPortUtil file.

PR #30677, opened 2024-02-01
[improve](backup) Skip all backup/restore jobs if max allowd option is set to 0

w41ter:
## Proposed changes

Issue Number: close #xxx

If there are too many backup/restore jobs, it may cause OOM. This PR allows the user to skip all backup/restore jobs if max_backup_restore_job_num_per_db is set to 0.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #30800, opened 2024-02-04
[improve](backup) Skip all backup/restore jobs if max allowd option is set to 0

w41ter:
Cherry-pick #30677


w41ter:
run buildall

PR #101, opened 2017-09-16
Added the fix to mysql build error.

zuyu:
This PR added the fix when building mysql error:

```
Inconsistency detected by ld.so: dl-version.c: 224: _dl_check_map_versions: Assertion `needed != ((void *)0)' failed!
```

It also removed the trailing whitespaces.

Finally, `max_align_t` error when building be may also happen in `Ubuntu 14.04 LTS`.

PR #30906, opened 2024-02-06
[Improve](Tablet Schema) Use deterministic way to serialize protobuf …

eldenmoon:
…(#101)

## Proposed changes

Issue Number: close #xxx

## Further comments

### DIFF
--- a/be/src/olap/tablet_schema.cpp
+++ b/be/src/olap/tablet_schema.cpp
@@ -20,6 +20,9 @@
 #include <gen_cpp/Descriptors_types.h>
 #include <gen_cpp/olap_file.pb.h>
 #include <glog/logging.h>
+#include <google/protobuf/io/coded_stream.h>
+#include <google/protobuf/io/zero_copy_stream.h>
+#include <google/protobuf/io/zero_copy_stream_impl_lite.h>
 
 #include <algorithm>
 #include <cctype>
@@ -27,6 +30,7 @@
 #include <cmath> // IWYU pragma: keep
 #include <memory>
 #include <ostream>
+#include <vector>
 
 #include "common/compiler_util.h" // IWYU pragma: keep
 #include "common/consts.h"
@@ -770,7 +774,7 @@ void TabletIndex::to_schema_pb(TabletIndexPB* index) const {
         index->add_col_unique_id(col_unique_id);
     }
     index->set_index_type(_index_type);
-    for (auto& kv : _properties) {
+    for (const auto& kv : _properties) {
         (*index->mutable_properties())[kv.first] = kv.second;
     }
     index->set_index_suffix_name(_escaped_index_suffix_path);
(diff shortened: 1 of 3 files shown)


eldenmoon:
run buildall

be/src/olap/tablet_schema.cpp
```diff
@@ -1382,4 +1386,13 @@ bool operator!=(const TabletSchema& a, const TabletSchema& b) {
     return !(a == b);
 }
 
+std::string TabletSchema::deterministic_string_serialize(const TabletSchemaPB& schema_pb) {
+    std::string output;
+    google::protobuf::io::StringOutputStream string_output_stream(&output);
+    google::protobuf::io::CodedOutputStream output_stream(&string_output_stream);
+    output_stream.SetSerializationDeterministic(true);
```
yiguolei, line 1393: Are you sure the deterministic serialize will take affect on map value?
void CodedOutputStream::SetSerializationDeterministic(
        bool value) 
Indicate to the serializer whether the user wants derministic serialization.

The default when this is not called comes from the global default, controlled by SetDefaultSerializationDeterministic.

What deterministic serialization means is entirely up to the driver of the serialization process (i.e. the caller of methods like WriteVarint32). In the case of serializing a proto buffer message using one of the methods of [MessageLite](https://protobuf.dev/reference/cpp/api-docs/google.protobuf.io.coded_stream/google.protobuf.message_lite#MessageLite), this means that for a given binary equal messages will always be serialized to the same bytes. This implies:

Repeated serialization of a message will return the same bytes.
Different processes running the same binary (including on different
machines) will serialize equal messages to the same bytes.

Note that this is not canonical across languages. It is also unstable across different builds with intervening message definition changes, due to unknown fields. Users who need canonical serialization (e.g. persistent storage in a canonical form, fingerprinting) should define their own canonicalization specification and implement the serializer using reflection APIs rather than relying on this API.

be/src/olap/tablet_schema.h
```diff
@@ -363,17 +366,17 @@ class TabletSchema {
     // Dump [(name, type, is_nullable), ...]
     string dump_structure() const {
         string str = "[";
-        for (auto p : _field_name_to_index) {
+        for (auto p : _cols) {
```
yiguolei, line 369: why change this code?

be/src/olap/tablet_schema.cpp
eldenmoon: Yes , map serialization is none deterministic by default, and `SetSerializationDeterministic` to true will make it deterministic.From the document above `This means that for a given binary equal messages will always be serialized to the same bytes`.

I've also tested and it's deterministic as expected

be/src/olap/tablet_schema.h
eldenmoon: `_field_name_to_index` is an unorded_map. change it to make output ordered

xiaokang (approved):
LGTM

qidaye (approved):

PR #30946, opened 2024-02-07
[Bug] doris cloud ms does not start even though process is not running

### LINKED ISSUE #30945: [Bug] doris cloud ms does not start even though process is not running
### Search before asking

### Version

master

### What's Wrong?

the doris cloud meta service script, start.sh fails to start the service even though there is no doris_cloud process running

$ bash bin/start.sh --meta-service
pid file existed, doris_cloud have already started, pid=107514
$ pgrep -f 107514 |grep doris_cloud
$ 

### What You Expected?

if doris_cloud is not running then the old pid file should be removed and proceed to start the process.

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

rohitrs1983:
## Proposed changes

Issue Number: close #30945 

if doris_cloud is not running then the old pid file should be removed and proceed to start the process.

pgrep matches the pid, this will not work; pgrep matched on process name
the existing code can we modified to fix this,
if ! pgrep -f doris_cloud 2>&1 | grep "${pid}" >/dev/null 2>&1; then

but this change also is not correct. if there is another process by the name doris_cloud then pgrep will return incorrect pid, which will again cause incorrect behavior.
rather than using pgrep, we should use kill command, passing the doris_cloud pid file as input.

## Further comments

### DIFF
--- a/cloud/script/start.sh
+++ b/cloud/script/start.sh
@@ -45,7 +45,7 @@ process=doris_cloud
 if [[ -f "${DORIS_HOME}/bin/${process}.pid" ]]; then
     pid=$(cat "${DORIS_HOME}/bin/${process}.pid")
     if [[ "${pid}" != "" ]]; then
-        if ! pgrep -f "${pid}" 2>&1 | grep doris_cloud >/dev/null 2>&1; then
+        if kill -0 "$(cat "${DORIS_HOME}/bin/${process}.pid")" >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


Nitin-Kashyap (approved):
please update description, why pgrep wont work, but kill -0 will

dataroaring (approved):
LGTM

dataroaring:
run buildall

LemonLiTree (approved):
LGTM

PR #30761, opened 2024-02-02
[BugFix](MultiCatalog) Fix oss file location is not avaiable in iceberg hadoop catalog

GoGoWen:
## Proposed changes

1, create iceberg hadoop catalog  like below:
CREATE CATALOG `iceberg_catalog` PROPERTIES (
"warehouse" = "s3a://xxx/xxx",
"type" = "iceberg",
"s3.secret_key" = "*XXX",
"s3.region" = "region",
"s3.endpoint" = "http://xxx.jd.local",
"s3.bucket" = "xxx-test",
"s3.access_key" = "xxxxx",
"iceberg.catalog.type" = "hadoop",
"fs.s3a.impl" = "org.apache.hadoop.fs.s3a.S3AFileSystem",
"create_time" = "2024-02-02 11:15:28.570"
);

2, run select * from iceberg_catalog.table limit 1;

will get  errCode = 2, detailMessage = Unknown file location nullnulls3a:/xxxx

expect: 
OK

also need to bp to branch-2.0

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #30992, opened 2024-02-08
[BugFix](MultiCatalog) Fix oss file location is not avaiable in iceberg hadoop catalog #30761

xiaokang:
cherry pick from #30761


xiaokang:
run buildall

PR #30994, opened 2024-02-08
[fix](dynamic partition) fix create too many dynamic partitions

yujun777:
For dynamic partitions,  the expect created partition num = dynamic_partition.end - dynamic_partition.start.
By default,  fe limit 500(Config.max_dynamic_partition_num) dynamic partitions.

But if dynamic_partition.start is near  int.min_value， the expect create partition num will be overflow and become a negative number.  Then check created partition num < Config.max_dynamic_partition_num is ok.

For example:  if start = -2147483647 (int.min_value + 1)， end = 1, then end - start =   -2147483648 < 0

So change the expect create partition num to long to avoid overflow.

## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #31039, opened 2024-02-18
[fix](dynamic partition) fix create too many dynamic partitions #30994

yujun777:
cherry pick from #30994


yujun777:
run buildall

PR #30972, opened 2024-02-07
[fix](repository) fix NullPointerException when execute create repository statement without properties

caiconghui:
## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #31069, opened 2024-02-18
[fix](repository) fix NullPointerException when execute create repository statement without properties (#30972)

caiconghui:
bp #30972


morningman:
run buildall

PR #31175, opened 2024-02-20
[regression-test] Analyze each table after loading all tables

gavinchou:
Analyze may rely on row count reported by BEs, if we analyze immediately
after loading to a table, the row count of the table may be still 0
which may lead to improper behavior of analyze manager.

This change will introduce extra 70 seconds wait time for the cases affected.
"70s" is slightly lager than default value (60s) of tablet_stat_update_interval_second fe.conf.
To reduce this duration, change tablet_stat_update_interval_second
to a smaller value, say 10 seconds and change 70s to 15s.


gavinchou:
run buildall

Jibing-Li (approved):
LGTM

yiguolei (approved):

PR #31231, opened 2024-02-21
[fix](bdbje) Initializing and starting the first fe node will cause an exception when the disk is insufficient.

liugddx:
## Proposed changes

Issue Number: close #xxx

38bffceac452bed9207665df2941c0d

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/journal/bdbje/BDBEnvironment.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/journal/bdbje/BDBEnvironment.java
@@ -207,7 +207,8 @@ public class BDBEnvironment {
                         LOG.warn("", e1);
                     }
                 } else {
-                    LOG.error("error to open replicated environment. will exit.", e);
+                    LOG.error("error to open replicated environment,"
+                            + "Maybe it's due to insufficient disk. will exit.", e);
                     System.exit(-1);
                 }
             }


liugddx:
run buildall

fe/fe-core/src/main/java/org/apache/doris/journal/bdbje/BDBEnvironment.java
```diff
@@ -207,7 +207,8 @@ public void setup(File envHome, String selfNodeName, String selfNodeHostPort,
                         LOG.warn("", e1);
                     }
                 } else {
-                    LOG.error("error to open replicated environment. will exit.", e);
+                    LOG.error("error to open replicated environment,"
+                            + "Maybe it's due to insufficient disk. will exit.", e);
```
SWJTU-ZhangLei, line 211: The image exception indicate the bdbje state is unknown and disallow to write,  adding a `insufficient disk` log infomation here may be confused in other exception.

liugddx: > The image exception indicate the bdbje state is unknown and disallow to write, adding a `insufficient disk` log infomation here may be confused when met other exception.

This is just to provide users with an idea to solve the problem. In fact, I couldn't get the status due to insufficient disk.

PR #31263, opened 2024-02-22
[feature](nereids) deal the slots that appear both in agg func and grouping sets

feiniaofeiafei:
## Proposed changes

select sum(a) from t group by grouping sets ((a)); The slot a  appearing both in agg func and grouping sets. 
Report an error: col_int_undef_signed cannot both in select list and aggregate functions when using GROUPING SETS/CUBE/ROLLUP, please use union instead.
 This PR removes the restriction and supports this situation.


924060929:
please add regression test case

starocean999:
run buildall

924060929:
run buildall

feiniaofeiafei:
run buildall

PR #25952, opened 2023-10-26
[pipeline](exec) disable shared scan in default and disable shared scan in limit with where scan

HappenLee:
## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #31342, opened 2024-02-23
[fix](plan) only scan node with limit and no predicate can reduce to 1 instance

morningman:
## Proposed changes

This PR #25952 introduce a opt that if a scan node has limit and predicates, use only 1 instance to save cup and memory.
But this is wrong because we can not guarantee that the predicates can truly help to prune the data.
So I modify the logic to remove this opt.
Now, only scan node with limit and NO predicate can reduce to only 1 instance.

## Further comments


morningman:
run buildall

yiguolei (approved):

HappenLee (approved):
LGTM

PR #31169, opened 2024-02-20
[fix](file-writer) avoid empty file for segment writer

morningman:
## Proposed changes

This #30703 allow writing empty file.
But internal segment file do not allow empty file.
So I add a new FileWriterOption `create_empty_file` to indicate whether to
create empty file.

This only effect s3/hdfs/broker file writer

After this,
1. For segment file, no empty file will be created.
2. For export operation, if there is no data, empty file will be created.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #31355, opened 2024-02-23
[fix](file-writer) avoid empty file for segment writer

morningman:
bp #31169


morningman:
run buildall

morningman:
run buildall

PR #31465, opened 2024-02-27
[fix] Fix the incorrect class name in the getLogger method call of MysqlTable.

ytwp:
Fix the incorrect class name in the getLogger method call of MysqlTable.


yiguolei (approved):

yiguolei:
run buildall

BePPPower (approved):
LGTM

PR #31475, opened 2024-02-27
[Fix](nereids)Window agg grouping sets Bug Fix

feiniaofeiafei:
```sql        
select a, c, sum(sum(b)) over(partition by c order by c rows between unbounded preceding and current row) 
from test_window_table2 group by grouping sets((a),( c)) having a > 1 order by 1,2,3;
```
for this kind of case:
sum(sum(col)) over, nereids has cannot find slot problem.
the output slot of repeat and aggregate is computed wrongly.
Only collecting the trival-agg  in NormalizeRepeat can fix this problem.


feiniaofeiafei:
run buildall

starocean999 (approved):

morrySnow (approved):

PR #31228, opened 2024-02-21
[opt](cancel) Cancel get result future  immediately if query is cancelled 

zhiqiang-hhhh:

PR #31490, opened 2024-02-27
[opt](cancel) Cancel get result future  immediately if query is cancelled (pick #31228)

zhiqiang-hhhh:
pick #31228


zhiqiang-hhhh:
run buildall

PR #31551, opened 2024-02-28
[fix](partial update) compaction may cause update failue

zhannngchen:
## Proposed changes

Issue Number: close #xxx

Under normal circumstances, `get_rowset_by_ids` does not need to consider the stale rowset, in other word, if a rowset id is not found in the normal rowset, we can ignore it. This is because even if we handle stale rowset here, we need to recalculate the new rowset generated by the corresponding compaction in the publish phase.

However, for partial update, ignoring the stale rowset may cause some keys to not be found in the flush phase (lookup_row_key returns KEY_NOT_FOUND), and thus be mistaken as new keys in the flush phase, which will cause the load to fail in the following two cases:
1. when strict_mode is enabled, new keys are not allowed to be added.
2. Some columns that need to be filled are neither nullable nor have a default value, in which case the value of the field cannot be filled as a new key, leading to a failure of the load.

## Further comments


zhannngchen:
run buildall

zhannngchen:
run buildall

liaoxin01 (approved):
LGTM

Yukang-Lian (approved):

PR #31578, opened 2024-02-29
[Fix](executor)Fix insert select forward carry workload_group

wangbo:
## Proposed changes
Fix insert select forward carry workload_group, otherwise insert won't using user expected group.


wangbo:
run buildall

yiguolei (approved):

HappenLee (approved):
LGTM

PR #31606, opened 2024-02-29
[fix](plan) only scan node with limit and no predicate can reduce to 1 instance (#31342)

morningman:
bp #31342


morningman:
run buildall

morningman:
run buildall

PR #31651, opened 2024-03-01
[refine](pipelineX) runtime state owner by the Task, not by the FragmentContext.

Mryange:
## Proposed changes

Issue Number: close #xxx

## Further comments


Mryange:
run buildall

Mryange:
run buildall

Mryange:
run buildall

Mryange:
run pipelinex_p0

be/src/pipeline/pipeline_x/pipeline_x_fragment_context.cpp
```diff
@@ -108,9 +108,8 @@ PipelineXFragmentContext::~PipelineXFragmentContext() {
     if (!_task_runtime_states.empty()) {
         // The memory released by the query end is recorded in the query mem tracker, main memory in _runtime_state.
```
yiguolei, line 109: not modify this logic, it is very dangerous.
Currently, there are many threads in doris, we use pipeline fragment context's weak ptr to make sure that the related resource is not released during another thread. All resources or objects that will be accessed between threads should be in pipeline fragment context. For example, the profile is in runtime state and the profile will also be accessed in async result writer thread. If the pipeline finished and the runtime state will be released, it will core.
And also the rpc thread callback.

PR #31671, opened 2024-03-01
[Improvement](executor)Routine load support workload group

wangbo:
## Proposed changes

Routine load support workload group
1 user can specify a workload group for routine load.
2 if user does not specify a workload group, then we could get workload group from user's property.
3 if user's propery does not have a valid workload group, then it will throw an exception.


wangbo:
run compile

wangbo:
run compile

wangbo:
run buildall

yiguolei (approved):

sollhui (commented):
please add doc to introduce how to use workload_group in routine load.

sollhui (approved):
LGTM

PR #31759, opened 2024-03-04
[fix] Fix the incorrect class name in the getLogger method call of MysqlTable. #31465

liaoxin01:
cherry pick from #31465


liaoxin01:
run buildall

PR #31403, opened 2024-02-26
[Fix](fe-core) The EliminateSortUnderSubquery rule and the EliminateOrderByConstant rule share the same ruleType

feelshana:
The EliminateSortUnderSubquery rule and the EliminateOrderByConstant rule share the same ruleType (#31402)

## Proposed changes

Issue Number: close #31402

seprate the ruletype from EliminateSortUnderSubquery and EliminateOrderByConstant

PR #31779, opened 2024-03-05
[Fix](fe-core) The EliminateSortUnderSubquery rule and the EliminateOrderByConstant rule share the same ruleType #31403

xiaokang:
cherry pick from #31403


xiaokang:
run buildall

PR #31767, opened 2024-03-05
[fix](Nereids) json_object can't be odd parameters, need even parameters

iwanttobepowerful:
## Proposed changes

Issue Number: close #31766

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #31907, opened 2024-03-07
[fix](Nereids) json_object can't be odd parameters, need even parameters #31767

xiaokang:
cherry pick from #31767


xiaokang:
run buildall

PR #31970, opened 2024-03-07
[enhancement](Nereids) Speedup PartitionPrunner

924060929:
## Proposed changes

This pr imporve the high QPS query by speed up PartitionPrunner
1. remove useless Date parse/format, use `LocalDate` instead
2. fast evaluate path for single value partition
3. change `Collection.stream()` to `ImmutableXxx.builderWithExpectedSize(n)` to skip useless method call and collection resize
4. change lots of if-else to switch
5. don't parse to string to compare dateLiteral, use int field compare instead

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/nereids/rules/expression/rules/OneRangePartitionEvaluator.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/nereids/rules/expression/rules/OneRangePartitionEvaluator.java
@@ -53,6 +53,7 @@ import org.apache.doris.nereids.util.Utils;
 import com.google.common.collect.BoundType;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableMap.Builder;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
@@ -62,11 +63,11 @@ import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiFunction;
-import java.util.stream.IntStream;
 
 /**
  * OneRangePartitionEvaluator.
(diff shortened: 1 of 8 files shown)


wm1581066:
Any performance data?

924060929:
run buildall

924060929:
> Any performance data?

100 threads parallel continuous send this sql which query an empty table
```sql
select  count(1),date_format(time_col,'%Y%m%d'),varchar_col1
from tbl 
where  partition_date>'2024-02-15'  and (varchar_col2 ='73130' or varchar_col3='73130') and time_col>'2024-03-04' 
  and  time_col<'2024-03-05'
group by date_format(time_col,'%Y%m%d'),varchar_col1 
order by date_format(time_col,'%Y%m%d') desc, varchar_col1 desc,count(1) asc
limit 1000
```

before: 1600 peak QPS, about 1300 avg QPS
after: 3100 peak QPS, about 2700 avg QPS

morrySnow (approved):

starocean999 (approved):

fe/fe-core/src/main/java/org/apache/doris/nereids/rules/expression/rules/OneRangePartitionEvaluator.java
```diff
@@ -102,31 +103,33 @@ public OneRangePartitionEvaluator(long partitionId, List<Slot> partitionSlots,
 
         PartitionRangeExpander expander = new PartitionRangeExpander();
         this.partitionSlotTypes = expander.computePartitionSlotTypes(lowers, uppers);
-        this.slotToType = IntStream.range(0, partitionSlots.size())
-                .mapToObj(index -> Pair.of(partitionSlots.get(index), partitionSlotTypes.get(index)))
-                .collect(ImmutableMap.toImmutableMap(Pair::key, Pair::value));
+        this.slotToType = Maps.newHashMapWithExpectedSize(partitionSlots.size() * 2);
```
starocean999, line 106: Minor: use Maps.newHashMapWithExpectedSize(partitionSlots.size()); may good enough, can update the code in next pr

924060929: I will change to 16 in next pr, because the num of partitionSlots usually <= 3

PR #31890, opened 2024-03-06
[Fix](load) Fix insert into wrong value

Yukang-Lian:
Problem: When the old optimizer processes an INSERT INTO statement that contains two quotation marks, it results in only one quotation mark being written into the database.

Reason: During syntax parsing, the old optimizer interprets two quotation marks as a single quotation mark.

Solution: Remove the logic that consolidates two quotation marks into one.

## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #31986, opened 2024-03-08
[Revert](load) Revert "[Fix](load) Fix insert into wrong value #31890"

Yukang-Lian:
Revert "Problem: When the old optimizer processes an INSERT INTO statement that contains two quotation marks, it results in only one quotation mark being written into the database. (#31890)"

This reverts commit 8c309652e04698f311b6c9158105352e8416c69a.

The final conclusion is that we need to use double quotes '' to replace single quotes ‘ when inserting data. The logic of the old optimizer is difficult to modify, because it is necessary to maintain compatibility with certain scenarios. So this PR should be reverted.

## Proposed changes

Issue Number: close #xxx

## Further comments


Yukang-Lian:
run buildall

zhannngchen (approved):
LGTM

yiguolei (approved):

PR #31995, opened 2024-03-08
[Optimize](ParquetReader) optimize parquet read on oss

GoGoWen:
## Proposed changes

In our situation,   we use Doris to query iceberg table which data file stored in cloud OSS. As the oss charge we with the number of api call,  we want to optimize the number of api call.

change:
 1, let buffer size configurable, as we have many small file, and later will want reuse this instead of read file again to reduce OSS  api call 
 2, use heap instead of stack for parquet footer read buffer, in order to reduce the risk of stack overflow

## Further comments


GoGoWen:
run buildall

morningman (dismissed):
LGTM

be/src/common/config.h
```diff
@@ -967,6 +967,9 @@ DECLARE_Int32(s3_transfer_executor_pool_size);
 DECLARE_Bool(enable_time_lut);
 DECLARE_mBool(enable_simdjson_reader);
 
+// the size of parquet file to read
```
AshinGau, line 970:
```suggestion
// the initial size of parquet footer to read
```

AshinGau (dismissed):

PR #32031, opened 2024-03-10
[statistic](cloud) Support force full analysis to sample

gavinchou:
Full analysis may take a lot of time due to lack of file cache in cloud mode, which may lead to analysis timeout when we run regression test.

We can force sample analysis for the analysis statements such as `analyze table $table with sync` to get rid of analysis timeout.

fe.conf `force_sample_analyze = true;`, it is `false` by default

Co-authored-by: zhengyu <freeman.zhang1992@gmail.com>


dataroaring:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #32067, opened 2024-03-11
[enhancement](Nereids) Optimize expression

924060929:
## Proposed changes

1. optimize expression comparison by
  a) flatter method call stack
  b) short circuit: if some simple field not equals, then return, and not compare to the big field, like `Alias.name`
2. lazy compute `Alias.name` which is computed by `toSql`. Now `Alias.toSlot()` will not generate long name immediately
3. cache `Expresssion.inputSlots`, it can save time when invoke this method multiple times
4.  always compute `Expression.unbound`, it can avoid traverse the big expression tree

this pr can save about 200ms when submit some long sqls


924060929:
run buildall

924060929:
run buildall

morrySnow (approved):

starocean999 (approved):

PR #32253, opened 2024-03-14
[fix](jni) remove 'push_down_predicates' and fix BE crash with decimal predicate

morningman:
## Proposed changes

Previously, for JNI scanner, BE will assemble a `push_down_predicates` from `colname_to_value_range`
and pass it to JNI scanner. and JNI scanner will determine how to use it to do the predicate pushdown.
But actually, this `push_down_predicates` is never used.

This PR removes the `push_down_predicates`.
This will also avoid BE crash when encountering decimal type predicate(a bug, not fixed, but skip it)

## Further comments

Actually the predicate passed to JNI scanner can be assembled and serialized on FE side, and pass to BE
then pass to JNI scanner directly, just like what PaimonJniScanner do(see `paimon_predicate`).

### DIFF
--- a/fe/be-java-extensions/java-common/src/main/java/org/apache/doris/common/jni/JniScanner.java
+++ b/fe/be-java-extensions/java-common/src/main/java/org/apache/doris/common/jni/JniScanner.java
@@ -33,6 +33,14 @@ public abstract class JniScanner {
     protected VectorTable vectorTable;
     protected String[] fields;
     protected ColumnType[] types;
+    @Deprecated
+    // This predicate is from BE, but no used.
+    // TODO: actually, we can generate the predicate for JNI scanner in FE's planner,
+    // then serialize it to BE, and BE pass it to JNI scanner directly.
+    // NO need to use this intermediate expression, because each JNI scanner has its
+    // own predicate expression format.
+    // For example, Paimon use "PaimonScannerUtils.decodeStringToObject(paimonPredicate)"
+    // to deserialize the predicate string to PaimonPredicate object.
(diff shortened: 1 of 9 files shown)


morningman:
run buildall

morningman:
run buildall

AshinGau (approved):

fe/be-java-extensions/java-common/src/main/java/org/apache/doris/common/jni/JniScanner.java
```diff
@@ -33,6 +33,14 @@ public abstract class JniScanner {
...
     protected ColumnType[] types;
+    @Deprecated
+    // This predicate is from BE, but no used.
+    // TODO: actually, we can generate the predicate for JNI scanner in FE's planner,
+    // then serialize it to BE, and BE pass it to JNI scanner directly.
+    // NO need to use this intermediate expression, because each JNI scanner has its
+    // own predicate expression format.
+    // For example, Paimon use "PaimonScannerUtils.decodeStringToObject(paimonPredicate)"
+    // to deserialize the predicate string to PaimonPredicate object.
     protected ScanPredicate[] predicates;
```
BePPPower, line 44: Is this member variable still in use? Why not delete it?

morningman: I leave the related class.
Maybe we can delete it when refactor this logic.
We still need a predicate pushdown framework

BePPPower (approved):

PR #32255, opened 2024-03-14
[fix](jni) remove 'push_down_predicates' and fix BE crash with decimal predicate (#32253)

morningman:
bp #32253


morningman:
run buildall

PR #32299, opened 2024-03-15
[fix](group commit) Fix compatibility issues on serializing and deserializing wal file

hust-hhb:
Now wal (be_exec_version is 3) deserializing to block using the newest way，after merging this commit https://github.com/apache/doris/commit/954750b59a8c6d6eb0525c5daa6ed26d5351b839, deserializing wal will use the older way, which may lead to construct a wrong block. This pr using the newest be_exec_version to deserialize wal file  to avoid compatibility problem.
## Proposed changes

Issue Number: close #xxx

## Further comments


hust-hhb:
run buildall

be/src/vec/exec/format/wal/wal_reader.cpp
gavinchou, line 1: It's better to add UT to prevent such compatibility issues.

hust-hhb: now we have vwal_scanner_test ut to check this problem, and revert the wal file to old version,next time we can find out the compatibility issues by using this old wal file

hust-hhb:
run buildall

gavinchou (approved):

mymeiyi:
I think the problem here is that the block deserializing method should be compatible with all `be_exec_version`, rather than for version 3, use version 2 deserializing method.

be/src/common/config.h
```diff
@@ -1198,6 +1198,7 @@ DECLARE_mInt32(group_commit_queue_mem_limit);
 // group_commit_wal_max_disk_limit=1024 or group_commit_wal_max_disk_limit=10% can be automatically identified.
 DECLARE_mString(group_commit_wal_max_disk_limit);
 DECLARE_Bool(group_commit_wait_replay_wal_finish);
+DECLARE_Int32(group_commit_wal_version);
```
mymeiyi, line 1201: should not be a config, because user may change it

hust-hhb:
run buildall

be/src/common/config.h
hust-hhb: already change the code,take a look please @mymeiyi

hust-hhb:
run buildall

mymeiyi (approved):

hust-hhb:
run buildall

hust-hhb:
run buildall

hust-hhb:
run buildall

mymeiyi:
run p0

dataroaring (approved):
LGTM

PR #32340, opened 2024-03-17
[opt](file-meta-cache) reduce file meta cache size and disable cache for some cases

morningman:
## Proposed changes

File meta cache on BE is used to cache the meta for external table's file such as parquet footer.
This cache is counted by number, not memory consumption.
So if the cache object is big(eg, a large parquet footer), the total memory consumption of this cache
will be large and causing OOM.

This PR mainly changes:

1. Add a new method `exceed_prune_limit()` for `CachePolicy`
    For `ObjLRUCache`, it always return true so that the minor of full gc on BE will prune the cache each time.

2. Reduce the default capability of file meta cache, from 20000 to 1000

    Also change the default capability of hdfs file handle cache, from 20000 to 1000

3. Change judgement of whether enable file meta cache when querying

    If the number of file need to be read is larger than the 1/3 of the file meta cache's capability, file meta cache
    will be disabled for this query. Because cache is useless if there are too many files.

## Further comments

### DIFF
--- a/be/src/runtime/memory/lru_cache_policy.h
+++ b/be/src/runtime/memory/lru_cache_policy.h
@@ -120,6 +120,9 @@ public:
 
     uint64_t new_id() { return _cache->new_id(); };
 
+    // Subclass can override this method to determine whether to do the minor or full gc
+    virtual bool exceed_prune_limit() { return mem_consumption() > CACHE_MIN_FREE_SIZE; }
+
     // Try to prune the cache if expired.
     void prune_stale() override {
         COUNTER_SET(_freed_entrys_counter, (int64_t)0);
@@ -127,7 +130,7 @@ public:
         if (_stale_sweep_time_s <= 0 && _cache == ExecEnv::GetInstance()->get_dummy_lru_cache()) {
             return;
         }
-        if (mem_consumption() > CACHE_MIN_FREE_SIZE) {
+        if (exceed_prune_limit()) {
             COUNTER_SET(_cost_timer, (int64_t)0);
             SCOPED_TIMER(_cost_timer);
             const int64_t curtime = UnixMillis();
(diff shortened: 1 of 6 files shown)


AshinGau (approved):

morningman:
run buildall

be/src/runtime/memory/lru_cache_policy.h
```diff
@@ -120,14 +120,17 @@ class LRUCachePolicy : public CachePolicy {
 
     uint64_t new_id() { return _cache->new_id(); };
 
+    // Subclass can override this method to determine whether to do the minor or full gc
```
xinyiZzz, line 123: inaccurate comment. gc type only related to memory status, minor gc exectue `prune_stale` and full gc exectue`prune_all`.
`exceed_prune_limit` only used for whether to exectue this cache prune.

be/src/util/obj_lru_cache.cpp
```diff
@@ -45,4 +45,11 @@ void ObjLRUCache::erase(const ObjKey& key) {
     }
 }
 
+bool ObjLRUCache::exceed_prune_limit() {
+    // just return true to prune all cached obj.
+    // Because ObjLRUCache is counted with number, not memory.
+    // Simple prune all
+    return true;
```
xinyiZzz, line 52: great, in addition to obj lru cache, other number lru cache should also have similar override.

xinyiZzz (approved):
LGTM

be/src/runtime/memory/lru_cache_policy.h
morningman: We are talking about the same thing

be/src/util/obj_lru_cache.cpp
morningman: OK

PR #32367, opened 2024-03-18
[opt](file-meta-cache) reduce file meta cache size and disable cache for some cases (#32340)

morningman:
pick part of #32340

2. Reduce the default capability of file meta cache, from 20000 to 1000

    Also change the default capability of hdfs file handle cache, from 20000 to 1000

3. Change judgement of whether enable file meta cache when querying

    If the number of file need to be read is larger than the 1/3 of the file meta cache's capability, file meta cache
    will be disabled for this query. Because cache is useless if there are too many files.


morningman:
run buildall

morningman:
run buildall

PR #32416, opened 2024-03-18
[Feature](hdfs) check the connectivity of hdfs resources

### LINKED ISSUE #32322: [Feature] Check if the hdfs resource is available when creating hdfs resource
### Search before asking

### Description

Fe would check whether the s3 resource is available or not when creating the s3 resource which set property `s3_validity_check` true. 

### Use case

When creating hdfs resource we can type the stmt like the following 
```sql
CREATE STORAGE RESOURCE IF NOT EXISTS hdfs_resource
        PROPERTIES (
            "type"="hdfs",
            "hdfs_validity_check"="false",
            "fs.defaultFS"="127.0.0.1:8020",
            "hadoop.username"="hive",
            "hadoop.password"="hive",
            "dfs.nameservices" = "my_ha",
            "dfs.client.failover.proxy.provider" = "org.apache.hadoop.hdfs.server.namenode.ha.ConfiguredFailoverProxyProvider"
        );
```

### Related issues

_No response_

### Code of Conduct

felixwluo:
## Proposed changes

Issue Number: close #32322

`Check hdfs resources when they are created`

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/HdfsResource.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/HdfsResource.java
@@ -17,17 +17,23 @@
 
 package org.apache.doris.catalog;
 
+import org.apache.doris.backup.Status;
 import org.apache.doris.common.DdlException;
 import org.apache.doris.common.proc.BaseProcResult;
 import org.apache.doris.common.security.authentication.AuthenticationConfig;
+import org.apache.doris.common.util.PrintableMap;
+import org.apache.doris.fs.remote.dfs.DFSFileSystem;
 import org.apache.doris.thrift.THdfsConf;
 import org.apache.doris.thrift.THdfsParams;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.gson.annotations.SerializedName;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.Map;
 
 /**
@@ -43,6 +49,7 @@ import java.util.Map;
  * );
  */
 public class HdfsResource extends Resource {
+    private static final Logger LOG = LogManager.getLogger(HdfsResource.class);
     public static final String HADOOP_FS_PREFIX = "dfs.";
     public static String HADOOP_FS_NAME = "fs.defaultFS";
     public static String HADOOP_SHORT_CIRCUIT = "dfs.client.read.shortcircuit";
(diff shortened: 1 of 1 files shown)


felixwluo:
run buildall

ByteYue:
I wonder could we add one regression test? You can refer to the hdfs cases for cold heat separation case to create one case for this.

fe/fe-core/src/main/java/org/apache/doris/catalog/HdfsResource.java
```diff
@@ -63,16 +71,63 @@ public HdfsResource(String name) {
         properties = Maps.newHashMap();
     }
 
+    private static void pingHdfs(String hdfsUri, Map<String, String> properties) throws DdlException {
+        String testFile = hdfsUri + "/test_hdfs_valid.txt";
+        Map<String, String> propertiesPing = new HashMap<>();
+        propertiesPing.put(HADOOP_FS_NAME, hdfsUri);
+        properties.putAll(propertiesPing);
+        DFSFileSystem dfsSystem = new DFSFileSystem(properties);
```
ByteYue, line 79: I've tested this pr with one hdfs resource and found out that this DFSFileSystem is not initialized correctly. The `operations` filed would be null if you only create one DFSFileSystem with `new DFSFileSystem`. Then you'll get
```
java.lang.NullPointerException: Cannot invoke "org.apache.doris.fs.operations.HDFSFileOperations.openWriter(org.apache.doris.fs.operations.OpParams)" because "this.operations" is null
        at org.apache.doris.fs.remote.dfs.DFSFileSystem.directUpload(DFSFileSystem.java:276) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.catalog.HdfsResource.pingHdfs(HdfsResource.java:83) ~[doris-fe.jar:1.2-SNAPSHOT]
```

You might consider using the `nativeFileSystem(String remotePath)` function to get one valid DFSFileSystem instance.

ByteYue (changes requested):

fe/fe-core/src/main/java/org/apache/doris/catalog/HdfsResource.java
ByteYue: The best practice would be to create one FIleSystem from the FileSystemFactory.java factory class instead of directly just new one fs.

felixwluo:
okay I'll adjust it and test it.

AlexYue ***@***.***>于2024年3月18日 周一22:22写道：

felixwluo:
run buildall

dataroaring:
We should add a regression test? @ByteYue Please provide hdfs config in pipeline.

PR #32424, opened 2024-03-18
[fix](cloud) Fix missing disk and cpu report in cloud mode

gavinchou:
Missing CPU reports may result in an incorrect PipelineExecutorSize (fixed to 1), affecting execution performance.

Reference code path of FE
```
getMinPipelineExecutorSize -> Backend.getPipelineExecutorSize 
public boolean updateCpuInfo(int cpuCores, int pipelineExecutorSize)
private static void cpuReport(long backendId, int cpuCores, int pipelineExecutorSize)
```


dataroaring (approved):
LGTM

deardeng (approved):
LGTM

PR #32285, opened 2024-03-15
[fix](s3 client)add default ca cert list for s3 client to avoid problem:'curlCode:77'

ryanzryu:
## Proposed changes

Issue Number: #23163

add a be config field: ca_cert_file_paths , this is a default ca cart file list for different OS
And selects a valid file for initialization of the s3 client
if All ca cert file path is invalid ，s3 client will no set this config and use s3‘s default config
## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #32485, opened 2024-03-19
[fix](s3 client)add default ca cert list for s3 client to avoid problem:'curlCode:77' #32285

xiaokang:
cherry pick from #32285


xiaokang:
run buildall

Issue #31517, opened 2024-02-28
[Bug] cast date to decimal get wrong result


zxealous:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

### Version

2.0.5 && master

### What's Wrong?

case:
 set experimental_enable_nereids_planner = true;
select cast(a as decimal) from (select cast("2000-01-01" as date) a) ta

wrong result on 2.0.4 && master
```
MySQL [(none)]> select cast(a as decimal) from (select cast("2000-01-01" as date) a) ta;
+----------------------------+
| cast(a as DECIMALV3(9, 0)) |
+----------------------------+
|                    1024033 |
+----------------------------+
1 row in set (0.066 sec)
```

right result:
```
mysql> select cast(a as decimal) from (select cast("2000-01-01" as date) a) ta;
+--------------------+
| cast(a as decimal) |
+--------------------+
|           20000101 |
+--------------------+
1 row in set (0.00 sec)
```

### What You Expected?

work well

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

AcKing-Sam:
Hi @Yukang-Lian , I want to try this issue, could you please assign it to me?

Yukang-Lian:
> Hi @Yukang-Lian , I want to try this issue, could you please assign it to me? I tried this demo on v2.0.5, and this bug still exists.

Done!

yongjinhou:
> Hi @Yukang-Lian , I want to try this issue, could you please assign it to me? I tried this demo on v2.0.5, and this bug still exists.

请问现在修复的怎么样了？

AcKing-Sam:
> > Hi @Yukang-Lian , I want to try this issue, could you please assign it to me? I tried this demo on v2.0.5, and this bug still exists.
> 
> 请问现在修复的怎么样了？

你好，正在进行中。 ∶)

Issue #32234, opened 2024-03-14
[Bug] Bug summary for doris 2.0.4


yongjinhou:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

### Version

目前2.0.4版本全量失败case有258个，目前针对失败case进行分类，先将各个分类的case全部修复完成后，再跑一轮回归测试，查看case的修复情况。
case修复流程：
1. 先查看case在master上是否有问题，若master已修复，则找到master上修复case的pr，cherry-pick到社区2.0分支
2. 若master上也存在问题，则将case修复并合入到master上，再cherry-pick 到2.0分支。
(注：修复的case需要增加regression-test的p0 case中，这样后续这类case可以作为代码合入的准入)

注意去对应的 issue 下认领完 assign 下，防止重复投入人力。
[https://github.com/apache/doris/issues/31524](https://github.com/apache/doris/issues/31524) 已修复
[https://github.com/apache/doris/issues/31516](https://github.com/apache/doris/issues/31516) 未修复
[https://github.com/apache/doris/issues/31552](https://github.com/apache/doris/issues/31552) 未修复
[https://github.com/apache/doris/issues/31517](https://github.com/apache/doris/issues/31517) 未修复
[https://github.com/apache/doris/issues/31514](https://github.com/apache/doris/issues/31514) 已修复
[https://github.com/apache/doris/issues/31518](https://github.com/apache/doris/issues/31518) 已修复
[https://github.com/apache/doris/issues/31938](https://github.com/apache/doris/issues/31938) 未修复
[https://github.com/apache/doris/issues/31957](https://github.com/apache/doris/issues/31957) 未修复

### What's Wrong?

无

### What You Expected?

无

### Anything Else?

无

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

AcKing-Sam:
@yongjinhou  Hi! I find another [bug issue](https://github.com/apache/doris/issues/32354) and I tested it on `master`. It seems that the bug still exists. In addition, could you please assign it to me? Thanks!

AcKing-Sam:
Hi, I made a PR to [#32517](https://github.com/apache/doris/issues/31517).

zhiqiang-hhhh:
https://github.com/apache/doris/issues/32773

[...]

PR #32518, opened 2024-03-20
[fix] (vectorization) fix cast from date/datetime to decimal

AcKing-Sam:
## Proposed changes

Issue Number: [#31517](https://github.com/apache/doris/issues/31517)
**[Bug] cast date to decimal get wrong result**

Below case got the wrong result:
```mysql
set experimental_enable_nereids_planner = true;
select cast(a as decimal) from (select cast("2000-01-01" as date) a) ta;

mysql> select cast(a as decimal) from (select cast("2000-01-01" as date) a) ta;
+----------------------------+
| cast(a as DECIMALV3(9, 0)) |
+----------------------------+
|                    1024033 |
+----------------------------+
1 row in set (0.066 sec)
```
After fix:
```mysql
mysql> select cast(a as decimal) from (select cast("2000-01-01" as date) a) ta;
+----------------------------+
| cast(a as DECIMALV3(9, 0)) |
+----------------------------+
|                   20000101 |
+----------------------------+
1 row in set (0.02 sec)

```

## Further comments

For this bug, I read the source code in `be/src/vec/functions/function_cast.h`:
```cpp
if constexpr (IsDataTypeDecimal<FromDataType> || IsDataTypeDecimal<ToDataType>) {
// ...
} else if constexpr (IsTimeType<FromDataType>) {
// ...
} else if constexpr (IsTimeV2Type<FromDataType>) {
// ...
} else {
// ...
}
``` 
From the if-condition branch, it is clear that `cast from date to decimal` will step into the first branch(from or to decimal), and the under-lying cast for date to uint is missed. This is the reason for this bug.

Therefore, I try to add cast date to int in the first branch.

Then, I read the source code in `be/src/vec/data_types/data_type_decimal.h`. I found the function `convert_to_decimals` is **only for casting between other integral types and decimals**. So I add the cast in `convert_to_decimal` before calling `convert_to_decimals`.

However, I think there must be some potential bugs when casting between date/datetime and decimal, which is a future work.


AcKing-Sam:
run buildall

AcKing-Sam:
run buildall

AcKing-Sam:
run buildall

AcKing-Sam:
run buildall

be/src/vec/data_types/data_type_decimal.h
```diff
@@ -655,6 +656,10 @@ void convert_to_decimal(typename ToDataType::FieldType* dst,
         for (size_t i = 0; i < size; ++i) {
             dst[i].value = FromFieldType(src[i] * multiplier.value + ((src[i] >= 0) ? 0.5 : -0.5));
         }
+    } else if constexpr (IsDateV2Type<FromDataType>) {
```
yiguolei, line 659: please add regression test

AcKing-Sam:
run buildall

be/src/vec/data_types/data_type_decimal.h
```diff
@@ -655,6 +656,10 @@ void convert_to_decimal(typename ToDataType::FieldType* dst,
...
+        for (size_t i = 0; i < size; ++i) {
+            dst[i].value = reinterpret_cast<const DateV2Value<DateV2ValueType>&>(src[i]).to_int64();
```
zhiqiang-hhhh, line 661: 1. maybe `assert_cast<const DateV2Value<FromFieldType>&>` is better
2. why convert result to int64? what if target Decimal is Decimal32 or Decimal128?

AcKing-Sam:
run buildall

zhiqiang-hhhh (approved):
LGTM

zhiqiang-hhhh:
@AcKing-Sam Do we have same issue for Datetime?

AcKing-Sam:
@zhiqiang-hhhh  
Hi, I just tested below in Doris:
```sql
mysql> select cast(c as decimal) from (select cast("9999-12-31" as datetime) c) t3;
ERROR 1105 (HY000): errCode = 2, detailMessage = (127.0.0.1)[E-124]Arithmetic overflow, convert failed from 703674110198153216, expected data is [-999999999, 999999999]
```
For MySQL, the result is:
```sql
+--------------------+
| cast(c as decimal) |
+--------------------+
|         9999999999 |
+--------------------+
```
I think for datetime, current version is okay because of the strict overflow detection in Doris.
I picked this issue from [Bug summary for doris 2.0.4](https://github.com/apache/doris/issues/32234).

zhiqiang-hhhh:
> @zhiqiang-hhhh Hi, I just tested below in Doris:
> 
> ```sql
> mysql> select cast(c as decimal) from (select cast("9999-12-31" as datetime) c) t3;
> ERROR 1105 (HY000): errCode = 2, detailMessage = (127.0.0.1)[E-124]Arithmetic overflow, convert failed from 703674110198153216, expected data is [-999999999, 999999999]
> ```
> 
> For MySQL, the result is:
> 
> ```sql
> +--------------------+
> | cast(c as decimal) |
> +--------------------+
> |         9999999999 |
> +--------------------+
> ```
> 
> I think for datetime, current version is okay because of the strict overflow detection in Doris. I picked this issue from [Bug summary for doris 2.0.4](https://github.com/apache/doris/issues/32234).

This is because we have incorrect integer value of Datetime. 

```sql
select cast(c as decimal(30,9)) from (select cast("9999-12-31" as datetime) c) t3
--------------

+------------------------------+
| cast(c as DECIMALV3(30, 9))  |
+------------------------------+
| 703674110198153216.000000000 |
+------------------------------+
1 row in set (0.02 sec)
```
So we need a similar fix for datetime too.

AcKing-Sam:
@zhiqiang-hhhh  Got it, I am willing to fix it! :)

zhiqiang-hhhh:
@AcKing-Sam btw, maybe we should modify logical in `convert_to_decimals` instead of using a temporary array which costs more memory and does not have performance benefits. For example, add check and do cast to int64 if from type is Date/Datetime.

Do you think which one is better?

AcKing-Sam:
> @AcKing-Sam btw, maybe we should modify logical in `convert_to_decimals` instead of using a temporary array which costs more memory and does not have performance benefits. For example, add check and do cast to int64 if from type is Date/Datetime.
> 
> Do you think which one is better?

Yes, I think the second way is better and I will try to change the committed code.
The code should be like:
```cpp
else if (type is date or datetime) {
// instead of using array: typename FromDataType::FieldType tmp[size];
// directly check here, may reuse some code in convert_to_decimals function
// if okay, directly change to int64
// else take an error
}
```

regression-test/suites/query_p0/cast/test_cast_from_date_to_decimal.groovy
```diff
@@ -0,0 +1,33 @@
...
+
+    test {
+        sql """select cast(b as decimal(7, 0)) from (select cast("1000-01-01" as date) b) t2;"""
+        exception "Arithmetic overflow, convert failed from 10000101, expected data is [-9999999, 9999999]"
+    }
+
+    qt_sql """select cast(c as decimal(8, 0)) from (select cast("9999-12-31" as date) c) t3;"""
+
+    qt_sql """select cast(c as decimal(38, 0)) from (select cast("9999-12-31" as date) c) t3;"""
+}
```
yiguolei, line 33: add some test cases such as:
1. 9999-12-31 23:59:59.999999
2. 0000-00-01
3. 0000-00-00
4. 1999-12-31 13:59:59.999
5. 2000-02-29 11:11:11
6. 2001-02-29 11:11:11

yiguolei:
@AcKing-Sam We should add some test cases for datetime with precision value.

AcKing-Sam:
@yiguolei  Got it, and thanks for the review. I am going to add the cast from datetime to decimal and add the cases. ;)

AcKing-Sam:
run buildall

AcKing-Sam:
@yiguolei  @zhiqiang-hhhh 

Hi, I added the cast from datetime to decimal. Also, I modified the logic in `cast_to_decimals()`.
However, there is an exception, to test this SQL:
`select cast(a as decimal(30, 9)) from (select cast("9999-12-31 23:59:59.999" as datetime) a) t1;`

In Doris, it directly get the exception:
`Conversion from Nullable(DateTimeV2) to Decimal(30, 9) is not supported.`

It looks like the cast from string to datetime cannot make sure the accuracy. Due to this reason, this sql tries to cast a `NULL` to decimal.
I think the continuous problem (**accuracy from string to datetime**) can be handled in another issue.

zhiqiang-hhhh:
> @yiguolei @zhiqiang-hhhh
> 
> Hi, I added the cast from datetime to decimal. Also, I modified the logic in `cast_to_decimals()`. However, there is an exception, to test this SQL: `select cast(a as decimal(30, 9)) from (select cast("9999-12-31 23:59:59.999" as datetime) a) t1;`
> 
> In Doris, it directly get the exception: `Conversion from Nullable(DateTimeV2) to Decimal(30, 9) is not supported.`
> 
> It looks like the cast from string to datetime cannot make sure the accuracy. Due to this reason, this sql tries to cast a `NULL` to decimal. I think the continuous problem (**accuracy from string to datetime**) can be handled in another issue.

You need to specify the scale of micro-second part.

```sql
select cast("9999-12-31 23:59:59.999" as datetime(3))
--------------

+--------------------------------------------------+
| cast('9999-12-31 23:59:59.999' as DATETIMEV2(3)) |
+--------------------------------------------------+
| 9999-12-31 23:59:59.999                          |
+--------------------------------------------------+
1 row in set (0.01 sec)
```

zhiqiang-hhhh:
@AcKing-Sam 

```sql
mysql [demo]>select *, cast(str as Datetime(3)) from str;
--------------
select *, cast(str as Datetime(3)) from str
--------------

+------+-------------------------+----------------------------+
| rid  | str                     | cast(str as DATETIMEV2(3)) |
+------+-------------------------+----------------------------+
|    1 | 9999-12-31 23:59:59.999 | 9999-12-31 23:59:59.999    |
+------+-------------------------+----------------------------+
1 row in set (0.09 sec)

mysql [demo]>select *, cast(str as Datetime) from str;
--------------
select *, cast(str as Datetime) from str
--------------

+------+-------------------------+----------------------------+
| rid  | str                     | cast(str as DATETIMEV2(0)) |
+------+-------------------------+----------------------------+
|    1 | 9999-12-31 23:59:59.999 | NULL                       |
+------+-------------------------+----------------------------+
1 row in set (0.03 sec)
```

AcKing-Sam:
@zhiqiang-hhhh 

Hi, I wonder what is the correct answer of this sql:
`select cast(a as decimal(30, 4)) from (select cast("9999-12-31 23:59:59.999999" as datetime(6)) a) t1;`

For MYSQL, the answer is:
```sql
+---------------------------+
| cast(a as decimal(30, 4)) |
+---------------------------+
|       99991231235960.0000 |
+---------------------------+
```
For my implementation, it is:
```sql
+-----------------------------+
| cast(a as DECIMALV3(30, 4)) |
+-----------------------------+
|         99991231235959.9999 |
+-----------------------------+
1 row in set (0.02 sec)
```

AcKing-Sam:
run buildall

zhiqiang-hhhh:
> @zhiqiang-hhhh
> 
> Hi, I wonder what is the correct answer of this sql: `select cast(a as decimal(30, 4)) from (select cast("9999-12-31 23:59:59.999999" as datetime(6)) a) t1;`
> 
> For MYSQL, the answer is:
> 
> ```sql
> +---------------------------+
> | cast(a as decimal(30, 4)) |
> +---------------------------+
> |       99991231235960.0000 |
> +---------------------------+
> ```
> 
> For my implementation, it is:
> 
> ```sql
> +-----------------------------+
> | cast(a as DECIMALV3(30, 4)) |
> +-----------------------------+
> |         99991231235959.9999 |
> +-----------------------------+
> 1 row in set (0.02 sec)
> ```

There is something strange about behaviour of mysql, it is rounding but finished halfway, and doris is just doing truncate.
I just tests pg and sql server ,they all have different behaviour. 

I think we can just keep doing truncate, but we need more conner test to make sure that we will do truncate in all situations like above.

be/src/vec/data_types/data_type_decimal.h
```diff
@@ -411,39 +412,86 @@ constexpr bool IsDataTypeDecimalOrNumber =
...
 // only for casting between other integral types and decimals
+// in addition, this function also handles cast from data/datetime to decimals
 template <typename FromDataType, typename ToDataType, bool multiply_may_overflow,
           bool narrow_integral, typename RealFrom, typename RealTo>
     requires IsDataTypeDecimal<FromDataType> && IsDataTypeDecimal<ToDataType>
 void convert_to_decimals(RealTo* dst, const RealFrom* src, UInt32 scale_from, UInt32 scale_to,
                          const typename ToDataType::FieldType& min_result,
-                         const typename ToDataType::FieldType& max_result, size_t size) {
+                         const typename ToDataType::FieldType& max_result, size_t size,
+                         bool is_from_date, bool is_from_datetime) {
```
zhiqiang-hhhh, line 422: we should check the source data type by deducting from template argument.

AcKing-Sam:
@zhiqiang-hhhh Got it. Thanks for your review and I will add more regression tests to it. ;)

be/src/vec/data_types/data_type_decimal.h
```diff
@@ -411,39 +412,86 @@ constexpr bool IsDataTypeDecimalOrNumber =
         IsDataTypeDecimal<DataType> || IsDataTypeNumber<DataType>;
 
 // only for casting between other integral types and decimals
+// in addition, this function also handles cast from data/datetime to decimals
 template <typename FromDataType, typename ToDataType, bool multiply_may_overflow,
           bool narrow_integral, typename RealFrom, typename RealTo>
     requires IsDataTypeDecimal<FromDataType> && IsDataTypeDecimal<ToDataType>
```
zhiqiang-hhhh, line 418: `requires (IsDataTypeDecimal || IsDateTimeV2Type || IsDateV2Type) && IsDataTypeDecimal<ToDataType>`

be/src/vec/data_types/data_type_decimal.h
```diff
@@ -662,7 +710,8 @@ void convert_to_decimal(typename ToDataType::FieldType* dst,
                                                       Decimal256, Decimal64>>;
         convert_to_decimals<DataTypeDecimal<DecimalFrom>, ToDataType, multiply_may_overflow,
                             narrow_integral>(dst, src, from_scale, to_scale, min_result, max_result,
-                                             size);
+                                             size, IsDateV2Type<FromDataType>,
+                                             IsDateTimeV2Type<FromDataType>);
```
zhiqiang-hhhh, line 714: IsDateTimeV2Type<FromDataType> should be move to function `convert_to_decimals`, and it will be something like
`convert_to_decimals(...) {
if constexpr (IsDateTimeV2Type<>) {
} else if constexpr (IsDateV2Type) {
} else {
   ...
}
}
`

zhiqiang-hhhh (changes requested):

AcKing-Sam:
run buildall

AcKing-Sam:
@zhiqiang-hhhh  Hi, I have added the deduction from template arguments. However, the bot said the function is a read-hard. Could you please give me some hints about how to optimize the code? Thanks a lot!

zhiqiang-hhhh:
> @zhiqiang-hhhh Hi, I have added the deduction from template arguments. However, the bot said the function is a read-hard. Could you please give me some hints about how to optimize the code? Thanks a lot!

@AcKing-Sam hello, just leave the bot alone

AcKing-Sam:
run buildall

PR #32387, opened 2024-03-18
[FIX](array_agg) fix array agg with other agg function

amorynan:
## Proposed changes
if we use array_agg func after other agg func like count, sum, here may make be core with core stack below
```
*** is nereids: 1 ***
*** tablet id: 0 ***
*** Aborted at 1710744527 (unix time) try "date -d @1710744527" if you are using GNU date ***
*** Current BE git commitID: b1cae7cc00 ***
*** SIGSEGV address not mapped to object (@0x1) received by PID 3108681 (TID 3127071 OR 0x7fd0ba580700) from PID 1; stack trace: ***
 0# doris::signal::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*) at /mnt/disk2/wangqiannan/amory/master/doris/be/src/common/signal_handler.h:421
 1# PosixSignals::chained_handler(int, siginfo*, void*) [clone .part.0] in /mnt/disk2/wangqiannan/tool/jdk-17.0.10/lib/server/libjvm.so
 2# JVM_handle_linux_signal in /mnt/disk2/wangqiannan/tool/jdk-17.0.10/lib/server/libjvm.so
 3# 0x00007FE05D36CB50 in /lib64/libc.so.6
 4# void doris::vectorized::PODArray<unsigned char, 4096ul, Allocator<false, false, false>, 15ul, 16ul>::push_back<unsigned char const&>(unsigned char const&) at /mnt/disk2/wangqiannan/amory/master/doris/be/src/vec/common/pod_array.h:385
 5# doris::vectorized::AggregateFunctionArrayAggData<doris::StringRef>::deserialize_and_merge(doris::vectorized::IColumn const&, unsigned long) at /mnt/disk2/wangqiannan/amory/master/doris/be/src/vec/aggregate_functions/aggregate_function_collect.h:398
 6# doris::vectorized::AggregateFunctionCollect<doris::vectorized::AggregateFunctionArrayAggData<doris::StringRef>, std::integral_constant<bool, false>, std::integral_constant<bool, true> >::deserialize_and_merge_vec(char* const*, unsigned long, char*, doris::vectorized::ColumnString const*, doris::vectorized::Arena*, unsigned long) const at /mnt/disk2/wangqiannan/amory/master/doris/be/src/vec/aggregate_functions/aggregate_function_collect.h:558
[...]

PR #32583, opened 2024-03-21
[compile](fix) fix gcc compile error

shuke987:
## Proposed changes
fcc reports error: comparing the result of pointer addition when compile
`
if (places[i] + offset)
`

Issue Number: close #xxx

## Further comments

### DIFF
--- a/be/src/vec/aggregate_functions/aggregate_function_collect.h
+++ b/be/src/vec/aggregate_functions/aggregate_function_collect.h
@@ -678,7 +678,7 @@ public:
                                             Arena* arena, const size_t num_rows) const override {
         if constexpr (ShowNull::value) {
             for (size_t i = 0; i != num_rows; ++i) {
-                if (static_cast<void*>(places[i] + offset) != nullptr) {
+                if (places[i] || offset) {
                     this->data(places[i] + offset)
                             .deserialize_and_merge(*assert_cast<const IColumn*>(column), i);
                 }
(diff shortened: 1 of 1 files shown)


shuke987:
run buildall

shuke987:
run buildall

shuke987:
run performance

be/src/vec/aggregate_functions/aggregate_function_collect.h
```diff
@@ -678,7 +678,7 @@ class AggregateFunctionCollect
                                             Arena* arena, const size_t num_rows) const override {
         if constexpr (ShowNull::value) {
             for (size_t i = 0; i != num_rows; ++i) {
-                if (places[i] + offset) {
+                if (places[i] || offset) {
```
cambyzju, line 681: We should not change the logic while fix compile error. 

offset is added from pr https://github.com/apache/doris/pull/32387, please have a review @amorynan

shuke987: place[i] is a pointer, and offset is an unsigned int, so if placces[i] != null or offset != 0, equals places[i] + offset != null

cambyzju: maybe the correct answer is like: `if(places[i] && offset < places[i].size())`, check `offset != 0 or == 0` is strange here.

zclllyybb: here below we use `places[i]+offset` as an pointer. so maybe `places[i] + offset != nullptr` is correct and more readable

PR #32614, opened 2024-03-21
[opt](pointer cast) better typeid_cast

zhiqiang-hhhh:


zclllyybb (approved):
LGTM

be/src/vec/common/typeid_cast.h
```diff
@@ -43,34 +43,20 @@
...
-        if (typeid(*from) == typeid(std::remove_pointer_t<To>)) {
-            return static_cast<To>(from);
-        }
-    } catch (const std::exception& e) {
-        throw doris::Exception(doris::ErrorCode::BAD_CAST, e.what());
-    }
-#else
-    if (typeid(*from) == typeid(std::remove_pointer_t<To>)) {
+    requires std::is_pointer_v<To>
+To typeid_cast(From* from) noexcept {
```
yiguolei, line 56: In debug mode, the previous code will print exception. It is useful to help us debug.

be/src/vec/common/typeid_cast.h
```diff
@@ -43,34 +43,20 @@
...
+    if ((typeid(From) == typeid(std::remove_pointer_t<To>)) ||
+        (from && typeid(*from) == typeid(std::remove_pointer_t<To>)))
         return static_cast<To>(from);
```
yiguolei, line 59: maybe we could always throw exception, not return nullptr.

zhiqiang-hhhh: > maybe we could always throw exception, not return nullptr.

https://github.com/apache/doris/blob/2c017399fbd109bb96ae09373f8dc46d8306e8a7/be/src/vec/columns/column_array.cpp#L513

the users of `typeid_cast` sometimes relay on the nullptr as their control flow, exception has bad performance when it is used as control flow.

be/src/vec/common/typeid_cast.h
zhiqiang-hhhh: I dont know whether `typeid_cast(nullptr)` is designed as a common use scenario

be/src/vec/common/typeid_cast.h
zclllyybb: 不是的，我们抛异常是在入参为空指针的时候才会抛。以前的行为，如果typeid不同，就是返回空指针。因为现在这个新写法应该不会带来更多的风险，应该是靠谱的。空指针的话肯定是代码有错才会出，这种情况下前、后的使用肯定会出问题，不必在这里处理成异常的。

PR #31610, opened 2024-02-29
[Refactor](insert) refactor insert command to support other type of table

morningman:
## Proposed changes

issue: #31442

The previous InsertIntoTableCommand and InsertExecutor only support olap table.
This PR refactor them so that we can support other type of target table in future.
**No logic is changed in this PR**

This PR mainly changes:

1. Move all insert related classes under `org.apache.doris.nereids.trees.plans.commands` to `org.apache.doris.nereids.trees.plans.commands.insert`
2. Extract some abstract classes:

	- `AbstractInsertExecutor`

		This is the parent class of Insert Executor. The common logic is as follow:

		```
		public void executeSingleInsert(StmtExecutor executor, long jobId) {
	        beforeExec();
	        try {
	            execImpl(executor, jobId);
	            onComplete();
	        } catch (Throwable t) {
	            onFail(t);
	        } finally {
	            executor.updateProfile(true);
	            QeProcessorImpl.INSTANCE.unregisterQuery(ctx.queryId());
	        }
	        afterExec(executor);
	    }
		```

		The derived classes should implement all abstract methods:

		1. beginTransaction();
		2. finalizeSink();
		3. beforeExec();
		4. onComplete();
		5. onFail();
		6. afterExec();

		All methods related to certain target table type should be in derived class.
		You can refer to `OlapInsertExecutor` for example.

		If you need add a new type of target table for insert operation.
		You need implement a `XXXInsertExecutor extends AbstractInsertExecutor`
		and implements the abstract methods.

	-  `InsertCommandContext`

		A class to store some parameters or flags of an insert opertion.
		This is an abstract class, you can derived it to add some fields or methods you need.
		See `OlapInsertCommandContext` for example

	- `InsertUtils`

		A helper class. Move all static method previous in `InsertExecutor` to this class.

3. The FE config `using_old_load_usage_pattern` is deprecated, remove it.

4. All `lock` related method in `ExternalTable` is meaningless, remove them

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #32665, opened 2024-03-22
[Fix](Test)Reduce the Sleep time and ensure that the corresponding thread resources are released after the case is executed.

CalvinKirs:
- When we cancel task , if the job sql contains the sleep function, it will sleep until the specified time even if it is interrupted. Therefore, this causes the https://github.com/apache/doris/pull/32534 case to fail.
- https://github.com/apache/doris/pull/31610 insert commond not set jobid


Gabriel39:
run buildall

CalvinKirs:
run buildall

morningman (approved):

zy-kkk (approved):

PR #32700, opened 2024-03-22
[Fix](TransientTask)Export tasks should only be run on the master node

CalvinKirs:
- Add thread name

- Export Task runs only on the master node, so it is necessary to explicitly start the corresponding resources. At the same time, refactor some code to avoid circular dependencies.
- missing log info (should use Log4j2)


CalvinKirs:
run buildall

morningman (approved):
LGTM

CalvinKirs:
run buildall

yiguolei (approved):

Issue #31892, opened 2024-03-06
[Bug] Error occurred while passing status up


wyxxxcat:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

### Version

master

### Error Information：
![alt text](https://krseoul.imgtbl.com/i/2024/03/06/65e824f4770f7.png)

### Error location
The status returned by the seek_to_position_in_page function needs to be passed to the upper layer by _seek_to_pos_in_page
However, changing line 1085 static_cast&lt;void&gt; to RETURN_IF_ERROR will result in invalid pos in regression testing
![alt text](https://krseoul.imgtbl.com/i/2024/03/06/65e82525958e9.png)

### What You Expected?

![alt text](https://krseoul.imgtbl.com/i/2024/03/06/65e863d6ab052.png)

### How to Reproduce?

First step:
change static_cast &lt;void&gt; to RETURN_IF_ERROR,and change the return value of _seek_to_pos_in_page to Status, add return Status::OK() to the end of the function, and then apply the macro RETURN_IF_ERROR wherever _seek_to_pos_in_page is called
Second step:
Execute sql
```
CREATE TABLE IF NOT EXISTS tbl_scalar_types_dup (
    `k1` bigint(11) NULL,
    `c_bool` boolean NULL,
    `c_tinyint` tinyint(4) NULL,
    `c_smallint` smallint(6) NULL,
    `c_int` int(11) NULL,
    `c_bigint` bigint(20) NULL,
    `c_largeint` largeint(40) NULL,
    `c_float` float NULL,
    `c_double` double NULL,
    `c_decimal` decimal(20, 3) NULL,
    `c_decimalv3` decimalv3(20, 3) NULL,
    `c_date` datev2 NULL,
    `c_datetime` datetimev2 NULL,
    `c_datev2` datev2 NULL,
    `c_datetimev2` datetimev2(0) NULL,
    `c_char` char(15) NULL,
    `c_varchar` varchar(100) NULL,
    `c_string` text NULL
) ENGINE=OLAP
DUPLICATE KEY(`k1`)
COMMENT 'OLAP'
DISTRIBUTED BY HASH(`k1`) BUCKETS 10
PROPERTIES("replication_num" = "1", "store_row_column" = "true");

INSERT INTO tbl_scalar_types_dup
[...]

PR #32725, opened 2024-03-23
[Bug] Error occurred while passing status up

wyxxxcat:
## Proposed changes

Issue Number: #31892

## Further comments

When querying a table with only NULL values, the function seek_to_position_in_page will misjudge and pass status upward, as shown in Figure, where data is not empty, but num_element is 0


wyxxxcat:
run buildall

zclllyybb:
run buildall

wyxxxcat:
run buildall

PR #28479, opened 2023-12-15
[fix](proxy) Fix getProxy frequently parse hostname to ip

xinyiZzz:
## Proposed changes

Each RPC between Doris FE and BE will call getProxy. Each getProxy will parse hostname and check whether the IP has changed.

When the hostname resolution server is unstable, frequent parse hostname will often fail and cause performance problems.

Therefore, only check whether the backend IP changes during heartbeat, and if so, recreate the rpc client.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #32869, opened 2024-03-26
[opt](fqdn) Add DNS Cache for FE and BE

morningman:
## Proposed changes

In previously, when enabling FQDN, Doris will call dns resolver to get IP from hostname
each time when 1) FE gets BE's grpc client. 2) BE gets other BE's brpc client.
So when in high concurrency case, the dns resolver be overloaded and failed to resolve hostname.

This PR mainly changes:

1. Add DNSCache for both FE and BE.
    The DNSCache will run on every FE and BE node. It has a cache, key is hostname and value is IP.
    Caller can get IP by hostname from this cache, and if hostname does not exist, it will try to resolve it
    and update the cache.
    In addition, DNSCache has a daemon thread to refresh the cache every 1 min, in case that the IP may
    be changed at anytime.

There are other implements of this dns cache:

1.  https://github.com/kaka11chen/doris/commit/36fed139974ed52dfa61b656f3e4d64f56a4185a
    This is for BE side, but it does not handle the IP change case.

3. https://github.com/apache/doris/pull/28479
    This is for FE side, but it can only work with Master FE. Other FE node will not be aware of the IP change.
    And there are a bunch of BackendServiceProxy, this PR only handle cache in one of them.

## Further comments

### DIFF
--- a/be/src/util/dns_cache.cpp
+++ b/be/src/util/dns_cache.cpp
@@ -1,0 +1,84 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include "util/dns_cache.h"
+
+#include "service/backend_options.h"
+#include "util/network_util.h"
+
+namespace doris {
+
+DNSCache::DNSCache() {
+    refresh_thread = std::thread(&DNSCache::_refresh_cache, this);
+    refresh_thread.detach();
+}
+
+DNSCache::~DNSCache() {
+    stop_refresh = true;
+    if (refresh_thread.joinable()) {
+        refresh_thread.join();
+    }
+}
+
+Status DNSCache::get(const std::string& hostname, std::string* ip) {
+    {
+        std::shared_lock<std::shared_mutex> lock(mutex);
+        auto it = cache.find(hostname);
+        if (it != cache.end()) {
+            *ip = it->second;
+            return Status::OK();
+        }
+    }
+    // Update if not found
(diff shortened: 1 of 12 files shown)


morningman:
run buildall

morningman:
run buildall

morningman:
run buildall

xinyiZzz (approved):
LGTM

kaka11chen (approved):
LGTM

be/src/util/dns_cache.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include "util/dns_cache.h"
+
+#include "service/backend_options.h"
+#include "util/network_util.h"
+
+namespace doris {
+
+DNSCache::DNSCache() {
+    refresh_thread = std::thread(&DNSCache::_refresh_cache, this);
+    refresh_thread.detach();
```
zhiqiang-hhhh, line 27: the detached thread can not be join by main thread

https://stackoverflow.com/questions/39174761/pthreads-can-i-detach-from-a-thread-and-then-join-in-main

morningman: I will fix it

be/src/util/dns_cache.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+DNSCache::~DNSCache() {
+    stop_refresh = true;
+    if (refresh_thread.joinable()) {
```
zhiqiang-hhhh, line 32: not joinable, so DNECache will be destoried immediately

be/src/util/dns_cache.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+void DNSCache::_refresh_cache() {
+    while (!stop_refresh) {
+        // refresh every 1 min
+        std::this_thread::sleep_for(std::chrono::minutes(1));
+        std::unordered_set<std::string> keys;
+        {
+            std::shared_lock<std::shared_mutex> lock(mutex);
+            std::transform(cache.begin(), cache.end(), std::inserter(keys, keys.end()),
```
zhiqiang-hhhh, line 74: it is very likely that object cache has already been destroyed, if main thread has already exited with --grace enabled.

be/src/util/dns_cache.cpp
```diff
@@ -0,0 +1,84 @@
...
+        cache[hostname] = real_ip;
+        LOG(INFO) << "update hostname " << hostname << "'s ip to " << real_ip;
+    }
+    return Status::OK();
+}
+
+void DNSCache::_refresh_cache() {
+    while (!stop_refresh) {
+        // refresh every 1 min
+        std::this_thread::sleep_for(std::chrono::minutes(1));
```
zhiqiang-hhhh, line 70: btw, after `_refresh_cache` thread is joinable, when we stop BE with grace enabled, we will wait at least 1 min ...

PR #25138, opened 2023-10-08
[feature](hive)Support hive tables after alter type.

hubgeter:
## Proposed changes

Issue Number: close #xxx
1.Reconstruct the logic of decode to read parquet. The parquet  reader first reads the data according to the parquet physical type, and then performs a type conversion.

2.Support hive alter table.

|    to\from     |  int     |smallint|tinyint|bigint|float|double|boolean|string|char|varchar|date|timestamp|decimal|
|---------|----------|--------|-------|------|-----|----------------|-------|------|----|-------|----|---------|-------|
|int      | &#10004; |&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;      |
|smallint | &#10004; |&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;
|tinyint  | &#10004; |&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;
|float    | &#10004; |&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;
|bigint   | &#10004; |&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;
|double   | &#10004; |&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;
|boolean  |  ||||||&#10004;
|string   | &#10004; |&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|
|char   | &#10004; |&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|
|varchar   | &#10004; |&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|&#10004;|
|date     |          | |||||||||&#10004;|
|timestamp|          |||||||||||&#10004;|
|decimal  |&#10004;|&#10004;|&#10004;|&#10004;|||||||||&#10004;|

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #32873, opened 2024-03-27
[feature](schema change) unified schema change for parquet and orc reader

AshinGau:
## Proposed changes

Following https://github.com/apache/doris/pull/25138, unified schema change interface for parquet and orc reader, and can be applied to other format readers as well.

## Supported Type Changes
More type changes are supported:
image

## ColumnTypeConverter
Unified schema change interface for all format readers:
- First, read the data according to the column type of the file into source column;
- Second, convert source column to the destination column with type planned by FE

## PhysicalToLogicalConverter
Convert parquet physical column to logical column
In parquet document(https://github.com/apache/parquet-format/blob/master/LogicalTypes.md),
Logical or converted type is the data type of column, physical type is the stored type of column chunk. eg, decimal type can be stored as INT32, INT64, BYTE_ARRAY, FIXED_LENGTH_BYTE_ARRAY, so there is a convert process from physical type to logical type. In addition, schema change will bring about a change in logical type.

In previous implementations, physical and logical conversion were mixed together, resulting in severe code complexity and bloating.
`PhysicalToLogicalConverter` strips away the conversion of logical type, and reuse `ColumnTypeConverter` to resolve schema change, allowing parquet reader to only focus on the conversion of physical types.

Therefore, tow layers converters are designed:
- First, read parquet data with the physical type
- Second, convert physical type to logical type
- Third, convert logical type to the final type planned by FE(schema change)

### Ultimate performance optimization:
1. If process of (First => Second) is consistent, eg. from BYTE_ARRAY to string, no additional copies and conversions will be introduced;
2. If process of (Second => Third) is consistent, eg. from decimal(12, 4) to decimal(8, 2), no additional copies and conversions will be introduced;
3. Null map is share among all processes, no additional copies and conversions will be introduced in null map;
4. Only create one physical column in physical conversion, and reused in each loop;
5. Only create one logical column in logical conversion, and reused in each loop;
6. FIXED_LENGTH_BYTE_ARRAY is read as ColumnUInt8 instead of ColumnString, so the underlying decoder has no process to decode string and use memory copy to read the data as a whole, and the conversion has no need to resolve the Offsets in ColumnString.

### Remaining Issues
1. Not support paimon timestamp type, because it has no logical or converted type, and is conflict with column type change from bigint to timestamp. Reference paimon issue: https://github.com/apache/paimon/pull/2739, 

## Further comments


be/src/vec/exec/format/column_type_convert.cpp
```diff
@@ -0,0 +1,330 @@
...
+    // change to string type
+    // example: decimal -> string
+    if (is_string_type(dst_primitive_type)) {
+        return _to_string_converter(src_type, dst_type);
+    }
+
+    // string type to other type
+    // example: string -> date
+    if (is_string_type(src_primitive_type)) {
+        return _cast_string_converter(src_type, dst_type);
```
morningman, line 305:
```suggestion
        return _from_string_converter(src_type, dst_type);
```

be/src/vec/data_types/data_type_struct.cpp
```diff
@@ -355,14 +355,7 @@ std::optional<size_t> DataTypeStruct::try_get_position_by_name(const String& nam
 }
 
 String DataTypeStruct::get_name_by_position(size_t i) const {
-    if (i == 0 || i > names.size()) {
```
morningman, line 358: Why removing this check?

be/src/vec/exec/format/orc/vorc_reader.cpp
```diff
@@ -532,6 +532,14 @@ std::vector<OrcPredicate> value_range_to_predicate(
         std::vector<orc::TypeKind>* unsupported_pushdown_types) {
     std::vector<OrcPredicate> predicates;
 
+    PrimitiveType src_type = OrcReader::convert_to_doris_type(type).type;
+    if (src_type != primitive_type) {
+        if (!(is_string_type(src_type) && is_string_type(primitive_type))) {
```
morningman, line 537: why?

be/src/vec/exec/format/parquet/parquet_column_convert.h
```diff
@@ -193,195 +106,189 @@ struct ConvertParams {
...
-    }
+/**
+ * Convert parquet physical column to logical column
+ * In parquet document(https://github.com/apache/parquet-format/blob/master/LogicalTypes.md),
+ * Logical or converted type is the data type of column, physical type is the stored type of column chunk.
+ * eg, decimal type can be stored as INT32, INT64, BYTE_ARRAY, FIXED_LENGTH_BYTE_ARRAY.
+ * So there is a convert process from physical type to logical type.
+ * In addition, Schema change will bring about a change in logical type.
+ *
+ * In previous implementations, physical and logical conversion were mixed together,
```
morningman, line 137: No need to add this in comment

be/src/vec/exec/format/column_type_convert.h
```diff
@@ -0,0 +1,539 @@
...
+    }
+};
+
+template <>
+struct SafeCastString<TYPE_LARGEINT> {
+    static bool safe_cast_string(
+            const char* startptr, const int buffer_size,
+            PrimitiveTypeTraits<TYPE_LARGEINT>::ColumnType::value_type* value) {
+        int64 cast_to_int = 0;
+        bool can_cast = safe_strto64(startptr, buffer_size, &cast_to_int);
```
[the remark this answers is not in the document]
AshinGau: Use `read_int_text_impl` instead.

be/src/vec/data_types/data_type_struct.cpp
AshinGau: This function is wrong, and there's no check like `Block::get_by_position`

be/src/vec/exec/format/orc/vorc_reader.cpp
AshinGau: Altering column type from `string` to `varchar` can still use push-down predicates.

AshinGau:
run buildall

morningman (dismissed):

AshinGau:
run buildall

morningman (dismissed):
LGTM

AshinGau:
run buildall

AshinGau:
run buildall

morningman (approved):

kaka11chen (approved):
LGTM

Issue #31442, opened 2024-02-27
[Feature] Support DDL/DML on external catalog


morningman:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

### Description

Before 2.0, Doris can only support query data from external catalog such as Hive, Hudi, etc.
Now we would like to support DDL/DML feature on external catalog

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

liugddx:
i want try

lexluo09:
I'd like to try

morningman:
We are working on the framework of this feature.
After the framework is stable, we'd like to invite more contributor to participate in developement.

wsjz:
https://github.com/apache/doris/pull/32274

PR #30198, opened 2024-01-22
[feature](multi-catalog)support hms catalog create and drop table/db

wsjz:
## Proposed changes

Issue #31442

1. rename old create/drop table to add/removeMemoryTable
2. add new create/drop table/db method
3. support hms catalog create/drop table/db

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #32898, opened 2024-03-27
[fix](iceberg) fix iceberg catalog bug and p2 test cases

morningman:
## Proposed changes

1. Fix iceberg catalog bug

    This PR #30198 change the logic of `IcebergHMSExternalCatalog.java`,
    to get locationUrl by calling hive metastore's `getCatalog()` method.
    But this method only exists in hive 3+. So it will fail if we using hive 2.x.

    I temporary remove this logic, because this logic is only used from iceberg table writing.
    Which is still under development. We will rethink this logic later.

2. Fix test cases

    Some of P2 test cases missed `order_qt`. And because the output format of the floating point
    type is changed, some result in `out` files need to be regenerated.

## Further comments

Issue #31442


morningman:
run buildall

AshinGau (approved):

wuwenchi (approved):

PR #32980, opened 2024-03-28
[improve](txn insert) txn insert support write to one table many times

mymeiyi:
### Purpose

We have supported `insert into select`(https://github.com/apache/doris/pull/31666), `update`(https://github.com/apache/doris/pull/33034) and `delete`(https://github.com/apache/doris/pull/33100) in transaction load.

But leave a problem that, one partition can only be written once in one transaction, because current transaction mechanism only support publish one version for one partition. This pr is to solve this problem.

In other words, this pr supports write to one table many times in one transaction like:
```
begin;
insert into t1 select ...;
insert into t1 select ...;
commit;
```

### Current implementation

* In Doris, one transaction is related to one txn_id

* BE use this txn_id to record the load info of partition_id, tablet, DeltaWriter... in `txn_manager`

* If writing to one partition twice in one txn, the above info in BE may be overwrited

* When FE commit the txn, it calcultes a new partition version, a version is related to a Rowset, but multiple loads in txn generate multiple Rowsets.

#### Introduce of sub_txn_id

To solve the above problem, the basic idea is to separate the txn_id in FE and BE. For multiple loads in one txn, we use sub_txn_id to distinguish the load for BE. 

One example: suppose table t has 2 partitions, p1 and p2. The current version of p1 is 3, p2 is 4.

|command|FE|BE|
|----------|--|--|
|BEGIN| mark txn is began| |
|insert into t partition(p1, p2)| 1. begin_txn -> txn_id;<br> 2. sub_txn_id1 = txn_id; | load use sub_txn_id1 |
|insert into t partition(p1)| generate sub_txn_id2 | load use sub_txn_id2 |
|insert into t partition(p1, p2)| generate sub_txn_id3 | load use sub_txn_id3 |
|COMMIT| commit_txn(calculate partition version):<br> * sub_txn_id1: p1(4), p2(5)<br> * sub_txn_id2: p1(5)<br> * sub_txn_id3: p1(6), p2(6)<br><br> publish_task:<br> * use sub_txn_id to submit publish version tasks to be | |

#### FE Meta

In addition, this pr change the storage of `TransationState` in bdbje to json format to make it compatible.

#### Isolation Level

Doris provides the `READ COMMITTED` isolation level. Please note the following:

* In a transaction, each statement reads the data that was committed at the time the statement began executing.
 
* In a transaction, each statement cannot read the modifications made by other statements within the same transaction. Please notice:

  1. For delete command, there are 2 implementations, one is delete condition, one is insert.
  
  2. If the delete condition is committed after the insert, the delete will work for the insert, for example:
  ```
  begin;
  insert into t select * from t1 where id = 1;
  delete from t where id = 1;
  commit;
  ```

  3. So we make the delete condition commit before the insert.

### User doc

https://github.com/apache/doris-website/pull/604


mymeiyi:
run buildall

mymeiyi:
run buildall

mymeiyi:
run buildall

fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
```diff
@@ -2290,4 +2480,177 @@ protected void updateMultiTableRunningTransactionTableIds(long transactionId, Li
...
+                        }
+
+                        publishResult = checkQuorumReplicas(transactionState, tableId, partition, tablet,
+                                loadRequiredReplicaNum, allowPublishOneSucc, newVersion, tabletSuccReplicas,
+                                tabletWriteFailedReplicas, tabletVersionFailedReplicas, publishResult, logs);
+                    }
+                }
+            }
+        }
+
```
dataroaring, line 2569: Here, we need assure that there is no hole at least for one replica.

dataroaring (commented):
one comment.

mymeiyi:
run buildall

mymeiyi:
run buildall

mymeiyi:
run buildall

mymeiyi:
run buildall

mymeiyi:
run buildall

mymeiyi:
run buildall

mymeiyi:
run buildall

mymeiyi:
run buildall

fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
mymeiyi: support it and add a ut `GlobalTransactionMgrTest#testCommitTransactionWithSubTxnAndReplicaFailed` check it.

mymeiyi:
run buildall

mymeiyi:
run buildall

regression-test/suites/insert_p0/txn_insert.groovy
```diff
@@ -328,19 +480,44 @@ suite("txn_insert") {
                 where txn_insert_dt4.id = txn_insert_dt2.id;
             """
             sql """
-                delete from txn_insert_dt2 where id = 1 or id = 5;
+                delete from txn_insert_dt2 where id = 1;
             """
             sql """
-                delete from txn_insert_dt5 partition(p_20000102) where id = 1 or id = 5;
+                delete from txn_insert_dt2 where id = 5;
+            """
```
dataroaring, line 487: If we do insert and delete on the same table, then there is a problem that, if the delete is a predicate and implemented via push handler, then the delete take effect on the insert result. while if the delete is in `delete from` format then the delete does not take effect on the insert result.

It is ambiguous for users.

dataroaring: BTW, delete sign also works as delete via push handler.

dataroaring: If a dba changes the replication num of a table between two insert to a table, then what happens.

dataroaring: Thread1 t2:      insert into table1
Thread2 t1: insert into table2 select table1, t3: insert into table2 select table1.

Is result of t2 visible to insert at t3?

dataroaring (commented):
A case ensure that group commit is disabled with begin commit.

dataroaring (commented):
Should we limit tables in a transaction? What if 10000 tables are written in a transaction?

fe/fe-core/src/main/java/org/apache/doris/transaction/TransactionEntry.java
```diff
@@ -163,39 +162,44 @@ public void setpLoadId(Types.PUniqueId pLoadId) {
         this.pLoadId = pLoadId;
     }
 
-    // Used for insert into select
-    public void beginTransaction(DatabaseIf database, TableIf table)
-            throws DdlException, BeginTransactionException, MetaNotFoundException, AnalysisException,
-            QuotaExceedException {
+    // Used for insert into select, return the sub_txn_id for this insert
+    public long beginTransaction(TableIf table) throws UserException {
         if (isTxnBegin()) {
```
dataroaring, line 167: Does isTxnBegin indicate there is already a insert into values? If so, maybe we should change name of isTxnBegin.

regression-test/suites/insert_p0/txn_insert.groovy
```diff
@@ -214,7 +228,145 @@ suite("txn_insert") {
...
+            try {
+                sql """ create table ${db2}.${table} like ${table} """
+                sql """ begin; """
+                sql """ insert into ${table} select * from ${table}_0; """
+                test {
+                    sql """ insert into $db2.${table} select * from ${table}_0; """
+                    exception """Transaction insert must be in the same database, expect db_id"""
+                }
+            } finally {
+                sql """rollback"""
```
dataroaring, line 276: Please add a case that does not finally rollback. We also should add a case for timeout without explicit rollback.

dataroaring (commented):
What if we change session var during a transaction, like timeout.

fe/fe-core/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -286,6 +303,33 @@ public boolean commitAndPublishTransaction(DatabaseIf db, List<Table> tableList,
         return dbTransactionMgr.waitForTransactionFinished(db, transactionId, publishTimeoutMillis);
     }
 
+    public boolean commitAndPublishTransaction(DatabaseIf db, long transactionId,
+            List<SubTransactionState> subTransactionStates, long timeoutMillis) throws UserException {
+        StopWatch stopWatch = new StopWatch();
+        stopWatch.start();
+        List<Table> tableList = subTransactionStates.stream().map(SubTransactionState::getTable).distinct()
```
dataroaring, line 310: Are table in tableList ordered? otherwise, deadlock may happen.

dataroaring (commented):
Add a parallel case, in which each instance writes to several 
 same tables continuously.

mymeiyi:
> A case ensure that group commit is disabled with begin commit.

done

mymeiyi:
> Add a parallel case, in which each instance writes to several same tables continuously.

done

mymeiyi:
run buildall

mymeiyi:
> What if we change session var during a transaction, like timeout.

When begin txn, calculate a timeout timestamp. So the changed timeout does not work.

regression-test/suites/insert_p0/txn_insert.groovy
mymeiyi: > Thread1 t2: insert into table1 Thread2 t1: insert into table2 select table1, t3: insert into table2 select table1.
> 
> Is result of t2 visible to insert at t3?

yes

mymeiyi:
run buildall

mymeiyi:
run buildall

regression-test/suites/insert_p0/txn_insert.groovy
mymeiyi: > If we do insert and delete on the same table, then there is a problem that, if the delete is a predicate and implemented via push handler, then the delete take effect on the insert result. while if the delete is in `delete from` format then the delete does not take effect on the insert result.
> 
> It is ambiguous for users.

For `delete from t1` and `insert t1` case, the delete does not work the next insert, see case 14.

mymeiyi:
run buildall

mymeiyi:
run buildall

mymeiyi:
run buildall

mymeiyi:
run buildall

mymeiyi:
run feut

mymeiyi:
run  buildall

yujun777 (commented):
need consider first time upgrade fe with old format history txns

fe/fe-core/src/main/java/org/apache/doris/transaction/TransactionState.java
```diff
@@ -695,41 +730,21 @@ public void prolongPublishTimeout() {
...
         } else {
-            out.writeBoolean(true);
-            txnCommitAttachment.write(out);
-        }
-        out.writeLong(callbackId);
-        out.writeLong(timeoutMs);
-        out.writeInt(tableIdList.size());
-        for (Long aLong : tableIdList) {
-            out.writeLong(aLong);
+            String json = Text.readString(in);
```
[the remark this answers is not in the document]
mymeiyi: In this case, the fe meta version is 131 and use the original `readFields` method to read txn

mymeiyi:
run buildall

mymeiyi:
run buildall

mymeiyi:
run p0

regression-test/suites/insert_p0/txn_insert.groovy
dataroaring: what if 
begin
insert into t1;
delete from t1;
insert into t2 select * from t1;
commit;

result of first insert is visible to delete but not visible to the second insert.

dataroaring: > > Thread1 t2: insert into table1 Thread2 t1: insert into table2 select table1, t3: insert into table2 select table1.
> > Is result of t2 visible to insert at t3?
> 
> yes

For clarity,
Thread1 t2: begin  insert into table1; commit
thread2 t1:begin insert into table 2 select table2; t3: insert into table2 select table1; commit

If yes, what's our isolation level? ReadCommitted?

fe/fe-core/src/main/java/org/apache/doris/transaction/TransactionState.java
```diff
@@ -298,6 +299,16 @@ public SchemaInfo(OlapTable olapTable) {
     // table id -> schema info
     private Map<Long, SchemaInfo> txnSchemas = new HashMap<>();
 
+    @Getter
+    @Setter
+    private List<SubTransactionState> subTransactionStates;
+    @Getter
+    @SerializedName(value = "subTxnIdToTableCommitInfo")
```
morningman, line 306: Use short name, such as:
```suggestion
    @SerializedName(value = "sti")
```

fe/fe-core/src/main/java/org/apache/doris/load/loadv2/LoadJobFinalOperation.java
```diff
@@ -38,13 +39,20 @@
  * It is used to edit the job final state.
  */
 public class LoadJobFinalOperation extends TxnCommitAttachment implements Writable {
+    @SerializedName(value = "id")
     private long id;
+    @SerializedName(value = "loadingStatus")
     private EtlStatus loadingStatus = new EtlStatus();
+    @SerializedName(value = "progress")
     private int progress;
+    @SerializedName(value = "loadStartTimestamp")
```
morningman, line 48: use short name. like `lst`.
Same comment for all other fields

morningman (changes requested):
This a large feature, so I suggest to add more detail description about what you did in this PR, including:
1. Design: At least you need to let reviewer know WHAT and HOW, the basic idea behind your design; The main class, etc.
2. Compatibility: Doris already supported `begin; insert values; insert values; commit`. What is the diff from this one? Is there any thing user need to be aware of before upgrading?

And I also see you modify other modules such as `delete` and `kafka routine load`? Why? Is there any thing need to be noticed?

Please ref to 
https://github.com/apache/doris/pull/33610
or
https://github.com/apache/doris/pull/33858

to leave a better PR comment.

mymeiyi:
run buildall

mymeiyi:
run buildall

mymeiyi:
run buildall

mymeiyi:
run p0

mymeiyi:
run cloud_p1

mymeiyi:
run buildall

mymeiyi:
run p0

mymeiyi:
run buildall

mymeiyi:
run buildall

fe/fe-core/src/main/java/org/apache/doris/transaction/TransactionEntry.java
```diff
@@ -261,22 +278,62 @@ public long abortTransaction()
         }
     }
 
+    private void beforeFinishTransaction() {
+        if (isTransactionBegan) {
+            List<Long> tableIds = transactionState.getTableIdList().stream().distinct().collect(Collectors.toList());
+            transactionState.setTableIdList(tableIds);
+            subTransactionStates.sort((s1, s2) -> {
+                if (s1.getSubTransactionType() == SubTransactionType.INSERT
```
morningman, line 286: Why using this order?

fe/fe-core/src/main/java/org/apache/doris/transaction/GlobalTransactionMgr.java
```diff
@@ -243,6 +245,22 @@ public void commitTransaction(long dbId, List<Table> tableList, long transaction
         dbTransactionMgr.commitTransaction(tableList, transactionId, tabletCommitInfos, txnCommitAttachment, false);
     }
 
+    /**
+     * @note callers should get all tables' write locks before call this api
+     */
+    public void commitTransaction(long dbId, List<Table> tableList, long transactionId,
```
morningman, line 251: the timeout is unused

fe/fe-core/src/main/java/org/apache/doris/transaction/TransactionEntry.java
mymeiyi: 1. We provide that, each statement cannot read the modifications made by other statements within the same transaction.
2. For delete command, there are 2 implementations,  one is delete condition, one is insert. 
3. If the delete condition is committed after the insert, the delete will work for the insert, for example:
```
begin;
insert into t select * from t1 where id = 1;
delete from t where id = 1;
commit;
```
4. So we make the delete condition commit before the insert.

dataroaring (dismissed):
LGTM

mymeiyi:
run buildall

mymeiyi:
run p0

morningman (approved):
LGTM

dataroaring (approved):
LGTM

mymeiyi:
run cloud_p1

dataroaring (approved):
LGTM

PR #33004, opened 2024-03-29
[Fix](Nereids) fix leading with inner join and right join

LiBinfeng-01:
## Proposed changes

Problem:
when plan with inner join and right join, some predicate and join can derive some null reject column, leading could change these column to nullable
Solved:
before leading, we can do eliminate outer join first to avoid message lost

## Further comments


LiBinfeng-01:
run buildall

LiBinfeng-01:
run buildall

PR #33075, opened 2024-03-31
[improvement](spill) avoid spill if memory is enough

jacktengg:
## Proposed changes

Issue Number: close #xxx

## Further comments

### DIFF
--- a/be/src/pipeline/exec/partitioned_aggregation_sink_operator.cpp
+++ b/be/src/pipeline/exec/partitioned_aggregation_sink_operator.cpp
@@ -71,12 +71,6 @@ Status PartitionedAggSinkLocalState::close(RuntimeState* state, Status exec_stat
         return Status::OK();
     }
     dec_running_big_mem_op_num(state);
-    {
-        std::unique_lock<std::mutex> lk(_spill_lock);
-        if (_is_spilling) {
-            _spill_cv.wait(lk);
-        }
-    }
     return Base::close(state, exec_status);
 }
 
@@ -168,10 +162,6 @@ Status PartitionedAggSinkOperatorX::sink(doris::RuntimeState* state, vectorized:
     auto* runtime_state = local_state._runtime_state.get();
     RETURN_IF_ERROR(_agg_sink_operator->sink(runtime_state, in_block, false));
     if (eos) {
-        local_state.profile()->add_info_string(
-                "Spilled", local_state._shared_state->is_spilled ? "true" : "false");
-        LOG(INFO) << "agg node " << id()
-                  << " sink eos, spilled: " << local_state._shared_state->is_spilled;
         if (local_state._shared_state->is_spilled) {
             if (revocable_mem_size(state) > 0) {
                 RETURN_IF_ERROR(revoke_memory(state));
@@ -240,9 +230,10 @@ Status PartitionedAggSinkLocalState::revoke_memory(RuntimeState* state) {
     LOG(INFO) << "agg node " << Base::_parent->id() << " revoke_memory"
               << ", eos: " << _eos;
     RETURN_IF_ERROR(Base::_shared_state->sink_status);
-    DCHECK(!_is_spilling);
-    _is_spilling = true;
-    _shared_state->is_spilled = true;
+    if (!_shared_state->is_spilled) {
+        _shared_state->is_spilled = true;
+        profile()->add_info_string("Spilled", "true");
+    }
 
     // TODO: spill thread may set_ready before the task::execute thread put the task to blocked state
     if (!_eos) {
@@ -252,7 +243,6 @@ Status PartitionedAggSinkLocalState::revoke_memory(RuntimeState* state) {
     Status status;
     Defer defer {[&]() {
         if (!status.ok()) {
-            _is_spilling = false;
             if (!_eos) {
                 Base::_dependency->Dependency::set_ready();
             }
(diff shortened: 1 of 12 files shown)


jacktengg:
run buildall

be/src/pipeline/exec/spill_sort_sink_operator.cpp
```diff
@@ -179,9 +179,20 @@ Status SpillSortSinkOperatorX::sink(doris::RuntimeState* state, vectorized::Bloc
             local_state._shared_state->in_mem_shared_state->sorter->data_size());
     if (eos) {
         if (_enable_spill) {
```
yiguolei, line 181: I think   spill_sort_sink_operator is used when _enable_spill== true. So that I think this variable is always true in this case??

be/src/pipeline/exec/partitioned_hash_join_probe_operator.cpp
```diff
@@ -783,6 +783,11 @@ Status PartitionedHashJoinProbeOperatorX::get_block(RuntimeState* state, vectori
                                                     bool* eos) {
     *eos = false;
     auto& local_state = get_local_state(state);
+    Defer defer([&] {
+        if (*eos) {
+            LOG(INFO) << "hash probe " << id() << " eos";
```
yiguolei, line 788: This is a debug message?

be/src/pipeline/exec/spill_sort_sink_operator.cpp
```diff
@@ -179,9 +179,20 @@ Status SpillSortSinkOperatorX::sink(doris::RuntimeState* state, vectorized::Bloc
...
-            if (revocable_mem_size(state) > 0) {
-                RETURN_IF_ERROR(revoke_memory(state));
+            local_state.profile()->add_info_string(
+                    "Spilled", local_state._shared_state->is_spilled ? "true" : "false");
+            LOG(INFO) << "sort node " << id()
```
yiguolei, line 184: This is a debug message?

be/src/pipeline/exec/partitioned_aggregation_source_operator.cpp
```diff
@@ -133,13 +133,16 @@ Status PartitionedAggSourceOperatorX::get_block(RuntimeState* state, vectorized:
     SCOPED_TIMER(local_state.exec_time_counter());
     RETURN_IF_ERROR(local_state._status);
 
-    RETURN_IF_ERROR(local_state.initiate_merge_spill_partition_agg_data(state));
+    if (local_state._shared_state->is_spilled) {
+        RETURN_IF_ERROR(local_state.initiate_merge_spill_partition_agg_data(state));
```
yiguolei, line 137: This method will merge block again. Do we need clear the previous partition data??

yiguolei, line 137: why call init merge every time?? It is only called the first time or meet previous partition's EOS?

be/src/pipeline/exec/partitioned_aggregation_sink_operator.cpp
```diff
@@ -168,13 +168,21 @@ Status PartitionedAggSinkOperatorX::sink(doris::RuntimeState* state, vectorized:
...
     if (eos) {
-        LOG(INFO) << "agg node " << id() << " sink eos";
-        if (revocable_mem_size(state) > 0) {
-            RETURN_IF_ERROR(revoke_memory(state));
-        } else {
-            for (auto& partition : local_state._shared_state->spill_partitions) {
-                RETURN_IF_ERROR(partition->finish_current_spilling(eos));
+        local_state.profile()->add_info_string(
+                "Spilled", local_state._shared_state->is_spilled ? "true" : "false");
+        LOG(INFO) << "agg node " << id()
```
yiguolei, line 173: This is debug log?

be/src/pipeline/exec/partitioned_aggregation_sink_operator.cpp
```diff
@@ -168,13 +168,21 @@ Status PartitionedAggSinkOperatorX::sink(doris::RuntimeState* state, vectorized:
     auto* runtime_state = local_state._runtime_state.get();
     RETURN_IF_ERROR(_agg_sink_operator->sink(runtime_state, in_block, false));
     if (eos) {
-        LOG(INFO) << "agg node " << id() << " sink eos";
-        if (revocable_mem_size(state) > 0) {
-            RETURN_IF_ERROR(revoke_memory(state));
-        } else {
-            for (auto& partition : local_state._shared_state->spill_partitions) {
-                RETURN_IF_ERROR(partition->finish_current_spilling(eos));
+        local_state.profile()->add_info_string(
```
yiguolei, line 171: Not meet EOS.
Shoud add this to profile when the spilled is set to true at first time.
Because we will provide realtime profile in the future, user will find his query is spilled to disk.
If we set it when eos, then user will not find ths spill task when the query not spilled finished.

be/src/pipeline/exec/partitioned_aggregation_sink_operator.cpp
jacktengg: If profile is not enabled, we can seach log to check if query is spilled.

be/src/pipeline/exec/partitioned_aggregation_source_operator.cpp
jacktengg: `initiate_merge_spill_partition_agg_data` will check whether the current partition still has data, if it is, it continue to output data from the current partition; if not, will merge the next partition.

be/src/pipeline/exec/spill_sort_sink_operator.cpp
jacktengg: We currently only support spill for full sort, but when creating the sort operatorX, it's hard to decide whether it's full sort.

It's in the `SortSinkOperatorX::prepare` phase to decide whether to use full sort, so we create 
 `SpillSortSinkOperatorX` when `enable_sort_spill=true` for all type of sort, but actually only enable spillable sort for full sort.

jacktengg:
run buildall

yiguolei (approved):
LGTM

mrhhsg (approved):
LGTM

PR #33110, opened 2024-04-01
[branch-2.0][fix](broker) fix broker writer close issue

morningman:
the broker writer may be closed even it is not opened yet.
If the broker writer is not opened, the fd is empty, the close will fail.
So check the open status of broker writer before closing it.

Also add kerberos debug options for broker process.


morningman:
run buildall

xiaokang (approved):
LGTM

PR #33170, opened 2024-04-02
[FIX](inverted_index) fix inverted index write array with _doc is empty

amorynan:
## Proposed changes
when a array in single row is null or array has all null elements , the _doc param may has empty field, in this situation if we also call add_document,  inverted index will call process_document() make sure field with inverted , but field is cleared, and _doc is empty which will make heap use after free, so we should avoid this situation happen!
Issue Number: close #xxx

## Further comments


amorynan:
run buildall

qidaye (approved):

amorynan:
run cloud_p1

airborne12 (approved):
LGTM

PR #33197, opened 2024-04-03
[Fix](executor)Fix normal group can not be appended when image exits

wangbo:
## Proposed changes
Fix normal group can not be appended when image exits.
Currently when FE images exits,then normal group could be put into WorkloadGroupMgr failed because image's  WorkloadGroup meta can overwrite normal group.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/resource/workloadgroup/WorkloadGroupMgr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/resource/workloadgroup/WorkloadGroupMgr.java
@@ -65,6 +65,9 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
 public class WorkloadGroupMgr implements Writable, GsonPostProcessable {
 
     public static final String DEFAULT_GROUP_NAME = "normal";
+
+    public static final Long DEFAULT_GROUP_ID = 1L;
+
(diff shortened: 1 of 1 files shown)


fe/fe-core/src/main/java/org/apache/doris/catalog/Env.java
```diff
@@ -1020,6 +1020,7 @@ public void initialize(String[] args) throws Exception {
         // 3. Load image first and replay edits
         this.editLog = new EditLog(nodeName);
         loadImage(this.imageDir); // load image file
+        this.workloadGroupMgr.appendInternalWorkloadGroup();
```
yiguolei, line 1023: Not call this method in load image method. We should isolate workload manager's init only in workload mgr.

fe/fe-core/src/main/java/org/apache/doris/resource/workloadgroup/WorkloadGroupMgr.java
```diff
@@ -137,6 +137,18 @@ public void resetQueryQueueProp() {
     }
 
     public WorkloadGroupMgr() {
+    }
+
+    public static WorkloadGroupMgr read(DataInput in) throws IOException {
+        String json = Text.readString(in);
+        return GsonUtils.GSON.fromJson(json, WorkloadGroupMgr.class);
```
yiguolei, line 144: Add check if the default workload group is inited here and put it into workload group manager if not exist.

yiguolei (approved):

Gabriel39 (approved):

PR #33235, opened 2024-04-03
[fix](kill_sql) support force cancel sql

Yulei-Yang:
In some scenarios, cancel query does not work, no matter auto cancel or manually cancel.
One case is when an error occurs during analyzing a sql.
image

## Proposed changes

Issue Number: close #xxx

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
@@ -1482,15 +1482,22 @@ public class StmtExecutor {
             }
             return;
         }
-        Coordinator coordRef = coord;
-        if (coordRef != null) {
-            coordRef.cancel();
-        }
         if (mysqlLoadId != null) {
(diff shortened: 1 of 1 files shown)


Yulei-Yang:
run buildall

fe/fe-core/src/main/java/org/apache/doris/qe/StmtExecutor.java
```diff
@@ -1482,16 +1482,23 @@ public void cancel() {
...
             Env.getCurrentEnv().getAnalysisManager().cancelSyncTask(context);
         }
+
+        Coordinator coordRef = coord;
+        if (coordRef != null) {
+            coordRef.cancel();
+        } else {
+            // force cancel, for scenerios when above strategies does not work, like sql error in analyze period
+            if (context != null) {
+                context.cleanup();
```
yiguolei, line 1498: The modification is very dangerouse. If you cleanup the context, we do not know the actual value in context and the analyzer may use it. And send some illegal plan to BE and BE may core.

PR #33242, opened 2024-04-03
[opt](hive) simplify the isSplitable method to avoid too many hadoop metrics

morningman:
The `inputFormat.isSplitable()` method will create FileSystem.
Each FileSystem will register a lot hadoop metrics, which will take a lot memory.
This PR simplify it to avoid calling `inputFormat.isSplitable()`.

Only for branch-2.0


morningman:
run buildall

PR #33423, opened 2024-04-09
[fix](restore) Fix the conflict IDs between two cluster

w41ter:
## Proposed changes

Issue Number: close #xxx

The meta of the restore may come from different clusters, so the original ID in the meta may conflict with the ID of the new cluster. For example, if a newly allocated ID happens to be the same as an original ID, the original one may be overwritten when executing `put`.

## Further comments


dataroaring (approved):
LGTM

swjtu-zhanglei (approved):

dataroaring (approved):
LGTM

PR #33432, opened 2024-04-09
[fix](catalog) fix resource is not reopen when rename catalog

zy-kkk:
## Proposed changes

Issue Number: close #xxx

During the renaming of `JdbcCatalog`, I noticed that the `jdbcClient` was being closed, resulting in exceptions during subsequent queries. This happens because the `removeCatalog` method is invoked when changing the name, which in turn calls the `onClose` method of the catalog. Ideally, the client should not be closed when renaming the catalog. However, to avoid extra checks in the `removeCatalog` method, we can simply execute `onRefresh` in the `addCatalog` method to address this issue.

## Further comments


zy-kkk:
run buildall

morningman (approved):
LGTM

morningman (approved):

AshinGau (approved):

PR #33553, opened 2024-04-11
[Fix](timezone) fix miss of expected rounding of Date type with timezone

zclllyybb:
## Proposed changes

Issue Number: close #xxx

1.

when we load a datetime literal with a timezone suffix to the `Date` type column. The rounding from time to date would be ignored before. like:
```
2024-04-11T06:00:13+08:00 ---> UTC timezone
got: 2024-04-11
```
but should be `2024-04-10`

now fixed it.

2.

add new header "time_zone" for alternative in streamload

## Further comments


zclllyybb:
run buildall

HappenLee (changes requested):
Add commit desc for pr

be/src/vec/runtime/vdatetime_value.cpp
```diff
@@ -2192,6 +2192,23 @@ bool DateV2Value<T>::from_date_str_base(const char* date_str, int len, int scale
         return false;
     }
 
+    // here maybe date changed because of rouding
```
HappenLee, line 2195: Add detail comment for the code.  what case, why we need the code.

zclllyybb:
run buildall

HappenLee (approved):
LGTM

yiguolei (approved):

PR #33722, opened 2024-04-16
[fix](Nereids) fix some sql cache consistence bug between multiple frontends

924060929:
## Proposed changes

1. fix some sql cache consistence bug between multiple frontends which introduced by #33262, fix by use row policy as the part of sql cache key. 
2. support dynamic update the num of fe manage sql cache key

### DIFF
--- a/fe/fe-common/src/main/java/org/apache/doris/catalog/ScalarType.java
+++ b/fe/fe-common/src/main/java/org/apache/doris/catalog/ScalarType.java
@@ -744,15 +744,19 @@ public class ScalarType extends Type {
             case DECIMAL128:
             case DECIMAL256:
             case DATETIMEV2: {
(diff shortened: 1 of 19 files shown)


924060929:
run buildall

924060929:
run buildall

924060929:
run buildall

morrySnow (approved):

yiguolei (changes requested):
Sync between multi FEs is very difficult.
We could simply using consistency hash like memcache to solve the problem.

924060929:
> Sync between multi FEs is very difficult. We could simply using consistency hash like memcache to solve the problem.

there hasn't any sync logic in this pr

yiguolei (approved):

PR #33824, opened 2024-04-18
[fix](merge-iterator) Fix mem leak when get next batch failed #33627

liaoxin01:
cherry pick from #33627


liaoxin01:
run buildall

PR #33836, opened 2024-04-18
[fix](statistics)Improve analyze timeout logic.

Jibing-Li:
Before, analyze_timeout will also affect analyze stmt, which may cause failed to cancel analyze task. This PR change analyze_timeout to only affect timeout value for executing analyze sql.

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/statistics/AnalysisManager.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/statistics/AnalysisManager.java
@@ -867,7 +867,7 @@ public class AnalysisManager implements Writable {
                 executor.submit(() -> {
                     try {
                         if (cancelled) {
-                            errorMessages.add("Query timeout or user cancelled."
+                            errorMessages.add("Query Timeout or user Cancelled."
                                     + "Could set analyze_timeout to a bigger value.");
                             return;
                         }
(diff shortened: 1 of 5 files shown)


Jibing-Li:
run buildall

fe/fe-core/src/main/java/org/apache/doris/qe/ConnectContext.java
```diff
@@ -995,7 +996,7 @@ public int getExecTimeout() {
             // particular for insert stmt, we can expand other type of timeout in the same way
             return Math.max(sessionVariable.getInsertTimeoutS(), sessionVariable.getQueryTimeoutS());
         } else if (executor != null && executor.isAnalyzeStmt()) {
-            return sessionVariable.getAnalyzeTimeoutS();
+            return StatisticConstants.ANALYZE_TIMEOUT_IN_SEC;
```
morningman, line 999: For `analyze` stmt, why use fixed timeout value?

Jibing-Li: AnalyzeStmt is the statement executed by mysql connection thread. Before, AnalyzeStmt was using analyze_timeout, but it brings the problem that AnalyzeStmt may be timeout but the internal analyze sql has not cancelled yet, in this case, the cancellation may fail. So here use a fixed timeout value for AnalyzeStmt and use analyze_timeout to control the analyze sql timeout. When analyze sql timeout, AnalyzeStmt will get an exception and return.

Jibing-Li:
run buildall

Jibing-Li:
run buildall

morningman (approved):
LGTM

wuwenchi (approved):

PR #33843, opened 2024-04-18
[Fix](nereids) fix bind order by expression logic

feiniaofeiafei:
select sum(a) c1, min(a) c1 from t1 group by b order by c1;
This pr make the bind logic in order by list same with mysql. Doris used to  report error in this situation, but mysql bind to the first c1.
When there are more than one c1 in agg outputs, the c1 in order by should bind to the first c1 in projection.


feiniaofeiafei:
run buildall

924060929 (approved):

starocean999 (approved):

PR #33853, opened 2024-04-18
[fix](memory) Allocator address sanitizers enable print stack trace 

xinyiZzz:
## Proposed changes

Allocator address sanitizers enable `dwarf_location_info_mode=DISABLED` stack trace, this will no stack line numbers, maybe cause performance reduce 1x

## Further comments

### DIFF
--- a/be/src/common/config.cpp
+++ b/be/src/common/config.cpp
@@ -1060,7 +1060,7 @@ DEFINE_mString(kerberos_krb5_conf_path, "/etc/krb5.conf");
 
 DEFINE_mString(get_stack_trace_tool, "libunwind");
 DEFINE_mString(dwarf_location_info_mode, "FAST");
-DEFINE_mBool(enable_address_sanitizers_with_stack_trace, "false");
+DEFINE_mBool(enable_address_sanitizers_with_stack_trace, "true");
 
 // the ratio of _prefetch_size/_batch_size in AutoIncIDBuffer
 DEFINE_mInt64(auto_inc_prefetch_size_ratio, "10");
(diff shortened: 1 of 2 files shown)


xinyiZzz:
run buildall

be/src/common/config.cpp
```diff
@@ -1059,7 +1059,7 @@ DEFINE_mString(kerberos_krb5_conf_path, "/etc/krb5.conf");
 
 DEFINE_mString(get_stack_trace_tool, "libunwind");
 DEFINE_mString(dwarf_location_info_mode, "FAST");
-DEFINE_mBool(enable_address_sanitizers_with_stack_trace, "false");
+DEFINE_mBool(enable_address_sanitizers_with_stack_trace, "true");
```
yiguolei, line 1062: not set to true by default

be/src/common/stack_trace.cpp
```diff
@@ -439,7 +439,8 @@ static StackTraceCache& cacheInstance() {
 
 static std::mutex stacktrace_cache_mutex;
 
-std::string toStringCached(const StackTrace::FramePointers& pointers, size_t offset, size_t size) {
+std::string toStringCached(const StackTrace::FramePointers& pointers, size_t offset, size_t size,
```
yiguolei, line 442: do not modify these very low API, it is very likely to import some bugs.

be/src/common/config.cpp
xinyiZzz: only take effect under DEBUG

be/src/common/stack_trace.cpp
xinyiZzz: i am 99.99% sure it's fine

yiguolei (approved):

wangbo (approved):
LGTM

PR #31111, opened 2024-02-19
[fix](s3client) fix unable to access COS 

morningman:
## Proposed changes

After this #30799, we upgrade the aws sdk version from 2.17.257 to 2.20.131.
The default endpoint url style logic seem changes.
So I set `pathStyleAccessEnabled(true)` to fix this.
Tested on S3, OSS, COS

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #30799, opened 2024-02-04
[enhance](iceberg)upgrade iceberg to 1.4.3

wuwenchi:
## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #33858, opened 2024-04-18
[Enhancement](multi-catalog) Rewrite `S3URI` to remove tricky virtual bucket mechanism and support different uri styles by flags.

kaka11chen:
### Issue:
Many domestic cloud vendors are compatible with the s3 protocol. However, early versions of s3 client will only generate path style http requests (https://github.com/aws/aws-sdk-java-v2/pull/763) when encountering endpoints that do not start with s3, while some cloud vendors only support virtual host style http request.

Therefore, Doris used `forceVirtualHosted` in `S3URI` to convert it into a virtual hosted path and implemented it through path style.
For example:
For s3 uri `s3://my-bucket/data/file.txt`, It will eventually be parsed into:
- virtualBucket: my-bucket
- Bucket: data (bucket must be set, otherwise the s3 client will report an error) Especially this step is particularly tricky because of the limitations of the s3 client.
- Key: file.txt

 The path style mode is used to generate an http request similar to the virtual host by setting the endpoint to virtualBucket + original endpoint, setting the bucket and key.
**However, the bucket and key here are inconsistent with the original concepts of s3, but the aws client happens to be able to generate an http request similar to the virtual host through the path style mode.**

However, after #30799 we have upgrade the aws sdk version from 2.17.257 to 2.20.131. The current aws s3 client can already generate a virtual host by third party by default style of http request. So in #31111 need to set the path style option, let the s3 client use doris' virtual bucket mechanism to continue working.

**Finally, the virtual bucket mechanism is too confusing and tricky, and we no longer need it with the new version of s3 client.**

### Resolution:

Rewrite `S3URI` to remove tricky virtual bucket mechanism and support different uri styles by flags.

This class represents a fully qualified location in S3 for input/output operations expressed as as URI.
 #### For AWS S3, URI common styles:
  - AWS Client Style(Hadoop S3 Style): `s3://my-bucket/path/to/file?versionId=abc123&partNumber=77&partNumber=88`
  - Virtual Host Style: `https://my-bucket.s3.us-west-1.amazonaws.com/resources/doc.txt?versionId=abc123&partNumber=77&partNumber=88`
  - Path Style: `https://s3.us-west-1.amazonaws.com/my-bucket/resources/doc.txt?versionId=abc123&partNumber=77&partNumber=88`
 
  Regarding the above-mentioned common styles, we can use <code>isPathStyle</code> to control whether to use path style
  or virtual host style.
  "Virtual host style" is the currently mainstream and recommended approach to use, so the default value of
  <code>isPathStyle</code> is false.
 
  #### Other Styles:
  - Virtual Host AWS Client (Hadoop S3) Mixed Style:
    `s3://my-bucket.s3.us-west-1.amazonaws.com/resources/doc.txt?versionId=abc123&partNumber=77&partNumber=88`
  - Path AWS Client (Hadoop S3) Mixed Style:
     `s3://s3.us-west-1.amazonaws.com/my-bucket/resources/doc.txt?versionId=abc123&partNumber=77&partNumber=88`
 
  For these two styles, we can use <code>isPathStyle</code> and <code>forceParsingByStandardUri</code>
  to control whether to use.
  Virtual Host AWS Client (Hadoop S3) Mixed Style: <code>isPathStyle = false && forceParsingByStandardUri = true</code>
  Path AWS Client (Hadoop S3) Mixed Style: <code>isPathStyle = true && forceParsingByStandardUri = true</code>
 
  When the incoming location is url encoded, the encoded string will be returned.
  For <code>getKey()</code>, <code>getQueryParams()</code> will return the encoding string

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/tablefunction/S3TableValuedFunction.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/tablefunction/S3TableValuedFunction.java
@@ -54,8 +54,6 @@ public class S3TableValuedFunction extends ExternalFileTableValuedFunction {
             ImmutableSet.of("access_key", "secret_key", "session_token", "region",
                     "ACCESS_KEY", "SECRET_KEY", "SESSION_TOKEN", "REGION");
 
-    private String virtualBucket = "";
-
     public S3TableValuedFunction(Map<String, String> properties) throws AnalysisException {
         // 1. analyze common properties
         Map<String, String> otherProps = super.parseCommonProperties(properties);
@@ -68,12 +66,17 @@ public class S3TableValuedFunction extends ExternalFileTableValuedFunction {
         forwardCompatibleDeprecatedKeys(otherProps);
 
         String usePathStyle = getOrDefaultAndRemove(otherProps, PropertyConverter.USE_PATH_STYLE, "false");
-        boolean forceVirtualHosted = isVirtualHosted(uriStr, Boolean.parseBoolean(usePathStyle));
-        S3URI s3uri = getS3Uri(uriStr, forceVirtualHosted);
-        String endpoint = forceVirtualHosted
-                ? getEndpointAndSetVirtualBucket(s3uri, otherProps) : s3uri.getBucketScheme();
+        String forceParsingByStandardUri = getOrDefaultAndRemove(otherProps,
+                PropertyConverter.FORCE_PARSING_BY_STANDARD_URI, "false");
+
+        S3URI s3uri = getS3Uri(uriStr, Boolean.parseBoolean(usePathStyle.toLowerCase()),
+                Boolean.parseBoolean(forceParsingByStandardUri.toLowerCase()));
+        String endpoint = otherProps.containsKey(S3Properties.ENDPOINT) ? otherProps.get(S3Properties.ENDPOINT) :
+                s3uri.getEndpoint().orElseThrow(() ->
+                        new AnalysisException(String.format("Properties '%s' is required.", S3Properties.ENDPOINT)));
         if (!otherProps.containsKey(S3Properties.REGION)) {
-            String region = S3Properties.getRegionOfEndpoint(endpoint);
+            String region = s3uri.getRegion().orElseThrow(() ->
+                    new AnalysisException(String.format("Properties '%s' is required.", S3Properties.REGION)));
             otherProps.put(S3Properties.REGION, region);
         }
         checkNecessaryS3Properties(otherProps);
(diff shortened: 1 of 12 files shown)


kaka11chen:
run buildall

kaka11chen:
buildall

kaka11chen:
run buildall

kaka11chen:
run buildall

regression-test/suites/load_p0/tvf/test_s3_tvf.groovy
```diff
@@ -140,20 +146,23 @@ suite("test_s3_tvf", "load_p0") {
         attributeList.add(new TvfAttribute(table, ["k00", "k01", "k02", "k03", "k04", "k05", "k06", "k07", "k08", "k09", "k10", "k11", "k12", "k13", "k14", "k15", "k16", "k17", "k18"], "", "", true)
                 .addProperty("uri", "s3://doris-build-1308700295.cos.ap-beijing.myqcloud.com/regression/load/data/basic_data_with_errors.csv")
                 .addProperty("format", "csv")
-                .addProperty("column_separator", "|"))
+                .addProperty("column_separator", "|")
+                .addProperty("force_parsing_by_standard_uri", "true"))
```
morningman, line 150: The default value of `force_parsing_by_standard_uri` is false.
So if we do this modification, user need to change their origin stmt?

fe/fe-core/src/main/java/org/apache/doris/tablefunction/S3TableValuedFunction.java
```diff
@@ -68,12 +66,17 @@ public S3TableValuedFunction(Map<String, String> properties) throws AnalysisExce
         forwardCompatibleDeprecatedKeys(otherProps);
 
         String usePathStyle = getOrDefaultAndRemove(otherProps, PropertyConverter.USE_PATH_STYLE, "false");
-        boolean forceVirtualHosted = isVirtualHosted(uriStr, Boolean.parseBoolean(usePathStyle));
-        S3URI s3uri = getS3Uri(uriStr, forceVirtualHosted);
-        String endpoint = forceVirtualHosted
-                ? getEndpointAndSetVirtualBucket(s3uri, otherProps) : s3uri.getBucketScheme();
+        String forceParsingByStandardUri = getOrDefaultAndRemove(otherProps,
```
morningman, line 69: Define the default value of `forceParsingByStandardUri` and `usePathStyle` somewhere.

regression-test/suites/load_p0/tvf/test_s3_tvf.groovy
kaka11chen: Yes.

fe/fe-core/src/main/java/org/apache/doris/tablefunction/S3TableValuedFunction.java
kaka11chen: OK.

regression-test/suites/load_p0/tvf/test_s3_tvf.groovy
kaka11chen: Will update document later.

kaka11chen:
run buildall

morningman:
run buildall

morningman (approved):
LGTM

wsjz (approved):
LGTM

PR #33875, opened 2024-04-19
[Fix](executor)Fix stream load can only use default group

wangbo:
## Proposed changes
In stream load, when be request a plan from FE, it has no session variable which means its session variables has default value.
But stream load currently get workload group from session, this means its group is always default group.
The correct approach is to stream load should get workload group from http args or user's property.
Currently PR is to fix the problem which stream load not gets workload group from user's property correctly.


wangbo:
run buildall

liaoxin01 (approved):
LGTM

xinyiZzz (approved):
LGTM

PR #33888, opened 2024-04-19
[opt](coord) use single instance only with small limit

morningman:
## Proposed changes

Followup #31342
When executing simple query like `select * from tbl limit 1`.
Doris will use only one instance for each scan node, to reduce the fan-out of simple queries.
But if the limit number is large, eg, 1 billion, this logic is not suitable.

So I add a new session variable `limit_rows_for_single_instance`, default is 10000.
Only if the limit number is lower than this value, Doris will use one instance for scan node.

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/ScanNode.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/ScanNode.java
@@ -746,8 +746,8 @@ public abstract class ScanNode extends PlanNode {
         return Integer.MAX_VALUE;
     }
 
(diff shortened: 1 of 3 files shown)


morningman:
run buildall

HappenLee (dismissed):
LGTM

yiguolei (dismissed):
LGTM

morningman:
run buildall

yiguolei (approved):

Gabriel39 (approved):

PR #33934, opened 2024-04-21
[fix](scan) fix ignore expr exec when _non_predicate_columns is empty  

Mryange:
## Proposed changes

Correct SQL
```
mysql [tpch_sf100]>select count(1) from lineitem, tmp where l_orderkey > (tmp.id + tmp.id);
+-----------+
| count(1)  |
+-----------+
| 600037884 |
+-----------+
```
Incorrect SQL
```
mysql [tpch_sf100]>select count(1) from lineitem, (select id + id as x from tmp) t where l_orderkey > x;
+-----------+
| count(1)  |
+-----------+
| 512150170 |
+-----------+
```
The reason for the error is that we have planned the RF, and we have logic to ignore expr execution. However, in cases where there is only one column, _non_predicate_columns will be empty.

Then, in this code...
```cpp
    if (!_is_need_vec_eval && !_is_need_short_eval && !_is_need_expr_eval) {
        if (_non_predicate_columns.empty()) {
            return Status::InternalError("_non_predicate_columns is empty");
        }
        RETURN_IF_ERROR(_convert_to_expected_type(_first_read_column_ids));
        RETURN_IF_ERROR(_convert_to_expected_type(_non_predicate_columns));
        _output_non_pred_columns(block);
        _output_index_result_column(nullptr, 0, block);
    }
```
The previous condition was "_is_need_vec_eval = true," and _non_predicate_columns was empty.
Later, when _is_need_vec_eval is set to false, it will output an empty block.

## Further comments

### DIFF
--- a/be/src/olap/rowset/segment_v2/segment_iterator.cpp
+++ b/be/src/olap/rowset/segment_v2/segment_iterator.cpp
@@ -1957,14 +1957,16 @@ void SegmentIterator::_replace_version_col(size_t num_rows) {
 uint16_t SegmentIterator::_evaluate_vectorization_predicate(uint16_t* sel_rowid_idx,
                                                             uint16_t selected_size) {
     SCOPED_RAW_TIMER(&_opts.stats->vec_cond_ns);
-    if (_is_need_vec_eval) {
-        _is_need_vec_eval = false;
-        for (const auto& pred : _pre_eval_block_predicate) {
-            _is_need_vec_eval |= (!pred->always_true());
+    bool all_pred_always_true = true;
+    for (const auto& pred : _pre_eval_block_predicate) {
+        if (!pred->always_true()) {
+            all_pred_always_true = false;
+            break;
         }
     }
-    if (!_is_need_vec_eval) {
-        for (uint32_t i = 0; i < selected_size; ++i) {
+    //If all predicates are always_true, then return directly.
+    if (all_pred_always_true || !_is_need_vec_eval) {
+        for (uint16_t i = 0; i < selected_size; ++i) {
             sel_rowid_idx[i] = i;
         }
         return selected_size;
@@ -2248,6 +2250,9 @@ Status SegmentIterator::_next_batch_internal(vectorized::Block* block) {
     }
 
     if (!_is_need_vec_eval && !_is_need_short_eval && !_is_need_expr_eval) {
+        if (_non_predicate_columns.empty()) {
+            return Status::InternalError("_non_predicate_columns is empty");
+        }
         RETURN_IF_ERROR(_convert_to_expected_type(_first_read_column_ids));
         RETURN_IF_ERROR(_convert_to_expected_type(_non_predicate_columns));
         _output_non_pred_columns(block);


Mryange:
run buildall

yiguolei (commented):
Add test for this case

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -1957,7 +1957,10 @@ void SegmentIterator::_replace_version_col(size_t num_rows) {
 uint16_t SegmentIterator::_evaluate_vectorization_predicate(uint16_t* sel_rowid_idx,
                                                             uint16_t selected_size) {
     SCOPED_RAW_TIMER(&_opts.stats->vec_cond_ns);
-    if (_is_need_vec_eval) {
+    // The expr can only be skipped when _non_predicate_columns is not empty;
+    // otherwise, skipping the expr could result in an empty block
+    // being output while _non_predicate_columns is empty.
+    if (_is_need_vec_eval && !_non_predicate_columns.empty()) {
```
yiguolei, line 1963: why the output will be empty block? since we already read the predicate column into current_return_columns and line 1970 will set all flags to true.

Mryange:
run buildall

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -1957,13 +1957,12 @@ void SegmentIterator::_replace_version_col(size_t num_rows) {
...
                                                             uint16_t selected_size) {
     SCOPED_RAW_TIMER(&_opts.stats->vec_cond_ns);
-    if (_is_need_vec_eval) {
-        _is_need_vec_eval = false;
-        for (const auto& pred : _pre_eval_block_predicate) {
-            _is_need_vec_eval |= (!pred->always_true());
-        }
-    }
-    if (!_is_need_vec_eval) {
+    bool all_pred_always_true =
```
yiguolei, line 1960: 不要这么写，这么写语义并不是很清楚。
比如_pre_eval_block_predicate 为空的时候，这个值是多少？ 先设置一个变量，然后再赋值。

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -1957,13 +1957,12 @@ void SegmentIterator::_replace_version_col(size_t num_rows) {
...
+            std::all_of(_pre_eval_block_predicate.begin(), _pre_eval_block_predicate.end(),
+                        [](const auto& pred) { return pred->always_true(); });
+
+    //If all predicates are always_true, then return directly.
+    if (all_pred_always_true) {
```
yiguolei, line 1965: 这里为什么不检查_is_need_vec_eval？ 按道理说应该是all_pred_always_true || ！_is_need_vec_eval

Mryange:
run buildall

Mryange:
run buildall

yiguolei (approved):
LGTM

BiteTheDDDDt:
addd p1 or p2 case

BiteTheDDDDt (approved):

PR #33935, opened 2024-04-21
[feature](iceberg)support read iceberg complex type，iceberg.orc format and position delete.

hubgeter:
## Proposed changes
1. support read iceberg complex type : struct map.
2. support read the ORC storage format of Iceberg tables
3. Supports reading of the iceberg table after schema change when using the orc storage format . In addition to the rename operation performed inside the struct. [iceberg schema change](https://iceberg.apache.org/docs/latest/evolution/#schema-evolution)
4. support iceberg position delete.

Issue Number: close #xxx

Related changes：
1. Add in orc_reader: perform filtering operation when the column to be filtered does not exist in the current orc file.
eg.  
iceberg_table  :  a int, b string  
add column ( c int ) to  iceberg_table.
sql : `select * from iceberg_table where c is not null.` 

## Further comments
todo :  
1. Modify the Doris document .


be/src/vec/exec/format/orc/vorc_reader.h
```diff
@@ -177,6 +177,12 @@ class OrcReader : public GenericReader {
     Status get_parsed_schema(std::vector<std::string>* col_names,
                              std::vector<TypeDescriptor>* col_types) override;
 
+    Status get_parsed_col_name_iceberg_ids(std::vector<std::string>* col_names,
```
AshinGau, line 180: Other components like `tvf` will get the file schema/metadata, How to provide a public interface to get the schema/metadata. `get_parsed_col_name_iceberg_ids` may be correct, but it is not elegant enough.

be/src/vec/exec/format/table/iceberg_reader.cpp
```diff
@@ -136,6 +138,32 @@ Status IcebergTableReader::init_reader(
 
     return status;
 }
+Status IcebergTableReader::init_reader_for_orc(
```
AshinGau, line 141: Can we abstract the abstract class for the iceberg reader, using parquet/orc as the concrete implementation? If there are reuse functions between two formats, it can be moved to the parent class. The code format like `if-else` often increases the difficulty of later maintenance and modification.

be/src/vec/exec/format/orc/vorc_reader.cpp
```diff
@@ -1684,6 +1702,10 @@ Status OrcReader::get_next_block(Block* block, size_t* read_rows, bool* eof) {
             for (auto& conjunct : _non_dict_filter_conjuncts) {
                 filter_conjuncts.emplace_back(conjunct);
             }
+            //include missing_columns != missing_columns ; missing_column is null; missing_column != file_columns etc...
+            for (auto& [missing_col, conjunct] : _lazy_read_ctx.predicate_missing_columns) {
```
AshinGau, line 1706: The parquet reader also has the same issue.

AshinGau (dismissed):

hubgeter:
run buildall

hubgeter:
run buildall

hubgeter:
run buildall

morningman (approved):
LGTM

AshinGau (approved):

PR #33976, opened 2024-04-22
[Fix](executor)Fix reset remote scan pool

wangbo:
## Proposed changes
1 add set min/max order to avoid meanless log.
2 fix default value error for remote_scan_thread_num.
3 add check value check inf FE


wangbo:
run buildall

liaoxin01 (approved):
LGTM

BiteTheDDDDt (approved):

PR #34003, opened 2024-04-23
[fix](fd) release fd for shutdown tablets

dataroaring:
When a tablet is shutdown, maybe there is a running compaction on it, the compaction would put fd in cache, so we should clear them when move the directory to the trash and fail the compaction due to shutdown state.

e.g.
t0: a compaction is running on tablet t;
t1: tablet t is set shutdown, files of tablet are removed from segment cache.
t2: the compaction stops and files of tablet are added to the segment cache again.
t3: tablet t is moved to trash due to it is not used any more.
t4: the files are deleted from trash.
t5: the tablet's cache in segment cache remains and the capacity consumed by the files is not freed.

When a stale rowset is expired, it is removed when it is not used any more, and remove() function deletes files and removes the cache. However, there is no harm in clearing cache when a rowset is expired.

We reconstruct code related to clearing cache in object oriented way.

## Proposed changes

Issue Number: close #xxx

## Further comments

### DIFF
--- a/be/src/olap/base_tablet.h
+++ b/be/src/olap/base_tablet.h
@@ -241,7 +241,7 @@ public:
 
     std::vector<RowsetSharedPtr> get_snapshot_rowset(bool include_stale_rowset = false) const;
 
-    virtual void clear_cache() { LOG(INFO) << "should not reach here"; }
+    virtual void clear_cache() = 0;
 
 protected:
     // Find the missed versions until the spec_version.
--- a/be/src/olap/base_tablet.h
+++ b/be/src/olap/base_tablet.h
@@ -241,6 +241,8 @@ public:
 
     std::vector<RowsetSharedPtr> get_snapshot_rowset(bool include_stale_rowset = false) const;
 
+    virtual void clear_cache() { LOG(INFO) << "should not reach here"; }
+
 protected:
     // Find the missed versions until the spec_version.
     //
--- a/be/src/olap/rowset/beta_rowset.cpp
+++ b/be/src/olap/rowset/beta_rowset.cpp
@@ -204,8 +204,8 @@ Status BetaRowset::remove() {
                 << ", version:" << start_version() << "-" << end_version()
                 << ", tabletid:" << _rowset_meta->tablet_id();
     // If the rowset was removed, it need to remove the fds in segment cache directly
-    SegmentLoader::instance()->erase_segments(_rowset_meta->rowset_id(),
-                                              _rowset_meta->num_segments());
+    clear_cache();
+
     auto fs = _rowset_meta->fs();
     if (!fs) {
         return Status::Error<INIT_FAILED>("get fs failed");
@@ -242,10 +242,6 @@ Status BetaRowset::remove() {
                         LOG(WARNING) << st.to_string();
                         success = false;
                     }
-                }
-                if (success) {
-                    RETURN_IF_ERROR(segment_v2::InvertedIndexSearcherCache::instance()->erase(
-                            inverted_index_file));
                 }
             }
         }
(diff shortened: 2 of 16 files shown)


dataroaring:
run buildall

dataroaring:
run buildall

dataroaring:
run buildall

dataroaring:
run buildall

be/src/olap/rowset/rowset.h
```diff
@@ -303,6 +303,9 @@ class Rowset : public std::enable_shared_from_this<Rowset> {
 
     std::string get_rowset_info_str();
 
+    virtual void clear_inverted_index_cache() { LOG(INFO) << "should not reach here"; }
```
gavinchou, line 306: LOG(ERROR)?

dataroaring: error is horrible for users.

be/src/olap/rowset/beta_rowset.cpp
```diff
@@ -124,6 +125,21 @@ Status BetaRowset::get_inverted_index_size(size_t* index_size) {
     return Status::OK();
 }
 
+void BetaRowset::clear_inverted_index_cache() {
```
Lchangliang, line 128: Is there any difference with BetaRowset::remove()?

dataroaring:
run buildall

be/src/olap/rowset/beta_rowset.cpp
dataroaring: remove() delets files too. When a tablet is moved to trash, the files are not deleted immediately. I let remove call clear_cache to reduce duplicated code.

dataroaring:
run buildall

gavinchou (approved):

Lchangliang (approved):

Lchangliang:
LGTM

be/src/olap/base_tablet.h
```diff
@@ -241,6 +241,8 @@ class BaseTablet {
 
     std::vector<RowsetSharedPtr> get_snapshot_rowset(bool include_stale_rowset = false) const;
 
+    virtual void clear_cache() { LOG(INFO) << "should not reach here"; }
```
platoneko, line 244:
```suggestion
    virtual void clear_cache() = 0;
```

be/src/olap/rowset/rowset.h
platoneko: Move this method to `BetaRowset` private field

be/src/olap/tablet.cpp
platoneko, line 1:
```c++
void CloudTablet::clear_cache() override {
    CloudTablet::recycle_cached_data(get_snapshot_rowset(true));
    _engine.tablet_mgr().erase_tablet(tablet_id());
}

void CloudTablet::recycle_cached_data(const std::vector<RowsetSharedPtr>& rowsets) {
    for (auto& rs : rowsets) {
        // Clear cached opened segments and inverted index cache in memory
        rs->clear_cache();
    }

    if (config::enable_file_cache) {
        for (const auto& rs : rowsets) {
            for (int seg_id = 0; seg_id < rs->num_segments(); ++seg_id) {
                auto seg_path = rs->segment_file_path(seg_id);
                auto file_key = io::BlockFileCache::hash(io::Path(seg_path).filename().native());
                auto* file_cache = io::FileCacheFactory::instance()->get_by_path(file_key);
                file_cache->remove_if_cached(file_key);
            }
        }
    }
}
```

platoneko (changes requested):

platoneko (dismissed):

dataroaring:
run buildall

dataroaring:
run buildall

dataroaring:
run buildall

dataroaring:
run buildall

platoneko (approved):

PR #34021, opened 2024-04-23
[fix](group commit) Fix the timeout to wait wal finished when schema change

mymeiyi:
## Proposed changes

1. modify regression case to add some schema change;
2. when schema change, there is a dead lock:
    * the schema change job hold table write lock, and wait for BE wal num is 0
    * the group_commit commit failed because can not get table write lock
    * the replay wal failed because can not get table read lock to generate a plan fragment
    
    so move the wait wal out of the table write lock.
3. If BE encounter memory hard limit, the load will stuck until the hard limit is ended. This may lead to the commit time interval of group commit is much larger than users' config, so improve the check timely.

## Further comments


mymeiyi:
run p0

dataroaring (approved):
LGTM

Yukang-Lian (approved):
LGTM

PR #34040, opened 2024-04-23
[fix](schema change) follow fe set sc fail replicas as bad #33569

yujun777:
cherry pick from #33569


yujun777:
run buildall

yujun777:
run buildall

PR #34055, opened 2024-04-24
[fix](nereids)prevent null pointer access if translate expression fails

starocean999:
## Proposed changes
pick from master https://github.com/apache/doris/pull/33990

## Further comments


starocean999:
run buildall

carlvinhust2012:
can you add some test case？

starocean999:
> can you add some test case？

I found the NPE in some unexpected scenario. Like user defined var is not replaced by literal in right place would cause the translation fail. After the bug is fixed, there is not NPE now. So this is just a fail-safe for unexpected fail in future.

PR #34074, opened 2024-04-24
[Fix](executor)Fix when Fe send empty wg list to be may cause query failed.

wangbo:
Currently FE may send an empty workload group list to be occasionally, this may cause be's workload group is shutdown, but never be deleted.
We need fix this case in BE, when FE send emptylist , be not delete any workload group.
Add more log to location this root reason, whether FE send emtpy list or missing some workload group.
I also find that all fe could send publish request to BE, but we expect only Master FE does this.

### DIFF
--- a/be/src/runtime/workload_group/workload_group_manager.cpp
+++ b/be/src/runtime/workload_group/workload_group_manager.cpp
@@ -83,9 +83,11 @@ void WorkloadGroupMgr::delete_workload_group_by_ids(std::set<uint64_t> used_wg_i
             auto workload_group_ptr = iter->second;
             if (used_wg_id.find(tg_id) == used_wg_id.end()) {
                 workload_group_ptr->shutdown();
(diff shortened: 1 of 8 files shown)


wangbo:
run buildall

be/src/runtime/workload_group/workload_group_manager.cpp
```diff
@@ -83,9 +83,11 @@ void WorkloadGroupMgr::delete_workload_group_by_ids(std::set<uint64_t> used_wg_i
             auto workload_group_ptr = iter->second;
             if (used_wg_id.find(tg_id) == used_wg_id.end()) {
                 workload_group_ptr->shutdown();
+                LOG(INFO) << "[topic_publish_wg] shutdown wg " << tg_id;
```
yiguolei, line 86: rename to wg_id

yiguolei (dismissed):
LGTM

yiguolei (approved):
LGTM

xinyiZzz (approved):
LGTM

PR #34112, opened 2024-04-25
[Fix](partial update) Fix rowset not found error when doing partial update

Yukang-Lian:
## Proposed changes

Issue Number: close #xxx

Issue: When users execute the update statement, they encounter a logic for partial column updates. For large datasets, during the update process, the error "the unmentioned column xxx should have default value or be nullable for newly inserted rows in non-strict mode partial update" may appear.

Cause: In the logic of partial column updates, the existing data columns are read first, and then the data is supplemented and written back. During the reading process, initialization involves initially fetching rowset IDs, and the actual rowset object is fetched only when needed later. However, between fetching the rowset IDs and the rowset object, compaction may occur, turning the old rowset into a stale rowset. If too much time passes, the stale rowset might be directly deleted. Thus, when the rowset object is needed for an update, it cannot be found. Although the update operation with partial column logic should be able to read all keys and should not encounter new keys, if the rowset disappears, the Backend (BE) will consider these keys as missing. Consequently, it will check whether other columns have default values or are nullable. If this check fails, the aforementioned error is thrown.

Solution: To avoid such issues during partial column updates, the initialization step should involve fetching both the rowset IDs and the shared pointer to the rowset object simultaneously. This ensures that the rowset can always be found during data retrieval.

## Further comments


Yukang-Lian:
test will be add soon.

be/src/olap/rowset/segment_v2/vertical_segment_writer.cpp
```diff
@@ -368,24 +368,9 @@ Status VerticalSegmentWriter::_append_block_with_partial_content(RowsInBlock& da
 
     std::vector<RowsetSharedPtr> specified_rowsets;
     {
+        sleep(120);
```
HHoflittlefish777, line 371: why sleep 120?

Yukang-Lian: This is just a draft. This sleep 120 will be added in debug point test.

Yukang-Lian:
run buildall

Yukang-Lian:
run buildall

Yukang-Lian:
run buildall

gavinchou (approved):
LGTM

be/src/olap/rowset_builder.cpp
```diff
@@ -134,10 +135,11 @@ Status BaseRowsetBuilder::init_mow_context(std::shared_ptr<MowContext>& mow_cont
         _rowset_ids.clear();
     } else {
         RETURN_IF_ERROR(tablet()->get_all_rs_id_unlocked(cur_max_version, &_rowset_ids));
+        rowset_ptrs = tablet()->get_rowset_by_ids(&_rowset_ids);
```
zhannngchen, line 138: DCHECK_EQ(_rowset_ids.size(), rowset_ptrs.size());

zhannngchen (approved):
LGTM

PR #34190, opened 2024-04-26
[fix](ES catalog)Make col != '' behavior consistent with SQL (#34151)

qidaye:
## Proposed changes

pick #34151 
In SQL syntax, `col != ''` equals `col.length() > 0`. 
It means that this column must exist in ES doc fields and its content is not empty. 
In this PR, we make a special translation for this binary predicate to keep the behavior of both consistent.

---------

Issue Number: close #xxx

## Further comments


qidaye:
run buildall

PR #34199, opened 2024-04-26
[fix](schema change) fix the defineName field is not the same when copying column

luwei16:
The copy constructor of the column class calls the getDefineName function, which causes the dest column defineName to have a specific value when src column defineName is null. The value of defineName is different after copying. So call getRealDefineName instead

```
public String getDefineName() {
    if (defineName != null) {
        return defineName;
    }
    return name;
}
```

```
public String getRealDefineName() {
    return defineName;
}
```


luwei16:
run buildall

fe/fe-core/src/main/java/org/apache/doris/catalog/Column.java
```diff
@@ -341,6 +341,10 @@ public String getDefineName() {
         return name;
     }
 
+    public String getRealDefineName() {
```
gavinchou, line 344: add comment

gavinchou (changes requested):

regression-test/suites/schema_change_p0/test_add_rename_column.groovy
```diff
@@ -0,0 +1,52 @@
...
+    sql """ ALTER TABLE ${tableName} add column c3 varchar(10) """
+
+    sql """ ALTER TABLE ${tableName} rename column c3 c4 """
+
+    sql """ truncate table ${tableName} """
+
+    qt_sql """ select * from ${tableName} """
+
+    sql """ insert into ${tableName} values (3, 'cc', 'dd') """
+
```
yiguolei, line 50: add sync,
In multi FE env, the result will not stable if not sync

luwei16: done

fe/fe-core/src/main/java/org/apache/doris/catalog/Column.java
luwei16: > add comment

done

luwei16:
run buildall

gavinchou (approved):

dataroaring (approved):
LGTM

PR #34201, opened 2024-04-26
[fix](schema change) fix the defineName field is not the same when copying column

luwei16:
The copy constructor of the column class calls the getDefineName function, which causes the dest column defineName to have a specific value when src column defineName is null. The value of defineName is different after copying. So call getRealDefineName instead

```
public String getDefineName() {
    if (defineName != null) {
        return defineName;
    }
    return name;
}
```

```
public String getRealDefineName() {
    return defineName;
}
```


luwei16:
run buildall

luwei16:
run buildall

yiguolei (approved):

PR #34072, opened 2024-04-24
[fix](pipeline-load) fix no error url when data quality error and total rows is negative

sollhui:
## Proposed changes

when data quality error, there is no error url
```
{
    "TxnId": 1798824628052992,
    "Label": "02b8d5e1-4b90-4b0b-b59f-7f06ebe13d49",
    "Comment": "",
    "TwoPhaseCommit": "false",
    "Status": "Fail",
    "Message": "[CANCELLED][END_OF_FILE]Encountered unqualified data, stop processing",
    "NumberTotalRows": 0,
    "NumberLoadedRows": 0,
    "NumberFilteredRows": 0,
    "NumberUnselectedRows": 0,
    "LoadBytes": 983607,
    "LoadTimeMs": 609,
    "BeginTxnTimeMs": 36,
    "StreamLoadPutTimeMs": 465,
    "ReadDataTimeMs": 1,
    "WriteDataTimeMs": 106,
    "CommitAndPublishTimeMs": 0
}
```

after bug fix:
```
{
    "TxnId": 1798824628052992,
    "Label": "02b8d5e1-4b90-4b0b-b59f-7f06ebe13d49",
    "Comment": "",
    "TwoPhaseCommit": "false",
    "Status": "Fail",
    "Message": "[DATA_QUALITY_ERROR]too many filtered rows",
    "NumberTotalRows": 3065,
    "NumberLoadedRows": 3012,
    "NumberFilteredRows": 53,
    "NumberUnselectedRows": 0,
    "LoadBytes": 983607,
    "LoadTimeMs": 609,
    "BeginTxnTimeMs": 36,
    "StreamLoadPutTimeMs": 465,
    "ReadDataTimeMs": 1,
    "WriteDataTimeMs": 106,
    "CommitAndPublishTimeMs": 0,
    "ErrorURL": "http://127.0.0.1:8249/api/_load_error_log?file=__shard_0/error_log_insert_stmt_bf4ff73cd2d1ff3a-4090cb75416fefb6_bf4ff73cd2d1ff3a_4090cb75416fefb6"
}
```

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #34204, opened 2024-04-27
[fix](pipeline-load) fix no error url when data quality error and total rows is negative #34072

liaoxin01:
cherry pick from #34072


liaoxin01:
run buildall

PR #34065, opened 2024-04-24
[improve](move-memtable) reduce default load stream per node to 2 for stream load

kaijchen:
## Proposed changes

The previous default value `20` is too large, reduce it to `2`.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #34205, opened 2024-04-27
[improve](move-memtable) reduce default load stream per node to 2 for stream load #34065

liaoxin01:
cherry pick from #34065


liaoxin01:
run buildall

PR #34324, opened 2024-04-30
[opt](log)password should not be output in log

390008457:
## Proposed changes

Passwords should not be output to logs in plain text. Instead, use MD5  to protect them.


390008457:
run beut: BE UT

dataroaring (approved):
LGTM

zy-kkk (approved):

PR #34341, opened 2024-04-30
[Fix](executor)Fix stream load IP based auth Failed.

wangbo:
## Proposed changes
Currenty when be request Fe for a plan, Fe miss client's IP and using 'user@%' to auth for workload group, this may cause some user config IP based user auth failed, so we need to use real user's IP to to auth.


morningman (dismissed):
LGTM

wangbo:
run buildall

wangbo:
run buildall

xiaokang:
@wangbo pls add testcase

wangbo:
> @wangbo pls add testcase

I will add test cas but may not in this pr. It's a little complicated.

morningman (approved):

yiguolei (approved):

PR #34374, opened 2024-05-01
[Fix](inverted index) fix memory leak when inverted index writer not close #33724

xiaokang:
cherry pick from #33724


xiaokang:
run buildall

PR #34377, opened 2024-05-02
[Fix](executor)normal group not auth

wangbo:
## Proposed changes
Currently for stream load, FE can not get client IP when auth, for user stream load not failed when upgrade from 2.0 to 2.1, normal group not auth at all.
This is a temporary fix should be removed after FE can get correct for stream load.


wangbo:
run buildall

yiguolei (approved):

morningman (approved):

PR #34385, opened 2024-05-02
[build](third-party) Support to download specified third-party only

gavinchou:
In the previous implementation, we have to download all third-party packages even if we build a specified third-party dependency, e.g.
```
sh build-third-party.sh brpc
```
it would also download all the other irrelevant third-party source files, say ali_sdk and rocskdb.
It wastes a lot of time if we only want to update a specific dependency, especially if the network quality is poor.

After this commit, we only need to download the specified third-party source files, it saves a lot of time when we do third-party updates manually.

### DIFF
--- a/thirdparty/build-thirdparty.sh
+++ b/thirdparty/build-thirdparty.sh
@@ -147,7 +147,7 @@ if [[ "${CLEAN}" -eq 1 ]] && [[ -d "${TP_SOURCE_DIR}" ]]; then
 fi
 
 # Download thirdparties.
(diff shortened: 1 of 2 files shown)


gavinchou:
run buildall

gavinchou:
run buildall

dataroaring (dismissed):
LGTM

gavinchou:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

Issue #32755, opened 2024-03-25
Enhancement of round based function 


zhiqiang-hhhh:
### Discussed in https://github.com/apache/doris/discussions/32086

We need to make round based functions support using column as its scale argument.

- [x] Truncate 
- [x] Round
- [x] Floor
- [x] Ceil

zhiqiang-hhhh:
truncate: https://github.com/apache/doris/pull/32746

superdiaodiao:
May I ask the other three enhancement tasks assigned to anyone? If not, could you assign to me?

zhiqiang-hhhh:
> May I ask the other three enhancement tasks assigned to anyone? If not, could you assign to me?

@superdiaodiao thanks for your attention, they are pending, and they are yours.

Just to let you know, the task is not difficult, but need much explore work in mysql.

Specifically, we need to make sure we have same behaviour with mysql, for example, in function truncate, when using literal as scale argument, its behaviour is different from using column as scale argument. I do not test the others in mysql, but I think they may have same problem.

Tips for implementation:
1. The pression of decimal should not change.
2. The scale of decimal may depend on the specific situation ( need to be same with mysql)
3. Need BEUT & regression test
4. docs update in https://github.com/apache/doris-website

Thanks your attention again, enjoy the process!

superdiaodiao:
> @superdiaodiao thanks for your attention, they are pending, and they are yours.
> 
> Just to let you know, the task is not difficult, but need much explore work in mysql.
> 
> Specifically, we need to make sure we have same behaviour with mysql, for example, in function truncate, when using literal as scale argument, its behaviour is different from using column as scale argument. I do not test the others in mysql, but I think they may have same problem.
> 
> Tips for implementation:
> 1. The pression of decimal should not change.
> 2. The scale of decimal may depend on the specific situation ( need to be same with mysql)
> 3. Need BEUT & regression test
> 4. docs update in https://github.com/apache/doris-website
> 
> Thanks your attention again, enjoy the process!

Alright, thank you very much~~~
I will try my best and may need your review in the future PR.

PR #34391, opened 2024-05-04
[enhancement](function) floor/ceil/round/round_bankers can use column as scale argument

superdiaodiao:
## Proposed changes

Issue Number: https://github.com/apache/doris/issues/32755

Like PR: #32746, we enhanced **floor/ceil/round/round_bankers** can use column as scale argument.

Meanwhile, we moved these round based function register into new file named round.cpp to accelerate compile process and enable code more readable.

What's more, we made truncate() can run with single const argument avoiding BE crash, like branch2.0's behavior or Clickhouse's.

Docs refer PR: https://github.com/apache/doris-website/pull/620

## Further comments


superdiaodiao:
run buildall

superdiaodiao:
run buildall

superdiaodiao:
run fe ut

superdiaodiao:
run feut

superdiaodiao:
run buildall

superdiaodiao:
run buildall

superdiaodiao (commented):
run buildall

be/src/vec/functions/round.h
```diff
@@ -696,49 +679,236 @@ class FunctionRounding : public IFunction {
         return Status::OK();
     }
 
-    ColumnNumbers get_arguments_that_are_always_constant() const override { return {1}; }
+    ColumnNumbers get_arguments_that_are_always_constant() const override { return {}; }
+    // SELECT number, func(123.345, 1) FROM numbers("number"="10")
+    // should NOT behave like two column arguments, so we can not use const column default implementation
+    bool use_default_implementation_for_constants() const override { return false; }
```
HappenLee, line 685: which case should not use default impl of constant? seems wired?

be/src/vec/functions/round.h
```diff
@@ -696,49 +679,236 @@ class FunctionRounding : public IFunction {
         return Status::OK();
     }
 
-    ColumnNumbers get_arguments_that_are_always_constant() const override { return {1}; }
+    ColumnNumbers get_arguments_that_are_always_constant() const override { return {}; }
```
HappenLee, line 682: maybe do not override?

be/src/vec/functions/round.h
superdiaodiao: This part tries to keep the same with truncate's behavior from [PR#32746.](https://github.com/apache/doris/pull/32746) Specifically, it said that:
> Function truncate can use column as its scale argument. The precision infer rule of truncate is same with mysql:
> 
> When using literal as scale, result scale is dependent on the integer literal;
> When using column as scale, result will always have same scale with input decimal.

be/src/vec/functions/round.h
superdiaodiao: OK

be/src/vec/functions/round.h
superdiaodiao: OK, I will optimize.

superdiaodiao:
run buildall

be/src/vec/functions/round.h
zhiqiang-hhhh: @HappenLee  @superdiaodiao for query like `SELECT number, func(123.345, 1) FROM numbers("number"="10")`, we should not use default impl for const. If we do, the result will be like `123.300`, but actually, we need `123.3` as result.

be/src/vec/functions/function_truncate.h
```diff
@@ -1,245 +0,0 @@
...
-
-template <typename Impl>
-class FunctionTruncate : public FunctionRounding<Impl, RoundingMode::Trunc, TieBreakingMode::Auto> {
-public:
-    static FunctionPtr create() { return std::make_shared<FunctionTruncate>(); }
-
-    ColumnNumbers get_arguments_that_are_always_constant() const override { return {}; }
-    // SELECT number, truncate(123.345, 1) FROM number("numbers"="10")
-    // should NOT behave like two column arguments, so we can not use const column default implementation
-    bool use_default_implementation_for_constants() const override { return false; }
```
zhiqiang-hhhh, line 78: `use_default_implementation_for_constants` should be override, and return value should be `false`. The reason is clear in comment.

be/src/vec/functions/round.h
```diff
@@ -696,49 +679,134 @@ class FunctionRounding : public IFunction {
...
             using Types = std::decay_t<decltype(types)>;
             using DataType = typename Types::LeftType;
 
             if constexpr (IsDataTypeNumber<DataType> || IsDataTypeDecimal<DataType>) {
                 using FieldType = typename DataType::FieldType;
-                res = Dispatcher<FieldType, rounding_mode, tie_breaking_mode>::apply_vec_const(
-                        column.column.get(), scale_arg);
+                // the SECOND argument is MISSING or CONST
+                if (arguments.size() == 1 ||
+                    is_column_const(*block.get_by_position(arguments[1]).column)) {
```
zhiqiang-hhhh, line 704: After use_default_impl_for_const returns false, for `SELECT number, truncate(123.345 , 1) FROM numbers("number"="10");` we should go into this branch.

be/src/vec/functions/round.h
```diff
@@ -696,49 +679,134 @@ class FunctionRounding : public IFunction {
...
+                    Int16 scale_arg = 0;
+                    if (arguments.size() == 2) {
+                        RETURN_IF_ERROR(
+                                get_scale_arg(block.get_by_position(arguments[1]), &scale_arg));
+                    }
+                    res = Dispatcher<FieldType, rounding_mode, tie_breaking_mode>::apply_vec_const(
```
zhiqiang-hhhh, line 710: If column_general is const, we will get wrong result if we use this method.

zhiqiang-hhhh (changes requested):
Need to process query like `SELECT number, truncate(123.345 , 1) FROM numbers("number"="10");` correctly.

be/src/vec/functions/round.h
superdiaodiao: Yes, I found it. @HappenLee wished that use constant implement as true, and I have to fix this.

be/src/vec/functions/function_truncate.h
```diff
@@ -1,245 +0,0 @@
...
-                }
-            }
-#endif
-
-            if (!call_on_index_and_data_type<void>(column_general.type->get_type_id(), call)) {
-                return Status::InvalidArgument("Invalid argument type {} for function {}",
-                                               column_general.type->get_name(), "truncate");
-            }
-            // Important, make sure the result column has the same size as the input column
-            res = ColumnConst::create(std::move(res), input_rows_count);
```
zhiqiang-hhhh, line 131: This logical should be reserved.

superdiaodiao: If constant implement is false, indeed it is necessary.

zhiqiang-hhhh:
If we use default implementation for const for round like functions, we will never meet two const arguments in execute_impl, since this property is discarded by common logical.

superdiaodiao:
> If we use default implementation for const for round like functions, we will never meet two const arguments in execute_impl, since this property is discarded by common logical.
> 
> 

Oh, you get the point. No wonder I just wanted to print the log but got nothing. Thanks~~~

superdiaodiao:
run buildall

be/src/vec/functions/round.h
```diff
@@ -630,9 +630,9 @@ struct Dispatcher {
 
             return col_res;
         } else {
-            throw doris::Exception(ErrorCode::INVALID_ARGUMENT,
-                                   "Unsupported column {} for function truncate",
-                                   const_col_general->get_name());
+            LOG(FATAL) << "__builtin_unreachable";
```
zhiqiang-hhhh, line 633: Actually I prefer using Exception instead of `LOG(FATAL)`, because we have no reason to explicitly make BE crash unless this logical error will lead to data incomplete or other problem worse than cluster shut down.

superdiaodiao: I saw vec_const part using fatal, should I change together?

be/test/vec/function/function_round_test.cpp
```diff
@@ -0,0 +1,1133 @@
...
+
+// input, scale_arg, expectation
+using FloatTestDataSet = std::vector<std::tuple<Float64, int, Float64>>;
+
+using DecimalTruncateFunction = FunctionRounding<DecimalRoundTwoImpl<TruncateName>,
+                                                 RoundingMode::Trunc, TieBreakingMode::Auto>;
+using DecimalFloorFunction = FunctionRounding<DecimalRoundTwoImpl<FloorName>, RoundingMode::Floor,
+                                              TieBreakingMode::Auto>;
+using DecimalCeilFunction =
+        FunctionRounding<DecimalRoundTwoImpl<CeilName>, RoundingMode::Ceil, TieBreakingMode::Auto>;
```
zhiqiang-hhhh, line 58: https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_ceil

ceil function in mysql only has one argument.

superdiaodiao: Yes, I noticed, but Doris already has two arguments since long time ago so I keep.

be/src/vec/functions/round.cpp
```diff
@@ -0,0 +1,50 @@
...
+void register_function_round(SimpleFunctionFactory& factory) {
+#define REGISTER_ROUND_FUNCTIONS(IMPL)                                                           \
+    factory.register_function<                                                                   \
+            FunctionRounding<IMPL<TruncateName>, RoundingMode::Trunc, TieBreakingMode::Auto>>(); \
+    factory.register_function<                                                                   \
+            FunctionRounding<IMPL<FloorName>, RoundingMode::Floor, TieBreakingMode::Auto>>();    \
+    factory.register_function<                                                                   \
+            FunctionRounding<IMPL<RoundName>, RoundingMode::Round, TieBreakingMode::Auto>>();    \
+    factory.register_function<                                                                   \
+            FunctionRounding<IMPL<CeilName>, RoundingMode::Ceil, TieBreakingMode::Auto>>();      \
```
zhiqiang-hhhh, line 35: maybe we can not register two arguments for ceil

be/src/vec/functions/round.h
```diff
@@ -564,16 +554,9 @@ struct Dispatcher {
         }
     }
 
-    // NOTE: This function is only tested for truncate
-    // DO NOT USE THIS METHOD FOR OTHER ROUNDING BASED FUNCTION UNTIL YOU KNOW EXACTLY WHAT YOU ARE DOING !!! only test for truncate
     static ColumnPtr apply_const_vec(const ColumnConst* const_col_general,
```
zhiqiang-hhhh, line 557: not ceil

zhiqiang-hhhh:
https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_ceil

ceil function in mysql and pg do not support using second argument.

So, @superdiaodiao could you please do a small modification so that we can detect the above situation for function ceil?

Just need to modify the logic in planner and do not register two argument for ceil on BE.

superdiaodiao:
> https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_ceil
> 
> ceil function in mysql and pg do not support using second argument.
> 
> So, @superdiaodiao could you please do a small modification so that we can detect the above situation for function ceil?
> 
> Just need to modify the logic in planner and do not register two argument for ceil on BE.

Doris has been two arguments for a long time, I think we shouldn't delete this behavior in this enhancement PR. What's more, Clickhouse supports two argument on ceil();

zhiqiang-hhhh:
> > https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_ceil
> > ceil function in mysql and pg do not support using second argument.
> > So, @superdiaodiao could you please do a small modification so that we can detect the above situation for function ceil?
> > Just need to modify the logic in planner and do not register two argument for ceil on BE.
> 
> Doris has been two arguments for a long time, I think we shouldn't delete this behavior in this enhancement PR. What's more, Clickhouse supports two argument on ceil();

ok, we can keep the code in this pr, and we can change behavior in another pr if necessary.

zhiqiang-hhhh (approved):
LGTM, excellent job

superdiaodiao:
run feut

HappenLee (approved):
LGTM

lide-reed (approved):
LGTM

PR #34392, opened 2024-05-04
[opt](ranger) modify and enhance the feature of ranger access controller

morningman:
## Proposed changes

1. Support more data mask
	
	Support data masks same as hive
	
	- MASK
	- MASK_SHOW_LAST_4
	- MASK_SHOW_FIRST_4
	- MASK_HASH
	- MASK_NULL
	- MASK_NONE
	- MASK_DATE_SHOW_YEAR
	- CUSTOM

2. Use `USAGE` priv instead of `SHOW_WORKLOAD_GROUP` and `SHOW_RESOURCES` for Resource and Workload Group

3. When checking the child level priv, also check its parent level priv

	eg, when checking tbl level priv, also check if user has database level priv.
	If user has has database level priv, he should have tbl leve priv automatically.

## Further comments

This should also update the ranger-doris-plugin


fe/fe-core/src/main/java/org/apache/doris/backup/Repository.java
```diff
@@ -746,12 +745,6 @@ public String getCreateStatement() {
         stmtBuilder.append("\nPROPERTIES\n(");
         Map<String, String> properties = new HashMap();
         properties.putAll(this.getRemoteFileSystem().getProperties());
-        // WE should not return the acturl secret key to user for safety consideration
```
morningman, line 749: Use `SENSITIVE_KEY` in `PrintableMap`, so no need to add `secretKeys` here

morningman:
run buildall

fe/fe-core/src/main/java/org/apache/doris/catalog/authorizer/ranger/RangerAccessController.java
```diff
@@ -135,12 +135,27 @@ public Optional<DataMaskPolicy> evalDataMaskPolicy(UserIdentity currentUser, Str
...
+                return Optional.empty();
+            case "CUSTOM":
+                String maskedValue = policy.getMaskedValue();
+                if (StringUtils.isEmpty(maskedValue)) {
+                    return Optional.empty();
+                }
+                return Optional.of(new RangerDataMaskPolicy(currentUser, ctl, db, tbl, col, policy.getPolicyId(),
+                        policy.getPolicyVersion(), maskType, maskedValue.replace("{col}", col)));
+            default:
+                String transformer = policy.getMaskTypeDef().getTransformer();
```
zddr, line 152: if ’maskType‘ is ‘CUSTOM’，’transformer‘ will be null?

zddr (approved):
LGTM

BePPPower (approved):

morningman:
run buildall

CalvinKirs (approved):

PR #34470, opened 2024-05-07
[feature](Nereids): remove useless project

jackwener:
## Proposed changes

If a projection select all columns, we can eliminate it

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/nereids/processor/post/EliminateProject.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/nereids/processor/post/EliminateProject.java
@@ -1,0 +1,51 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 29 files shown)


fe/fe-core/src/main/java/org/apache/doris/nereids/processor/post/EliminateProject.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    public Plan visitPhysicalProject(PhysicalProject<? extends Plan> project, CascadesContext ctx) {
+        Plan child = project.child();
+        Plan newChild = child.accept(this, ctx);
+        if (project.isAllSlots()) {
+            Set<Slot> projects = project.getProjects().stream().map(Slot.class::cast).collect(Collectors.toSet());
+            Set<Slot> outputSet = newChild.getOutputSet();
+            if (outputSet.equals(projects)) {
+                return ((AbstractPhysicalPlan) newChild).copyStatsAndGroupIdFrom(project);
+            }
```
924060929, line 45: I think this two set should change to list.
The project is useless if and only if the order, output number, and each slots are equals.

e.g. the top project maybe use to change the output order:
```sql
-- the tbl output: id, name
select name, id
from tbl
```

jackwener:
run buildall

fe/fe-core/src/main/java/org/apache/doris/nereids/processor/post/EliminateProject.java
jackwener: `ResultSink` already confirm the order of output

924060929:
why not use `EliminateUnnecessaryProject` and change it to support `PhysicalProject`?

PR #34481, opened 2024-05-07
[Fix](executor)Add more log for runtime status report

wangbo:
## Proposed changes
1 Add log for time cost and timeout report query num for report runtime status.
2 catch Fe exception for first report and reopen report


wangbo:
run buildall

wangbo:
run buildall

PR #34494, opened 2024-05-07
[Refactor](executor) rename workload schedule policy to workload policy

wangbo:
Rename workload schedule policy to workload policy,  main changes are in the following aspects:
1 rename metadata class, this may cause upgrade failed if user config policy.
2 rename thrift class.
3 rename schema table from workload_schedule_policy to workload policy.


wangbo:
run buildall

PR #34364, opened 2024-04-30
[fix](csv-reader) fix column split error when there is escape character

liaoxin01:
## Proposed changes

Issue Number: close #xxx

A row of data happened to have an escape character truncated, mistaking the next character for an enclose, resulting in a delimitation error.

**for example:** 
origin data
```
1,2,3,"{~"title~": ~"abced~", ~"id~": 1, ~"name~": ~"user1~", ~"email~": ~"user1@example.com~"}"
```
curl  --location-trusted -u root: -T data -H "format:csv" -H "column_separator:," -H "format:csv" -H 'enclose:"' -H "trim_double_quotes:true" -H 'escape:~' http://127.0.0.1:8040/api/test/load_test_2/_stream_load

Truncated when a row of data is read
```
1,2,3,"{~"title~": ~"abced~", ~"id~": 1, ~"name~": ~
"user1~", ~"email~": ~"user1@example.com~"}"
```
The quotation  before `user1` is considered an enclose character.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #34505, opened 2024-05-08
[fix](csv-reader) fix column split error when there is escape character #34364

liaoxin01:
cherry pick from #34364


liaoxin01:
run buildall

PR #33714, opened 2024-04-16
[fix](tablet invert index) fix tablet invert index leaky caused by auto partition

yujun777:
BUG:

When creating a partition,   it will call InternalCatalog.createPartitionWithIndices  and the partition's tablet will add to tablet invert index.  And drop or discard a partition, need to clean its tablet in tablet invert index.

We found a issue oneline: user create a table with auto partition, they concurrent insert data, and then  concurrent create partitions with the same partition name.  Then the later created partition will raise exists partition eror.  But it forget to clear the related tablets in tablet invert index.  

## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #34511, opened 2024-05-08
[fix](tablet invert index) fix tablet invert index leaky caused by auto partition #33714

yujun777:
cherry pick from #33714


yujun777:
run buildall

PR #34518, opened 2024-05-08
[bugfix](hive)fix error for writing to hive for 2.1

wuwenchi:
## Proposed changes

mirror #34520

When the nameserver specified by `defaultFS` is different from the nameserver of the actual table, write failure may occur.
Moreover, when the orc writing fails, it still goes to the commit process (actually, after the writing fails, FE should be able to get the error message from BE, and should not go to the commit process)

Fix:
1. When orc writing fails, an exception is thrown. The upstream captures the exception and returns it to status for FE to obtain.
2. When creating fs in the writing process, if the writing path contains a nameserver, then use this nameserver directly to create fs. (In this way, when writing to hive, there is no need to configure defaultFS in the catalog)

## Further comments


be/src/io/file_factory.cpp
```diff
@@ -94,6 +94,16 @@ Status FileFactory::create_file_writer(TFileType::type type, ExecEnv* env,
...
         THdfsParams hdfs_params = parse_properties(properties);
         std::shared_ptr<io::HdfsFileSystem> fs;
+        // If the destination path contains a schema, use the schema directly.
+        // If not, use defaultFS.
+        // Otherwise a write error will occur.
+        string::size_type idx = path.find("://");
+        if (idx != string::npos) {
+            idx = path.find("/", idx + 3);
+            if (idx != string::npos) {
+                hdfs_params.fs_name = path.substr(0, idx);
```
wsjz, line 104: give a comment to show that parse fs_name from hdfs://host:port and hdfs://dfsnameservice

PR #34547, opened 2024-05-08
[fix](txn insert) Txn insert stuck if execute on observer fe

mymeiyi:
## Proposed changes

When client use txn insert in legacy planner and connect to observer fe, it will stuck. 
Because the packet is null and fe does not return to client.
The reason of packet is null is that the statement does not forward to master.

## Further comments


mymeiyi:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #33897, opened 2024-04-19
Use jemalloc's experimental features to hook jemalloc

adonis0147:
## Proposed changes

Issue Number: close #33570

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #32865, opened 2024-03-26
[Fix](jdk) Fix jdk17 crash on some envs.

kaka11chen:
## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #34578, opened 2024-05-09
[fix](memory) Fix jdk17 and jemalloc hook not compatible on some envs

xinyiZzz:
## Proposed changes

Now in order to support malloc hook, compiling jemalloc will add tow compilation options `--with-jemalloc-prefix=je` and `--disable-cxx`, then overwrite malloc/free as follows to allow BE to use jemalloc:
```
#define ALIAS(doris_fn) __attribute__((alias(#doris_fn), used))
void* malloc(size_t size) __THROW ALIAS(doris_malloc);
void free(void* p) __THROW ALIAS(doris_free);
void* realloc(void* p, size_t size) __THROW ALIAS(doris_realloc);
void* calloc(size_t n, size_t size) __THROW ALIAS(doris_calloc);
void cfree(void* ptr) __THROW ALIAS(doris_cfree);
```

but after such overwrite like this, Doris BE on jdk17 cannot be started in some environments.
there are three solutions:
1. Modify overwrite malloc/free, fixed in https://github.com/apache/doris/pull/33897
2. Modify BE dynamically loads JVM, fixed in https://github.com/apache/doris/pull/32865
3. Commit of this PR, modify jemalloc compilation options, after remove `--with-jemalloc-prefix=je`, there is not need to overwrite malloc/free, BE will use Jemalloc by default. 

So,  if need to use memory hook, this PR still cannot solve this problem.
Looking forward to modify BE dynamically loads JVM, which will not only solve this problem, also fix other crashes caused by BE use JVM.

## Further comments


xinyiZzz:
run buildall

xinyiZzz:
run buildall

be/src/runtime/fragment_mgr.cpp
```diff
@@ -837,6 +837,26 @@ Status FragmentMgr::exec_plan_fragment(const TPipelineFragmentParams& params,
     std::shared_ptr<QueryContext> query_ctx;
     RETURN_IF_ERROR(_get_query_ctx(params, params.query_id, true, query_ctx));
     SCOPED_ATTACH_TASK_WITH_ID(query_ctx->query_mem_tracker, params.query_id);
+    LOG(INFO) << "222222 ";
```
yiguolei, line 840: remove this log

xinyiZzz: sorry, i forget..

xinyiZzz:
run buildall

xinyiZzz:
run buildall

morningman (approved):
LGTM

yiguolei (approved):

PR #34606, opened 2024-05-09
[Fix](nereids) modify the binding aggregate function in order by

feiniaofeiafei:
Do same job with pr #32758 and pr #33843.
Because the master branch2.0 differ in binding order by expressions. So this pr is not a cherry-pick, but actually does same job.
For each order by expression:
If it does not have aggregate functions, firstly bind expression from aggregate outputs. If not found, then bind expression from aggregate child outputs.
If it has aggregate functions, firstly bind expression from aggregate outputs which do not have aggregate functions. If not found, then bind expression from aggregate child outputs.


feiniaofeiafei:
run buildall

feiniaofeiafei:
run buildall

feiniaofeiafei:
run feut

924060929 (approved):

PR #34666, opened 2024-05-10
[chore](test) nereids support window function but some cases does not open yet (#33098)

LiBinfeng-01:
## Proposed changes

cherry-pick #33098


LiBinfeng-01:
run buildall

PR #34244, opened 2024-04-28
[improve](shcema change)fix alter table faild when  modify multiple columns with column changed positions

kobe6th:
## Proposed changes

Issue Number: close #xxx

when try to modify multiple columns with changed positions, attempting to modify a column that has already been moved will result in a 'column not found' error.

CREATE TABLE `test` (
`id` BIGINT NOT NULL,
`v1` BIGINT NULL,
`v2` VARCHAR(128) NULL,
`v3` VARCHAR(128) NULL
 ) ENGINE=OLAP
 UNIQUE KEY(`id`)
 DISTRIBUTED BY HASH(`id`) BUCKETS 1
 PROPERTIES (
 "replication_allocation" = "tag.location.default: 1"
 );
Query OK, 0 rows affected (0.00 sec)

ALTER TABLE `test` MODIFY COLUMN `v2` VARCHAR(512) NULL AFTER `v1`, MODIFY COLUMN `v3` VARCHAR(512) NULL AFTER `v2`;
ERROR 1105 (HY000): errCode = 2, detailMessage = Column[v2] does not exists

when column datatype change, shadow prefix will be add to column name, we can use NonShadowName to find this column.

PR #34687, opened 2024-05-11
[branch-1.2](shcema change)fix alter table faild when modify multiple columns with column changed positions

kobe6th:
cherry pick from #34244

Problem and Cause:
when try to modify multiple columns with changed positions, attempting to modify a column that has already been moved will result in a 'column not found' error.

Solved:
when column datatype change, shadow prefix will be add to column name, we can use NonShadowName to find this column.


kobe6th:
run buildall

PR #34705, opened 2024-05-11
[feature](ui) web profile download use plain text

Mryange:
## Proposed changes

download the plain text of the profile, reduce the overhead of  string replace , converting JSON format.
```java
    @RequestMapping(path = "/query_profile/{" + ID + "}", method = RequestMethod.GET)
    public Object profile(@PathVariable(value = ID) String id) {
        String profile = ProfileManager.getInstance().getProfile(id);
        if (profile == null) {
            return ResponseEntityBuilder.okWithCommonError("ID " + id + " does not exist");
        }
        profile = profile.replaceAll("\n", "</br>");
        profile = profile.replaceAll(" ", "&nbsp;&nbsp;");
        return ResponseEntityBuilder.ok(profile);
    }
```

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/httpv2/controller/QueryProfileController.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/httpv2/controller/QueryProfileController.java
@@ -59,6 +59,15 @@ public class QueryProfileController extends BaseController {
         return ResponseEntityBuilder.ok(profile);
     }
 
+    @RequestMapping(path = "/query_profile/text/{" + ID + "}", method = RequestMethod.GET)
+    public Object profileText(@PathVariable(value = ID) String id) {
+        String queryProfileStr = ProfileManager.getInstance().getProfile(id);
+        if (queryProfileStr == null) {
+            return "query id " + id + " not found";
+        }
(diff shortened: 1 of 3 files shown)


Mryange:
run buildall

fe/fe-core/src/main/java/org/apache/doris/httpv2/controller/QueryProfileController.java
```diff
@@ -59,6 +59,15 @@ public Object profile(@PathVariable(value = ID) String id) {
         return ResponseEntityBuilder.ok(profile);
     }
 
+    @RequestMapping(path = "/query_profile/text/{" + ID + "}", method = RequestMethod.GET)
```
yiguolei, line 62: when this API will be used?

Mryange: web use download

HappenLee, line 62: we better improve the performance of `replaceAll("\n", "</br>");`

PR #34712, opened 2024-05-11
[fix](multi-table-load) fix be core when partial table load failed

sollhui:
### bug
When using single stream multi table load, be will core dump if partial table load failed.

### reason
After some tables load fail, it will cause **exec_task** to return, leading to ctx destruction. When the execution plans of other tables are called back, it will cause the core.

### solution
Wait all table plans exec finish.

### DIFF
--- a/be/src/io/fs/multi_table_pipe.cpp
+++ b/be/src/io/fs/multi_table_pipe.cpp
@@ -32,6 +32,7 @@
 #include "runtime/fragment_mgr.h"
 #include "runtime/runtime_state.h"
 #include "runtime/stream_load/new_load_stream_mgr.h"
+#include "util/debug_points.h"
 #include "util/thrift_rpc_helper.h"
 #include "util/thrift_util.h"
 #include "util/time.h"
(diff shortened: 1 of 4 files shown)


HHoflittlefish777:
run buildall

be/src/io/fs/multi_table_pipe.cpp
```diff
@@ -250,6 +249,8 @@ Status MultiTablePipe::exec_plans(ExecEnv* exec_env, std::vector<ExecParam> para
             CHECK(false);
         }
 
+        _inflight_cnt++;
+
         RETURN_IF_ERROR(exec_env->fragment_mgr()->exec_plan_fragment(
```
liaoxin01, line 254: if exec_plan_fragment return an error, we shouldn't add 1 to _inflight_cnt.

HHoflittlefish777:
run buildall

liaoxin01 (dismissed):
LGTM

HHoflittlefish777:
run buildall

HHoflittlefish777:
run buildall

XuJianxu (approved):
LGTM

liaoxin01 (approved):

PR #34773, opened 2024-05-13
Buf fix in be broker module.

XuPengfei-1020:
## Proposed changes

Issue Number: close #xxx

This pr contains tow small bug fix about snapshop_loader and broker module.

1. Uncheckd erase action of vector may cause be run failure of core dump.
2. Close the local file writer after data has been wrote.

### DIFF
--- a/be/src/io/fs/broker_file_system.cpp
+++ b/be/src/io/fs/broker_file_system.cpp
@@ -416,7 +416,7 @@ Status BrokerFileSystem::download_impl(const Path& remote_file, const Path& loca
         RETURN_IF_ERROR(local_writer->append({read_buf.get(), read_len}));
     } // file_handler should be closed before calculating checksum
 
-    return Status::OK();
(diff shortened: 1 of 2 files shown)


gavinchou (approved):

be/src/io/fs/broker_file_system.cpp
```diff
@@ -415,6 +415,7 @@ Status BrokerFileSystem::download_impl(const Path& remote_file, const Path& loca
 
         RETURN_IF_ERROR(local_writer->append({read_buf.get(), read_len}));
     } // file_handler should be closed before calculating checksum
+    RETURN_IF_ERROR(local_writer->close());
```
gavinchou, line 418: we may just `return local_writer->close()`
line 420 can be ommited

XuPengfei-1020: fixed

PR #22923, opened 2023-08-14
[Improvement] Optimize count operation for iceberg

wuwenchi:
## Proposed changes

Iceberg has its own metadata information, which includes count statistics for table data. If the table does not contain equli'ty delete, we can get the count data of the current table directly from the count statistics.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #34775, opened 2024-05-13
[Opt](Iceberg) Only initialize one split if the statement can push down count

zhangbutao:
## Proposed changes
https://github.com/apache/doris/pull/22923 did a good optimization for iceberg count. I think we can end the `get splits loop `early as one split is enough if the statement can push down count. This can reduce the query time if iceberg table has many splits. 

Issue Number: close #xxx

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/datasource/iceberg/source/IcebergScanNode.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/datasource/iceberg/source/IcebergScanNode.java
@@ -212,60 +212,73 @@ public class IcebergScanNode extends FileQueryScanNode {
         HashSet<String> partitionPathSet = new HashSet<>();
         boolean isPartitionedTable = icebergTable.spec().isPartitioned();
 
+        TPushAggOp aggOp = getPushDownAggNoGroupingOp();
+        boolean canPushCount = aggOp.equals(TPushAggOp.COUNT) && getCountFromSnapshot() > 0;
+
         CloseableIterable<FileScanTask> fileScanTasks = TableScanUtil.splitFiles(scan.planFiles(), splitSize);
         try (CloseableIterable<CombinedScanTask> combinedScanTasks =
                 TableScanUtil.planTasks(fileScanTasks, splitSize, 1, 0)) {
-            combinedScanTasks.forEach(taskGrp -> taskGrp.files().forEach(splitTask -> {
-                String dataFilePath = normalizeLocation(splitTask.file().path().toString());
+            for (CombinedScanTask taskGrp : combinedScanTasks) {
+                for (FileScanTask splitTask : taskGrp.files()) {
+                    String dataFilePath = normalizeLocation(splitTask.file().path().toString());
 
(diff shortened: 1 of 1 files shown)


fe/fe-core/src/main/java/org/apache/doris/datasource/iceberg/source/IcebergScanNode.java
```diff
@@ -212,60 +212,73 @@ private List<Split> doGetSplits() throws UserException {
...
-                        Object obj = structLike.get(i, Object.class);
-                        String value = String.valueOf(obj);
-                        PartitionField partitionField = fields.get(i);
-                        if (partitionField.transform().isIdentity()) {
-                            Type type = structType.fieldType(partitionField.name());
-                            if (type != null && type.typeId().equals(Type.TypeID.DATE)) {
-                                // iceberg use integer to store date,
-                                // we need transform it to string
-                                value = DateTimeUtil.daysToIsoDate((Integer) obj);
+            for (CombinedScanTask taskGrp : combinedScanTasks) {
```
zhangbutao, line 221: I didn't find a better way to end the loop in `forEach`, so i use `for ` to replace `forEach`. :(

fe/fe-core/src/main/java/org/apache/doris/datasource/iceberg/source/IcebergScanNode.java
```diff
@@ -212,60 +212,73 @@ private List<Split> doGetSplits() throws UserException {
...
+                        split.setDeleteFileFilters(getDeleteFileFilters(splitTask));
                     }
+                    split.setTableFormatType(TableFormatType.ICEBERG);
+                    splits.add(split);
 
-                    // Counts the number of partitions read
-                    partitionPathSet.add(structLike.toString());
+                    // End loop early as one split is enough if the statement can push down count
+                    if (canPushCount) {
+                        break;
```
zhangbutao, line 269: This is what i want to do. End the entire loop early to avoid a lot of useless spilts if the statement can push down count.

zhangbutao:
run buildall

zhangbutao:
@wuwenchi Could you give some suggestion about this change? Thanks.

morningman (commented):
If you want to do this optimization, why not just create a dummy IcebergSplit?
So we even don't need to call `TableScanUtil.planTasks`

zhangbutao:
> If you want to do this optimization, why not just create a dummy IcebergSplit? So we even don't need to call `TableScanUtil.planTasks`

@morningman Thanks for your suggestion! You are right, create a dummy IcebergSplit is a better approach than this PR. But i found that BE need a real iceberg spilt to do some code logic, and we need do some odd check in BE side to let the BE accept the dummy IcebergSplit.

Acutually, i think the `count `pushdown is a metadata operation which can be done in fe side, and so we can skip to initialize the split and no need the operation in BE side. This can be improve increase efficiency of `count `operation.

I just submitted a new PR https://github.com/apache/doris/pull/34928, Please take a look if you have time.
Thanks.

PR #34778, opened 2024-05-13
[opt](routine-load) optimize routine load task allocation algorithm

sollhui:
The current task allocation algorithm takes into account the cache information of the object pool. If there are still more than half of the idle slots in the previously executed BE, they will be allocated to the previous BE.

However, in the actual process of execution in the production environment, there have been serious imbalances.

So we optimized the algorithm based on the following reasons:
1. We believe that the benefits of load balance outweigh the benefits of object pool cache, we try to find the one with the most idle slots as much as possible.
2. On the basis of selecting the maximum idle slot be, try to reuse the object cache as much as possible.


HHoflittlefish777:
run buildall

HHoflittlefish777:
run buildall

liaoxin01 (approved):
LGTM

XuJianxu (approved):
LGTM

liaoxin01 (approved):

liaoxin01:
run performance

PR #34852, opened 2024-05-14
[fix](profile) Load profile need to be registered to get real-time profile

zhiqiang-hhhh:
Profile need to be registered not only when it is finished, but also when it is running. So that user could get realtime profile of Load command.


zhiqiang-hhhh:
run buildall

yiguolei (approved):

mrhhsg (approved):
LGTM

PR #34871, opened 2024-05-14
[Fix](bloom filter) Fix bloom filter memory leak

Yukang-Lian:
## Proposed changes

Issue Number: close #xxx

Issue: Doris occasionally encounters an issue where memory usage becomes exceptionally high and does not decrease. The leaked memory is occupied by Bloom filters stored in memory.

Reason: The segment cache stores segment objects read from files into memory. It functions as an LRU cache with an eviction strategy: when the number of segments exceeds the maximum number, or the total memory size of segment objects in the cache exceeds the maximum usage, it evicts the older segments. However, there is a piece of logic in the code that first reads the segment object into memory, assuming it occupies memory size A, then places the read segment object into the cache (at this point, the cache considers the segment object size to be A). It then reads the segment's Bloom filter from the file and assigns it to the segment's Bloom filter member variable, assuming the Bloom filter occupies memory size B. Thus, the total size of the segment object at this point is A+B. However, the cache does not update this size, leading to the actual size of the segment object stored in the cache (A+B) being larger than the size considered by the cache (A). When the number of segment objects in the cache increases to a certain extent, the used memory will surge dramatically. However, the cache does not perceive the size as reaching the eviction limit, so it does not evict the segment objects. In such cases, a memory leak issue arises.

Solution: Since each segment object only reads the Bloom filter once, the issue can be resolved by changing the logic from reading the segment, placing it into the cache, and then reading the Bloom filter to reading the segment, reading the Bloom filter, and then placing it into the cache.

## Further comments

### DIFF
--- a/be/src/olap/base_tablet.cpp
+++ b/be/src/olap/base_tablet.cpp
@@ -541,7 +541,7 @@ Status BaseTablet::lookup_row_key(const Slice& encoded_key, bool with_seq_col,
         if (UNLIKELY(segment_caches[i] == nullptr)) {
             segment_caches[i] = std::make_unique<SegmentCacheHandle>();
             RETURN_IF_ERROR(SegmentLoader::instance()->load_segments(
-                    std::static_pointer_cast<BetaRowset>(rs), segment_caches[i].get(), true));
+                    std::static_pointer_cast<BetaRowset>(rs), segment_caches[i].get(), true, true));
         }
         auto& segments = segment_caches[i]->get_segments();
         DCHECK_EQ(segments.size(), num_segments);
--- a/be/src/olap/rowset/segment_v2/segment.cpp
+++ b/be/src/olap/rowset/segment_v2/segment.cpp
@@ -284,11 +284,13 @@ Status Segment::_parse_footer(SegmentFooterPB* footer) {
 
 Status Segment::_load_pk_bloom_filter() {
 #ifdef BE_TEST
-    // for BE UT "segment_cache_test"
-    return _load_pk_bf_once.call([this] {
-        _meta_mem_usage += 100;
-        return Status::OK();
-    });
+    if (_pk_index_meta == nullptr) {
+        // for BE UT "segment_cache_test"
+        return _load_pk_bf_once.call([this] {
+            _meta_mem_usage += 100;
+            return Status::OK();
+        });
+    }
 #endif
     DCHECK(_tablet_schema->keys_type() == UNIQUE_KEYS);
     DCHECK(_pk_index_meta != nullptr);
--- a/be/src/olap/rowset/segment_v2/segment.cpp
+++ b/be/src/olap/rowset/segment_v2/segment.cpp
@@ -283,6 +283,13 @@ Status Segment::_parse_footer(SegmentFooterPB* footer) {
 }
 
 Status Segment::_load_pk_bloom_filter() {
+#ifdef BE_TEST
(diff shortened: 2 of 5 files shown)


Yukang-Lian:
run buildall

Yukang-Lian:
run buildall

be/src/olap/rowset/segment_v2/segment.cpp
```diff
@@ -290,6 +290,10 @@ Status Segment::_load_pk_bloom_filter() {
         return _load_pk_bf_once.call([this] {
             RETURN_IF_ERROR(_pk_index_reader->parse_bf(_file_reader, *_pk_index_meta));
             _meta_mem_usage += _pk_index_reader->get_bf_memory_size();
+            SegmentCache::CacheKey cache_key(_rowset_id, _segment_id);
```
zhannngchen, line 293: It's better to decouple cache management and load index/bloomfilter

zhannngchen: Sometimes, we might call `_load_pk_bloom_filter()` on `Segment` objects that not in cache. Not all Segment object should in SegmentCache

Yukang-Lian:
run buildall

Yukang-Lian:
run buildall

be/src/olap/base_tablet.cpp
```diff
@@ -548,6 +553,15 @@ Status BaseTablet::lookup_row_key(const Slice& encoded_key, bool with_seq_col,
 
         for (auto id : picked_segments) {
             Status s = segments[id]->lookup_row_key(encoded_key, with_seq_col, with_rowid, &loc);
+            SegmentCache::CacheKey cache_key(rs->rowset_id(), segments[id]->id());
+            auto cache_value = std::make_unique<SegmentCache::CacheValue>();
+            cache_value->segment = segments[id];
+            try {
+                SegmentLoader::instance()->update_segment(cache_key, *cache_value);
```
zhannngchen, line 560: `load_pk_index_and_bf` only would be called once, but `lookup_row_key` might be called for billion times, if you also call `update_segment` for so many times ,the LRU write lock would become bottleneck.

Yukang-Lian:
run buildall

be/src/olap/base_tablet.cpp
```diff
@@ -538,10 +538,10 @@ Status BaseTablet::lookup_row_key(const Slice& encoded_key, bool with_seq_col,
             continue;
         }
 
-        if (UNLIKELY(segment_caches[i] == nullptr)) {
+        if (LIKELY(segment_caches[i] == nullptr)) {
```
zhannngchen, line 541: should be UNLIKELY

zhannngchen (dismissed):
LGTM

Yukang-Lian:
run buildall

Yukang-Lian:
run buildall

Yukang-Lian:
run buildall

Yukang-Lian:
run buildall

Yukang-Lian:
run p0

zhannngchen (dismissed):

Yukang-Lian:
run buildall

Yukang-Lian:
run buildall

Yukang-Lian:
run buildall

Yukang-Lian:
run buildall

zhannngchen (approved):
LGTM

dataroaring (approved):
LGTM

PR #34913, opened 2024-05-15
[fix](mow) the recycling timing of compaction input rowsets should rely on config compacted_rowset_retention_seconds

hust-hhb:
## Proposed changes
now in cloud mode, compaction input rowsets's recycling timing  is rely on retention_seconds, and this retention_seconds should only rely on config compacted_rowset_retention_seconds to avoid deleting these rowsets too early, becasue these rowset is reading by last loading task to calculate delete bitmap. 
Issue Number: close #xxx

## Further comments


hust-hhb:
run buildall

hust-hhb:
after discusssion, it's better not change this, make retention_seconds enough big can solve this problem

PR #31742, opened 2024-03-04
[refactor](profile)Refactor active_queries() schema

wangbo:
## Proposed changes

Refactor active_queries() schema.

```
mysql [(none)]>desc function active_queries();
+------------------+--------+------+-------+---------+-------+
| Field            | Type   | Null | Key   | Default | Extra |
+------------------+--------+------+-------+---------+-------+
| QueryId          | TEXT   | No   | false | NULL    | NONE  |
| StartTime        | TEXT   | No   | false | NULL    | NONE  |
| QueryTimeMs      | BIGINT | No   | false | NULL    | NONE  |
| WorkloadGroupId  | BIGINT | No   | false | NULL    | NONE  |
| Database         | TEXT   | No   | false | NULL    | NONE  |
| FrontendInstance | TEXT   | No   | false | NULL    | NONE  |
| Sql              | TEXT   | No   | false | NULL    | NONE  |
+------------------+--------+------+-------+---------+-------+
7 rows in set (0.00 sec)
```

PR #34953, opened 2024-05-16
[fix](show proc) Fix insert into command cannot be seen in show proc

zhiqiang-hhhh:
fix insert into command can not be shown by `show proc '/current_query_stmts'`;

reference https://github.com/apache/doris/pull/31742/


zhiqiang-hhhh:
run buildall

PR #34978, opened 2024-05-16
[Fix](executor)Fix query has no coord add may cause report failed.

wangbo:
## Proposed changes

1 Add a column query_type for schema table ```backend_active_tasks```.
2 For external query, they has no coord addr, should skip it when report audit log to Fe.
3 Add strictly check for coord addr and QueryType in debug mode.

### DIFF
--- a/be/src/runtime/runtime_query_statistics_mgr.h
+++ b/be/src/runtime/runtime_query_statistics_mgr.h
@@ -44,7 +44,8 @@ class Block;
 
 class QueryStatisticsCtx {
 public:
-    QueryStatisticsCtx(TNetworkAddress fe_addr) : _fe_addr(fe_addr) {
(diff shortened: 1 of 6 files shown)


wangbo:
run buildall

be/src/runtime/runtime_query_statistics_mgr.h
```diff
@@ -57,6 +58,7 @@ class QueryStatisticsCtx {
     std::vector<std::shared_ptr<QueryStatistics>> _qs_list;
     bool _is_query_finished;
     TNetworkAddress _fe_addr;
+    TQueryType::type _query_type;
```
yiguolei, line 61: It should be const member

wangbo:
run buildall

yiguolei (approved):

mrhhsg (approved):
LGTM

PR #35029, opened 2024-05-17
[fix](catalog) fix the that failed to check if input format is splitable

morningman:
## Proposed changes

Introduced from #33242
When we check supported inputformat in a `Set<String>`, we should use string, not object

## Further comments


morningman:
run buildall

PR #35073, opened 2024-05-20
[profile](pipeline) Add key metrics for pipeline initialization

Gabriel39:
## Proposed changes

image

## Further comments

### DIFF
--- a/be/src/pipeline/pipeline_fragment_context.cpp
+++ b/be/src/pipeline/pipeline_fragment_context.cpp
@@ -221,89 +221,103 @@ Status PipelineFragmentContext::prepare(const doris::TPipelineFragmentParams& re
     if (_prepared) {
         return Status::InternalError("Already prepared");
     }
-    _num_instances = request.local_params.size();
-    _total_instances = request.__isset.total_instances ? request.total_instances : _num_instances;
     _runtime_profile = std::make_unique<RuntimeProfile>("PipelineContext");
     _prepare_timer = ADD_TIMER(_runtime_profile, "PrepareTime");
     SCOPED_TIMER(_prepare_timer);
+    _build_pipelines_timer = ADD_TIMER(_runtime_profile, "BuildPipelinesTime");
+    _init_context_timer = ADD_TIMER(_runtime_profile, "InitContextTime");
+    _plan_local_shuffle_timer = ADD_TIMER(_runtime_profile, "PlanLocalShuffleTime");
+    _build_tasks_timer = ADD_TIMER(_runtime_profile, "BuildTasksTime");
+    _prepare_all_pipelines_timer = ADD_TIMER(_runtime_profile, "PrepareAllPipelinesTime");
+    {
+        SCOPED_TIMER(_init_context_timer);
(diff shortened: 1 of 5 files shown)


Gabriel39:
run buildall

be/src/runtime/fragment_mgr.cpp
```diff
@@ -266,6 +266,14 @@ void FragmentMgr::coordinator_callback(const ReportStatusRequest& req) {
             }
 
             if (enable_profile) {
+                DCHECK(req.profile != nullptr);
+                TDetailedReportParams detailed_param;
+                detailed_param.__isset.fragment_instance_id = false;
+                detailed_param.__isset.profile = true;
+                detailed_param.__isset.loadChannelProfile = false;
```
yiguolei, line 273: why loadChannelProfile is always false?

be/src/pipeline/pipeline_fragment_context.cpp
```diff
@@ -1568,7 +1586,7 @@ Status PipelineFragmentContext::send_report(bool done) {
     ReportStatusRequest req {true,
                              exec_status,
                              runtime_states,
-                             nullptr,
+                             _runtime_profile.get(),
```
yiguolei, line 1589: why it is nullptr in the past?

be/src/runtime/fragment_mgr.cpp
Gabriel39: Because load profile reporting will not reach here.

be/src/pipeline/pipeline_fragment_context.cpp
Gabriel39: This is for non-pipeline profile if pipeline mode is off

zhiqiang-hhhh: will non-pipeline go into this method?

Gabriel39: > will non-pipeline go into this method?

No. But non-pipeline will use this `ReportStatusRequest ` struct so we keep this field here.

zhiqiang-hhhh (commented):
Basically the `_runtime_profile` in PipelineFragmentContext is now used for FragmentLevel profile, so it is passed to `ReportStatusRequest`.

Unfortunately, we have two profile reporting mechanism ... so same modification need to be applied to `PipelineFragmentContext::collect_realtime_profile_x` and `RuntimeQueryStatiticsMgr::create_report_exec_status_params_x`

Gabriel39:
> Basically the `_runtime_profile` in PipelineFragmentContext is now used for FragmentLevel profile, so it is passed to `ReportStatusRequest`.
> 
> Unfortunately, we have two profile reporting mechanism ... so same modification need to be applied to `PipelineFragmentContext::collect_realtime_profile_x` and `RuntimeQueryStatiticsMgr::create_report_exec_status_params_x`

I think we need an overall code refactoring for this to unify query profile and load profile. This is not aimed in this PR.

BiteTheDDDDt (approved):

Mryange (approved):

PR #35081, opened 2024-05-20
[opt](parquet-writer) Specify the row group size when writing data to Parquet files.

BePPPower:
Previously, each Block data would write a row group every time when data was written to a Parquet file.
Which causing too many row group in a single parquet file. The the reading performance on this kind of file
is very bad.

This PR use Arrow's `WriteRecordBatch` interface, which allows caching data to be written to Parquet.
We can control the size of the row group cache."

Currently, simply accumulate the size of the blocks to determine the cache size.

### DIFF
--- a/be/src/vec/runtime/vparquet_transformer.cpp
+++ b/be/src/vec/runtime/vparquet_transformer.cpp
@@ -70,6 +70,8 @@
 
 namespace doris::vectorized {
 
+const uint64_t min_row_group_size = 128 * 1024 * 1024; // 128MB
+
 ParquetOutputStream::ParquetOutputStream(doris::io::FileWriter* file_writer)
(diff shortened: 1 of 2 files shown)


BePPPower:
run buildall

BePPPower:
run p0

BePPPower:
run buildall

morningman (dismissed):
LGTM

kaka11chen:
This solution will consume a lot of memory, so we need to discuss a better solution. @BePPPower

BePPPower:
run buildall

BePPPower:
run performance

morningman (approved):
LGTM

kaka11chen (approved):
LGTM

morningman (approved):

PR #35095, opened 2024-05-21
[fix](FragmentMgr) Should not use operator[] for _query_ctx_map

zhiqiang-hhhh:
Ref: https://en.cppreference.com/w/cpp/container/unordered_map/operator_at

Returns a reference to the value that is mapped to a key equivalent to key or x respectively, performing an insertion if such key does not already exist.

The original code would make `_query_ctx_map` has key `0-0` and value null-pointer.


zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run cloud_p1

zhiqiang-hhhh:
image

crash on p0 is not related to this pr, re run

zhiqiang-hhhh:
run p0

yiguolei (approved):

HappenLee (approved):
LGTM

PR #35105, opened 2024-05-21
[fix](broker load) Make Config.enable_pipeline_load works as expected for BrokerLoad

zhiqiang-hhhh:
Ref https://github.com/apache/doris/blob/a8af8104f016c691674ceb1817620b8354270dcb/fe/fe-core/src/main/java/org/apache/doris/load/loadv2/LoadLoadingTask.java#L155

Default value of enablePipelineEngine in Coordinator is false, but Config.enable_pipeline_load is true by default. This leading to some unexpected behaviours.


zhiqiang-hhhh:
run buildall

yiguolei (approved):

liaoxin01 (approved):
LGTM

PR #35110, opened 2024-05-21
[profile](Brokerload) Support get broker load profile

zhiqiang-hhhh:
Profile of broker load can not be collected since its profile structure is different from query. The id of BrokerLoad Profile is a string representation of long type, and its real id used for profile updating should be LoadLoadingTask id.

This pr fixed the above problem.

A known bug is: when broker load is cancelled, its profile state can not  be updated correctly.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/common/util/ProfileManager.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/util/ProfileManager.java
@@ -17,6 +17,7 @@
 
 package org.apache.doris.common.util;
 
+import org.apache.doris.catalog.Env;
 import org.apache.doris.common.AnalysisException;
 import org.apache.doris.common.AuthenticationException;
 import org.apache.doris.common.ClientPool;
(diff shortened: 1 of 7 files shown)


zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

fe/fe-core/src/main/java/org/apache/doris/common/util/ProfileManager.java
```diff
@@ -326,28 +328,72 @@ private static TGetRealtimeExecStatusResponse getRealtimeQueryProfile(
...
+                    tmp.beAddress = addr;
+                    involvedBackends.add(tmp);
+                }
             }
+        } else {
+            Long loadJobId = (long) -1;
+            try {
+                loadJobId = Long.parseLong(id);
+            } catch (Exception e) {
+                return futures;
```
yiguolei, line 358: 在java 里，遇到错误直接抛异常。不要返回这种invalid的值。

yiguolei (approved):

mrhhsg (approved):
lgtm

PR #35115, opened 2024-05-21
[compile](mac) Fix compile on mac os

zhiqiang-hhhh:


zhiqiang-hhhh:
run buildall

adonis0147 (changes requested):
I don't think this fix is suitable.

zhiqiang-hhhh:
> I don't think this fix is suitable.

Please add some suggestion instead of just giving a conclusion.

adonis0147:
> > I don't think this fix is suitable.
> 
> Please add some suggestion instead of just giving a conclusion.

Hi @zhiqiang-hhhh , I feel sorry for the reply. I validated another fix just now. I think the following modification is more suitable.

```shell
/Users/runner/work/doris/doris/be/src/vec/functions/function_string.h:3024:22: error: call to 'abs' is ambiguous
        frac_value = std::abs(frac_value / multiplier) + 5;
                     ^~~~~~~~
```

Another fix without defining functions in namespace `std`:
```cpp
// do rounding, so we need to reserve 3 digits.
auto multiplier = common::exp10_i128(std::abs(static_cast<int>(scale - 3)));
// do devide first to avoid overflow
// after round frac_value will be positive by design.
frac_value = std::abs(frac_value) / multiplier + 5;
```

We can just make `frac_value` positive due to `multiplier` is positive.

zhiqiang-hhhh:
> ```c++
> frac_value
> ```

`T` could be Int128, `std::abs` seems not support using `__int128` directly unless we are using libstdc++. I am wondering whether the fix you mentioned can work..

zhiqiang-hhhh:
@adonis0147 I made a test, and modified as your suggestion, it works. So this one is closed.

PR #35126, opened 2024-05-21
[bugfix](stdcallonce) replace std callonce with a lock because it is not exception safe

yiguolei:
## Proposed changes

According to cppreference https://en.cppreference.com/w/cpp/thread/call_once. If an exception throws from the call method, the flag is not set and the caller method could catch the exception and handle it.

But I tried with clang16 and gcc11。 And also sometime clang17 will core. The core stack is as follows:

img_v3_02b3_2eb61bac-8dcd-4de6-aad9-51e6d188f5bg

So that doris should not depend on it.

## Further comments

### DIFF
--- a/be/src/util/once.h
+++ b/be/src/util/once.h
@@ -110,7 +110,10 @@ public:
 
     // Return the stored result. The result is only meaningful when `has_called() == true`.
     ReturnType stored_result() const {
-        std::lock_guard l(_flag_lock);
+        if (!has_called()) {
+            // Could not return status if the method not called.
+            throw std::exception();
+        }
         if (_eptr) {
             std::rethrow_exception(_eptr);
         }
@@ -120,7 +123,7 @@ public:
 private:
     std::atomic<bool> _has_called;
     // std::once_flag _once_flag;
-    mutable std::mutex _flag_lock;
+    std::mutex _flag_lock;
     std::exception_ptr _eptr;
     ReturnType _status;
 };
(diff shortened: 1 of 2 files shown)


yiguolei:
run buildall

be/src/util/once.h
```diff
@@ -77,6 +99,7 @@ class DorisCallOnce {
 private:
     std::atomic<bool> _has_called;
     std::once_flag _once_flag;
```
gavinchou, line 101: Is _once_flag still needed?

yiguolei: Useless, I removed it.

yiguolei:
run buildall

be/src/util/once.h
```diff
@@ -52,14 +52,36 @@ class DorisCallOnce {
...
+            return _status;
+        }
+        std::lock_guard l(_flag_lock);
+        // After lock, should check again. Another thread may call
+        // at the same time during first time check has called.
+        if (has_called()) {
+            return _status;
+        }
+        _status = fn();
+        _has_called.store(true, std::memory_order_release);
```
zclllyybb, line 84: should `store` then `fn()`?

zclllyybb (changes requested):
In test should test concurrency thoroughly. maybe just use `std::thread` or `std::async` to start many enough thread

be/src/util/once.h
yiguolei: No, if exception occured during fn, the flag should be false.

yiguolei: No, if stored then other thread will check the has called == true, and will access the result. It is wrong.

yiguolei:
run buildall

Gabriel39 (approved):

mrhhsg (approved):
lgtm

PR #35137, opened 2024-05-21
[fix](routine-load) fix data source properties do not persist in edit log

sollhui:
## Proposed changes

Data source properties do not persist in edit log will affect the alter command and fail to take effect in the following two scenarios:

1. FE leader change result in the loss of values changed by the alter command.
- Progress is {"0":"39"}
-  use `ALTER ROUTINE LOAD FOR lineitem FROM kafka (     "kafka_partitions" = "0",     "kafka_offsets" = "100");` to alter progress
- fe down and leader change.
- kafka offsets is the value before the alter change. which is {"0":"39"}, except is {"0":"99"}

2. Restarting FE may also result in the loss of values changed by the alter command.


HHoflittlefish777:
run buildall

gavinchou (approved):

dataroaring (approved):
LGTM

PR #35189, opened 2024-05-22
[Feat](planner)create table stmt offer default distribution attribute :random distribution and auto bucket

feiniaofeiafei:
Create table stmt do not need to specify distribution, the default distribution is random distribution and auto bucket.


morrySnow (approved):

feiniaofeiafei:
run buildall

yiguolei (approved):

PR #35194, opened 2024-05-22
[improve](jdbc catalog) Remove all property checks during create

zy-kkk:
## Proposed changes

Issue Number: close #xxx

Previously, in order to prevent users from writing wrong parameters when creating a jdbc catalog, we checked all properties. However, we found that when the parent class adds a new property, the subclass cannot sense the parent class's newly added property and can only manually add a check list. In order to avoid this trouble, I will delete this check for now and will check it in a better way later.

## Further comments


zy-kkk:
run buildall

morningman (approved):
LGTM

wuwenchi (approved):

Jibing-Li (approved):

PR #35195, opened 2024-05-22
[fix](pipeline) Fix resource leak if prepare failed

Gabriel39:
## Proposed changes

Now we think a query is done only if all instances are done. But instances account is computed by FE which is an expected values. If query failed in prepare phase and some instances are not ready, this condition will never be met .

## Further comments


Gabriel39:
run buildall

PR #35248, opened 2024-05-23
[refactor](query) Make query finish with context released safely

Gabriel39:
## Proposed changes

We rely on a count number to decide if we can release a query context. However, count is not always reliable. If query is failed in preparation phase, the fragment number computed by FE may not be equal to its actual initialized fragments. So we here make query context a weak ptr which is release when all fragments finish instead of relying a reference count.

## Further comments


Gabriel39:
run buildall

Gabriel39:
run buildall

Gabriel39:
run buildall

yiguolei (approved):

BiteTheDDDDt (approved):

PR #35272, opened 2024-05-23
[Fix](Nereids) fix leading hint should have all tables in one query block (#33517)

LiBinfeng-01:
cherry-pick: #33517


LiBinfeng-01:
run buildall

PR #35314, opened 2024-05-23
[Fix](group commit) Fix group commit block queue mem estimate faule

Yukang-Lian:
## Proposed changes

Issue Number: close #xxx

**Problem:** When `group commit=async_mode` and NULL data is imported into a `variant` type column, it causes incorrect memory statistics for group commit backpressure, leading to a stuck issue.
**Cause:** In group commit mode, blocks are first added to a queue in batches using `add block`, and then blocks are retrieved from the queue using `get block`. To track memory usage during backpressure, we add the block size to the memory statistics during `add block` and subtract the block size from the memory statistics during `get block`. However, for `variant` types, during the `add block` write to WAL, serialization occurs, which can merge types (e.g., merging `int` and `bigint` into `bigint`), thereby changing the block size. This results in a discrepancy between the block size during `get block` and `add block`, causing memory statistics to overflow.
**Solution:** Record the block size at the time of `add block` and use this recorded size during `get block` instead of the actual block size. This ensures consistency in the memory addition and subtraction.

## Further comments

### DIFF
--- a/be/src/runtime/group_commit_mgr.cpp
+++ b/be/src/runtime/group_commit_mgr.cpp
@@ -62,9 +62,27 @@ Status LoadBlockQueue::add_block(RuntimeState* runtime_state,
     RETURN_IF_ERROR(status);
     if (block->rows() > 0) {
         if (!config::group_commit_wait_replay_wal_finish) {
-            _block_queue.push_back(block);
+            _block_queue.emplace_back(block);
             _data_bytes += block->bytes();
+            int before_block_queues_bytes = _all_block_queues_bytes->load();
             _all_block_queues_bytes->fetch_add(block->bytes(), std::memory_order_relaxed);
+            std::stringstream ss;
+            ss << "[";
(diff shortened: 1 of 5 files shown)


Yukang-Lian:
test will be added soon

Yukang-Lian:
run buildall

Yukang-Lian:
run buildall

be/src/runtime/group_commit_mgr.cpp
```diff
@@ -184,8 +220,27 @@ void LoadBlockQueue::_cancel_without_lock(const Status& st) {
     status = st;
     while (!_block_queue.empty()) {
         {
```
gavinchou, line 222: What is this code block indention for?

Yukang-Lian:
run buildall

mymeiyi (approved):

dataroaring (approved):
LGTM

PR #35329, opened 2024-05-24
[fix](shuffle) Do not return error if local recvr is null

Gabriel39:
## Proposed changes

errCode = 2, detailMessage = (172.30.0.248)[INTERNAL_ERROR]_local_recvr is null: 27

If downstream exchange operator finish earlier (e.g. if limited rows has reached), _local_recvr will be released. And then exchange sink operator try to get _local_recvr will be failed.
In fact, if downstream exchange operator finished, exchange sink operator can ignore it.

## Further comments

### DIFF
--- a/be/src/pipeline/exec/exchange_sink_operator.cpp
+++ b/be/src/pipeline/exec/exchange_sink_operator.cpp
@@ -155,12 +155,17 @@ Status ExchangeSinkLocalState::open(RuntimeState* state) {
         size_t dep_id = 0;
         for (auto* channel : channels) {
             if (channel->is_local()) {
-                _local_channels_dependency.push_back(channel->get_local_channel_dependency());
-                DCHECK(_local_channels_dependency[dep_id] != nullptr);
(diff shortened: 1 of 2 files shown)


Gabriel39:
run buildall

be/src/pipeline/exec/exchange_sink_operator.cpp
```diff
@@ -155,12 +155,17 @@ Status ExchangeSinkLocalState::open(RuntimeState* state) {
         size_t dep_id = 0;
         for (auto* channel : channels) {
             if (channel->is_local()) {
-                _local_channels_dependency.push_back(channel->get_local_channel_dependency());
```
yiguolei, line 158: why should we extract this dependency dynamically?
The channel is inited during prepare.
We should extract these dependency during open？
I think we should extract it during prepare stage and then we could avoid such problems.

yiguolei (approved):

BiteTheDDDDt (approved):

PR #35419, opened 2024-05-27
[Branch-2.0](35210)[fix](create-table)The CREATE TABLE IF NOT EXISTS AS SELECT statement should refrain from performing any INSERT operations if the table already exists

CalvinKirs:
#35210


CalvinKirs:
run buildall

PR #35456, opened 2024-05-27
[Fix](JobSchedual) Modify the default value of `async_task_consumer_thread_num`

BePPPower:
## Proposed changes

Issue Number: close #xxx

When `Export` statements are executed concurrently, the background uses `Job schedule` to manage export tasks. Previously, the default value of `async_task_consumer_thread_num` was 5, meaning that regardless of the concurrency setting, a maximum of only 5 threads could execute concurrently.

On the other hand, not only `Export` uses `Job schedule`, but other scheduled tasks might also use `Job schedule`, leading to a shortage of thread resources

Now, we have found that in many scenarios, `Export` needs to be set to a high concurrency value and run concurrently according to that high value. Clearly, `async_task_consumer_thread_num = 5` is no longer sufficient, so we have changed the default value of `async_task_consumer_thread_num` to 64

## Further comments

### DIFF
--- a/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
@@ -1704,7 +1704,7 @@ public class Config extends ConfigBase {
      * if we have a lot of async tasks, we need more threads to consume them. Sure, it's depends on the cpu cores.
      */
     @ConfField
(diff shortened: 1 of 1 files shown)


CalvinKirs (approved):

BePPPower:
run buildall

morningman (approved):
Add description to explain why changing this config

PR #35474, opened 2024-05-27
[fix](statistics) correct update rows when doing multi-table load

freemandealer:
rows of only one table is updated correctly, need to merge all table commit infos.

## Proposed changes

Issue Number: close #xxx

## Further comments

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/transaction/PublishVersionDaemon.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/transaction/PublishVersionDaemon.java
@@ -302,7 +302,14 @@ public class PublishVersionDaemon extends MasterDaemon {
     private List<TPartitionVersionInfo> generatePartitionVersionInfos(TableCommitInfo tableCommitInfo,
             TransactionState transactionState, Map<Long, Set<Long>> beIdToBaseTabletIds) {
         try {
-            beIdToBaseTabletIds.putAll(getBaseTabletIdsForEachBe(transactionState, tableCommitInfo));
+            Map<Long, Set<Long>> map = getBaseTabletIdsForEachBe(transactionState, tableCommitInfo);
+            map.forEach((beId, newSet) -> {
+                beIdToBaseTabletIds.computeIfPresent(beId, (id, orgSet) -> {
+                    orgSet.addAll(newSet);
+                    return orgSet;
+                });
+                beIdToBaseTabletIds.putIfAbsent(beId, newSet);
+            });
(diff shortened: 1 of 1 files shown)


freemandealer:
run buildall

fe/fe-core/src/main/java/org/apache/doris/transaction/PublishVersionDaemon.java
```diff
@@ -302,7 +302,14 @@ private List<TPartitionVersionInfo> generatePartitionVersionInfos(Collection<Tab
...
             TransactionState transactionState, Map<Long, Set<Long>> beIdToBaseTabletIds) {
         try {
-            beIdToBaseTabletIds.putAll(getBaseTabletIdsForEachBe(transactionState, tableCommitInfo));
+            Map<Long, Set<Long>> map = getBaseTabletIdsForEachBe(transactionState, tableCommitInfo);
+            map.forEach((beId, newSet) -> {
+                beIdToBaseTabletIds.computeIfPresent(beId, (id, orgSet) -> {
+                    orgSet.addAll(newSet);
+                    return orgSet;
+                });
+                beIdToBaseTabletIds.putIfAbsent(beId, newSet);
```
gavinchou, line 311: Is line 311 necessary? 
The lib says that for computeIfPresent
> If the specified key is not already associated with a value (or is mapped to null), attempts to compute its value using the given mapping function and enters it into this map unless null.
If the function returns null no mapping is recorded. If the function itself throws an (unchecked) exception, the exception is rethrown, and no mapping is recorded. The most common usage is to construct a new object serving as an initial mapped value or memoized result, as in:
 map.computeIfAbsent(key, k -> new Value(f(k)));

Jibing-Li:
```
    default V computeIfPresent(K key,
            BiFunction<? super K, ? super V, ? extends V> remappingFunction) {
        Objects.requireNonNull(remappingFunction);
        V oldValue;
        if ((oldValue = get(key)) != null) {
            V newValue = remappingFunction.apply(key, oldValue);
            if (newValue != null) {
                put(key, newValue);
                return newValue;
            } else {
                remove(key);
                return null;
            }
        } else {
            return null;
        }
    }
```
Look at the outer else branch, it seems like we need to keep line 311. Because it doesn't put new value to the map when the key doesn't exist.

Jibing-Li (approved):
LGTM

gavinchou (approved):

PR #35497, opened 2024-05-28
[fix](Nereids) fix Illegal aggregate node: group by and output is empty

924060929:
fix Illegal aggregate node: group by and output is empty, introduced by #33091

PR #35498, opened 2024-05-28
[fix](Nereids) fix Illegal aggregate node: group by and output is empty

924060929:
cherry pick from #35497
and fix macos compile failed by std::abs


924060929:
run buildall

PR #35472, opened 2024-05-27
[fix](compaction) fix repeatedly picking tablets with disable auto compaction

luwei16:

PR #35505, opened 2024-05-28
[fix](compaction) fix repeatedly picking tablets with disable auto compaction (#35472)

luwei16:
pick master #35472


luwei16:
run buildall

PR #35535, opened 2024-05-28
[opt](audit) add timeout for audit log load and modify the label format

morningman:
## Proposed changes

1. Add a new global variable: `audit_plugin_load_timeout`, default is 600 sec
   Avoid using default stream load, which is 4 hours, too long for audit log load.

2. Modify the label of audit log load
   Add millisecond to avoid same label from different FE.


deardeng (approved):
LGTM

gavinchou (approved):

dataroaring (approved):
LGTM

PR #35537, opened 2024-05-28
[fix](snappy) avoid potential buffer overflow

dataroaring:
If skip more than once when available is zero, then a buffer overflow occurs.

## Proposed changes

photo-size-5-6244711526321733357-y

Issue Number: close #xxx

## Further comments

### DIFF
--- a/be/src/util/block_compression.cpp
+++ b/be/src/util/block_compression.cpp
@@ -669,7 +669,7 @@ public:
     // REQUIRES: Available() >= n
     void Skip(size_t n) override {
         _available -= n;
-        while(n > 0) {
+        while (n > 0) {
             auto left = _slices[_cur_slice].size - _slice_off;
             if (left > n) {
                 // n can be digest in current slice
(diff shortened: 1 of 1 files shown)


dataroaring:
run buildall

dataroaring:
run buildall

gavinchou (approved):

be/src/util/block_compression.cpp
```diff
@@ -669,7 +669,7 @@ class SnappySlicesSource : public snappy::Source {
     // REQUIRES: Available() >= n
     void Skip(size_t n) override {
         _available -= n;
```
cambyzju, line 671: 1. this function has requirement for the caller:  `// REQUIRES: Available() >= n`, if the caller can not make sure this, maybe the caller have some bug;
2. After the change in this pr, `_available` may be corrupted;
3. If we really want to make `Skip` safe, I think only add a check at the begin of the function is better:
 ```if (unlikely(_available < n)) { return; }```

dataroaring: when available = 0 and n = 0, the requirement is met.

liaoxin01 (approved):
LGTM

PR #35539, opened 2024-05-28
[fix](compaction) fix inaccurate compaction cnt

luwei16:
## bug
Before this bug is fixed, when the rpc of commit compaction job and the rpc of sync tablet meta are parallel and out of order, the compaction cnt may be inaccurate.

for example

1 cumu_compaction_cnt = 10
2 meta service, commit compaction job rpc, cumu_compaction_cnt++ (thread 1)
3 meta service, get tablet meta rpc, get cumu_compaction_cnt = 11      (thread 2)
4 BE，get tablet meta rpc return，update tablet cumu_compaction_cnt = 11 (thread 2)
5 BE，commit compaction job rpc return，update cumu_compaction_cnt = 12 (thread 1)


luwei16:
run buildall

gavinchou (approved):

dataroaring (approved):
LGTM

PR #35445, opened 2024-05-27
[fix](group commit) should set wal id in runtime_state when building pipeline task

hust-hhb:
after commit https://github.com/apache/doris/commit/85026759d6ed8d89240beb1b546b08c67db098d7, stream load using pipeline engine to load data, wal id should be setted in runtime_state when building pipeline task, like plan_fragment_executor does. Wal id in runtime_state is used to find wal file.

## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #35552, opened 2024-05-28
[fix](group commit) should set wal id in runtime_state when building pipeline task

hust-hhb:
pick from master #35445

## Proposed changes

Issue Number: close #xxx


hust-hhb:
run buildall

PR #35461, opened 2024-05-27
[fix](binlog) Fix add partition record sql

w41ter:
## Proposed changes

Issue Number: close #xxx

1. support adding a temporary partition
2. remove extra parentheses in the list partition value set
3. support unpartitioned partition item

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #35564, opened 2024-05-29
[fix](binlog) Fix add partition record sql

w41ter:
1. support adding a temporary partition
2. remove extra parentheses in the list partition value set
3. support unpartitioned partition item

Cherry-pick #35461


w41ter:
run buildall

PR #33817, opened 2024-04-18
[feature](fe) Advance the next id before transfering to master

w41ter:
## Proposed changes

Issue Number: close #xxx

Advance the id generator, ensuring it doesn't roll back.

If we need to support time travel, the next id cannot be rolled back to avoid
errors in the corresponding relationship of the metadata recorded in BE/MS.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #35568, opened 2024-05-29
[feature](fe) Advance the next id before transfering to master

w41ter:
If the FE metadata needs to be rolled back (such as time travel) and the ID generator is rolled back, the corresponding relationship of the metadata in BE/MS may be abnormal. This patch attempts to use the latest timestamp to advance the next id of ID generator before becoming master each time. This can ensure that in most scenarios, the rollback of metadata will not cause the ID generator to be rolled back.

Cherry-pick #33817


w41ter:
run buildall

Issue #34228, opened 2024-04-28
[Enhancement] Add some default value function for doris


Yukang-Lian:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

### Description

1. localtime, localtimestamp
2. now
3. bitmap_empty
4. hll_empty
5. e
6. pi
7. uuid_numeric
8. random

### Solution

Reference #34211 

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Z-SWEI:
I want to try

Yukang-Lian:
> I want to try

Which one do you want to try?

Z-SWEI:
> > I want to try
> 
> Which one do you want to try?

2. now
3. bitmap_empty
7. uuid_numeric
8. random

Dragonliu2018:
I want to try 1 and 6.

my-vegetable-has-exploded:
I'd like to try hll_empty. Thanks.

PR #35574, opened 2024-05-29
[Feature](column) do not change column nullable on ColumnDef

BiteTheDDDDt:
## Proposed changes
do not change column nullable on ColumnDef

1. bitmap/hll/quantile_state will directly change nullable to false
    1. Currently, complex types do not support nullable in the storage layer. If it is set to null, it should be coredump.
    2. It is meaningless to set complex types to null. For example, bitmap can use bitmap_empty to represent null to avoid     nullable overhead.
    3. Users reported that if it is a not null bitmap. When inserting into, it is necessary to explicitly declare bitmap_empty, which is troublesome. However, the work of supporting the default value of bitmap_empty for complex columns has been claimed by the community. https://github.com/apache/doris/issues/34228
    4. Consider changing it here so that if null is explicitly specified when creating a table, an error will be reported. If it is not explicitly specified (the default will be null), it is not null.
2. The aggregate type REPLACE_IF_NOT_NULL will directly change nullable to true
    1. If it is a not null column, using REPLACE_IF_NOT_NULL is meaningless, which is equivalent to REPLACE.
    2. Consider changing it here so that if not nullable is encountered, an error will be reported, and a prompt to use REPLACE will be given.


BiteTheDDDDt:
run buildall

BiteTheDDDDt:
run buildall

BiteTheDDDDt:
run buildall

BiteTheDDDDt:
run buildall

BiteTheDDDDt:
run buildall

BiteTheDDDDt:
run buildall

morrySnow (approved):

Mryange (approved):

PR #35203, opened 2024-05-22
[fix](fe) Fix drop frontend cause `bdbje` and `fe memory` inconsistent state

swjtu-zhanglei:
* dropFrontend use `ReplicationGroupAdmin` to remove bdbje node, in some corner case
  `ReplicationGroupAdmin.removeMember` success but `editLog.logRemoveFrontend` fail
   will cause inconsistent node info between `fe memory` and `bdbje`

PR #35612, opened 2024-05-29
[fix](fe) fix start fe failed when recover fe in different host

924060929:
## Proposed changes

An exception introduced by #35203 cause the frontend start failed

In this cases:
1. all frontends are down and recover frontend in different host by the original metadata
2. or the only frontend's ip changed, and start with `-r` option to recover from original metadata


swjtu-zhanglei (approved):

morrySnow (approved):

924060929:
run buildall

PR #35615, opened 2024-05-29
[fix](inverted index) Remove float/double support in index writer

qidaye:
## Proposed changes

`float` and `double` is not allowed to build inverted index.
We remove them in `inverted_index_writer` to keep consistent with FE. And to avoid unnecessary exception.


qidaye:
run buildall

airborne12 (approved):
LGTM

eldenmoon (approved):
LGTM

PR #35623, opened 2024-05-30
[fix](mtmv) Fix get mv statistics plan wrong and optimize code usage

seawinde:
## Proposed changes

1. get mv statistics, the relation id should be getted from scan plan, fix it.
2. optimize the `isGroupEquals` method, add `MaterializationContext` param which maybe used to control the decide group by equals logic.
3. remove `set enable_nereids_timeout = false;` setting in mv rewrite regression test.


seawinde:
run buildall

morrySnow (approved):

starocean999 (approved):

PR #35196, opened 2024-05-22
[improvement](jdbcScan) Change the mysql function that does not support pushdown in JdbcScan to Config

zy-kkk:
## Proposed changes

Issue Number: close #xxx

We noticed that mysql does not have a negative function, so the function pushdown rule prohibits it

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #35631, opened 2024-05-30
[branch-2.1][improvement](JdbcScan) Change the mysql function that does not support pushdown in JdbcScan to Config

zy-kkk:
## Proposed changes

Issue Number: close #xxx

pk #35196

### DIFF
--- a/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
@@ -151,6 +151,10 @@ public class Config extends ConfigBase {
                     + "The default is * to allow all, if set to empty, all are not allowed"})
     public static String jdbc_driver_secure_path = "*";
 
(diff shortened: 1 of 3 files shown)


fe/fe-common/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -151,6 +151,10 @@ public class Config extends ConfigBase {
                     + "The default is * to allow all, if set to empty, all are not allowed"})
     public static String jdbc_driver_secure_path = "*";
 
+    @ConfField(description = {"MySQL Jdbc Catalog mysql 不支持下推的函数",
+            "MySQL Jdbc Catalog mysql does not support pushdown functions"}, options = {"date_trunc", "money_format"})
```
cambyzju, line 155: why we need options here?

morningman (approved):
LGTM

PR #35632, opened 2024-05-30
[refactor](cancel) disable cancel instance for pipeline

Gabriel39:
## Proposed changes

For pipeline engine, query will be cancelled only if all pipeline tasks should be stopped. So we should disable `cancel_instance` which is to stop fragment instances in non-pipeline engine.

### DIFF
--- a/be/src/runtime/fragment_mgr.cpp
+++ b/be/src/runtime/fragment_mgr.cpp
@@ -997,6 +997,9 @@ void FragmentMgr::cancel_instance(const TUniqueId instance_id, const Status reas
     std::shared_ptr<PlanFragmentExecutor> non_pipeline_ctx;
     {
         std::lock_guard<std::mutex> state_lock(_lock);
+        DCHECK(!_pipeline_map.contains(instance_id))
+                << " Pipeline tasks should be canceled by query instead of instance! Query ID: "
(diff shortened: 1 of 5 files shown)


Gabriel39:
run buildall

Gabriel39:
run buildall

be/src/runtime/fragment_mgr.cpp
```diff
@@ -997,25 +997,16 @@ void FragmentMgr::cancel_instance(const TUniqueId instance_id, const Status reas
...
-        if (is_pipeline_instance) {
-            auto itr = _pipeline_map.find(instance_id);
-            if (itr != _pipeline_map.end()) {
-                pipeline_ctx = itr->second;
-            } else {
-                LOG(WARNING) << "Could not find the pipeline instance id:" << print_id(instance_id)
-                             << " to cancel";
-                return;
-            }
+        CHECK(!_pipeline_map.contains(instance_id))
```
yiguolei, line 1000: DO NOT use check to avoid core online.
Using DCHECK and print warning message please.

Gabriel39:
run buildall

yiguolei (approved):

BiteTheDDDDt (approved):

PR #35649, opened 2024-05-30
[fix](recyclebin) Fix replay erase table NPE

w41ter:
NPE may occur in some special scenarios so we will take some precautions here.


w41ter:
run buildall

xiaokang (approved):
LGTM

PR #35653, opened 2024-05-30
[fix](fe) Add check editlog size mechanism for backupJob

swjtu-zhanglei:
* When creating a backupJob with huge of tables in a database, it can cause backupJob editlog size over 2GB and bdbje will throw exception because of ByteBuffer overflow

## Proposed changes

Issue Number: close #xxx

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/journal/bdbje/BDBJEJournal.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/journal/bdbje/BDBJEJournal.java
@@ -707,4 +707,27 @@ public class BDBJEJournal implements Journal { // CHECKSTYLE IGNORE THIS LINE: B
         }
         return bdbEnvironment.getNotReadyReason();
     }
+
+    @Override
+    public boolean exceedMaxJournalSize(short op, Writable writable) throws IOException {
+        JournalEntity entity = new JournalEntity();
+        entity.setOpCode(op);
+        entity.setData(writable);
+
+        DataOutputBuffer buffer = new DataOutputBuffer(OUTPUT_BUFFER_INIT_SIZE);
+        entity.write(buffer);
+
+        DatabaseEntry theData = new DatabaseEntry(buffer.getData());
+
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("opCode = {}, journal size = {}", op, theData.getSize());
+        }
+
+        // 1GB
+        if (theData.getSize() > (1 << 30)) {
+            return true;
+        }
+
+        return false;
+    }
 }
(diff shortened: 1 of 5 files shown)


SWJTU-ZhangLei:
run buildall

SWJTU-ZhangLei:
run buildall

fe/fe-core/src/main/java/org/apache/doris/journal/bdbje/BDBJEJournal.java
```diff
@@ -707,4 +707,27 @@ public String getNotReadyReason() {
         }
         return bdbEnvironment.getNotReadyReason();
     }
+
+    @Override
+    public boolean exceedMaxJournalSize(short op, Writable writable) throws IOException {
```
gavinchou, line 712: Is param `op` mandatory for this check?

fe/fe-core/src/main/java/org/apache/doris/persist/EditLog.java
```diff
@@ -2102,4 +2102,17 @@ public String getNotReadyReason() {
         }
         return "";
     }
+
+    public boolean exceedMaxJournalSize(BackupJob job) {
```
gavinchou, line 2106: It seems that we can change it to the following form
```suggestion
    public boolean exceedMaxJournalSize(Writable w) {
```

fe/fe-core/src/main/java/org/apache/doris/persist/EditLog.java
```diff
@@ -2102,4 +2102,17 @@ public String getNotReadyReason() {
...
+        try {
+            return exceedMaxJournalSize(OperationType.OP_BACKUP_JOB, job);
+        } catch (Exception e) {
+            LOG.warn("exceedMaxJournalSize exception:", e);
+        }
+        return true;
+    }
+
+    private boolean exceedMaxJournalSize(short op, Writable writable) throws IOException {
```
gavinchou, line 2115: This function is redundant if we do not care about `op`

SWJTU-ZhangLei: > This function is redundant if we do not care about `op`

opCode is important, we will print opCode

dataroaring:
please add a regression test in p2.

dataroaring (approved):
LGTM

gavinchou (approved):

dataroaring:
run buildall

SWJTU-ZhangLei:
run buildall

dataroaring:
run buildall

dataroaring (approved):
LGTM

PR #35662, opened 2024-05-30
[feature](workloadgroup)Add workload condition query used memory

wangbo:
## Proposed changes
Add workload condition query used memory, we can kill queries based on memory usage.

1  create a policy which can kill query used memory exceeds 100M.
```
create workload policy memory_used_policy conditions(query_be_memory_bytes > 104857600) actions(cancel_query);
```

2 submit a query.
```
mysql [hits]>insert into hits2 select * from hits;
ERROR 1105 (HY000): errCode = 2, detailMessage = (10.16.10.8)[INTERNAL_ERROR]query 81c39e2b3ecf461c-bc78ad6d9b6173d2 cancelled by workload policy memory_used_policy, id:29028
```

### DIFF
--- a/be/src/runtime/runtime_query_statistics_mgr.cpp
+++ b/be/src/runtime/runtime_query_statistics_mgr.cpp
@@ -621,6 +621,8 @@ void RuntimeQueryStatiticsMgr::get_metric_map(
     metric_map.emplace(WorkloadMetricType::QUERY_TIME, std::to_string(query_time_ms));
     metric_map.emplace(WorkloadMetricType::SCAN_ROWS, std::to_string(ret_qs.get_scan_rows()));
     metric_map.emplace(WorkloadMetricType::SCAN_BYTES, std::to_string(ret_qs.get_scan_bytes()));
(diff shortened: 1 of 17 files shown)


wangbo:
run buildall

yiguolei:
the metric name should be consistent with other metrics in be, for example:
be_scan_rows，The number of rows scanned by an SQL within a single BE process, and if there are multiple concurrency, it is the cumulative value of multiple concurrency.
be_scan_bytes，The number of bytes scanned by an SQL within a single BE process, and if there are multiple concurrency, it is the cumulative value of multiple concurrency.
query_time，The running time of an SQL on a single BE process, measured in milliseconds.

So that I think the name should be be_memory_used_bytes,

yiguolei (dismissed):

yiguolei:
run buildall

wangbo:
run buildall

mrhhsg (approved):
LGTM

xinyiZzz (approved):
LGTM

PR #35665, opened 2024-05-30
[bugfix](cache) should calcuate memusage when create segment

yiguolei:
## Proposed changes

Issue Number: close #xxx


yiguolei:
run buildall

PR #35674, opened 2024-05-30
 [improvement](mtmv) Improve the mv rewrite performance by optimize code usage

seawinde:
## Proposed changes

Improve the performance from two points, one is optimize decide model method and  another is to reuse the mv struc info:
1. Instead of use` java.util.List#containsAll `by `java.util.Set#containsAll` in method `AbstractMaterializedViewRule#decideMatchMode`
2. Reuse the mv struct info in different query, because mv struct info is immobile. 
Notes: tableBitSet in struct info is relevant to the `statementContext` in `cascadesContext`, if reuse the mv struct info for different query, we should re generate table bitset and construct new struct info with method `StructInfo#withTableBitSet`


seawinde:
run buildall

morrySnow (approved):

seawinde:
run buildall

keanji-x (approved):

PR #35679, opened 2024-05-30
[opt](log) refine the FE logger

morningman:
## Background

Previously, FE logs were written to files. The main FE logs include fe.log, fe.warn.log, fe.audit.log, fe.out, and fe.gc.log.
In a K8s deployment environment, logs usually need to be output to standard output, and then other components process the log stream.

## Solution

This PR made the following changes:

1. Modified the log4j configuration template

	- When started with `--daemon`, logs are still written to various files, and the format remains unchanged.
	- When started with `--console`, all logs are output to standard output and marked with different prefixes:
	
		- `StdoutLogger`: logs for standard output
		- `StderrLogger`: logs for standard error output
		- `RuntimeLogger`: logs for fe.log or fe.warn.log
		- `AuditLogger:` logs for fe.audit.log
		- No prefix: logs for fe.gc.log
		
		Examples are as follows:
		
		```
		RuntimeLogger 2024-06-03 14:54:51,229 INFO (binlog-gcer|62) [BinlogManager.gc():359] begin gc binlog
		```

2. Added a new FE config: `enable_file_logger`

	Defaults to true. Indicates that logs will be recorded to files regardless of the startup method. For example, if it is started with `--console`, the log will be output to both the file and the standard output. If it is `false`, the log will not be recorded in the file regardless of the startup method.

3. Optimized the log format of standard output

	The byte streams of stdout and stderr are captured. The logs previously outputted using `System.out` will be captured in fe.log for unified management.


morningman:
run buildall

morningman:
run buildall

morrySnow (approved):

intelligentfu (approved):

fe/fe-core/src/main/java/org/apache/doris/journal/bdbje/BDBEnvironment.java
```diff
@@ -105,8 +105,7 @@ public void setup(File envHome, String selfNodeName, String selfNodeHostPort,
             DbResetRepGroup resetUtility = new DbResetRepGroup(
                     envHome, PALO_JOURNAL_GROUP, selfNodeName, selfNodeHostPort);
             resetUtility.reset();
-            LOG.warn("WARNING: metadata recovery mode, group has been reset.");
-            System.out.println("WARNING: metadata recovery mode, group has been reset.");
+            LOG.warn("metadata recovery mode, group has been reset.");
```
w41ter, line 108: Printing this type of log to stdout is critical for troubleshooting.

fe/fe-core/src/main/java/org/apache/doris/persist/EditLog.java
```diff
@@ -446,8 +446,7 @@ public static void loadJournal(Env env, Long logId, JournalEntity journal) {
                     Frontend fe = (Frontend) journal.getData();
                     env.replayDropFrontend(fe);
                     if (fe.getNodeName().equals(Env.getCurrentEnv().getNodeName())) {
-                        System.out.println("current fe " + fe + " is removed. will exit");
-                        LOG.info("current fe " + fe + " is removed. will exit");
+                        LOG.warn("current fe {} is removed. will exit", fe);
```
w41ter, line 449: Same problem

PR #35681, opened 2024-05-30
[enhancement](from_unixtime) function from_unixtime support using decimal as argument

zhiqiang-hhhh:
Keep same with mysql, from_unixtime can use decimal as its first argument.
When format argument is not supplied, the scale of datetime will be same with input decimal.
For example, `select from_unixtime(123.123)` will get `1970-01-01 08:02:03.123`. 
The largest scale of datetime is 6, so `select from_unixtime(123.1234567);` will be `1970-01-01 08:02:03.123457`, rounding up, like mysql.


zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

regression-test/suites/datatype_p0/date/test_from_unixtime.groovy
zclllyybb, line 1: should also add some case 
1. over the bound so get NULL
2. big exactly get not NULL

be/src/vec/functions/function_from_unixtime.cpp
```diff
@@ -0,0 +1,301 @@
...
+            throw doris::Exception(ErrorCode::INVALID_ARGUMENT,
+                                   "from_unixtime() requires at least 1 argument, 0 provided.");
+        }
+
+        if constexpr (argument_num == ArgumentNumber::Two) {
+            if (arguments.size() != 2) {
+                throw doris::Exception(ErrorCode::INTERNAL_ERROR,
+                                       "from_unixtime() requires 2 arguments, {} provided.",
+                                       arguments.size());
+            }
```
zclllyybb, line 188: I think we shouldn't do checks like these. It's part of BE's function match. this is redundant

be/src/vec/functions/function_from_unixtime.cpp
```diff
@@ -0,0 +1,301 @@
...
+        return make_nullable(std::make_shared<DataTypeString>());
+    }
+
+    Status execute_impl(FunctionContext* context, Block& block, const ColumnNumbers& arguments,
+                        size_t result, size_t input_rows_count) const override {
+        const ColumnPtr source_col = block.get_by_position(arguments[0]).column;
+        const DataTypePtr data_type = block.get_by_position(arguments[0]).type;
+        const cctz::time_zone& time_zone = context->state()->timezone_obj();
+
+        const auto* nullable_column = check_and_get_column<ColumnNullable>(source_col.get());
```
zclllyybb, line 219: if so, maybe we should return override `default_impl_for_null` to false.
or you can just ignore the business of null input. just return `ColumnNullable` which nullmap gets true from the function logic, like out-of-bound or something. then at last our default impl framework will combine the nullmap with the input nullmap.

be/src/vec/runtime/vdatetime_value.cpp
```diff
@@ -3367,16 +3370,28 @@ void DateV2Value<T>::from_unixtime(int64_t timestamp, int32_t nano_seconds,
...
+    if (nano_seconds % 10 >= 5) {
+        nano_seconds += 10;
+    }
+
+    if (nano_seconds >= common::exp10_i32(scale + 1)) {
+        timestamp += 1;
+        nano_seconds = 0;
+    }
+
+    int32_t millisecond = (nano_seconds /= 10) * common::exp10_i32(6 - scale);
```
zclllyybb, line 3390: 6 width precision is microsecond, not millisecond

zclllyybb (commented):
good for the generic logic. still need a little fix

be/src/vec/functions/function_from_unixtime.cpp
zhiqiang-hhhh: Actually in clickhouse, function signature check is almost alway added.

be/src/vec/functions/function_from_unixtime.cpp
zhiqiang-hhhh: makes sence

regression-test/suites/datatype_p0/date/test_from_unixtime.groovy
zhiqiang-hhhh: fixed by

`insert into from_unixtime_table values (5, "32536771199.999999"), (6, "32536771199.9999995"), (7, "32536771200.0");`

be/src/vec/runtime/vdatetime_value.cpp
zhiqiang-hhhh: fixed

zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

PR #35684, opened 2024-05-31
[fix](regression)Remove useless case which may cause preHeat npe. (#35582)

Jibing-Li:
backport https://github.com/apache/doris/pull/35582

Remove useless case which may cause preHeat npe.


Jibing-Li:
run buildall

Jibing-Li:
run external

PR #35690, opened 2024-05-31
[fix](s3) move s3 providers to fe-common to be accessible for jni reader

AshinGau:
## Proposed changes

`PropertyConverter.setS3FsAccess` has add customized s3 providers:
```
public static final List<String> AWS_CREDENTIALS_PROVIDERS = Arrays.asList(
            DataLakeAWSCredentialsProvider.class.getName(),
            TemporaryAWSCredentialsProvider.class.getName(),
            SimpleAWSCredentialsProvider.class.getName(),
            EnvironmentVariableCredentialsProvider.class.getName(),
            IAMInstanceCredentialsProvider.class.getName());
```
And these providers are set as configuration value of `fs.s3a.aws.credentials.provider`, which will be used as configuration to build s3 reader in JNI readers. However, `DataLakeAWSCredentialsProvider` is in `fe-core`, that is not dependent by JNI readers, so we have to move s3 providers to `fe-common'.


AshinGau:
run buildall

morningman:
run buildall

morningman (approved):
LGTM

wuwenchi (approved):

PR #35722, opened 2024-05-31
[Bugfix] Fix nullptr cache object because init file cache concurrently

Lchangliang:
The file caches are inited in be/src/runtime/exec_env_init.cpp:init_file_cache_factory concurrently. The field `_path_to_cache` and `_caches` is not protected by lock. So BE may coredump because the cache object is nullptr.

### DIFF
--- a/be/src/io/cache/block_file_cache_factory.cpp
+++ b/be/src/io/cache/block_file_cache_factory.cpp
@@ -92,16 +92,20 @@ Status FileCacheFactory::create_file_cache(const std::string& cache_base_path,
     }
     auto cache = std::make_unique<BlockFileCache>(cache_base_path, file_cache_settings);
     RETURN_IF_ERROR(cache->initialize());
-    _path_to_cache[cache_base_path] = cache.get();
-    _caches.push_back(std::move(cache));
+    {
+        std::lock_guard lock(_mtx);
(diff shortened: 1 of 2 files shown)


Lchangliang:
run buildall

gavinchou (approved):

be/src/io/cache/block_file_cache_factory.cpp
```diff
@@ -92,16 +92,20 @@ Status FileCacheFactory::create_file_cache(const std::string& cache_base_path,
     }
     auto cache = std::make_unique<BlockFileCache>(cache_base_path, file_cache_settings);
     RETURN_IF_ERROR(cache->initialize());
-    _path_to_cache[cache_base_path] = cache.get();
-    _caches.push_back(std::move(cache));
+    {
```
cambyzju, line 95: At the beginnig of this function, the filesystem operations also have concurrency problem? such as:
```
fs->exists
fs->create_directory
fs->delete_directory
fs->create_directory
```

Lchangliang: the dir path is not same.

Lchangliang:
run buildall

dataroaring (approved):
LGTM

PR #35724, opened 2024-05-31
[improvement](build index)Make build index and clone mutually exclusive and add timeout for index change job

qidaye:
## Proposed changes

Currently the index change job and clone task can be executed at the same time. If the clone task gets stuck at this point, it will cause the index change job to get stuck as well and keep retrying.
To solve this problem, we can refer to alter job and make index change job exclusive with clone task, and introduce the timeout to prevent infinite retries of build index.

Add the following checks and status in FE. 
1. Check if table is stable (build index is not allowed when clone is in progress)
1.1. Tablet is HEALTHY.
1.2. Whether the tablet is included in the Tablet scheduler, if so, it means the current tablet is doing clone.
2. When creating the index change job, set the timeout at the same time.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/alter/IndexChangeJob.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/alter/IndexChangeJob.java
@@ -37,6 +37,7 @@ import org.apache.doris.common.io.Writable;
 import org.apache.doris.common.util.TimeUtils;
 import org.apache.doris.persist.gson.GsonUtils;
 import org.apache.doris.task.AgentBatchTask;
+import org.apache.doris.task.AgentTask;
 import org.apache.doris.task.AgentTaskExecutor;
 import org.apache.doris.task.AgentTaskQueue;
 import org.apache.doris.task.AlterInvertedIndexTask;
@@ -46,6 +47,7 @@ import org.apache.doris.thrift.TTaskType;
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.google.gson.annotations.SerializedName;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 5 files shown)


qidaye:
run buildall

qidaye:
run cloud_p1

qidaye:
run buildall

qidaye:
run feut

fe/fe-core/src/main/java/org/apache/doris/alter/IndexChangeJob.java
```diff
@@ -307,10 +353,36 @@ protected void runWaitingTxnJob() throws AlterCancelException {
...
+            for (AgentTask task : tasks) {
+                if (task.getFailedTimes() > 3) {
+                    LOG.warn("alter inverted index task failed: " + task.getErrorMsg());
+                    List<Long> failedBackends = failedTabletBackends.computeIfAbsent(task.getTabletId(),
+                            k -> Lists.newArrayList());
+                    failedBackends.add(task.getBackendId());
+                    int expectSucceedTaskNum = tbl.getPartitionInfo()
+                            .getReplicaAllocation(task.getPartitionId()).getTotalReplicaNum();
+                    int failedTaskCount = failedBackends.size();
+                    if (expectSucceedTaskNum - failedTaskCount < expectSucceedTaskNum / 2 + 1) {
```
xiaokang, line 380: expectSucceedTaskNum is based on partition. failedTaskCount is based on task. They are not on the same dimension.

xiaokang (commented):
pls add testcase

fe/fe-core/src/main/java/org/apache/doris/alter/IndexChangeJob.java
qidaye: failedTaskCount is the same tablet failed count. expectSucceedTaskNum is the tablet's replica number. 
This check is checking the tasks failed on same tablet.

qidaye:
run buildall

qidaye:
run buildall

xiaokang (approved):
LGTM

eldenmoon (approved):
LGTM

PR #35734, opened 2024-05-31
[Feature](IP) support ipv4/ipv6 with inverted index and conjuncts for query

amorynan:
## Proposed changes
support data type ipv4/ipv6 with inverted index 
and then we can query like "> or < or >= or <=  or in/not in " this conjuncts expr for ip with inverted index speeding up
Issue Number: close #xxx


morrySnow (dismissed):

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -573,7 +573,7 @@ Status ColumnReader::_load_inverted_index_index(
                         "create StringTypeInvertedIndexReader error: {}", e.what());
             }
         }
-    } else if (is_numeric_type(type)) {
+    } else if (is_numeric_type(type) || is_ip_type(type)) {
```
xiaokang, line 576: is_numeric_type() should be the same as field_is_numeric_type() and is_ip_type() can be removed here.

be/src/olap/utils.h
```diff
@@ -260,6 +260,10 @@ constexpr bool is_numeric_type(const FieldType& field_type) {
            field_type == FieldType::OLAP_FIELD_TYPE_BOOL;
 }
 
+constexpr bool is_ip_type(const FieldType& field_type) {
```
xiaokang, line 263: not necessary

be/src/olap/rowset/segment_v2/inverted_index_writer.cpp
```diff
@@ -695,6 +695,8 @@ Status InvertedIndexColumnWriter::create(const Field* field,
         M(FieldType::OLAP_FIELD_TYPE_DECIMAL128I)
         M(FieldType::OLAP_FIELD_TYPE_DECIMAL256)
         M(FieldType::OLAP_FIELD_TYPE_BOOL)
+        M(FieldType::OLAP_FIELD_TYPE_IPV4)
+        M(FieldType::OLAP_FIELD_TYPE_IPV6)
```
xiaokang, line 699: Where is the code for transform encoding from little endian to big endian?

airborne12: in key_coder's full_encoding function.

amorynan: writer & reader part both has key_coder's full_encoding function before add_doc and seek in bkd

be/src/olap/rowset/segment_v2/column_reader.cpp
amorynan: ok I will make the same

amorynan:
run buildall

be/src/olap/rowset/segment_v2/inverted_index_writer.cpp
xiaokang: ok

xiaokang (approved):
LGTM

airborne12 (approved):
LGTM

PR #30366, opened 2024-01-25
[fix](partial update) mishandling of exceptions in the publish phase may result in data loss

zhannngchen:
## Proposed changes

Issue Number: close #xxx

We found in our stress test environment that when performing concurrent partial column updates, data was lost or inconsistent due to not to handle write exceptions of the newly generated segment well in publish phase:

1. when two concurrent partial column update job try to update an identical key at the same time, additional conflict handling needs to be done in the publish phase
2. and the load task with a higher version needs to generate the new rows and write them to the newly generated segmnet, while marking the corresponding key in the original segment for deletion
3. this deletion marker is updated in txn_info->delete_bitmap in real time, when writing the new segment failed, these deletion markers for the original segment are not rolled back
4. and the next time the publish task retries, it will not process these keys because they have already been marked for deletion.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #35739, opened 2024-05-31
[fix](partial update) duplicate key occur when BE restart after conflict concurrent partial update 

zhannngchen:
## Proposed changes

Issue Number: close #xxx

1. In #30366 , in order to avoid that some incomplete delete bitmap left in `txn_info->delete_bitmap` when publish failed, we make a copy of `txn_info->delete_bitmap` before we start to compute the delete bitmap
2. this copy is not updated back to `txn_info->delete_bitmap` after `rowset->rowset_meta()->merge_rowset_meta()` is successful
3. `txnManager::publish_txn()` saves the contents of `txn_info->delete_bitmap` to RocksDB after the call to `update_delete_bitmap()`, due to the issue in step 2, bitmap generated during publish is not saved to RocksDB, so if BE restarts at this point, this part of the incremental delete bitmap will be lost
4. it will result in duplicated keys on querying

### DIFF
--- a/be/src/olap/base_tablet.cpp
+++ b/be/src/olap/base_tablet.cpp
@@ -1114,7 +1114,7 @@ void BaseTablet::_remove_sentinel_mark_from_delete_bitmap(DeleteBitmapPtr delete
     }
 }
 
-Status BaseTablet::update_delete_bitmap(const BaseTabletSPtr& self, const TabletTxnInfo* txn_info,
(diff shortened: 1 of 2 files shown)


zhannngchen:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #32429, opened 2024-03-18
[fix] Fix unrecognized column name delete handler #32404

gavinchou:
branch-2.0 #32404

PR #35742, opened 2024-05-31
[fix](delete) Fix unrecognized column name delete handler (#32429)

gavinchou:
pick doris-master #32429


gavinchou:
run buildall

PR #35608, opened 2024-05-29
[feature](be jvm monitor)append enable_jvm_monitor in be.conf to control jvm monitor.

hubgeter:
## Proposed changes

before pr : #35023
In order to prevent doris_be from crashing when collecting jvm information due to jvm incompatibility issues, you can set `enable_jvm_monitor = true / false` in `be.conf` to enable the jvm metrics.

The default value of `enable_jvm_monitor` is false.

When JVM monitoring has 30 consecutive exceptions, turn off JVM information collection and set all values ​​to 0.

Issue Number: close #xxx

PR #35764, opened 2024-06-01
[feature](be jvm monitor)append enable_jvm_monitor in be.conf to control jvm monitor. (#35608)

morningman:
bp #35608


morningman:
run buildall

PR #35766, opened 2024-06-01
[Fix](partial update) Handle delete predicate in partial Update

Yukang-Lian:
## Proposed changes

Issue Number: close #xxx


dataroaring (changes requested):
Please add a regression case.

Yukang-Lian:
run buildall

PR #35770, opened 2024-06-02
[fix](rpc) fix transfer large data and enable transfer_large_data_by_brpc by default

liaoxin01:
## Proposed changes

Issue Number: close #xxx

when send large data by rpc，load will report the following error:
```
add batch req success but status isn't ok, err: [INTERNAL_ERROR]PStatus: (172.200.0.1)[INTERNAL_ERROR]fail to add batch in load channel. unknown load_id=0000000000000000-0000000000000000.

```

The maximum length of protobuf is 2GB. When it exceeds 2GB, we will use the brpc HTTP method. However, the block data was not removed during encoding, resulting in protobuf exceeding the length and encoding failure.

The config of transfer_large_data_by_brpc was enabled by default on branch-2.0, so we also enable it on master.


liaoxin01:
run buildall

dataroaring (approved):
LGTM

dataroaring:
run performance

yiguolei (approved):

dataroaring:
run performance

PR #35773, opened 2024-06-03
[feat](Nereids): Optimize Join Penalty Calculation Based on Build Side Data Volume

keanji-x:
## Proposed changes

This PR introduces an optimization that adjusts the penalty applied during join operations based on the volume of data on the build side. Specifically, when the number of rows and width of the tables being joined are equal, the materialization costs are now considered more accurately. The update ensures that joins with a larger dataset on the build side incur a higher penalty, improving overall query performance and resource allocation.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/nereids/cost/CostModelV1.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/nereids/cost/CostModelV1.java
@@ -365,11 +365,11 @@ class CostModelV1 extends PlanVisitor<Cost, PlanContext> {
                 leftRowCount += 1;
             }
             if (probeStats.getWidthInJoinCluster() == buildStats.getWidthInJoinCluster()
-                    && probeStats.computeSize() < buildStats.computeSize()) {
+                    && probeStats.computeTupleSize() < buildStats.computeTupleSize()) {
                 // When the number of rows and the width on both sides of the join are the same,
                 // we need to consider the cost of materializing the output.
                 // When there is more data on the build side, a greater penalty will be given.
-                rightRowCount += 1;
+                leftRowCount += 1e-3;
             }
         }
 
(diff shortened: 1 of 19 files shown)


XieJiann:
run buildall

fe/fe-core/src/main/java/org/apache/doris/nereids/cost/CostModelV1.java
```diff
@@ -364,6 +364,13 @@ public Cost visitPhysicalHashJoin(
             } else if (probeStats.getWidthInJoinCluster() < buildStats.getWidthInJoinCluster()) {
                 leftRowCount += 1;
             }
+            if (probeStats.getWidthInJoinCluster() == buildStats.getWidthInJoinCluster()
+                    && probeStats.computeSize() < buildStats.computeSize()) {
+                // When the number of rows and the width on both sides of the join are the same,
```
xzj7019, line 369: The case has the same row count but different width, so we should use the "rowCount" and "computeTupleSize" to judge, instead of computeSize?

XieJiann: There is an equal row count condition outside already. And the computeTupleSize is private. When row count is same, the computSize is same to computeTupleSize

XieJiann:
run buildall

fe/fe-core/src/main/java/org/apache/doris/nereids/cost/CostModelV1.java
xzj7019: LGTM

XieJiann:
run buildall

XieJiann:
run buildall

xzj7019 (approved):

morrySnow (approved):

PR #35725, opened 2024-05-31
[Test](regression-test): do not use path style to access aliyun OSS

Thearas:
Change-Id: I09394806b418c1fba421c09f3aa6fc7ec3d2f5ad

## Proposed changes

This PR continues <https://github.com/apache/doris/pull/35666>.

PR #35666, opened 2024-05-30
[Test](regression-test): Normalize S3 http url

Thearas:
Change-Id: Ib992240a2538ac9d969df9f8dc109df8c38cae21

## Proposed changes

`bucket` should be in domain instead of HTTP path. Aliyun OSS does not support path style anymore.

PR #35788, opened 2024-06-03
[Test](regression-test): do not use path style to access s3

Thearas:
## Proposed changes

CP from #35666 and #35725.


yiguolei:
Add the related PR in master branch please.

PR #35801, opened 2024-06-03
[fix](regression) Fix txn_insert execute timeout

mymeiyi:
## Proposed changes

1. The execute timeout is set to 5 seconds and forget to rest, which will make `txn_insert` execute timeout.
2. Add a `insert overwrite` case which is not supported in txn insert.


mymeiyi:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #35805, opened 2024-06-03
[fix](third-party) enable keepalive on socket created by libevent

dataroaring:
## Proposed changes

We encounter a problem that tcp connections created by libevent is in established while there is no corresponding endpoint in the other endpoint. So we enable keepalive to recycle the tcp connection.


dataroaring:
run buildall

gavinchou (approved):

morningman (approved):
LGTM

PR #35860, opened 2024-06-04
[fix](mtmv)Mtmv support row column

zddr:
when create mtmv with property `'store_row_column' = 'true'`,and `select * from mtmv order by xxx limit xxx`
be will core,before MTMV not have hidden column `__DORIS_ROW_STORE_COL__`


zddr:
run buildall

morrySnow (approved):

seawinde (approved):

eldenmoon (approved):
LGTM

PR #35878, opened 2024-06-04
[Bug](drop) fix drop function has error when it's not exists

zhangstar333:
## Proposed changes

```
drop function xxx
drop function if exists xxx
```
as when drop UDTF function, we need drop two function, one is normal, another is the _outer function.
getFunction() should call firstly and always.  as dropFunctionImpl() will remove it from env, then we can't find it anymore.
so can't check if (ifExists) firstly, when it's false, we have no change call getFunction()
ifExists should only used to check whether throw execption.

Issue Number: close #xxx

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/Database.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/Database.java
@@ -748,14 +748,16 @@ public class Database extends MetaObject implements Writable, DatabaseIf<Table> 
 
     public synchronized void dropFunction(FunctionSearchDesc function, boolean ifExists) throws UserException {
         Function udfFunction = null;
-        if (ifExists) {
-            try {
-                // here we must first getFunction, as dropFunctionImpl will remove it
-                udfFunction = getFunction(function);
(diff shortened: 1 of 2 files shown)


zhangstar333:
run buildall

fe/fe-core/src/main/java/org/apache/doris/catalog/Database.java
```diff
@@ -748,14 +748,16 @@ public synchronized void replayAddFunction(Function function) {
...
-                udfFunction = getFunction(function);
-            } catch (AnalysisException e) {
-                // ignore it, as drop it if exist, so can't sure it must exist
+        try {
+            // here we must first getFunction, as dropFunctionImpl will remove it
+            udfFunction = getFunction(function);
+        } catch (AnalysisException e) {
+            if (!ifExists) {
+                throw new UserException(e);
             }
```
HappenLee, line 757: !ifExists just return

zhangstar333:
run buildall

HappenLee (approved):
LGTM

wangbo (approved):
LGTM

PR #35913, opened 2024-06-05
[fix](scan) In-predicate should not be pushed down for non-key column

mrhhsg:
## Proposed changes

The in-predicate associated with non-key columns should not be pushed down to the storage layer (as aggregation is required after reading it from storage).

PR #35914, opened 2024-06-05
[fix](scan) In-predicate should not be pushed down for non-key column

mrhhsg:
## Proposed changes

pick #35913


mrhhsg:
run buildall

mrhhsg:
run buildall

PR #35924, opened 2024-06-05
[fix](statistics)Fix partition table health rate evaluation bug.

Jibing-Li:
The older Doris version doesn't support partition stats collection, so when user upgrade their cluster, the metadata doesn't contain partition level update rows for each table. In this case, auto collector thought this table haven't collect partition stats,  which will trigger a new collection. Since the partition level update rows are only updated after load operation, if user stop loading, auto analyzing will keep collecting this table for ever. 
This pr is to fix this problem, set the partition update rows to 0 for the old version tables.


morrySnow (approved):

Jibing-Li:
run buildall

morningman (approved):

PR #35942, opened 2024-06-05
[opt](log) refine the BE logger 

morningman:
Followup #35679

## Background

Previously, BE logs were written to files. The main FE logs include be.INFO, be.WARNING, be.out, be.gc.log and jni.log.
In a K8s deployment environment, logs usually need to be output to standard output, and then other components process the log stream.

## Solution

This PR made the following changes:

1. Modified the glog config:

	- When started with `--daemon`, logs are still written to various files, and the format remains unchanged.
	- When started with `--console`, be.INFO's log is output to standard output and marked with prefix `RuntimeLogger`.
		
		Examples are as follows:
		
		```
		RuntimeLogger I20240605 23:41:20.426553 4137369 runtime_query_statistics_mgr.cpp:245] Report profile thread stopped
		```

2. Added a new BE config: `enable_file_logger`

	Defaults to true. Indicates that logs will be recorded to files regardless of the startup method. For example, if it is started with `--console`, the log will be output to both the file and the standard output. If it is `false`, the log will not be recorded in the file regardless of the startup method.

## TODO

- Need to handle jni.log in next PR

### DIFF
--- a/be/CMakeLists.txt
+++ b/be/CMakeLists.txt
@@ -66,6 +66,9 @@ else()
     add_definitions(-DBOOST_STACKTRACE_USE_NOOP)
 endif()
 
+# enable glog custom prefix
+add_definitions(-DGLOG_CUSTOM_PREFIX_SUPPORT)
+
 # Options
 option(GLIBC_COMPATIBILITY "Enable compatibility with older glibc libraries." ON)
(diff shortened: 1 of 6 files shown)


morningman:
run buildall

intelligentfu (approved):
done

CalvinKirs (approved):

PR #35886, opened 2024-06-05
[fix](multicast) should not ignore Status of block::merge

jacktengg:
## Proposed changes

Issue Number: close #xxx

PR #35945, opened 2024-06-06
[fix](multicast) should not ignore Status of block::merge #35886

jacktengg:
## Proposed changes

Pick #35886


jacktengg:
run buildall

PR #35948, opened 2024-06-06
[fix](nereids)should not push down constant output expr in window expression

starocean999:
`SELECT 0 AS x1, (CASE WHEN t.a IS NOT NULL THEN 1.0 / count(1) OVER (PARTITION BY t.b) ELSE 0 END) AS x2 FROM t`

0 AS x1 was pushed down to bottom project. Then it will be used to normalize case when expr, and change it to
(CASE WHEN t.a IS NOT NULL THEN 1.0 / count(1) OVER (PARTITION BY t.b) ELSE **x1** END) AS x2

the literal 0 was replaced by x1 accidentally. This pr fix it by keep constant expr in output list

Issue Number: close #xxx


starocean999:
run buildall

morrySnow (approved):

924060929 (approved):

PR #35951, opened 2024-06-06
[fix](inverted index)Delete tmp dirs when BE starts to avoid tmp file…

qidaye:
…s left by last crash

## Proposed changes

When BE crashes, there may be tmp files left in the tmp dir, so we remove and rebuild the tmp dir every time we start BE to prevent rubbish data from occupying the disk.

### DIFF
--- a/be/src/olap/compaction.cpp
+++ b/be/src/olap/compaction.cpp
@@ -663,8 +663,7 @@ Status Compaction::do_inverted_index_compaction() {
         }
     }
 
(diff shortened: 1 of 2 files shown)


qidaye:
run buildall

qidaye:
run performance

airborne12 (approved):
LGTM

eldenmoon (approved):
LGTM

PR #35956, opened 2024-06-06
[fix](mtmv)Mtmv support row column (#35860)

zddr:
pick from master: https://github.com/apache/doris/pull/35860

when create mtmv with property `'store_row_column' = 'true'`,and `select
* from mtmv order by xxx limit xxx` be will core,before MTMV not have hidden column
`__DORIS_ROW_STORE_COL__`

## Proposed changes

Issue Number: close #xxx


zddr:
run buildall

zddr:
run feut

PR #35957, opened 2024-06-06
[bugfix](iceberg)Fix the datafile path error issue

wuwenchi:
## Proposed changes

Doris will convert the schema in FIleSystem to achieve the function of natively reading files. For example, `s3a` will be converted to `s3`.
The position delete file of iceberg will record the full path of the datafile, which includes the schema.
When comparing datafile with position delete, the converted path cannot be used, but the original datafile path must be used.
So, added variables for the original path.


wuwenchi:
run buildall

morningman (approved):
LGTM

zy-kkk (approved):

PR #35961, opened 2024-06-06
[fix](nereids)should not push down constant output expr in window expression

starocean999:
pick from master https://github.com/apache/doris/pull/35948

## Proposed changes

Issue Number: close #xxx


starocean999:
run buildall

PR #35969, opened 2024-06-06
[BugFix](StoragePolicy) fix modify partition's storage policy with different resource issue succeed with execption 

Johnnyssc:
## Proposed changes

1. when we try to modify a partition's storage policy which have different resource with the previous one, we get error message "currently do not support change origin storage policy to another one with different resource", but the modification same also succeed!

2. if we modify partition storage policy with policy_1 based on resource1, after that we set this table's storage policy with policy_2, it can still succeed and partitions are all set to policy_2, which may result in some cold data with policy_1 missing 

expect:
1. the modification fail with "currently do not support change origin storage policy to another one with different resource".

2. if set table policy ops after modify partition policy ops, it will be failed with exception

## Further comments


Johnnyssc:
run buildall

Johnnyssc:
run buildall

Johnnyssc:
run buildall

Johnnyssc:
run buildall

Johnnyssc:
run buildall

PR #32793, opened 2024-03-25
[fix](gc) Skip tablet dir without schema hash dir in path gc

platoneko:
## Proposed changes

Skip empty tablet dir (without schema hash dir) in path gc.
Adjusting path gc process for better readability.
Old:
```
1. perform path scan to get all tablet directory `tablet_paths`
2. gc unused tablet dir in `tablet_paths`, and erase unused or invalid tablet dir from `tablet_paths`
3. gc unused rowsets under left `tablet_paths`
```
New:
```
for shard_path in list(data_dir)
    for tablet_path in list(shard_path)
        gc tablet dir
            gc rowsets under tablet dir 
```

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #35978, opened 2024-06-06
Pick "[fix](gclog) Skip tablet dir without schema hash dir in path gc (#32793)"

platoneko:
## Proposed changes
Pick "[fix](gclog) Skip tablet dir without schema hash dir in path gc (#32793)"


platoneko:
run buildall

PR #35309, opened 2024-05-23
[fix](paimon)fix paimon cache bug

suxiaogang223:
## Proposed changes
Issue Number: close #35024 
This bug is because the fe incorrectly sets the update time of paimon catalog, causing the be to be unable to update paimon's schema in time.
```c++
    private void initTable() {
        PaimonTableCacheKey key = new PaimonTableCacheKey(ctlId, dbId, tblId, paimonOptionParams, dbName, tblName);
        TableExt tableExt = PaimonTableCache.getTable(key);
        if (tableExt.getCreateTime() < lastUpdateTime) {
            LOG.warn("invalidate cache table:{}, localTime:{}, remoteTime:{}", key, tableExt.getCreateTime(),
                    lastUpdateTime);
            PaimonTableCache.invalidateTableCache(key);
            tableExt = PaimonTableCache.getTable(key);
        }
        this.table = tableExt.getTable();
        paimonAllFieldNames = PaimonScannerUtils.fieldNames(this.table.rowType());
        if (LOG.isDebugEnabled()) {
            LOG.debug("paimonAllFieldNames:{}", paimonAllFieldNames);
        }
    }
```

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #35981, opened 2024-06-06
[bugfix](paimon)paimon's field length judgment error

wuwenchi:
## Proposed changes

 Introduced from #35309
`fields.length` indicates the fields we want to read. If a table has 10 fields, but we only select a few fields, then the length of `fields.length` will definitely be less than 10. 
So the judgment condition here should be greater than.


wuwenchi:
run buildall

morningman (approved):
LGTM

zddr (approved):

PR #34990, opened 2024-05-17
[fix](multi-catalog)put java udf to custom lib

wsjz:
## Proposed changes

when use java udf , we can put its jars in custom_lib of FE and BE in advance， and then we can create function without file property:
```
CREATE FUNCTION java_udf_add_one(int) RETURNS int PROPERTIES (
    "symbol"="org.apache.doris.udf.AddOne",
    "always_nullable"="true",
    "type"="JAVA_UDF"
);
```

and we can use file property for custom jar:  

```
CREATE FUNCTION java_udf_add_one(int) RETURNS int PROPERTIES (
    "file"="file:///path/to/java-udf-demo.jar",
    "symbol"="org.apache.doris.udf.AddOne",
    "always_nullable"="true",
    "type"="JAVA_UDF"
);
```

on FE, we use null location if the 'file' property is not set
on BE, we get jar path if both file path location and checksum are null

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #35986, opened 2024-06-06
[fix](multi-catalog)the classpath of custom lib should put back of fe jars

wsjz:
## Proposed changes

see https://github.com/apache/doris/pull/34990
user class maybe use some tripartite lib, we should put them back of fe jars to avoid conflicts


hello-stephen (approved):
LGTM

morningman (approved):
LGTM

PR #35987, opened 2024-06-06
[Fix](feut) Fix distribute hint test with reorder cross join

LiBinfeng-01:
when checking plan shape of distribute hint test, the shape of it does not changed but projection has changed. So it needed to be updated


morningman:
run buildall

PR #35995, opened 2024-06-06
[improve](group commit) Modify group commit case and modify cancel status

mymeiyi:
## Proposed changes

1. Modify group commit p2 case to use nereids since https://github.com/apache/doris/pull/32523 support it.
2. We get many same logs when group commit is cancelled: 
    ```
    group_commit_mgr.cpp:182] cancel group commit, instance_id=xx, label=xx, status=[OK]
    ```
    the problem is that `runtime_state->is_cancelled` is true, but `runtime_state->cancel_reason()` is `OK`


dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #35997, opened 2024-06-06
[Pick 2.1](inverted index) fix tokenize function wrong result when params with space seperator 

airborne12:
## Proposed changes

Issue Number: close #xxx

Pick from (#32671)


airborne12:
run buildall

PR #36000, opened 2024-06-06
[Bug](function) fix week function get wrong result

zhangstar333:
## Proposed changes

before:
```
select week(cast('0000-01-01' as DATEV2), cast(2 as INT));
+----------------------------------------------------+
| week(cast('0000-01-01' as DATEV2), cast(2 as INT)) |
+----------------------------------------------------+
|                                                 52 |
+----------------------------------------------------+
1 row in set (0.00 sec)
```
 but it's should return 1 as result,
 mode 2 is start with a Sunday day as first week in this year.
 and special case for 0000-01-01, as it's SATURDAY, calculate result of 52 is
 last year, so it's meaningless.


zhangstar333:
run buildall

zclllyybb (commented):
lgtm for the fix. but we should add both fe-ut and be-ut for them. maybe cover the first 7 days and last 7 days for year 0000

zhangstar333:
run buildall

zhangstar333:
run buildall

zclllyybb (approved):
LGTM

HappenLee (approved):
LGTM

PR #36004, opened 2024-06-06
[fix](eq_for_null) fix incorrect logic in function eq_for_null

zhiqiang-hhhh:
Make sure we can process ColumnConst correctly

### DIFF
--- a/be/src/vec/functions/comparison_equal_for_null.cpp
+++ b/be/src/vec/functions/comparison_equal_for_null.cpp
@@ -87,7 +87,8 @@ public:
                 const ColumnNullable* nullable_right_col = nullptr;
                 if (right_const) {
                     nullable_right_col = assert_cast<const ColumnNullable*>(
-                            &(assert_cast<const ColumnConst*>(col_right.column.get())->get_data_column()));
+                            &(assert_cast<const ColumnConst*>(col_right.column.get())
+                                      ->get_data_column()));
                 } else {
                     nullable_right_col = assert_cast<const ColumnNullable*>(col_right.column.get());
                 }
@@ -106,7 +107,8 @@ public:
                 const ColumnNullable* nullable_left_col = nullptr;
                 if (left_const) {
                     nullable_left_col = assert_cast<const ColumnNullable*>(
-                            &(assert_cast<const ColumnConst*>(col_left.column.get())->get_data_column()));
+                            &(assert_cast<const ColumnConst*>(col_left.column.get())
+                                      ->get_data_column()));
                 } else {
                     nullable_left_col = assert_cast<const ColumnNullable*>(col_left.column.get());
                 }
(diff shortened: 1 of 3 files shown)


zhiqiang-hhhh:
run buildall

be/src/vec/functions/comparison_equal_for_null.cpp
```diff
@@ -75,33 +79,62 @@ class FunctionEqForNull : public IFunction {
         } else if (left_only_null) {
             auto right_type_nullable = col_right.type->is_nullable();
             if (!right_type_nullable) {
+                // right_column is not nullable, so result is all false.
                 block.get_by_position(result).column =
                         ColumnVector<UInt8>::create(input_rows_count, 0);
```
zclllyybb, line 84: these result columns could be wrap in `ColumnConst`. if then we use this column, it could be accelerated

be/src/vec/functions/comparison_equal_for_null.cpp
```diff
@@ -75,33 +79,62 @@ class FunctionEqForNull : public IFunction {
...
+                    nullable_right_col = assert_cast<const ColumnNullable*>(col_right.column.get());
+                }
+                // left column is all null, so result has same nullmap with right column.
                 block.get_by_position(result).column =
                         nullable_right_col->get_null_map_column().clone_resized(input_rows_count);
             }
             return Status::OK();
         } else if (right_only_null) {
             auto left_type_nullable = col_left.type->is_nullable();
             if (!left_type_nullable) {
```
zclllyybb, line 102: maybe put positive branch in `if` is better

zclllyybb (approved):
LGTM. some nit(not block):

be/src/vec/functions/comparison_equal_for_null.cpp
zhiqiang-hhhh: Can not do this, return ColumnConst may have conflict with default impl for const.

be/src/vec/functions/comparison_equal_for_null.cpp
zhiqiang-hhhh: bug fix should not be mixed with refactor

be/src/vec/functions/comparison_equal_for_null.cpp
```diff
@@ -75,33 +79,62 @@ class FunctionEqForNull : public IFunction {
...
+                const ColumnNullable* nullable_right_col = nullptr;
+                if (right_const) {
+                    nullable_right_col = assert_cast<const ColumnNullable*>(
+                            &(assert_cast<const ColumnConst*>(col_right.column.get())
+                                      ->get_data_column()));
+                } else {
+                    nullable_right_col = assert_cast<const ColumnNullable*>(col_right.column.get());
+                }
+                // left column is all null, so result has same nullmap with right column.
                 block.get_by_position(result).column =
```
HappenLee, line 96: if right_const, here do `clone_reisze` is error;

zhiqiang-hhhh:
run buildall

HappenLee (approved):
LGTM

PR #36009, opened 2024-06-07
[branch-2.1](jvm) disable BE's jvm metrics

morningman:
disable  BE's jvm metrics on external p0, because there is some issue with ASAN when BE exit.


morningman:
run buildall

morningman:
run buildall

PR #35491, opened 2024-05-28
[fix](backup) save finished state for local repo backup

w41ter:
## Proposed changes

Issue Number: close #xxx

The backup finished state in the local repo is not logged. So if there exist multiple local repo backups, and FE was restarted, then these backup records will be reset to the UPLOAD_INFO state.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #36014, opened 2024-06-07
[fix](backup) save finished state for local repo backup

w41ter:
Cherry-pick #35491

The backup finished state in the local repo is not logged. So if there exist multiple local repo backups, and FE was restarted, then these backup records will be reset to the UPLOAD_INFO state.


w41ter:
run buildall

PR #36024, opened 2024-06-07
[fix](in expr) fix error result when in expr has null value  and  left expr is 0 

Mryange:
## Proposed changes

```
create table t2 (id int, c1 int);
insert into t2 values(1, null);
 select 0 in (c1, null) from t2; -- should return null，but 1
```


Mryange:
run buildall

HappenLee (approved):
LGTM

Gabriel39 (approved):

PR #36028, opened 2024-06-07
[Fix](explain) fix tablet showing problem (#35830)

LiBinfeng-01:
cherry-pick: #35830 
tablets could not showed in explain string when selected partitions too large, so it need to be show in next line


LiBinfeng-01:
run buildall

morningman:
run buildall

PR #36033, opened 2024-06-07
[improvement](JDBC Catalog) Properly release resources in JdbcClient

zy-kkk:
### What changes were proposed in this pull request?

This pull request introduces modifications to the `closeClient` method in the `JdbcClient` class to ensure proper cleanup of the custom class loader used for loading JDBC drivers. The existing implementation does not address the potential memory leak associated with class loaders when they are not properly closed after their usage ends.

### Why are the changes needed?

Class loaders, when not closed properly, can lead to memory leaks as the objects they load may not be eligible for garbage collection. This is especially critical in environments where multiple JDBC drivers are loaded and unloaded during the lifecycle of the application, such as in database federated queries or multi-tenant databases.

### Changes made:

- Added a check to close the `URLClassLoader` instance explicitly if it is not null.
- Nullified the class loader reference to aid the garbage collection process, thereby reducing the risk of memory leaks.

These changes help in ensuring that all resources are properly released when the JDBC client is closed, maintaining the robustness and efficiency of the resource management within the application.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/datasource/jdbc/client/JdbcClient.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/datasource/jdbc/client/JdbcClient.java
@@ -34,6 +34,7 @@ import lombok.Getter;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import java.io.IOException;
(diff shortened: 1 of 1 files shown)


zy-kkk:
run buildall

zy-kkk:
run buildall

PR #36042, opened 2024-06-07
[opt](parquet-writer) Specify the row group size when writing data to Parquet files. (#35081)

morningman:
bp #35081


morningman:
run buildall

PR #36046, opened 2024-06-07
[fix](statistics)Fix analyze twice after drop stats bug.

Jibing-Li:
1. After drop stats when partition analyze function enabled. Auto collector will collect the dropped table twice. Because int the first time analyzing, it doesn't write the partition update rows record to tableStats. In this case, the collector will consider the table has not collect partition level stats, which will trigger the second time collection. This pr is to fix this bug.
2. Set correct trigger type for partition analyze.
3. Sleep for 2 minutes before auto collection thread begin to work. This is waiting for BEs report tablet row count.


Jibing-Li:
run buildall

morrySnow (approved):

morningman (approved):

PR #36049, opened 2024-06-07
[bugfix](paimon)paimon's field length judgment error for 2.1

wuwenchi:
## Proposed changes

bp #35981


morningman:
run buildall

PR #35729, opened 2024-05-31
[fix](mtmv)fix mtmv show partition error when base table dropped

zddr:
when partition of base table dropped,is sync of show partitions is false

when base table dropped,`UnsyncTables` display error message, such as: errCode=2, detailMessage = Unknown table '16011'

PR #36051, opened 2024-06-07
[fix](mtmv)fix mtmv show partition error when base table dropped (#35…

zddr:
…729)

pick from master:https://github.com/apache/doris/pull/35729

when partition of base table dropped,is sync of show partitions is false

when base table dropped,`UnsyncTables` display error message, such as: errCode=2, detailMessage = Unknown table '16011'

## Proposed changes

Issue Number: close #xxx


zddr:
run buildall

PR #36055, opened 2024-06-07
[fix](nereids)decimal and datetime literal comparison should compare datatype too

starocean999:
this pr revert https://github.com/apache/doris/pull/35948 and fix the bug in correct way.

The root cause is decimal literal's comparison only compare value, we should also compare their datatype as well as datetime literal.

suppose 
a=0 with type decimal(2,1)
b=0 with type decimal(3,1)
we have:
a.equals(b) returns false

## Proposed changes

Issue Number: close #xxx


starocean999:
run buildall

starocean999:
run buildall

morrySnow (approved):

englefly (approved):

PR #36064, opened 2024-06-07
[fix](nereids)decimal and datetime literal comparison should compare datatype too

starocean999:
pick from master https://github.com/apache/doris/pull/36055

Issue Number: close #xxx


starocean999:
run buildall

PR #36071, opened 2024-06-08
[fix](pipeline) failure if memory is not enough

dataroaring:
## Proposed changes

Issue Number: close #xxx

### DIFF
--- a/be/src/pipeline/pipeline_fragment_context.cpp
+++ b/be/src/pipeline/pipeline_fragment_context.cpp
@@ -337,13 +337,13 @@ Status PipelineFragmentContext::prepare(const doris::TPipelineFragmentParams& re
     for (PipelinePtr& pipeline : _pipelines) {
         SCOPED_TIMER(_prepare_all_pipelines_timer);
         pipeline->children().clear();
-        RETURN_IF_ERROR(pipeline->prepare(_runtime_state.get()));
+        RETURN_IF_ERROR_OR_CATCH_EXCEPTION(pipeline->prepare(_runtime_state.get()));
     }
 
     {
         SCOPED_TIMER(_build_tasks_timer);
         // 5. Build pipeline tasks and initialize local state.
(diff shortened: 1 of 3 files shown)


dataroaring:
run buildall

gavinchou (approved):

be/src/pipeline/pipeline_task.cpp
```diff
@@ -302,7 +302,7 @@ Status PipelineTask::execute(bool* eos) {
     }
     // The status must be runnable
     if (!_opened && !_fragment_context->is_canceled()) {
-        RETURN_IF_ERROR(_open());
```
Gabriel39, line 305: Exception is already caught by caller

be/src/pipeline/pipeline_fragment_context.cpp
```diff
@@ -337,13 +337,13 @@ Status PipelineFragmentContext::prepare(const doris::TPipelineFragmentParams& re
     for (PipelinePtr& pipeline : _pipelines) {
         SCOPED_TIMER(_prepare_all_pipelines_timer);
         pipeline->children().clear();
-        RETURN_IF_ERROR(pipeline->prepare(_runtime_state.get()));
+        RETURN_IF_ERROR_OR_CATCH_EXCEPTION(pipeline->prepare(_runtime_state.get()));
```
Gabriel39, line 340: Maybe we should catch exception by `prepare`'s caller

be/src/vec/exprs/vexpr.cpp
```diff
@@ -343,7 +343,7 @@ Status VExpr::create_tree_from_thrift(const std::vector<TExprNode>& nodes, int*
     // create root expr
     int root_children = nodes[*node_idx].num_children;
     VExprSPtr root;
-    RETURN_IF_ERROR(create_expr(nodes[*node_idx], root));
```
Gabriel39, line 346: Catch exception in `FoldConstantExecutor::fold_constant_vexpr`

PR #36075, opened 2024-06-08
[log](FE) Add switch to control whether print log before execution

gavinchou:
The audit log prints the query when the query finished, however, we cannot known when the query starts exactly.
To solve this problem, add a config option
enable_print_request_before_execution, default false. It may be set to true by default in the future.


gavinchou:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #36082, opened 2024-06-10
[Improvementation](runtime-filter) do not sync filter size when rf has local target

BiteTheDDDDt:
## Proposed changes
do not sync filter size when rf has local target


BiteTheDDDDt:
run buildall

PR #36094, opened 2024-06-11
[fix](snappy) avoid potential buffer overflow (#35537)

dataroaring:
pick #35537

If skip more than once when available is zero, then a buffer overflow occurs.

photo-size-5-6244711526321733357-y

## Proposed changes

Issue Number: close #xxx


dataroaring:
run buildall

PR #35287, opened 2024-05-23
[fix](auto-partition) fix auto partition load lost data in multi sender

kaijchen:
## Proposed changes

Change `use_cnt` mechanism for incremental (auto partition) channels and streams, it's now dynamically counted.
Use `close_wait()` of regular partitions as a synchronize point to make sure all sinks are in close phase before closing any incremental (auto partition) channels and streams.
Add dummy (fake) partition and tablet if there is no regular partition in the auto partition table.

Replace #34740

Co-authored-by: zhaochangle <zhaochangle@selectdb.com>

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #36098, opened 2024-06-11
Revert "[fix](auto-partition) fix auto partition load lost data in multi sender (#35287)"

zclllyybb:
Reverts apache/doris#35630 because it brought some more damaging bugs. we will fix it and merge in next version


zclllyybb:
run buildall

PR #36104, opened 2024-06-11
[enhance](Vault) Refresh default storage vault info when doing instance check in FE

ByteYue:
## Proposed changes

Previously the default vault info would only be refreshed when user tries to set default storage vault. The default vault info might be lost if the FE crashed and restarted. This pr tries to add one routine work which would refresh the default vault info.


ByteYue:
run buildall

dataroaring (approved):
LGTM

platoneko (approved):

PR #36107, opened 2024-06-11
[opt](Nereids) Optimize Join Penalty Calculation Based on Build Side Data Volume (#35773)

keanji-x:
## Proposed changes
#35773

This PR introduces an optimization that adjusts the penalty applied during join operations based on the volume of data on the build side. Specifically, when the number of rows and width of the tables being joined are equal, the materialization costs are now considered more accurately. The update ensures that joins with a larger dataset on the build side incur a higher penalty, improving overall query performance and resource allocation.

Issue Number: close #xxx


XieJiann:
run buildall

XieJiann:
run buildall

PR #36110, opened 2024-06-11
[fix](function) error scale set in unix_timestamp

Mryange:
## Proposed changes

```
mysql [test]>set DEBUG_SKIP_FOLD_CONSTANT = true;

mysql [test]>select cast(unix_timestamp("2024-01-01",'yyyy-MM-dd') as bigint);
+------------------------------------------------------------+
| cast(unix_timestamp('2024-01-01', 'yyyy-MM-dd') as BIGINT) |
+------------------------------------------------------------+
|                                           1704038400000000 |
+------------------------------------------------------------+
```
now
```
mysql [test]>select cast(unix_timestamp("2024-01-01",'yyyy-MM-dd') as bigint);
+------------------------------------------------------------+
| cast(unix_timestamp('2024-01-01', 'yyyy-MM-dd') as BIGINT) |
+------------------------------------------------------------+
|                                                 1704038400 |
+------------------------------------------------------------+
1 row in set (0.01 sec)
```

The column does not have a scale set, but the cast uses the scale to perform the cast.

### DIFF
--- a/be/src/vec/functions/function_timestamp.cpp
+++ b/be/src/vec/functions/function_timestamp.cpp
@@ -702,7 +702,7 @@ struct UnixTimeStampStrImpl {
         std::tie(col_right, format_const) =
                 unpack_if_const(block.get_by_position(arguments[1]).column);
 
-        auto col_result = ColumnDecimal<Decimal64>::create(input_rows_count, 0);
+        auto col_result = ColumnDecimal<Decimal64>::create(input_rows_count, 6);
         auto null_map = ColumnVector<UInt8>::create(input_rows_count);
         auto& col_result_data = col_result->get_data();
         auto& null_map_data = null_map->get_data();
--- a/regression-test/data/query_p0/sql_functions/datetime_functions/test_date_function.out
+++ b/regression-test/data/query_p0/sql_functions/datetime_functions/test_date_function.out
@@ -380,6 +380,9 @@ February
 -- !sql_ustamp7 --
 1196389819.1235
 
+-- !sql_ustamp8 --
+1704038400
+
 -- !sql --
 0
 
--- a/regression-test/suites/query_p0/sql_functions/datetime_functions/test_date_function.groovy
+++ b/regression-test/suites/query_p0/sql_functions/datetime_functions/test_date_function.groovy
@@ -441,6 +441,7 @@ suite("test_date_function") {
     // UNIX_TIMESTAMP
     def unin_timestamp_str = """ select unix_timestamp() """
     assertTrue(unin_timestamp_str[0].size() == 1)
+    sql """set DEBUG_SKIP_FOLD_CONSTANT = true;"""
     qt_sql_ustamp1 """ select unix_timestamp('2007-11-30 10:30:19') """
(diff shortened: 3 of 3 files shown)


Mryange:
run buildall

be/src/vec/functions/function_timestamp.cpp
```diff
@@ -702,7 +702,7 @@ struct UnixTimeStampStrImpl {
         std::tie(col_right, format_const) =
                 unpack_if_const(block.get_by_position(arguments[1]).column);
 
-        auto col_result = ColumnDecimal<Decimal64>::create(input_rows_count, 0);
+        auto col_result = ColumnDecimal<Decimal64>::create(input_rows_count, 6);
```
zhiqiang-hhhh, line 705: why not using scale from planner? like `block.get_by_position(result).type->get_scale()`

Mryange: Since the return type is fixed,
static DataTypePtr get_return_type_impl(const ColumnsWithTypeAndName& arguments) {
    return make_nullable(std::make_shared<DataTypeDecimal<Decimal64>>(16, 6));
}

Mryange: > why not using scale from planner? like `block.get_by_position(result).type->get_scale()`

Since the return type is fixed,
static DataTypePtr get_return_type_impl(const ColumnsWithTypeAndName& arguments) {
    return make_nullable(std::make_shared<DataTypeDecimal<Decimal64>>(16, 6));
}

zhiqiang-hhhh: does FE have same assumption?

Mryange: > does FE have same assumption?

    private static final List<FunctionSignature> SIGNATURES = ImmutableList.of(
            FunctionSignature.ret(IntegerType.INSTANCE).args(),
            FunctionSignature.ret(DecimalV3Type.WILDCARD).args(DateTimeV2Type.SYSTEM_DEFAULT),
            FunctionSignature.ret(IntegerType.INSTANCE).args(DateV2Type.INSTANCE),
            FunctionSignature.ret(IntegerType.INSTANCE).args(DateTimeType.INSTANCE),
            FunctionSignature.ret(IntegerType.INSTANCE).args(DateType.INSTANCE),
            FunctionSignature.ret(DecimalV3Type.createDecimalV3Type(16, 6)).args(VarcharType.SYSTEM_DEFAULT,
                    VarcharType.SYSTEM_DEFAULT),
            FunctionSignature.ret(DecimalV3Type.createDecimalV3Type(16, 6)).args(StringType.INSTANCE,
                    StringType.INSTANCE)
    );

zhiqiang-hhhh: what if this signature changed?

Mryange: > what if this signature changed?

The range of UNIX_TIMESTAMP is fixed and cannot be changed.

HappenLee (approved):
LGTM

wangbo (approved):
LGTM

PR #36127, opened 2024-06-11
[branch-2.1](test) fix some tests in external p0

morningman:
Also move the analysis exception of "Not support insert with partition spec in hive catalog."
from create sink phase to bind sink phase.
So that when `set enable_fallback_to_original_planner=false;`, the return error will be correct.

master: #36149


morningman:
run buildall

morrySnow (approved):

morningman:
run buildall

PR #35871, opened 2024-06-04
[enhancement](Nereids) support 4 phases distinct aggregate with full distribution

924060929:
## Proposed changes

The origin implementation of 4 phases distinct aggregate only support the pattern which not contains `group by`, and only one distinct aggregate function

for example:
```sql
select count(distinct sex), sum(age)
from student
```

This pr complement the 4 phases distinct aggregate with full distribution, to avoid data skew in the `group by`.

for example
```sql
select sex, sum(distinct age)
from student
group by sex;
```
The sex only contains two distinct values, `male` and `female`, and the table store millions rows.
Shuffle by the `sex` cause the data skew and lots of instances process empty rows.

The 4 phase aggregate shuffle `sex, age` to distinct rows first, so more instances can do parallel distinct, the plan shape will like this:
```

PhysicalAggregate(groupBy=[sex], output=[sex, sum(partial_sum(age))], mode=BUFFER_TO_RESULT)
                                        |
                         PhysicalDistribute(columns=[sex])
                                        |
PhysicalAggregate(groupBy=[sex], output=[sex, partial_sum(age)], mode=INPUT_TO_BUFFER)
                                        |
    PhysicalAggregate(groupBy=[sex, age], output=[sex, age], mode=BUFFER_TO_BUFFER)
                                        |
                         PhysicalDistribute(columns=[sex, age])   # more columns to shuffle avoid data skew
                                        |
PhysicalAggregate(groupBy=[sex, age], output=[sex, age], mode=INPUT_TO_BUFFER)
                                        |
                          PhysicalOlapScan(name=student)
```

PR #36128, opened 2024-06-11
[fix](Nereids) fix four phase aggregation compute wrong result

924060929:
## Proposed changes

fix a bug introduced by #35871

Backend can process distinct aggregate by two method:
1. use group by key to process distinct, I call it: external distinct of aggregate function
2. use `multi_distinct_xxx` to process distinct, in the corresponding aggregate function, I call it: internal distinct of aggregate function

This sql should process distinct by the multi distinct function when use four phase aggregation:
The first two phase use group by key: id,  field1 and field2.
The last two phase use group by key: id. This group by key can not process distinct in the aggregate function, so we should use internal distinct method in four phase aggregate function, if contains multi distinct column(e.g. field1 and field2) 

```sql
select /*+SET_VAR(disable_nereids_rules='TWO_PHASE_AGGREGATE_SINGLE_DISTINCT_TO_MULTI,TWO_PHASE_AGGREGATE_WITH_MULTI_DISTINCT,THREE_PHASE_AGGREGATE_WITH_DISTINCT,THREE_PHASE_AGGREGATE_WITH_COUNT_DISTINCT_MULTI')*/
  id,
  count(distinct field1),
  count(distinct field2)
from agg_4_phase_tbl2
group by id
```


924060929:
run buildall

morrySnow (approved):

starocean999 (approved):

PR #36144, opened 2024-06-11
[Fix](inverted index) fix memory leak in inverted index writer for array values

airborne12:
## Proposed changes

Issue Number: close #xxx

Add ownership flag to Field's TokenStream value and Analyzer's Reader value, free them when tokenizer finalize.

### DIFF
--- a/be/src/olap/rowset/segment_v2/inverted_index_writer.cpp
+++ b/be/src/olap/rowset/segment_v2/inverted_index_writer.cpp
@@ -404,12 +404,15 @@ public:
                             _parser_type != InvertedIndexParserType::PARSER_NONE) {
                             // in this case stream need to delete after add_document, because the
                             // stream can not reuse for different field
+                            bool own_token_stream = true;
+                            bool own_reader = true;
                             std::unique_ptr<lucene::util::Reader> char_string_reader = nullptr;
                             RETURN_IF_ERROR(create_char_string_reader(char_string_reader));
                             char_string_reader->init(v->get_data(), v->get_size(), false);
                             ts = _analyzer->tokenStream(new_field->name(),
                                                         char_string_reader.release());
-                            new_field->setValue(ts);
+                            _analyzer->set_ownReader(own_reader);
+                            new_field->setValue(ts, own_token_stream);
                         } else {
                             new_field_char_value(v->get_data(), v->get_size(), new_field);
                         }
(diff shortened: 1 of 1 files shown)


be/src/olap/rowset/segment_v2/inverted_index_writer.cpp
```diff
@@ -404,12 +404,15 @@ class InvertedIndexColumnWriterImpl : public InvertedIndexColumnWriter {
                             _parser_type != InvertedIndexParserType::PARSER_NONE) {
                             // in this case stream need to delete after add_document, because the
                             // stream can not reuse for different field
+                            bool own_token_stream = true;
```
xiaokang, line 407: Only used once

xiaokang (commented):
pls add a testcase

be/src/olap/rowset/segment_v2/inverted_index_writer.cpp
airborne12: yes, to make code more readable.

airborne12: by the way, no need more case, p0 case can cause the memory leak.

be/src/olap/rowset/segment_v2/inverted_index_writer.cpp
```diff
@@ -421,7 +424,6 @@ class InvertedIndexColumnWriterImpl : public InvertedIndexColumnWriter {
                     // if this array is null, we just ignore to write inverted index
                     RETURN_IF_ERROR(add_document());
                     _doc->clear();
-                    _CLDELETE(ts);
```
amorynan, line 424: why delete here will make memory leak?

amorynan, line 424: why delete here will make memory leak?

zzzxl1993 (approved):
LGTM

be/src/olap/rowset/segment_v2/inverted_index_writer.cpp
airborne12: The Field now takes ownership of the TokenStream, eliminating the need for manual deletion.

airborne12: The field now takes ownership of the TokenStream, eliminating the need for manual deletion.

qidaye (dismissed):

qidaye:
run  buildall

airborne12:
run buildall

qidaye (approved):

PR #36159, opened 2024-06-12
[Bug](function) fix yearweek function get wrong result

zhangstar333:
## Proposed changes
before:
```
mysql [(none)]>select yearweek(cast('0000-01-01-00:00:00.00001' as DATETIMEV2(5)), cast(2 as INT));
+------------------------------------------------------------------------------+
| yearweek(cast('0000-01-01-00:00:00.00001' as DATETIMEV2(5)), cast(2 as INT)) |
+------------------------------------------------------------------------------+
|                                                                          -48 |
+------------------------------------------------------------------------------+
```
but it's should return 1 as result,
and special case for 0000-01-01, calculate result of -48 is last year, so it's meaningless.


zhangstar333:
run buildall

morrySnow (approved):

zhangstar333:
run buildall

zclllyybb (approved):
LGTM

PR #36167, opened 2024-06-12
[fix](rpc) fix transfer large data and enable transfer_large_data_by_brpc by default #35770

liaoxin01:
cherry pick from #35770


liaoxin01:
run buildall

PR #31746, opened 2024-03-04
[fix](broker-load) fix broker load statement type conversion failure

liaoxin01:
## Proposed changes

Issue Number: close #xxx

when fe restart，replaying broker load job may report an error.

LOAD_JOB=32991219, transaction_id={0}, error_msg={Failed to execute load with error: org.apache.doris.analysis.UnifiedLoadStmt cannot be cast to org.apache.doris.analysis.LoadStmt

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #36168, opened 2024-06-12
[fix](broker-load) fix broker load statement type conversion failure #31746

liaoxin01:
cherry pick from #31746


liaoxin01:
run buildall

PR #36177, opened 2024-06-12
[fix](jdbc catalog) fix jdbc mysql client match jsonb type

zy-kkk:
When using mysql catalog to connect to Doris, some lower versions of Doris's Json type will be displayed as JsonB, and we need to be compatible with it


xiaokang (approved):
LGTM

morningman:
run buildall

zy-kkk:
run buildall

morningman (approved):

PR #36179, opened 2024-06-12
[branch-2.0][fix](jdbc catalog) fix jdbc mysql client match jsonb type

zy-kkk:
When using mysql catalog to connect to Doris, some lower versions of Doris's Json type will be displayed as JsonB, and we need to be compatible with it

bp #36177


zy-kkk:
run buildall

xiaokang (approved):
LGTM

PR #36184, opened 2024-06-12
[fix](planner)remove constant expr in window function's partition and order exprs

starocean999:
the legacy planner reports error if there is constant expr in window function's partition or order by keys. This pr just remove these unnecessary constant exprs and make planner happy.
now:
`SELECT row_number() OVER (partition by 1 order by 2) from t`
will be changed to 
`SELECT row_number() OVER () from t  `

## Proposed changes

Issue Number: close #xxx


starocean999:
run buildall

morrySnow (approved):

englefly (approved):

PR #36190, opened 2024-06-12
[fix](inverted index)Delete tmp dirs when BE starts to avoid tmp files left by last crash

qidaye:
When BE crashes, there may be tmp files left in the tmp dir, so we remove and rebuild the tmp dir every time we start BE to prevent rubbish data from occupying the disk.

pick from master #35951


qidaye:
run buildall

xiaokang (approved):
LGTM

PR #36198, opened 2024-06-12
[fix](regression) Fix test_disable_move_memtable

mymeiyi:
## Proposed changes

when `set enable_fold_constant_by_be = false`, the real error is `too many filtered row`:
```
Reason: column(k10) value is incorrect while strict mode is true, src value is 2022. src line [1 10 1000 1 1 1.000 a 2022 2022 a 1 1 hello 1];
```


mymeiyi:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

zhannngchen (approved):
LGTM

mymeiyi:
run p0

PR #36200, opened 2024-06-12
[improvement](inverted index)Avoid parsing ignore_above in add_values and return error status when rowid_conversion is failed. (#32890)

qidaye:
… and return error status when rowid_conversion is failed. (#32890)

pick from master #32890


qidaye:
run buildall

PR #36208, opened 2024-06-12
[Fix](inverted index) fix memory leak of inverted index writer for array values

airborne12:
## Proposed changes

Issue Number: close #xxx

fix inverted index writer's field leak

### DIFF
--- a/be/src/olap/rowset/segment_v2/inverted_index_writer.cpp
+++ b/be/src/olap/rowset/segment_v2/inverted_index_writer.cpp
@@ -395,9 +395,9 @@ public:
                     } else {
                         // now we temp create field . later make a pool
                         if (Status st = create_field(&new_field); st != Status::OK()) {
-                            LOG(ERROR)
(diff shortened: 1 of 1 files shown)


be/src/olap/rowset/segment_v2/inverted_index_writer.cpp
```diff
@@ -385,13 +385,6 @@ class InvertedIndexColumnWriterImpl : public InvertedIndexColumnWriter {
                     if (null_map[j] == 1) {
                         continue;
                     }
-                    // now we temp create field . later make a pool
-                    if (Status st = create_field(&new_field); st != Status::OK()) {
```
amorynan, line 389: here to create field would make field memory leak?

airborne12:
run buildall

amorynan (approved):
LGTM

xiaokang (approved):
LGTM

PR #26721, opened 2023-11-09
[Revert](merge-on-write) Don't use delete bitmap to mark delete for rows with delete sign when sequence column doesn't exist 

bobhan1:
## Proposed changes

This PR reverts the optimization for delete sign in https://github.com/apache/doris/pull/24011

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #36210, opened 2024-06-12
[fix](partial update) partial update should not read old fileds from rows with delete sign

### LINKED ISSUE #34296: [Bug] insert after delete where like ‘%%’ on paritial update mode will get no data
### Search before asking

### Version

master doris-4.0.0-preview-6b9f132

### What's Wrong?

 insert after delete where like ‘%%’ on paritial update mode will get no data

### What You Expected?

get insert data after delete where like ‘%%’ 

### How to Reproduce?

           
            create table test2 (
                k varchar(200) null,
                v int null
            ) unique key (k) distributed by hash(k) buckets 1
            properties("replication_num" = "1",
            "enable_unique_key_merge_on_write"="true",
            "disable_auto_compaction"="true");

            set enable_unique_key_partial_update=true;
            set enable_insert_strict = false;
            insert into test2(k,v) values('a1',100);
            select * from test2 order by k;
            delete from test2 where k like '%a%';
            select * from test2 order by k;
            insert into test2(k,v) values('a1',100);
            select * from test2 order by k;

get no result from last query

### Anything Else?

_No response_

### Code of Conduct

zhannngchen:
## Proposed changes

Issue Number: close #34296

1. When partial update filling in the missing fields, if a load job previously wrote data with a delete sign, it will also read out the data in the column with the delete sign, so that the newly written data will also become invisible
2. This problem was fixed in #24877, but was introduced again in #26721, and was never found because the case was changed to the wrong output in #26721.
3. The fix in #24877 didn't take into account the handling of concurrent conflicts in the publish phase, the current PR adds this part of the handling, and adds the corresponding case.


zhannngchen:
run buildall

zhannngchen:
run buildall

Yukang-Lian (approved):
LGTM

dataroaring (dismissed):
LGTM

zhannngchen:
run buildall

dataroaring (approved):
LGTM

PR #36211, opened 2024-06-12
[fix](Azure) Use correct http range assignation to Azure Block Blob Client

ByteYue:
## Proposed changes

The DownloadToBlobOptions' range field is one `Nullable` type which should be directly use operator= to assign one value.
Also this pr tries to print the e.what message for better debug.

### DIFF
--- a/be/src/io/fs/azure_obj_storage_client.cpp
+++ b/be/src/io/fs/azure_obj_storage_client.cpp
@@ -26,6 +26,7 @@
 #include <azure/storage/blobs/rest_client.hpp>
 #include <azure/storage/common/storage_credential.hpp>
 #include <azure/storage/common/storage_exception.hpp>
+#include <exception>
 #include <iterator>
 
 #include "common/logging.h"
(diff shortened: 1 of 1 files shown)


ByteYue:
run buildall

be/src/io/fs/azure_obj_storage_client.cpp
```diff
@@ -84,9 +85,9 @@ ObjectStorageUploadResponse AzureObjStorageClient::upload_part(const ObjectStora
         Azure::Core::IO::MemoryBodyStream memory_body(
                 reinterpret_cast<const uint8_t*>(stream.data()), stream.size());
         client.StageBlock(std::to_string(part_num), memory_body);
-    } catch (Azure::Storage::StorageException& e) {
```
gavinchou, line 87: Also catch std::exception

gavinchou (changes requested):
Also catch std::exception

ByteYue:
run buildall

gavinchou (approved):

dataroaring (approved):
LGTM

PR #36217, opened 2024-06-13
[fix](statistics)Fix sync analyze job timeout block bug.

Jibing-Li:
Fix sync analyze job timeout block bug. When a task of a analyze job timeout, it should throw an exception instead of finish silently.


Jibing-Li:
run buildall

PR #36226, opened 2024-06-13
[Fix](sql function) memory overflow to the left of string address when do_money_format has small negative value

biohazard4321:
see the unit test.
in do_money_format, some case like -123.45，no comma would set in the first loop iteration，but j -= 4 contains comma position.
so next loop iteration would cause buffer overflow. in some condition may corrupt some critical memory then coredump.

## Proposed changes

Issue Number: close #xxx

### DIFF
--- a/be/src/vec/functions/function_string.h
+++ b/be/src/vec/functions/function_string.h
@@ -2585,7 +2585,7 @@ static StringRef do_money_format(FunctionContext* context, const string& value) 
     if (!is_positive) {
         *result_data = '-';
     }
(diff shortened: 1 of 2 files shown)


biohazard4321:
run buildall

BiteTheDDDDt (approved):

biohazard4321:
run p0

zhiqiang-hhhh (approved):

biohazard4321:
run arm

biohazard4321:
run cloud_p0

HappenLee (approved):
LGTM

PR #36243, opened 2024-06-13
[fix](statistics)Skip unknown stats while show cached partition stats.

Jibing-Li:
Don't show anything if cached partition stats for a column doesn't exists. Before it will show default value which is misleading.


morrySnow (approved):

Jibing-Li:
run buildall

morningman (approved):
LGTM

PR #36067, opened 2024-06-07
[fix](script) Set DYLD_LIBRARY_PATH for macOS in be startup script

zy-kkk:
### Description
- Added logic to set `DYLD_LIBRARY_PATH` for macOS in the `setup_java_env` function of the startup script.
- This change ensures that the necessary Java dynamic libraries are correctly loaded on macOS.

### Changes
- Modified `setup_java_env` function to include `DYLD_LIBRARY_PATH` setup for macOS.

PR #36257, opened 2024-06-13
[fix](script) Set DYLD_LIBRARY_PATH for macOS in be nohup startup script

zy-kkk:
Following #36067, I found that some MacOS will block the passing of DYLD_LIBRARY_PATH in the case of nohup, so we need to actively set env during nohup


zy-kkk:
run buildall

xinyiZzz (approved):
LGTM

924060929 (approved):

PR #36264, opened 2024-06-13
[test]Add query queue check

wangbo:
## Proposed changes
This suite is used for when all p0 test finish, then call this suite to check whether workload group's query num is a valid value.


wangbo:
run buildall

wangbo:
run buildall

PR #36267, opened 2024-06-13
[fix](group commit) Fix group commit in nereids

mymeiyi:
## Proposed changes

1. The group commit in nereids use `isGroupCommit` variable in `ConnectContext` to mark if it is a group commit, but not reset. If executes 2 sqls, the first is group_commit, the second is not group_commit, the `isGroupCommit` is true for the second sql and the data_sink is `GroupCommitBlockSink`
2. Move the `isGroupCommitAvailablePlan` to the end when check `canGroupCommit` because it's cpu expensive


mymeiyi:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #36277, opened 2024-06-13
[Optimization](Variant) Using correct column name to go fast execute …

eldenmoon:
…path  in executing match expressions

The correct column name should be like `CAST value.name(Nullable(Variant)) TO Nullable(String)` since the predicate is `cast(value['name'] as string) match 'xxx'`

Query : SELECT value FROM test.records WHERE entity_id = "xxxx"  value['text3']  MATCH_ALL ' 就是 射手 男 的 克'  OR ( value['text83']  MATCH_ALL ' 考试 的 孩子 们 有 福利 ' ) OR (  value['text15']  MATCH_ALL '得 台湾 女星 许' )  LIMIT 0, 100

from 3s ~ 0.3s

## Proposed changes

Issue Number: close #xxx


eldenmoon:
run buildall

eldenmoon:
run buildall

zzzxl1993 (approved):
LGTM

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -2692,13 +2715,16 @@ void SegmentIterator::_calculate_pred_in_remaining_conjunct_root(
 
     auto node_type = expr->node_type();
     if (node_type == TExprNodeType::SLOT_REF) {
+        auto slot_expr = std::dynamic_pointer_cast<doris::vectorized::VSlotRef>(expr);
```
airborne12, line 2718: need if (slot_expr != nullptr) ?

airborne12 (approved):
LGTM

qidaye (approved):

be/src/olap/rowset/segment_v2/segment_iterator.cpp
eldenmoon: if (node_type == TExprNodeType::SLOT_REF)  ， it must be doris::vectorized::VSlotRef

PR #36280, opened 2024-06-13
[Featrue](default value) add pi as default value

Dragonliu2018:
## Proposed changes

Issue Number: refers https://github.com/apache/doris/issues/34228

add pi as default value, such as:
```sql
  CREATE TABLE IF NOT EXISTS t1
  (
      k TINYINT,
      v double not null DEFAULT PI
  )
  UNIQUE KEY(k)
  DISTRIBUTED BY HASH(k)
```


fe/fe-core/src/main/java/org/apache/doris/nereids/trees/plans/commands/info/DefaultValue.java
```diff
@@ -40,6 +41,8 @@ public class DefaultValue {
     public static DefaultValue BITMAP_EMPTY_DEFAULT_VALUE = new DefaultValue(ZERO);
     // default "value", "[]" means empty array
     public static DefaultValue ARRAY_EMPTY_DEFAULT_VALUE = new DefaultValue("[]");
+    // default "value", "3.14159265358979323846" means pi
+    public static DefaultValue PI_DEFAULT_VALUE = new DefaultValue("3.14159265358979323846", PI);
```
zclllyybb, line 45: could you explain why this scale is proper. do we need more precision?

Dragonliu2018: refer to <math.h>:
```c++
# define M_PI		3.14159265358979323846	/* pi */
```

Dragonliu2018: use `M_PIl` in <math.h> maybe better?
```c
# define M_PIl		3.141592653589793238462643383279502884L /* pi */
```

zclllyybb: I think in legal range, more precision is better as default value. you could try the max precision which double could represent exactly

Dragonliu2018: Get.
`M_PI` is enough, because the precision of type double is 15~16 digits. but it is not adequate for computation using long double. in this case, use `M_PIl`.

Yukang-Lian (approved):
LGTM

Yukang-Lian:
run buildall

zclllyybb (approved):
LGTM

dataroaring (approved):
LGTM

PR #36296, opened 2024-06-14
[improvement](build index)Make build index and clone mutually exclusi…

qidaye:
…ve and add timeout for index change job (#35724)

Currently the index change job and clone task can be executed at the same time. If the clone task gets stuck at this point, it will cause the index change job to get stuck as well and keep retrying. To solve this problem, we can refer to alter job and make index change job exclusive with clone task, and introduce the timeout to prevent infinite retries of build index.

Add the following checks and status in FE.
1. Check if table is stable (build index is not allowed when clone is in progress)
1.1. Tablet is HEALTHY.
1.2. Whether the tablet is included in the Tablet scheduler, if so, it means the current tablet is doing clone.
2. When creating the index change job, set the timeout at the same time.

---------
pick from master #35724


qidaye:
run buildall

qidaye:
run buildall

PR #36317, opened 2024-06-14
[Fix](Variant) fix variant schema change may cause invalid block schema and write missing blocks

eldenmoon:
…ma and write missing blocks

1. update_rowset_schema should not update _context.tablet_schema, since it's used as src schema for schema change, if it's changed may lead to wrong schema with block
2. buffering block during schema change should merge first and then flush, otherwise  may lost current block

## Proposed changes

Issue Number: close #xxx

### DIFF
--- a/be/src/olap/rowset/segment_creator.cpp
+++ b/be/src/olap/rowset/segment_creator.cpp
@@ -315,14 +315,15 @@ Status SegmentCreator::add_block(const vectorized::Block* block) {
     size_t block_row_num = block->rows();
     size_t row_avg_size_in_bytes = std::max((size_t)1, block_size_in_bytes / block_row_num);
     size_t row_offset = 0;
-
     if (_segment_flusher.need_buffering()) {
+        RETURN_IF_ERROR(_buffer_block.merge(*block));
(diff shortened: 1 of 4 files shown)


eldenmoon:
run buildall

eldenmoon:
run buildall

be/src/olap/rowset/segment_creator.cpp
```diff
@@ -315,14 +315,15 @@ Status SegmentCreator::add_block(const vectorized::Block* block) {
     size_t block_row_num = block->rows();
     size_t row_avg_size_in_bytes = std::max((size_t)1, block_size_in_bytes / block_row_num);
     size_t row_offset = 0;
-
     if (_segment_flusher.need_buffering()) {
+        RETURN_IF_ERROR(_buffer_block.merge(*block));
```
xiaokang, line 319: need more discussion

xiaokang (commented):
pls add testcase

eldenmoon:
> pls add testcase

done

eldenmoon:
run buildall

xiaokang (approved):
LGTM

csun5285 (approved):
LGTM

Issue #34118, opened 2024-04-25
[Bug] 有倒排索引后导入异常


cst4049:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

### Version

版本 2.0.3

### What's Wrong?

倒排索引的问题真不少

问题1： 字段名为中文字段时创建倒排索引成功， 搜索时异常报错
问题2: 搜索数据量小于2亿条数据时 倒排索引不比like强多少。 数据量超过2亿条数据后， 倒排索引直接超时， like耗时几分钟依然能返回数据 （数据内包含中文，uuid等 ）。 比elastic的差太多了

问题3: 倒排索引创建成功后， routine load导入一段时间后报错！ routine load 无法启动后不消费！

W0425 02:39:40.453413 84852 status.h:393] meet error status: [INTERNAL_ERROR][INTERNAL_ERROR]close wait failed coz rpc error

        0#  doris::stream_load::VNodeChannel::close_wait(doris::RuntimeState*) at /var/local/ldb-toolchain/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/basic_string.h:187
        1#  std::_Function_handler<void (std::shared_ptr<doris::stream_load::VNodeChannel> const&), doris::stream_load::VOlapTableSink::close(doris::RuntimeState*, doris::Status)::$_0>::_M_invoke(std::_Any_data const&, std::shared_ptr<doris::stream_load::VNodeChannel> const&) at /root/src/doris-2.0/be/src/common/status.h:442
        2#  doris::stream_load::VOlapTableSink::close(doris::RuntimeState*, doris::Status) at /root/src/doris-2.0/be/src/vec/sink/vtablet_sink.h:400
        3#  doris::PlanFragmentExecutor::open_vectorized_internal() at /root/src/doris-2.0/be/src/common/status.h:354
        4#  doris::PlanFragmentExecutor::open() at /root/src/doris-2.0/be/src/common/status.h:354
        5#  doris::FragmentExecState::execute() at /var/local/ldb-toolchain/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/tuple:180
        6#  doris::FragmentMgr::_exec_actual(std::shared_ptr<doris::FragmentExecState>, std::function<void (doris::RuntimeState*, doris::Status*)> const&) at /root/src/doris-2.0/be/src/common/status.h:442
[...]

PR #36321, opened 2024-06-14
[fix](inverted index)Support Chinese column name with inverted index

qidaye:
## Proposed changes

1. `std::string` to `std::wstring` conversion only supports ASCII characters. For non-ASCII characters, we need to use `StringUtil::string_to_wstring`
2. Fix index_tool check_terms_stats_v2 and add field info to print

Issue Number: #34118


qidaye:
run buildall

zzzxl1993 (approved):
LGTM

airborne12 (approved):
LGTM

be/src/olap/rowset/segment_v2/inverted_index_reader.cpp
```diff
@@ -353,7 +353,7 @@ Status FullTextIndexReader::query(OlapReaderStatistics* stats, RuntimeState* run
         }
 
         std::unique_ptr<lucene::search::Query> query;
-        query_info.field_name = std::wstring(column_name.begin(), column_name.end());
```
xiaokang, line 356: grep std::wstring in the code and fix all

qidaye: done

qidaye:
run buildall

xiaokang (approved):
LGTM

PR #36352, opened 2024-06-16
[feat](agg) support second key

hqx871:
## Proposed changes

Issue Number: close #xxx

For uniq and agg table, sometimes user will compute md5 value of some key columns and put md5 as key while put those key as value. For example
- when key column is too many, using md5 as key will reduce readed key columns as agg/uniq table will need to read all the key columns even though user do not select those key columns
- user may need to use array as key, but currently doris does not support

However, predicates of value column can not be well pushed down. To improve those cases, I add a table property 'second_key' to save those key columns and make ScanNode to push predicates of those key columns to rowset.


hqx871:
run buildall

dataroaring (changes requested):
It complicates the code much.

PR #36363, opened 2024-06-17
[fix](statistics)Fix stats analyze p0 case. (#36251)

Jibing-Li:
Fix stats p0 case. Def a variable before using it.
backport https://github.com/apache/doris/pull/36251


Jibing-Li:
run buildall

PR #36365, opened 2024-06-17
[fix](nereids)disable NullSafeEqualToEqual rule

starocean999:
## Proposed changes

NullSafeEqualToEqual depends on join conjunct's nullable info. But the nullable value may change after this rule. So convert from <=> to = may be wrong. We disable this rule for now and fix it later


starocean999:
run buildall

morrySnow (approved):

jacktengg (approved):

PR #33894, opened 2024-04-19
[Fix](meta) Add serialize name for auto partition FE properties

zclllyybb:
## Proposed changes

Issue Number: close #xxx

Currently there are extremely occasional instances of `partitionExpr` changes in Auto Partition, which we suspect may be a cause. Keep watching after the modification.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #36366, opened 2024-06-17
[branch-2.1](meta) Add serialize name for auto partition FE properties (#33894)

zclllyybb:
## Proposed changes

Issue Number: close #xxx

Currently there are extremely occasional instances of `partitionExpr` changes in Auto Partition, which we suspect may be a cause. Keep watching after the modification.

pick https://github.com/apache/doris/pull/33894


zclllyybb:
no need to pick.

PR #36375, opened 2024-06-17
[fix](inverted index)Support Chinese column name with inverted index #36321

qidaye:
1. `std::string` to `std::wstring` conversion only supports ASCII characters. For non-ASCII characters, we need to use `StringUtil::string_to_wstring`
2. Fix index_tool check_terms_stats_v2 and add field info to print

Issue Number: #34118
pick from master #36321


qidaye:
run buildall

qidaye:
run p0

qidaye:
run buildall

PR #36390, opened 2024-06-17
[feature](statistics)Refresh partition stats cache while analyzing.

Jibing-Li:
While doing partition level analyzing, we need to update the partition stats cache for all FEs, so that the cache are updated after analyzing finished. This pr is to support this.


morrySnow (approved):

Jibing-Li:
run buildall

morningman (approved):
LGTM

PR #36402, opened 2024-06-17
[fix](hudi) disable fs.impl.cache to avoid FE OOM

morningman:
Sometime when using Hudi Catalog, there will be lots of objects like
`org.apache.hadoop.hdfs.client.impl.DfsClientConf` in FE memory,
which belongs to the Hadoop FileSystem's Cache.

Disable this Cache to avoid FE OOM by setting
`fs.xxx.impl.disable.cache` to true


AshinGau (approved):

morningman:
run buildall

wuwenchi (approved):

PR #36438, opened 2024-06-18
[draft](mtmv) Add pre check to avoid loop

seawinde:
## Proposed changes

draft mv rewrite check to avoid loop


seawinde:
already contained by other pr

PR #36449, opened 2024-06-18
[fix](regression)Disable auto analyze before running mtmv test.

Jibing-Li:
Disable auto analyze before running mtmv test. Because auto analyze result may overwrite manual analyze result.


Jibing-Li:
run buildall

morrySnow (commented):
should wait auto analyze real stop?

morrySnow (approved):

morningman (approved):

PR #36457, opened 2024-06-18
[fix](regression)Disable auto analyze before running mtmv test.

Jibing-Li:
Disable auto analyze before running mtmv test. Because auto analyze result may overwrite manual analyze result.
backport: https://github.com/apache/doris/pull/36449


Jibing-Li:
run buildall

PR #36461, opened 2024-06-18
[opt](function)Remove input validation in FunctionDateOrDateTimeToSomething.

Mryange:
## Proposed changes

For functions like FunctionDateOrDateTimeToSomething that take date-type inputs, the input data should not contain invalid values. If there are invalid values, they should be set to null.

before
```
mysql [test]>select count(to_date(dt)) from dates;
+--------------------+
| count(to_date(dt)) |
+--------------------+
|           64000000 |
+--------------------+
1 row in set (0.32 sec)
```

now
```
mysql [test]>select count(to_date(dt)) from dates;
+--------------------+
| count(to_date(dt)) |
+--------------------+
|           63985581 |
+--------------------+
1 row in set (0.09 sec)
```

### DIFF
--- a/be/src/vec/functions/date_time_transforms.h
+++ b/be/src/vec/functions/date_time_transforms.h
@@ -311,7 +311,8 @@ struct TransformerToStringTwoArgument {
     }
 };
 
-template <typename FromType, typename ToType, typename Transform>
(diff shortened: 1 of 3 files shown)


Mryange:
run buildall

Mryange:
run buildall

PR #36275, opened 2024-06-13
[Fix](segment iterator) fix shrink non-char column coredump

airborne12:
## Proposed changes

Issue Number: close #xxx

If we execute a delete predicate in an inverted index table and then query something from it, it will cause a core dump in shrink_char_type_column_suffix_zero. This occurs because the delete predicate column ID in _char_type_idx is incorrectly shifted to the result column inserted by the inverted index result.

coredump stack like:
```
COULD NOT CREATE A LOGGINGFILE 20240604-145331!F20240604 14:53:31.991016 24178 column.h:134] Cannot get_shrinked_column() column Const(UInt8)
*** Check failure stack trace: ***
F20240604 14:53:31.991436 24158 column.h:134] Cannot get_shrinked_column() column Const(UInt8)
*** Check failure stack trace: ***
F20240604 14:53:31.991436 24158 column.h:134] Cannot get_shrinked_column() column Const(UInt8)F20240604 14:53:31.991645 24160 column.h:134] Cannot get_shrinked_column() column Const(UInt8)

#0  0x00007fa0bf784387 in ?? () from /mnt/disk2/tengjianping/local/ldb_toolchain/lib/libc.so.6
#1  0x0000557dc8e617a0 in google::LogMessage::Flush() ()
#2  0x0000557dc8e65599 in google::LogMessageFatal::~LogMessageFatal() ()
#3  0x0000557dbe4ec939 in doris::vectorized::IColumn::get_shrinked_column (this=0x7fa094ac16a0) at /home/zcp/repo_center/doris_release/doris/be/src/vec/columns/column.h:134
#4  0x0000557dc236d8a2 in doris::vectorized::Block::shrink_char_type_column_suffix_zero (this=0x7f9a372a4b70, char_type_idx=...)
    at /home/zcp/repo_center/doris_release/doris/be/src/vec/core/block.cpp:1126
#5  0x0000557dbee9ea95 in doris::segment_v2::SegmentIterator::_next_batch_internal (this=<optimized out>, block=0x7f9a372a4b70)
    at /home/zcp/repo_center/doris_release/doris/be/src/olap/rowset/segment_v2/segment_iterator.cpp:2408
#6  0x0000557dbee9cc5a in doris::segment_v2::SegmentIterator::next_batch(doris::vectorized::Block*)::$_0::operator()() const (this=<optimized out>)
    at /home/zcp/repo_center/doris_release/doris/be/src/olap/rowset/segment_v2/segment_iterator.cpp:2090
[...]

PR #36468, opened 2024-06-18
[Pick 2.1](segment iterator) fix shrink non-char column coredump

airborne12:
## Proposed changes
Pick from  (#36275)


airborne12:
run buildall

PR #36292, opened 2024-06-14
[fix](inverted index) fix the issue where need_read_data does not handle non-pushdown conditions

zzzxl1993:
…dle non-pushdown conditions

## Proposed changes

1. Fixed the previous optimization of https://github.com/apache/doris/pull/35346 not handling non-pushdown conditions.

PR #36491, opened 2024-06-18
[fix](inverted index) fix the issue where need_read_data does not handle non-pushdown conditions

zzzxl1993:
pick from https://github.com/apache/doris/pull/36292


zzzxl1993:
run buildall

PR #36502, opened 2024-06-19
[BUG][COLOCATE] fix colocate balance bug that bucket index cannot be set correctly #27741

yujun777:
cherry pick from #27741


yujun777:
run buildall

PR #36507, opened 2024-06-19
[enhance](mtmv)when calculating the availability of MTMV, no longer consider refresh state

zddr:
Before the modification, if the last refresh of the materialized view failed, the entire materialized view cannot be used for transparent rewriting. In fact, many partitions may be available, so this modification was made

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/mtmv/MTMVRewriteUtil.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/mtmv/MTMVRewriteUtil.java
@@ -20,7 +20,6 @@ package org.apache.doris.mtmv;
 import org.apache.doris.catalog.MTMV;
 import org.apache.doris.catalog.Partition;
 import org.apache.doris.common.AnalysisException;
-import org.apache.doris.mtmv.MTMVRefreshEnum.MTMVRefreshState;
 import org.apache.doris.mtmv.MTMVRefreshEnum.MTMVState;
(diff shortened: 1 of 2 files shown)


zddr:
run buildall

zddr:
run buildall

fe/fe-core/src/main/java/org/apache/doris/mtmv/MTMVRewriteUtil.java
```diff
@@ -62,8 +61,7 @@ public static Collection<Partition> getMTMVCanRewritePartitions(MTMV mtmv, Conne
             return res;
         }
         // check mv is normal
-        if (!(mtmv.getStatus().getState() == MTMVState.NORMAL
-                && mtmv.getStatus().getRefreshState() == MTMVRefreshState.SUCCESS)) {
+        if (mtmv.getStatus().getState() != MTMVState.NORMAL) {
```
seawinde, line 64: MTMVState#INIT means in building? May  MTMVState#INIT should not return

seawinde: mtmv.getStatus().getState() == MTMVState.NORMAL && mtmv.getStatus().getRefreshState() !=  MTMVRefreshState.INIT maybe better

zddr:
run buildall

seawinde (approved):

morrySnow (approved):

PR #36513, opened 2024-06-19
[enhance](mtmv)reduce the behavior of triggering the mtmv state to change to schema_change

zddr:
The following behavior will change the state of the materialized view to SCHEMA-CHANGE
- column rename
- drop column
- modify column
- replace table
- table rename
- drop table


zddr:
run buildall

zddr:
run p0

zddr:
run buildall

fe/fe-core/src/main/java/org/apache/doris/analysis/AlterTableClause.java
```diff
@@ -40,4 +40,6 @@ public void setTableName(TableName tableName) {
     }
 
     public abstract boolean allowOpMTMV();
+
+    public abstract boolean needChangeMTMVState();
```
seawinde, line 44: this may be add a default impl and value such as true, when add new cluse later, maybe safe

seawinde (approved):

fe/fe-core/src/main/java/org/apache/doris/analysis/AlterTableClause.java
zddr: This is intentional to avoid ignoring this method when creating new classes, and adding some comments in other PRs

morrySnow (approved):

PR #36536, opened 2024-06-19
[Fix](Variant) fix variant schema change may cause invalid block schema and write missing blocks #36317

eldenmoon:
…ma and write missing blocks (#36317)

buffering block during schema change should merge first and then flush, otherwise may lost current block

## Proposed changes

Issue Number: close #xxx


eldenmoon:
run buildall

eldenmoon:
run buildall

PR #36547, opened 2024-06-19
[fix](mtmv)fix when related table drop partition,mv partition is sync

zddr:
for example:
if mv's partition [mv_p1] is related with t1's partition [p1,p2]

when p2 dropped,
mv_p1'isSync should be false

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/mtmv/MTMVRefreshSnapshot.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/mtmv/MTMVRefreshSnapshot.java
@@ -18,6 +18,7 @@
 package org.apache.doris.mtmv;
 
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import com.google.gson.annotations.SerializedName;
 import org.apache.commons.collections.MapUtils;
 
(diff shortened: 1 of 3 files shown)


zddr:
run buildall

fe/fe-core/src/main/java/org/apache/doris/mtmv/MTMVRefreshSnapshot.java
```diff
@@ -46,6 +47,14 @@ public boolean equalsWithRelatedPartition(String mtmvPartitionName, String relat
         return relatedPartitionSnapshot.equals(relatedPartitionCurrentSnapshot);
     }
 
+    public Set<String> getSnapshotPartitions(String mtmvPartitionName) {
+        MTMVRefreshPartitionSnapshot partitionSnapshot = partitionSnapshots.get(mtmvPartitionName);
+        if (partitionSnapshot == null) {
+            return Sets.newHashSet();
```
seawinde, line 53: Maybe use ImmutebleSet.of() is better

seawinde (approved):

morrySnow (approved):

PR #36551, opened 2024-06-19
[fix](txn_manager) Add ingested rowsets to unused rowsets when removing txn

w41ter:
Generally speaking, as long as a rowset has a version, it can be considered not to be in a pending state.  However, if the rowset was created through ingesting binlogs, it will have a version but should still be considered in a pending state because the ingesting txn has not yet been committed.

This PR updates the condition for determining the pending state. If a rowset is COMMITTED, the txn should be allowed to roll back even if a version exists.


w41ter:
run buildall

w41ter:
run buildall

platoneko (approved):

dataroaring (approved):
LGTM

PR #36561, opened 2024-06-19
[fix](statistics)Skip large partition while calculate healthy value.

Jibing-Li:
1. Skip large partition while calculate column healthy value, because large partition may be skipped while analyzing. This could avoid keep adding the column to analyze job queue.
2. Move huge_partition_lower_bound_rows to GlobalSession
3. Fix sample job doesn't record correct start time bug.
4. Manual full analyze for partition table doesn't need to skip large partition and fallback to sample collection.


Jibing-Li:
run buildall

morrySnow (approved):

morningman (approved):
LGTM

PR #36611, opened 2024-06-20
[regression-test](case) do statistics table compaction before quit regression tests

shuke987:
## Proposed changes
In regression, we will check whether there are pipeline tasks after all test cases were finished. 
But in regression-tests, creating and dropping tables are frequency operations, which will cause multiple delete predicater in statistics table rowsets, which will cause statistics operations running slowly.
This pr will try to do a full compaction in statitics in nonConccurent mode, try to solve the problem. 

Issue Number: close #xxx


shuke987:
run buildall

hello-stephen (approved):
LGTM

jacktengg (approved):

PR #30145, opened 2024-01-19
[Enhancement](inverted index) support inverted index format v2

airborne12:
## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #36625, opened 2024-06-20
[fix](index compaction)Change index_id from int32 to int64 to avoid overflow

qidaye:
master is fix by #30145


qidaye:
run buildall

xiaokang (approved):
LGTM

PR #36618, opened 2024-06-20
[Fix](inverted index) fix wrong opt for pk no need read data

airborne12:
## Proposed changes

The optimization to avoid reading data for primary keys (PK) in segment_iterator can lead to incorrect results when there are function expressions involving the same PK column in the predicate.

PR #36633, opened 2024-06-20
[Pick 2.0](inverted index) fix wrong opt for pk no need read data

airborne12:
## Proposed changes

Pick from #36618


airborne12:
run buildall

PR #36635, opened 2024-06-20
[fix](map) fix core dump when upgrading from 1.2.x to 2.0.x with map datatype column 

amorynan:
## Proposed changes

Issue Number: close #xxx

### DIFF
--- a/be/src/olap/rowset/segment_v2/column_reader.cpp
+++ b/be/src/olap/rowset/segment_v2/column_reader.cpp
@@ -137,6 +137,7 @@ Status ColumnReader::create(const ColumnReaderOptions& opts, const ColumnMetaPB&
         }
         case FieldType::OLAP_FIELD_TYPE_MAP: {
             // map reader now has 3 sub readers for key, value, offsets(scalar), null(scala)
+            DCHECK(meta.children_columns_size() == 3 || meta.children_columns_size() == 4);
             std::unique_ptr<ColumnReader> key_reader;
             RETURN_IF_ERROR(ColumnReader::create(opts, meta.children_columns(0),
                                                  meta.children_columns(0).num_rows(), file_reader,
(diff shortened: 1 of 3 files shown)


amorynan:
run buildall

be/src/olap/rowset/segment_v2/column_reader.cpp
```diff
@@ -137,6 +137,7 @@ Status ColumnReader::create(const ColumnReaderOptions& opts, const ColumnMetaPB&
         }
         case FieldType::OLAP_FIELD_TYPE_MAP: {
             // map reader now has 3 sub readers for key, value, offsets(scalar), null(scala)
+            DCHECK(meta.children_columns_size() == 3 || meta.children_columns_size() == 4);
```
xiaokang, line 140: why not only 3?

be/src/olap/rowset/segment_v2/column_writer.cpp
```diff
@@ -262,6 +262,11 @@ Status ColumnWriter::create(const ColumnWriterOptions& opts, const TabletColumn*
         }
         case FieldType::OLAP_FIELD_TYPE_MAP: {
             DCHECK(column->get_subtype_count() == 2);
+            if (column->get_subtype_count() < 2) {
+                return Status::InvalidArgument(
+                        "If you upgraded from version 1.2.*, please drop the "
+                        "map column and wait a while to re-create the map column.");
```
xiaokang, line 268: "drop the MAP column and wait a while to re-create the map column" => "DROP the MAP columns and then ADD the MAP columns back."

be/src/olap/rowset/segment_v2/column_writer.cpp
```diff
@@ -262,6 +262,11 @@ Status ColumnWriter::create(const ColumnWriterOptions& opts, const TabletColumn*
         }
         case FieldType::OLAP_FIELD_TYPE_MAP: {
             DCHECK(column->get_subtype_count() == 2);
+            if (column->get_subtype_count() < 2) {
+                return Status::InvalidArgument(
+                        "If you upgraded from version 1.2.*, please drop the "
```
eldenmoon, line 267: return Status::InternalError to print stack, InvalidArgument will not print stack

be/src/olap/rowset/segment_v2/column_reader.cpp
amorynan: if map column is nullable, meta.children_columns(3) means to read null map

amorynan:
run buildall

PR #34923, opened 2024-05-15
[fix](planner) fix unrequired slot bug when join node introduced by #25204

nextdreamblue:
## Proposed changes

fix bug introduced by #25204

fix issue #35009

before fix, join node will retain some slots, which are not materialized and unrequired. 
join node need remove these slots and not make them be output slots.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #36646, opened 2024-06-21
(fix)(planner) fix could not resolve slot descriptor issue

Yulei-Yang:
bp #34923 and add more regression test case


Yulei-Yang:
run buildall

lide-reed (approved):
LGTM

PR #36643, opened 2024-06-21
[Fix](func) CoreDump and Result Error in percentile function

HappenLee:
## Proposed changes

Before：
```
select percentile(kbint, 0.6) from fn_test where kbint > 100;
+----------------------------------------+
| percentile(kbint, cast(0.6 as DOUBLE)) |
+----------------------------------------+
|                                   0 |
+----------------------------------------+
```

After:
```
select percentile(kbint, 0.6) from fn_test where kbint > 100;
+----------------------------------------+
| percentile(kbint, cast(0.6 as DOUBLE)) |
+----------------------------------------+
|                                   NULL |
+----------------------------------------+
```

PR #36648, opened 2024-06-21
[Fix](func) CoreDump and Result Error in percentile function

HappenLee:
## Proposed changes

cherry pick #36643


xiaokang:
run buildall

PR #36651, opened 2024-06-21
[fix](regression) Fix data_model_p0/unique/test_unique_table in cloud mode

mymeiyi:
## Proposed changes

The txn insert into mow table failed, because the commit_txn for mow table is doing by fe, but the `StreamLoadContext::is_mow_table()` return false, cause the commit_txn is doing by be. The`StreamLoadContext::is_mow_table()` return false because `put_result.__isset.pipeline_params` is false.


mymeiyi:
run buildall

dataroaring (approved):
LGTM

gavinchou (approved):

PR #36655, opened 2024-06-21
[improvement](mysql catalog) disable mysql AbandonedConnectionCleanup Thread

zy-kkk:
When using mysql catalog, mysql jdbc driver will generate an `AbandonedConnectionCleanupThread` for each database connection. This is a virtual reference, which will accumulate over time as database connections are constantly created, eventually causing OOM. Therefore, in our usage scenario, we need to turn off this thread because our database connection recycling depends on the connection pool.
But please note that this switch is only for MySQL JDBC Driver versions greater than 8.0.22


zy-kkk:
run buildall

morningman (approved):
LGTM

wuwenchi (approved):

AshinGau (approved):

PR #36657, opened 2024-06-21
Fix insert into mode is not restriction for BE storage_flood_stage_usage_percent param

dh-cloud:
# Steps to reproduce:
## 1,Create Doris ENV

```
3 FE + 4 BE
FE have  one data path：/data1
BE have  one data path：/data1
```

## 2,Create table

```
create table test(
    id                bigint       not null,
    name               varchar(100)
)
duplicate key(id)
distributed by hash(id) buckets 1
properties (
   "replication_num" = "3"
);
```

## 3,modify all BE param & restart doris

```
BE.storage_flood_stage_usage_percent=1
-- set to 100T
BE.storage_flood_stage_left_capacity_bytes=1099511627776
```
## 4,insert data

```
insert into test  values (1,'test_name');
```

## 5,result
```
Query OK, 1 row affected (0.09 sec)
{'label':'label_1bfb0be2ee8b424f_8f9a3a25ca638bc1', 'status':'VISIBLE', 'txnId':'4419'}
```

## 6,after fix:

```
ERROR 1105 (HY000) at line 1: errCode = 2, detailMessage = (be0)[E-232]disk 10878111130781813261 exceed capacity limit.

```
[ci skip]

## Proposed changes

Issue Number: close #xxx


gavinchou:
run buildall

be/src/runtime/load_stream.cpp
```diff
@@ -178,6 +179,19 @@ Status TabletStream::add_segment(const PStreamHeader& header, butil::IOBuf* data
...
+    DataDir* data_dir = nullptr;
+
+    TabletSharedPtr tablet = StorageEngine::instance()->tablet_manager()->get_tablet(header.tablet_id());
+    if (tablet != nullptr) {
+        data_dir = tablet->data_dir();				 
+    }
+
+    if (data_dir != nullptr &&
+        data_dir->reach_capacity_limit(stat.data_size + stat.index_size)) {
+        LOG(INFO) << "segment size is " << (stat.data_size + stat.index_size);
```
cambyzju, line 191: maybe debug log is better.
After disk usage is high, lots of these logs will take large disk space.

dh-cloud: already change to DEBUG

PR #36659, opened 2024-06-21
[improvement](sqlserver catalog) Configurable whether to use encrypt when connecting to SQL Server using the catalog

zy-kkk:
In previous versions, we used druid as the default JDBC connection pool, which can use custom decryption to parse the certificate when SQL Server encryption is turned on. However, in the new version, after changing HikariCP as the default connection pool, the SQLServer certificate cannot be parsed, so encryption needs to be turned off for normal use. Therefore, a parameter is added to decide whether to disable SQLServer encryption. It is not disabled by default.


zy-kkk:
run buildall

morningman (approved):
LGTM

dataroaring (approved):
LGTM

zy-kkk:
run external

zy-kkk:
run buildall

PR #36666, opened 2024-06-21
[fix](exception) throw std::runtime_error leads to BE coredump

jacktengg:
## Proposed changes

Issue Number: close #xxx

`std::runtime_error` is not caught which leads to be coredump.


jacktengg:
run buildall

zhiqiang-hhhh (approved):

Gabriel39 (approved):

PR #36678, opened 2024-06-21
[Fix]check if fe set current connect fe

wangbo:
## Proposed changes
check if fe set current connect fe, when do gray upgrade, fe may not set this value.


wangbo:
run buildall

wangbo:
run buildall

PR #36637, opened 2024-06-20
[Fix](inverted index) fix wrong no need read data when need_remaining_after_evaluate

airborne12:
## Proposed changes

When using an equal predicate on a column that applies an inverted index with a parser, it requires remaining_after_evaluate. In this situation, we cannot optimize the column without reading the data.

PR #36684, opened 2024-06-21
[Pick 2.1](inverted index) fix wrong no need read data when need_remaining_after_evaluate

airborne12:
When using an equal predicate on a column that applies an inverted index with a parser, it requires remaining_after_evaluate. In this situation, we cannot optimize the column without reading the data.

## Proposed changes

From (#36637)


airborne12:
run buildall

PR #36687, opened 2024-06-21
[Fix](inverted index) fix wrong no need read data when same column in inverted index and like function

airborne12:
## Proposed changes

SQL like SELECT COUNT() FROM table WHERE a >= '2024-06-15 00:00:00' AND b LIKE '%xxx%' AND b MATCH 'xxx' for inverted index tables may cause incorrect optimization in the segment iterator, leading to the no-read-data optimization being applied incorrectly.


airborne12:
run buildall

eldenmoon (approved):
lgtm

zzzxl1993 (approved):
LGTM

PR #35990, opened 2024-06-06
[feature-wip](Cloud) Support Azure Cloud Blob in FE

ByteYue:
## Proposed changes

As one subsequent pr of #35307, this pr tries to support Azure Cloud in FE. After this pr, we should be able to use Azure Blob for Cold Heat Separation and Storage Compute Separation along with backup restore to Azure Cloud.

For the data lakes support, you can add corresponding sdks to FE and implements the corresponding features.

PR #36695, opened 2024-06-22
[chore](Regression) Add provider conf in regression's pipeline conf

ByteYue:
## Proposed changes

Previously the conf of pipeline doesn't contain the provider of the Object Storage it uses. When adding azure's support like #35990, user should specify the vendor they use or just doesn't set the "provider" property when using S3-like object storage. So i add this conf.


ByteYue:
run buildall

gavinchou (approved):

Yukang-Lian (approved):

PR #36704, opened 2024-06-22
[bug](prepared statement) fix double type value incorrect in legacy planner

xy720:
## Proposed changes

Issue Number: close #xxx

fix double type value incorrect.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/LiteralExpr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/LiteralExpr.java
@@ -358,6 +358,7 @@ public abstract class LiteralExpr extends Expr implements Comparable<LiteralExpr
                 break;
             case 5: // MYSQL_TYPE_DOUBLE
                 literalExpr = LiteralExpr.create("0", Type.DOUBLE);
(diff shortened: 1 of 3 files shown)


xy720:
run buildall

fe/fe-core/src/main/java/org/apache/doris/analysis/LiteralExpr.java
```diff
@@ -358,6 +358,7 @@ public static LiteralExpr getLiteralByMysqlType(int mysqlType, boolean isUnsigne
                 break;
             case 5: // MYSQL_TYPE_DOUBLE
                 literalExpr = LiteralExpr.create("0", Type.DOUBLE);
+                literalExpr.setType(Type.DOUBLE);
```
zclllyybb, line 361: if double should set it type, should we just change `DOUBLE` branch in `LiteralExpr::create`?

xy720: LiteralExpr::create is used in many places, it's best not to change it.

lide-reed (approved):
LGTM

cambyzju (approved):

PR #36722, opened 2024-06-24
[Optimize](Cooldown) Remove the time limit for table creation and specified time cooling strategy

qzsee:
## Proposed changes

Issue Number: close #xxx

Because some users migrated to doris from other products, they do not need to query some historical data, but it needs to be stored in doris for future query. At this time, they need to cool down directly without specifying a future time.


qzsee:
run buildall

PR #36745, opened 2024-06-24
[opt](mysql serde) Avoid core dump when converting invalid block to mysql result #28069

zhiqiang-hhhh:
cherry pick from #28069


zhiqiang-hhhh:
run buildall

PR #36753, opened 2024-06-24
[fix](move-memtable) check segment num when closing each tablet

kaijchen:
## Proposed changes

Previously, there is chance that sender failed to send some data while the receiver being unaware of.
This will cause lost data if some segments are skipped.

This PR fixes the problem by including checks in both sender and receiver.
When sender failed to send rpc, LoadStreamStub will mark the involved tablets failed.
Each sender will send segment num for each tablet in CLOSE_LOAD,
and receivers (LoadStream) will sum up and check total segment nums.


kaijchen:
run buildall

kaijchen:
run buildall

kaijchen:
run buildall

kaijchen:
run buildall

be/src/runtime/load_stream.cpp
```diff
@@ -307,11 +312,17 @@ Status IndexStream::close(const std::vector<PTabletID>& tablets_to_commit,
...
     // open all need commit tablets
     for (const auto& tablet : tablets_to_commit) {
+        if (_id != tablet.index_id()) {
+            continue;
+        }
         TabletStreamSharedPtr tablet_stream;
         auto it = _tablet_streams_map.find(tablet.tablet_id());
-        if (it == _tablet_streams_map.end() && _id == tablet.index_id()) {
+        if (it == _tablet_streams_map.end()) {
             RETURN_IF_ERROR(
```
liaoxin01, line 321: It should be added to the list of failed tablets instead of directly returning a failure.

liaoxin01:
Please add an injection regression case.

be/src/runtime/load_stream.cpp
kaijchen: This PR doesn't change the original logic.
Let's fix this in a later PR.

kaijchen:
run buildall

kaijchen:
run buildall

kaijchen:
run buildall

kaijchen:
run buildall

kaijchen:
run buildall

liaoxin01 (approved):
LGTM

dataroaring (approved):
LGTM

PR #36770, opened 2024-06-24
[fix](mtmv) Fix union rewrite by materialized view regression test instability

seawinde:
## Proposed changes

When union rewrite by materialized view, the final plan chosen by CBO is instability.
So the regression test only check mv is rewritten successful or not, doesn't check is chosen by CBO or not.
Optimize to make sure chosen by CBO would be anther pr to fix this thoroughly。


seawinde:
run buildall

morrySnow (approved):

zfr9527 (approved):
LGTM

PR #36772, opened 2024-06-24
[Fix](stream-load) Fix stream load stuck under high concurrency

liaoxin01:
## Proposed changes

When the concurrency of streamload exceeds the number of threads in the remote scanner, streamload may get stuck. The reason is that the libevent thread blocks and waits for streamload to complete, and when there is no intersection between the tasks handled by the scanner thread and the libevent thread, it gets stuck.
The solution is to convert the synchronous waiting tasks of libevent into asynchronous execution by using callbacks in the streamload executor thread.

### DIFF
--- a/be/src/http/action/stream_load.cpp
+++ b/be/src/http/action/stream_load.cpp
@@ -33,6 +33,7 @@
 #include <time.h>
 
 #include <algorithm>
+#include <functional>
 #include <future>
 #include <map>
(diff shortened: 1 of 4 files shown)


liaoxin01:
run buildall

gavinchou (approved):
LGTM

dataroaring (approved):
LGTM

PR #36776, opened 2024-06-24
[fix](array)fix array_except/union for left const return only one row result

amorynan:
## Proposed changes
if left param in array_except/union is const, it will make ret column only has one row column which will make result wrong

Issue Number: close #xxx

### DIFF
--- a/be/src/vec/functions/array/function_array_set.h
+++ b/be/src/vec/functions/array/function_array_set.h
@@ -181,7 +181,11 @@ private:
         constexpr auto execute_left_column_first = Impl::Action::execute_left_column_first;
         size_t current = 0;
         Impl impl;
-        for (size_t row = 0; row < left_data.offsets_ptr->size(); ++row) {
+        size_t row_size = left_data.offsets_ptr->size();
+        if constexpr (LCONST) {
+            row_size = right_data.offsets_ptr->size();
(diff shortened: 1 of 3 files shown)


amorynan:
run buildall

be/src/vec/functions/array/function_array_set.h
```diff
@@ -181,7 +181,11 @@ struct ArraySetImpl {
         constexpr auto execute_left_column_first = Impl::Action::execute_left_column_first;
         size_t current = 0;
         Impl impl;
-        for (size_t row = 0; row < left_data.offsets_ptr->size(); ++row) {
+        size_t row_size = left_data.offsets_ptr->size();
```
zhiqiang-hhhh, line 184: what if right const

amorynan: right and left is all not const ==>  left_data.offsets_ptr->size()
right const && left is not const ==> left_data.offsets_ptr->size()
right is not const && left is const ==> right.offsets_ptr->size()
right and left is all const ==>  left_data.offsets_ptr->size()

zhiqiang-hhhh (approved):

eldenmoon (approved):
lgtm

PR #36798, opened 2024-06-25
[Revert](stream-load)Revert fix stream load stuck under high concurrency #36772

liaoxin01:
Reverts apache/doris#36772
It may cause some regression cases to fail. The general reason is that there were some abnormal when load load, this modification will return the HTTP response to the client in advance, but the client data has not been sent yet, which may cause some problems at this time.


liaoxin01:
run buildall

hello-stephen (approved):

dataroaring (approved):
LGTM

PR #36805, opened 2024-06-25
[feat](job)Internal job cancellation immediately and the strong association with the STARTS parameter

CalvinKirs:
## Proposed changes

For internal tasks, such as MTMV, the start time may already be set, or the time may be adjusted immediately.


CalvinKirs:
run buildall

fe/fe-core/src/main/java/org/apache/doris/analysis/CreateJobStmt.java
```diff
@@ -149,6 +150,8 @@ public void analyze(Analyzer analyzer) throws UserException {
         if (null != startsTimeStamp) {
             if (startsTimeStamp.equalsIgnoreCase(CURRENT_TIMESTAMP_STRING)) {
                 jobExecutionConfiguration.setImmediate(true);
+                //To avoid immediate re-scheduling, set the start time of the timer 100ms before the current time.
+                timerDefinition.setStartTimeMs(System.currentTimeMillis() - 100L);
```
zddr, line 154: what will happen if not set startTimemMs?

fe/fe-core/src/main/java/org/apache/doris/job/scheduler/JobScheduler.java
```diff
@@ -124,6 +124,10 @@ public void scheduleOneJob(T job) throws JobException {
                 schedulerInstantJob(job, TaskType.SCHEDULED, null);
             }
         }
+        if (job.getJobConfig().isImmediate() && JobExecuteType.ONE_TIME.equals(job.getJobConfig().getExecuteType())) {
```
zddr, line 127: why JobExecuteType == ONE_TIME ,can not setLatestSchedulerTimeMs(System.currentTimeMillis())?

zddr (commented):
add case?

fe/fe-core/src/main/java/org/apache/doris/analysis/CreateJobStmt.java
CalvinKirs: The syntax for Create Job doesn't immediately schedule the job. For a one-time schedule, setting the start time to CURRENT_TIMESTAMP achieves immediate scheduling. Therefore, we still need to set the startTime parameter; otherwise, the job query will lack a start time.

fe/fe-core/src/main/java/org/apache/doris/job/scheduler/JobScheduler.java
CalvinKirs: This variable is not persistent, so for a one-time, immediately scheduled task, there's no need to set it.

zddr (approved):

CalvinKirs:
run buildall

CalvinKirs:
run buildall

CalvinKirs:
run buildall

CalvinKirs:
run external

JNSimba (approved):
LGTM

PR #36825, opened 2024-06-25
[Fix](inverted index) fix wrong segment file path when doing segcompaction

airborne12:
## Proposed changes

We can't using normal segment path like rowsetid_segid.dat when create segcompaction segment writer.


airborne12:
run buildall

qidaye (approved):
LGTM

platoneko (approved):

PR #36865, opened 2024-06-26
[opt](mow) reduce memory usage for mow table compaction

cambyzju:
**The problem:**
Huge memory usage while compact mow table, following is a example, compaction peak memory 45.19 GB used while `input_row_num=701398908, output_row_num=175349727, filtered_row_num=526049181`

```
succeed to do base compaction is_vertical=1. tablet=785333.186591797.61463385baeb9d59-9167a604aac7738e, output_version=[0-42], current_max_version=42, disk=/data/cdw/doris/be/storage, segments=73, input_rowset_size=16124232924, output_rowset_size=3981194537, input_row_num=701398908, output_row_num=175349727, filtered_row_num=526049181, merged_row_num=0. elapsed time=402.428s. cumulative_compaction_policy=size_based, compact_row_per_second=1742918
```

```
MemTrackerLimiter Label=BaseCompaction:785333, Type=compaction, Limit=-1.00 B(-1 B), Used=45.19 GB(48518743241 B), Peak=45.19 GB(48519793889 B)
```

**How to Fix:**
The reason is `missed_rows` and `location_map` are very expensive. So we opt it:
1. Only save `missed_rows` set while we really need;
2. add a config `enable_missing_rows_correctness_check` to control whether we need to collect `missed_rows`
3. only when `enable_rowid_conversion_correctness_check` opened, we should collect `location_map`;

**After fix**
after fix, the memory could reduce from 45.19 GB to 9.89 GB

### DIFF
--- a/be/src/common/config.cpp
+++ b/be/src/common/config.cpp
@@ -1121,7 +1121,7 @@ DEFINE_mBool(enable_mow_compaction_correctness_check_core, "false");
 // rowid conversion correctness check when compaction for mow table
 DEFINE_mBool(enable_rowid_conversion_correctness_check, "false");
 // missing rows correctness check when compaction for mow table
-DEFINE_mBool(enable_missing_rows_correctness_check, "true");
+DEFINE_mBool(enable_missing_rows_correctness_check, "false");
 // When the number of missing versions is more than this value, do not directly
 // retry the publish and handle it through async publish.
 DEFINE_mInt32(mow_publish_max_discontinuous_version_num, "20");
--- a/be/src/common/config.cpp
+++ b/be/src/common/config.cpp
@@ -1120,6 +1120,8 @@ DEFINE_mBool(enable_merge_on_write_correctness_check, "true");
 DEFINE_mBool(enable_mow_compaction_correctness_check_core, "false");
 // rowid conversion correctness check when compaction for mow table
 DEFINE_mBool(enable_rowid_conversion_correctness_check, "false");
+// missing rows correctness check when compaction for mow table
+DEFINE_mBool(enable_missing_rows_correctness_check, "true");
 // When the number of missing versions is more than this value, do not directly
 // retry the publish and handle it through async publish.
 DEFINE_mInt32(mow_publish_max_discontinuous_version_num, "20");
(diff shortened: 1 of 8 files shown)


cambyzju:
run buildall

be/src/olap/compaction.cpp
```diff
@@ -881,22 +881,30 @@ Status CompactionMixin::modify_rowsets() {
         _tablet->tablet_schema()->cluster_key_idxes().empty()) {
         Version version = tablet()->max_version();
         DeleteBitmap output_rowset_delete_bitmap(_tablet->tablet_id());
-        std::set<RowLocation> missed_rows;
-        std::map<RowsetSharedPtr, std::list<std::pair<RowLocation, RowLocation>>> location_map;
+        std::unique_ptr<RowLocationSet> missed_rows;
+        if ((config::enable_missing_rows_correctness_check ||
```
zhannngchen, line 885:
```suggestion
        #ifndef NDEBUG
            config::enable_missing_rows_correctness_check = true
         #endif
```

zhannngchen, line 885: If the binary is debug version, we should enable the check to discover issues in our test pipeline
```
#ifndef NDEBUG
        config::enable_missing_rows_correctness_check = true
#endif
```

be/src/common/config.cpp
```diff
@@ -1118,6 +1118,8 @@ DEFINE_mBool(enable_merge_on_write_correctness_check, "true");
 DEFINE_mBool(enable_mow_compaction_correctness_check_core, "false");
 // rowid conversion correctness check when compaction for mow table
 DEFINE_mBool(enable_rowid_conversion_correctness_check, "false");
+// missing rows correctness check when compaction for mow table
+DEFINE_mBool(enable_missing_rows_correctness_check, "true");
```
zhannngchen, line 1122: Since the missing row check is costive, we can disable it by default

cambyzju: done

zhannngchen (approved):
LGTM

be/src/olap/compaction.cpp
cambyzju: added in be.conf

dataroaring (approved):
LGTM

dataroaring:
run buildall

liaoxin01 (approved):

xy720 (approved):
LGTM

zhannngchen:
run p0

PR #36869, opened 2024-06-26
[fix](inverted index)Make build index operation only affect base index

qidaye:
## Proposed changes

When building index on table with mv, we should only send index build jobs to base index.
Otherwise the build process may be stuck.
This PR make the build index operation only be allowed in base index and add a test case.


qidaye:
run buildall

qidaye:
run feut

airborne12 (approved):
LGTM

zzzxl1993 (approved):
LGTM

PR #36892, opened 2024-06-27
[fix](regression) fix txn_insert case

mymeiyi:
## Proposed changes

The delete predicate on mow table is implementated in insert, so in txn insert, it is an insert rather than delete operation, the commit order is changed, so the result is changed.


mymeiyi:
run buildall

mymeiyi:
run buildall

zhannngchen (approved):
LGTM

dataroaring (approved):
LGTM

PR #36917, opened 2024-06-27
[feature](Azure) Implement azure table value function

ByteYue:
## Proposed changes

Formerly we forbid using azure as s3 tvf's backend storage, this pr implements the corresponding logic. User should specify the provider property to azure when using azure tvf.


fe/fe-core/src/main/java/org/apache/doris/tablefunction/S3TableValuedFunction.java
```diff
@@ -113,6 +109,26 @@ public S3TableValuedFunction(Map<String, String> properties) throws AnalysisExce
         }
     }
 
+    private String constructEndpoint(Map<String, String> properties, S3URI s3uri) throws AnalysisException {
```
gavinchou, line 112: Do we have to check ak sk for all kinds of providers?

ByteYue: No need to do this.

ByteYue:
run buildall

ByteYue:
run buildall

gavinchou (approved):

morningman (approved):

PR #36926, opened 2024-06-27
[fix](group commit) Group commit http stream should not begin txn

mymeiyi:
## Proposed changes

After nereids support group_commit, the http_stream in group commit will also do some check like insert into values, and change the group_commit flag to false. 
Then will begin a txn. This is hang schema change.


gavinchou (approved):

morrySnow (approved):

mymeiyi:
run buildall

PR #36938, opened 2024-06-27
[opt](function) opt nullif function by nullif(null,x) = null

Mryange:
## Proposed changes

https://dbfiddle.uk/Ud5Ng3Ui

nullif(null,x) = null
after
```

mysql [test10]>select count(nullif(null,number)) from numbers("number" = "800000000");
+-----------------------------+
| count(nullif(NULL, number)) |
+-----------------------------+
|                           0 |
+-----------------------------+
1 row in set (2.18 sec)

```
now
```
mysql [test10]>select count(nullif(null,number)) from numbers("number" = "800000000");
+-----------------------------+
| count(nullif(NULL, number)) |
+-----------------------------+
|                           0 |
+-----------------------------+
1 row in set (0.86 sec)
```

### DIFF
--- a/be/src/vec/functions/nullif.cpp
+++ b/be/src/vec/functions/nullif.cpp
@@ -86,6 +86,12 @@ public:
     // nullIf(col1, col2) == if(col1 = col2, NULL, col1)
     Status execute_impl(FunctionContext* context, Block& block, const ColumnNumbers& arguments,
                         size_t result, size_t input_rows_count) const override {
+        if (block.get_by_position(arguments[0]).column->only_null()) {
+            // nullIf(null,x) => null
+            block.get_by_position(result).column =
+                    block.get_by_position(arguments[0]).column->clone_resized(input_rows_count);
+            return Status::OK();
+        }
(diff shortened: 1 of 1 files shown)


Mryange:
run buildall

superdiaodiao (commented):
`bool use_default_implementation_for_nulls() const override { return false; }`

What if we change the `false` to `true`? Would it be better?

Mryange:
> `bool use_default_implementation_for_nulls() const override { return false; }`
> 
> What if we change the `false` to `true`? Would it be better?

if use_default_implementation_for_nulls = true 
 nullif(3,null) = null 
but 
nullif(3,null) = 3

superdiaodiao:
> > `bool use_default_implementation_for_nulls() const override { return false; }`
> > What if we change the `false` to `true`? Would it be better?
> 
> if use_default_implementation_for_nulls = true nullif(3,null) = null but nullif(3,null) = 3

OK, get.

Mryange:
run buildall

PR #36940, opened 2024-06-27
[branch-2.0][improvement](sqlserver catalog) Can configure whether to use encryption when connecting to SQL Server using the catalog

zy-kkk:
pick #36659
pick #37015
In previous versions, we used druid as the default JDBC connection pool, which can use custom decryption to parse the certificate when SQL Server encryption is turned on. However, in the new version, after changing HikariCP as the default connection pool, the SQLServer certificate cannot be parsed, so encryption needs to be turned off for normal use. Therefore, a parameter is added to decide whether to disable SQLServer encryption. It is not disabled by default.


zy-kkk:
run buildall

morningman:
run buildall

PR #36873, opened 2024-06-26
[fix](oom) avoid oom when a lot of tablets fail on load

dataroaring:
## Proposed changes

When a lot of tablets fail when loading, then detailed information would cause oom.

Issue Number: close #xxx

PR #36944, opened 2024-06-27
[fix](oom) avoid oom when a lot of tablets fail on load

dataroaring:
pick #36873

## Proposed changes

Issue Number: close #xxx


dataroaring:
run buildall

PR #36452, opened 2024-06-18
[fix](statistics)Use ConcurrentHashMap to avoid ConcurrentModificationException

Jibing-Li:
OlapTable idToPartition Map should use ConcurrentHashMap to avoid ConcurrentModificationException.

PR #36950, opened 2024-06-27
[fix](statistics)Use ConcurrentHashMap to avoid ConcurrentModificationException (#36452)

Jibing-Li:
OlapTable idToPartition Map should use ConcurrentHashMap to avoid ConcurrentModificationException.
backport: https://github.com/apache/doris/pull/36452


Jibing-Li:
run buildall

PR #36960, opened 2024-06-27
[fix](autoinc) avoid duplicated auto inc when role of fe changes

dataroaring:
## Proposed changes

Issue Number: close #xxx

PR #36961, opened 2024-06-27
[fix](autoinc) avoid duplicated auto inc when role of fe changes

dataroaring:
## Proposed changes

pick #36960 

Issue Number: close #xxx


dataroaring:
run buildall

PR #36966, opened 2024-06-27
[BugFix] Correct cgroup mem info cache

hongkunxu:
## Proposed changes
After upgrading to Doris 2.1.3, we noticed that the "sys available memory" in be.INFO continuously decreases until it falls below the warning water mark, leading to persistent garbage collection (GC) despite the actual memory usage being very low. And The cache in cgroup mem info is always 0. Consequently, I identified an error in the calculation of available memory in cgroup memory:

1. The memory information for cgroup memory is stored in the file "memory.stat" rather than "memory.meminfo" (in fact, the "memory.meminfo" file does not exist). You can see the files under the cgroup path in the attached screenshot1.
2. The output content of "memory.stat" is shown in the screenshot1 below.

image
image

My change is about two steps:
1. Modified the file name for mem info in cgroup.
2. Modified the process for extracting the cache from cgroup.


xinyiZzz:
run buildall

xinyiZzz (approved):
LGTM 
`memory.stat` is used by all Linux OS, `memory.meminfo` is only used in CentOS.

xinyiZzz:
run buildall

xinyiZzz (approved):
LGTM

dataroaring (approved):
LGTM

xinyiZzz:
run buildall

dataroaring (approved):
LGTM

PR #36997, opened 2024-06-28
[Fix](variant) ignore serialization of nothing type

eldenmoon:
1. fix variant should not serialize nothing type
2. fix unstable cases
Issue Number: close #xxx


eldenmoon:
run buildall

eldenmoon:
run buildall

be/src/vec/data_types/data_type_object.cpp
```diff
@@ -113,6 +119,8 @@ char* DataTypeObject::serialize(const IColumn& column, char* buf, int be_exec_ve
         // 2.2 serialize subcolumn
         buf = type->serialize(entry->data.get_finalized_column(), buf, be_exec_version);
     }
+    // serialize num of subcolumns
+    *reinterpret_cast<uint32_t*>(size_pos) = num_of_columns;
```
xiaokang, line 123: variable size_pos is only used once and can be elimilated.

eldenmoon: no， it's need to be recorded, and assigned then

xiaokang (approved):

qidaye (approved):

PR #37001, opened 2024-06-28
[fix](merge-on-write) when full clone failed, duplicate key might occur

zhannngchen:
## Proposed changes

Issue Number: close #xxx

introduced by #31268

full clone failure may produce duplicate keys in mow table
the bug would be triggered in the following condition:
1. replica 0 miss version
2. replica 0 try to do full clone from other replicas
3. the full clone failed and the delete bitmap is overrided incorrectly
4. replica 0 try to do incremental clone again and this time the clone succeed
5. incremental clone can't fix the delete bitmap overrided by previous failed full clone
6. duplicate key occurred

solution:
for full clone, don't override the delete bitmap, use `merge()` method instead.


zhannngchen:
run buildall

liaoxin01 (approved):
LGTM

dataroaring (approved):
LGTM

PR #37009, opened 2024-06-28
[fix](mtmv)fix mtmv dead lock

zddr:
Cause of occurrence：
- dropBaseTable: Holding db's writeLock ,notify mtmv alter status to schema_change,need mv's writeLock
- task(insert overwrite):Holding mv's readLock,when generage plan need db's readLock

fix:
- mtmv alter status to schema_change need mv's writeMvLock instead of mv's writeLock


zddr:
run buildall

morrySnow (approved):

dataroaring (approved):
LGTM

PR #37021, opened 2024-06-28
[fix](jdbc catalog) fix jdbc table sink writer close nullptr and add log

zy-kkk:
fix jdbc table sink writer close nullptr and add log


zy-kkk:
run buildall

morningman (approved):
LGTM

PR #37031, opened 2024-06-29
[opt](binlog) support rename binlog

lsy3993:
## Proposed changes

Rename operator doesn't have binlog now. This PR will create binlog when execute rename. The rename operator means : 
1. rename table : ALTER TABLE table1 RENAME table2;
2. rename rollup : ALTER TABLE example_table RENAME ROLLUP rollup1 rollup2;
3. rename partition : ALTER TABLE example_table RENAME PARTITION p1 p2;

After SQL analyzing, we can get the old and new table name (rollup name or partition name) , then record the info to binlog, so we can use the info from binlog.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/persist/TableInfo.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/persist/TableInfo.java
@@ -42,10 +42,16 @@ public class TableInfo implements Writable {
 
     @SerializedName("nT")
     private String newTableName;
+    @SerializedName("oT")
+    private String oldTableName;
     @SerializedName("nR")
     private String newRollupName;
+    @SerializedName("oR")
+    private String oldRollupName;
     @SerializedName("nP")
     private String newPartitionName;
+    @SerializedName("oP")
+    private String oldPartitionName;
 
     public TableInfo() {
         // for persist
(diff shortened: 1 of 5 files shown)


dataroaring:
run buildall

lsy3993:
run buildall

lsy3993:
run buildall

lsy3993:
run buildall

lsy3993:
run buildall

fe/fe-core/src/main/java/org/apache/doris/persist/TableInfo.java
```diff
@@ -131,7 +170,14 @@ public void readFields(DataInput in) throws IOException {
         partitionId = in.readLong();
 
         newTableName = Text.readString(in);
+        oldTableName = Text.readString(in);
```
w41ter, line 173: The read order of fields is important. Since the default serialize/deserialize way of `TableInfo` has changed to GSON, after VERSION 134, there is no need to read and set fields in the `readFields` method.

w41ter (changes requested):

fe/fe-core/src/main/java/org/apache/doris/persist/TableInfo.java
lsy3993: done

w41ter (approved):

CalvinKirs (approved):
There are no problems with the changes made to the ENV file.

CalvinKirs:
trigger ci

CalvinKirs:
run buildall

lsy3993:
run buildall

PR #37040, opened 2024-06-30
[opt](catalog) add some profile for parquet reader and change meta cache config

morningman:
## Proposed changes

This PR mainly changes:
1. add new BE config `enable_parquet_page_index`
    Default is true, if set to false, the parquet reader will not use page index to filter data.
    This is only for debug purpose, in case sometimes the page index
    filter wrong data.

2. Add new FE config `max_hive_partition_table_cache_num`
    Separator from `max_hive_table_cache_num`. This config is used to set the max cache number
    of `partitionValuesCache`, which is for partition values list of partitioned hive table.

3. Reduce the default expire time of meta cache from 86400 to 24400.
    The hive table partitioned by day, if the expire time of cache is 86400,
    it may always fetching the stale cache value after a day pass by.

### DIFF
--- a/be/src/common/config.cpp
+++ b/be/src/common/config.cpp
@@ -1313,6 +1313,11 @@ DEFINE_Bool(enable_file_logger, "true");
 // The minimum row group size when exporting Parquet files. default 128MB
 DEFINE_Int64(min_row_group_size, "134217728");
 
+// If set to false, the parquet reader will not use page index to filter data.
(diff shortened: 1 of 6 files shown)


morningman:
run buildall

AshinGau (approved):

AshinGau:
LGTM

wuwenchi (approved):

PR #37051, opened 2024-07-01
[conf](brpc) Reduce brpc threads

Gabriel39:
## Proposed changes

Issue Number: close #xxx


Gabriel39:
run buildall

yiguolei (changes requested):
Not modify the default value since only one user have problem. Maybe it is brpc's bug.

PR #37053, opened 2024-07-01
[fix](cloud) Allow access to MS during the replay

w41ter:
In some metadata designs, the process involves writing to the edit log first and then calling RPC to delete the data. The latter might fail, so it is reasonable to continue calling RPC to delete the data during replay. PR #36856 argues that not calling RPC in the checkpoint thread is problematic.


w41ter:
run buildall

w41ter:
run buildall

dataroaring (approved):
LGTM

gavinchou (approved):

PR #37072, opened 2024-07-01
[Bug](QueryStatistics) use std::atomic<int64_t> for returned_rows to save data

zhangstar333:
## Proposed changes

when set enable_parallel_result_sink = true, it's will be plan multi result sink operator,
and if set dry_run_query = true, it's will sum all result sink operator of return rows,
so need use std::atomic<int64_t> for returned_rows to save data.


HappenLee (approved):
LGTM

BiteTheDDDDt (approved):

PR #36263, opened 2024-06-13
[fix](null safe equal join) fix coredump if both sides of the conjunct is not nullable

jacktengg:
## Proposed changes

For a null safe equal join, FE may generate a plan that both sides of the conjuct are not nullable, which is a valid plan, BE will coredump in this situation.

Issue Number: close #xxx

Coredump stacktrace:
```
#0  0x00007fff042caacf in raise () from /lib64/libc.so.6
#1  0x00007fff0429dea5 in abort () from /lib64/libc.so.6
#2  0x00005555a7ec7cad in ?? ()
#3  0x00005555a7eba34a in google::LogMessage::Fail() ()
#4  0x00005555a7ebd436 in google::LogMessage::SendToLog() ()
#5  0x00005555a7eb9e80 in google::LogMessage::Flush() ()
#6  0x00005555a7ebdc79 in google::LogMessageFatal::~LogMessageFatal() ()
#7  0x000055557824c9ac in doris::vectorized::IColumn::get_raw_data (this=0x60400387df90)
    at /mnt/disk2/tengjianping/doris-39/be/src/vec/columns/column.h:590
#8  0x00005555a4a85f32 in doris::vectorized::MethodKeysFixed<doris::JoinHashTable<unsigned long, HashCRC32<unsigned long> >, false>::pack_fixeds<unsigned long> (this=0x6110086dcb50, 
    row_numbers=21, key_columns=std::vector of length 1, capacity 1 = {...}, 
    nullmap_columns=std::vector of length 0, capacity 0, 
    result=std::vector of length 21, capacity 21 = {...})
    at /mnt/disk2/tengjianping/doris-39/be/src/vec/common/hash_table/hash_map_context.h:405
#9  0x00005555a4a83a77 in doris::vectorized::MethodKeysFixed<doris::JoinHashTable<unsigned long, HashCRC32<unsigned long> >, false>::init_serialized_keys (this=0x6110086dcb50, 
    key_columns=std::vector of length 1, capacity 1 = {...}, num_rows=21, null_map=0x0, 
    is_join=true, is_build=true, bucket_size=32)
    at /mnt/disk2/tengjianping/doris-39/be/src/vec/common/hash_table/hash_map_context.h:470
[...]

PR #37073, opened 2024-07-01
[fix](null safe equal join) fix coredump if both sides of the conjunct is not nullable #36263

jacktengg:
## Proposed changes

Bp #36263


jacktengg:
run buildall

HappenLee (approved):
LGTM

PR #37086, opened 2024-07-01
[fix](multi-catalog) Revert #36575 and check nullptr of data column

AshinGau:
## Proposed changes

Revert #36575, because `VScanner::get_block` will check `DCHECK(block->rows() == 0)`, so block should be cleared when `eof = true`.


AshinGau:
run buildall

morningman (approved):
LGTM

kaka11chen (approved):
LGTM

PR #37087, opened 2024-07-01
[fix](statistics)Fix partition record update rows bug. Remove useless code.

Jibing-Li:
Fix partition record update rows bug. 
When enable_partition_analyze set to false, we shouldn't update column's update rows map, only update it when enable_partition_analyze is true. Because when it's false, the analyze task didn't collect partition level stats. If we update column's update rows in this case, it will cause auto analyze failed to trigger partition collection because the health checker thought the update rows are update-to-date.


Jibing-Li:
run buildall

morrySnow (approved):

Jibing-Li:
run cloud_p1

morningman (approved):
LGTM

PR #37093, opened 2024-07-01
[opt](log) Remove unnecessary warning log

zhiqiang-hhhh:
https://ask.selectdb.com/questions/D1N41/fe-yi-zhi-shua-yi-chang-ri-zhi-qeprocessorimpl-reportexecstatus-248-reportexecstatus-runtime-error-query-with-type-select-does-not-exist/E1851

When enable_profile = true or report_succeed=true, it is very likely fe/fe-core/src/main/java/org/apache/doris/qe/QeProcessorImpl.java::reportExecStatus will print much warning log. It is not necesary.


zhiqiang-hhhh:
run buildall

PR #37106, opened 2024-07-01
[refactor](nereids) refactor analyze view

924060929:
## Proposed changes

The Analyzer of NereidsPlanner use different rules to analyze normal plan and view, to prevent the plans in the views analyze multiple times, because some rules can not apply multiple times, say, decimal type coercion, if this rule applied multiple times, it will generate a wrong result.

But this design is trick. Normally, after process the LogicalView, the whole plan tree in the LogicalView should not contains unbound plan, but the current situation is not like this. And this problem block development of some rules, so I refactor it:
1. the Analyzer will not traverse the children of the LogicalView
2. After link the LogicalView to the outer plan tree, the whole plan tree of the LogicalView will not contains unbound plan
3. analyze view and table use the same rules, keep it simple

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/nereids/CascadesContext.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/nereids/CascadesContext.java
@@ -240,19 +240,11 @@ public class CascadesContext implements ScheduleContext {
     }
 
     public Analyzer newAnalyzer() {
-        return newAnalyzer(false);
-    }
-
-    public Analyzer newAnalyzer(boolean analyzeView) {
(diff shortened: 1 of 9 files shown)


924060929:
run buildall

924060929:
run buildall

morrySnow (dismissed):

924060929:
run buildall

924060929:
run buildall

morrySnow (approved):

feiniaofeiafei (approved):

starocean999 (approved):

PR #36840, opened 2024-06-26
[fix](mtmv)fix MTMVTask log has null pointer

zddr:
`LOG.info("finish create mtmv task, task: {}", task);`
==> finish create mtmv task, task: [!!!org.apache.doris.job.extensions.mtmv.MTMVTask@xxx=>java.lang.NullPointerException:null!!!]

Suspected to be a problem with lombok, therefore displaying the implementation of a toString method

PR #37131, opened 2024-07-02
[fix](mtmv)fix MTMVTask log has null pointer (#36840)

zddr:
pick https://github.com/apache/doris/pull/36840


zddr:
run buildall

PR #36928, opened 2024-06-27
[fix](auth)support check priv when tvf use resource

zddr:
current, follow sql not check priv, we need check `usage priv` of `test_resource`
```
SELECT * FROM S3 (
    "uri" = "xxx",
    "format" = "orc",
    "resource" = "test_resource",
    "use_path_style" = "true"
);
````

PR #37132, opened 2024-07-02
[fix](auth)support check priv when tvf use resource (#36928)

zddr:
pick https://github.com/apache/doris/pull/36928


zddr:
run buildall

zddr:
run buildall

PR #37135, opened 2024-07-02
[fix](txn insert) Forbid delete condition in txn insert

mymeiyi:
## Proposed changes

Currently, there are two kinds of delete in doris, one is delete-where which is implemented via fast path, while the other one is delete-using which is implemented via insert into alike.

The first one can read writings in the same txn while the second one can not. This is confusing for users, so we just disable delete after insert and update now.

doc: https://github.com/apache/doris-website/pull/820


mymeiyi:
run buildall

mymeiyi:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):
LGTM

PR #37146, opened 2024-07-02
[opt](catalog) add some profile for parquet reader and change meta cache config (#37040)

morningman:
bp #37040


morningman:
run buildall

PR #37148, opened 2024-07-02
[opt](function)Optimize BinaryArithmetic by handling the null column internally.

Mryange:
## Proposed changes

The default implementation of null handling requires the following steps:

1. Construct a new block, removing nullable.
2. Execute.
3. Generate a null map based on input parameters.
4. Destroy the new block.

These processes do not have significant overhead, but because the BinaryArithmetic series of functions are very fast, they might seem slow in comparison. Therefore, an optimization is made here: moving the null handling into BinaryArithmetic can reduce the overhead of steps 1 and 4.

expr time from 2.222s to 2.045s


Mryange:
run buildall

PR #36871, opened 2024-06-26
[test](tvf) move p2 tvf tests from p2 to p0

BePPPower:
## Proposed changes

Issue Number: close #xxx

Move test `test_tvf_p2`, `test_path_partition_keys` and `test_hdfs_tvf_compression` from external_table_p2 to external_table_p0,
so that developer can run it locally with docker env.

PR #37150, opened 2024-07-02
[test](tvf) move p2 tvf tests from p2 to p0 (#36871)

BePPPower:
## Proposed changes

bp: #36871 
Move test `test_tvf_p2`, `test_path_partition_keys` and `test_hdfs_tvf_compression` from external_table_p2 to external_table_p0, so that developer can run it locally with docker env.

## Proposed changes

Issue Number: close #xxx


BePPPower:
run buildall

PR #36668, opened 2024-06-21
[Fix](csv_reader) Add a session variable to control whether empty rows in CSV files are read as NULL values.

BePPPower:
## Proposed changes

Issue Number: close #xxx

Previously, empty rows in CSV files were ignored by Doris.
Now, add a session variable to control whether empty rows in CSV files are read as NULL values.

PR #37153, opened 2024-07-02
[Fix](csv_reader) Add a session variable to control whether empty rows in CSV files are read as NULL values

BePPPower:
bp: #36668

Previously, empty rows in CSV files were ignored by Doris. Now, add a session variable to control whether empty rows in CSV files are read as NULL values.


BePPPower:
run buildall

PR #37159, opened 2024-07-02
[improvement] Log workload group's thread num when publish

wangbo:
## Proposed changes
Some user feedback that BE may crash because of thread resource exhaustion after running for a period of time.
So log workload group's threadpool real thread num and config num when every time FE publish.
When the same case happens, we can find workload group's thread debug info in ```be.INFO``` to check whether workload group has thread leak, or it could just be an unreasonable configuration.
Besides I remove some useless code and do little code refactor.


wangbo:
run buildall

wangbo:
run buildall

HappenLee (approved):
LGTM

wangbo:
run buildall

wangbo:
run buildall

xinyiZzz (approved):
LGTM

PR #35489, opened 2024-05-28
[fix](tvf) Partition columns in CTAS need to be compatible with the STRING type of external tables/TVF

BePPPower:
## Proposed changes

Issue Number: close #xxx

Partition columns in CTAS need to be compatible with the STRING type of external tables/TVF

1. For TVF queries with partition columns in the file path, map these partition columns to the `VARCHAR(65533)` type. (Previously, they were of `STRING` type.)

3. If you use the `STRING` type of an external table as a partition column, CTAS will automatically convert it to `VARCHAR(65535)` type."

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #37161, opened 2024-07-02
[fix](tvf) Partition columns in CTAS need to be compatible with the STRING type of external tables/TVF

BePPPower:
bp: #35489

Partition columns in CTAS need to be compatible with the STRING type of external tables/TVF

1. For TVF queries with partition columns in the file path, map these partition columns to the `VARCHAR(65533)` type. (Previously, they were of `STRING` type.)

3. If you use the `STRING` type of an external table as a partition column, CTAS will automatically convert it to `VARCHAR(65535)` type."


BePPPower:
run buildall

BePPPower:
run buildall

PR #36490, opened 2024-06-18
[Fix](export/outfile) Support compression when exporting data to Parquet / ORC.

BePPPower:
## Proposed changes

Issue Number: close #xxx

When using Export/Outfile, we can use properties `compress_type` to specify the compression method.

Currently, Parquet file format supports `SNAPPY`, `GZIP`, `BROTLI`, `ZSTD`, `LZ4`, `LZO`, `BZ2` and `PLAIN` compression methods, default value is `SNAPPY`.

 Orc file format supports `PLAIN`, `SNAPPY`, `ZLIB`, and `ZSTD` compression methods, default value is `ZLIB`.  

Docs: https://github.com/apache/doris-website/pull/764

PR #37167, opened 2024-07-02
[Fix](export/outfile) Support compression when exporting data to Parquet / ORC.

BePPPower:
bp: #36490

When using Export/Outfile, we can use properties `compress_type` to specify the compression method.

Currently, Parquet file format supports `SNAPPY`, `GZIP`, `BROTLI`, `ZSTD`, `LZ4`, `LZO`, `BZ2` and `PLAIN` compression methods, default value is `SNAPPY`.

 Orc file format supports `PLAIN`, `SNAPPY`, `ZLIB`, and `ZSTD` compression methods, default value is `ZLIB`.


BePPPower:
run buildall

PR #37168, opened 2024-07-02
[fix](nereids) the stop watch is not reset

mymeiyi:
## Proposed changes

When use JDBC to insert into Doris, after execute several sqls, will get `Nereids cost too much time (> 30s )`, but the audit log shows that the sql only cost several  milliseconds. The reason is that the stop watch is never reset.


dataroaring (approved):
LGTM

morrySnow (approved):

PR #37171, opened 2024-07-02
[fix](compaction) fix null pointer when retrieving CPU load average

luwei16:
1 When enable_system_metrics = false, system_metrics is a null pointer
2 When BE exits, system_metrics has been destructed, and the compaction background thread may still be running

### DIFF
--- a/be/src/util/system_metrics.cpp
+++ b/be/src/util/system_metrics.cpp
@@ -870,7 +870,7 @@ void SystemMetrics::get_disks_io_time(std::map<std::string, int64_t>* map) {
 }
 
 double SystemMetrics::get_load_average_1_min() {
-    if (!_load_average_metrics) {
+    if (_load_average_metrics) {
         return _load_average_metrics->load_average_1_minutes->value();
     } else {
         return 0;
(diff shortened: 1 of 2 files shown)


luwei16:
run buildall

dataroaring (dismissed):
LGTM

be/src/util/system_metrics.cpp
```diff
@@ -870,7 +870,11 @@ void SystemMetrics::get_disks_io_time(std::map<std::string, int64_t>* map) {
 }
 
 double SystemMetrics::get_load_average_1_min() {
-    return _load_average_metrics->load_average_1_minutes->value();
+    if (!_load_average_metrics) {
```
gavinchou, line 873: it seems incorrect

luwei16:
run buildall

be/src/util/system_metrics.cpp
luwei16: done

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #36705, opened 2024-06-22
[Fix](multi-catalog) Fix the transaction is not removed in abnormal situations by removing transaction in finally block.

kaka11chen:
## Proposed changes

[Fix] (multi-catalog) Fix the transaction is not removed in abnormal situations by removing transaction in `finally` block.

PR #37172, opened 2024-07-02
[Fix](multi-catalog) Fix the transaction is not removed in abnormal situations by removing transaction in finally block.

kaka11chen:
## Proposed changes

Backport #36705.


kaka11chen:
run buildall

PR #37101, opened 2024-07-01
[feature](ES Catalog) map nested/object type in ES to JSON type in Doris

qidaye:
## Proposed changes

1. `nested`/`object` can map to `json` type in Doris, and can be analyzed with json functions.
2. Add some cases for `json_extract`.

PR #37182, opened 2024-07-02
[feature](ES Catalog) map nested/object type in ES to JSON type in Doris (#37101)

qidaye:
backport #37101


qidaye:
run buildall

PR #37082, opened 2024-07-01
[fix](manager)fix manager interface case.

hubgeter:
## Proposed changes

Issue Number: close #xxx

PR #37184, opened 2024-07-02
[fix](manager)fix manager interface case. 

hubgeter:
## Proposed changes
bp #37082 
 
Issue Number: close #xxx


hubgeter:
run buildall

PR #37185, opened 2024-07-02
[improve](ip)improve cidr query for looking up ips

amorynan:
## Proposed changes

Issue Number: close #xxx


be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -1344,8 +1344,9 @@ Status SegmentIterator::_apply_inverted_index() {
             // _common_expr_columns has all column ids from _common_expr_ctxs_push_down
             // if current bitmap is already empty just return
             if (!(expr_ctx->root()->node_type() == TExprNodeType::FUNCTION_CALL &&
-                  expr_ctx->root()->fn().name.function_name ==
```
zzzxl1993, line 1347: 1.  Add an inverted index field of string type called "type".
2.  Test the query: select * from tbl where (array_contains(inventors, 's') or type = '2');
3.  The root node being of type compound_or may prevent index optimization.

PR #37187, opened 2024-07-02
[fix](HadoopLz4BlockCompression)Fixed the bug that HadoopLz4BlockCompression creates _decompressor every time it decompresses.

hubgeter:
## Proposed changes

Problem:
HadoopLz4BlockCompression uses a singleton mode.
Every time its decompression function is called, a new object will be created for its member variable _decompressor. In a multi-threaded scenario, new objects will be created for _decompressor at the same time.

Fix method:
During the construction phase of HadoopLz4BlockCompression, a new object is created for its member variable _decompressor. When calling the decompression function, there is no need to create a new object for _decompressor.

Issue Number: close #xxx

### DIFF
--- a/be/src/util/block_compression.cpp
+++ b/be/src/util/block_compression.cpp
@@ -25,6 +25,8 @@
         defined(__i386) || defined(_M_IX86)
 #include <libdeflate.h>
 #endif
+#include <glog/log_severity.h>
+#include <glog/logging.h>
 #include <limits.h>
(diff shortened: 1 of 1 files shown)


hubgeter:
run buildall

hubgeter:
run p0

morningman (approved):
LGTM

AshinGau (approved):

PR #37043, opened 2024-06-30
[fix](hive) support find serde info from both tbl properties and serde properties

morningman:
Some hive table set serde properties in tblproperties.
We need to support it

PR #37188, opened 2024-07-02
[fix](hive) support find serde info from both tbl properties and serde properties (#37043)

morningman:
bp #37043


morningman:
run buildall

morningman:
run buildall

PR #37220, opened 2024-07-03
[improvement](routine load)add a method to quit before guard lock

wonderzerg:
## Proposed changes
When routine load`s workload is heavy, sometimes hangs at guard lock.

Issue Number: close #xxx

Add a method to avoid an invalid lock being entered.


wonderzerg:
run buildall

PR #37221, opened 2024-07-03
[fix](spill) Adjust the logic for triggering spilling

mrhhsg:
## Proposed changes

1. If 'query_weighted_limit` was not set, do not trigger spilling.
2. Using `process_memory_usage` to represent the process memory usage is more reasonable.


jacktengg (approved):

HappenLee (approved):
LGTM

PR #37228, opened 2024-07-03
[opt](binlog) support binlog for rename operator

lsy3993:
Rename operator doesn't have binlog now. This PR will create binlog when execute rename. The rename operator means :

1. rename table : ALTER TABLE table1 RENAME table2;
2. rename rollup : ALTER TABLE example_table RENAME ROLLUP rollup1 rollup2;
3. rename partition : ALTER TABLE example_table RENAME PARTITION p1 p2;

After SQL analyzing, we can get the old and new table name (rollup name or partition name) , then record the info to binlog, so we can use the info from binlog.


dataroaring:
run buildall

lsy3993:
no need to branch-2.1 so close

PR #37233, opened 2024-07-03
[Fix](regression) Fix cluster privilege for workload group case in cloud mode

wangbo:
## Proposed changes
When regression test test a cloud mode Doris cluster, non-root user should set a cluster name, otherwise query may failed.


wangbo:
run buildall

gavinchou (approved):

BiteTheDDDDt (approved):

PR #37240, opened 2024-07-03
[fix](TVF) Enable azure tvf without specifying s3 bucket property.

ByteYue:
## Proposed changes

For azure blob storage, the region property is unnecessary. But the legacy tvf code would check if the properties contains bucket, so this pr tries to set one mock bucket property for azure tvf.


ByteYue:
run buildall

fe/fe-core/src/main/java/org/apache/doris/tablefunction/S3TableValuedFunction.java
```diff
@@ -57,6 +57,10 @@ public class S3TableValuedFunction extends ExternalFileTableValuedFunction {
 
     public S3TableValuedFunction(Map<String, String> properties) throws AnalysisException {
         final boolean isAzureTvf = AzureProperties.checkAzureProviderPropertyExist(properties);
+        // Azure could run without region
+        if (isAzureTvf) {
+            properties.put(S3Properties.REGION, "REGION");
```
gavinchou, line 62: "dummy-region"

gavinchou:
Design needs to be review, pending merging

gavinchou (approved):

dataroaring (approved):
LGTM

PR #20907, opened 2023-06-16
[Enhancement](Compaction) Caculate all committed rowsets delete bitmaps when do comapction

Yukang-Lian:
## Proposed changes

Issue Number: close #xxx

Here we will calculate all the rowsets delete bitmaps which are committed but not published to reduce the calculation pressure of publish phase.

Step1: collect this tablet's all committed rowsets' delete bitmaps.

Step2: calculate all rowsets' delete bitmaps which are published during compaction.

Step3: write back updated delete bitmap and tablet info.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #37242, opened 2024-07-03
[opt](merge-on-write) Skip considering rowsets produced by compaction from commited txns' rowsets in publish phase

bobhan1:
## Proposed changes

https://github.com/apache/doris/pull/20907 let compaction calculates the delete bitmap of the output rowset from committed txns' rowsets to reduce the calculation pressure of publish phase. But this output rowset will be considered as new rowset in the publish phase and will be calculated again in the publish phase. This PR let the publish phase calculation skip considering these rowsets.

### DIFF
--- a/be/src/olap/base_tablet.cpp
+++ b/be/src/olap/base_tablet.cpp
@@ -1066,6 +1066,15 @@ void BaseTablet::add_sentinel_mark_to_delete_bitmap(DeleteBitmap* delete_bitmap,
     }
 }
 
+void BaseTablet::add_skip_mark_to_delete_bitmap(DeleteBitmap* delete_bitmap,
+                                                std::vector<RowsetSharedPtr> rowsets) {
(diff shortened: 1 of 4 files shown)


dataroaring:
run buildall

be/src/olap/base_tablet.cpp
```diff
@@ -1066,6 +1066,15 @@ void BaseTablet::add_sentinel_mark_to_delete_bitmap(DeleteBitmap* delete_bitmap,
     }
 }
 
+void BaseTablet::add_skip_mark_to_delete_bitmap(DeleteBitmap* delete_bitmap,
```
zhannngchen, line 1069: you don't need to add a mark, just check the rowset's version, if the version is not single version, e.g [x-x], it means that the new rowset is generated by compaction, you can skip it directly

PR #37245, opened 2024-07-03
[Fix](Variant) fix should not implicit cast to variant

eldenmoon:
query like
```
SELECT "1234"["abc"]
```
will be casted to variant implicitly

## Proposed changes

Issue Number: close #xxx


eldenmoon:
run buildall

PR #37246, opened 2024-07-03
[fix](cloud) add lock when get visible version from cache

sollhui:
## Proposed changes

introduced by https://github.com/apache/doris/pull/34615

When get visible version from cache, the visible version may be updated by load, so it is necessary to add lock when get visible version.


sollhui:
run buildall

PR #37265, opened 2024-07-04
[fix](txn insert) sub txn map is not thread safe

mymeiyi:
## Proposed changes

The sub txn map in DatabaseTransactionMgr is not thread safe

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
@@ -136,7 +136,7 @@ public class DatabaseTransactionMgr {
 
     // transactionId -> final status TransactionState
     private final Map<Long, TransactionState> idToFinalStatusTransactionState = Maps.newHashMap();
-    private final Map<Long, Long> subTxnIdToTxnId = Maps.newHashMap();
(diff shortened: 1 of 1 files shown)


mymeiyi:
run buildall

fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
```diff
@@ -2830,13 +2830,24 @@ private PublishResult checkQuorumReplicas(TransactionState transactionState, lon
     }
 
     protected void addSubTransaction(long transactionId, long subTransactionId) {
-        subTxnIdToTxnId.put(subTransactionId, transactionId);
+        writeLock();
+        try {
+            subTxnIdToTxnId.put(subTransactionId, transactionId);
```
gavinchou, line 2835: `subTxnIdToTxnId` is a concurrent map, is it necessary to lock it?

mymeiyi: it's not concurrent map `private final Map<Long, Long> subTxnIdToTxnId = Maps.newHashMap()`

dataroaring (dismissed):
LGTM

mymeiyi:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #37266, opened 2024-07-04
[fix](trino-connector) Fix the problem of reading the `DATE` type from Trino too slowly

BePPPower:
Previously, we use `LocalDate.parse(DateTimeUtils.printDate(block.getInt(position, 0)), DateTimeFormatter.ISO_DATE)` to handle DATE type from Trino.

Now, we use `LocalDate.ofEpochDay(block.getInt(position, 0))` to handle DATE type from Trino.
This will result in a 20-fold speed improvement.


BePPPower:
run buildall

morningman (approved):
LGTM

hubgeter (approved):

PR #33726, opened 2024-04-16
[Fix](planner) fix create view star except and modify cast to sql

feiniaofeiafei:
fix old planner create view bug when the select has star except and cast(sth to time)

PR #37274, opened 2024-07-04
[fix](planner) fix bug of select stmt toSql

feiniaofeiafei:
PR #33726 does not correctly fix the create view problem of the legacy optimizer caused by the toSql bug (* except syntax). The results of the regression test cases are also problematic. This pr fix the problem.


feiniaofeiafei:
run buildall

feiniaofeiafei:
run p0

morrySnow (approved):

starocean999 (approved):

PR #37287, opened 2024-07-04
[chore](cloud) Define std::stringstream in local

w41ter:


w41ter:
run buildall

cloud/src/meta-service/meta_service.cpp
```diff
@@ -1591,13 +1594,13 @@ static bool check_delete_bitmap_lock(MetaServiceCode& code, std::string& msg, st
     return true;
 }
 
-static bool process_pending_delete_bitmap(MetaServiceCode& code, std::string& msg,
-                                          std::stringstream& ss, std::unique_ptr<Transaction>& txn,
```
gavinchou, line 1595: We have to check if `ss` already has something in it, say, 
```
ss << "test";  // called somewhere
...
process_pending_delete_bitmap(..., ss, ...);
...
```

PR #37294, opened 2024-07-04
[Fix](Timezone)The time zone should not be fixed to UTC+8.

CalvinKirs:
## Proposed changes

It is generally more reasonable to use the system's default time zone rather than binding the default time zone to UTC+8. This approach ensures that the application will behave correctly in different environments and regions.

**For countries that observe DST, this might require extra attention.**


CalvinKirs:
run buildall

morrySnow (approved):

JNSimba (approved):
LGTM

PR #37298, opened 2024-07-04
[enhance](Recycler) Catch RequestFailedException for all azure request

ByteYue:
## Proposed changes

Previously we only catch storageException for azure's request. But it could also throw RequestFailedException which is the base class for storageException. So we directly catch RequestFailedException here.


cloud/src/recycler/azure_obj_client.cpp
```diff
@@ -45,7 +45,7 @@ template <typename Func>
 ObjectStorageResponse do_azure_client_call(Func f, std::string_view url, std::string_view key) {
     try {
         f();
-    } catch (Azure::Storage::StorageException& e) {
+    } catch (Azure::Core::RequestFailedException& e) {
```
gavinchou, line 48: what about `catch (...)`?

ByteYue:
run buildall

platoneko (approved):

gavinchou (approved):

PR #37308, opened 2024-07-04
[enhance](Azure) Use s3Uri to specify the object's bucket and key for azure in FE

ByteYue:
Previously when using s3 load on azure blob storage, user should specify the s3.bucket property. But actually we can get the bucket information from the data infile uri.


ByteYue:
run buildall

gavinchou (approved):

dataroaring (approved):
LGTM

PR #37312, opened 2024-07-04
[enhancement](cloud) support TTL file cache evict through LRU

freemandealer:
# Motivation and Basic Ideas
Originally, the TTL file cache only evicts when it expires. If TTL data fills the cache, new TTL data won't fit in the cache and thus switch to SKIP_CACHE mode, which is a performance killer that overruns the S3 downloader with tons of small IOs.

This commit enables evicting the TTL cache actively through LRU beside the original TTL expiration.

# Performance tests
We can set up a scenario where the TTL cache is full by using a small file cache space (e.g. 5GB). We use table comsumer_ttl in the regression test attached with the PR and load data several times (e.g. 20GB) to populate the TTL cache.

With this setting, we execute query `select count(*) from customer_ttl where C_ADDRESS like '%ea%' and C_NAME like '%a%' and C_COMMENT like '%b%'` separately while enable_ttl_cache_evict_using_lru = true/false in the be conf. The results shows that the performance is increased by 43x during the tests:

| ttl_cache_evict_using_lru feature | enabled | disabled (baseline) |
| --------------------------------- | ------- | ------------------- |
| query execution time              | 8.5s    | 342s                |

Note that the result heavily depends on the ratio of cache size and the amount of data that the query incurs, along with S3 server performance, etc, so the result is only here for reference.

Such improvement is achieved by the active eviction of the TTL cache , which effectively reduces the possibility of SKIP_CACHE. During the tests:

| ttl_cache_evict_using_lru feature | enabled | disabled (baseline) |
| --------------------------------- | ------- | ------------------- |
| SKIP_CACHE occurence              | 0   | 257.1K               |

### DIFF
--- a/be/src/io/cache/block_file_cache.cpp
+++ b/be/src/io/cache/block_file_cache.cpp
@@ -56,6 +56,8 @@ BlockFileCache::BlockFileCache(const std::string& cache_base_path,
             _cache_base_path.c_str(), "file_cache_ttl_cache_size", 0);
     _cur_normal_queue_element_count_metrics = std::make_shared<bvar::Status<size_t>>(
             _cache_base_path.c_str(), "file_cache_normal_queue_element_count", 0);
+    _cur_ttl_cache_lru_queue_size_metrics = std::make_shared<bvar::Status<size_t>>(
+            _cache_base_path.c_str(), "file_cache_ttl_cache_lru_queue_size", 0);
     _cur_normal_queue_cache_size_metrics = std::make_shared<bvar::Status<size_t>>(
             _cache_base_path.c_str(), "file_cache_normal_queue_cache_size", 0);
     _cur_index_queue_element_count_metrics = std::make_shared<bvar::Status<size_t>>(
@@ -84,6 +86,8 @@ BlockFileCache::BlockFileCache(const std::string& cache_base_path,
                             7 * 24 * 60 * 60);
     _normal_queue = LRUQueue(cache_settings.query_queue_size, cache_settings.query_queue_elements,
                              24 * 60 * 60);
+    _ttl_queue = LRUQueue(std::numeric_limits<int>::max(), std::numeric_limits<int>::max(),
+                          std::numeric_limits<int>::max());
 
     LOG(INFO) << fmt::format(
             "file cache path={}, disposable queue size={} elements={}, index queue size={} "
@@ -213,7 +217,8 @@ void BlockFileCache::use_cell(const FileBlockCell& cell, FileBlocks* result, boo
         result->push_back(cell.file_block);
     }
 
-    if (cell.file_block->cache_type() != FileCacheType::TTL) {
+    if (cell.file_block->cache_type() != FileCacheType::TTL ||
+        config::enable_ttl_cache_evict_using_lru) {
         auto& queue = get_queue(cell.file_block->cache_type());
         DCHECK(cell.queue_iterator) << "impossible";
         /// Move to the end of the queue. The iterator remains valid.
@@ -280,7 +285,14 @@ FileBlocks BlockFileCache::get_impl(const UInt128Wrapper& hash, const CacheConte
             if (st.ok()) {
                 auto& queue = get_queue(origin_type);
                 queue.remove(cell.queue_iterator.value(), cache_lock);
-                cell.queue_iterator.reset();
+                if (config::enable_ttl_cache_evict_using_lru) {
+                    auto& ttl_queue = get_queue(FileCacheType::TTL);
+                    cell.queue_iterator = ttl_queue.add(
+                            cell.file_block->get_hash_value(), cell.file_block->offset(),
+                            cell.file_block->range().size(), cache_lock);
+                } else {
+                    cell.queue_iterator.reset();
+                }
                 st = cell.file_block->update_expiration_time(context.expiration_time);
             }
             if (!st.ok()) {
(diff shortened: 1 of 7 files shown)


be/src/io/cache/block_file_cache.cpp
```diff
@@ -796,6 +817,58 @@ bool BlockFileCache::try_reserve_for_ttl(size_t size, std::lock_guard<std::mutex
...
+            FileBlockSPtr file_block = cell->file_block;
+            if (file_block) {
+                std::lock_guard block_lock(file_block->_mutex);
+                remove(file_block, cache_lock, block_lock);   // 维护 time_to_hash &  hash_to_map
+            }
+        };
+
+        std::for_each(to_evict.begin(), to_evict.end(), remove_file_block_if);
+
+        if (is_overflow(removed_size, size, cur_cache_size)) {
```
gavinchou, line 862: just `return is_overflow(removed_size, size, cur_cache_size)`

dataroaring:
run buildall

be/src/io/cache/block_file_cache.cpp
```diff
@@ -796,6 +818,70 @@ bool BlockFileCache::try_reserve_for_ttl(size_t size, std::lock_guard<std::mutex
     return true;
 }
 
+bool BlockFileCache::try_reserve_for_ttl(size_t size, std::lock_guard<std::mutex>& cache_lock) {
+    if (try_reserve_for_ttl_without_lru(size, cache_lock)) {
```
Lchangliang, line 822:
```
if (!try_reserve_for_ttl_without_lru(size, cache_lock) && config::enable_ttl_cache_evict_using_lru) {
}
```

be/src/io/cache/block_file_cache.cpp
```diff
@@ -796,6 +818,70 @@ bool BlockFileCache::try_reserve_for_ttl(size_t size, std::lock_guard<std::mutex
...
+            }
+        }
+
+        auto remove_file_block_if = [&](FileBlockCell* cell) {
+            FileBlockSPtr file_block = cell->file_block;
+            if (file_block) {
+                std::lock_guard block_lock(file_block->_mutex);
+                auto hash = cell->file_block->get_hash_value();
+                remove(file_block, cache_lock, block_lock);
+                if (_files.find(hash) == _files.end()) {
```
Lchangliang, line 859: it dont need to do this

be/src/io/cache/block_file_cache.cpp
```diff
@@ -1463,6 +1553,7 @@ void BlockFileCache::run_background_operation() {
                                                _index_queue.get_capacity(cache_lock) -
                                                _normal_queue.get_capacity(cache_lock) -
                                                _disposable_queue.get_capacity(cache_lock));
+        _cur_ttl_cache_lru_queue_size_metrics->set_value(_ttl_queue.get_capacity(cache_lock));
```
Lchangliang, line 1556: add `_cur_ttl_queue_element_count_metrics`

freemandealer: sure thing

be/src/io/cache/block_file_cache.cpp
freemandealer: > it dont need to do this
why?

be/src/io/cache/block_file_cache.cpp
freemandealer: cannot see the improvement of code correctness and readability of such change

freemandealer:
run buildall

freemandealer:
run buildall

freemandealer:
run buildall

freemandealer:
run buildall

freemandealer:
run p0

gavinchou (dismissed):

Lchangliang (approved):

be/src/io/cache/block_file_cache.cpp
```diff
@@ -796,6 +820,70 @@ bool BlockFileCache::try_reserve_for_ttl(size_t size, std::lock_guard<std::mutex
...
+
+            if (cell->releasable()) {
+                auto& file_block = cell->file_block;
+
+                std::lock_guard block_lock(file_block->_mutex);
+                DCHECK(file_block->_download_state == FileBlock::State::DOWNLOADED);
+                to_evict.push_back(cell);
+
+                removed_size += cell_size;
+            }
```
dataroaring, line 852: duplicated codes in try_reserve_for_ttl_without_lru.

be/src/io/cache/block_file_cache.cpp
```diff
@@ -796,6 +820,70 @@ bool BlockFileCache::try_reserve_for_ttl(size_t size, std::lock_guard<std::mutex
...
+                remove(file_block, cache_lock, block_lock);
+                if (_files.find(hash) == _files.end()) {
+                    if (auto iter = _key_to_time.find(hash);
+                        _key_to_time.find(hash) != _key_to_time.end()) {
+                        auto _time_to_key_iter = _time_to_key.equal_range(iter->second);
+                        while (_time_to_key_iter.first != _time_to_key_iter.second) {
+                            if (_time_to_key_iter.first->second == hash) {
+                                _time_to_key_iter.first =
+                                        _time_to_key.erase(_time_to_key_iter.first);
+                                break;
```
dataroaring, line 869: too many idents here, we should supress indents via function call.

dataroaring (approved):
LGTM

freemandealer:
run buildall

PR #37297, opened 2024-07-04
[pipeline](fix) Set upstream operators always runnable once source operator closed

Gabriel39:
## Proposed changes

Some kinds of source operators has a 1-1 relationship with a sink operator (such as AnalyticOperator). We must ensure AnalyticSinkOperator will not be blocked if AnalyticSourceOperator already closed.

PR #37325, opened 2024-07-05
[pipeline](fix) Set upstream operators always runnable once source op…

Gabriel39:
…erator closed (#37297)

Some kinds of source operators has a 1-1 relationship with a sink operator (such as AnalyticOperator). We must ensure AnalyticSinkOperator will not be blocked if AnalyticSourceOperator already closed.

pick #37297

## Proposed changes

Issue Number: close #xxx


Gabriel39:
run buildall

PR #37295, opened 2024-07-04
[fix](fe) Fix the sql of AddPartitionRecord

w41ter:
The range field is accidentally compared to DUMMY_ITEM.

It was introduced by #35461.

The case is improved in selectdb/ccr_syncer#118.

Affected versions:
- 2.0.11 ~ 2.0.12
- 2.1.4

PR #37341, opened 2024-07-05
[fix](fe) Fix the sql of AddPartitionRecord

w41ter:
Cherry-pick #37295

The range field is accidentally compared to DUMMY_ITEM.

It was introduced by #35461.


w41ter:
run buildall

PR #36237, opened 2024-06-13
[Opt](streamload) Commit txn on BE coordinator in cloud mode to avoid RPC and lock overhead

gavinchou:
In cloud mode, FE master is not mandatory for committing load txn, and there is overhead (locks and PRC) if we commit load txn via FE master.

PR #37347, opened 2024-07-05
[fix](cloud) fix routine load job state can not changed correctly

sollhui:
## Proposed changes

introduce by https://github.com/apache/doris/pull/36237

If meets some error like `out of range`, routine load job state should change from running to pause.

When rollback transaction, the RPC will send to meta service directly, which can not change state by transaction information. 

This pr sends RPC to FE first to solve this problem.


sollhui:
run buildall

liaoxin01 (approved):
LGTM

dataroaring (approved):
LGTM

PR #37353, opened 2024-07-05
[regression](plugin) Support different object storage for recycler plugin

ByteYue:
Former implementation only supports recycler plugin for s3, this pr makes azure also available.


regression-test/framework/src/main/groovy/org/apache/doris/regression/util/ObjectListUtil.groovy
```diff
@@ -0,0 +1,92 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
gavinchou, line 1: where is azure?

platoneko (dismissed):

ByteYue:
run buildall

gavinchou (approved):

platoneko (approved):

PR #37354, opened 2024-07-05
[Improvement](join) catch memory allocate fail on hash join build

BiteTheDDDDt:
## Proposed changes
 catch memory allocate fail on hash join build


BiteTheDDDDt:
run buildall

HappenLee (approved):
LGTM

PR #36632, opened 2024-06-20
[chore](routine-load) increase routine load job default max batch size and rows

sollhui:
Most users only care about the size of **max_batch_interval**, but in order to achieve an interval effect, they have to configure **max_batch_rows** and **max_batch_size** according to the characteristics of the data. By adjusting these two default values, users do not need to worry about configuration in most scenarios.

PR #37388, opened 2024-07-07
[branch-2.1](routine-load) increase routine load job default max batch size and rows 

sollhui:
pick #36632

Most users only care about the size of **max_batch_interval**, but in order to achieve an interval effect, they have to configure **max_batch_rows** and **max_batch_size** according to the characteristics of the data. By adjusting these two default values, users do not need to worry about configuration in most scenarios.

## Proposed changes

Issue Number: close #xxx


sollhui:
run buildall

PR #37398, opened 2024-07-07
[improve](group commit) Pick Modify group commit case and modify cancel status (#35995)

mymeiyi:
Pick https://github.com/apache/doris/pull/35995


mymeiyi:
run buildall

PR #36408, opened 2024-06-17
[fix](group commit) Fix the incorrect group commit count in log; fix the core in get_first_block

mymeiyi:
## Proposed changes
1. For group commit in sync_mode, we always get `this group commit includes 0 loads` when finish inener load, because the destruction of inner load is before the user load
2. Fix:
```
*** SIGSEGV invalid permissions for mapped object (@0x55ee31bf2585) received by PID 2459843 (TID 2460361 OR 0x7f5702977700) from PID 834610565; stack trace: ***
 0# doris::signal::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*) at /home/zcp/repo_center/doris_master/doris/be/src/common/signal_handler.h:421
 1# PosixSignals::chained_handler(int, siginfo*, void*) [clone .part.0] in /usr/lib/jvm/jdk-17.0.2/lib/server/libjvm.so
 2# JVM_handle_linux_signal in /usr/lib/jvm/jdk-17.0.2/lib/server/libjvm.so
 3# 0x00007F580B170090 in /lib/x86_64-linux-gnu/libc.so.6
 4# std::_Function_handler<void (), doris::GroupCommitTable::get_first_block_load_queue(long, long, doris::UniqueId const&, std::shared_ptr<doris::LoadBlockQueue>&, int, std::shared_ptr<doris::MemTrackerLimiter>)::$_0>::_M_invoke(std::_Any_data const&) at /var/local/ldb_toolchain/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/std_function.h:291
 5# doris::ThreadPool::dispatch_thread() in /mnt/hdd01/STRESS_ENV/be/lib/doris_be
```

PR #37405, opened 2024-07-07
[fix](group commit) Pick Fix the incorrect group commit count in log; fix the core in get_first_block (#36408)

mymeiyi:
Pick https://github.com/apache/doris/pull/36408/


mymeiyi:
run buildall

PR #37416, opened 2024-07-08
[refine](agg)Adjust the value of STREAMING_HT_MIN_REDUCTION

Mryange:
## Proposed changes

A size of 16MB is too large in many cases.


Mryange:
run buildall

PR #31303, opened 2024-02-23
[fix](routine-load) fix routine load job throw exception after commit

sollhui:
## Proposed changes

```
[RoutineLoadJob.afterCommitted():1032] after committed failed
java.util.NoSuchElementException: No value present
        at java.util.Optional.get(Optional.java:135) ~[?:1.8.0_382]
        at org.apache.doris.load.routineload.RoutineLoadJob.afterCommitted(RoutineLoadJob.java:1025) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.transaction.CloudGlobalTransactionMgr.commitTransaction(CloudGlobalTransactionMgr.java:346) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.transaction.CloudGlobalTransactionMgr.commitTransaction(CloudGlobalTransactionMgr.java:253) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.transaction.CloudGlobalTransactionMgr.commitAndPublishTransaction(CloudGlobalTransactionMgr.java:595) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.service.FrontendServiceImpl.loadTxnCommitImpl(FrontendServiceImpl.java:1670) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.service.FrontendServiceImpl.loadTxnCommit(FrontendServiceImpl.java:1606) ~[doris-fe.jar:1.2-SNAPSHOT]
        at sun.reflect.GeneratedMethodAccessor101.invoke(Unknown Source) ~[?:?]
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_382]
        at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_382]
        at org.apache.doris.service.FeServer.lambda$start$0(FeServer.java:59) ~[doris-fe.jar:1.2-SNAPSHOT]
        at com.sun.proxy.$Proxy47.loadTxnCommit(Unknown Source) ~[?:?]
        at org.apache.doris.thrift.FrontendService$Processor$loadTxnCommit.getResult(FrontendService.java:3382) ~[fe-common-1.2-SNAPSHOT.jar:1.2-SNAPSHOT]
[...]

PR #37432, opened 2024-07-08
[branch-2.0](routine-load) fix routine load job throw exception after commit (#31303)

sollhui:
pick #31303


sollhui:
run buildall

PR #32021, opened 2024-03-09
[opt](routine-load) optimize allocate task to be algorithm for load balance

sollhui:
## Proposed changes

The task always running on one BE may cause the BE to run too many tasks simultaneously while the other BE is not running few task. Here, a simple limitation is added to optimize this situation.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #37434, opened 2024-07-08
[opt](routine-load) optimize allocate task to be algorithm for load balance (#32021)

sollhui:
pick #32021


sollhui:
run buildall

PR #32638, opened 2024-03-21
[fix](routine-load) avoid routine load pause for check transaction status fail

sollhui:
## Proposed changes

routine load will pause in follow case:

1. routine load task visible and changed transaction status
2. user show task which cause task transaction status changed to visible
3. routineLoadTaskInfo.getTxnStatus() != TransactionStatus.COMMITTED check fail

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #37444, opened 2024-07-08
[fix](routine-load) avoid routine load pause for check transaction status fail (#32638)

sollhui:
pick #32638


sollhui:
run buildall

PR #35901, opened 2024-06-05
[fix](routine-load) dealing with the high watermark of Kafka may fallback

sollhui:
## Proposed changes

When Kafka roll upgrade, value of `high watermark` may briefly fallback.

PR #37454, opened 2024-07-08
[fix](routine-load) dealing with the high watermark of Kafka may fallback (#35901)

sollhui:
pick (#35901)


sollhui:
run buildall

PR #37464, opened 2024-07-08
[fix](memory) Fix Jemalloc Cache Memory Tracker

xinyiZzz:
## Proposed changes

Doris uses Jemalloc as default Allocator, Jemalloc Cache consists of two parts:
- Thread Cache, cache a specified number of Pages in Thread Cache.
- Dirty Page, memory Page that can be reused in all Arenas.

1. Metadata should not be counted as cache, this will cause memory GC to be delayed, leading to BE OOM.
2. Fix Jemalloc dirty page memory size, previous code used dirty page number * page size (4K on x86), which is much smaller than the actual memory. the fix is ​​the sum of dirty page memory of all size classes of extents.


xinyiZzz:
run buildall

be/src/util/mem_info.cpp
```diff
@@ -85,9 +86,9 @@ void MemInfo::refresh_allocator_mem() {
     // https://jemalloc.net/jemalloc.3.html
     // https://www.bookstack.cn/read/aliyun-rds-core/4a0cdf677f62feb3.md
     _s_allocator_cache_mem.store(get_je_all_arena_metrics("tcache_bytes") +
-                                         get_je_metrics("stats.metadata") +
                                          get_je_all_arena_metrics("pdirty") * get_page_size(),
                                  std::memory_order_relaxed);
+    _s_allocator_metadata_mem.store(get_je_metrics("stats.metadata"), std::memory_order_relaxed);
```
yiguolei, line 91: add comment here to explain why not count metadata cache to allocator cache mem size.

xinyiZzz: done

xinyiZzz:
run buildall

xinyiZzz:
run buildall

yiguolei (approved):
LGTM

HappenLee (approved):
LGTM

PR #37466, opened 2024-07-08
[fix](fe) Add check editlog size mechanism for backupJob (#35653)

swjtu-zhanglei:
* When creating a backupJob with huge of tables in a database, it can cause backupJob editlog size over 2GB and bdbje will throw exception because of ByteBuffer overflow

pick #35653

## Proposed changes

Issue Number: close #xxx


SWJTU-ZhangLei:
run buildall

PR #33825, opened 2024-04-18
[fix](schema change) fix schema change check does not calculate reader merged rows

luwei16:
### 1 bugfix
In the unique table, when there are multiple segments in a rowset and the same key exists in different segments, reading the rowset will merge these same keys. However, during schema change check, these merged keys are not calculated. May cause schema change check to fail
### 2 add config
Add configuration to ignore schema change verification results

## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #37467, opened 2024-07-08
[fix](schema change) fix schema change check does not calculate reader merged rows (#33825)

luwei16:
pick #33825


luwei16:
run buildall

PR #37469, opened 2024-07-08
[chore](dependencies)Remove Unnecessary Dependencies

CalvinKirs:
## Proposed changes
**devtools**: This dependency is only needed during local development, so we should set its scope to runtime.
**canal** :  The Spring dependency for Canal is primarily used for Pulsar's Bean replication. Since binlogload does not require any MQ dependencies, we can remove the corresponding Spring jar.  https://github.com/alibaba/canal/blob/master/client/src/main/java/com/alibaba/otter/canal/client/kafka/protocol/KafkaFlatMessage.java
**kafka** :Fe does not directly interact with Kafka; only UT uses Kafka libraries, so its dependency scope should be set to 'test'.

### DIFF
--- a/fe/fe-core/pom.xml
+++ b/fe/fe-core/pom.xml
@@ -93,6 +93,7 @@ under the License.
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-devtools</artifactId>
+            <scope>runtime</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


morningman (dismissed):
LGTM

CalvinKirs:
run buildall

morningman (approved):
LGTM

CalvinKirs:
run buildall

JNSimba (approved):

PR #37470, opened 2024-07-08
[fix](schema change) fix the defineName field is not the same when copying column (#34199)

luwei16:
pick #34199


luwei16:
run buildall

dataroaring:
run buildall

PR #37328, opened 2024-07-05
[feat](fe) Add a counter metric for recording large editlog write

swjtu-zhanglei:

PR #37474, opened 2024-07-08
[improve](fe) Add a counter metric for recording large editlog write (#37328)

swjtu-zhanglei:
pick #37328


SWJTU-ZhangLei:
run buildall

PR #32913, opened 2024-03-27
[fix](schemachange) schema change from not null to null

dataroaring:
1. Use equals instead of == for type comparing
2. null bitmap size is reiszed by size of ref column.

## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #37487, opened 2024-07-08
[fix](schema-change) Fix schema-change from non-null to null

Lchangliang:
https://github.com/apache/doris/pull/32913


Lchangliang:
run buildall

PR #35778, opened 2024-06-03
[fix](dynamic partition) fix dynamic partition thread met uncatch exception

yujun777:
dynamic partition thread get table's drop clause met an error due to dynamic_partition.start = -99999999

```
2024-05-23 10:40:32,072 ERROR (DynamicPartitionScheduler|53) [Daemon.run():118] daemon thread got exception. name: DynamicPartitionScheduler
org.apache.doris.nereids.exceptions.AnalysisException: date/datetime literal [+271768-09-11 00:00:00] is invalid
	at org.apache.doris.nereids.trees.expressions.literal.DateLiteral.normalize(DateLiteral.java:202) ~[doris-fe.jar:1.2-SNAPSHOT]
	at org.apache.doris.nereids.trees.expressions.literal.DateTimeLiteral.determineScale(DateTimeLiteral.java:107) ~[doris-fe.jar:1.2-SNAPSHOT]
	at org.apache.doris.nereids.types.DateTimeV2Type.forTypeFromString(DateTimeV2Type.java:91) ~[doris-fe.jar:1.2-SNAPSHOT]
	at org.apache.doris.nereids.trees.expressions.literal.DateTimeV2Literal.<init>(DateTimeV2Literal.java:38) ~[doris-fe.jar:1.2-SNAPSHOT]
	at org.apache.doris.catalog.PartitionKey.getDateTimeLiteral(PartitionKey.java:120) ~[doris-fe.jar:1.2-SNAPSHOT]
	at org.apache.doris.catalog.PartitionKey.createPartitionKey(PartitionKey.java:98) ~[doris-fe.jar:1.2-SNAPSHOT]
	at org.apache.doris.clone.DynamicPartitionScheduler.getDropPartitionClause(DynamicPartitionScheduler.java:431) ~[doris-fe.jar:1.2-SNAPSHOT]
	at org.apache.doris.clone.DynamicPartitionScheduler.executeDynamicPartition(DynamicPartitionScheduler.java:555) ~[doris-fe.jar:1.2-SNAPSHOT]
	at org.apache.doris.clone.DynamicPartitionScheduler.runAfterCatalogReady(DynamicPartitionScheduler.java:641) ~[doris-fe.jar:1.2-SNAPSHOT]
	at org.apache.doris.common.util.MasterDaemon.runOneCycle(MasterDaemon.java:58) ~[doris-fe.jar:1.2-SNAPSHOT]
[...]

PR #37488, opened 2024-07-08
[fix](create table) create table fail not write drop table editlog

yujun777:
Fix:  for creating table,  if create dynamic partition throw exception (pr #35778),  then create table will fail,  then drop table,  but it forget to write an editlog.


yujun777:
run buildall

deardeng (approved):
LGTM

yujun777:
run buildall

dataroaring (approved):
LGTM

PR #36672, opened 2024-06-21
[enhancement](compaction) adjust compaction concurrency based on compaction score and workload

luwei16:
1 Resolved the issue where the priority queue did not reserve slots for cumulative compaction.

2 When considering compaction task priorities, introduced metrics for CPU and memory usage rates. When the compaction score is low, and CPU or memory usage is high, reduce the number of compaction tasks generated and allocate CPU and memory resources to queries or load.

3 Integrated the logic of the priority queue and concurrency control together, removing the previous priority code.

PR #37491, opened 2024-07-08
[[enhancement](compaction) adjust compaction concurrency based on compaction score and workload

luwei16:
adjust compaction concurrency based on compaction score and workload #36672
fix null pointer when retrieving CPU load average #37171


luwei16:
run buildall

PR #37495, opened 2024-07-08
[fix](func)fix array_with_const with larger than max_array_size

amorynan:
## Proposed changes
when we use func array_with_const with large size there has not limit so, we will make be core like this
截屏2024-07-08 20 41 43

Issue Number: close #xxx

### DIFF
--- a/be/src/vec/functions/array/function_array_with_constant.cpp
+++ b/be/src/vec/functions/array/function_array_with_constant.cpp
@@ -91,9 +91,9 @@ public:
         array_sizes.reserve(input_rows_count);
         for (size_t i = 0; i < input_rows_count; ++i) {
             auto array_size = num->get_int(i);
-            if (UNLIKELY(array_size < 0)) {
-                return Status::RuntimeError("Array size can not be negative in function:" +
-                                            get_name());
+            if (UNLIKELY(array_size < 0) || UNLIKELY(array_size > max_array_size_as_field)) {
+                return Status::RuntimeError("Array size should in range(0, {}) in function: {}",
+                                            max_array_size_as_field, get_name());
             }
             offset += array_size;
             offsets.push_back(offset);
--- a/regression-test/suites/datatype_p0/nested_types/query/array_functions/test_array_functions_array_with_const.groovy
+++ b/regression-test/suites/datatype_p0/nested_types/query/array_functions/test_array_functions_array_with_const.groovy
@@ -16,6 +16,7 @@
 // under the License.
 
 suite("test_array_functions_array_with_const", "p0") {
+    sql "set enable_nereids_planner=false;"
     //array_with_constant
     qt_old_sql "SELECT 'array_with_constant';"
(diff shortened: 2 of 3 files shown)


amorynan:
run buildall

regression-test/suites/datatype_p0/nested_types/query/array_functions/test_array_functions_array_with_const.groovy
```diff
@@ -16,6 +16,7 @@
 // under the License.
 
 suite("test_array_functions_array_with_const", "p0") {
+    sql "set enable_nereids_planner=false;"
```
eldenmoon, line 19: why set to false? Is Nererids planner ok for this case?

amorynan: below in this groovy we have nereids query

eldenmoon (approved):
LGTM

regression-test/suites/datatype_p0/nested_types/query/array_functions/test_array_functions_array_with_const.groovy
```diff
@@ -31,7 +32,16 @@ suite("test_array_functions_array_with_const", "p0") {
                 SELECT array_with_constant(-231.37104, -138); 
                 """
     } catch (Exception ex) {
-        assertTrue(ex.getMessage().contains("Array size can not be negative in function:array_with_constant"))
+        assertTrue(ex.getMessage().contains("Array size should in range(0, 1000000) in function"))
+    }
+
+    // -- {server for large array}
+    try {
```
zclllyybb, line 39: don't use `try-catch` to do sql failure test. it doesn't work at all. if a sql didn't fail, the test will success. it's not as expected. please use `test{ sql exception }` instead

zclllyybb (changes requested):

be/src/vec/functions/array/function_array_with_constant.cpp
```diff
@@ -91,9 +91,9 @@ class FunctionArrayWithConstant : public IFunction {
         array_sizes.reserve(input_rows_count);
         for (size_t i = 0; i < input_rows_count; ++i) {
             auto array_size = num->get_int(i);
-            if (UNLIKELY(array_size < 0)) {
-                return Status::RuntimeError("Array size can not be negative in function:" +
-                                            get_name());
+            if (UNLIKELY(array_size < 0) || UNLIKELY(array_size > max_array_size_as_field)) {
+                return Status::RuntimeError("Array size should in range(0, {}) in function: {}",
```
xiaokang, line 95: add array_size actual value in error message for debug

xiaokang (approved):

be/src/vec/functions/array/function_array_with_constant.cpp
amorynan: array_size has already in error message

PR #37512, opened 2024-07-09
[bugfix](paimon)adding dependencies for `clang`

wuwenchi:
## Proposed changes

When paimon uses the `hms` type catalog and reads data in jni mode, it needs to use the `org.apache.commons.lang.StringUtils` class. 
(This problem is not tested in the pipeline environment because the pipeline environment automatically generates the `java-udf-case-jar-with-dependencies.jar` for testing, which contains the `lang` package.)


wuwenchi:
run buildall

morningman (approved):
LGTM

CalvinKirs (approved):

PR #33574, opened 2024-04-12
[fix](create table) Fix create table exception without cleaning the e…

deardeng:
…nvironment

## Proposed changes

Issue Number: close #xxx

修复建表过程中抛出异常后，db 内仍然保留table信息。
在修复前，打桩异常发现，建表流程中抛异常，还是能show出来

img_v3_029q_7061711e-d9f5-4ac5-a6c6-bde175a7513g

er0Nf6CQOm

还能show tablets 出来

img_v3_029q_ddd92989-cca1-4c2e-aa6b-cff1fd8be5dg

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #37516, opened 2024-07-09
[fix](create table)  fix create table fail cause meta leaky #37488, #33574

yujun777:
cherry-pick:  #37488,  #33574


yujun777:
run buildall

PR #37520, opened 2024-07-09
[opt](regression-test) More test case to trace problem of datetime v2 round

zhiqiang-hhhh:
We have some problem in other rep. The problem can not be reproduced on master, so refine the test case to trace problem.


zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
image

zhangstar333 (approved):

yiguolei (approved):

PR #37527, opened 2024-07-09
[fix](load) The NodeChannel should be canceled when failed to add block #37500

liaoxin01:
cherry pick from #37500


liaoxin01:
run buildall

PR #37531, opened 2024-07-09
[fix](publish) Fix publish failed because because "task" is null

mymeiyi:
## Proposed changes

```
2024-07-08 00:43:27,149 ERROR (PUBLISH_VERSION|33) [PublishVersionDaemon.runAfterCatalogReady():73] errors while publish version to all backends
java.lang.NullPointerException: Cannot invoke "org.apache.doris.task.PublishVersionTask.isFinished()" because "task" is null
at org.apache.doris.transaction.PublishVersionDaemon.lambda$tryFinishTxn$0(PublishVersionDaemon.java:163) ~[doris-fe.jar:1.2-SNAPSHOT]
at java.util.HashMap.forEach(HashMap.java:1421) ~[?:?]
at org.apache.doris.transaction.PublishVersionDaemon.tryFinishTxn(PublishVersionDaemon.java:160) ~[doris-fe.jar:1.2-SNAPSHOT]
at org.apache.doris.transaction.PublishVersionDaemon.publishVersion(PublishVersionDaemon.java:96) ~[doris-fe.jar:1.2-SNAPSHOT]
at org.apache.doris.transaction.PublishVersionDaemon.runAfterCatalogReady(PublishVersionDaemon.java:70) ~[doris-fe.jar:1.2-SNAPSHOT]
at org.apache.doris.common.util.MasterDaemon.runOneCycle(MasterDaemon.java:58) ~[doris-fe.jar:1.2-SNAPSHOT]
at org.apache.doris.common.util.Daemon.run(Daemon.java:116) ~[doris-fe.jar:1.2-SNAPSHOT]
```

1. When try finish one txn, catch the exception to make the failed txn does not block the other txns.
2. In the original way, when commit txn, add a <be_id, null publish task> to publish tasks, and then when publish txn, reset the null publish task to a new publish task. 
    This pr modify it to when commit txn, record the involved be ids, and then when publish txn, generate the publish tasks to all involved bes.
3. There is also a bug of `tableIdToTabletDeltaRows` in transaction state, it records all ready txn infos, because the variable scope is out of `for (TransactionState transactionState : readyTransactionStates)`

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/transaction/TransactionState.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/transaction/TransactionState.java
@@ -822,7 +822,7 @@ public class TransactionState implements Writable {
     public void pruneAfterVisible() {
         publishVersionTasks.clear();
         tableIdToTabletDeltaRows.clear();
-        // TODO if subTransactionStates can be cleared?
+        involvedBackends.clear();
     }
 
     public void setSchemaForPartialUpdate(OlapTable olapTable) {
--- a/fe/fe-core/src/main/java/org/apache/doris/transaction/TransactionState.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/transaction/TransactionState.java
@@ -389,7 +389,7 @@ public class TransactionState implements Writable {
     }
 
     public void addPublishVersionTask(Long backendId, PublishVersionTask task) {
-        if (this.subTxnIdToTableCommitInfo.isEmpty()) {
+        if (this.subTxnIds == null) {
             this.publishVersionTasks.put(backendId, Lists.newArrayList(task));
         } else {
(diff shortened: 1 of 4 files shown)


fe/fe-core/src/main/java/org/apache/doris/transaction/TransactionState.java
```diff
@@ -822,7 +822,7 @@ public String getErrMsg() {
     public void pruneAfterVisible() {
         publishVersionTasks.clear();
         tableIdToTabletDeltaRows.clear();
-        // TODO if subTransactionStates can be cleared?
+        involvedBackends.clear();
```
yujun777, line 825: add @SerializedName to involvedBackends ?

yujun777 (approved):
LGTM

dataroaring (approved):
LGTM

PR #37539, opened 2024-07-09
[fix](dynamic partition) drop partition exclude history_partition_num

yujun777:
FIX:   

When dropping dynamic partition,  PR #35778 will use math.max(start,  -history_partition_num) as the first partition,  but it may delete users' partitions if they specify both start and history_partition_num  inappropriately.  For safety reason,  revert this behavious changed,  only use start as the first partition when dropping partitions. 

For those who had specified a very small start value,   drop partitions will catch an exception ,   and stop dropping this table's partition and then record this error in dynamic info.  Users can   use command `SHOW DYNAMIC PARTITION TABLES FROM DBXXX` to know this error.  From this error, it will give user hint to modify start if they really specify a error start.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/clone/DynamicPartitionScheduler.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/clone/DynamicPartitionScheduler.java
@@ -467,10 +467,11 @@ public class DynamicPartitionScheduler extends MasterDaemon {
             // AnalysisException: keys.size is always equal to column.size, cannot reach this exception
             // IllegalArgumentException: lb is greater than ub
             String hint = "'dynamic_partition.start' = " + dynamicPartitionProperty.getStart()
-                    + ", maybe it's too small, can use alter table sql to increase it.";
+                    + ", maybe it's too small, can use alter table sql to increase it. ";
             LOG.warn("Error in gen reservePartitionKeyRange. db: {}, table: {}. {}",
                     db.getFullName(), olapTable.getName(), hint, e);
-            recordDropPartitionFailedMsg(db.getFullName(), olapTable.getName(), hint + e.getMessage(), olapTable.getId());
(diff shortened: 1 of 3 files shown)


yujun777:
run buildall

zclllyybb:
need re-format FE

yujun777:
run buildall

yujun777:
run external

yujun777:
run p1

yujun777:
run cloud_p1

yujun777:
run p0

deardeng (approved):
LGTM

yujun777:
run p1

yujun777:
run cloud_p1

yujun777:
run external

dataroaring (dismissed):
LGTM

dataroaring:
run buildall

dataroaring:
run buildall

dataroaring (approved):
LGTM

fe/fe-core/src/main/java/org/apache/doris/clone/DynamicPartitionScheduler.java
```diff
@@ -585,10 +590,12 @@ public void executeDynamicPartition(Collection<Pair<Long, Long>> dynamicPartitio
                     addPartitionClauses = getAddPartitionClause(db, olapTable, partitionColumn, partitionFormat,
                             executeFirstTime);
                 }
+                clearDropPartitionFailedMsg(olapTable.getId());
```
zxealous, line 593: why call clearDropPartitionFailedMsg()  here?

zxealous: This will cause the status to remain NORMAL even if adding partitions fails.

PR #37546, opened 2024-07-09
[fix](publish) Pick Fix publish failed because because task is null

mymeiyi:
## Proposed changes

Pick https://github.com/apache/doris/pull/37531

This pr catch the exception to make the failed txn does not block the other txns.


mymeiyi:
run buildall

mymeiyi:
run p0

PR #37547, opened 2024-07-09
[enhance](HDFS) Add one config to control hdfs memory limiter

ByteYue:
Previously when loading into hdfs cluster, we would try to do memory control.  Though it's one approximate memory recording and limit control solution, it's not so accuracy. So we add one config to control whether using the memory limit controller or not.


ByteYue:
run buildall

ByteYue:
run buildall

gavinchou (approved):
Add description: behavior and when and how to use this config

dataroaring (approved):
LGTM

PR #37549, opened 2024-07-09
[fix](auth)fix fe can not restart when replay create row policy log for 2.0

zddr:
The fixed issue is the same as https://github.com/apache/doris/pull/37342

There are significant differences in metadata, so instead of directly picking, errors are ignored to achieve the effect of not affecting FE startup


zddr:
run buildall

PR #37560, opened 2024-07-09
[Fix](branch-2.1) Fix partial update regression test

Yukang-Lian:
## Proposed changes

Issue Number: close #xxx

Version 2.1 does not support partial update for variant type data, so we delete this case.


Yukang-Lian:
run buildall

zhannngchen:
plz add description to explain why to delete this case

PR #32243, opened 2024-03-14
[improve](load) try lock 30ms to get base_migration_lock in rowset builder

kaijchen:
## Proposed changes

Try lock for 30ms to get base_migration_lock in rowset builder, instead of failing immediately.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #37576, opened 2024-07-10
[improve](load) try lock 30ms to get base_migration_lock in delta writer (#32243)

kaijchen:
cherry-pick #32243


kaijchen:
run buildall

PR #34897, opened 2024-05-15
[fix](load) skip sending cancel rpc if VNodeChannel is not inited

kaijchen:
## Proposed changes

Fix the following coredump

```cpp
Core was generated by `/mnt/hdd01/DorisCloudS3Stress/cluster0/be/lib/doris_be'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  pthread_sigmask (how=2, newmask=<optimized out>, oldmask=0x0) at pthread_sigmask.c:48
48      pthread_sigmask.c: No such file or directory.
[Current thread is 1 (Thread 0x7fdbf91f1700 (LWP 3649251))]
(gdb) bt
#0  pthread_sigmask (how=2, newmask=<optimized out>, oldmask=0x0) at pthread_sigmask.c:48
#1  0x00007ffa17acd71e in PosixSignals::chained_handler(int, siginfo*, void*) [clone .part.0] ()
   from /usr/lib/jvm/java-17-openjdk-amd64/lib/server/libjvm.so
#2  0x00007ffa17ace206 in JVM_handle_linux_signal () from /usr/lib/jvm/java-17-openjdk-amd64/lib/server/libjvm.so
#3  <signal handler called>
#4  doris::vectorized::VNodeChannel::cancel (this=this@entry=0x7ff4ef107190, cancel_msg=...)
    at /home/zcp/repo_center/doris_branch-3.0/doris/be/src/vec/sink/writer/vtablet_writer.cpp:912
#5  0x00005646777a937b in doris::vectorized::VTabletWriter::_cancel_all_channel(doris::Status)::$_0::operator()(std::shared_ptr<doris::vectorized::VNodeChannel> const&) const (this=<optimized out>, ch=...)
    at /home/zcp/repo_center/doris_branch-3.0/doris/be/src/vec/sink/writer/vtablet_writer.cpp:1388
#6  std::__invoke_impl<void, doris::vectorized::VTabletWriter::_cancel_all_channel(doris::Status)::$_0&, std::shared_ptr<doris::vectorized::VNodeChannel> const&>(std::__invoke_other, doris::vectorized::VTabletWriter::_cancel_all_channel(doris::Status)::$_0&, std::shared_ptr<doris::vectorized::VNodeChannel> const&) (__f=..., __args=...) at /var/local/ldb_toolchain/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/invoke.h:61

...

(gdb) f 4
[...]

PR #37577, opened 2024-07-10
[fix](load) skip sending cancel rpc if VNodeChannel is not inited (#34897)

kaijchen:
cherry-pick #34897


kaijchen:
run buildall

PR #37439, opened 2024-07-08
[fix](partition) Fix test partition regression case unstabitily

deardeng:
## Proposed changes

Issue Number: close #xxx

PR #37583, opened 2024-07-10
[fix](partition) Fix test partition regression case unstabitily #37439

deardeng:
cherry pick from #37439


deardeng:
run buildall

PR #37587, opened 2024-07-10
[fix](auto bucket) Fix hit not support alter estimate_partition_size #33670

deardeng:
cherry pick from #33670


deardeng:
run buildall

PR #32991, opened 2024-03-28
[Enhancement](merge-on-write) Support  dynamic delete bitmap cache

Yukang-Lian:
The default delete bitmap cache is set to 100MB, which can be insufficient and cause performance issues when the amount of user data is large. To mitigate the problem of an inadequate cache, we will take the larger of 5% of the total memory and 100MB as the delete bitmap cache size.

## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #37602, opened 2024-07-10
[cherry-pick](branch-2.0)  Pick "[Enhancement](merge-on-write) Support  dynamic delete bitmap cache (#32991)"

Yukang-Lian:
The default delete bitmap cache is set to 100MB, which can be insufficient and cause performance issues when the amount of user data is large. To mitigate the problem of an inadequate cache, we will take the larger of 5% of the total memory and 100MB as the delete bitmap cache size.

pick #32991

## Proposed changes

Issue Number: close #xxx


Yukang-Lian:
run buildall

dataroaring (approved):
LGTM

PR #37622, opened 2024-07-10
[fix](test) fix case conflict between test_tvf_based_broker_load and test_broker_load

liaoxin01:
## Proposed changes

Issue Number: close #xxx

PR #37631, opened 2024-07-11
[fix](test) fix case conflict between test_tvf_based_broker_load and test_broker_load #37622

liaoxin01:
cherry pick from #37622


liaoxin01:
run buildall

PR #35929, opened 2024-06-05
[Refactor]Refacor query queue

wangbo:
## Proposed changes
Use CompletableFuture to refactor query queue to simplify code for better readability.

PR #37632, opened 2024-07-11
[bugfix](wg) should not remove token directly, there maybe concurrency problem

yiguolei:
remove token api will not check token state is ready to run. If the token is set to ready to run before call remove token, then the currentRunningQueryNum is not decrement. 

imported by PR https://github.com/apache/doris/pull/35929/files


yiguolei:
run buildall

PR #37636, opened 2024-07-11
[fix](split) remove retry when fetch split batch failed

AshinGau:
## Proposed changes

We need to remove the retry logic for failed to fetch split batch. Originally, this was implemented to handle cases where the cached client connection might have been lost and needed to be reestablished. However, this retry mechanism can lead to data loss. For instance, if a batch of data has already been sent, retrying can cause this batch to be lost without the receiver being aware of it.


AshinGau:
run buildall

wuwenchi (approved):

morningman (approved):
LGTM

PR #37649, opened 2024-07-11
[Update](inverted index) Add column name to debug point for "no need to read data" optimization

airborne12:
## Proposed changes

We need to verify the false negative situation when applying the "no need to read data" optimization in inverted index predicates. In this PR, we add the column name to the debug point parameters, allowing us to check if the optimization is functioning correctly.


qidaye (approved):
LGTM

zzzxl1993 (approved):
LGTM

airborne12:
run buildall

PR #37662, opened 2024-07-11
[fix](query cancel) Fix query is cancelled when it comes from follower FE

zhiqiang-hhhh:
In some rear cases, the rpc port of follower FE is not updated in time, the value of rpc port of this follower in heartbeat will be 0, but actually it is still running. Query from the follower FE will be cancelled by be until rpc port is updated correctly on BE. 

This pr fixes the problem on BE by detecting above situation, and avoid cancel query in this situation.

### DIFF
--- a/be/src/runtime/fragment_mgr.cpp
+++ b/be/src/runtime/fragment_mgr.cpp
@@ -39,6 +39,7 @@
 #include <thrift/transport/TTransportException.h>
 #include <unistd.h>
 
+#include <algorithm>
 #include <atomic>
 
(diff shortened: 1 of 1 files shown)


zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run external

yiguolei (approved):
LGTM

xinyiZzz (approved):
LGTM

PR #37545, opened 2024-07-09
[fix](statistics)Fix column cached stats size bug.

Jibing-Li:
Fix column cached stats average size calculate bug. Average size is double type, when calculating, it is calculated by totalSize/rowCount. We need to cast totalSize to double, otherwise the result is not accurate, only keep the integer part.

PR #37666, opened 2024-07-11
[fix](statistics)Fix column cached stats size bug. (#37545)

Jibing-Li:
backport: https://github.com/apache/doris/pull/37545


Jibing-Li:
run buildall

PR #31689, opened 2024-03-01
[feature](pipelineX) support paritition tablet sink shuffle

zhangstar333:
## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #37681, opened 2024-07-11
[fix](sleep) sleep with character const make be crash

### LINKED ISSUE #37327: [Bug] sleep function with character constant(such as 'a', '1.1') leads to be crash
### Search before asking

### Version

2.1

### What's Wrong?

`select sleep('1.1')` leads to be crash

### What You Expected?

just as 2.0version, be will not crash

### How to Reproduce?

Just execute `select sleep('1.1')`
图片

### Anything Else?

_No response_

### Code of Conduct

cambyzju:
## Proposed changes

Issue Number: close https://github.com/apache/doris/issues/37327

Come from: #31689

After fix be will not crash any more.

```
mysql> select cast('1.1' as int);
+--------------------+
| cast('1.1' as INT) |
+--------------------+
|               NULL |
+--------------------+
1 row in set (0.04 sec)

mysql> select sleep('1.1');
+---------------------------+
| sleep(cast('1.1' as INT)) |
+---------------------------+
|                      NULL |
+---------------------------+
1 row in set (0.02 sec)

```

### DIFF
--- a/be/src/vec/functions/function_utility.cpp
+++ b/be/src/vec/functions/function_utility.cpp
@@ -62,7 +62,7 @@ public:
     size_t get_number_of_arguments() const override { return 1; }
 
     DataTypePtr get_return_type_impl(const DataTypes& arguments) const override {
-        if (arguments[0].get()->is_nullable()) {
+        if (arguments[0]->is_nullable()) {
             return make_nullable(std::make_shared<DataTypeUInt8>());
         }
         return std::make_shared<DataTypeUInt8>();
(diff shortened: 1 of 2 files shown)


cambyzju:
run buildall

zhangstar333 (dismissed):

be/src/vec/functions/function_utility.cpp
```diff
@@ -74,19 +74,12 @@ class FunctionSleep : public IFunction {
```
zhiqiang-hhhh, line 74: Could you please add some comment to explain why we not using default_impl for const? 
"Sleep function should not be executed during open stage, this will makes fargment preparint too long, so we do not use default impl."

cambyzju:
run buildall

be/src/vec/functions/function_utility.cpp
```diff
@@ -62,31 +62,26 @@ class FunctionSleep : public IFunction {
...
     bool use_default_implementation_for_nulls() const override { return false; }
 
+    // Sleep function should not be executed during open stage, this will makes fragment prepare
+    // waiting too long, so we do not use default impl.
     bool use_default_implementation_for_constants() const override { return false; }
 
     Status execute_impl(FunctionContext* context, Block& block, const ColumnNumbers& arguments,
                         size_t result, size_t input_rows_count) const override {
-        ColumnPtr& argument_column = block.get_by_position(arguments[0]).column;
+        const auto& argument_column =
```
zclllyybb, line 79: use `unpack_if_const` may be better

yiguolei (approved):
LGTM

be/src/vec/functions/function_utility.cpp
cambyzju: there are already lots of convert_to_full_column_if_const used, maybe next time, we refractor all these used is better.

lide-reed (approved):
LGTM

zhiqiang-hhhh (approved):

PR #37688, opened 2024-07-11
[fix](compile) replace shared_ptr::unique() with std::shared_ptr::use_count() == 1

Hastyshell:
## Proposed changes

`std::shared_ptr::unique` is supported until cpp20. It was disabled by default since llvm-18. [see](https://releases.llvm.org/18.1.0/projects/libcxx/docs/ReleaseNotes/18.html#id4).

Therefore, replace it with std::shared_ptr::use_count() == 1 for better compability. (Or we should work with _LIBCPP_ENABLE_CXX20_REMOVED_SHARED_PTR_UNIQUE which is ungraceful.)


TangSiyang2001:
run buildall

gavinchou (approved):

yiguolei (approved):
LGTM

PR #37151, opened 2024-07-02
[Fix](paritial update) Fix the case of partial update failing in cloud mode

Yukang-Lian:
## Proposed changes

Problem: The `test_new_partial_update` case fails to run in cloud mode, but it passes in local mode. 

Reason: In PR #35917, we introduced a new table attribute `enable_mow_light_delete`. When executing schema changes with `alter table set xxx` statements, the local and cloud modes process the logic differently. The cloud mode has its unique processing logic, which was not addressed in the mentioned PR, leading to failures in the cloud environment.

Solution: To resolve the issue, we need to complete the missing schema change logic for the cloud mode. Once this is implemented, the problem should be resolved.

PR #35917, opened 2024-06-05
[Fix](delete command) Mark delete sign when do delete command in MoW table

Yukang-Lian:
## Proposed changes

close #34551 

Problem: As shown in the issue above, if a key deleted by a delete statement is written to by updating only certain columns, the data will not display correctly.

Reason: The delete statement deletes the data by writing a delete predicate, which is stored in the rowset meta and applied during data retrieval to filter the data. However, partial column updates do not consider the effect of the delete predicate when reading the original data. The imported key should be considered as a new key (since it has already been deleted), but it is actually treated as an old key. Therefore, only some columns are updated, leading to incorrect results.

Solution: Consider the delete predicate during partial column updates, but this method will result in reading more columns, as shown in #35766. Thus, in this PR, we change the delete operation in the mow table from writing a delete predicate to writing a delete sign, which effectively resolves the issue.

PR #37689, opened 2024-07-11
[Enhancement](delete) Modify some p2 delete cases' property

Yukang-Lian:
## Proposed changes

Issue Number: close #xxx

In #35917 and #37151, we changed MOW table default delete command from delete predicate to delete sign. It makes sure the correctness during partial update but leads to slowdowns. Actually, if there is no partial update, delete predicate will not lead to data fault. Delete data by delete predicate or delete sign can be controlled by a table property "enable_light_delete". If "enable_light_delete=true", we execute delete command by delete predicate. Otherwise, we execute delete command by delete sign.

In p2 cases, there are lots of cases with large data need to delete and do not perform partial column update operations. Therefore, in order to make it faster, we change some cases default create table clause.


Yukang-Lian:
run buildall

Yukang-Lian:
run buildall

dataroaring (dismissed):
LGTM

bobhan1 (approved):

Yukang-Lian:
run buildall

gavinchou (approved):

PR #37697, opened 2024-07-12
[fix](jni_connector) fix that be core when jni_connector open fails

BePPPower:
## Proposed changes

Here are two problems:
1. `appendDataTimeNs` will be `NULL` if TrinoConnector open fails before call to `parseRequiredTypes`
2. It does not catch exception in method `get_statistics()` and this exception is not caught until `close()` method. Therefore, it will cause BE core dump in `JniConnector::close()` at `LOG(FATAL)`


BePPPower:
run buildall

BePPPower:
run buildall

morningman (approved):
LGTM

hello-stephen (approved):

PR #37701, opened 2024-07-12
[improvement](mv) Catch exception when init sync materialized view context

seawinde:
## Proposed changes
Init sync materialized view context happend after analyze, If fail would cause query or insert into fail.
This is inacceptable, so catch the exception and recored the warn log.


seawinde:
run buildall

morrySnow (approved):

starocean999 (approved):

PR #37702, opened 2024-07-12
[fix](hudi) return empty if there is no commit implemented

AshinGau:
## Proposed changes

Return empty set when there is no commit implemented, instead of throwing error:
```
HoodieException("No instants to incrementally pull")
```


AshinGau:
run buildall

morningman (approved):
LGTM

the test case will be added later

wuwenchi (approved):

PR #37747, opened 2024-07-13
[fix](inverted index) Error parsing NULL_LITERAL

zzzxl1993:
## Proposed changes

1. NULL_LITERAL and other LITERAL will enter the else process incorrectly
2. SELECT * FROM table_name WHERE column_name NOT IN (1, 2, NULL); The NOT IN operator returns an empty result set when the collection contains NULL.


xiaokang:
run buildall

zzzxl1993:
run buildall

PR #37568, opened 2024-07-09
[fix](dockerfile) Switch repos to point to to vault.centos.org because CentOS 7 is EOL

morningman:
Fix `Could not resolve host: mirrorlist.centos.org; Unknown error`

PR #37763, opened 2024-07-14
[fix](dockerfile) Switch repos to point to to vault.centos.org because CentOS 7 is EOL (#37568)

morningman:
bp #37568


morningman:
run buildall

PR #37707, opened 2024-07-12
[fix](query cancel) Fix query is cancelled when it comes from follower FE #37662

zhiqiang-hhhh:
cherry pick from #37662

PR #37772, opened 2024-07-15
[FIX] should check fe host standing when coordinator is not found. 

zhiqiang-hhhh:
fix https://github.com/apache/doris/pull/37707


zhiqiang-hhhh:
run buildall

PR #37775, opened 2024-07-15
[fix](sleep) sleep with character const make be crash (#37681)

cambyzju:
cherry-pick #37681 to branch-2.1


cambyzju:
run buildall

PR #37788, opened 2024-07-15
[fix](nereids) fix aggr node colocate flag which local shuffle depending on

xzj7019:
Local shuffle depends on the operator node's isColocate flag, to decide the shuffle columns.
Current aggr node's  isColocate true setting is too strict to support shuffle-colocate scenarios. 
This pr releases the unnecessary restriction (only support table scan) and sets the flag true expectly.

PR #37789, opened 2024-07-15
[fix](nereids) fix aggr node colocate flag local shuffle depending on 

xzj7019:
## Proposed changes

pick from https://github.com/apache/doris/pull/37788


xzj7019:
run buildall

PR #37643, opened 2024-07-11
[Bug](join) fix broadcast join running when hash table build not finished

zhangstar333:
## Proposed changes
before when PipelineTask close, will be set sink operator always ready.
but not all sink could be running normal, like some instance of join which not build hash table,
it's need to wait until others build hash table finished and then shared from it.

```
F20240710 17:29:09.628299 221449 hashjoin_build_sink.cpp:582] Check failed: _shared_hash_table_context->signaled
 0# doris::signal::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*) at /mnt/disk2/zhangsida/doris/be/src/common/signal_handler.h:421
 1# 0x00007FEF9BF64B50 in /lib64/libc.so.6
 2# gsignal in /lib64/libc.so.6
 3# __GI_abort in /lib64/libc.so.6
 4# 0x0000559C8BD8BE8D in /mnt/disk2/zhangsida/doris/output/be/lib/doris_be
 5# 0x0000559C8BD7E52A in /mnt/disk2/zhangsida/doris/output/be/lib/doris_be
 6# google::LogMessage::SendToLog() in /mnt/disk2/zhangsida/doris/output/be/lib/doris_be
 7# google::LogMessage::Flush() in /mnt/disk2/zhangsida/doris/output/be/lib/doris_be
 8# google::LogMessageFatal::~LogMessageFatal() in /mnt/disk2/zhangsida/doris/output/be/lib/doris_be
 9# doris::pipeline::HashJoinBuildSinkOperatorX::sink(doris::RuntimeState*, doris::vectorized::Block*, bool) at /mnt/disk2/zhangsida/doris/be/src/pipeline/exec/hashjoin_build_sink.cpp:582
10# doris::pipeline::PipelineTask::execute(bool*)::$_1::operator()() const at /mnt/disk2/zhangsida/doris/be/src/pipeline/pipeline_task.cpp:361
11# doris::pipeline::PipelineTask::execute(bool*) at /mnt/disk2/zhangsida/doris/be/src/pipeline/pipeline_task.cpp:364
12# doris::pipeline::TaskScheduler::_do_work(unsigned long) at /mnt/disk2/zhangsida/doris/be/src/pipeline/task_scheduler.cpp:138
13# doris::pipeline::TaskScheduler::start()::$_0::operator()() const at /mnt/disk2/zhangsida/doris/be/src/pipeline/task_scheduler.cpp:64

```

PR #37792, opened 2024-07-15
[Bug](join) fix broadcast join running when hash table build not finished

zhangstar333:
in pr https://github.com/apache/doris/pull/37643
wants fix the bug of join which not build hash table, but running early and not wait the finished signal.
But this may be a normal phenomenon, as it should allows all sinks to run when the source operator have closed,
so here return eof status directly when signal == false.

## Proposed changes

Issue Number: close #xxx


zhangstar333:
run buildall

yiguolei (approved):
LGTM

Gabriel39 (approved):

PR #37349, opened 2024-07-05
[fix](function) fix core when input not null array in foreach function

Mryange:
## Proposed changes

error code
```C++
return creator_without_type::create<AggregateFunctionForEach>(transform_arguments, true,
                                                                      nested_function);
```
"transform_arguments is an internal type of array. All internal types of the array are null, so an array that is not null was mistakenly treated as a null array."

core
```
INFO: java_cmd /mnt/disk2/yanxuecheng/java17/bin/java
INFO: jdk_version 17
start BE in local mode
F20240705 15:45:48.466758 3587508 assert_cast.h:45] Bad cast from type:doris::vectorized::ColumnArray* to doris::vectorized::ColumnNullable const*
*** Check failure stack trace: ***
    @     0x5612b1cf86a6  google::LogMessage::SendToLog()
    @     0x5612b1cf50f0  google::LogMessage::Flush()
    @     0x5612b1cf8ee9  google::LogMessageFatal::~LogMessageFatal()
    @     0x5612a3373e74  assert_cast<>()
    @     0x5612a778e60d  doris::vectorized::AggregateFunctionNullUnaryInline<>::add_batch_single_place()
    @     0x5612b0d0dba1  doris::vectorized::AggFnEvaluator::execute_single_add()
    @     0x5612b0c62b52  doris::pipeline::AggSinkLocalState::_execute_without_key()
    @     0x5612b0d0a90c  doris::pipeline::AggSinkLocalState::Executor<>::execute()
    @     0x5612b0c653df  doris::pipeline::AggSinkOperatorX::sink()
    @     0x5612b1bf27c8  doris::pipeline::PipelineTask::execute()::$_1::operator()()
    @     0x5612b1bf15c6  doris::pipeline::PipelineTask::execute()
    @     0x5612b1c0e8e7  doris::pipeline::TaskScheduler::_do_work()
    @     0x5612b1c0f72c  doris::pipeline::TaskScheduler::start()::$_0::operator()()
    @     0x5612b1c0f705  std::__invoke_impl<>()
    @     0x5612b1c0f6b5  _ZSt10__invoke_rIvRZN5doris8pipeline13TaskScheduler5startEvE3$_0JEENSt9enable_ifIX16is_invocable_r_vIT_T0_DpT1_EES6_E4typeEOS7_DpOS8_
    @     0x5612b1c0f5bd  std::_Function_handler<>::_M_invoke()
    @     0x5612a1400905  std::function<>::operator()()
[...]

PR #37798, opened 2024-07-15
[fix](function) fix core when input not null array in foreach functio…

Mryange:
## Proposed changes
https://github.com/apache/doris/pull/37349
error code
```C++
return creator_without_type::create<AggregateFunctionForEach>(transform_arguments, true,
                                                                      nested_function);
```
"transform_arguments is an internal type of array. All internal types of the array are null, so an array that is not null was mistakenly treated as a null array."

## Proposed changes

Issue Number: close #xxx


Mryange:
run buildall

PR #37071, opened 2024-07-01
[fix](routine-load) fix auto resume invalid when FE leader change

sollhui:
## Proposed changes

We meet routine load pause and never be auto resume even if it meet the conditions.
```
                  Id: 134305
                Name: lineitem_balance_dup_persistent_weekly_persistent_flow_weekly
          CreateTime: 2024-06-27 19:54:13
           PauseTime: 2024-06-28 23:02:46
             EndTime: NULL
              DbName: regression_test_stress_load_long_duration_load
           TableName: lineitem_balance_dup_persistent_weekly
        IsMultiTable: false
               State: PAUSED
      DataSourceType: KAFKA
      CurrentTaskNum: 0
       JobProperties: {"max_batch_rows":"550000","timezone":"Asia/Shanghai","send_batch_parallelism":"1","load_to_single_tablet":"false","column_separator":"','","line_delimiter":"\n","current_concurrent_number":"0","delete":"*","partial_columns":"false","merge_type":"APPEND","exec_mem_limit":"2147483648","strict_mode":"false","jsonpaths":"","max_batch_interval":"10","max_batch_size":"409715200","fuzzy_parse":"false","partitions":"*","columnToColumnExpr":"","whereExpr":"*","desired_concurrent_number":"100","precedingFilter":"*","format":"csv","max_error_number":"0","max_filter_ratio":"1.0","json_root":"","strip_outer_array":"false","num_as_string":"false"}
DataSourceProperties: {"topic":"test-topic-persistent-weekly-new","currentKafkaPartitions":"","brokerList":"xxx"}
    CustomProperties: {"kafka_default_offsets":"OFFSET_BEGINNING","group.id":"test-consumer-group","client.id":"test-client-id"}
           Statistic: {"receivedBytes":2234836231654,"runningTxns":[],"errorRows":0,"committedTaskNum":1019074,"loadedRows":11693905636,"loadRowsRate":119675,"abortedTaskNum":13556,"errorRowsAfterResumed":0,"totalRows":11693905636,"unselectedRows":0,"receivedBytesRate":22871277,"taskExecuteTimeMs":97713660}
[...]

PR #37804, opened 2024-07-15
Revert "[fix](routine-load) fix auto resume invalid when FE leader change (#37071)"

sollhui:
revert (#37071)

It will only call the `write` method to write the edit log when creating the routine load job, but will not use this method when change job state to pause, but use the logic:
```
Env.getCurrentEnv().getEditLog().logOpRoutineLoadJob(new RoutineLoadOperation(id, jobState));
```
Therefore, using the logic:
```
@SerializedName("pr")
protected ErrorReason pauseReason;
```
only persist pauseReason when create job, rather than pause job, which means `pauseReason` will not be persist when it is assigned due to pause, causing auto resume still invalid if FE leader change when job pause.


sollhui:
run buildall

dataroaring (approved):
LGTM

liaoxin01 (approved):

PR #37333, opened 2024-07-05
[Feature](rpc) support set brpc_idle_timeout_sec and enable thrift socket keep alive

BiteTheDDDDt:
## Proposed changes
In some special circumstances (such as network instability), the brpc server will always retain many tcp connections.

PR #37808, opened 2024-07-15
[Feature](rpc) support set brpc_idle_timeout_sec and enable thrift so…

BiteTheDDDDt:
pick from #37333


BiteTheDDDDt:
run buildall

PR #37812, opened 2024-07-15
[improve](test)Using Awaitility instead of Thread.sleep

CalvinKirs:
## Proposed changes

Awaitility: Waits until a specific condition is met, which makes the tests more reliable and less prone to intermittent failures. It ensures that the test only proceeds when the expected condition is true.
Thread.sleep: Introduces a fixed delay, which can lead to flaky tests. If the condition is met sooner than the sleep time, the test unnecessarily waits. Conversely, if the condition is met after the sleep time, the test will fail.

### DIFF
--- a/regression-test/framework/pom.xml
+++ b/regression-test/framework/pom.xml
@@ -401,5 +401,13 @@ under the License.
             <artifactId>netty-all</artifactId>
             <version>4.1.104.Final</version>
         </dependency>
+        <dependency>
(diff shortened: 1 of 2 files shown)


CalvinKirs:
run buildall

morningman (approved):
LGTM

dataroaring (approved):
LGTM

gavinchou (approved):

PR #37815, opened 2024-07-15
[test](case) fix agg_nullable test case get wrong result

zhangstar333:
## Proposed changes
seems this PR https://github.com/apache/doris/pull/37340 change the output of character.
now the explain sql of array character have the num of 255.


zhangstar333:
have fixed in PR: https://github.com/apache/doris/pull/37797

PR #37276, opened 2024-07-04
[case](paimon/iceberg)move cases from p2 to p0

wuwenchi:
## Proposed changes

move cases from p2 to p0

PR #37818, opened 2024-07-15
[case](paimon/iceberg)move cases from p2 to p0 for 2.1 (#37276)

wuwenchi:
move cases from p2 to p0

bp: #37276


wuwenchi:
run buildall

PR #37820, opened 2024-07-15
[fix](auth)fix fe can not restart when replay create row policy log (…

zddr:
pick: https://github.com/apache/doris/pull/37342


zddr:
run buildall

PR #37822, opened 2024-07-15
[fix](mtmv) use isManagedTable instead of check table type (#34287)

zddr:
pick: https://github.com/apache/doris/pull/34287


zddr:
run buildall

PR #37842, opened 2024-07-15
[fix](inverted index) Error parsing NULL_LITERAL

zzzxl1993:
## Proposed changes

1. NULL_LITERAL and other LITERAL will enter the else process incorrectly
2. SELECT * FROM table_name WHERE column_name NOT IN (1, 2, NULL); The NOT IN operator returns an empty result set when the collection contains NULL.


zzzxl1993:
run buildall

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -2811,7 +2812,8 @@ void SegmentIterator::_calculate_pred_in_remaining_conjunct_root(
     } else if (_is_literal_node(node_type)) {
         auto v_literal_expr = static_cast<const doris::vectorized::VLiteral*>(expr.get());
         _column_predicate_info->query_values.insert(v_literal_expr->value());
-    } else {
+    } else if (node_type == TExprNodeType::BINARY_PRED || node_type == TExprNodeType::MATCH_PRED ||
```
airborne12, line 2815: Remember to refactor in order to unify all the code to the same behavior.

airborne12 (approved):
LGTM

xiaokang (approved):
LGTM

PR #37848, opened 2024-07-15
[fix](cloud) provide a conf to enable/disable streamload commit on be

freemandealer:
Add BE config `enable_stream_load_commit_txn_on_be` default false

When to enable commit on be?
Bypassing FE during commit is good for load performance, especially when doing frequent streamloading.

When to disable commit on be?
The current implementation of the bypass will mess up the statistics, thus the query plan optimizer. So disable it when you encountering any troubles. Such problem will be fixed in future work.


freemandealer:
run buildall

gavinchou (approved):

freemandealer:
run external

PR #36802, opened 2024-06-25
[fix](delete) fix the error message for valid decimal data

felixwluo:
## Proposed changes

Issue Number: close #36775

PR #37854, opened 2024-07-15
[fix](delete) fixed special decimal values

felixwluo:
## Proposed changes

Related PR：#36802
`what this PR fixes is to remove values similar to 0.1234567`


felixwluo:
run buildall

PR #37863, opened 2024-07-16
[fix](Nereids) fix fe fold constant failed when using like function

LiBinfeng-01:
cherry-pick: #37616


LiBinfeng-01:
run buildall

PR #37865, opened 2024-07-16
[fix](regression) fix mow flaky case for cloud

zhannngchen:
## Proposed changes

Issue Number: close #xxx

the case `test_cloud_mow_delete_bitmap_calc_timeout` is flaky, it failed due to the following exception
```
Exception:
java.sql.SQLException: errCode = 2, detailMessage = (127.0.0.1)[INTERNAL_ERROR]failed to initialize storage reader. tablet=532938472811, res=[INTERNAL_ERROR]{} does not need to read data
```
but the exception is injected by other case, I suspect these cases don't clear the debug points for some unexpected reason.

to fix the flaky case, add a clear command at the first


zhannngchen:
run buildall

dataroaring (approved):
LGTM

bobhan1 (approved):

PR #37870, opened 2024-07-16
[Fix](sql function) memory overflow to the left of string address when do_money_format has small negative value #36226

zhiqiang-hhhh:
cherry pick from #36226


zhiqiang-hhhh:
run buildall

yiguolei:
run buildall

PR #37801, opened 2024-07-15
[fix](delete) Incorrect precision detection for the decimal type in condition.​

mrhhsg:
## Proposed changes

For precision like Decimal(7,7),  the value "0.1234567" should be valid(the integer part is 0).

PR #37904, opened 2024-07-16
[fix](delete) Incorrect precision detection for the decimal type in condition.​ (#37801)

mrhhsg:
## Proposed changes

pick #37801

For precision like Decimal(7,7), the value "0.1234567" should be valid(the integer part is 0).

## Proposed changes

Issue Number: close #xxx


mrhhsg:
run buildall

PR #37915, opened 2024-07-16
[enhancement](nereids) speedup sql cache with use variable as partition predicate

924060929:
## Proposed changes

follow up #37090

support reuse sql cache when use variable as partition predicate and variable change:
```sql
set @dt='2024-07-16';
-- create cache 1
select * from tbl where dt = @dt;

set @dt='2024-07-17';
-- create cache 2, will not invalidate cache 1
select * from tbl where dt = @dt;

set @dt='2024-07-16';
-- reuse cache 1
select * from tbl where dt = @dt;
```


924060929:
run buildall

morrySnow (approved):

feiniaofeiafei (approved):

PR #37928, opened 2024-07-16
[fix](regression) fix some txn_insert and group_commit case

mymeiyi:
## Proposed changes

1. the `TLoadTxnCommitRequest` used for multi table load should sort tables in table id orders, because the publish should get table write lock in order to prevent dead lock
2. When publish a transaction with multi tables, may cost a long time. Because there is only 1 publish thread, this will block the other txns.
    
    Here is how it happen:
   * the txn contains 2 tables: t0 and t1, publish get `t0 write lock`, and wait for `t1 write lock`
   * at the same time, client `insert into t0 select from t1` concurrently. `t1 read lock` is held by nereids planner, and this planner wait for `t0 read lock` until 1 minute timeout. 
   * the `t1 read lock` is held by multi load threads which continuously comes from client, so the publish wait a long time for `t1 write lock`.

   the fix is that:
   * when publish, we get table write lock with a timeout, if failed, the publish will retry in next loop. Make one txn does not block the other txns
   * for nereids planner, get table read lock in id order to find lock conflict more promptly


dataroaring:
run buildall

mymeiyi:
run buildall

dataroaring (dismissed):
LGTM

gavinchou (dismissed):

gavinchou:
run fe_ut

gavinchou:
run cloud_p1

mymeiyi:
run buildall

mymeiyi:
run fe_ut

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #37932, opened 2024-07-16
[fix](nereids)acos function should return null literal instead of NaN value

starocean999:
select acos(cast(1.1 as double)) should return NULL instead of NaN

## Proposed changes

Issue Number: close #xxx


starocean999:
run buildall

morrySnow (approved):

924060929 (approved):

PR #37729, opened 2024-07-12
[fix](colocate) fix colocate join while multi tables

cambyzju:
after https://github.com/apache/doris/pull/37361, in multi-tables cases, we should no depend on `SlotReference.getTable()`

PR #37937, opened 2024-07-16
[fix)(colocate join) fix wrong use of colocate join (#37361)(#37729)

cambyzju:
cherry-pick #37361 and #37729 to branch-2.0


cambyzju:
run buildall

PR #37760, opened 2024-07-14
[fix](MoW) fix MoW & segcompaction conflict on cache of temp segment

freemandealer:
MoW will update delete bitmap during load, and the page cache could be modified by segcompaction. Disable page cache touchs when doing segcompaction could solve this problem.

## Proposed changes

Issue Number: close #xxx

PR #37938, opened 2024-07-16
[fix](MoW) fix MoW & segcompaction conflict on cache of temp segment (pick #37760)

freemandealer:
pick #37760

MoW will update delete bitmap during load, and the page cache could be modified by segcompaction. Disable page cache touchs when doing segcompaction could solve this problem.

## Proposed changes

Issue Number: close #xxx


freemandealer:
run buildall

PR #37943, opened 2024-07-16
[enhancement](nereids) speedup sql cache with use variable as partition predicate

924060929:
cherry pick from #37915


924060929:
run buildall

PR #37959, opened 2024-07-16
[Fix](function) fix coredump for MULTI_MATCH_ANY

zclllyybb:
## Proposed changes

Issue Number: close #xxx

fix BE crash. now get:
```sql
mysql> select MULTI_MATCH_ANY(cast('aaaaa' as STRING), cast(['aaaaa', '12345!@#$%^&*()_', '', '中文测试', '123', 'abs ss 123'] as ARRAY < STRING > ));
ERROR 1105 (HY000): errCode = 2, detailMessage = (10.16.10.8)[INVALID_ARGUMENT][E33] Compile regexp expression failed. got Embedded start anchors not supported.. some expressions may be illegal
```


zclllyybb:
run buildall

Mryange (approved):

be/src/vec/functions/regexps.h
```diff
@@ -136,42 +136,45 @@ Regexps constructRegexps(const std::vector<String>& str_patterns,
...
+        }
     }
 
     /// We allocate the scratch space only once, then copy it across multiple threads with hs_clone_scratch
     /// function which is faster than allocating scratch space each time in each thread.
     hs_scratch_t* scratch = nullptr;
     err = hs_alloc_scratch(db, &scratch);
 
-    /// If not HS_SUCCESS, it is guaranteed that the memory would not be allocated for scratch.
-    if (err != HS_SUCCESS) LOG(FATAL) << "Could not allocate scratch space for hyperscan";
```
HappenLee, line 173: so what should the code run if `if (err != HS_SUCCESS)`

zclllyybb: it will be Exception thrown in code above. this check will never be true.

zclllyybb:
run buildall

HappenLee (approved):
LGTM

PR #37961, opened 2024-07-17
[Fix](InternalSchema) Compute nodes should not be used for Internal schema three replica (#36130)

morningman:
bp #36130


morningman:
run buildall

PR #37924, opened 2024-07-16
[fix](dynamic partition) fix create dynamic partition with overlap not throw exception

yujun777:
PR #35778  will throw exception if dynamic partition overlap with others.  Don't throw exception due to considering no behaviour change.  Just ignore create this partition.

SQL:
```
PARTITION BY RANGE(`dt`)(
PARTITION `phistory` VALUES less than ('2024-01-01')
)
DISTRIBUTED BY HASH(`user_id`, `room_id`) BUCKETS 3
PROPERTIES (
"replication_allocation" = "tag.location.default: 1",
"dynamic_partition.enable" = "true",
"dynamic_partition.time_unit" = "DAY",
"dynamic_partition.end" = "10",
"dynamic_partition.prefix" = "p",
"dynamic_partition.create_history_partition" = "true",
"dynamic_partition.history_partition_num" = "250"
); 
```

error:
```
ERROR 1105 (HY000): errCode = 2, detailMessage = errCode = 2, detailMessage = errCode = 2, detailMessage = Range [types: [DATEV2]; keys: [0000-01-01]; ..types: [DATEV2]; keys: [2024-01-01]; ) is intersected with range: [types: [DATEV2]; keys: [2023-12-29]; ..types: [DATEV2]; keys: [2023-12-30]; ) 
```

PR #37966, opened 2024-07-17
[fix](dynamic partition) fix create dynamic partition with overlap not throw exception #37924

yujun777:
cherry-pick: #37924


yujun777:
run buildall

PR #37930, opened 2024-07-16
[fix](case)Fix unusual case in manager_interface.

hubgeter:
## Proposed changes
fix in case : println   -> logger.info. 
Fix the case of multiple clusters.

Issue Number: close #xxx

PR #37967, opened 2024-07-17
[fix](case)Fix unusual case in manager_interface.(#37930)

hubgeter:
bp #37930 
## Proposed changes

Issue Number: close #xxx


hubgeter:
run buildall

PR #37985, opened 2024-07-17
[bug](delete) fix delete random distributed tbl

DarvenDuan:
## Proposed changes

Bug description:
In PR https://github.com/apache/doris/pull/33630, Doris supports auto aggregation for random distributed table, but it not only effects query statements, so if we delete from a random distributed table, will get an error because of unexpectedly rewriting.
```
CREATE TABLE `test_tbl` (
  `k` INT NULL,
  `v` BIGINT SUM NULL
) ENGINE=OLAP
AGGREGATE KEY(`k`)
DISTRIBUTED BY RANDOM BUCKETS AUTO;

mysql > delete from test_tbl where k=1;
ERROR 1105 (HY000): errCode = 2, detailMessage = Where clause only supports compound predicate, binary predicate, is_null predicate or in predicate.
```

fix:
Check whether it is a query statement before rewriting.


DarvenDuan:
run buildall

morrySnow (commented):
add test case please

DarvenDuan:
> add test case please

done

morrySnow (approved):

dataroaring (approved):
LGTM

PR #37995, opened 2024-07-17
[fix](case)fix s3 tvf case failed in pipeline of 2.1

zddr:
default value `enable_fallback_to_original_planner` is true, resource priv check only support in nereids,so need set enable_fallback_to_original_planner=false;


zddr:
run buildall

yiguolei:
run buildall

PR #38006, opened 2024-07-17
[fix](compaction) Fix data race when fetching base compaction score from cloud tablet

Hastyshell:
## Proposed changes

When calculating base compaction score on cloud tablet, data race may cause rowset meta become nullptr and cause a crash.

Therefore added a read lock as protection in this case.


TangSiyang2001:
run buildall

luwei16 (approved):

dataroaring (approved):
LGTM

PR #38031, opened 2024-07-17
[Fix](load) Fix the channel leak when close wait has been cancelled

liaoxin01:
When the close_wait is called, the NodeChannel has already been marked as cancelled, but close_wait will set _is_closed to true. When it actually sends a cancel request to the downstream LoadChannel, it finds that _is_closed has already been set to true, so it will not send an RPC request, causing a LoadChannel leak.


dataroaring (dismissed):
LGTM

liaoxin01:
run buildall

gavinchou (approved):

dataroaring (approved):
LGTM

PR #37724, opened 2024-07-12
[fix](publish) Catch exception in genPublishTask to make one failed txn does not block the other txns

mymeiyi:
## Proposed changes

If any exception(such as NullPointerException) is thrown in `genPublishTask` when publish, the publish for all txns will fail.
This pr catch the exception to make the failed txn does not block other txns.

PR #38042, opened 2024-07-18
[fix](publish) Pick Catch exception in genPublishTask to make one failed txn does not block the other txns (#37724)

mymeiyi:
Pick https://github.com/apache/doris/pull/37724


mymeiyi:
run buildall

PR #34584, opened 2024-05-09
[pipeline](fix) Avoid to use a freed dependency when cancelled

Gabriel39:
## Proposed changes

*** Query id: 0-0 ***
*** is nereids: 0 ***
*** tablet id: 0 ***
*** Aborted at 1715228866 (unix time) try "date -d @1715228866" if you are using GNU date ***
*** Current BE git commitID: 6849cd521a ***
*** SIGSEGV address not mapped to object (@0x41) received by PID 9394 (TID 11207 OR 0x7f2db5012700) from PID 65; stack trace: ***
 0# doris::signal::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*) at /home/zcp/repo_center/doris_master/doris/be/src/common/signal_handler.h:421
 1# PosixSignals::chained_handler(int, siginfo*, void*) [clone .part.0] in /usr/lib/jvm/java-17-openjdk-amd64/lib/server/libjvm.so
 2# JVM_handle_linux_signal in /usr/lib/jvm/java-17-openjdk-amd64/lib/server/libjvm.so
 3# 0x00007F35F1187090 in /lib/x86_64-linux-gnu/libc.so.6
 4# std::atomic::operator bool() const at /var/local/ldb_toolchain/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/atomic:87
 5# doris::pipeline::Dependency::set_ready() at /home/zcp/repo_center/doris_master/doris/be/src/pipeline/dependency.cpp:56
 6# doris::pipeline::PipelineTask::clear_blocking_state() at /home/zcp/repo_center/doris_master/doris/be/src/pipeline/pipeline_task.h:199
 7# doris::pipeline::PipelineFragmentContext::cancel(doris::PPlanFragmentCancelReason const&, std::__cxx11::basic_string, std::allocator > const&) at /home/zcp/repo_center/doris_master/doris/be/src/pipeline/pipeline_fragment_context.cpp:219
 8# doris::QueryContext::cancel_pipeline_context(int, doris::PPlanFragmentCancelReason const&, std::__cxx11::basic_string, std::allocator > const&) at /home/zcp/repo_center/doris_master/doris/be/src/runtime/query_context.cpp:261
[...]

PR #38046, opened 2024-07-18
[pipeline](fix) Avoid to use a freed dependency when cancelled (#34584)

Gabriel39:
## Proposed changes

pick #34584


Gabriel39:
run buildall

PR #38036, opened 2024-07-18
[fix] (compile) Resolve time_point type mismatch in s3_rate_limiter.cpp in mac

Tech-Circle-48:
## Proposed changes

Issue Number: close #xxx

A compilation error in `s3_rate_limiter.cpp` due to a type mismatch between `std::chrono::system_clock::time_point` and `std::chrono::high_resolution_clock::time_point`，Updated the `_update_remain_token` function to accept `std::chrono::high_resolution_clock::time_point` instead of `std::chrono::system_clock::time_point` on MacOS。

PR #38049, opened 2024-07-18
[fix][compile] Resolve time_point type mismatch on MacOS and Linux

zy-kkk:
revert #38036
We cannot change the time_point type directly, but use macros to distinguish different compilation platforms, refer to #34054


zy-kkk:
run buildall

ByteYue (approved):

morningman (approved):

PR #35707, opened 2024-05-31
[test](mtmv)add increment create mtmv case

zfr9527:
More examples of the derivation of the partition mapping of the materialized view have been supplemented. The change situation of the right table of the join that was missed before has been added this time.

Existing mtmv examples:
select l_shipdate, l_orderkey, O_ORDERDATE, 
        count(orders_left_2.O_ORDERDATE) over (partition by lineitem_left_2.L_SHIPDATE order by lineitem_left_2.L_ORDERKEY) as window_count 
        from lineitem_left_2 
        left join orders_left_2 
        on lineitem_left_2.l_orderkey = orders_left_2.o_orderkey 
        group by l_shipdate, l_orderkey, O_ORDERDATE;

This time supplement examples in the right table that contain special operators:
select l_shipdate, l_orderkey, t.O_ORDERDATE, 
        count(t.O_ORDERDATE) over (partition by lineitem_left_2.L_SHIPDATE order by lineitem_left_2.L_ORDERKEY) as window_count 
        from lineitem_left_2 
        left join (select O_ORDERDATE, o_orderkey, count(O_ORDERDATE) over (partition by O_ORDERDATE order by o_orderkey) from orders_left_2 group by O_ORDERDATE, o_orderkey) as t
        on lineitem_left_2.l_orderkey = t.o_orderkey 
        group by l_shipdate, l_orderkey, t.O_ORDERDATE

PR #38054, opened 2024-07-18
[test](mtmv)add increment create mtmv case

zfr9527:
pick from master #35707 

## Proposed changes

[test](mtmv)add increment create mtmv case


zfr9527:
run buildall

PR #36562, opened 2024-06-19
[test](mtmv)Add group by aggregate negative case

zfr9527:

PR #38055, opened 2024-07-18
[test](mtmv)Add group by aggregate negative case

zfr9527:
pick from master #36562 

## Proposed changes

[test](mtmv)Add group by aggregate negative case


zfr9527:
run buildall

PR #38058, opened 2024-07-18
[enhancement](runtimefilter) fix potential core in runtime filter sync filter size

yiguolei:
## Proposed changes
IRuntimeFilter maybe deconstructed before the rpc finished, so that could not use a raw pointer in closure. Has to use the context's shared ptr.


yiguolei:
run buildall

BiteTheDDDDt (approved):

HappenLee (approved):
LGTM

PR #36391, opened 2024-06-17
[fix](test)fix regression test case failure

starocean999:
## Proposed changes

Issue Number: close #xxx

PR #38066, opened 2024-07-18
[fix](test)fix regression test case failure

starocean999:
## Proposed changes

pick from master https://github.com/apache/doris/pull/36391


starocean999:
run buildall

PR #38023, opened 2024-07-17
[enhancement](memtable) use shared ptr for flush token since it is shared between memtable write thread and flush thread

yiguolei:
## Proposed changes

Issue Number: close #xxx

PR #38068, opened 2024-07-18
[enhancement](memtable) use shared ptr for flush token since it is shared between memtable write thread and flush thread (#38023)

yiguolei:
pick https://github.com/apache/doris/pull/38023
---------

## Proposed changes

Issue Number: close #xxx


yiguolei:
run buildall

PR #38076, opened 2024-07-18
[fix](cloud) fix inconsistent reponse between cloud mode and local mode for streamload 2PC

liaoxin01:
fix  the regression test case of flink_connector_response for cloud p0.


dataroaring (approved):
LGTM

gavinchou (approved):

PR #38082, opened 2024-07-18
[fix](profile) Task state of query profile is not set correctly

zhiqiang-hhhh:
Task state in connection context will only be updated after profile is updated. So task state of profile should be set to query state of coordinator.


zhiqiang-hhhh:
run buildall

yiguolei (approved):

HappenLee (approved):
LGTM

PR #38089, opened 2024-07-18
[Fix](iceberg) Fix Iceberg count pushdown with limit

zhangbutao:
## Proposed changes

Issue Number: close #xxx

Iceberg count pushdown with limit always return 0. This change fix this issue.
```
set enable_count_push_down_for_external_table=true;
select count(*) from sample_mor_parquet limit 1;  -- will always return 0
```


zhangbutao:
run buildall

fe/fe-core/src/main/java/org/apache/doris/nereids/NereidsPlanner.java
```diff
@@ -610,8 +610,8 @@ public Optional<ResultSet> handleQueryInFe(StatementBase parsedStmt) {
         }
 
         if (physicalPlan instanceof PhysicalResultSink
-                && physicalPlan.child(0) instanceof PhysicalHashAggregate && !getScanNodes().isEmpty()
-                && getScanNodes().get(0) instanceof IcebergScanNode) {
+                && !getScanNodes().isEmpty() && getScanNodes().get(0) instanceof IcebergScanNode
+                && getScanNodes().get(0).getPushDownAggNoGroupingOp().equals(TPushAggOp.COUNT)) {
```
924060929, line 614: This condition is trick because it can only process this shape:
```
     PhysicalResultSink
            |
    PhysicalHashAggregate
            |
PhysicalFileScan(table=IcebergScan)
```

If there has another PhysicalPlan between PhysicalHashAggregate and PhysicalFileScan, it will compute a wrong result

```
     PhysicalResultSink
            |
    PhysicalHashAggregate
            |
    PhysicalHashAggregate
            |
PhysicalFileScan(table=IcebergScan)
```

for example:

```sql
select max(cnt) from (select count(*) as cnt from iceberg_table)a
```


I think you should remove this code in NereidsPlanner, the correct approach is
1. add `PhysicalCountIcebergTable` plan
2. add a rule in Rewrite to rewrite `LogicalAggregate(LogicalFileScan().when(scan -> scan.getTable instance IcebergTable)` to `PhysicalCountIcebergTable`
3. translate `PhysicalCountIcebergTable` like `PhysicalOneRowRelation`
5. finally the PhysicalCountIcebergTable can be compute by constant value

zhangbutao: @924060929 Many thanks for your suggestion!  Will check this later.

PR #38090, opened 2024-07-18
[fix](mtmv)Mtmv support force replica 2.1

zddr:
master do not have bug, because has call `PropertyAnalyzer.getInstance().rewriteOlapProperties`

2.1 not have this method ,should call `PropertyAnalyzer.rewriteReplicaAllocationProperties`


zddr:
run buildall

PR #35835, opened 2024-06-04
[fix](regression test) test_alter_colocate_group due to force_olap_table_replication_num

yujun777:
```
Exception in alter_p2/test_alter_colocate_group.groovy(line 94):

)
DISTRIBUTED BY HASH(k2) BUCKETS 5
PROPERTIES
(
"colocate_with" = "group_2",
"replication_num" = "1"
);
"""

sql """
^^^^^^^^^^^^^^^^^^^^^^^^^ERROR LINE^^^^^^^^^^^^^^^^^^^^^^^^^
CREATE TABLE tbl3
(
`uuid` varchar(255) NULL,
`action_datetime` date NULL
)
DUPLICATE KEY(uuid)
PARTITION BY RANGE(action_datetime)()
DISTRIBUTED BY HASH(uuid) BUCKETS 4
PROPERTIES
(

Exception:
java.sql.SQLSyntaxErrorException: Colocate tables must have same replication allocation:

{ tag.location.default: 1 }
should be

{ tag.location.default: 3 }
at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:120)
```

## Proposed changes

Issue Number: close #xxx

PR #38092, opened 2024-07-18
[fix](regression test) test_alter_colocate_group due to force_olap_table_replication_num #35835

yujun777:
cherry pick from #35835


yujun777:
run buildall

PR #38093, opened 2024-07-18
[enhancement](runtimefilter) fix potential core in runtime filter sync filter size (#38058)

yiguolei:
pick #38058

## Proposed changes
IRuntimeFilter maybe deconstructed before the rpc finished, so that could not use a raw pointer in closure. Has to use the context's shared ptr.

---------

## Proposed changes

Issue Number: close #xxx


yiguolei:
run buildall

PR #38108, opened 2024-07-19
[improve](test)Using Awaitility instead of Thread.sleep (#37812)

CalvinKirs:
## Proposed changes

Awaitility: Waits until a specific condition is met, which makes the tests more reliable and less prone to intermittent failures. It ensures that the test only proceeds when the expected condition is true. Thread.sleep: Introduces a fixed delay, which can lead to flaky tests. If the condition is met sooner than the sleep time, the test unnecessarily waits. Conversely, if the condition is met after the sleep time, the test will fail.

(cherry picked from commit 8d249a2562b15825a06a53a5a530a532a0ca2454)

## Proposed changes

Issue Number: close #37812

### DIFF
--- a/regression-test/framework/pom.xml
+++ b/regression-test/framework/pom.xml
@@ -286,5 +286,13 @@ under the License.
             <artifactId>jcc</artifactId>
             <version>11.5.8.0</version>
         </dependency>
+        <dependency>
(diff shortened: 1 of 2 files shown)


CalvinKirs:
run buildall

PR #38125, opened 2024-07-19
[Fix](load) Fix the channel leak when close wait has been cancelled #38031

liaoxin01:
cherry pick from #38031


liaoxin01:
run buildall

PR #38147, opened 2024-07-19
[Fix](JsonReader) fix json with duplicate key entry may result out of  bound exception

eldenmoon:
#38146


eldenmoon:
run buildall

PR #38148, opened 2024-07-19
[fix](nereids)mv selection should only consider visible indexes

starocean999:
the indexes in olap table are not always visible. Sometimes, schema change will introduce invisible index into olap table. And we should only use getVisibleIndexIdToMeta() to get visible indexes as candidates for mv selection.


starocean999:
run buildall

starocean999:
run buildall

starocean999:
run buildall

morrySnow (approved):

seawinde (approved):

PR #38162, opened 2024-07-19
[Fix](JobManager)Release the lock immediately after modifying job metadata to avoid holding the lock for an extended period.

CalvinKirs:
## Proposed changes

When deleting a job in MV, a DB lock is held. However, due to the larger lock granularity of the Job Manager, this may prevent obtaining the lock. This issue arises because during the creation of a job, immediately scheduled tasks need to perform their initial scheduling and compensate for tasks within the time window. To alleviate this issue, we are attempting to reduce the lock granularity.

PR #38163, opened 2024-07-20
[Fix](JobManager)Release the lock immediately after modifying job metadata to avoid holding the lock for an extended period.#38162

CalvinKirs:
## Proposed changes
#38162


CalvinKirs:
run buildall

yiguolei (approved):

PR #38175, opened 2024-07-21
[fix](statistics)Remove old partition stats in the next analyze after insert overwrite.

Jibing-Li:
Remove old partition stats in the next analyze after insert overwrite. Because insert overwrite will use new partitions to replace old ones, which means all partitions are new and the old partition stats should be removed.


Jibing-Li:
run buildall

Jibing-Li:
run feut

Jibing-Li:
run cloud_p1

Jibing-Li:
run cloud_p1

zfr9527 (approved):
LGTM

morrySnow (approved):

PR #37782, opened 2024-07-15
[enhancement](compaction) Control the parallelism for urgent compaction tasks

Hastyshell:
## Proposed changes

For some urgent compaction tasks, their submittion should take parallelism into account.

Currently, we apply the control policy for data loading in specific. Other source of urgent tasks are considered as eager.

PR #38189, opened 2024-07-22
[enhancement](compaction) Control the parallelism for urgent compacton tasks (#37782)

Hastyshell:
## Proposed changes

pick #37782

For some urgent compaction tasks, their submittion should take parallelism into account.

Currently, we apply the control policy for data loading in specific. Other source of urgent tasks are considered as eager.


TangSiyang2001:
run buildall

PR #38200, opened 2024-07-22
[fix](pipeline) Fix blocked tasks if query is canceled before opening

Gabriel39:
## Proposed changes

pick #38206


Gabriel39:
run buildall

PR #38203, opened 2024-07-22
[Fix](outfile) FE check the hdfs URI of outfile

BePPPower:
1. Previously, if the root path of the HDFS URI started with two slashes, the outfile would be successfully exported without errors, but the exported path would not be the expected path.
Currently, we will delete repeated '/' which specified by users in FE.

2. move the test case for outfile HDFS from p2 to p0.


BePPPower:
run buildall

BePPPower:
run buildall

morningman (dismissed):
LGTM

BePPPower:
run buildall

morningman (approved):
LGTM

kaka11chen (approved):
LGTM

BePPPower:
run cloud_p0 cloud_p1

PR #38212, opened 2024-07-22
[enhance](mtmv)refresh hms table before run mtmv task

zddr:
There are some caches in the HMS catalog, which can result in the inability to maintain consistency with Hive side data even after a successful refresh of the materialized view.
In this PR, the cache of relevant HMS tables will be cleared before refreshing the materialized view to solve this problem


zddr:
run buildall

zddr:
run buildall

zddr:
run buildall

regression-test/suites/mtmv_p0/test_hive_refresh_mtmv.groovy
```diff
@@ -157,9 +148,6 @@ suite("test_hive_refresh_mtmv", "p0,external,hive,external_docker,external_docke
                                     """
```
seawinde, line 148: after refresh materialized view, maybe we should add `order_qt` to check the mv result is right or not

morrySnow (approved):

seawinde (approved):

PR #38211, opened 2024-07-22
[fix](regression test) fix test_report_version_missing due to set force_olap_table_replication_num

yujun777:
error:

```
Exception in control_p0/test_report_version_missing.groovy(line 68):

succ = true
break
}
}
if (succ)

{ break }
Thread.sleep(60000)
}
assertTrue(succ)
^^^^^^^^^^^^^^^^^^^^^^^^^ERROR LINE^^^^^^^^^^^^^^^^^^^^^^^^^

} finally

{ GetDebugPoint().disableDebugPointForAllBEs("Tablet.build_tablet_report_info.version_miss") }
}

Exception:
org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:40)
at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:35)
at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:179)
at sun.reflect.GeneratedMethodAccessor55.invoke(Unknown Source)
```

logs:
```
2024-07-22 03:20:55,050 INFO (tablet checker|201) Add tablet to pending queue, tablet id: 807781, status: VERSION_INCOMPLETE, state: PENDING, type: REPAIR, priority: HIGH, tablet size: 0, visible version: -1, committed version: -1
2024-07-22 03:20:55,061 INFO (tablet checker|201) finished to check tablets. unhealth/total/added/in_sched/not_ready: 1/87388/1/0/0, cost: 87 ms
2024-07-22 03:20:55,464 INFO (tablet scheduler|201) balance or tablet scheduler is disabled. skip selecting tablets for balance
2024-07-22 03:20:55,464 INFO (tablet scheduler|201) create clone task to repair replica, tabletId=807781, replica=[replicaId=807783, BackendId=10008, version=2, dataSize=346, rowCount=10, lastFailedVersion=3, lastSuccessVersion=2, lastFailedTimestamp=1721589639451, schemaHash=1701605109, state=NORMAL, isBad=false], visible version 2, tablet status VERSION_INCOMPLETE
[...]

PR #38232, opened 2024-07-23
[fix](regression test) fix test_report_version_missing due to set force_olap_table_replication_num #38211

yujun777:
cherry pick from #38211


yujun777:
run buildall

PR #38237, opened 2024-07-23
[feature](workloadgroup) use slot num to control memory distribution among one workload group

yiguolei:
## Proposed changes

1. use wg_mem_limit / total slots count as default query mem limit.
2.  user could specify query_slot_count to adjust the mem limit for specific query.
3. user could also specify enable_query_hard_limit to indicate use hard limit for specific query. But we do not expect user could modify this.
4. total slot number == max_concurrency refers https://docs.aws.amazon.com/redshift/latest/dg/r_wlm_query_slot_count.html


yiguolei:
run buildall

yiguolei:
run buildall

PR #38198, opened 2024-07-22
[fix](load data) decommission replica don't load data when it misses versions

yujun777:
BUG:  when a decommission replica misses versions (last failed version > 0),   it will cause a bug:
1. It cann't do compaction due to it has miss versions;
2. It cann't  cloning the missing rowsets because it's decommissioned.

So if it continue to load data, its rowset will increase  util it cause error -235(TOO MANY VERSION).

Fix:  don't write data to the decommission replica if it miss rowsets.

PR #38257, opened 2024-07-23
[fix](load data) decommission replica don't load data when it misses versions #38198

yujun777:
cherry pick from #38198


yujun777:
run buildall

PR #34220, opened 2024-04-28
[fix](insert) Insert init plan error may forget to abort txn

mymeiyi:
## Proposed changes

When FE generate plan for insert into command:
1. begin txn and finalize sink
2. if finalize sink throws exception, the txn is not aborted

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #38260, opened 2024-07-23
[fix](insert) Pick Insert init plan error may forget to abort txn (#34220)

mymeiyi:
Pick https://github.com/apache/doris/pull/34220


mymeiyi:
run buildall

PR #38262, opened 2024-07-23
[fix](group commit) Fix prepare stmt setNull return too many filtered rows error

mymeiyi:
When use prepare statement and setNull to do group commit:
```
stmt.setString(3, "WAGERNO");
stmt.setNull(4, Types.INTEGER);
```
we may get `too many rows filtered` rows.


mymeiyi:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #38209, opened 2024-07-22
[Fix](inverted index) skip index compaction for inverted index V2

airborne12:
## Proposed changes

Enabling index compaction in the inverted index V2 format currently causes unexpected errors, especially in the case of tables with hybrid indexes, such as BKD index and Fulltext index together.

PR #38273, opened 2024-07-24
[Pick 3.0](inverted index) skip index compaction for inverted index V2 

airborne12:
## Proposed changes

Enabling index compaction in the inverted index V2 format currently causes unexpected errors, especially in the case of tables with hybrid indexes, such as BKD index and Fulltext index together.

backport #38209


airborne12:
run buildall

PR #38275, opened 2024-07-24
[fix](broker) Broker already had baidu afs support but can't really use, so i fix it

biohazard4321:
## Proposed changes

Issue Number: close #xxx

Add afs to the scheme options so broker can recognize afs location,  and change default afs impl to pure java implementation.


biohazard4321:
run buildall

yiguolei (approved):

dataroaring (approved):
LGTM

PR #38276, opened 2024-07-24
[fix](group commit) Fix prepare stmt setNull return too many filtered rows error (#38262)

mymeiyi:
Pick https://github.com/apache/doris/pull/38262


mymeiyi:
run buildall

mymeiyi:
run buildall

PR #38284, opened 2024-07-24
[fix](array_range) fix array_range func for large param which should return error

amorynan:
## Proposed changes
if array_range with large size , which will make be oom , so we should avoid this with max_array_size_as_field  
Issue Number: close #xxx


amorynan:
run buildall

eldenmoon (approved):
LGTM

zclllyybb (approved):
LGTM

xiaokang (approved):

amorynan:
run beut

PR #38287, opened 2024-07-24
[opt](mow) opt mow lookup with sequence column

cambyzju:
Compare primary key without sequence column is expensive,
if full key is exact_match, the primary key without sequence column should also the same,
do not need to compare again.


cambyzju:
run buildall

bobhan1 (approved):

cambyzju:
run buildall

bobhan1:
run external

zhannngchen (approved):
LGTM

PR #37362, opened 2024-07-05
[fix](auth)Add some log in auth case

zfr9527:
## Proposed changes

[fix](auth)Add some log in auth case

PR #38289, opened 2024-07-24
[fix](auth)Add some log in auth case

zfr9527:
pick from master #37362 

## Proposed changes

[fix](auth)Add some log in auth case


zfr9527:
run buildall

PR #37892, opened 2024-07-16
[fix](function) the bucket number arg of width_bucket should be a positive integer value

mrhhsg:
## Proposed changes

Issue Number: close #xxx

PR #38295, opened 2024-07-24
[fix](function) the bucket number arg of width_bucket should be a positive integer value (#37892)

mrhhsg:
## Proposed changes

pick #37892


mrhhsg:
run buildall

PR #38224, opened 2024-07-22
[fix](compaction) fix duplicate key in agg/mor table caused by ordered data compaction

luwei16:

PR #38299, opened 2024-07-24
[fix](compaction) fix duplicate key in agg/mor table caused by ordered data compaction (#38224)

luwei16:
pick master #38224


luwei16:
run buildall

PR #38310, opened 2024-07-24
[feature](audit) add new FE config to skip audit for certain user

morningman:
Sometime we don't want to audit operation from certain user in audit log or audit table.
Add a new FE config `skip_audit_user_list`.
Default is empty, which means all operations will be recorded.
When you want to ignore some user's operation, you can set this config like:

```
skip_audit_user_list=user1
--or
skip_audit_user_list=user1,user2
```


morningman:
run buildall

gavinchou (approved):

zddr (approved):

PR #38318, opened 2024-07-24
[Fix](Variant) fix outof range exception after calling function element_at for variant type

eldenmoon:
…nt_at for variant type

src subcolumns empty but src row count may not be 0, so need to padding rows to at least src.size() and finlize at last

```
[CANCELLED][E6] Parameter out of bound in IColumnStr<T>::insert_range_from me !
```
## Proposed changes

Issue Number: close #xxx


eldenmoon:
run buildall

amorynan (approved):
LGTM

xiaokang (approved):
LGTM

PR #37492, opened 2024-07-08
[enhance](Cooldown) Use config to control whether use cooldown replica for scanning first

ByteYue:
## Proposed changes

Issue Number: close #xxx

PR #38322, opened 2024-07-24
Pick "[enhance](Cooldown) Use config to control whether use cooldown replica for scanning first"

ByteYue:
## Proposed changes

Same as master #37492


ByteYue:
run buildall

PR #38325, opened 2024-07-24
[fix](load) fix no error url for stream load

liaoxin01:
Previously, it was determined whether to output an error URL based on whether num_filtered_rows was greater than 0. However, in abnormal situations, the statistics of num_filtered_rows may not be accurate, resulting in no error URL output. As long as there is data  in the error log file indicating filtered rows, the URL can be returned to the client without the need to determine the number of filtered rows.


liaoxin01:
run buildall

liaoxin01:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

gavinchou (approved):

PR #38353, opened 2024-07-25
[fix](cloud) ensure error log upload to S3 before returning load result

sollhui:


sollhui:
run buildall

sollhui:
run buildall

PR #38369, opened 2024-07-25
[fix](merge-on-write) segcompaction should process delete bitmap if necessary

zhannngchen:
## Proposed changes

Issue Number: close #xxx

When loading data to a unique key table with sequence column, some data in current load job might be marked as delete due to a lower sequence value. 
If there's many segments in such load job, segcompaction might be triggered, which don't process the delete bitmap currently, will cause data correctness issue

For example:
1. we have 4 segments in current load job initially, and due to seq column, some rows are marked as deleted
2. after segcompaction, if we don't process the delete bitmap, it's content is still corresponding to the old segment layout, and row 7,14,15 is not mark deleted correctly on new generated segment 1.
3. in this PR, we convert old delete bitmap to fit new segment layout, it use similar way as base/cumulative compaction to convert delete bitmaps on old layout to new one, but the rowid conversion is simpler

whiteboard_exported_image-2

### DIFF
--- a/be/src/olap/merger.cpp
+++ b/be/src/olap/merger.cpp
@@ -336,16 +336,12 @@ Status Merger::vertical_compact_one_group(
 }
 
 // for segcompaction
-Status Merger::vertical_compact_one_group(int64_t tablet_id, ReaderType reader_type,
-                                          const TabletSchema& tablet_schema, bool is_key,
-                                          const std::vector<uint32_t>& column_group,
-                                          vectorized::RowSourcesBuffer* row_source_buf,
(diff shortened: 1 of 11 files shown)


zhannngchen:
run buildall

zhannngchen:
run buildall

zhannngchen:
run buildall

zhannngchen:
run buildall

zhannngchen:
run buildall

dataroaring (approved):
LGTM

hust-hhb (approved):
LGTM

PR #38382, opened 2024-07-25
[Fix](Variant) implement resize and set num_rows right when `subcolum…

eldenmoon:
…ns` is empty

## Proposed changes

Issue Number: close #xxx


eldenmoon:
run buildall

PR #38349, opened 2024-07-25
[Improvement](bloom filter) Forbid small bloom filter

Gabriel39:
## Proposed changes

Bloom filter has a expected filter ratio when data is enough. This PR forbid too small bloom filter which has a big bias for filter ratio.

PR #38392, opened 2024-07-25
[Improvement](bloom filter) Forbid small bloom filter (#38349)

Gabriel39:
Bloom filter has a expected filter ratio when data is enough. This PR forbid too small bloom filter which has a big bias for filter ratio.

## Proposed changes
pick #38349


Gabriel39:
run buildall

PR #37461, opened 2024-07-08
[fix](spill) Duplicate calls to Dependency::set_ready() in hash join

mrhhsg:
## Proposed changes

Duplicate calling the function `Dependency::set_ready()` will cause pipeline tasks to be scheduled incorrectly.

PR #38399, opened 2024-07-26
[fix](spill) Duplicate calls to Dependency::set_ready() in hash join(#37461)

mrhhsg:
## Proposed changes

pick #37461
Duplicate calling the function `Dependency::set_ready()` will cause pipeline tasks to be scheduled incorrectly.


mrhhsg:
run buildall

PR #38408, opened 2024-07-26
[feature](schema change) unified schema change for parquet and orc reader (#32873)

hubgeter:
## Proposed changes
bp #32873 

Scenario: Reading a hive table after adding fields to a struct column
Since there are still problems with reading tables in parquet and text formats on the master in this scenario, only tables in orc format are picked here and some cases are added.


hubgeter:
run buildall

be/src/vec/data_types/data_type_struct.cpp
```diff
@@ -355,14 +355,7 @@ std::optional<size_t> DataTypeStruct::try_get_position_by_name(const String& nam
...
 String DataTypeStruct::get_name_by_position(size_t i) const {
-    if (i == 0 || i > names.size()) {
-        fmt::memory_buffer error_msg;
-        fmt::format_to(error_msg, "Index of tuple element ({}) if out range ([1, {}])", i,
-                       names.size());
-        LOG(FATAL) << fmt::to_string(error_msg);
-    }
-
-    return names[i - 1];
+    return names[i];
```
morningman, line 358: Why removing this check?

hubgeter: because : https://github.com/apache/doris/pull/32873/files#r1555136892

morningman (approved):
LGTM

PR #38413, opened 2024-07-26
[Fix](Variant) check storage type before apply `get_row_ranges_by_dict`

eldenmoon:
1. storage type maybe missmatch with predicate type eg. cast(v['a'] as date) = '2020-01-01'. v['a'] in storage is string type, but predicate type is date type

2. serialization object with empty subcolumns may lost num_rows, so need to record num_rows and set back num_rows in serdes

## Proposed changes

Issue Number: close #xxx


eldenmoon:
run buildall

be/src/vec/data_types/data_type_object.cpp
```diff
@@ -121,6 +126,11 @@ char* DataTypeObject::serialize(const IColumn& column, char* buf, int be_exec_ve
     }
     // serialize num of subcolumns
     *reinterpret_cast<uint32_t*>(size_pos) = num_of_columns;
+    if (num_of_columns == 0) {
+        // Make sure the num_of_rows in ColumnObject not lost
+        *reinterpret_cast<uint32_t*>(buf) = column_object.rows();
```
xiaokang, line 131: 1. Why only serialize num_rows when num_of_columns == 0?
2. It may cause error when rolling upgrade be if serialization is changed.

be/src/olap/rowset/segment_v2/segment_iterator.cpp
```diff
@@ -699,6 +699,11 @@ Status SegmentIterator::_get_row_ranges_from_conditions(RowRanges* condition_row
         if (_opts.io_ctx.reader_type == ReaderType::READER_QUERY) {
             RowRanges dict_row_ranges = RowRanges::create_single(num_rows());
             for (auto cid : cids) {
+                if (!_segment->can_apply_predicate_safely(cid,
```
xiaokang, line 702: There is check for can_apply_predicate_safely() in bf and zonemap branch. Why it's not check in this branch before? Does this modification affect other logic in this branch?

be/src/vec/data_types/data_type_object.cpp
eldenmoon: fixed

be/src/olap/rowset/segment_v2/segment_iterator.cpp
eldenmoon: only takes effect when storage type and predicate type ares missmatched typically in variant predicates with cast

eldenmoon:
run buildall

eldenmoon:
run buildall

xiaokang (approved):
LGTM

amorynan (approved):
LGTM

eldenmoon:
run performance

eldenmoon:
run buildall

PR #38332, opened 2024-07-24
[Opt](load) don't print stack when some errors occur for stream load

liaoxin01:
## Proposed changes

Issue Number: close #xxx

PR #38418, opened 2024-07-26
[Opt](load) don't print stack when some errors occur for stream load #38332

liaoxin01:
cherry pick from #38332


liaoxin01:
run buildall

PR #38011, opened 2024-07-17
[fix](auth)fix create mv not need alter priv of table

zddr:
case will add in another pr

PR #38423, opened 2024-07-26
[fix](auth) fix create mv not need alter priv of table (#38011)

zddr:
pick: https://github.com/apache/doris/pull/38011


zddr:
run buildall

PR #38428, opened 2024-07-26
[fix](statistics)Add row count to AnalysisJob and update tableStats using this row count after analyze.

Jibing-Li:
Add row count to AnalysisJob and update tableStats using this row count after analyze.
Take row count to consideration in stats health calculation.


Jibing-Li:
run buildall

PR #38437, opened 2024-07-28
[fix](profile) Fix read load profile failed

zhiqiang-hhhh:
Profile ID of load task is a long string, `DebugUtil.parseTUniqueIdFromString` will throw exception instead of null when we pass long string to it. So we should handle the exception inside of parseProfileFileName.


zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

dataroaring (approved):
LGTM

yiguolei (approved):

PR #29799, opened 2024-01-10
[Improvement](function) optimization for substr with ascii string

BiteTheDDDDt:
## Proposed changes
optimization for substr with ascii string
```sql
select count(substr(l_comment, 1, 10)) from lineitem;
```

before:
图片

after:
图片

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #38467, opened 2024-07-29
[fix](function)Fix substr utf8 error when pos is larger than str size

koarz:
## Proposed changes

Issue Number: close #xxx

https://github.com/apache/doris/pull/29799#issuecomment-2247334457
When the str parameter of substr is a utf8 character, the correct result should return empty when pos > size of str, but this case is not handled.


koarz:
run buildall

koarz:
run buildall

PR #38469, opened 2024-07-29
[improvement](drop tablet) reduce shard lock wait time

yujun777:
drop tablet's code:

```
    lock_guard(tablet_manager.shard.lock);

    delete tablet from tablet map

     {
            lock_guard(tablet.meta_lock)
            ...
     }

     ...
```

Sometimes,  tablet meta lock may held by other threads for a long time,    it will cause shard.lock also wait for a long time.

For creating tablet thread, it also need to acquire the shard.lock,  so creating tablet thread will also wait for a long time, then creating partition will be timeout.

Fix:   shard.lock   don't wait the time of the tablet meta lock.


yujun777:
run buildall

deardeng (approved):
LGTM

dataroaring (approved):
LGTM

PR #38470, opened 2024-07-29
[Fix](Nereids) fix fold constant rule by be cause case when exception

LiBinfeng-01:
Problem:
when use be fold constant rule, whenclause in case when expression would be folded into constant, so fe would throw an exception cause of whenclause can not be constant
Solved:
When be fold constant rule fold whenclause success, fe would make this constant as first priority default value


LiBinfeng-01:
run buildall

PR #38480, opened 2024-07-29
[chore](fe/be) Log the cloud/local mode

w41ter:


w41ter:
run buildall

fe/fe-core/src/main/java/org/apache/doris/DorisFE.java
```diff
@@ -403,6 +403,13 @@ private static void printVersion() {
         LOG.info("Build info: {}", Version.DORIS_BUILD_INFO);
         LOG.info("Build hash: {}", Version.DORIS_BUILD_HASH);
         LOG.info("Java compile version: {}", Version.DORIS_JAVA_COMPILE_VERSION);
+
+        if (Config.isCloudMode()) {
+            LOG.info("Run FE in the cloud mode, cloud_unique_id: {}, meta_service_endpoint: {}",
```
gavinchou, line 408: ditto

dataroaring (commented):
It is better to print in fe.out and be.out, because fe.log and be.INFO can be rolled out.

PR #38490, opened 2024-07-29
[fix](simdjson) fix simdjson reader will parse given root twice when data is object array 

amorynan:
## Proposed changes
if we set jsonroot is not empty , and data is object array , here will meet status like this:
image

Issue Number: close #xxx


amorynan:
run buildall

eldenmoon (approved):
LGTM

xiaokang (approved):
LGTM

PR #38511, opened 2024-07-30
[fix](cast) differences between FE and BE in parsing strings to integers.

Mryange:
## Proposed changes

```
mysql [(none)]>select cast('3.000' as int);
+----------------------+
| cast('3.000' as INT) |
+----------------------+
|                    3 |
+----------------------+

mysql [(none)]>set debug_skip_fold_constant = true;

mysql [(none)]>select cast('3.000' as int);
+----------------------+
| cast('3.000' as INT) |
+----------------------+
|                 NULL |
+----------------------+
```


Mryange:
run buildall

Mryange:
run buildall

PR #38533, opened 2024-07-30
[enhance](tags)Support authentication of tags used by users when creating tables and partitions

whutpencil:
## Proposed changes

When users create tables or partitions, FE will not verify whether they have the corresponding tag BE permission. Resource tags can only be used for queries and loads. This PR adds a configuration that, if enabled, only allows users to create tables or partitions under BE with tag permission

Issue Number: close #xxx


morningman (commented):
LGTM, but need to add FE unit test

PR #38545, opened 2024-07-30
[fix](regression) fix prepare_insert when execute prepare stmt in observer fe

mymeiyi:
observer fe will forward insert statements to master fe, and forward does not support prepare protocol


dataroaring (approved):
LGTM

dataroaring:
run buildall

Yukang-Lian (approved):

PR #38526, opened 2024-07-30
[fix](nereids) fix insert stmt throw MultiCastDataSink cannot be cast to DataStreamSink

924060929:
## Proposed changes

fix `insert ... with ... select ...`, which not use some cte, and throw an exception:
```
errCode = 2, detailMessage = class org.apache.doris.planner.MultiCastDataSink cannot be cast to class org.apache.doris.planner.DataStreamSink (org.apache.doris.planner.MultiCastDataSink and org.apache.doris.planner.DataStreamSink are in unnamed module of loader 'app')
```

PR #38547, opened 2024-07-30
[fix](nereids) fix insert stmt throw MultiCastDataSink cannot be cast to DataStreamSink (#38526)

924060929:
cherry pick from #38526


924060929:
run buildall

PR #38568, opened 2024-07-31
[pipeline](fix) Fix blocking task which is not triggered by 2nd RPC

Gabriel39:
## Proposed changes

Once a query is cancelled due to any reason, BE may not receive 2nd RPC from FE. If so, we must ensure the execution dependency is ready so tasks will not be blocked.


Gabriel39:
run buildall

yiguolei (approved):
LGTM

BiteTheDDDDt (approved):

PR #38581, opened 2024-07-31
[Fix](inverted index) fix wrong no need read data when same column in inverted index and like function #36687

airborne12:
cherry pick from #36687


airborne12:
run buildall

PR #37898, opened 2024-07-16
[case][fix](iceberg)move rest cases from p2 to p0 and fix iceberg version issue

wuwenchi:
## Proposed changes

move cases from p2 to p0

PR #38589, opened 2024-07-31
[case][fix](iceberg)move rest cases from p2 to p0 and fix iceberg version issue for 2.1 (#37898)

wuwenchi:
bp: #37898


wuwenchi:
run buildall

PR #38421, opened 2024-07-26
[fix](catalog)fix db name may be null in NotificationEvent

zddr:
cause by: https://github.com/apache/doris/pull/29059

PR #38596, opened 2024-07-31
[fix](catalog)fix db name may be null in NotificationEvent (#38421)

zddr:
pick: https://github.com/apache/doris/pull/38421


zddr:
run buildall

PR #38603, opened 2024-07-31
[regression](spark)Add spark to read doris predicate pushdown case

caoliang-web:
## Proposed changes

Issue Number: close #xxx

Add spark to read doris predicate pushdown case


caoliang-web:
run buildall

PR #38608, opened 2024-07-31
[fix](hist) Fix unstable result of aggregrate function hist

zhiqiang-hhhh:
* Target

Fix unstable result of hist function when involving null value.

* Reproduce

test result of `regression-test/suites/query_p0/sql_functions/aggregate_functions/test_aggregate_all_functions2.groovy` is unstable, sql `SELECT histogram(k7, 5) FROM baseall` will sometimes acts like the second argument is not passed in.

* Root reason

We have short-circuit in AggregateFunctionNullVariadicInline, when this row is NULL, the value will not be added by the nested function. Implementation of histogram relies on its add method to get its seconds argument, when we have an all null value block, histogram will not get its seconds arg even if sql is like `select(k7, 5)`, so a max_bucket_num with default value 128 is serialized. When we do merging, and happens to deserialize the above block at last, the max_bucket_num in merge stage will be assigned to 128, and this leads to the wrong result.

* Fix by

Init value of max_bucket_num is assigned to 0, when we do merging, we will discard this aggregated data if its max_bucket_num is 0.


zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

be/src/vec/aggregate_functions/aggregate_function_histogram.h
```diff
@@ -186,13 +198,18 @@ class AggregateFunctionHistogram final
...
+        if constexpr (has_input_param) {
+            Int32 input_max_num_buckets =
+                    assert_cast<const ColumnInt32*>(columns[1])->get_element(row_num);
+            if (input_max_num_buckets <= 0) {
+                throw doris::Exception(ErrorCode::INVALID_ARGUMENT,
+                                       "Invalid max_num_buckets {}, row_num {}",
+                                       input_max_num_buckets, row_num);
+            }
             this->data(place).set_parameters(
                     assert_cast<const ColumnInt32*>(columns[1])->get_element(row_num));
```
zhangstar333, line 210: here could use input_max_num_buckets directly

be/src/vec/aggregate_functions/aggregate_function_histogram.h
```diff
@@ -86,7 +92,9 @@ struct AggregateFunctionHistogramData {
     }
 
     void merge(const AggregateFunctionHistogramData& rhs) {
-        if (!rhs.max_num_buckets) {
+        // if rhs.max_num_buckets == 1, it means the input block for serialization is all null
```
zhangstar333, line 95: rhs.max_num_buckets == 1,  seems is means == -1?

zhiqiang-hhhh: yes, i will fix

be/src/vec/aggregate_functions/aggregate_function_histogram.h
zhiqiang-hhhh: yes

zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
@zhangstar333 I changed the default value from -1 to 0, so that we can fix the problem in an upgrade friendly way.

zhiqiang-hhhh:
run external

zhangstar333 (approved):

HappenLee (approved):
LGTM

PR #38611, opened 2024-07-31
[fix](group commit) add debug log show why group commit not work; delete wal when replay success

mymeiyi:
## Proposed changes

1. add debug log to show which condition mismatch, cause group commit insert does not work
2. delete wal immediately after replay, the original logic is replay all wals and then delete
3. add a replay wal case


dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #38612, opened 2024-07-31
[bug](expr)fix operator do_projections should use local_state intermediate_projections

zhangstar333:
should use local_state->_intermediate_projections, this is clone from parent intermediate_projections.
So it can run independently between different instance operators without affecting each other。
It needs to run in some extreme multi-threaded environments, so it's not easy to add test cases.


zhangstar333:
run buildall

HappenLee:
Add TEST and DESC of the PR

zhangstar333:
run buildall

yiguolei (approved):

HappenLee (approved):
LGTM

PR #38617, opened 2024-07-31
[Bug](fix) fix ubsan use int32_t pointer access bool value

HappenLee:
## Proposed changes

fix ubsan use int32_t pointer access bool value is ub:
```
 load of misaligned address 0x7fff94267926 for type 'const int32_t', which requires 4 byte alignment
```

result is error


HappenLee:
run buildall

Mryange (approved):

yiguolei (approved):

PR #38244, opened 2024-07-23
[refactor](catalog) set "use_meta_cache" default to true

morningman:
This is PR #33610 introduce a new feature of `use_meta_cache=true`.
Now I will set this property default to true.
For all newly created catalog, is this property is not specified, it will be true.
For all previously created catalog, this property is still false.

PR #38619, opened 2024-07-31
[fix](multicatalog) fix metacache is null issue when show_dbs

Yulei-Yang:
Brought by: #38244

how to reproduce:
create a hms catalog named hive with property use_meta_cache = false, after that restart fe, and run sql `switch hive; show databases;`, will get error：
```sql
ERROR 1105 (HY000): NullPointerException, msg: java.lang.NullPointerException: Cannot invoke "org.apache.doris.datasource.metacache.MetaCache.listNames()" because "this.metaCache" is null
```

reason, property use_meta_cache of newly created hms catalog after pr #38244 was force convert to true, but field  `useMetaCache` of class ExternalCatalog is not converted at the same time.


Yulei-Yang:
run buildall

Yulei-Yang:
run cloud_p0

morningman (approved):
LGTM

wuwenchi (approved):

PR #38627, opened 2024-07-31
[config](array) add config for create inverted index on array type

amorynan:
## Proposed changes
same action from master
Issue Number: close #xxx


amorynan:
run buildall

PR #38562, opened 2024-07-31
[Fix](memory) Fix allocator.h compiling failed on mac.

kaka11chen:
## Proposed changes

Fix allocator.h compiling failed on mac which introduced by #37234.

PR #37257, opened 2024-07-04
[Feature](multi-catalog) Add memory tracker for orc reader/writer and arrow parquet writer。

kaka11chen:
## Proposed changes

backport #37234

PR #38646, opened 2024-08-01
[Fix](memory) Fix allocator.h compiling failed on mac.

kaka11chen:
Backport #38562. Fix allocator.h compiling failed on mac which introduced by #37257.


kaka11chen:
run buildall

PR #36209, opened 2024-06-12
[Fix](inverted index) fix memory leak for index compaction

airborne12:
## Proposed changes

Issue Number: close #xxx

Need to delete source directory and dest directory when doing index compaction

PR #38654, opened 2024-08-01
[Fix](inverted index) fix memory leak for index compaction #36209

airborne12:
cherry pick from #36209


airborne12:
run buildall

PR #38659, opened 2024-08-01
[fix](group commit) Pick add debug log show why group commit not work; delete wal when replay success (#38611)

mymeiyi:
Pick https://github.com/apache/doris/pull/38611


mymeiyi:
run buildall

PR #38663, opened 2024-08-01
[fix](regression)Set enable_partition_analyze session back to false in finally block.

Jibing-Li:
Set enable_partition_analyze session back to false in finally block. Otherwise, enable_partition_analyze will stay in true when case fails.


Jibing-Li:
run buildall

morrySnow (approved):

zddr (approved):

PR #38501, opened 2024-07-29
[fix](test)Fix unstable index compaction fault injection case

qidaye:
And add index_storage_format = v1 to index compaction cases

PR #38673, opened 2024-08-01
[fix](test)Fix unstable index compaction fault injection case (#38501)

qidaye:
## Proposed changes

bp #38501


qidaye:
run buildall

PR #36812, opened 2024-06-25
[enhancement](index compaction) Enable index compaction by default

qidaye:
## Proposed changes

Issue Number: close #xxx

PR #38676, opened 2024-08-01
[enhancement](index compaction) Enable index compaction by default (#36812)

qidaye:
## Proposed changes

bp #36812


qidaye:
run buildall

PR #38687, opened 2024-08-01
[fix](test) fix `test_numbers` case

BePPPower:
we use `order_qt` rather than `qt` to promise the order of results.


BePPPower:
run buildall

morningman (approved):
LGTM

wuwenchi (approved):

PR #38692, opened 2024-08-01
[feature](StorageVault) Support setting path format for storage vault

platoneko:
## Proposed changes

Support setting path format for storage vault


platoneko:
run buildall

PR #38657, opened 2024-08-01
[fix](local exchange) Fix local exchange blocked by a huge data block

Gabriel39:
## Proposed changes

If a huge block is push into local exchanger, it will be blocked due to concurrent problems. This PR use a unique lock to resolve it .

PR #38693, opened 2024-08-01
[fix](local exchange) Fix local exchange blocked by a huge data block…

Gabriel39:
… (#38657)

If a huge block is push into local exchanger, it will be blocked due to concurrent problems. This PR use a unique lock to resolve it .

## Proposed changes

Issue Number: close #xxx


Gabriel39:
run buildall

PR #38694, opened 2024-08-01
[pipeline](fix) Fix blocking task which is not triggered by 2nd RPC (…

Gabriel39:
…#38568)

Once a query is cancelled due to any reason, BE may not receive 2nd RPC from FE. If so, we must ensure the execution dependency is ready so tasks will not be blocked.

## Proposed changes

Issue Number: close #xxx


Gabriel39:
run buildall

PR #38698, opened 2024-08-01
[fix](mtmv)fix second level MTMV will refresh all partition

zddr:
if:
-  create mv2 as select * from mv1;
- create mv1 as select * from t1;
- t1 has 2 partitions: p1,p2

when t1 insert data to p1;

mv1 will refresh p1

mv2 will refresh p1 and p2

fix:
should not check if data is sync between mv2 and t1


zddr:
run buildall

morrySnow (approved):

seawinde (approved):

PR #38712, opened 2024-08-01
[fix](array_range) fix array_range func for large param which should return error #38284

amorynan:
## Proposed changes
backport: https://github.com/apache/doris/pull/38284
Issue Number: close #xxx


amorynan:
run buildall

PR #38718, opened 2024-08-01
[bug](parquet)Fix the problem that the parquet reader reads the missing sub-columns of the struct and fails.

hubgeter:
## Proposed changes

Fix the problem that the parquet reader reads the missing sub-columns of the struct and fails.
for example : 
 suppose we have a column  `array<struct<a:int>>` .  And this column has two data
```
[{1},{2},{3}]
[{4},{5}]
```
Then we add a subcolumn b to the struct . Now the column structure is like this `array<struct<a:int,b:string>>`
The expected data for the query is as follows, instead of an error  :
 
```
[{1,null},{2,null},{3,null}]
[{4,null},{5,null}]
```


hubgeter:
run buildall

hubgeter:
run p0

morningman (approved):
LGTM

kaka11chen (approved):
LGTM

PR #38731, opened 2024-08-01
[fix](function) stddev with DecimalV2 type will result in an error

Mryange:
## Proposed changes

The stddev function has a separate implementation for the DecimalV2 type, but there are issues with the implementation. Given that there is almost no existing data for DecimalV2, it will be removed here. For be, upgrading to this situation will result in an error directly.
```
SELECT STDDEV(data) FROM DECIMALV2_10_0_DATA;
ERROR 1105 (HY000): errCode = 2, detailMessage = (127.0.0.1)[INTERNAL_ERROR]Agg Function stddev(decimal(10,0)) is not implemented
```
After removing DecimalV2, parameters of type DecimalV2 will be converted to double for calculations.


Mryange:
run buildall

Mryange:
run buildall

HappenLee (approved):
LGTM

Gabriel39 (approved):

PR #38737, opened 2024-08-01
[Bug](fix) fix coredump case in (not null, null) execpt (not null, t null) case

HappenLee:
## Proposed changes

fix coredump case in (not null, null) execpt (not null, t null) case， column not null insert  from nullable column cause coredump


HappenLee:
run buildall

zhangstar333 (approved):

yiguolei (approved):

PR #38756, opened 2024-08-02
[Bug](fix) fix coredump case in (not null, null) execpt (not null, not null) case

HappenLee:
## Proposed changes

Issue Number: close #38612


HappenLee:
run buildall

PR #38732, opened 2024-08-01
[fix](index compaction)Remove return error in merged rows check

qidaye:
## Proposed changes

Log here just for debugging, do not return error.

PR #38757, opened 2024-08-02
[fix](index compaction)Remove return error in merged rows check #38732

qidaye:
## Proposed changes

backport #38732


qidaye:
run buildall

PR #38779, opened 2024-08-02
[Fix](Nereids) remove some unstable and not important cases

LiBinfeng-01:
Problem:
test_distribute is test for shuffle/broadcast hint, but it can not be sure to force fixed the shape of plan. So it may be cause some unstable problem when use multi instances of be or unstable due to the cost of network
Solved:
remove it temperately until find more reliable way to fixed the shape of plan completely regardless of volatility of environment


LiBinfeng-01:
run buildall

englefly (approved):

starocean999 (approved):

PR #38781, opened 2024-08-02
[fix](recycler) Fix uninitialized num_entries when list a non-exist directory (#38782)

platoneko:
## Proposed changes

`hdfsListDirectory` won't set `num_entries` when it returns nullptr


platoneko:
run buildall

PR #38556, opened 2024-07-31
[opt](join) check datatype of intermediate slots in hash join

mrhhsg:
## Proposed changes

Issue Number: close #xxx

PR #38792, opened 2024-08-02
[opt](join) check datatype of intermediate slots in hash join (#38556)

mrhhsg:
## Proposed changes

pick #38556


mrhhsg:
run buildall

PR #38799, opened 2024-08-02
[fix](gson) Fix Expr deserialize compatibility

w41ter:
Since GsonBuilder.create() adds all registered factories to GSON in reverse order, and ExprAdapterFactory is registered before the RuntimeTypeAdapterFactory for Expr, ExprAdapterFactory will not be executed.

This PR adjusts their registration order. Now, it will first check-in ExprAdapterFactory whether to use the pre-134 deserialize method and then attempt to use the RuntimeTypeAdapterFactory for Expr.class.


w41ter:
run buildall

dataroaring (approved):
LGTM

gavinchou (approved):

PR #38808, opened 2024-08-02
[fix](fs) Use full path as file cache key

platoneko:
## Proposed changes

In path v1, the segment file path is "data/${shard_id}/${tablet_id}/${rowset_id}/${seg_id}.dat", which means that segment files of different rowsets can have the same file name (e.g. "0.dat"). However, the initial file cache key is based on the file name, which will lead to cache key collisions. Additionally, to maintain the generality of `CachedRemoteFileReader`, we should not assume that all files accessed by `CachedRemoteFileReader` have globally unique file names. Therefore, it is necessary to use the full path as the cache key.


platoneko:
run buildall

be/src/io/cache/block_file_cache.cpp
gavinchou, line 1: Add UT to cover all newly added branches, make sure nothing is missed.

gavinchou (changes requested):

PR #38402, opened 2024-07-26
[fix](routine load) fix enclose and escape can not set in routine load job

sollhui:
Fix enclose and escape can not set in routine load job.

PR #38825, opened 2024-08-04
[fix](routine load) fix enclose and escape can not set in routine load job (#38402)

sollhui:
pick (#38402)


sollhui:
run buildall

PR #38833, opened 2024-08-04
[enhance](server) add metrics to show thread cpu usage

cosven:
## Proposed changes

Add a metric `thread_cpu_total` to show cpu usage by each thread, which improves observability.

Before, when a instance costs too much CPU, user does not know which task consumes it. User needs to run `top -H` on the machine to see the cpu usage details. Now, user just needs a simple check of the metrics.

Major implementation changes
1. add `ThreadMgr::instance` static method and `ThreadMgr::update_threads_metrics` method
2. add three methods for `ThreadMgr::ThreadDescriptor` class: `register_metric`, `update_metric`, `deregister_metric`
3. call `update_threads_metrics` to update threads metrics periodically `calculate_metrics_thread `

## Usage example

User can query thread pool usage by the following promql
image

For example, check the cpu usage of `CumuCompaction` thread pool 
image

### DIFF
--- a/be/src/common/daemon.cpp
+++ b/be/src/common/daemon.cpp
@@ -67,9 +67,9 @@
 #include "util/network_util.h"
 #include "util/perf_counters.h"
 #include "util/system_metrics.h"
+#include "util/thread.h"
 #include "util/thrift_util.h"
 #include "util/time.h"
-#include "util/thread.h"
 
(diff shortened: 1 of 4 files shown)


cosven:
run performance

cosven:
run buildall

PR #38356, opened 2024-07-25
[fix](compaction) fix missing flush column during compaction

luwei16:
#37960 introduces this problem

PR #38251, opened 2024-07-23
[fix](compaction) fix the issue of writing segment with 0 rows during compaction

luwei16:
https://github.com/apache/doris/pull/37960 introduced this issue

PR #37960, opened 2024-07-16
[fix](compaction) fix mismatch between segment key and value column rows during compaction

luwei16:
When a block is splitted to 3 segments, old code just handles 2 and the last is overlowed.

PR #38835, opened 2024-08-04
[fix](compaction) fix mismatch between segment key and value column rows during compaction (#37960)(#38251)(#38356)

luwei16:
pick master #37960 #38251 #38356


luwei16:
run buildall

luwei16:
run buildall

PR #38842, opened 2024-08-05
[fix](txn) fix abortTxn by label does not acquire table write lock (#38777)

mymeiyi:
pick https://github.com/apache/doris/pull/38777


mymeiyi:
run buildall

PR #38850, opened 2024-08-05
[fix](regression) fix prepare_insert when execute prepare stmt in observer fe (#38545)

mymeiyi:
pick https://github.com/apache/doris/pull/38545


mymeiyi:
run buildall

PR #38855, opened 2024-08-05
[fix](testcase) add order by to fix unstable output of passwordLeaked #38813

xiaokang:
cherry pick from #38813


xiaokang:
run buildall

PR #38857, opened 2024-08-05
[fix asan core](memoryleak) should deconstruct filter entity in scope of query context

yiguolei:
## Proposed changes

Merge controller operation may create new memory and save in controller's object pool. So that should deconstruct the shared ptr in scope of query context to avoid asan check core.


yiguolei:
run buildall

BiteTheDDDDt (approved):

xinyiZzz (approved):
LGTM

PR #38859, opened 2024-08-05
[enhancement](querycancel) print detail message when query is cancelled

yiguolei:
## Proposed changes

Issue Number: close #xxx


yiguolei:
run buildall

PR #38866, opened 2024-08-05
[fix](function) Results for stddev with nan and inf are unstable.

Mryange:
## Proposed changes

This is a very strange issue.
```
mysql [test10]>set parallel_pipeline_task_num = 1;

mysql [test10]>SELECT STDDEV(t.ARG0) FROM (SELECT TEMPDATA . data, TABLE0.ARG0 FROM TEMPDATA CROSS JOIN (SELECT data AS ARG0 FROM DOUBLEDATA) AS TABLE0) t;
+--------------+
| stddev(ARG0) |
+--------------+
|         NULL |
+--------------+

mysql [test10]>set parallel_pipeline_task_num = 48;

mysql [test10]>SELECT STDDEV(t.ARG0) FROM (SELECT TEMPDATA . data, TABLE0.ARG0 FROM TEMPDATA CROSS JOIN (SELECT data AS ARG0 FROM DOUBLEDATA) AS TABLE0) t;
+--------------+
| stddev(ARG0) |
+--------------+
|          inf |
+--------------+
```
This function performs squaring operations, and due to differences in computation order, it might produce different values such as inf and nan. In MySQL, this will directly result in an error due to exceeding the double range. For performance reasons, we are uniformly changing it to nan


Mryange:
run buildall

HappenLee (approved):
LGTM

Gabriel39 (approved):

PR #38876, opened 2024-08-05
[fix](hist) Fix unstable result of aggregrate function hist #38608

zhiqiang-hhhh:
cherry pick from #38608


zhiqiang-hhhh:
run buildall

PR #38878, opened 2024-08-05
[fix](multi table) fix single stream multi table access denied in cloud mode

sollhui:
Create single stream multi table job to load Kafka stream, error happened:
```
failed to get stream load plan, errCode = 2, detailMessage = Access denied for user
'doris_test@null' (using password: YES) '}
```

It is do not need check password for requests initiated internally for already check it when connection.


sollhui:
run buildall

dataroaring (approved):
LGTM

liaoxin01 (approved):
LGTM

PR #38882, opened 2024-08-05
[improvement](mtmv) Support to use view when create mtmv

seawinde:
## Proposed changes

Support to use mv when create async materialized view
Such as view def is as following:
```sql
        create view view1 as
            select o_shippriority, o_comment,
            count(distinct case when o_shippriority > 1 and o_orderkey IN (1, 3) then o_custkey else null end) as cnt_1,
            count(distinct case when O_SHIPPRIORITY > 2 and o_orderkey IN (2) then o_custkey else null end) as cnt_2,
            sum(o_totalprice) as sum_1,
            max(o_totalprice) as max_1,
            min(o_totalprice) as min_1,
            count(*)
            from orders
            where o_orderdate = '2023-12-09'
            group by
            o_shippriority,
            o_comment;
```

Can create materialized view like following:
```sql
        CREATE MATERIALIZED VIEW mv1_name
        BUILD IMMEDIATE REFRESH AUTO ON MANUAL
        DISTRIBUTED BY RANDOM BUCKETS 2
        PROPERTIES (
        'replication_num' = '1'
        )
        AS
        SELECT cnt_1, cnt_2, sum_1, min_1 from view1;
```


seawinde:
run buildall

seawinde:
run buildall

seawinde:
run buildall

seawinde:
run buildall

seawinde:
run buildall

morrySnow (dismissed):

zddr:
`org.apache.doris.mtmv.MTMVPlanUtil#getBaseViews` now will return empty set.
We need to implement it or remove the baseViews variable from MTMVRelation if not needed

seawinde:
if view in mtmv, we always expand the view in analyze. So when query rewrite by materialized view or refresh the materialized view, the view doesn't exist.
if we want to support rewritting by materialized view based view, then we consider the `MTMVPlanUtil#getBaseViews`

seawinde:
run buildall

seawinde:
run buildall

fe/fe-core/src/main/java/org/apache/doris/mtmv/MTMVPlanUtil.java
```diff
@@ -83,21 +81,30 @@ public static MTMVRelation generateMTMVRelation(MTMV mtmv, ConnectContext ctx) {
     }
 
     public static MTMVRelation generateMTMVRelation(Plan plan) {
-        return new MTMVRelation(getBaseTables(plan, true), getBaseTables(plan, false), getBaseViews(plan));
+        return new MTMVRelation(getBaseTables(plan, true, true),
+                getBaseTables(plan, false, false), getBaseViews(plan));
```
zddr, line 85: should explandView？otherwise，data change in baseTable,mv will not refresh auto

fe/fe-core/src/main/java/org/apache/doris/nereids/rules/exploration/mv/InitMaterializationContextHook.java
```diff
@@ -78,7 +78,8 @@ public void initMaterializationContext(CascadesContext cascadesContext) {
      * @param cascadesContext current cascadesContext in the planner
      */
     protected void doInitMaterializationContext(CascadesContext cascadesContext) {
-        TableCollectorContext collectorContext = new TableCollectorContext(Sets.newHashSet(), true);
+        TableCollectorContext collectorContext = new TableCollectorContext(Sets.newHashSet(),
+                true, true);
```
zddr, line 82: ditto

seawinde: For query plan, this should not extend, we only get avalible mvs by one level table or view. this is for performance

seawinde:
run buildall

PR #38883, opened 2024-08-05
[fix](group commit) fix repaly wal check label status

mymeiyi:
## Proposed changes

When replay wal, it firstly abort the txn with the label but does not check the abort result.
And when begin txn of replay, if FE returns `LabelAlreadyUsedException`, it consider the load is success in previous group commit load or repaly wal, and delete this wal directly.
But `LabelAlreadyUsedException` means there is a txn with this label, the txn may be in `PREPARE / RUNNING / COMMITTED / VISIBLE` status(the abort txn in first step may fail), so replay wal should check both `LabelAlreadyUsedException` and txn status is `COMMITTED / VISIBLE`.

This pr also add a case for replay wal with schema change.


mymeiyi:
run buildall

mymeiyi:
run buildall

dataroaring (approved):
LGTM

bobhan1 (approved):

Yukang-Lian (approved):

PR #38890, opened 2024-08-05
[Chore](pipeline) set PipelineFragmentContext::_timeout

Gabriel39:
## Proposed changes

Now we use `query_timeout` to set a timeout value for queries. But for pipelineX engine, Doris do not use it so each query will not end before EOS. This PR fix it.

pick #35328


Gabriel39:
run buildall

PR #38899, opened 2024-08-05
[fix](inverted index) Fix the issue where the inverted index file_cache was not being counted

zzzxl1993:
## Proposed changes

1. inverted index file_cache profile


zzzxl1993:
run buildall

PR #38902, opened 2024-08-05
[Fix](multi-catalog) Fix not throw error when call close() in hive/iceberg writer.

kaka11chen:
## Proposed changes

Backport #38987.

[Fix] (multi-catalog) Fix not throw error when call close() in hive/iceberg writer. 

When the file writer closes(), it will sync buffer to commit. Therefore, sometimes data is written only when close() is called, which can expose some errors. For example, hdfs_file_writer. Therefore, this error needs to be captured in the entire close process.


kaka11chen:
run buildall

kaka11chen:
run buildall

PR #38885, opened 2024-08-05
[fix](pipeline) Fix mem control in local exchanger

Gabriel39:
## Proposed changes

If a block (>128M) is dequeue by local exchange source operator and it is the last block, both of source operators and sink operators will be hang. This PR fixed it.

PR #38910, opened 2024-08-06
[fix](pipeline) Fix mem control in local exchanger (#38885)

Gabriel39:
## Proposed changes

If a block (>128M) is dequeue by local exchange source operator and it is the last block, both of source operators and sink operators will be hang. This PR fixed it.

pick #38885

## Proposed changes

Issue Number: close #xxx


Gabriel39:
run buildall

PR #35313, opened 2024-05-23
[fix](group commit) Fix test_group_commit_async_wal_msg_fault_injection case

mymeiyi:
## Proposed changes

The test_group_commit_async_wal_msg_fault_injection case may get:

```
terminate called after throwing an instance of 'std::future_error'
  what():  std::future_error: Promise already satisfied
*** Query id: 68dda9ae3c7e4815-aeac2090983b45e2 ***
*** tablet id: 0 ***
*** Aborted at 1716427902 (unix time) try "date -d @1716427902" if you are using GNU date ***
*** Current BE git commitID: 3175a62a849 ***
*** SIGABRT unknown detail explain (@0x1e5a) received by PID 7770 (TID 8350 OR 0x7f0d3e7a5700) from PID 7770; stack trace: ***
 0# doris::signal::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*) at /mnt/disk3/pipeline/repo/selectdb-core_selectdb-cloud-dev/selectdb-core/be/src/common/signal_handler.h:435
 1# 0x00007F0E6DBF7090 in /lib/x86_64-linux-gnu/libc.so.6
 2# raise at ../sysdeps/unix/sysv/linux/raise.c:51
 3# abort at /build/glibc-SzIz7B/glibc-2.31/stdlib/abort.c:81
 4# __gnu_cxx::__verbose_terminate_handler() [clone .cold] at ../../../../libstdc++-v3/libsupc++/vterminate.cc:75
 5# __cxxabiv1::__terminate(void (*)()) at ../../../../libstdc++-v3/libsupc++/eh_terminate.cc:48
 6# 0x0000565170AE5721 in /mnt/disk1/selectdb-cloud-dev-asan/cluster0/be/lib/doris_be
 7# 0x0000565170AE5874 in /mnt/disk1/selectdb-cloud-dev-asan/cluster0/be/lib/doris_be
 8# std::__throw_future_error(int) at ../../../../../libstdc++-v3/src/c++11/future.cc:76
 9# std::__future_base::_State_baseV2::_M_set_result(std::function ()>, bool) at /var/local/ldb_toolchain/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/future:411
10# std::promise::set_value(doris::Status&&) at /var/local/ldb_toolchain/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/future:1137
[...]

PR #38911, opened 2024-08-06
[fix](group commit) Fix test_group_commit_async_wal_msg_fault_injection case (#35313)

mymeiyi:
pick https://github.com/apache/doris/pull/35313


mymeiyi:
run buildall

PR #38924, opened 2024-08-06
[chore](sql) Forbid show hidden columns and create table with hidden column (#38796)

924060929:
cherry pick from #38796


924060929:
run buildall

PR #38891, opened 2024-08-05
[fix](index) fix inverted index compound file entry size int32 overflow

xiaokang:
change FileInfo::filesize from int32_t to int64_t to avoid overflow

PR #38928, opened 2024-08-06
[fix](index) fix inverted index compound file entry size int32 overflow #38891

xiaokang:
pick master  #38891


xiaokang:
run buildall

PR #38475, opened 2024-07-29
[fix](multicatalog) fix hadoop authenticator not inited for existing hms catalog

Yulei-Yang:
authenticator for hms type catalog has been refactored to be initted only on creating，but existing catalog are not initted. 

Issue:
```sql
switch hive;
show databases;
ERROR 1105 (HY000): HMSClientException, msg: java.lang.NullPointerException: Cannot invoke "org.apache.doris.common.security.authentication.HadoopAuthenticator.doAs(java.security.PrivilegedExceptionAction)" because "this.hadoopAuthenticator" is null
```

PR #38930, opened 2024-08-06
[fix](multicatalog) fix hadoop authenticator not inited for existing hms catalog.

kaka11chen:
Backport #38475.


kaka11chen:
run buildall

PR #38907, opened 2024-08-06
[Fix](inverted index) fix missed array inverted index null bitmap

airborne12:
## Proposed changes

We forgot to set null bitmap in inverted index for array values.
In this PR, we added it.

PR #38933, opened 2024-08-06
[Fix](inverted index) fix missed array inverted index null bitmap #38907

airborne12:
cherry pick from #38907


airborne12:
run buildall

PR #38085, opened 2024-07-18
[Improvement](runtime-filter) use shared ptr to instead object pool to store runtime filters

BiteTheDDDDt:
## Proposed changes
use shared ptr to instead object pool to store runtime filters

PR #38965, opened 2024-08-06
[Bug](rf) fix rf of minmax cast data as string_ref maybe return wrong result

zhangstar333:
## Proposed changes
 this PR: https://github.com/apache/doris/pull/38085 have refactor the store of min/max in RF.
and now will be use std::string to store data, so if still reinterpret_cast<const StringRef*> maybe get wrong data.

Not merge to 2.1, because 38085 is not merged


zhangstar333:
run buildall

yiguolei (approved):

BiteTheDDDDt (approved):

PR #38982, opened 2024-08-06
[fix](cancel) Fix wrong status when query failed

zhiqiang-hhhh:
Pipeline task A failed with `Status A`, pipeline task B should also fail with same `Status A` instead of just returning a CANCELLED.


zhiqiang-hhhh:
run buildall

yiguolei (approved):

zclllyybb (approved):

PR #38992, opened 2024-08-07
[fix](MoW) fix MoW & segcompaction conflict on cache of temp segment …

dataroaring:
…(#37760)

MoW will update delete bitmap during load, and the page cache could be modified by segcompaction. Disable page cache touchs when doing segcompaction could solve this problem.

## Proposed changes

Issue Number: close #xxx

## Proposed changes

Issue Number: close #xxx


dataroaring:
run buildall

PR #38997, opened 2024-08-07
[fix](group commit) fix repaly wal check label status (#38883)

mymeiyi:
pick https://github.com/apache/doris/pull/38883


mymeiyi:
run buildall

PR #39018, opened 2024-08-07
[Fix](merge-on-write) Fix FE may use the staled response to wrongly commit txn

bobhan1:
## Problem

consider the following scenarios for merge-on-write table in cloud mode
### Scenario 1: Load-Load Conflict
1. load txn1 tries to commit version n and gets the delete bitmap update lock
2. load txn1 begins to calculate delete bitmap on BEs, this is a heavy calculating process and lasts long
3. load txn2 tries to commit version n and gets the delete bitmap update lock because load txn1's delete bitmap update lock has expired
4. load txn1's delete bitmap update lock expires and load txn2 get the delete bitmap update lock
5. load txn2 commits successfully with version n and release the delete bitmap update lock
6. load txn1 fails to commit due to timeout of the calculation of delete bitmap
7. load txn1 retries the commit process with version n+1, gets the bitmap update lock and sends delete bitmap calculation task to BEs
8. BE fails to register this new calculation task because there is a task with the same signatrure(txn_id) running in the task_worker_pool
9. BE finishes the calculation of delete bitmap and report success status to FE
10. load txn1 commits successfully with n+1

Finally, load txn1 failed to calculate delete bitmap for version n from load txn2
### Scenario 2: Load-Compaction Conflict
1. load txn tries to commit and gets the delete bitmap update lock
2. load txn collects rowset_ids and submit a delete bitmap calculation task to the threadpool for the diff rowsets. But the theadpool is full, so the task is queued in the threadpool.
3. load txn's delete bitmap update lock expired and a compaction job on the same tablet finished successfully.
4. load txn fails to commit due to timeout of the calculation of delete bitmap
5. load txn retries the commit process, gets the bitmap update lock and sends delete bitmap calculation task to BEs
6. BE fails to register this new calculation task because there is a task with the same signatrure(txn_id) running in the task_worker_pool
7. BE finishes the calculation of delete bitmap and report success status to FE
8. load txn1 commits successfully

Finally, load txn failed to calculate delete bitmap for the compaction produced by compaction

## Solution
The root cause of the above failures is that when the commit process is retried many times, FE may use the previous stale success response from BEs and commit txns. One solution for that problem is that FE attaches an unique id within the delete bitmap calculation task sent to BE and BE takes it in the response for FE to check if the response is for the current latest task. However, since `calculate_delete_bitmap_task_timeout_seconds` can not change adaptively based on actual computation time currently, if the delete bitmap calculation always consumes more time than the timeout of the delete bitmap calculation, FE will retry the commit process infinitely which causes live lock.

This PR let the BE's response take the compaction stats(to avoid load-compaction conflict) and versions(to avoid load-load conflict) from the task request and let the FE compares it with the current task's to know that if there is any compaction or load finished during the time periods since the current load get the delete bitmap lock due to lock expiration. If so, the current txn should retry or abort. If not, the current txn can commit successfully.


bobhan1:
run buildall

zhannngchen (approved):
LGTM

Yukang-Lian (approved):

zhannngchen (changes requested):
add regression cases

regression-test/suites/fault_injection_p0/cloud/test_cloud_mow_stale_resp_load_load_conflict.groovy
```diff
@@ -0,0 +1,89 @@
...
+
+        // the first load
+        t1 = Thread.start {
+            sql "insert into ${table1} values(1,999,999),(2,888,888);"
+        }
+
+        // wait util the first load's delete bitmap update lock expired
+        // to ensure that the second load can take the delete bitmap update lock
+        // Config.delete_bitmap_lock_expiration_seconds = 10s
+        Thread.sleep(11 * 1000)
```
zhannngchen, line 58: you'd better to set Config.delete_bitmap_lock_expiration_seconds independently here, since we plan to enlarge the default value of this config soon, which may cause the case fail
you can try `setFeConfigTemporary`?

bobhan1:
run buildall

zhannngchen (dismissed):
LGTM

bobhan1:
run buildall

zhannngchen (approved):
LGTM

bobhan1:
run performance

PR #39026, opened 2024-08-07
[Bug](rf) fix rf of in filter cast data as different class type maybe return wrong result

zhangstar333:
two point have changed:
1. in batch_assign function: 
const std::string& string_value = column.stringval();
if call **insert(&string_value)**, will cast as string_ref: reinterpret_cast<const **StringRef***>(data), this maybe error;
```
    void insert(const void* data) override {
        if (data == nullptr) {
            _contains_null = true;
            return;
        }

        const auto* value = reinterpret_cast<const StringRef*>(data);
        std::string str_value(value->data, value->size);
        _set.insert(str_value);
    }
```

2. in batch_copy function, will cast void_value as T*
but the it->get_value() return is StringRef, so need change T as StringRef
```
template <typename T>
void batch_copy(PInFilter* filter, HybridSetBase::IteratorBase* it,
                void (*set_func)(PColumnValue*, const T*)) {
    while (it->has_next()) {
        const void* void_value = it->get_value();
        auto origin_value = reinterpret_cast<const T*>(void_value);
        set_func(filter->add_values(), origin_value);
        it->next();
    }
}
```

## Proposed changes

Issue Number: close #xxx


zhangstar333:
run buildall

HappenLee (approved):
LGTM

BiteTheDDDDt (approved):

PR #39044, opened 2024-08-07
[Fix](lakesoul) Fix CVEs and some filter pushdown issues

xuchen-plus:
## Proposed changes

1. Addressed CVEs caused by hadoop and spark. Unused packages are excluded from LakeSoul's depdencies.
2. Fix som filter pushdown issues.


CalvinKirs:
run buildall

fe/be-java-extensions/lakesoul-scanner/pom.xml
```diff
@@ -48,68 +48,18 @@ under the License.
...
-                    <groupId>commons-logging</groupId>
-                    <artifactId>commons-logging</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>com.google.code.findbugs</groupId>
-                    <artifactId>jsr305</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>org.apache.spark</groupId>
+                    <groupId>*</groupId>
```
CalvinKirs, line 58: If the classifier is set to "shaded," then the following exclusions will be meaningless, as the shaded jar itself will contain all the dependencies.

fe/fe-core/pom.xml
```diff
@@ -565,58 +565,87 @@ under the License.
...
+            <groupId>org.postgresql</groupId>
+            <artifactId>postgresql</artifactId>
+            <version>42.7.3</version>
+        </dependency>
+        <dependency>
+            <groupId>org.scala-lang</groupId>
+            <artifactId>scala-library</artifactId>
+            <version>${scala.version}</version>
+            <scope>compile</scope>
+        </dependency>
```
CalvinKirs, line 642: why do we need to include these jars? Is it necessary?

fe/be-java-extensions/lakesoul-scanner/pom.xml
xuchen-plus: Because lakesoul ship both original and shaded jar, the pom.xml file of lakesoul still includes all the dependencies even though the classifier is shaded. That's why we need to exclude all the dependencies here.

fe/fe-core/pom.xml
xuchen-plus: 1. postgresql is a dependency of lakesoul, and its version in lakesoul is 42.7.3. However in FE, the version will be overriden to 42.3.8 (by spring-boot-dependencies I believe), which has CVE issues. So We override the version again here.
2. scala-library is required by lakesoul in both compile and runtime, but its original scope is provided. So we override the scope here.
3. antlr4 declaration here is unnecessary, I'll remove it and push again.

morningman (approved):
LGTM

morningman:
run buildall

xinyiZzz (approved):
LGTM

PR #39056, opened 2024-08-07
[Improment]support drop be tag

wangbo:
## Proposed changes
1 support remove backend property;
```
alter system modify backend "localhost:9250" set ("tag.workload_group"="");
```

2 support modify backend property without carrying all properties;
```
alter system modify backend "localhost:9250" set ("tag.workload_group"="g1");
```
Apart from role and location, other tags can be removed by set null value, because role and location not only exits in tagMap but also in Backend class's field.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/ModifyBackendClause.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/ModifyBackendClause.java
@@ -20,7 +20,6 @@ package org.apache.doris.analysis;
 import org.apache.doris.common.AnalysisException;
 import org.apache.doris.common.Config;
 import org.apache.doris.common.util.PropertyAnalyzer;
-import org.apache.doris.resource.Tag;
 
 import com.google.common.collect.Maps;
 import lombok.Getter;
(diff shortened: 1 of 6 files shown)


wangbo:
run buildall

fe/fe-core/src/main/java/org/apache/doris/analysis/ModifyBackendClause.java
```diff
@@ -20,7 +20,6 @@
 import org.apache.doris.common.AnalysisException;
 import org.apache.doris.common.Config;
 import org.apache.doris.common.util.PropertyAnalyzer;
-import org.apache.doris.resource.Tag;
```
yiguolei, line 23: please add many many tests for this PR. Because it is very important.

fe/fe-core/src/test/java/org/apache/doris/clone/TabletRepairAndBalanceTest.java
```diff
@@ -180,6 +180,64 @@ private static void alterTable(String sql) throws Exception {
         Env.getCurrentEnv().getAlterInstance().processAlterTable(alterTableStmt);
     }
 
+    @Test
+    public void test_alter_be_tag() throws Exception {
+        Assert.assertEquals(5, backends.size());
```
yiguolei, line 185: java 里是驼峰，不是下划线的风格

wangbo:
run buildall

wangbo:
run buildall

PR #39064, opened 2024-08-07
[fix](expr) Enhance SQL Expression Handling by Introducing printSqlInParens to CompoundPredicate

zy-kkk:
This PR enhances SQL expression handling by introducing a printSqlInParens flag to the CompoundPredicate class and setting it to true. This ensures that expressions in a CompoundPredicate are always enclosed in parentheses, similar to how it is handled in a BinaryPredicate. . By removing redundant handling of CompoundPredicate, the logic for handling SQL generation in the JdbcScanNode node has been simplified as the updated toSql method now handles this uniformly across expressions.


zy-kkk:
run buildall

fe/fe-core/src/main/java/org/apache/doris/analysis/CompoundPredicate.java
```diff
@@ -94,7 +94,26 @@ public String toSqlImpl() {
             Preconditions.checkState(op == Operator.NOT);
             return "NOT " + getChild(0).toSql();
         } else {
-            return getChild(0).toSql() + " " + op.toString() + " " + getChild(1).toSql();
+            String left = getChild(0).toSql();
```
morningman, line 97: why not adding `()` for both left and right?
like: `(left) op (right)`

zy-kkk:
run buildall

zy-kkk:
run buildall

zy-kkk:
run buildall

zy-kkk:
run buildall

zy-kkk:
run cloud_p0

morningman (approved):
LGTM

CalvinKirs (approved):

PR #38331, opened 2024-07-24
[Fix](partial update) Persist partial_update_info in RocksDB in case of BE restart after a partial update has commited

bobhan1:
## Proposed changes
If a partial update has conflict with another load during publish phase, it should combine the two load's data into one to get the corrrect result. This procedure needs partial update info. But If BE crashed after the partial update load has committed, the partial update info will be missing becasuse it's not persisted and will not be restored in `DataDir::load()`. This PR persists partial update info in RocksDB before the txn is commited and remove it after the publish phase.
Before https://github.com/apache/doris/pull/25147, partial update info is persisted with tablet_schema in RocksDB. https://github.com/apache/doris/pull/25147 split partial update info from tablet schema but forget to handle the persistence logic.

branch-2.1-pick: https://github.com/apache/doris/pull/39035
branch-2.0-pick: https://github.com/apache/doris/pull/39078

PR #39066, opened 2024-08-08
[Fix](merge-on-write) Fix UBSan error cause by member call on null pointer

bobhan1:
## Proposed changes

fix member call on null pointer in https://github.com/apache/doris/pull/38331
image

### DIFF
--- a/be/src/olap/txn_manager.cpp
+++ b/be/src/olap/txn_manager.cpp
@@ -747,8 +747,13 @@ void TxnManager::force_rollback_tablet_related_txns(OlapMeta* meta, TTabletId ta
             }
         }
     }
-    static_cast<void>(
-            RowsetMetaManager::remove_tablet_related_partial_update_info(meta, tablet_id));
+    if (meta != nullptr) {
+        Status st = RowsetMetaManager::remove_tablet_related_partial_update_info(meta, tablet_id);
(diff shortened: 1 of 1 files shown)


bobhan1:
run buildall

be/src/olap/txn_manager.cpp
```diff
@@ -747,8 +747,10 @@ void TxnManager::force_rollback_tablet_related_txns(OlapMeta* meta, TTabletId ta
             }
         }
     }
-    static_cast<void>(
-            RowsetMetaManager::remove_tablet_related_partial_update_info(meta, tablet_id));
+    if (meta != nullptr) {
+        static_cast<void>(
+                RowsetMetaManager::remove_tablet_related_partial_update_info(meta, tablet_id));
+    }
```
yiguolei, line 753: why cast the result as static void?
If you do not need check the result, I think you should modify the method to return void.

bobhan1:
run buildall

zhannngchen (approved):

bobhan1:
run p1

bobhan1:
run cloud_p1

bobhan1:
run performance

dataroaring (approved):
LGTM

bobhan1:
run performance

PR #38995, opened 2024-08-07
[fix](regression) fix test_group_commit_timeout failed due to different error message

mymeiyi:
the case is failed:
```
(test_group_commit_timeout.groovy:48) - failed errCode = 2, detailMessage = query is timeout, killed by timeout checker
```

PR #39068, opened 2024-08-08
[fix](regression) fix test_group_commit_timeout failed due to different error message (#38995)

mymeiyi:
pick https://github.com/apache/doris/pull/38995


mymeiyi:
run buildall

PR #39003, opened 2024-08-07
[fix](group commit) Modify group commit commit/abort txn timeout as stream load

mymeiyi:

PR #39069, opened 2024-08-08
[fix](group commit) Modify group commit commit/abort txn timeout as stream load (#39003)

mymeiyi:
pick https://github.com/apache/doris/pull/39003


mymeiyi:
run buildall

PR #36392, opened 2024-06-17
[fix](be) return correct canceled status from scanner

mrhhsg:
## Proposed changes

Issue Number: close #xxx

PR #39111, opened 2024-08-08
[fix](be) return correct canceled status from scanner (#36392)

mrhhsg:
## Proposed changes

pick #36392


mrhhsg:
run buildall

mrhhsg:
run buildall

PR #38790, opened 2024-08-02
[Fix](load) The value of the index id printed in the log is incorrect

liaoxin01:

PR #39131, opened 2024-08-09
[Fix](load) The value of the index id printed in the log is incorrect #38790

liaoxin01:
cherry pick from #38790


liaoxin01:
run buildall

PR #39164, opened 2024-08-09
[fix](schema-change) Fix potential data race when a schema change jobs is set to cancelled but the table state is still SCHEMA_CHANGE

Hastyshell:
## Proposed changes

Set job cancel state after table state changed to normal.


TangSiyang2001:
run buildall

dataroaring (approved):
LGTM

bobhan1 (approved):

PR #39167, opened 2024-08-09
[fix](mtmv) second level MTMV always refresh all partition by mistake…

zddr:
pick: 
https://github.com/apache/doris/pull/38698


zddr:
run buildall

PR #39168, opened 2024-08-09
[fix](s3_tvf)Fixed s3 tvf missing http header when parsing endpoint from uri.

hubgeter:
## Proposed changes

Issue Number: close #xxx


hubgeter:
run buildall

PR #39172, opened 2024-08-09
[Chore](runtime-filter) avoid dcheck fail when rf merge failed

BiteTheDDDDt:
## Proposed changes
 avoid dcheck fail when rf merge failed
there is a difference in the logic of calculating bf size between 2.1 and 3.0, so the merge will fail
this pr is to prevent core dump caused by dcheck failure during rolling upgrade.


BiteTheDDDDt:
run buildall

dataroaring (approved):
LGTM

zhiqiang-hhhh (approved):

PR #39121, opened 2024-08-08
[improvement](balance) don't balance tablet which has unfinish alter job

yujun777:
Improvement:   don't balance tablets that have unfished alter job.

Also fix partition rebalancer may balance colocate tablets.

PR #39202, opened 2024-08-12
[improvement](balance) don't balance tablet which has unfinish alter job #39121

yujun777:
cherry pick from #39121


yujun777:
run buildall

PR #39206, opened 2024-08-12
[Fix](JsonReader) Return correct status when parse failed

eldenmoon:
When using `JsonFunctions::extract_from_object`, but input obj is not object type but other types like null, then Status should be `simdjson::INCORRECT_TYPE` to fill the default value in the later process. Example, json path is `$.city.name`, but input json is `{"city" : null}` then `Status::NotFound` should be returned, in the following, column will be filled with default values.

## Proposed changes

Issue Number: close #xxx


eldenmoon:
run buildall

eldenmoon:
run buildall

xiaokang (dismissed):
LGTM

eldenmoon:
run buildall

qidaye (approved):
LGTM

amorynan (approved):
LGTM

PR #39077, opened 2024-08-08
[opt](test)Increase the timeout of http request

qidaye:
## Proposed changes

`show_nested_index_file` http api's response may be large with variant table, so we increase the timeout.

PR #39024, opened 2024-08-07
[opt](test)Add expected indices count check

qidaye:
## Proposed changes

Only when the expected_indices_count is 0, the tablet may not have the index file.

PR #38443, opened 2024-07-29
[improvement](test)Use show_nested_index_file to check inverted index

qidaye:
## Proposed changes

Use show_nested_index_file to check inverted index instead of calc_file_crc

PR #39232, opened 2024-08-12
[improvement](test)Use show_nested_index_file to check inverted index(#38443 #39024  #39077 )

qidaye:
## Proposed changes

bp:#38443 
#39024  
#39077


qidaye:
run buildall

PR #39233, opened 2024-08-12
[fix](nereids) fix partitionTopN choosing under multi winexprs

xzj7019:
intro by #38393

Fix the cases whose window function both contains row_number and other types but only the other types contains pushing down filter.


xzj7019:
run buildall

morrySnow (approved):

englefly (approved):

PR #39248, opened 2024-08-12
[fix](inverted index)Add exception check when write bkd index

qidaye:
## Proposed changes

We are not catching the exception when add values in `bkd_writer`, if error throws, BE will run into segment fault.
So we add the exception check here to avoid coredump.

### DIFF
--- a/be/src/olap/rowset/segment_v2/inverted_index_writer.cpp
+++ b/be/src/olap/rowset/segment_v2/inverted_index_writer.cpp
@@ -362,7 +362,7 @@ public:
                 _rid++;
             }
         } else if constexpr (field_is_numeric_type(field_type)) {
-            add_numeric_values(values, count);
+            RETURN_IF_ERROR(add_numeric_values(values, count));
         }
         return Status::OK();
     }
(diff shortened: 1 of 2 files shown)


qidaye:
run buildall

airborne12 (approved):
LGTM

be/src/olap/rowset/segment_v2/inverted_index_writer.cpp
```diff
@@ -520,23 +512,33 @@ class InvertedIndexColumnWriterImpl : public InvertedIndexColumnWriter {
...
+            std::string new_value;
+            size_t value_length = sizeof(CppType);
 
-        _value_key_coder->full_encode_ascending(&value, &new_value);
-        _bkd_writer->add((const uint8_t*)new_value.c_str(), value_length, _rid);
+            DBUG_EXECUTE_IF("InvertedIndexColumnWriterImpl::add_value_bkd_writer_add_throw_error", {
+                _CLTHROWA(CL_ERR_IllegalArgument, ("packedValue should be length=xxx"));
+            });
 
-        _rid++;
```
xiaokang, line 539: why delete `_rid++` ?

qidaye: In order to reuse `add_value` function in `add_array_values`, I moved the `_rid++` to the caller `add_numeric_values` in line 519.

xiaokang (approved):
LGTM

PR #39251, opened 2024-08-12
[fix](function) fix error return type in corr(float32,float32) 

Mryange:
## Proposed changes

```
mysql [test11]>select corr(cast(x as float),cast(y as float)) from test_corr;
ERROR 1105 (HY000): errCode = 2, detailMessage = (127.0.0.1)[INTERNAL_ERROR]column_type not match data_types in agg node, column_type=Nullable(Float64), data_types=Nullable(Float32),column name=

```


zclllyybb (approved):
LGTM

HappenLee (approved):
LGTM

PR #39104, opened 2024-08-08
[fix](ES Catalog)Check isArray before parse json to array

qidaye:
## Proposed changes

Elasticsearch does not have an explicit array type, but one of its fields can contain [0 or more values](https://www.elastic.co/guide/en/elasticsearch/reference/current/array.html).
When the field has one value and we map it as array type in Doris, it will run into segment fault while parsing it.
So we add a check before we parse json to array.

Issue Number: close #39102

PR #39273, opened 2024-08-13
[fix](ES Catalog)Check isArray before parse json to array (#39104)

qidaye:
## Proposed changes

bp #39104


qidaye:
run buildall

PR #39284, opened 2024-08-13
[fix](sync mv) fix rewrite wrongly with sync mv

justfortaste:
## Proposed changes

Issue Number: close #xxx

Aggregate table rewrite wrongly with sync mv:

drop table if  exists t1;
CREATE TABLE IF NOT EXISTS t1 (
                k int,
                a int,
                int_value int sum,
                char_value char(10) max,
                date_value date max
            )
            ENGINE=OLAP
            aggregate KEY(k,a)
            DISTRIBUTED BY HASH(k) BUCKETS 2 properties("replication_num" = "1");
insert into t1 values
    (1,1,1,'a', '2020-12-01'),
    (2,2,2,'b', '2021-12-01'),
    (3,2,2,'c', '2022-12-01'),
    (4,2,4,'c', '2023-12-01');

mysql> select a from t1;
+------+
| a    |
+------+
|    2 |
|    1 |
|    2 |
|    2 |
+------+
4 rows in set (0.01 sec)

CREATE MATERIALIZED VIEW v_t1
        AS 
     select k%2 as kk,a,max(date_value) from t1 group by kk, a;
select JobName from mv_infos('database'='test') where Name = 'v_t1';
select sleep(2);

// after rewire with v_t1, the result only have 3 rows
mysql> select a from t1;
+------+
| a    |
+------+
|    2 |
|    2 |
|    1 |
+------+
3 rows in set (0.01 sec)

## solution

When rewriting query with aggregate table's sync mv, it is not only necessary to check the key size, but also to check all key columns;


liutang123:
run buildall

regression-test/suites/nereids_rules_p0/mv/join/inner/inner_join_x.groovy
```diff
@@ -0,0 +1,93 @@
...
+    """
+
+    sql """
+    insert into t2 values(1,2),(2,2);
+    """
+
+    sleep(2000)
+
+    def query =  """
+    select min(t1.date_value) from t1 inner join t2 on t1.a=t2.a group by t2.k;
```
cambyzju, line 76: 1、add explain check to make sure mv is used or not;
2、then check the result is correct;

regression-test/suites/nereids_rules_p0/mv/join/inner/inner_join_x.groovy
```diff
@@ -0,0 +1,93 @@
...
+                date_value date max
+            )
+            ENGINE=OLAP
+            aggregate KEY(k,a)
+            DISTRIBUTED BY HASH(k) BUCKETS 2 properties("replication_num" = "1")
+
+    """
+
+    sql """
+     create materialized view v_t1 as select k%2 as kk,a, sum(int_value), max(date_value) from t1 group by kk, a;
```
cambyzju, line 44: add another case, replace `k%2` to k, then the mv should used?

liutang123:
Please fix this bug in old planner, its logic is in `MaterializedViewSelector.compensateCandidateIndex`.

regression-test/suites/nereids_rules_p0/mv/join/inner/inner_join_x.groovy
starocean999, line 76: t1.date_value is a max agg column. so what's the expected behavior using min(t1.date_value)?

justfortaste: min or max hava no effect. I think the key problem is "k/2"  not same as "k"

justfortaste:
run buildall

justfortaste:
run buildall

lide-reed (dismissed):
LGTM

starocean999:
run buildall

regression-test/suites/nereids_rules_p0/mv/join/inner/inner_join_x.groovy
```diff
@@ -0,0 +1,97 @@
...
+    sql "set runtime_filter_mode=OFF"
+    sql "set enable_sync_mv_cost_based_rewrite=false"
+
+
+    sql """
+    drop table if  exists t1;
+    """
+
+    sql """
+    CREATE TABLE IF NOT EXISTS t1 (
```
starocean999, line 30: Add a case for duplicate key table too

fe/fe-core/src/main/java/org/apache/doris/nereids/rules/rewrite/mv/SelectMaterializedIndexWithoutAggregate.java
```diff
@@ -234,6 +234,7 @@ public static LogicalOlapScan select(
             // So only base index and indexes that have all the keys could be used.
             List<MaterializedIndex> candidates = table.getVisibleIndex().stream()
                     .filter(index -> table.getKeyColumnsByIndexId(index.getId()).size() == baseIndexKeySize)
+                    .filter(index -> containAllKeyColumns(table, index))
```
starocean999, line 237: Does duplicate key table need this check?

starocean999:
run buildall

liutang123 (approved):
LGTM

starocean999 (approved):

PR #35515, opened 2024-05-28
[Fix](tvf) Pass through user-defined properties

BePPPower:
## Proposed changes

Issue Number: close #xxx

Previously, irrelevant properties passed by users when using TVF were directly ignored by Doris.
Now, we retain and pass these additional user-defined properties to the S3 SDK.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #39285, opened 2024-08-13
[Fix](tvf) Pass through user-defined properties 

BePPPower:
bp: #35515

Previously, irrelevant properties passed by users when using TVF were directly ignored by Doris.
Now, we retain and pass these additional user-defined properties to the S3 SDK.


BePPPower:
run buildall

BePPPower:
run p0

PR #39298, opened 2024-08-13
[Fix](nereids) fix condition function partition prune

feiniaofeiafei:
Current partition prune logic cannot correctly deal with condition function in predicates. For example, where if(substring('abc',1,2) <> "case", a > 0, b > 0), if the first parameter is not fold by FoldConstantRule, then our current logic will also intersect the partition range of b with b>0. If one of the partitions is empty, the partition will be trimmed. However, substring('abc',1,2) <> "case" is true, the predicate b>0 is not useful and should not be involved in partition trimming.
This pr remove the logic that make the result to "Boolean.False" when considering the child ranges and finding an empty range.


feiniaofeiafei:
run buildall

morrySnow (dismissed):

feiniaofeiafei:
run cloud_p0

feiniaofeiafei:
run cloud_p1

feiniaofeiafei:
run buildall

924060929 (approved):

morrySnow (approved):

PR #39302, opened 2024-08-13
[Fix](cloud) CloudStorageEngine should wait compaction task done when stoping it

hust-hhb:
When stoping doris, CloudStorageEngine should wait compaction task done, like StorageEngine does, if we don't wait, the compaction task run in background may core beacause these task will access CloudStorageEngine to get _txn_delete_bitmap_cache


hust-hhb:
run buildall

gavinchou (approved):

dataroaring (approved):
LGTM

PR #39308, opened 2024-08-13
[fix](protocol) CLIENT_MULTI_STATEMENTS not used actually

cambyzju:
in pr https://github.com/apache/doris/pull/36759, we try to output result depend CLIENT_MULTI_STATEMENTS,

but in `MysqlProto::negotiate` function, we call `context.setCapability(context.getServerCapability())` function set capability in context use server default flags instead of client flags.

So here, we save  CLIENT_MULTI_STATEMENTS flag in MysqlChannel like flag CLIENT_DEPRECATE_EOF


cambyzju:
run buildall

lide-reed (approved):
LGTM

morrySnow (approved):

PR #39311, opened 2024-08-13
[fix](routine load) should update progress before handle transaction state transform

sollhui:
Update progress maybe throw exception, causing offset has been persisted on edit log or meta service, but the memory data has not been updated. It will cause repeated consumption.


sollhui:
run buildall

liaoxin01 (approved):
LGTM

dataroaring (approved):
LGTM

dataroaring (approved):
LGTM

PR #39313, opened 2024-08-13
[fix](cloud) fix routine load job progress cache incorrect in cloud mode

sollhui:
Routine load job progress cache incorrect in cloud mode in the following scenario:
1. schedule thread update cloud progress, get old transaction value in fdb.
2. routine load task commit transaction and update progress cache.
3. update cloud progress RPC return, and change progress value to old value which is incorrect.

This PR solves the problem that may occur in the storage computation separation mode by not allowing small values to overwrite large values.


sollhui:
run buildall

liaoxin01 (approved):
LGTM

dataroaring (approved):
LGTM

PR #39322, opened 2024-08-13
[fix](function)timediff with now function  causes a error signature

Mryange:
## Proposed changes

```
mysql [(none)]>select round(timediff(now(),'2024-08-15')/60/60,2);
ERROR 1105 (HY000): errCode = 2, detailMessage = argument 1 requires datetimev2 type, however 'now()' is of datetime type
```
The reason is that the function parameter types were modified in expectedInputTypes, which led to no match being found. The code here is from a long time ago. Because the precision of datetimev2 could not be deduced in the past, a separate implementation was made here. This code can be safely deleted.


Mryange:
run buildall

HappenLee (approved):
LGTM

zclllyybb (approved):
LGTM

PR #39341, opened 2024-08-14
[improvement](jdbc catalog) Compatible with ojdbc6 by adding version check

zy-kkk:
In previous versions, we used a method based on JDBC 4.2 to read data, so it was equivalent to abandoning support for ojdbc6. However, we recently found that a large number of users still use Oracle version 11g, which will have some unexpected compatibility issues when using ojdbc8 to connect. Therefore, I use version verification to make it compatible with both ojdbc6 and ojdbc8, so that good compatibility can be obtained through ojdbc6, and better reading efficiency can be obtained through ojdbc8.


zy-kkk:
run buildall

morningman (approved):
LGTM

CalvinKirs (approved):

PR #39346, opened 2024-08-14
[fix](catalog) set timeout for split fetch

morningman:
When fetch splits in batch, BE will send rpc to FE to fetch batch of splits.
The FE may be blocked when listing file from hdfs, causing BE block too.
This PR add timeout on FE side to avoid BE block.


morningman:
run buildall

kaka11chen (approved):
LGTM

suxiaogang223 (approved):

PR #39347, opened 2024-08-14
[fix](function)timediff with now function  causes a error signature (…

Mryange:
…#39322)
https://github.com/apache/doris/pull/39322
The derivation of precision for the datetime constant in version 2.0 is incorrect, it tends to be derived as the maximum precision.

```
mysql [(none)]>select round(timediff(now(),'2024-08-15')/60/60,2);
ERROR 1105 (HY000): errCode = 2, detailMessage = argument 1 requires datetimev2 type, however 'now()' is of datetime type
```
The reason is that the function parameter types were modified in expectedInputTypes, which led to no match being found. The code here is from a long time ago. Because the precision of datetimev2 could not be deduced in the past, a separate implementation was made here. This code can be safely deleted.

## Proposed changes

Issue Number: close #xxx


Mryange:
run buildall

Mryange:
run buildall

PR #39360, opened 2024-08-14
[fix](stream load) do not throw exception but skip record when can not find database

sollhui:
When fetch stream load record from BE node, if can not find database, StreamLoadRecordMgr will throw exception and the remaining records will not be recorded in memory. 

For example: Ten stream load records were pulled, and the database associated with the stream load of the first record was deleted by the user. Therefore, the pull will end, resulting in the remaining nine records not being consumed recorded in memory.

This pr do not throw exception but skip record when can not find database to solve this problem.


sollhui:
run buildall

sollhui:
run buildall

liaoxin01 (approved):
LGTM

dataroaring (approved):
LGTM

Jibing-Li:
run buildall

PR #39361, opened 2024-08-14
[fix](file cache) Fix data race of rowset meta when download segment data

gavinchou:
The original impl. _rs_metas is not protected correctly when doing traversal. We have to access the rowset meta via tablet level interface to ensure integrity.


platoneko (approved):

gavinchou:
run buildall

bobhan1 (approved):

Yukang-Lian (approved):

PR #39378, opened 2024-08-14
revert [improvement](mv) Support to use cast when create sync materialized view #38008

seawinde:
## Proposed changes

this is brought by https://github.com/apache/doris/pull/38008
if use cast(FLOOR(MINUTE(time) / 15) as decimal(9, 0)) in group by clause when sync materialized view. if downgrade from 2.1.6 to 2.1.5 or upgrade 2.1.6 to 3.0.0
this may cause fe can not run. So revert the function.


seawinde:
run buildall

PR #37330, opened 2024-07-05
[fix](function)adjust aggregate function's nullable property

zhangstar333:
## Proposed changes

in this PR have change some age function nullable property.
aggregate functions can be divided into 3 types by their nullable properties:
- AlwaysNullable
- AlwaysNotNullable
- NullableAggregateFunction

this PR code of FE part from https://github.com/apache/doris/pull/37215, the author is @starocean999

PR #39380, opened 2024-08-14
[Fix-2.1](function) fix wrong nullable signature of function corr

zclllyybb:
## Proposed changes

Issue Number: close #xxx

before `corr(nullable_x, nullable_y)` will core dump. not fixed.
no need to patch in master because the refactor https://github.com/apache/doris/pull/37330 already changed the implementation context


zclllyybb:
run buildall

HappenLee (approved):
LGTM

PR #37794, opened 2024-07-15
[Fix](Variant) handle scalar variant with none string root

eldenmoon:
1. If columns is ColumnObject(Int32), root cast to ColumnString will result crash
2. variant with invalid json will be stored as raw string instead of throwing exception
## Proposed changes

Issue Number: close #xxx

PR #39394, opened 2024-08-15
[Fix](Row store) fix row store with invalid json string in variant type

eldenmoon:
Previous we allow invalid text as variant in PR #37794 and store as string type.But in encoding rowstore we CHECK the json is valid and store as jsonb binary field.In this PR we support the invalid json encoding as row store

## Proposed changes

Issue Number: close #xxx


eldenmoon:
run buildall

amorynan (approved):
LGTM

xiaokang (dismissed):
LGTM

be/src/vec/data_types/serde/data_type_object_serde.cpp
```diff
@@ -98,8 +98,13 @@ void DataTypeObjectSerDe::write_one_cell_to_jsonb(const IColumn& column, JsonbWr
     JsonbParser json_parser;
     // encode as jsonb
     bool succ = json_parser.parse(value_str.data(), value_str.size());
-    // maybe more graceful, it is ok to check here since data could be parsed
-    CHECK(succ);
+    if (!succ) {
```
xiaokang, line 101: if (succ) {
  // write binary
} else {
  // write string
}

eldenmoon: done

eldenmoon:
run buildall

xiaokang (approved):
LGTM

PR #39339, opened 2024-08-14
[fix](group commit) fix group commit core if be inject FragmentMgr.exec_plan_fragment.failed

mymeiyi:
```
*** SIGSEGV address not mapped to object (@0x0) received by PID 1898955 (TID 1900522 OR 0x7f4f94abc640) from PID 0; stack trace: ***
 0# doris::signal::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*) at /home/zcp/repo_center/doris_branch-3.0.2-tmp/doris/be/src/common/signal_handler.h:421
 1# PosixSignals::chained_handler(int, siginfo*, void*) [clone .part.0] in /usr/lib/jvm/java-17-openjdk-amd64/lib/server/libjvm.so
 2# JVM_handle_linux_signal in /usr/lib/jvm/java-17-openjdk-amd64/lib/server/libjvm.so
 3# 0x00007F5335001520 in /lib/x86_64-linux-gnu/libc.so.6
 4# brpc::Socket::Write(brpc::SocketMessagePtr<void>&, brpc::Socket::WriteOptions const*) in /mnt/disk1/STRESS_ENV/be/lib/doris_be
 5# brpc::policy::HttpResponseSender::~HttpResponseSender() in /mnt/disk1/STRESS_ENV/be/lib/doris_be
 6# brpc::policy::HttpResponseSenderAsDone::~HttpResponseSenderAsDone() in /mnt/disk1/STRESS_ENV/be/lib/doris_be
 7# std::_Function_handler<void (), doris::PInternalService::group_commit_insert(google::protobuf::RpcController*, doris::PGroupCommitInsertRequest const*, doris::PGroupCommitInsertResponse*, google::protobuf::Closure*)::$_0>::_M_invoke(std::_Any_data const&) at /var/local/ldb-toolchain/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/std_function.h:291
 8# doris::WorkThreadPool<false>::work_thread(int) at /home/zcp/repo_center/doris_branch-3.0.2-tmp/doris/be/src/util/work_thread_pool.hpp:159
```

PR #39396, opened 2024-08-15
[fix](group commit) fix group commit core if be inject FragmentMgr.exec_plan_fragment.failed (#39339)

mymeiyi:
pick https://github.com/apache/doris/pull/39339


mymeiyi:
run buildall

mymeiyi:
run buildall

PR #39180, opened 2024-08-09
[fix](jdbc scan) Remove the `conjuncts.remove` call in JdbcScan

zy-kkk:
In #37565, due to the change in the calling order of finalize, the final generated Plan will be missing the PREDICATES that have been pushed down in Jdbc. Although this behavior is correct, before perfectly handling the push down of various PREDICATES, we need to keep all conjuncts to ensure that we can still filter data normally when the data returned by Jdbc is a superset.

PR #37565, opened 2024-07-09
[Fix](ScanNode) Move the finalize phase of ScanNode to after the end of the Physical Translate phase.

BePPPower:
## Proposed changes

Currently, Doris first obtains splits and then performs projection. After column pruning, it calls `updateRequiredSlots` to update the scanRange information. However, the Trino connector's column pruning pushdown needs to be completed before obtaining splits.

Therefore, we move the finalize phase of `ScanNode` to after the end of the `Physical Translate` phase, so that `createScanRangeLocations` can use the final columns which have been pruning.

PR #39407, opened 2024-08-15
[branch-2.1][fix](jdbc scan) Remove the `conjuncts.remove` call in JdbcScan 

zy-kkk:
pick (#39180)

In #37565, due to the change in the calling order of finalize, the final generated Plan will be missing the PREDICATES that have been pushed down in Jdbc. Although this behavior is correct, before perfectly handling the push down of various PREDICATES, we need to keep all conjuncts to ensure that we can still filter data normally when the data returned by Jdbc is a superset.


zy-kkk:
run buildall

zy-kkk:
run  feut

PR #37793, opened 2024-07-15
[enhancement](log)Updated the default setting of sys_log_mode in fe.conf for better performance

Baymine:
## Proposed changes

Updated the default setting of `sys_log_mode` in `fe.conf` from `NORMAL` to `ASYNC` for better performance.

PR #39419, opened 2024-08-15
[fix](fe-log) add position info in async mode

morningman:
followup #37793
NORMAL:  sync flush + with location
ASYNC(default): async flush + with location
BRIEF: async flush + without location


gavinchou (dismissed):

caiconghui:
this config is used to improve log4j performance, so I suggest that to make best performance for brief mode without location  and async flush, and the default config is async which is with location but async flush

caiconghui:
so we need to modify illustration for this config

zclllyybb (approved):

caiconghui (approved):
LGTM

gavinchou (approved):

gavinchou:
run buildall

morningman:
run performance

PR #39447, opened 2024-08-15
[Fix](load) Fix the incorrect src value printed in the error log when strict mode is true

liaoxin01:
When there is a column order mapping, an incorrect src value will be printed because the wrong column was used.


liaoxin01:
run buildall

dataroaring (approved):
LGTM

gavinchou (approved):

PR #39448, opened 2024-08-15
[fix](profile) Make auto_profile_threshold_ms work on FE

zhiqiang-hhhh:
The `auto_profile_threshold_ms` parameter indicates that a profile will only be generated if the query time exceeds this threshold. For load jobs, since each load task has a profile, this parameter applies to each load task, indicating that a profile will be generated only if the load task's execution time exceeds this threshold.

The default value of `auto_profile_threshold_ms` is -1, which means that as long as the user has enabled profiling, the profile will be collected. This behavior is consistent with the current older version.

When the user sets `auto_profile_threshold_ms` to 4 seconds:

1. On the BE (Backend) side, while reporting the profile, it will check the execution time, and if it is less than 4 seconds, it will not report it(#38548 ).
2. . On the FE (Frontend) side, it will calculate based on the execution time observed by the coordinator.
   1. If the query completion time is less than 4 seconds, it will directly delete the profile without saving it or waiting for the report to finish. The associated execution profile will also be deleted.
   3. If the query time is more than 4 seconds, the profile will not be deleted and needs to be saved. It will also be persisted when memory is insufficient and needs to be evicted.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/common/profile/Profile.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/profile/Profile.java
@@ -103,7 +103,7 @@ public class Profile {
     private long queryFinishTimestamp = Long.MAX_VALUE;
     private Map<Integer, String> planNodeMap = Maps.newHashMap();
     private int profileLevel = MergedProfileLevel;
-    private long autoProfileDurationMs = 500;
+    private long autoProfileDurationMs = -1;
     // Profile size is the size of profile file
     private long profileSize = 0;
 
@@ -629,5 +629,21 @@ public class Profile {
 
     public long getProfileSize() {
         return this.profileSize;
+    }
+
+    public boolean shouldBeRemoveFromMemory() {
+        if (!this.isQueryFinished) {
+            return false;
+        }
+
+        if (this.profileHasBeenStored()) {
+            return false;
+        }
+
+        if (this.queryFinishTimestamp - this.summaryProfile.getQueryBeginTime() >= autoProfileDurationMs) {
+            return false;
+        }
+
+        return true;
     }
 }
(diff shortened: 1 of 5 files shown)


zhiqiang-hhhh:
run buildall

zhiqiang-hhhh:
run buildall

fe/fe-core/src/main/java/org/apache/doris/common/profile/Profile.java
```diff
@@ -625,4 +630,20 @@ public void deleteFromStorage() {
...
+    public boolean shouldBeRemoveFromMemory() {
+        if (!this.isQueryFinished) {
+            return false;
+        }
+
+        if (this.profileHasBeenStored()) {
+            return false;
+        }
+
+        if (this.queryFinishTimestamp - this.summaryProfile.getQueryBeginTime() >= autoProfileDurationMs) {
```
yiguolei, line 643: 为什么有这个条件？

fe/fe-core/src/main/java/org/apache/doris/common/util/ProfileManager.java
```diff
@@ -913,8 +905,98 @@ public void markExecutionProfileFinished(TUniqueId queryId) {
...
+    private void preventExecutionProfileLeakage() {
+        StringBuilder stringBuilder = new StringBuilder();
+        int executionProfileNum = 0;
+        writeLock.lock();
+        try {
+            // This branch has two purposes:
+            // 1. discard profile collecting if its collection not finished in 5 seconds after query finished.
+            // 2. prevent execution profile from leakage. If we have too many execution profiles in memory,
+            // we will remove execution profiles of query that has finished in 5 seconds ago.
+            if (queryIdToExecutionProfiles.size() > 2 * Config.max_query_profile_num) {
```
yiguolei, line 922: 这里不用加这个if 检查了，因为未来这个值可能很大

yiguolei (approved):

morningman (approved):

PR #39391, opened 2024-08-15
[fix](group commit) fix group commit insert rpc may stuck

mymeiyi:
1. when exec_plan_fragment failed and PipelineFragmentContext is not constructed, the group_commit_insert rpc will stuck
2. the `LoadBlockQueue.add_block.back_pressure_time_out` debug point is not work

PR #39458, opened 2024-08-16
[fix](group commit) fix group commit insert rpc may stuck (#39391)

mymeiyi:
pick https://github.com/apache/doris/pull/39391


mymeiyi:
run buildall

PR #39470, opened 2024-08-16
[fix](util) Fix DorisOnce not accurate once when exception happend

Hastyshell:
## Proposed changes

In cpp, exception catch relies on stack unwind and will destruct all stack variables before entering catch block.

In the original DorisOnce, if exception happended when `fn` called, the `_flag_lock` will be released before `_has_called` flag in set. The race condition may make it actually called twice.

This PR use std::once_falg and std::call_once to refactor the DorisOnce.

1. Support call once semantic and return result when `call()` is called. If an exception thrown, the exception will be rethrown whenever called the `call()`.
2. Supports to get is called status through function `is_called()`.
3. Support to get produced value through `stored_result()`. If exception throwed, will rethrow this exception. If it is not called, will throw an exception to indicate it.

### DIFF
--- a/be/src/util/once.h
+++ b/be/src/util/once.h
@@ -51,7 +51,7 @@ namespace doris {
 template <typename ReturnType>
 class DorisCallOnce {
 public:
-    DorisCallOnce() : _has_called(false) {}
+    DorisCallOnce() : _once_flag(false) {}
 
     // this method is not exception safe, it will core when exception occurs in
     // callback method. I have tested the code https://en.cppreference.com/w/cpp/thread/call_once.
@@ -89,16 +89,15 @@ public:
             return _status;
         }
         try {
+            // This memory order make sure both status and eptr is set
+            // and will be seen in another thread.
+            _once_flag.store(true, std::memory_order_release);
             _status = fn();
         } catch (...) {
             // Save the exception for next call.
             _eptr = std::current_exception();
-            _has_called.store(true, std::memory_order_release);
             std::rethrow_exception(_eptr);
         }
-        // This memory order make sure both status and eptr is set
-        // and will be seen in another thread.
-        _has_called.store(true, std::memory_order_release);
         return _status;
     }
 
@@ -109,7 +108,7 @@ public:
         // std::memory_order_acquire here and std::memory_order_release in
         // init(), taken together, mean that threads can safely synchronize on
         // _has_called.
-        return _has_called.load(std::memory_order_acquire);
+        return _once_flag.load(std::memory_order_acquire);
     }
 
     // Return the stored result. The result is only meaningful when `has_called() == true`.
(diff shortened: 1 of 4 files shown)


TangSiyang2001:
run buildall

TangSiyang2001:
run buildall

TangSiyang2001:
run buildall

TangSiyang2001:
run buildall

be/src/util/once.h
```diff
@@ -89,16 +89,15 @@ class DorisCallOnce {
             return _status;
         }
         try {
+            // This memory order make sure both status and eptr is set
+            // and will be seen in another thread.
+            _once_flag.store(true, std::memory_order_release);
             _status = fn();
         } catch (...) {
             // Save the exception for next call.
             _eptr = std::current_exception();
```
TangSiyang2001, line 98: how shall we handle exception here to make it visible intime for other threads?

TangSiyang2001: I got a good idea, will update this PR later.

TangSiyang2001:
run buildall

TangSiyang2001:
run buildall

TangSiyang2001:
run buildall

TangSiyang2001:
run buildall

TangSiyang2001:
run buildall

TangSiyang2001:
run p0

be/src/util/once.h
```diff
@@ -42,95 +48,67 @@ namespace doris {
...
-    // occurred.
     template <typename Fn>
-    ReturnType call(Fn fn) {
-        // Avoid lock to improve performance
-        if (has_called()) {
-            if (_eptr) {
-                std::rethrow_exception(_eptr);
+        requires std::is_invocable_r_v<R, Fn>
+    R call(Fn fn) {
+        std::call_once(_once_flag, [&] {
```
yiguolei, line 66: In the past, we use std call once, but there are 2 main reasons that we implement our own call once:
1. std call once could not deal with exception logic correctly.
2. when it is core during std call once, it's callstack will include another thread, it is not very clear.

yiguolei (changes requested):
https://learn.microsoft.com/en-us/cpp/cpp/exceptions-and-stack-unwinding-in-cpp?view=msvc-170

Are you sure the lock will be destroyed?

TangSiyang2001:
The call once `called twice` problem is in the formal version DorisOnce, the current version has no problem, this fix is not in demand, so close this PR.

TangSiyang2001:
> https://learn.microsoft.com/en-us/cpp/cpp/exceptions-and-stack-unwinding-in-cpp?view=msvc-170
> 
> Are you sure the lock will be destroyed?

Misdirect by a similar case, the lock is out of try{}, and will not be destroyed.

PR #39479, opened 2024-08-16
[fix](json function) json function parse failed should not cancel load and need record url and rows statistic

sollhui:
when using routine load  as follows to consume Kafka stream, there is no exception but data can not consume. 
```
CREATE ROUTINE LOAD jobName ON tableName
                COLUMNS TERMINATED BY ",",
                COLUMNS(k1,k2,v1,v2,v3,tmp,v4=json_unquote(tmp))
                PROPERTIES
                (
                    "max_batch_interval" = "5",
                    "max_batch_rows" = "300000",
                    "max_batch_size" = "209715200"
                )
                FROM KAFKA
                (
                    "kafka_broker_list" = "${externalEnvIp}:${kafka_port}",
                    "kafka_topic" = "${kafkaCsvTpoics[0]}",
                    "property.kafka_default_offsets" = "OFFSET_BEGINNING"
                );
```

The root cause is the json function ` json unquote` will cancel load directly and do not record url and rows statistic if parses data error, such as data `"aaaaaaaabbbbbbbbb"""`,making routine load data quality checker invalid and always retry to consume Kafka data.


sollhui:
run buildall

sollhui:
run buildall

PR #39357, opened 2024-08-14
[fix](delete) Only apply regex check on delete condition str for non-lsc tables

Hastyshell:
## Proposed changes

Light schema change capable tables will work on delete sub predicate v2 and doesn't need this check.

PR #39500, opened 2024-08-16
[fix](delete) Only apply regex check on delete condition str for non-lsc tables (#39357)

Hastyshell:
## Proposed changes

Light schema change capable tables will work on delete sub predicate v2 and doesn't need this check.


TangSiyang2001:
run buildall

PR #39351, opened 2024-08-14
[enhancement](schema-change) Record detailed fail reason for schema change tasks

Hastyshell:
## Proposed changes

Expose the error msg from BE as the real fail reason recorded for schema change tasks. To avoid too much memory usage, we just pick one among all to record.

PR #39501, opened 2024-08-16
[enhancement](schema-change) Record detailed fail reason for schema change tasks (#39351)

Hastyshell:
## Proposed changes

pick #39351

Expose the error msg from BE as the real fail reason recorded for schema change tasks. To avoid too much memory usage, we just pick one among all to record.


TangSiyang2001:
run buildall

PR #39525, opened 2024-08-18
[fix](routine load) add read lock to fix some concurrent bugs (#39242)

sollhui:
pick #39242


sollhui:
run buildall

yiguolei:
run beut

PR #39526, opened 2024-08-18
[fix](routine load) should update progress before handle transaction state transform (#39311)

sollhui:
pick (#39311)

Update progress maybe throw exception, causing offset has been persisted on edit log or meta service, but the memory data has not been updated. It will cause repeated consumption.


sollhui:
run buildall

PR #38350, opened 2024-07-25
[enhance](Backup) Do connectivity check when creating repository

ByteYue:
## Proposed changes

Previously when creating repository, FE would not do connectivity check. It might result in confusing error when using backup restore.

PR #39538, opened 2024-08-18
[enhance](Backup) Do connectivity check when creating repository (#38350)

gavinchou:
Previously when creating repository, FE would not do connectivity check. It might result in confusing error when using backup restore.

pick #38350


gavinchou:
run buildall

PR #39581, opened 2024-08-19
[Bug](column) fix append_data_by_selector_impl reserve too mush useless memory

zhangstar333:
## Proposed changes

```
for (auto* place : local_state._value_places) {
                            SCOPED_TIMER(local_state._selector_block_timer);
                            RETURN_IF_ERROR(place->append_block_by_selector(input_block, eos));
}
    for (int i = 0; i < mutable_columns.size(); ++i) {
        columns[i]->append_data_by_selector(mutable_columns[i], _selector);
    }
```
the columns[I] size is 4096, and _selector size is 1;
so the (4096 - 1) rows is useless;
in a usercase, the block have 4096 rows and 34 columns.
so need bytes is:
1 block * 4096 rows * 4096 reserve * 16 bytes * 34 column * 48 instance = 438086664192 bytes;

it's will cause the query canceled as not have enough memory.

### DIFF
--- a/be/src/vec/columns/column_impl.h
+++ b/be/src/vec/columns/column_impl.h
@@ -43,6 +43,9 @@ void IColumn::append_data_by_selector_impl(MutablePtr& res, const Selector& sele
                                selector.size(), num_rows);
     }
     DCHECK_GE(end, begin);
+    // here wants insert some value from this column, and the nums is (end - begin)
+    // and many be this column num_rows is 4096, but only need insert num is (1 - 0) = 1
+    // so can't call res->reserve(num_rows), it's will be too mush waste memory
     res->reserve(res->size() + (end - begin));
 
     for (size_t i = begin; i < end; ++i) {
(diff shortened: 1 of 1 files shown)


zhangstar333:
run buildall

be/src/vec/columns/column_impl.h
```diff
@@ -42,8 +42,8 @@ void IColumn::append_data_by_selector_impl(MutablePtr& res, const Selector& sele
                                "Size of selector: {} is larger than size of column: {}",
                                selector.size(), num_rows);
     }
-
-    res->reserve(num_rows);
+    DCHECK_GE(end, begin);
+    res->reserve(end - begin);
```
yiguolei, line 46: add comment here to explain why need using this variable to reserve memory not using num rows

yiguolei (approved):
LGTM

zhangstar333:
run buildall

zclllyybb (approved):
LGTM

PR #39587, opened 2024-08-20
[Fix](load) Fix the incorrect src value printed in the error log when strict mode is true #39447

liaoxin01:
cherry pick from #39447


liaoxin01:
run buildall

PR #39611, opened 2024-08-20
[opt](memory) Modify memory gc conf and add `crash_in_alloc_large_memory_bytes`

xinyiZzz:
1. faster frequency and small batch memory GC, which helps query stability.
2. when alloc memory larger than crash_in_alloc_large_memory_bytes will crash, default -1 means disabled. if you need a core dump to analyze large memory allocation, modify this parameter to crash when large memory allocation occur will help


xinyiZzz:
run buildall

yiguolei (approved):

dataroaring (approved):
LGTM

PR #39613, opened 2024-08-20
[Fix] issues where certain syntax does not work correctly in some bash …

cfbber:
…environments.

## Proposed changes

Issue Number: close #xxx


cfbber:
bin/run-tpcds-queries.sh: line 138: syntax error near unexpected token `<'

PR #39620, opened 2024-08-20
[fix](regression) wrong bthread context when executing fault injection

freemandealer:
This injection is meant to test gc during data loading. The gc process cannot executing in bthread context by design, because of the `DCHECK(bthread_sefl()==0)` in LocalFileReader in Segment Cache free procedure. So I start a new thread to do the job.

## Proposed changes

Issue Number: close #xxx


kaijchen (approved):
LGTM

dataroaring (approved):
LGTM

hello-stephen (approved):

PR #39638, opened 2024-08-20
[fix](cast) Fix cast as boolean

zhiqiang-hhhh:
* Target 

Fix the behavior of cast numeric as boolean.

Unlike mysql, doris has real boolean type( https://doris.apache.org/docs/2.0/table-design/data-type/). The only possible values of a boolean type in doris should be 0 or 1. 

Currently, the computation of type bool is replaced with UInt8, which means we could get values like 2,3,4 for boolean column.

* Fix

This pr fix the computation of cast numeric as boolean, and add some check when we do result serialization.
We need more work in scanner

* Future work

Do value convert in scanner. To make sure we will only get 0/1 for bool column even though value in storage is others.


zhiqiang-hhhh:
run buildall

PR #39642, opened 2024-08-20
[chore](script message) More convenient error msg for nofile open.

zhiqiang-hhhh:
`ulimit -n 655350` will only take effect on current session, ask user to modify `/etc/security/limits.conf` to make the config work for every time we login.

New message like;
```
Set max number of open file descriptors to a value greater than 60000.
Ask your system manager to modify /etc/security/limits.conf and append content like
  * soft nofile 655350
  * hard nofile 655350
and then run 'ulimit -n 655350' to take effect on current session.
```


zhiqiang-hhhh:
run buildall

yiguolei (approved):
LGTM

jacktengg (approved):

PR #39343, opened 2024-08-14
[bugfix](external)add check of engine and catalog types

wuwenchi:
## Proposed changes

Currently, we can only create hive table under hms type catalog and iceberg table under iceberg catalog.

PR #39643, opened 2024-08-20
[bugfix](external)add check of engine and catalog types for 2.1 #39343

wuwenchi:
## Proposed changes

bp #39343

Currently, we can only create hive table under hms type catalog and iceberg table under iceberg catalog.


morningman:
run buildall

PR #39647, opened 2024-08-20
[chore](log) Log written rows when result sink close

zhiqiang-hhhh:
In some cases, audit log in fe says a query did not return any rows. So we log the information when result sink close to help us do check.


Jibing-Li (approved):

yiguolei (approved):

Jibing-Li:
run buildall

zhiqiang-hhhh:
run p1

yiguolei (approved):

PR #39675, opened 2024-08-21
[fix](nereids)prevent null pointer exception if datetime value overflows

starocean999:
## Proposed changes
pick from master https://github.com/apache/doris/pull/39482


starocean999:
run buildall

PR #39679, opened 2024-08-21
[fix](cluster key) forbid cluster key and remove case

mymeiyi:
branch-2.1 does not support mow cluster key


mymeiyi:
run buildall

dataroaring (approved):
LGTM

PR #39568, opened 2024-08-19
[fix](local shuffle) Fix correctness for bucket hash shuffle exchanger

Gabriel39:
## Proposed changes

For query plan

image

we will plan local exchangers  and get a new plan

image

and the hash join operator will get probe and build data which are different distributed (one is HASH shuffle and another is Bucket hash shuffle). This PR fix it.

PR #39691, opened 2024-08-21
[fix](local shuffle) Fix correctness for bucket hash shuffle exchange…

Mryange:
…r (#39568)

For query plan

image

we will plan local exchangers  and get a new plan

image

and the hash join operator will get probe and build data which are different distributed (one is HASH shuffle and another is Bucket hash shuffle). This PR fix it.

## Proposed changes

Issue Number: close #xxx


Mryange:
run buildall

PR #39695, opened 2024-08-21
[Enhancement](FE) Make scheduler aware of interval changes immediately

zclllyybb:
## Proposed changes

Issue Number: close #xxx

Before, if `dynamic_partition_check_interval_seconds` is 86400 now and we want to set it to 1, the set operation will not go into effect in 86400 seconds in worst situation.
Now all set operation will norify scheduler thread and set the interval immediately.

Now only applied for `DynamicPartitionScheduler`. should check other daemons in future.

### DIFF
--- a/fe/fe-common/src/main/java/org/apache/doris/common/ConfigBase.java
+++ b/fe/fe-common/src/main/java/org/apache/doris/common/ConfigBase.java
@@ -123,6 +123,10 @@ public class ConfigBase {
         }
     }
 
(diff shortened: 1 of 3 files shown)


zclllyybb:
run buildall

zclllyybb:
run buildall

zhangstar333 (approved):

CalvinKirs (approved):

PR #39699, opened 2024-08-21
[fix](funcs) map struct construct funcs

amorynan:
## Proposed changes
we should not change the column which is passed from a block when we execute a function , because this block may be used in other expression or other exec node
Issue Number: close #xxx


amorynan:
run buildall

amorynan:
run buildall

HappenLee (approved):
LGTM

eldenmoon (approved):
LGTM

PR #39700, opened 2024-08-21
[fix](hive) report error with escape char and null format

suxiaogang223:
## Proposed changes

Because be did not process escape char and null format when reading the hive text table, an error was reported when fe found that this value was not the default value.


suxiaogang223:
run buildall

suxiaogang223:
run buildalll

suxiaogang223:
run buildalll

suxiaogang223:
run buildall

morningman (approved):
LGTM

BePPPower (approved):

PR #39616, opened 2024-08-20
[fix](auditlog)Record return row count in audit log for internal query.

Jibing-Li:
Record return row count in audit log for internal query. Before, the returned row count for internal select in audit log is always 0.

PR #39701, opened 2024-08-21
[fix](auditlog)Record return row count in audit log for internal query. (#39616)

Jibing-Li:
backport: https://github.com/apache/doris/pull/39616


Jibing-Li:
run buildall

PR #39714, opened 2024-08-21
[fix](mtmv) Fix regression test unsteadiness

seawinde:
## Proposed changes

1. Add explain memo plan for problem location easier
2.  Fix regression test unsteadiness because exception msg difference between master and cloud


seawinde:
run buildall

morrySnow (approved):

PR #39721, opened 2024-08-21
[opt](file-cache) add evict file number per round

morningman:
Previously, when getting block from file cache, it may try to evict
lots of blocks to reserve capacity for lru cache. This operation may take long time
while hold the lock, causing other operation blocked.

This PR add a new BE config `file_cache_max_evict_num_per_round`,
default is 1000, so that it will not hold lock for a long time.


morningman:
run buildall

PR #39730, opened 2024-08-21
[Pick][Improment]publish workload to BE by tag (#38486)

wangbo:
A workload group's tag property may be three cases as below: 1 empty string, null or '', it could be published to all BE. 2 a value match some BE' location, then the workload group could only be published to the BE with same tag.
3 not an empty string, but some invalid string which can not math any BE's location, then it could not be published any BE.

## Proposed changes

Issue Number: close #xxx


wangbo:
run buildall

PR #39736, opened 2024-08-21
[fix](Nereids) forbid sql cache when use nondeterministic function

924060929:
## Proposed changes

The sql cache maybe return outdated cache because we cannot detect the the result of non-deterministric function changed, so I forbid sql cache when use nondeterministic function.

This problem not exists in the 2.1.3 because I refactor it


924060929:
run buildall

924060929:
run buildall

924060929:
run buildall

PR #39743, opened 2024-08-21
[fix](agg function) incorrect result of map agg

mrhhsg:
## Proposed changes

In the function `void add(const Field& key_, const Field& value)`, it should not return while the key exists in the set.


mrhhsg:
run buildall

mrhhsg:
run buildall

yiguolei (approved):

jacktengg (approved):

PR #39744, opened 2024-08-21
[fix](ctas) fix NPE when ctas with old planner and varchar issue

morningman:
Fix 2 bugs:
1. 
introduced from #35489
When the target table is unpartitioned with text type column from external table,
NPE will be thrown.
Only happen when using old planner.
The new planner works well.

2.
If first column is varchar type, the new planner will change it to string type
and then failed because first column can not be string type.


morningman:
run buildall

BePPPower (approved):
LGTM

BePPPower:
run buildall

morrySnow (dismissed):

morrySnow (approved):

BePPPower (approved):

morningman:
run buildall

PR #38168, opened 2024-07-20
[fix](memory) Refactor refresh workload groups weighted memory ratio and record refresh interval memory growth

xinyiZzz:
1. Modify `refresh_wg_memory_info` to `refresh_wg_weighted_memory_ratio`,
- Previously, `total memory used` and `weighted memory` of wordload groups were refreshed periodically, and they no longer change during the refresh period.
- Now only `weighted memory ratio` is refreshed periodically, and `weighted memory` will use (`total memory used` + `refresh interval memory growth`) * `weighted memory ratio` in real time. Wordload groups and queries only save `weighted memory ratio`, which is more accurate.
- wordload groups `total memory used` is refreshed when `make_memory_tracker_snapshots` is executed, which is called by `workload gorups memory GC` and `refresh_wg_weighted_memory_ratio`.

2. `Thread Context` add workload group weak ptr, and record workload group memory ratio refresh interval memory growth.
- if thread in query or load, initialize with memory tracker, query id and workload group wptr.
- else not query or load, initialize with memory tracker, empty query id and default normal workload group.

PR #39760, opened 2024-08-22
[branch-2.1](memory) Refactor refresh workload groups weighted memory ratio and record refresh interval memory growth

xinyiZzz:
pick #38168
overwrites changes in #37221 on workload_group_manager.cpp. If need to pick 37221, ignore it.


xinyiZzz:
run buildall

PR #39775, opened 2024-08-22
[fix](cloud) should do before commit check in cloud mode

sollhui:
In cloud mode, there is no before commit check when commit transaction. For routine load, job may be paused during sub task execution. If these transactions are committed, the problem is:
1. The progress seen by the user may not be accurate. It will consume data repeatedly if they create a new job consumption using this progress.
2. The user's modification offset operation may be overwritten by the offset of the transaction callback.


sollhui:
run buildall

sollhui:
run buildall

sollhui:
run buildall

liaoxin01 (approved):
LGTM

XuJianxu (approved):
LGTM

PR #39786, opened 2024-08-22
[fix](commit) Fix does not skip commit if txn state is committed or visible

mymeiyi:
introduced in #32980
if a txn is committed twice, the check txn state is committed or visible and skip commit logic does not work

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/transaction/DatabaseTransactionMgr.java
@@ -680,7 +680,10 @@ public class DatabaseTransactionMgr {
         return writeDetail;
     }
 
-    private void checkTransactionStateBeforeCommit(Database db, List<Table> tableList, long transactionId,
+    /**
+     * @return true if the transaction need to commit, otherwise false
+     */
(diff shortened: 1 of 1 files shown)


mymeiyi:
run buildall

yujun777 (changes requested):
```
void   XXXTransaction() {
        readlock()
         txn = getTransaction()
         checkTransactionStateBeforeXXX() 
        readunlock()

         do something no need lock()

         writeLock()
         checkTransactionStateBeforeXXX() 
          ...
         writeUnlock()
}
```


maybe in write lock scope,  check transaction status again is better ?

here XXXTransaction  XXX should include PreCommit/Commit/Abort

mymeiyi:
run fe_ut

mymeiyi:
run feut

dataroaring (approved):
LGTM

liaoxin01:
Please add a description of what scenario triggers the issue.

dataroaring (changes requested):
There should be a lock hold when check and commit. Otherwise, two parallel thread may check ok and commit together.  It seems that lock on trasaction state is enough.

dataroaring (approved):
LGTM

liaoxin01 (approved):
LGTM

PR #39787, opened 2024-08-22
[chore](log) Optimization for log on FE and BE

zhiqiang-hhhh:
At this stage, when an exception occurs, we are logging excessive and redundant information on both the frontend (FE) and backend (BE). This PR attempts to streamline and reduce these logs.


zhiqiang-hhhh:
run buildall

PR #39788, opened 2024-08-22
[improve](statistics)Record index row count update time.

Jibing-Li:
Record index row count update time. So the planner could use it to decide which row count to use, reported row count or analyzed row count.


Jibing-Li:
run buildall

englefly (approved):

morrySnow (approved):

PR #37670, opened 2024-07-11
[Fix](merge-on-write) cloud mow table should sync rowsets in publish phase if compaction on other BE finished during this load

bobhan1:
## Proposed changes
Due to https://github.com/apache/doris/pull/35838, when executing load job, BE will not `sync_rowsets()` in publish phase if a compaction job is finished on another BE on the same tablet between the commit phase and the publish phase of the current load job. This PR let the meta service return the tablet compaction stats along with the getDeleteBitmapUpdateLockResponse to FE and FE will send them to BE to let the BE know whether it should `sync_rowsets()` due to compaction on other BEs.

PR #39791, opened 2024-08-22
[fix](cloud-mow) FE should not request partition stats by snapshot read in calcDeleteBitmapForMow

zhannngchen:
## Proposed changes

Issue Number: close #xxx

#37670 Let FE call get_delete_bitmap_update_lock in calcDeleteBitmapForMow to get the latest compaction stats of the corresponding partition at the same time, so that in the downstream calcDelete bitmap task, it can let BE determine if there is a concurrent compaction conflict.

However, this PR uses snapshot read when fetching the partition stats, which makes it possible for us to fetch outdated stats, thus allowing the BE to miss the conflict processing of concurrent compaction, and generating duplicate keys.


zhannngchen:
run buildall

dataroaring (approved):
LGTM

bobhan1 (approved):

PR #39801, opened 2024-08-22
[fix](nereids) Remove Nondeterministic to avoid ambiguity when decidea expression is nondeterministic or not

seawinde:
## Proposed changes

In https://github.com/apache/doris/pull/36111, we add `isDeterministic` method in class `ExpressionTrait` to identify the expression is deterministic or not.

But `unix_timestamp` doesn't extend Nondeterministic, but it is not deterministic when it's children is empty. and is not deterministic when children is not empty. If we use` instanceOf Nondeterministic `to indentify if expression is  is not deterministic, that is confused.

So we do something as fllowing:
1. Remove Nondeterministic class, and use `isDeterministic` to indentify it's deterministic.
2. Add `containsNondeterministic` method in `ExpressionTrait` to identify it contains nondeterministic expression or not.
3. `isDeterministic` only identify current expression is deterministic or not. would identify if contains nondeterministic or not


fe/fe-core/src/main/java/org/apache/doris/common/NereidsSqlCacheManager.java
```diff
@@ -396,7 +396,8 @@ private boolean usedVariablesChanged(List<Variable> currentVariables, SqlCacheCo
             Variable currentVariable = currentVariables.get(i);
             Variable cachedVariable = cachedUsedVariables.get(i);
             if (!Objects.equals(currentVariable, cachedVariable)
-                    || cachedVariable.getRealExpression().anyMatch(Nondeterministic.class::isInstance)) {
+                    || cachedVariable.getRealExpression().anyMatch(expr -> expr instanceof ExpressionTrait
+                    && !((ExpressionTrait) expr).isDeterministic())) {
```
924060929, line 400: You can remove `expr instanceof ExpressionTrait` because Expression is always ExpressionTrait

fe/fe-core/src/main/java/org/apache/doris/nereids/trees/expressions/functions/ExpressionTrait.java
```diff
@@ -79,23 +78,16 @@ default boolean foldable() {
...
-        }
-        return isDeterministic;
+        return true;
+    }
+
+    /**
+     * Identify the expression is containing deterministic expr or not
+     */
+    default boolean containsNondeterministic() {
+        return anyMatch(expr -> expr instanceof ExpressionTrait && !((ExpressionTrait) expr).isDeterministic());
```
924060929, line 91: remove `expr instanceof ExpressionTrait`

seawinde:
run buildall

924060929 (approved):

starocean999 (approved):

PR #39806, opened 2024-08-22
[fix](paimon) fix not able to read paimon data from hdfs with HA

morningman:
When reading Paimon data via JNI, the hadoop properties should be passed to JNI part.
Otherwise, it may miss some properties and return error like:
`CAUSED BY: IllegalArgumentException: java.net.UnknownHostException: hdfs-cluster`


morningman:
run buildall

wuwenchi (approved):

kaka11chen (approved):
LGTM

PR #35457, opened 2024-05-27
[opt](fe) Optimize fe show table statistics

swjtu-zhanglei:
* Remove `olapTable.readLock` when show table statistics

## Proposed changes

Issue Number: close #xxx

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #39807, opened 2024-08-22
[optimize](table stat) Reduce lock when get table statistics

xy720:
## Proposed changes

Issue Number: this pr will pick part of #35457 to reduce the potential of dead lock with following example:

时间点 | create table线程1 | create table线程2 | table stat线程 | truncate线程
-- | -- | -- | -- | --
t1 | add table1 read lock |   |   |  
t2 |   | add table2 read lock |   |  
t3 |   |   | try add table1 write lock，blocked |  
t4 |   |   |   | try add table2 write lock，blocked
t5 | try add table2 read lock，Due to a thread attempting to add the table2 write lock earlier, it was forced to queue up |   |   |  
t6 |   | try add table1 read lock，Due to a thread attempting to add the table1 write lock earlier, it was forced to queue up |   |  

This deadlock situation usually occurs when it is a fair lock. So we should optimize the read lock.

This PR places the updates of table statistics(such as data size/replica count) in the Table Stats Thread to be done on a scheduled basis, so that there is no need to add a read lock when obtaining table statistics.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/OlapTable.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/OlapTable.java
@@ -86,6 +86,8 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import lombok.Getter;
+
 /**
  * Internal representation of tableFamilyGroup-related metadata. A OlaptableFamilyGroup contains several tableFamily.
  * Note: when you add a new olap table property, you should modify TableProperty class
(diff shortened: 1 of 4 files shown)


fe/fe-core/src/main/java/org/apache/doris/catalog/OlapTable.java
```diff
@@ -148,6 +150,8 @@ public enum OlapTableState {
 
     private TableProperty tableProperty;
 
+    private volatile Statistics statistics = new Statistics();
```
liutang123, line 153: When fe restarted, the `statistics` has not inited and the dataSize will be 0. At this point, `createTable` behavior will not be restricted.
Should we deal with this situation?

xy720: As we discuss, if statistics is not inited, we should try init it first.
But I think we may first implement this logic in master branch.

lide-reed (approved):
LGTM

PR #39821, opened 2024-08-23
[branch-2.0](thrift) fix TLoadTxnBeginRequest backend_id's field no

yujun777:
Make backend_id's field no the same with master branch.

For upgrading, change backend_id's field no is safe,   because old fe can torrent with TLoadTxnBeginRequest not setting backend id.

backend_id was introduce by #39317


yujun777:
run buildall

PR #39825, opened 2024-08-23
[Fix](Nerieds) insert into table with default not null but not included in insert schema

LiBinfeng-01:
example:
create table t(k1 int not null, k2 int not null);
insert into t (k1) values (101);
it should be failed because we don't know what value should k2 be


LiBinfeng-01:
run buildall

morrySnow (dismissed):

LiBinfeng-01:
run buildall

morrySnow (approved):

starocean999 (approved):

PR #39802, opened 2024-08-22
[fix](arrow-flight-sql) Fix regression-test arrow flight port

xinyiZzz:
Fix regression-test arrow flight port

PR #39127, opened 2024-08-08
[fix](arrow-flight-sql) Fix exceed user property max connection cause `Reach limit of connections`

xinyiZzz:
Limit the number of arrow flight connections for a single user to less than the user property max_user_connections / 2, default 50.

PR #39836, opened 2024-08-23
[branch-2.1](arrow-flight-sql) Fix exceed user property max connection cause Reach limit of connections

xinyiZzz:
pick #39127
pick #39802


xinyiZzz:
run buildall

PR #39841, opened 2024-08-23
[Fix](Variant) use uinque id to access column reader

eldenmoon:
Currently, the variant type is not supported rename column because its column reader accesses columns by path rather than by unique ID. If the name is modified, the column reader may not locate the column correctly.So we should access by unique id 

## Proposed changes

Issue Number: close #xxx


eldenmoon:
run buildall

be/src/olap/rowset/segment_v2/hierarchical_data_reader.cpp
```diff
@@ -34,14 +36,15 @@ namespace segment_v2 {
 Status HierarchicalDataReader::create(std::unique_ptr<ColumnIterator>* reader,
                                       vectorized::PathInData path,
                                       const SubcolumnColumnReaders::Node* node,
-                                      const SubcolumnColumnReaders::Node* root) {
+                                      const SubcolumnColumnReaders::Node* root,
+                                      const vectorized::DataTypePtr& root_type) {
```
xiaokang, line 40: root_type is not used

be/src/olap/field.h
```diff
@@ -28,6 +28,7 @@
 #include "olap/types.h"
 #include "olap/utils.h"
 #include "runtime/collection_value.h"
+#include "runtime/descriptors.h"
```
xiaokang, line 31: unused

xiaokang (dismissed):
LGTM

eldenmoon:
run buildall

be/src/olap/rowset/segment_v2/hierarchical_data_reader.cpp
eldenmoon: done

be/src/olap/field.h
eldenmoon: done

xiaokang (dismissed):
LGTM

eldenmoon:
run buildall

qidaye (approved):

amorynan (approved):
LGTM

PR #39865, opened 2024-08-23
[fix](hive)Modify the Hive notification event processing method when using meta cache and add parameters to the Hive catalog. (#39239)

morningman:
bp #39239


morningman:
run buildall

PR #39877, opened 2024-08-24
[fix](paimon) fix crash when enable cache with paimon deletion vector

morningman:
missing IOContext param, cause NPE.
When init row filter for paimon reader, it will read deletion file, which need io context


morningman:
run buildall

wuwenchi (approved):

kaka11chen (approved):
LGTM

PR #39883, opened 2024-08-24
[feat]reserve and spill

mrhhsg:
## Proposed changes

Issue Number: close #xxx

### DIFF
--- a/be/src/pipeline/exec/aggregation_sink_operator.cpp
+++ b/be/src/pipeline/exec/aggregation_sink_operator.cpp
@@ -23,6 +23,7 @@
 #include "common/status.h"
 #include "pipeline/exec/operator.h"
 #include "runtime/primitive_type.h"
+#include "runtime/thread_context.h"
 #include "vec/common/hash_table/hash.h"
 #include "vec/exprs/vectorized_agg_fn.h"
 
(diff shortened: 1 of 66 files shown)


be/src/pipeline/exec/aggregation_source_operator.cpp
```diff
@@ -22,17 +22,13 @@
...
+#include "runtime/thread_context.h"
 #include "vec/exprs/vectorized_agg_fn.h"
+#include "vec/exprs/vexpr_fwd.h"
 
 namespace doris::pipeline {
 
-AggLocalState::AggLocalState(RuntimeState* state, OperatorXBase* parent)
-        : Base(state, parent),
-          _get_results_timer(nullptr),
-          _serialize_result_timer(nullptr),
```
yiguolei, line 32: why delete these code

be/src/pipeline/exec/operator.h
```diff
@@ -221,6 +234,24 @@ class PipelineXLocalStateBase {
     vectorized::Block _origin_block;
 };
 
+class ScopedMemTracker {
+public:
+    ScopedMemTracker(size_t& counter) : _counter(counter), _mem_tracker("ScopedMemTracker") {
```
yiguolei, line 239: the output value should use pointer, not use reference

be/src/pipeline/exec/aggregation_sink_operator.cpp
```diff
@@ -394,6 +399,9 @@ Status AggSinkLocalState::_merge_with_serialized_key_helper(vectorized::Block* b
 Status AggSinkLocalState::_merge_without_key(vectorized::Block* block) {
     SCOPED_TIMER(_merge_timer);
     DCHECK(_agg_data->without_key != nullptr);
+
+    _memory_usage_last_executing = 0;
+    ScopedMemTracker mem_tracker(_memory_usage_last_executing);
```
yiguolei, line 404: 不要这么定义，参考SCOPED_TIMER。
定义一个宏，然后传入一个counter的指针。

PR #39621, opened 2024-08-20
[fix](stat) handle overflow of memory stat if load failed

dataroaring:
## Proposed changes

Issue Number: close #xxx

PR #39887, opened 2024-08-24
[fix](stat) handle overflow of memory stat if load failed (#39621)

dataroaring:
## Proposed changes

pick #39621 

Issue Number: close #xxx


dataroaring:
run buildall

PR #39892, opened 2024-08-24
[Fix-2.1](testcase) Fix some testcases unstable results due to lack of orderby

zclllyybb:
## Proposed changes

Issue Number: close #xxx

some sql's order-by column has same value, add more order-by column to fix it.


zclllyybb:
run buildall

PR #39893, opened 2024-08-24
[Opt](parquet/orc-reader) Opt get dict ids in _rewrite_dict_predicates().

kaka11chen:
## Proposed changes

[Opt] (parquet/orc-reader) Opt get dict ids in _rewrite_dict_predicates().

The original implementation needed to obtain the dict code through dict value when generating dictionary filter conditions. In order to obtain the dict code, a `dict_value_to_code` hash map needs to be constructed. We found that when the amount of data is large, it is time-consuming to construct the `dict_value_to_code` hash map, and the current implementation will also construct it when there is no filter condition. Because the dict code is the position of the dictionary element, this PR records the position when filtering the dict value, so as to directly obtain the dict code.


kaka11chen:
run buildall

kaka11chen:
run buildall

kaka11chen:
run buildall

morningman (dismissed):
LGTM

wuwenchi (approved):

kaka11chen:
run buildall

kaka11chen:
run beut

morningman (approved):

CalvinKirs (approved):

PR #39899, opened 2024-08-26
[fix](group commit) Fix group commit error log when decommission

mymeiyi:
Decommission backend will check the group commit wal num is 0.
But there is a small bug of log:
```
boolean hasWal = checkWal(backend);  // the hasWal should be empty wal

hasWal ? "; and has unfinished WALs" : "" // so if log print 'and has unfinished WALs', the wal is 0 actually
```


mymeiyi:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #37196, opened 2024-07-03
[chore][fe] Get meta returns dropped partitions in binlog manager

w41ter:
The CCR syncer needs to know the distribution of tables, partitions, indexes, and replicas when synchronizing binlogs. If a partition is deleted before the binlog synchronization is complete, the CCR syncer cannot continue synchronizing. This PR will record the deleted partitions and include them in the get meta response, allowing the CCR syncer to filter out the binlogs that belong to these partitions.

The CCR syncer part PR is selectdb/ccr-syncer#117.

PR #39909, opened 2024-08-26
[fix](binlog) Fix NPE when recover binlogs

w41ter:
The field partition ID of DropPartitionInfo was added in PR: apache/doris#37196, the old version doesn't contain this field so it will be null.


w41ter:
run buildall

Yukang-Lian (approved):

dataroaring (approved):
LGTM

PR #39910, opened 2024-08-26
[fix](binlog) Fix NPE when recover binlogs (#39909)

w41ter:
Cherry-pick #39909 

The field partition ID of DropPartitionInfo was added in PR: apache/doris#37196, the old version doesn't contain this field so it will be null.


w41ter:
run buildall

PR #39935, opened 2024-08-26
[bug](regression-test)fix regression test fail

DarvenDuan:
## Proposed changes

After pr #37985, explain will not be rewritten for querying random distributed table, so remove explain action in regression test case.


DarvenDuan:
run buildall

yiguolei (approved):
LGTM

starocean999 (approved):

PR #37014, opened 2024-06-28
[bug](function)fix json_replace check return type error

zhangstar333:
## Proposed changes
1. fix the return type dcheck error:
```
mysql [test]>select (json_replace(a, '$.fparam.nested_2', "qwe")) from json_table_2 limit 1;
ERROR 1105 (HY000): errCode = 2, detailMessage = (10.16.10.8)[INTERNAL_ERROR]Function json_replace get failed, expr is VectorizedFnCall[json_replace](arguments=a, String, String, String,return=Nullable(String)) and return type is Nullable(String).
```

2.  improve the json_replace/json_insert/json_set function execute of not convert const column, test about could faster 1s on 1000w table rows

PR #39938, opened 2024-08-26
[bug](function)fix json_replace check return type error (#37014)

zhangstar333:
cherry-pick from master https://github.com/apache/doris/pull/37014
1. fix the return type dcheck error:
```
mysql [test]>select (json_replace(a, '$.fparam.nested_2', "qwe")) from json_table_2 limit 1;
ERROR 1105 (HY000): errCode = 2, detailMessage = (10.16.10.8)[INTERNAL_ERROR]Function json_replace get failed, expr is VectorizedFnCall[json_replace](arguments=a, String, String, String,return=Nullable(String)) and return type is Nullable(String).
```

2. improve the json_replace/json_insert/json_set function execute of not convert const column, test about could faster 1s on 1000w table rows

## Proposed changes

Issue Number: close #xxx


zhangstar333:
run buildall

PR #39946, opened 2024-08-26
[fix](catalog) hide some properties in show create catalog stmt

morningman:
the `use_meta_cache` and `create_time` property should not be visible in
show create catalog stmt.
These properties are not for user.
But you can still see them in `show catalog xxx` stmt


morningman:
run buildall

wuwenchi (approved):

kaka11chen (approved):
LGTM

PR #39358, opened 2024-08-14
[fix](function) fix error return type in mod(float32,BigInt)

Mryange:
## Proposed changes

```
CREATE TABLE testdb (
    K1 BIGINT,
    K2 FLOAT
) properties("replication_num" = "1");
insert into testdb values(1,1.1);
select mod(k1,k2) from testdb;
mysql [test10]>select mod(k1,k2) from testdb;
ERROR 1105 (HY000): errCode = 2, detailMessage = (127.0.0.1)[INTERNAL_ERROR]Function mod get failed, expr is VectorizedFnCall[mod](arguments=K1, K2,return=Nullable(Float32)) and return type is Nullable(Float32).
```

PR #39971, opened 2024-08-27
[fix](function) fix error return type in mod(float32,BigInt) (#39358)

Mryange:
## Proposed changes
https://github.com/apache/doris/pull/39358
```
CREATE TABLE testdb (
    K1 BIGINT,
    K2 FLOAT
) properties("replication_num" = "1");
insert into testdb values(1,1.1);
select mod(k1,k2) from testdb;
mysql [test10]>select mod(k1,k2) from testdb;
ERROR 1105 (HY000): errCode = 2, detailMessage = (127.0.0.1)[INTERNAL_ERROR]Function mod get failed, expr is VectorizedFnCall[mod](arguments=K1, K2,return=Nullable(Float32)) and return type is Nullable(Float32).
```

## Proposed changes

Issue Number: close #xxx


Mryange:
run buildall

PR #39975, opened 2024-08-27
[improve](routine load) delay schedule EOF tasks to avoid too many small transactions

sollhui:
We encountered a scenario where a large number of small transactions were generated, resulting in an impact on query performance:
Kafka's data comes in batches of very small data every very short time, which leads to tasks being frequently scheduled and ending very quickly, resulting in a large number of small transactions.

To solve this problem, we delay the scheduling of tasks that perceive EOF, which would not delay data consumption, for perceiving EOF indicates that the consumption speed is greater than the production speed.


sollhui:
run buildall

sollhui:
run buildall

sollhui:
run buildall

dataroaring (approved):
LGTM

liaoxin01 (approved):
LGTM

PR #39981, opened 2024-08-27
[Fix-2.1](testcase) fix an unstable semi-structure case result

zclllyybb:
## Proposed changes

Issue Number: close #xxx


zclllyybb:
run buildall

PR #27800, opened 2023-11-30
[Improvement](schema-change) Reserve some memory for use by other parts except hold block of schem…

BiteTheDDDDt:
…a change job

## Proposed changes
Reserve some memory for use by other parts except hold block of schema change job

```cpp
W0922 15:57:10.657173 62550 beta_rowset_reader.cpp:289] failed to read next block: [MEM_LIMIT_EXCEEDED]PreCatch error code:11, [E11] Memory limit exceeded:[MEM_LIMIT_EXCEEDED]failed alloc size 1.00 MB, exceeded tracker:<EngineAlterTabletTask#baseTabletId=61523:newTabletId=61950>, limit 2.00 GB, peak used 2.00 GB, current used 2.00 GB, exec node:<>, execute msg:Allocator mem tracker check failed. backend 172.30.199.214 process memory used 12.95 GB, limit 402.55 GB. Can `set exec_mem_limit=8G` to change limit, details see be.INFO.
0. /root/src/doris-2.0/be/src/common/stack_trace.cpp:302: StackTrace::tryCapture() @ 0x000000000b8a51a7 in /opt/app/apache-doris-2.0.1/be/lib/doris_be
```

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #39995, opened 2024-08-27
[cherrypick](schema-change) cherrypick to Reserve some memory for use by other par…

GoGoWen:
cp #27800  to fix schema change failure


GoGoWen:
run buildall

PR #40007, opened 2024-08-27
[opt] change remote scan pool size

morningman:
1. Use Config validator to init the scan pool size
2. The origin remote pool size is `max(512, core * 10)`, which is too large, change it to `max(96, core * 2)`


morningman:
run buildall

PR #40011, opened 2024-08-28
[chore](log) Log written rows when result sink close #39647

zhiqiang-hhhh:
cherry pick from #39647


zhiqiang-hhhh:
run buildall

PR #40027, opened 2024-08-28
[fix](cloud) Fix cloud auto start and add a regression case

deardeng:
1. Fix the cluster being suspended and select not waking up the cluster. The reason is that all be nodes in the cluster are inactive, the cluster is skipped, and the cluster that needs to be woken up cannot be found, and the wake-up logic will not be reached. And delete the redundant function getAuthorizedCloudCluster
2. Add check after resume cluster, there must be at least one alive be in the cluster.
3. add auto start regression case

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/cloud/catalog/CloudReplica.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/cloud/catalog/CloudReplica.java
@@ -148,6 +148,13 @@ public class CloudReplica extends Replica {
     }
 
     private long getBackendIdImpl(String cluster) {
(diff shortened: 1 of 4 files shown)


deardeng:
run buildall

gavinchou (commented):
This PR, describe below, what is the bug and how was it fixed, and then the removed segment of code, as well as changing the return type from void to string, does it involve a behavior change?

gavinchou (dismissed):

dataroaring (dismissed):
LGTM

deardeng:
run buildall

gavinchou (approved):

dataroaring (approved):
LGTM

PR #40044, opened 2024-08-28
[fix](compile) compile fail on MacOS because 'bits/types/struct_timespec.h' not found (#39567)

cambyzju:
pick #39567 to branch-2.1


cambyzju:
run buildall

PR #40059, opened 2024-08-28
[improve](cloud-mow) splitting large rowset meta into multiple KVs

hust-hhb:
now large rowset meta save to fdb will fail, it will report error `failed to commit rowset: failed to save rowset meta, err=Value length exceeds limit`, rowset meta  need to split into multiple KVs when value is bigger than 90kb.


hust-hhb:
run buildall

dataroaring (changes requested):
Please add ut and regression test.

gavinchou (changes requested):
This implementation is dangerous and not well considered with recycling procedure.
Further discussion should be taken.

hust-hhb:
run buildall

hust-hhb:
run buildall

PR #40069, opened 2024-08-28
[Fix](Variant) variant fallthrough with inverted index

eldenmoon:
When reading from segment, the schema type is variant, if we check type valid in `get_inverted_index`, the result should always return nullptr(since variant type it self does not support inverted index), but the actual storage could be `string` or etc.So we should ignore the type check and return the correct inverted index iterators

introduced by #36163
## Proposed changes

Issue Number: close #xxx


eldenmoon:
run buildall

airborne12 (dismissed):
LGTM

csun5285 (approved):
LGTM

eldenmoon:
run buildall

eldenmoon:
run buildall

qidaye (approved):

PR #40058, opened 2024-08-28
[fix](compaction) fix heap-use-after-free caused by compaction sample info resize

luwei16:

PR #40077, opened 2024-08-28
[fix](compaction) fix heap-use-after-free caused by compaction sample info resize (#40058)

luwei16:


luwei16:
run buildall

dataroaring:
run buildall

PR #40071, opened 2024-08-28
[fix](ut)fix be enable_http_auth ut

hubgeter:
## Proposed changes
before pr #39577

PR #39577, opened 2024-08-19
[Enhancement](http)Add http authentication to all API interfaces under be 8040.

hubgeter:
## Proposed changes
Add http authentication to all API interfaces under be 8040
The `enable_all_http_auth parameter` in be.conf can control the switch.

Issue Number: close #xxx

PR #40088, opened 2024-08-29
[fix](ut)fix be enable_http_auth ut (#40071)

hubgeter:
bp #40071 

## Proposed changes
before pr #39577


hubgeter:
run buildall

PR #40093, opened 2024-08-29
[Feat](Nereids) add no use index hint

LiBinfeng-01:
support hint like:
/*+ no_use_index(tablename indexname) */ which forbid tablename(indexname) to be choose
or /*+ no_use_index */ which forbid all indexes to be choose


LiBinfeng-01:
run buildall

PR #40099, opened 2024-08-29
[fix](nereids) fix union all instance number

xzj7019:
pick from https://github.com/apache/doris/pull/39999


xzj7019:
run buildall

PR #40189, opened 2024-08-30
[Bug](expr) execute expr should use local states instead of operators

zhangstar333:
## Proposed changes
The expr of operator  cannot be executed concurrently, 
should use local state's expr.


zhangstar333:
run buildall

Gabriel39 (approved):

HappenLee (approved):
LGTM

PR #40201, opened 2024-08-30
[Bug](function) fix encryption_function return wrong result at const column

zhangstar333:
## Proposed changes

the string object have error of order, should be src, key,iv,mode

### DIFF
--- a/be/src/vec/functions/function_encryption.cpp
+++ b/be/src/vec/functions/function_encryption.cpp
@@ -335,7 +335,7 @@ struct EncryptionAndDecryptFourImpl {
         return Status::OK();
     }
 
-    static void vector_const(const ColumnString* column, StringRef iv_arg, StringRef key_arg,
+    static void vector_const(const ColumnString* column, StringRef key_arg, StringRef iv_arg,
                              StringRef mode_arg, size_t input_rows_count,
                              ColumnString::Chars& result_data, ColumnString::Offsets& result_offset,
                              NullMap& null_map) {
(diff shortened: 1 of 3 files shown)


zhangstar333:
run buildall

yiguolei (approved):

be/src/vec/functions/function_encryption.cpp
```diff
@@ -335,7 +335,7 @@ struct EncryptionAndDecryptFourImpl {
         return Status::OK();
     }
 
-    static void vector_const(const ColumnString* column, StringRef iv_arg, StringRef key_arg,
```
cambyzju, line 338: we should add: `use_default_implementation_for_constants to false`?

otherwise we still can not go to vector_const branch.

zclllyybb (approved):
LGTM

be/src/vec/functions/function_encryption.cpp
zclllyybb: we can. this is vector_const, not const_const

zhiqiang-hhhh (approved):
LGTM

be/src/vec/functions/function_encryption.cpp
cambyzju: please add some log or have a debug to make sure it.

the default const impl will convert const columns to normal columns at first, then call execute_impl function.

image

HappenLee (commented):
why performance opt pr Introduce bugs ？ Please add
 more test for all · encryption_function`

PR #40205, opened 2024-08-30
[fix](case) drop table if exists before create it

hello-stephen:
## Proposed changes

Issue Number: close #xxx


hello-stephen:
skip buildall

PR #40212, opened 2024-08-31
[opt](inverted index) cache result sign is initialized only once

zzzxl1993:
## Proposed changes

1. optimize the performance issue caused by slow result sign generation due to too many values in the in condition.

### DIFF
--- a/be/src/vec/exprs/vexpr.cpp
+++ b/be/src/vec/exprs/vexpr.cpp
@@ -617,7 +617,14 @@ bool VExpr::fast_execute(Block& block, const ColumnNumbers& arguments, size_t re
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


zzzxl1993:
run buildall

be/src/vec/exprs/vexpr.cpp
```diff
@@ -617,7 +617,14 @@ bool VExpr::fast_execute(Block& block, const ColumnNumbers& arguments, size_t re
         return false;
     }
 
-    std::string result_column_name = gen_predicate_result_sign(block, arguments, function_name);
+    // cache value is initialized only once.
```
xiaokang, line 620: It can be done in open without lock.

xiaokang (approved):

PR #40216, opened 2024-08-31
[fix](regression) fix regression failure of window_funnel caused by duplicated table name

jacktengg:
BP #40217


jacktengg:
run buildall

PR #40219, opened 2024-08-31
[Bug](expr) execute expr should use local states instead of operators (#40189)

zhangstar333:
## Proposed changes

cherry-pick from master #40189


zhangstar333:
run buildall

PR #37783, opened 2024-07-15
[test](inverted index)Add upgrade and downgrade test cases for inverted_index

ChenPeng2013:
## Proposed changes

Issue Number: close #xxx

add case

PR #40221, opened 2024-08-31
[test](inverted index)Add upgrade and downgrade test cases for inverted_index (#37783)

ChenPeng2013:
add case


airborne12:
run buildall

PR #40227, opened 2024-09-01
[fix](jdbc catalog) Fix incorrect format of FullDriverUrl under Windows

Chen768959:
When debugging in a Windows environment, the `jdbc_drivers_dir` format is `D:\xxx\xxx`. 
The `getFullDriverUrl()` will return `file://D:\xxx\xxx`, 
image
which does not comply with URI format specifications and will prevent the JDBC jar from being loaded in Windows. 
The correct FullDriverUrl format should be `file:///D:\xxx\xxx`.


Chen768959:
run buildall

Chen768959:
Could you please help review this? Thank you all. @morrySnow @xiaokang @dataroaring

morrySnow (approved):

PR #40228, opened 2024-09-01
[fix](mysql catalog) Fixed exception in dorisTypeToDoris when using MySQL catalog.

Chen768959:
When using specific MySQL JDBC packages (such as mysql-connector-java-8.3.0), their `CHAR` or `VARCHAR` metadata strings lack the (len) suffix. 
Consequently, related queries will result in errors due to attempts at converting `CHAR` to int.
clipboard

This pr ensures compatibility with the metadata formats of both `CHAR` and `VARCHAR` types by providing the maximum lengths.


Chen768959:
run buildall

PR #40002, opened 2024-08-27
[fix](arrow-flight-sql) Fix return result from FE Arrow Flight server error `0.0.0.0:xxx, connection refused`

xinyiZzz:
Fix: #36340
Introduced in: #34850

```
adbc_driver_manager.OperationalError: IO: [FlightSQL] connection error: desc = "transport: Error while dialing: dial tcp 0.0.0.0:8029: connect: connection refused" (Unavailable; DoGet: endpoint 0: [uri:"grpc+tcp://0.0.0.0:8029"])
```

When query results are returned from FE Arrow Flight server, and Doris FE and ADBC ​​Client are not on the same machine, such as `show tables;`, `show databases;`, the IP in FlightInfo returned by FE to the client is wrong, and `0.0.0.0` is returned instead of the real IP of FE. The client fails to pull the results from `0.0.0.0`.

No problem when query results are returned from BE Arrow Flight server, such as the normal query statement `select * from xxx`.

PR #40229, opened 2024-09-01
[branch-2.1](arrow-flight-sql) Fix return result from FE Arrow Flight server error 0.0.0.0:xxx, connection refused

xinyiZzz:
pick #40002


xinyiZzz:
run buildall

Issue #39798, opened 2024-08-22
[Bug] Dup key columns or unique key columns with varchar columns will fail to modify lenth


TangSiyang2001:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

### Version

all

### What's Wrong?

Due to historical reasons, there are currently bugs when increasing the length of varchar types for dup key and uniq key: the agg type is not unified.

```java
public void checkSchemaChangeAllowed(Column other) throws DdlException {
        if (Strings.isNullOrEmpty(other.name)) {
            throw new DdlException("Dest column name is empty");
        }

        if (!ColumnType.isSchemaChangeAllowed(type, other.type)) {
            throw new DdlException("Can not change " + getDataType() + " to " + other.getDataType());
        }

        if (type.isNumericType() && other.type.isStringType()) {
            try {
                Integer lSize = type.getColumnStringRepSize();
                Integer rSize = other.type.getColumnStringRepSize();
                if (rSize < lSize) {
                    throw new DdlException(
                            "Can not change from wider type " + type.toSql() + " to narrower type "
                                    + other.type.toSql());
                }
            } catch (TypeException e) {
                throw new DdlException(e.getMessage());
            }
        }

        if (!Objects.equals(this.aggregationType, other.aggregationType)) {
            throw new DdlException("Can not change aggregation type");
        }
        ...
```

```java
private boolean processModifyColumn(ModifyColumnClause alterClause, OlapTable olapTable,
                                        Map<Long, LinkedList<Column>> indexSchemaMap) throws DdlException {
        Column modColumn = alterClause.getColumn();
        boolean lightSchemaChange = false;

        if (KeysType.AGG_KEYS == olapTable.getKeysType()) {
            if (modColumn.isKey() && null != modColumn.getAggregationType()) {
                throw new DdlException("Can not assign aggregation method on key column: " + modColumn.getName());
            } else if (!modColumn.isKey() && null == modColumn.getAggregationType()) {
                throw new DdlException("Aggregate method must be specified for value column: " + modColumn.getName());
            }
        } else if (KeysType.UNIQUE_KEYS == olapTable.getKeysType()) {
            if (null != modColumn.getAggregationType()) {
                throw new DdlException("Can not assign aggregation method" + " on column in Unique data model table: "
                        + modColumn.getName());
            }
            if (!modColumn.isKey()) {
[...]

PR #39319, opened 2024-08-13
[enhancement](schema-change) Support light schema change on hash columns and agg key columns with varchar type to change length

Hastyshell:
## Proposed changes

1. Schema change should rebuild distribution info after modifying columns, especially distribution columns. Or it may cause dynamic partition failed when checking distribution columns' equality.
2. Support hash key columns to do light schema change. For unique key or dup key columns, could not be enabled temporarily due to some historical reasons. See #39798 .

PR #40236, opened 2024-09-02
[enhancement](schema-change) Support light schema change on hash columns and agg key columns with varchar type to change length (#39319)

Hastyshell:
## Proposed changes
pick https://github.com/apache/doris/pull/39319
1. Schema change should rebuild distribution info after modifying columns, especially distribution columns. Or it may cause dynamic partition failed when checking distribution columns' equality.
2. Support hash key columns to do light schema change. For unique key or dup key columns, could not be enabled temporarily due to some historical reasons. See #39798 .


TangSiyang2001:
run buildall

PR #40247, opened 2024-09-02
[cherry-pick](branch-21) fix encryption_function return wrong result at const column (#40201)

zhangstar333:
## Proposed changes
cherry-pick from master #40201
the string object have error of order, should be src, key,iv,mode


zhangstar333:
run buildall

PR #40260, opened 2024-09-02
[fix](cloud) Fix missing privilege to storage vault after restarting FE

gavinchou:
The previous implement forgets to build `storageVaultPrivTable` (in-memory) after loading auth information from image, which means the privileges are persisted but unable to use after restarting FE.

Note: a new image will be generated after FE restarts or a checkpoint is triggered,
the newly generated will be used when FE starts.

### DIFF
--- a/regression-test/suites/vaults/privilege_restart.groovy
+++ b/regression-test/suites/vaults/privilege_restart.groovy
@@ -1,0 +1,178 @@
+
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


regression-test/suites/vaults/privilege_restart.groovy
```diff
@@ -0,0 +1,178 @@
...
+                    )
+        """
+    }
+
+    //==========================================================================
+    // check the prepared users and tables
+    //==========================================================================
+    def allUsers = (sql " SHOW all grants ").stream().map(row -> row[0]).collect(Collectors.toSet())
+    logger.info("all users: ${allUsers}")
+    def userPresent = !(allUsers.stream().filter(i -> i.contains(user1)).collect(Collectors.toSet()).isEmpty())
```
deardeng, line 99: groovy，有个 find {} , 搜下 case，能找到例子，这个写的太难看了

deardeng: allUsers.find { user1 == it }

Yukang-Lian (approved):

liaoxin01 (approved):
LGTM

PR #40261, opened 2024-09-02
[improvement](jdbc catalog)Optimize JDBC Catalog refresh to reduce frequent client creation.

zy-kkk:
In the previous JDBC Catalog refresh behavior, each refresh would close and recreate the JdbcClient. During this process, we observed that some JDBC drivers create classloader-level shared threads when the JdbcClient is repeatedly instantiated. These threads cannot be garbage collected. Since we use the JdbcClient as the context class loader for the JDBC driver, frequent creation leads to a buildup of non-recyclable shared threads in the JVM. This PR changes the refresh behavior to update the cache only, rather than recreating the client. Recreating the client is unnecessary when the Catalog configuration has not changed.


zy-kkk:
run buildall

zy-kkk:
run buildall

morningman (approved):
LGTM

PR #40111, opened 2024-08-29
[fix](ES Catalog)Do not push down limit to ES when predicates can not be processed by ES.

qidaye:
## Proposed changes

Issue Number: close #xxx

PR #40265, opened 2024-09-02
[fix](ES Catalog)Do not push down limit to ES when predicates can not be processed by ES. (#40111)

qidaye:
bp #40111


qidaye:
run buildall

PR #40289, opened 2024-09-02
[fix](execution) fix wrong check when blocking result sink

morningman:
Otherwise, when client stop fetch result data, the result sink will still adding batch to the result queue,
which causing BE OOM.

### DIFF
--- a/be/src/runtime/buffer_control_block.cpp
+++ b/be/src/runtime/buffer_control_block.cpp
@@ -312,12 +312,12 @@ void PipBufferControlBlock::set_dependency(
 }
 
 void PipBufferControlBlock::_update_dependency() {
-    if (_result_sink_dependency &&
-        (_batch_queue_empty || _buffer_rows < _buffer_limit || _is_cancelled)) {
-        _result_sink_dependency->set_ready();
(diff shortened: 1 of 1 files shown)


morningman:
run buildall

BiteTheDDDDt (approved):

HappenLee (approved):
LGTM

be/src/runtime/buffer_control_block.cpp
```diff
@@ -316,7 +316,7 @@ void PipBufferControlBlock::_update_dependency() {
         (_batch_queue_empty || _buffer_rows < _buffer_limit || _is_cancelled)) {
```
yiguolei, line 316: if (_result_sink_dependency){
if (_batch_queue_empty || _buffer_rows < _buffer_limit || _is_cancelled)) {
_result_sink_dependency->set_ready();
} else {
_result_sink_dependency->block();
}
}

yiguolei: The old code it too complicated,  I think it is just an else logic.

morningman:
run buildall

yiguolei (approved):

PR #40290, opened 2024-09-02
[fix](decimal) throw overflow exception if result is NaN of Infinit when converting from decimal to float

jacktengg:
## Proposed changes

When casting `decimal256` to `float`, BE first converting `decimal256` to `long double`, and then converting to `float`, but for some values, the resulting float value is `std::inf`, this PR check this situation and report `Arithmetic overflow`.

CPP test:
```
int main()
{
        long double ld = std::stold("1.23456779999999999998e+67");
        std::cout << "long double: " << ld << std::endl;
        float f = static_cast<float>(ld);
        if (std::isnan(f)) {
            std::cout << "long double to float is nan" << std::endl;
        } else if (std::isinf(f)) {
            std::cout << "long double to float is inf" << std::endl;
        } else {
            std::cout << f << std::endl;
        }   
        double d = static_cast<double>(ld);
        if (std::isnan(d)) {
            std::cout << "long double to double is nan" << std::endl;
        } else if (std::isinf(d)) {
            std::cout << "long double to double is inf" << std::endl;
        } else {
            std::cout << "long double to double is " << d << std::endl;
        }   
        return 0;
}
```

result:
```
long double: 1.23457e+67
long double to float is inf
long double to double is 1.23457e+67
```


jacktengg:
run buildall

regression-test/suites/datatype_p0/decimalv3/test_decimal256_cast.groovy
```diff
@@ -41,4 +41,29 @@ suite("test_decimal256_cast") {
...
+    properties (
+        'replication_num' = '1'
+    );
+    """
+    sql """
+        insert into dec256cast_to_float values  (1, "12345678.000000000000000000000000000000001");
+    """
+    test {
+        sql """
+            select cast(v1 as float) from dec256cast_to_float;
```
yiguolei, line 61: also add cast for literal and test fold constant is true or false

regression-test/suites/datatype_p0/decimalv3/test_decimal256_cast.groovy
```diff
@@ -41,4 +41,29 @@ suite("test_decimal256_cast") {
         select cast('0.000000000000000000000000000000000000000000000000000000000000000000000012345678901' as decimalv3(76,0));
     """
 
+    sql """
+        drop table  if exists dec256cast_to_float;
```
yiguolei, line 45: The previous cast test only test const literal, should also add some test to insert the data to a table and cast them during select.

regression-test/suites/datatype_p0/decimalv3/test_decimal256_cast.groovy
```diff
@@ -41,4 +41,29 @@ suite("test_decimal256_cast") {
         select cast('0.000000000000000000000000000000000000000000000000000000000000000000000012345678901' as decimalv3(76,0));
     """
 
+    sql """
```
yiguolei, line 44: 字面常量的测试：包括打开常量折叠，关闭常量折叠
变量的测试：把数据插入到表中，然后select 执行函数
cast： 别的类型，到decimal； decimal 到别的类型。
不论是字面量还是变量，都要测试异常的值。

jacktengg:
run buildall

yiguolei (approved):

zclllyybb (approved):
LGTM

PR #40309, opened 2024-09-03
[fix](regression-test) Sync after insert to get stable results

Hastyshell:
## Proposed changes

As title.


TangSiyang2001:
run buildall

PR #40314, opened 2024-09-03
[fix](ES Catalog)Do not extract doc_values of field with ignore_above setting

qidaye:
## Proposed changes

Issue Number: close #xxx


qidaye:
run buildall

qidaye:
run buildall

morningman (dismissed):
LGTM

fe/fe-core/src/main/java/org/apache/doris/catalog/EsTable.java
```diff
@@ -52,7 +52,9 @@
 @Getter
 @Setter
 public class EsTable extends Table implements GsonPostProcessable {
-    public static final Set<String> DEFAULT_DOCVALUE_DISABLED_FIELDS = new HashSet<>(Collections.singletonList("text"));
+    // refrence: https://www.elastic.co/guide/en/elasticsearch/reference/current/doc-values.html
+    public static final Set<String> DEFAULT_DOCVALUE_DISABLED_FIELDS =
+            new HashSet<>(Arrays.asList("text", "annotated_text"));
```
xiaokang, line 57: What about match_only_text?

eldenmoon (dismissed):
LGTM

fe/fe-core/src/main/java/org/apache/doris/datasource/es/MappingPhase.java
```diff
@@ -130,6 +130,10 @@ private static void resolveDocValuesFields(SearchContext searchContext, ObjectNo
                         if (docValue) {
                             docValueField = colName;
                         }
+                    } else if (innerTypeObject.has("ignore_above")) {
```
xiaokang, line 133: we need to test "doc_values": true

xiaokang: consider implict "ignore_above": 256

fe/fe-core/src/main/java/org/apache/doris/catalog/EsTable.java
qidaye: `match_only_text`'s behavior is the same with `text` and should be included in here.

fe/fe-core/src/main/java/org/apache/doris/datasource/es/MappingPhase.java
qidaye: 1. The default value of `doc_values`  is `true`, its behavior is consistent with explicit set to `true`.
2. [keyword param](https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html#keyword-params)
> [ignore_above](https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html)
Do not index any string longer than this value. Defaults to 2147483647 so that all values would be accepted. Please however note that default dynamic mapping rules create a sub keyword field that overrides this default by setting ignore_above: 256. 

`ignore_above` default value is `2147483647`. 
The special case is the keyword field generated by dynamic mapping, whose `ignore_above` setting is `256` , which will be shown in mapping explicitly, and we have taken it into consideration.

qidaye:
run buildall

qidaye:
run buildall

xiaokang (approved):
LGTM

zzzxl1993 (approved):
LGTM

PR #36743, opened 2024-06-24
[Chore](execution) remove unused function get_least_supertype

BiteTheDDDDt:
## Proposed changes
remove unused function get_least_supertype

PR #40317, opened 2024-09-03
[ut](data type) fix get_common_type_test

eldenmoon:


eldenmoon:
run buildall

be/test/vec/core/get_common_type_test.cpp
```diff
@@ -143,10 +146,8 @@ INSTANTIATE_TEST_SUITE_P(data_type, LeastSuperTypeTest,
                                  {"Float32 Float64", "Float64"},
                                  {"Float32 UInt16 Int32", "Float64"},
                                  {"Float32 Int16 UInt32", "Float64"},
-                                 {"Date Date", "Date"},
```
zhangstar333, line 146: it's should be fixed at be/src/vec/data_types/get_least_supertype.cpp line:288
add datev2 and datetimev2 
`        if (which.is_date_v2()) {
            *type = std::make_shared<DataTypeDateV2>();
            return;
        }
        if (which.is_date_time_v2()) {
            *type = std::make_shared<DataTypeDateTimeV2>();
            return;
        }`

be/test/vec/core/get_common_type_test.cpp
```diff
@@ -143,10 +146,8 @@ INSTANTIATE_TEST_SUITE_P(data_type, LeastSuperTypeTest,
...
-                                 {"DateTime DateTime", "DateTime"},
                                  {"String String String", "String"},
-                                 {"Int8 String", nullptr},
```
zhangstar333, line 149: I think could remove those, no need change to Jsonb

be/test/vec/core/get_common_type_test.cpp
eldenmoon: @zhangstar333 related code removed by PR #36743

be/test/vec/core/get_common_type_test.cpp
eldenmoon: need to change to jsonb， since they conflict

be/test/vec/core/get_common_type_test.cpp
zhangstar333: image

be/test/vec/core/get_common_type_test.cpp
zhangstar333: could remove those test case of return nullptr
now return jsonb seems the default result as get_least_supertype_jsonb.
so those test case is useless

be/test/vec/core/get_common_type_test.cpp
eldenmoon: done

be/test/vec/core/get_common_type_test.cpp
eldenmoon: done

eldenmoon:
run buildall

zhangstar333 (approved):

yiguolei (approved):
LGTM

PR #40318, opened 2024-09-03
[fix](mtmv)Fix that after insert overwrite, the partition version will always be 2

zddr:
The replaced partition version should inherit from the original partition


zddr:
run buildall

PR #40333, opened 2024-09-03
[fix](cloud-mow) make cloud_txn_delete_bitmap_cache's  expired time more reasonable

hust-hhb:
Now cloud_txn_delete_bitmap_cache's expired time is relay rely on txn_timeout_s, however if the cost time of calculating delete bitmap is bigger than txn_timeout_s, updating the publish status of cloud_txn_delete_bitmap_cache will be failed, because this cache may be remove by the cleaning thread. In rountine load, this txn_timeout_s is rely on max_batch_interval session variable, so if someone use small interval  to submit rountine load task, cloud_txn_delete_bitmap_cache may be removed when calculating delete bitmap task is going, so we need to set a min expired time on cloud_txn_delete_bitmap_cache to avoid this scenario.

### DIFF
--- a/be/src/cloud/cloud_txn_delete_bitmap_cache.cpp
+++ b/be/src/cloud/cloud_txn_delete_bitmap_cache.cpp
@@ -154,11 +154,11 @@ void CloudTxnDeleteBitmapCache::set_tablet_txn_info(
 }
 
 Status CloudTxnDeleteBitmapCache::update_tablet_txn_info(TTransactionId transaction_id,
-                                                       int64_t tablet_id,
-                                                       DeleteBitmapPtr delete_bitmap,
-                                                       const RowsetIdUnorderedSet& rowset_ids,
-                                                       PublishStatus publish_status,
-                                                       TxnPublishInfo publish_info) {
+                                                         int64_t tablet_id,
+                                                         DeleteBitmapPtr delete_bitmap,
+                                                         const RowsetIdUnorderedSet& rowset_ids,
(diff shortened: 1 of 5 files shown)


hust-hhb:
run buildall

be/src/cloud/cloud_txn_delete_bitmap_cache.cpp
```diff
@@ -184,26 +184,34 @@ void CloudTxnDeleteBitmapCache::update_tablet_txn_info(TTransactionId transactio
...
         int64_t current_time = duration_cast<std::chrono::seconds>(
                                        std::chrono::system_clock::now().time_since_epoch())
                                        .count();
-        if (iter->first > current_time) {
+        if (it->first > current_time) {
             break;
         }
-        auto txn_iter = _txn_map.find(iter->second);
-        if ((txn_iter != _txn_map.end()) && (iter->first == txn_iter->second.txn_expiration)) {
+        if (*(_txn_map.find(it->second)->second.publish_status) != PublishStatus::SUCCEED) {
```
zhannngchen, line 205: If you don't clear the txn_info if it's not in SUCCEED status, the expiration time is use less
And if the txn failed, when would it be cleared?

hust-hhb:
run buildall

zhannngchen (dismissed):
LGTM

hust-hhb:
run buildall

zhannngchen (approved):

Yukang-Lian (approved):

PR #40351, opened 2024-09-04
Revert "[enhancement](index compaction) Enable index compaction by default (#36812)"

qidaye:
Reverts apache/doris#38676


qidaye:
run buildall

PR #40363, opened 2024-09-04
[fix](array/map) Fix semi-structure data column

Gabriel39:
## Proposed changes

Function `convert_to_full_column_if_const` will return a shared pointer with the origin column and add 1 to `use count` of this column if it is not a Array/Map column.

For Array/Map columns, `convert_to_full_column_if_const` will return a new pointer which shares the internal nested columns with the origin columns.

During execution, a column which is exclusive means it is not shared by another pointer so we can use this column as we want otherwise we should copy the data into another exclusive column.

Because `convert_to_full_column_if_const` will return a new pointer for Array/Map columns, we can not just `use count == 1` to decide if the column is exclusive.

### DIFF
--- a/be/src/vec/columns/column_array.h
+++ b/be/src/vec/columns/column_array.h
@@ -232,6 +232,10 @@ public:
                        ->get_number_of_dimensions(); /// Every modern C++ compiler optimizes tail recursion.
     }
 
+    bool is_exclusive() const override {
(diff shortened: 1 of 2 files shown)


Gabriel39:
run buildall

amorynan:
please add cases!

PR #40367, opened 2024-09-04
[fix](multi table) do not use strlen to calculate the length of msg

sollhui:
Meet code dump when using single stream multi table load:
```
SUMMARY: AddressSanitizer: heap-buffer-overflow /root/doris/be/src/io/fs/multi_table_pipe.cpp:99:22 in doris::io::MultiTablePipe::dispatch(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, char const*, unsigned long, doris::Status (doris::io::KafkaConsumerPipe::*)(char const*, unsigned long))
```

1. It is hard to guaranteed that  msg is a C-style string ending in '\0' character. If not, it may cause the core dump to access memory out of bounds.
2. It is not need to calculate the length of msg twice.

Therefore,  deleting the logic that using strlen to calculate the length of msg.

### DIFF
--- a/be/src/io/fs/multi_table_pipe.cpp
+++ b/be/src/io/fs/multi_table_pipe.cpp
@@ -96,7 +96,7 @@ std::string MultiTablePipe::parse_dst_table(const char* data, size_t size) {
 
 Status MultiTablePipe::dispatch(const std::string& table, const char* data, size_t size,
                                 AppendFunc cb) {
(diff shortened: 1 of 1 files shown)


sollhui:
run buildall

sollhui:
run buildall

liaoxin01 (approved):
LGTM

dataroaring (approved):
LGTM

PR #40374, opened 2024-09-04
[fix](mtmv) Generate mtmv cache should use ADMIN user, and rewritten plan should not check privilege

seawinde:
## Proposed changes
1. In query rewrite by materialized view, If user has the select privilege for table `lineitem` and `orders` but not mv1.
and mv def is as following:
```sql
            CREATE MATERIALIZED VIEW mv1
            BUILD IMMEDIATE REFRESH AUTO ON MANUAL 
            DISTRIBUTED BY RANDOM BUCKETS 1 
            PROPERTIES ('replication_num' = '1') 
            AS 
            select l_shipdate, o_orderdate, l_partkey, l_suppkey,
            sum(o_totalprice) as sum_total,
            max(o_totalprice) as max_total,
            min(o_totalprice) as min_total,
            count(*) as count_all,
            bitmap_union(to_bitmap(case when o_shippriority > 1 and o_orderkey IN (1, 3) then o_custkey else null end)) as bitmap_union_basic
            from lineitem
            left join orders on lineitem.l_orderkey = orders.o_orderkey and l_shipdate = o_orderdate
            group by
            l_shipdate,
            o_orderdate,
            l_partkey,
            l_suppkey;
```
if query is as follwoing, we should rewrite by mv1 sucessfully. Because the privilege check for table is passed, though user has no select privilege for mv1, but can use the rewritten plan with mv1.
```
            select t1.l_partkey, t1.l_suppkey, o_orderdate,
            sum(o_totalprice),
            max(o_totalprice),
            min(o_totalprice),
            count(*),
            count(distinct case when o_shippriority > 1 and o_orderkey IN (1, 3) then o_custkey else null end)
            from (select * from lineitem where l_shipdate = '2023-12-11') t1
            left join orders on t1.l_orderkey = orders.o_orderkey and t1.l_shipdate = o_orderdate
            group by
            o_orderdate,
            l_partkey,
            l_suppkey;
```

2. We use the `admin` user when refresh materialized view. So genenrate mtmv cache, we also need to use the `admin` user.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/MTMV.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/MTMV.java
@@ -299,7 +299,13 @@ public class MTMV extends OlapTable {
         }
         // Concurrent situations may result in duplicate cache generation,
         // but we tolerate this in order to prevent nested use of readLock and write MvLock for the table
-        MTMVCache mtmvCache = MTMVCache.from(this, connectionContext, true);
+        MTMVCache mtmvCache;
+        try {
+            // Should new context with ADMIN user
+            mtmvCache = MTMVCache.from(this, MTMVPlanUtil.createMTMVContext(this), true);
+        } finally {
+            connectionContext.setThreadLocalInfo();
+        }
         writeMvLock();
         try {
             this.cache = mtmvCache;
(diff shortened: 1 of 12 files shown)


seawinde:
run buildall

morrySnow (dismissed):

seawinde:
run buildall

seawinde:
run buildall

zddr (approved):

morrySnow (approved):

PR #40322, opened 2024-09-03
[Opt](delete) Skip newly inserted rows check in non-strict mode partial update if the row's delete sign is marked

bobhan1:
## Proposed changes

branch-2.1-pick: https://github.com/apache/doris/pull/40383
branch-2.0-pick: https://github.com/apache/doris/pull/40382

PR #40382, opened 2024-09-04
[Opt](delete) Skip newly inserted rows check in non-strict mode partial update if the row's delete sign is marked #40322

bobhan1:
picks https://github.com/apache/doris/pull/40322


bobhan1:
run buildall

bobhan1:
run p0

PR #40385, opened 2024-09-04
[fix](ES Catalog)Fix int parse error when querying by doc_values

qidaye:
## Proposed changes

When querying by doc_values, the result may be embrace by quotes, which will lead the int parsing error.


qidaye:
run buildall

morningman (approved):

eldenmoon (approved):
LGTM

PR #40395, opened 2024-09-04
[Opt](scanner-scheduler) Opt scanner scheduler starvation issue by yield signal.

kaka11chen:
### Issue
When a scanner scheduler is stuck in executing a scan task, other scan tasks will starve and have no chance to execute, which will affect other queries. Currently, the scan task hopes to scan as much data as possible to reduce the overhead of scheduling switching. Currently, it hopes to obtain up to 10MB of data in `doris_scanner_row_bytes`. However, if a query scans a table with many rows of data, but the filtering rate is very high, the filter will eventually filter out a lot of data and will never get 10MB of data. It will keep getting and executing expression filtering, which will cause other scan tasks to starve.

### Solution
The current solution is to set a yield signal. After executing for a maximum of 1s, the state will be set to yield. When the scan task executes some time-consuming tasks, it needs to slice to do it. Each slice area checks the yield signal. If it is found to be in the yield state, it will exit early and make room for other scan tasks to execute.
We originally used the original number of rows in some places in the code to express similar semantics, but I feel that it is not accurate in time. The parquet reader may skip many rows due to indexes, but sometimes it does not.

### DIFF
--- a/be/src/common/config.cpp
+++ b/be/src/common/config.cpp
@@ -252,6 +252,10 @@ DEFINE_Int32(remote_split_source_batch_size, "10240");
 DEFINE_Int32(doris_max_remote_scanner_thread_pool_thread_num, "-1");
 // number of olap scanner thread pool queue size
 DEFINE_Int32(doris_scanner_thread_pool_queue_size, "102400");
+// number of scanner yield signal thread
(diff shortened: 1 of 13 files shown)


kaka11chen:
run buildall

kaka11chen:
run buildall

kaka11chen:
run buildall

PR #40398, opened 2024-09-05
[fix](nestedtype)fix nested type with is_exclusive

amorynan:
## Proposed changes
is_exclusive in column_array/map/struct has wrong semantic , we should make sure it's nested column is also is_exclusive which can make behavior right in some operator  like join.
Issue Number: close #xxx


amorynan:
run buildall

be/src/vec/columns/column_array.h
```diff
@@ -122,6 +122,11 @@ class ColumnArray final : public COWHelper<IColumn, ColumnArray> {
     const char* get_family_name() const override { return "Array"; }
```
eldenmoon, line 122: what about ColumnObject?

Gabriel39 (dismissed):

amorynan:
run buildall

be/src/vec/columns/column_object.cpp
```diff
@@ -1877,6 +1877,10 @@ void ColumnObject::for_each_imutable_subcolumn(ImutableColumnCallback callback)
     }
 }
 
+bool ColumnObject::is_exclusive() const {
```
eldenmoon, line 1880: if subcolumns size is 2, then the first is returned as result, is this correct?

be/src/vec/columns/column_object.cpp
```diff
@@ -1877,6 +1877,10 @@ void ColumnObject::for_each_imutable_subcolumn(ImutableColumnCallback callback)
...
+    for_each_imutable_subcolumn([](const auto& subcolumn) { return subcolumn.is_exclusive(); });
```
eldenmoon, line 1881: bool is_exclusive = true
for_each_imutable_subcolumn([](const auto& subcolumn) { if (!subcolumn.is_exclusive()) {is_exclusive = false} });
return is_exclusive && IColumn::is_exclusive();;

amorynan:
run buildall

eldenmoon (approved):
LGTM

Gabriel39 (approved):

PR #40399, opened 2024-09-05
[chore](log) Do not print profile in be log if query comes from FE

zhiqiang-hhhh:
For stream load and queries from external connector, we will print profile in be.log. For others, should not do it because insert into select and broker load already saved profile in  FE.


zhiqiang-hhhh:
run buildall

yiguolei (changes requested):
Do not modify it. Broker load and insert into select need this logic to debug.

liaoxin01 (approved):
LGTM

yiguolei (approved):

PR #39747, opened 2024-08-22
[fix](auth)Fix ordinary users can see the processes of other users

zddr:
Fix when `set show_all_fe_connection="true"; `At this time, ordinary users can see the processes of other users,

PR #40417, opened 2024-09-05
[fix](auth) ordinary users can see the processes of other users (#39747)

zddr:
pick: https://github.com/apache/doris/pull/39747


zddr:
run buildall

PR #40438, opened 2024-09-05
[Bug](compatibility) fix stddev_samp function coredump when upgrade

zhangstar333:
## Proposed changes

```
            if (IAggregateFunction::version < AGG_FUNCTION_NULLABLE) {
                return make_nullable(Data::get_return_type());
            } else {
                return Data::get_return_type();
            }
```

before check two version , and then get return type,
but now in branch-21, the IAggregateFunction::version have update to 5,
and also AGG_FUNCTION_NEW=5
so the check will not get right return type.

### DIFF
--- a/be/src/vec/aggregate_functions/aggregate_function_stddev.h
+++ b/be/src/vec/aggregate_functions/aggregate_function_stddev.h
@@ -126,8 +126,6 @@ struct BaseData {
         count += 1;
     }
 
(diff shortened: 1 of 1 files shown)


zhangstar333:
run buildall

hello-stephen (approved):

BiteTheDDDDt (approved):

PR #40388, opened 2024-09-04
[enhance](variant) throw exception when field type is not supported in cast elimation

eldenmoon:

PR #40448, opened 2024-09-05
[enhance](variant) throw exception when field type is not supported in cast elimation 

eldenmoon:
#40388


eldenmoon:
run buildall

PR #40463, opened 2024-09-06
[fix](cloud) should do check before abort transaction

sollhui:
When routine load task transaction is abort, it should do check before abort transaction, otherwise, it may cause concurrent modifications to the `routineLoadTaskInfoList`, which in the Java language may result in elements in the list being null, leading to a loop throwing NullPointerException during scheduling and making it impossible to schedule routine load task to consume Kafka stream.


sollhui:
run buildall

liaoxin01 (approved):
LGTM

dataroaring (approved):
LGTM

PR #38003, opened 2024-07-17
[fix](move-memtable) multi replica tables should tolerate minority failures

kaijchen:
## Proposed changes

Load job for multi replica tables shouldn't fail immediately on any single replica errors.
Errors should be recorded and reported for individual replica of tablets, and checked on commit info.

PR #40477, opened 2024-09-06
[fix](move-memtable) multi replica tables should tolerate minority failures (#38003)

kaijchen:
backport #38003


kaijchen:
run buildall

PR #40479, opened 2024-09-06
[fix](catalog) should return error if try using a unknown database

morningman:
Previous, when `use db`, and `db` does not exist,
no error return.
But it should return error: `Uknown database xxx`


morningman:
run buildall

morningman:
run buildall

zy-kkk (approved):
LGTM

wuwenchi (approved):

PR #40122, opened 2024-08-29
[fix](jdbc catalog) Fix type recognition error when using `query` tvf to query doris

zy-kkk:
Using string to match Doris type will not work with query tvf, so use field matching instead

PR #40481, opened 2024-09-06
[fix](jdbc catalog) Fix type recognition error when using `query` tvf to query doris

zy-kkk:
pick  (#40122)

Using string to match Doris type will not work with query tvf, so use field matching instead


zy-kkk:
run buildall

PR #40108, opened 2024-08-29
[fix](auth)fix case should grant after create view

zddr:
cause by:https://github.com/apache/doris/pull/39597

PR #40488, opened 2024-09-06
[fix](auth)fix case should grant after create view (#40108)

zddr:
pick: https://github.com/apache/doris/pull/40108


zddr:
run buildall

PR #40493, opened 2024-09-06
[opt](scan) unify the local and remote scan bytes stats for all scanners

morningman:
Previously, only olap table's query has local and remote bytes read statistics.
This PR add these stats for all scanners.

1. Use `CachedRemoteFileReader` no matter `enable_file_cache` is true or false

	Previously, if `enable_file_cache` is true, we use `CachedRemoteFileReader`.
	Otherwise, we use raw file reader to read data.
	
	In order to unify the query stats, in this PR, I use `CachedRemoteFileReader`
	no matter `enable_file_cache` is true or false.
	
	When reading data, if cache is disable, `CachedRemoteFileReader` will use
	the raw file reader in it directly.
	
2. Add `_update_bytes_and_rows_read()` interface in `VScanner`

	This method will be called after each `get_block()` method.
	It will update the scan bytes and rows in query statistics.
	So that we can get real time statistics when querying system table `backend_active_tasks`
	
3. Add `REMOTE_SCAN_BYTES` and `LOCAL_SCAN_BYTES` columns in `backend_active_tasks`

	`REMOTE_SCAN_BYTES` is bytes read from remote fs.
	`LOCAL_SCAN_BYTES` is bytes read from local disks.
	
	And `SCAN_BYTES` is now the sum of  `REMOTE_SCAN_BYTES` and `LOCAL_SCAN_BYTES`

4. Add new columns for audit log table

	- `local_scan_bytes`
	- `remote_scan_bytes`
	- `shuffle_bytes`
	- `shuffle_rows`
	- `cloud_cluster_name`

### DIFF
--- a/be/src/io/fs/buffered_reader.cpp
+++ b/be/src/io/fs/buffered_reader.cpp
@@ -394,7 +394,7 @@ Status MergeRangeFileReader::_fill_box(int range_index, size_t start_offset, siz
 // there exists occasions where the buffer is already closed but
 // some prior tasks are still queued in thread pool, so we have to check whether
 // the buffer is closed each time the condition variable is notified.
-void PrefetchBuffer::reset_offset(size_t offset) {
+void PrefetchBuffer::reset_offset(size_t offset, const IOContext* io_ctx) {
     {
         std::unique_lock lck {_lock};
         if (!_prefetched.wait_for(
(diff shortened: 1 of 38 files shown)


morningman:
run buildall

morningman:
run buildall

be/src/io/fs/buffered_reader.cpp
```diff
@@ -415,10 +415,36 @@ void PrefetchBuffer::reset_offset(size_t offset) {
...
             [buffer_ptr = shared_from_this()]() { buffer_ptr->prefetch_buffer(); });
 }
 
+void PrefetchBuffer::_update_and_reset_io_context(const IOContext* io_ctx) {
+    if (io_ctx) {
+        // If file_cache_stats is set in the input parameter,
+        // first need to update the last statistics in _owned_cache_stats
+        // to the file_cache_stats in the input parameter.
+        // Then reset _owned_cache_stats
+        if (io_ctx->file_cache_stats) {
```
zhiqiang-hhhh, line 436: potential data race ?

be/src/io/fs/buffered_reader.cpp
```diff
@@ -415,10 +415,36 @@ void PrefetchBuffer::reset_offset(size_t offset) {
...
+    // its life cycle may be different from that of the caller.
+    // Therefore, before submitting PrefetchBuffer to the thread pool,
+    // we need to copy io_ctx to _owned_io_ctx to ensure that
+    // the life cycle of IOContext is consistent with that of PrefetchBuffer.
+    // _update_and_reset_io_context(io_ctx);
     _prefetch_status = ExecEnv::GetInstance()->buffered_reader_prefetch_thread_pool()->submit_func(
             [buffer_ptr = shared_from_this()]() { buffer_ptr->prefetch_buffer(); });
 }
 
+void PrefetchBuffer::_update_and_reset_io_context(const IOContext* io_ctx) {
```
zhiqiang-hhhh, line 430: when does this method is called?

PR #40037, opened 2024-08-28
[chore](routine load)  make error msg clear if routine load name illegal

sollhui:
Make error msg clear if routine load name illegal(such as job name is too long).

PR #40499, opened 2024-09-07
[chore](routine load)  make error msg clear if routine load name illegal (#40037)

sollhui:
pick (#40037)


sollhui:
run buildall

PR #40505, opened 2024-09-07
[fix](hive)Modify the Hive notification event processing method when using meta cache and add parameters to the Hive catalog. (#39239) 

hubgeter:
bp #39239
just pick : fix event npe   
## Proposed changes

Issue Number: close #xxx


hubgeter:
run buildall

PR #40563, opened 2024-09-09
[Fix](inverted index) gc TEMP colum when next_batch in segment iterator

airborne12:
## Proposed changes

Issue Number: close #xxx


airborne12:
run p1

PR #40575, opened 2024-09-10
[Fix](inverted index) gc TEMP colum when next_batch in segment iterator #40563

airborne12:
cherry pick from #40563


airborne12:
run buildall

airborne12:
run p1

PR #40579, opened 2024-09-10
[optimize](schema change) optimize of schema change for cooldown data

DarvenDuan:
## Proposed changes
This PR solves 2 problems of schema change for cooldown data:
1. Schema change will build a new tablet for base tablet, and all replicas do this separately, so the cooldown data SC writes all replicas to the remote storage, But the cooldown rowsets  just need one master replica to cooldown, the other replicas follow the master. To optimize this issue, Doris can just write the new rowsets to the local storage, and the tablet will cooldown this rowsets automatically for just one mater replica.

2. In schema change job and rollup job, FE generates `createReplicaTask`, which specifies the storage policy of the tablets to create, table's storage policy may be empty if we just set the partitions' storage policy, so we use partition's storage policy instead of table.


DarvenDuan:
run buildall

DarvenDuan:
run compile

DarvenDuan:
run buildall

DarvenDuan:
run p0

DarvenDuan:
run buildall

PR #40324, opened 2024-09-03
[Fix](inverted index) fix profile metrics for inverted index after refactor

airborne12:
## Proposed changes
Make profile metrics right after #38908

PR #40583, opened 2024-09-10
[Fix](inverted index) fix profile metrics for inverted index after refactor #40324

airborne12:
cherry pick from #40324


airborne12:
run buildall

PR #40338, opened 2024-09-03
[Fix](inverted index) fix comparison for string after refactor compound

airborne12:
## Proposed changes
Untokenized strings exceed ignore_above, they are written as null, causing range query errors.
Introduced by #38908

PR #40585, opened 2024-09-10
[Fix](inverted index) fix comparison for string after refactor compound #40338

airborne12:
cherry pick from #40338


airborne12:
run buildall

PR #40232, opened 2024-09-02
[Chore](inverted index) remove useless code in scan operator for normalize compound and match predicate

airborne12:
## Proposed changes
Remove unused code in scan operator for normalizing compound and match predicate

PR #40587, opened 2024-09-10
[Chore](inverted index) remove useless code in scan operator for normalize compound and match predicate #40232

airborne12:
cherry pick from #40232


airborne12:
run buildall

PR #40612, opened 2024-09-10
[fix](restore) Persist the sqlMode field of the View

w41ter:
During initialization, the View will parse inlineViewRef again according to the sqlMode value. Therefore, sqlMode must be persisted.


morrySnow (approved):

dataroaring (approved):
LGTM

PR #40615, opened 2024-09-10
[fix](restore) Remove the sqlMode from the view signature

w41ter:
To keep compatible with the old version, without sqlMode, if the signature of views are the same, we should have the same sqlMode. (since the sqlMode doesn't affect the parsing of inlineViewDef, otherwise the parsing will fail)


dataroaring (approved):
LGTM

gavinchou (approved):

PR #40616, opened 2024-09-10
Revert "[improvement](statistics)Return -1 to neredis if report olap table row count for new table is not done for all tablets. (#40457)"

yiguolei:
Reverts apache/doris#40540


yiguolei:
skip buildall

PR #40619, opened 2024-09-10
[Fix](nereids) fix partition prune not expression compute logic

feiniaofeiafei:
In partition pruning, whether an expression supports NOT computation depends on the rejectNot boolean flag. If rejectNot is true, it indicates support for the computation. For example, consider partition column a with a range of (1,10) in partition1. For the expression NOT (a > 5), the computation process involves first evaluating a > 5. Combined with the partition range, this results in 5 < a < 10 for a > 5. Then, applying NOT (a > 5) leads to the range 1 < a <= 5 if rejectNot is true, meaning the calculation is valid.

However, this calculation should only be supported for certain operators like <, <=, >, >= and =. In this pull request (PR), the default value of rejectNot is set to false, reducing the range of expressions that support NOT computation. This change prevents unsupported expressions, such as LIKE, from producing incorrect results.


feiniaofeiafei:
run buildall

PR #40624, opened 2024-09-10
[fix](regression) speed up some group commit case

mymeiyi:
group commit interval is default 10s, cause the case cost too much time


mymeiyi:
run buildall

dataroaring (approved):
LGTM

gavinchou (approved):

PR #40625, opened 2024-09-10
[Fix](Job)Replace BlockingWaitStrategy with LiteTimeoutBlockingWaitStrategy to avoid deadlock issues.

CalvinKirs:
## Proposed changes

FYI https://issues.apache.org/jira/browse/LOG4J2-1221
- BlockingWaitStrategy is a wait strategy used in the Disruptor framework that blocks the thread when the ring buffer is full or not yet available for publishing.

When threads are blocked, they are waiting for space in the ring buffer to become available, which can lead to potential deadlocks if not managed properly.
Timeout Handling:

- LiteTimeoutBlockingWaitStrategy provides a timeout for waiting threads. If the buffer is not ready within the timeout period, the thread is released, preventing it from being blocked indefinitely.
Reduced Risk of Deadlocks:

- By avoiding indefinite blocking, this strategy reduces the risk of deadlocks caused by threads waiting on each other. The timeout allows the system to handle scenarios where resources are temporarily

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/job/disruptor/TaskDisruptor.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/job/disruptor/TaskDisruptor.java
@@ -50,7 +50,7 @@ public class TaskDisruptor<T> {
                          WaitStrategy waitStrategy, WorkHandler<T>[] workHandlers,
                          EventTranslatorVararg<T> eventTranslator) {
         disruptor = new Disruptor<>(eventFactory, ringBufferSize, threadFactory,
-                ProducerType.SINGLE, waitStrategy);
+                ProducerType.MULTI, waitStrategy);
         disruptor.handleEventsWithWorkerPool(workHandlers);
         this.eventTranslator = eventTranslator;
         disruptor.start();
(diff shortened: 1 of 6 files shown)


fe/fe-core/src/main/java/org/apache/doris/job/disruptor/TaskDisruptor.java
```diff
@@ -68,9 +71,16 @@ public void start() {
...
-    public void publishEvent(Object... args) {
-        RingBuffer<T> ringBuffer = disruptor.getRingBuffer();
-        ringBuffer.publishEvent(eventTranslator, args);
+    public boolean publishEvent(Object... args) {
+        try {
+            RingBuffer<T> ringBuffer = disruptor.getRingBuffer();
+            ringBuffer.publishEvent(eventTranslator, args);
+            return true;
+        } catch (Exception e) {
+            LOG.error("Failed to publish event", e);
```
morningman, line 80:
```suggestion
            LOG.warn("Failed to publish event", e);
```

fe/fe-core/src/main/java/org/apache/doris/job/executor/TimerJobSchedulerTask.java
```diff
@@ -44,7 +44,10 @@ public void run(Timeout timeout) {
                 log.info("job status is not running, job id is {}, skip dispatch", this.job.getJobId());
                 return;
             }
-            dispatchDisruptor.publishEvent(this.job);
+            if (dispatchDisruptor.publishEvent(this.job)) {
```
morningman, line 47:
```suggestion
            if (!dispatchDisruptor.publishEvent(this.job)) {
```

CalvinKirs:
run buildall

CalvinKirs:
run buildall

morningman (approved):
LGTM

morrySnow (approved):

PR #40566, opened 2024-09-09
[fix](regression-test) Fix scanner profile test failed occasionally

zhiqiang-hhhh:
Collection of profile is async. So wait 5 seconds before get profile.

adaptive_pipeline_task_serial_read_on_limit will fail on multi backends env, since tablets num on each be will not be 10.

PR #40639, opened 2024-09-10
[fix](regression-test) Fix scanner profile test failed occasionally #40566

zhiqiang-hhhh:
cherry pick from #40566


zhiqiang-hhhh:
run buildall

PR #40641, opened 2024-09-10
[Opt](scanner-scheduler) Opt scanner scheduler starvation issue.

kaka11chen:
### Issue
When a scanner scheduler is stuck in executing a scan task, other scan tasks will starve and have no chance to execute, which will affect other queries. Currently, the scan task hopes to scan as much data as possible to reduce the overhead of scheduling switching. Currently, it hopes to obtain up to 10MB of data in `doris_scanner_row_bytes`. However, if a query scans a table with many rows of data, but the filtering rate is very high, the filter will eventually filter out a lot of data and will never get 10MB of data. It will keep getting and executing expression filtering, which will cause other scan tasks to starve.

### Solution
The current solution is to check `max_run_time_ms` by `MonotonicStopWatch`. After executing for a maximum of 1s,  it will yield self's task for other tasks. When the scan task executes some time-consuming tasks, it needs to slice to do it.

### DIFF
--- a/be/src/common/config.cpp
+++ b/be/src/common/config.cpp
@@ -284,6 +284,8 @@ DEFINE_mInt32(doris_scan_range_max_mb, "1024");
 DEFINE_mInt32(doris_scanner_row_num, "16384");
 // single read execute fragment row bytes
 DEFINE_mInt32(doris_scanner_row_bytes, "10485760");
+// single read execute fragment max run time millseconds
+DEFINE_mInt32(doris_scanner_max_run_time_ms, "1000");
 // (Advanced) Maximum size of per-query receive-side buffer
 DEFINE_mInt32(exchg_node_buffer_size_bytes, "20485760");
 DEFINE_mInt32(exchg_buffer_queue_capacity_factor, "64");
--- a/be/src/vec/exec/scan/scanner_scheduler.cpp
+++ b/be/src/vec/exec/scan/scanner_scheduler.cpp
@@ -236,6 +236,8 @@ void ScannerScheduler::_scanner_scan(std::shared_ptr<ScannerContext> ctx,
         Thread::set_thread_nice_value();
     }
 #endif
+    MonotonicStopWatch max_run_time_watch;
+    max_run_time_watch.start();
     scanner->update_wait_worker_timer();
     scanner->start_scan_cpu_timer();
     Status status = Status::OK();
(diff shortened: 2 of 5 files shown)


kaka11chen:
run buildall

kaka11chen:
run buildall

kaka11chen:
run buildall

morningman (dismissed):
LGTM

be/src/vec/exec/scan/vscanner.cpp
```diff
@@ -127,9 +127,11 @@ Status VScanner::get_block(RuntimeState* state, Block* block, bool* eof) {
 
             // 2. Filter the output block finally.
             {
-                auto* timer = _local_state->_filter_timer;
-                SCOPED_TIMER(timer);
-                RETURN_IF_ERROR(_filter_output_block(block));
+                if (!block->empty()) {
```
zhiqiang-hhhh, line 130: in what situation will the block in queue be empty?

be/src/common/config.cpp
```diff
@@ -281,6 +281,8 @@ DEFINE_mInt32(doris_scan_range_max_mb, "1024");
 DEFINE_mInt32(doris_scanner_row_num, "16384");
 // single read execute fragment row bytes
 DEFINE_mInt32(doris_scanner_row_bytes, "10485760");
+// single read execute fragment max run time millseconds
+DEFINE_mInt32(doris_scanner_max_run_time_ms, "1000");
```
yiguolei, line 285: change it to session variable?

kaka11chen:
run buildall

be/src/common/config.cpp
kaka11chen: I think it is a system configuration, similar with `doris_scanner_row_bytes` or `doris_scanner_row_num`, should not changed by session level.

be/src/vec/exec/scan/scanner_scheduler.cpp
```diff
@@ -270,6 +272,10 @@ void ScannerScheduler::_scanner_scan(std::shared_ptr<ScannerContext> ctx,
                     eos = true;
                     break;
                 }
+                if (max_run_time_watch.elapsed_time() / 1e6 >
+                    config::doris_scanner_max_run_time_ms) {
```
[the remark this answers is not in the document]
kaka11chen: ok

zhiqiang-hhhh (approved):
LGTM

kaka11chen:
run buildall

morningman (approved):

PR #40671, opened 2024-09-11
[fix](default value) Avoid default precision value of 0 for default expressions during deserialization

zy-kkk:
This PR fixes the issue where the precision of default value expressions was incorrectly set to `0` during deserialization. The `gsonPostProcess` method no longer assigns a default precision of `0` if the precision is not provided. Additionally, logic has been added to ensure that expressions without precision do not incorrectly apply a precision of `0`.


zy-kkk:
run buildall

morrySnow (approved):

BePPPower (approved):

PR #40679, opened 2024-09-11
[opt](function) Optimize instr by avoiding repeated traversal of strl.

Mryange:
## Proposed changes
Previously, get_char_len was repeatedly traversed during execution. Now, it is precomputed for optimization.

```C++
    static int execute(const StringRef& strl, const StringRef& strr) {
        if (strr.size == 0) {
            return 1;
        }
        StringSearch search(&strr);
        // Hive returns positions starting from 1.
        int loc = search.search(&strl);
        if (loc > 0) {
            size_t len = std::min((size_t)loc, strl.size);
            loc = simd::VStringFunctions::get_char_len(strl.data, len);
        }
        return loc + 1;
    }
```
Issue Number: close #xxx

### DIFF
--- a/be/src/vec/functions/function_string.cpp
+++ b/be/src/vec/functions/function_string.cpp
@@ -229,19 +229,28 @@ struct StringInStrImpl {
     using ResultDataType = DataTypeInt32;
     using ResultPaddedPODArray = PaddedPODArray<Int32>;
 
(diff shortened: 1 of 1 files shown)


Mryange:
run buildall

Mryange:
run p0

PR #40683, opened 2024-09-11
[bugfix](hive)Remove redundant split operations

wuwenchi:
## Proposed changes

1. In explain, we don't need to get all the splits.
2. All scans in Hive share the same scheduleExecutor. If a table with a large amount of data is specified during explain, the background will continue to scan after the explain is completed. If the user queries another small Hive table at this time, the operation will be stuck.


wuwenchi:
run buildall

morningman (approved):
LGTM

suxiaogang223 (approved):

PR #38908, opened 2024-08-06
[Refactor](inverted index) refactor inverted index compound predicates evaluate logic

airborne12:
## Proposed changes

This PR addresses several key issues related to the compound condition support in the inverted index, and optimization for index skipping without returning to the table:

1. **Unified Handling of `expr` and `column predicate`**:
   - Combined the processing of inverted index-related `column predicate` and `expr`.
   - Ensured that compound conditions involving both `column predicate` and `expr` are processed uniformly to reduce complexity and improve robustness.
   
2. **Optimized the Execution of Compound Conditions**:
   - Removed the logic in `scan_operator` that normalized compound predicates by pushing down logic to `_common_expr_ctxs_push_down` where `expr` contexts are managed.
   - Added `evaluate_inverted_index` support to the `vexpr` and function layers, such as `function comparison` and `function collection_in`.
      - Introduced new data structures in `VExprContext` to store results from `evaluate_inverted_index`, thus facilitating quick lookup and application of these results during execution.

PR #40689, opened 2024-09-11
[Fix](inverted index) Fix wrong need read data opt when enable_common_expr_pushdown is disabled

airborne12:
## Proposed changes

If common_expr_pushdown is disabled, we cannot guarantee that all conditions are processed by the inverted index after refactor of #38908 
Consider a scenario where there is a column predicate and an expression involving the same column in the SQL query,
such as 'a < 0' and 'abs(a) > 1'. 
This could potentially lead to errors.


airborne12:
run buildall

BiteTheDDDDt (approved):

csun5285 (approved):
LGTM

xiaokang (approved):
LGTM

PR #40696, opened 2024-09-11
[bugfix](loadmem) load memory is too small because flush memory is not right

yiguolei:
## Proposed changes

image

to block will clear input block and will release some memory, so if scoped, then the result will be small


yiguolei:
run buildall

xinyiZzz (approved):
LGTM

kaijchen (approved):

PR #40703, opened 2024-09-12
[Fix](inverted index) Fix wrong need read data opt when enable_common_expr_pushdown is disabled #40689

airborne12:
cherry pick from #40689


airborne12:
run buildall

PR #40707, opened 2024-09-12
[branch-2.1][Fix](Job)Replace BlockingWaitStrategy with LiteTimeoutBlockingWaitStrategy to avoid deadlock issues. (#40625)

CalvinKirs:
…

FYI https://issues.apache.org/jira/browse/LOG4J2-1221

- BlockingWaitStrategy is a wait strategy used in the Disruptor framework that blocks the thread when the ring buffer is full or not yet available for publishing.

When threads are blocked, they are waiting for space in the ring buffer to become available, which can lead to potential deadlocks if not managed properly.
Timeout Handling:

- LiteTimeoutBlockingWaitStrategy provides a timeout for waiting threads. If the buffer is not ready within the timeout period, the thread is released, preventing it from being blocked indefinitely. Reduced Risk of Deadlocks:

- By avoiding indefinite blocking, this strategy reduces the risk of deadlocks caused by threads waiting on each other. The timeout allows the system to handle scenarios where resources are temporarily

(cherry picked from commit 087048f2b2f06eae15e80e7ffa209610c9e3f173)

## Proposed changes

Issue Number: close #xxx


yiguolei:
run buildall

PR #40708, opened 2024-09-12
[bugfix](hive)Handle exceptions when submitting tasks fails to prevent dead loop

wuwenchi:
## Proposed changes

When the thread pool is full, an exception will be thrown and we need to handle this exception


wuwenchi:
run buildall

morningman (approved):
LGTM

suxiaogang223 (approved):

PR #40714, opened 2024-09-12
[fix](memory) Allows to enable memory tracker accuracy detection via Config

xinyiZzz:
add `crash_in_memory_tracker_inaccurate`, If memory tracker value is inaccurate, BE will crash. usually used in test environments, default value is false.


xinyiZzz:
run buildall

be/src/common/config.cpp
```diff
@@ -150,6 +150,9 @@ DEFINE_mInt64(stacktrace_in_alloc_large_memory_bytes, "2147483648");
 
 DEFINE_mInt64(crash_in_alloc_large_memory_bytes, "-1");
 
+// If memory tracker value is inaccurate, BE will crash. usually used in test environments, default value is false.
+DEFINE_mBool(crash_in_memory_tracker_inaccurate, "false");
```
yiguolei, line 154: also modify this parameter in p0 p1 external table pipeline

yiguolei (dismissed):

be/src/common/config.cpp
xinyiZzz: done

xinyiZzz:
run buildall

xinyiZzz:
run buildall

yiguolei (approved):

wangbo (approved):
LGTM

PR #40720, opened 2024-09-12
[fix](compile) failed on MacOS because size_t is not uint64_t on MacOS (#39296)

cambyzju:
pick #39296 to branch-2.1


cambyzju:
run buildall

PR #40721, opened 2024-09-12
fix the createCatalog in CatalogMgr.java

zhangm365:
## Proposed changes

The process of `createCatalog` will be changed now:

1. First, check if `catalogName` exists;
2. If it exists, directly throw a prompt error message;
3. If it does not exist, proceed with the process of creating the catalog.

**Previous process:**
1. First, the catalog is directly created;
2. Then, `catalogName` is checked for existence.

**Problem:** This logic will result in multiple catalogs being created when the `create catalog` statement is executed multiple times, and it will not prompt an error message indicating that the catalog has already been created.

**Outcome:** The `createFromStmt` method will instantiate a new object, which may lead to a memory leak.


zhangm365:
run buildall

zhangm365:
run buildall

PR #40336, opened 2024-09-03
[bugfix](hms)add default value check for hms catalog with dlf

wuwenchi:
## Proposed changes

1. `dlf` is not supported default values, so throw an exception.
2. The database name needs to be case-insensitive and consistent with hive
3. Improve error message prompt information

PR #40723, opened 2024-09-12
[bugfix](hms)add default value check for hms catalog with dlf (#40336)

wuwenchi:
## Proposed changes

bp #40336


wuwenchi:
run buildall

PR #37194, opened 2024-07-03
[improve](function) opt aes_encrypt/decrypt function to handle const column

zhangstar333:
## Proposed changes
```
select to_base64(aes_encrypt(column,'F3229A0B371ED2D9441B830D21A390C3', '0123456789'));
In most case of user, the second and third argument are const literal.
so handle the special case.

mysql [test]>select count(aes_encrypt(a, 'seed1seed2seed3','0123456789')) from json_table_2;
+-----------------------------------------------------------+
| count(aes_encrypt(a, '***', '0123456789', 'AES_256_CBC')) |
+-----------------------------------------------------------+
|                                                  10000001 |
+-----------------------------------------------------------+
1 row in set (5.20 sec)



mysql [test]>select count(aes_encrypt(a, 'seed1seed2seed3','0123456789')) from json_table_2;
+-----------------------------------------------------------+
| count(aes_encrypt(a, '***', '0123456789', 'AES_256_CBC')) |
+-----------------------------------------------------------+
|                                                  10000001 |
+-----------------------------------------------------------+
1 row in set (7.05 sec)

```

PR #40726, opened 2024-09-12
[fix](encrypt) wrong mode arg of encrypt and decrypt function make BE crash

cambyzju:
Not supported mode arg in sm4_encrypt and sm4_decrypt make BE crash.

Intro by https://github.com/apache/doris/pull/37194


zhangstar333 (approved):

zclllyybb (approved):
LGTM

PR #40731, opened 2024-09-12
[fix](nullable) Fix correctness problem caused by nullable judgement

Gabriel39:
## Proposed changes

In PR #13289, `has_null` is cached in NullableColumn so we get a better performance if we need to use the nullable property.

But recently we find a serious correctness problem due to this. Now I m still not clear why this will cause this problem and I just revert this part to ensure we get a correct result. 

After this PR, we need to continue to find the root cause.


Gabriel39:
run buildall

Gabriel39:
run buildall

PR #40732, opened 2024-09-12
[fix](nullable) Fix correctness problem caused by nullable judgement

Gabriel39:
## Proposed changes

pick #40731


Gabriel39:
run buildall

PR #40752, opened 2024-09-12
[Fix](Nereids) remove unstable cases

LiBinfeng-01:
remove unstable cases which does not totally controlled by hint


LiBinfeng-01:
run buildall

englefly (approved):

PR #40756, opened 2024-09-12
[fix](correctness) need to update flag of _has_null in ColumnNullable::filter_by_selector

mrhhsg:
## Proposed changes

If the flag `_need_update_has_null` is false after `ColumnNullable::filter_by_selector`, `has_null` will return incorrect value.


mrhhsg:
run buildall

zclllyybb (approved):
LGTM

cambyzju:
Why I test the regression test case on 2.1.6 and 2.0.13, both got the same result as the output in this pr.

Is there any other steps to reproduce the problem?

regression-test/suites/correctness/test_column_nullable_cache.groovy
```diff
@@ -0,0 +1,64 @@
...
+ // Unless required by applicable law or agreed to in writing,
+ // software distributed under the License is distributed on an
+ // "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ // KIND, either express or implied.  See the License for the
+ // specific language governing permissions and limitations
+ // under the License.
+
+
+/*
+The varchar column type of the view needs to be a specific number, not an asterisk
```
cambyzju, line 20: useless comments?

zclllyybb:
This pr is replaced by https://github.com/apache/doris/pull/40769

zclllyybb:
> Why I test the regression test case on 2.1.6 and 2.0.13, both got the same result as the output in this pr.
> 
> Is there any other steps to reproduce the problem?

to reproduce it, you need a environment has at least 3 BEs

PR #40761, opened 2024-09-12
[Bug](partition_topn) fix partition_topn not reset output rows after do_partition_topn_sort

zhangstar333:
## Proposed changes

should reset all variables about output rows after do_partition_topn_sort and get_next from sorter
if we not reset all variables. it's will be regard as  have output many rows
but the sorter still have remaing many data need to output.


zhangstar333:
run buildall

be/src/pipeline/exec/partition_sort_sink_operator.h
```diff
@@ -66,7 +66,12 @@ struct PartitionSortInfo {
 };
 
 static constexpr size_t INITIAL_BUFFERED_BLOCK_BYTES = 64 << 20;
+
+#ifndef NDEBUG
+static constexpr size_t PARTITION_SORT_ROWS_THRESHOLD = 10;
```
HappenLee, line 71: why change the default value？

HappenLee, line 71: why change the default value？

HappenLee (approved):
LGTM

BiteTheDDDDt (approved):

be/src/pipeline/exec/partition_sort_sink_operator.h
zhangstar333: for debug

PR #37669, opened 2024-07-11
(cloud-merge) Support to abort txn when coordinate be restart and do schema change

Lchangliang:
When do schema change, it will check whether transactions of the given database which txnId is less than 'watershedTxnId' are finished. If is not, schema change will be waiting.
When the coordinator restart (whatever fe/be), the txn belong the cooridinator will be hang until timeout. So we need to abort it in time.
There are two optimizations.
1. Check BE's lastStartTime when get it heart beat. If current lastStartTime is large than lastStartTime in fe memory. Abort all the hang txns belong  this BE.
2. Check conflict txns when do schema change. If txns is failed (Maybe coordinator be/fe restart), abort it directly.

PR #40765, opened 2024-09-12
[Fix](regression test) Fix schema change concurrent with txn regression case

Yukang-Lian:
## Proposed changes

PR #37669 introduced a new config `enable_abort_txn_by_checking_conflict_txn=true`. This config will abort schema while loading txn and restart coordinator, which causes regression test failed. We set it to false to fix this test.


dataroaring:
run buildall

cjj2010 (approved):
LGTM

dataroaring (approved):
LGTM

PR #8245, opened 2022-02-25
fix restore with partitions cause some query fail

GoGoWen:
# Proposed changes

Issue Number: close #8244

## Problem Summary:

Describe the overview of changes.

## Checklist(Required)

1. Does it affect the original behavior: (Yes/No/I Don't know)
2. Has unit tests been added: (Yes/No/No Need)
3. Has document been added or modified: (Yes/No/No Need)
4. Does it need to update dependencies: (Yes/No)
5. Are there any changes that cannot be rolled back: (Yes/No)

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #40768, opened 2024-09-12
[enhancement]support query when only restoring a few partitions of this table

Johnnyssc:
If query execute in one table that is restoring data, query may failed with Exception occasionally.
Exception info :" Table state is not NORMAL: 'RESTORING'".

But mostly restoreJob only effects some partitions in this table, not all of them, so that the other partitions need support query successfully.

The reason why this problem happen occasionally is that `Analyzer.resolveTableRef` check olapTable.state and partition.state without readLock, so fix it by check partition.state only.

ps: set restore status for partitions in this pr:https://github.com/apache/doris/pull/8245


Johnnyssc:
run build all

Johnnyssc:
run buildall

PR #40771, opened 2024-09-12
[improvement](cloud rebalance) make multiple fe consistent of tablet's backend

yujun777:
When given a tablet,  for multiple frontends,  this tablet's backends maybe different.   Because when a tablet's backend shutdown,  the frontends will immedidately change the tablet to a live backend using hash alg. Because changed no write editlog,  so the changed is not aync.  It may result in a tablet relocating to different backends for different frontends. To fix this problem, each tablet introduces two be id:  primary be id and secondary be id.  Only master fe can change primary be id actively, and it will write an editlog. So it will ensure all frontends have the same primary be. For a tablet,  if its primary be is alive, then use it as its backend,  if its primary be is dead,  then can chose a secondary be as its backend.

More details:
1.  when be dead over 1h(Config.rehash_tablet_after_be_dead_seconds), change none-colocate tablet's primary be id to its secondary be id;
2.  don't balance on colocate table, and colocate table don't save their primary be id;
3.  colocate table rehash on "alive be + be dead less 1h(Config.rehash_tablet_after_be_dead_seconds)" sets,  but if the gotten be is dead, change to an alive be. This both ensure availability and avoid all tablets rehashing when be dead for a short time;
4.  colocate tablet rehash on group id + bucket index;
5.  rehash  prefer to skip the decommission be;
6.  'show backends' will print the right tablet num of each backend;
7.  cloud tablet show PrimaryBackendId

### DIFF
--- a/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
@@ -235,7 +235,7 @@ public class Config extends ConfigBase {
             "连续写多批 BDBJE 日志后需要短暂停顿。这里最大的连写次数。",
             "After writting multiple batching BDBJE continuously, need a short rest. "
                     + "Indicates the writting count before a rest"})
-    public static long batch_edit_log_continuous_count_for_rest = 100000;
+    public static long batch_edit_log_continuous_count_for_rest = 1000;
 
     @ConfField(description = {"元数据同步的容忍延迟时间，单位为秒。如果元数据的延迟超过这个值，非主 FE 会停止提供服务",
             "The toleration delay time of meta data synchronization, in seconds. "
--- a/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
@@ -219,6 +219,14 @@ public class Config extends ConfigBase {
             "The log roll size of BDBJE. When the number of log entries exceeds this value, the log will be rolled"})
     public static int edit_log_roll_num = 50000;
 
+    @ConfField(mutable = true, masterOnly = true, description = {
+            "批量 BDBJE 日志包含的最大条目数", "The max number of log entries for batching BDBJE"})
+    public static int batch_edit_log_max_item_num = 100;
+
+    @ConfField(mutable = true, masterOnly = true, description = {
+            "批量 BDBJE 日志包含的最大长度", "The max size for batching BDBJE"})
+    public static long batch_edit_log_max_byte_size = 640 * 1024L;
+
     @ConfField(description = {"元数据同步的容忍延迟时间，单位为秒。如果元数据的延迟超过这个值，非主 FE 会停止提供服务",
             "The toleration delay time of meta data synchronization, in seconds. "
                     + "If the delay of meta data exceeds this value, non-master FE will stop offering service"})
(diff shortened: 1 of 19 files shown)


yujun777:
run buildall

yujun777:
run buildall

yujun777:
run buildall

yujun777:
run performance

yujun777:
run buildall

fe/fe-common/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -3045,8 +3053,14 @@ public static int metaServiceRpcRetryTimes() {
     @ConfField(mutable = true, description = {"存算分离模式下是否开启大事务提交，默认false"})
     public static boolean enable_cloud_txn_lazy_commit = false;
 
-    @ConfField(mutable = true, description = {"存算分离模式下，当tablet分布的be异常，是否立即映射tablet到新的be上，默认true"})
-    public static boolean enable_immediate_be_assign = true;
+    @ConfField(mutable = true, masterOnly = true,
+            description = {"存算分离模式下，当tablet分布的be异常，是否立即映射tablet到新的be上，默认false"})
+    public static boolean enable_immediate_be_assign = false;
```
gavinchou, line 3058: this seems a behavior change

gavinchou:
Add more background and  why we choose the solution and what is the trade off

fe/fe-common/src/main/java/org/apache/doris/common/Config.java
yujun777: > this seems a behavior change

默认要开启新算法。这个开关关了，那还是走旧算法

yujun777:
run buildall

yujun777:
run p0

yujun777:
run p0

fe/fe-core/src/main/java/org/apache/doris/cloud/catalog/CloudReplica.java
```diff
@@ -93,51 +97,114 @@ private boolean isColocated() {
...
-        String clusterName = ((CloudSystemInfoService) Env.getCurrentSystemInfo()).getClusterNameByClusterId(cluster);
+    public long getColocatedBeId(String cluster) throws ComputeGroupException {
+        List<Backend> bes = ((CloudSystemInfoService) Env.getCurrentSystemInfo())
+                .getBackendsByClusterId(cluster).stream()
+                .filter(be -> !be.isQueryDisabled()).collect(Collectors.toList());
         if (bes.isEmpty()) {
+            LOG.warn("failed to get available be, clusterId: {}", cluster);
             throw new ComputeGroupException(
-                String.format("There are no Backend nodes in the current compute group %s", clusterName),
+                String.format("There are no Backend nodes in the current compute group %s", cluster),
```
deardeng, line 107: 最好打clusterName,这个报错用户可见，clusterid 不太好看

yujun777:
run buildall

fe/fe-core/src/main/java/org/apache/doris/cloud/catalog/CloudReplica.java
yujun777: had fixed

yujun777:
run buildall

yujun777:
run cloud_p0

deardeng (approved):
LGTM

yujun777:
run buildall

yujun777:
run buildall

yujun777:
run buildall

yujun777:
run feut

yujun777:
run feut

gavinchou (approved):

PR #38889, opened 2024-08-05
[feature](statistics)Support get row count for JDBC external table.

Jibing-Li:
Support get row count for JDBC external table by executing select count(1) from table.
Return -1 when external table row count is unknown.

PR #40773, opened 2024-09-12
[fix](jdbc catalog) Fixed FE memory leak by enabling weak references in HikariCP

zy-kkk:
When FE collects statistics, its ThreadLocal will hold the ThreadLocal of HikariCP in JdbcClient, making it difficult for HikariCP to be gc, so this problem can be alleviated by enabling weak references of HikariCP.
Introduced from #38889


morningman (approved):
LGTM

BePPPower (approved):

PR #40777, opened 2024-09-12
[Enhancement] Support recover partitions to a new table

Vallishp:
## Proposed changes

Issue Number: close #xxx

Support recovering a partition to new table.
1) new table will be created based on partition schema.
2) this feature will be helpful for user to recover dat, if there is a schema change after the drop of partition.

Still coding in progress


fe/fe-core/src/main/java/org/apache/doris/catalog/CatalogRecycleBin.java
```diff
@@ -833,49 +885,54 @@ public synchronized void recoverPartition(long dbId, OlapTable table, String par
...
-        // check if schema change
-        Partition recoverPartition = recoverPartitionInfo.getPartition();
-        Set<Long> tableIndex = table.getIndexIdToMeta().keySet();
-        Set<Long> partitionIndex = recoverPartition.getMaterializedIndices(IndexExtState.ALL).stream()
-                .map(i -> i.getId()).collect(Collectors.toSet());
-        if (!tableIndex.equals(partitionIndex)) {
-            throw new DdlException("table's index not equal with partition's index. table's index=" + tableIndex
-                    + ", partition's index=" + partitionIndex);
-        }
+        if (newTableName != null) {
```
dataroaring, line 888: isNullOrEmpty is better.

regression-test/suites/catalog_recycle_bin_p0/recover_partition_with_newtable.groovy
```diff
@@ -0,0 +1,66 @@
...
+
+    // drop partition
+    sql """ ALTER TABLE ${table} DROP PARTITION p1; """
+    sql """ ALTER TABLE ${table} DROP PARTITION p2; """
+    sql """ ALTER TABLE ${table} DROP PARTITION p3; """ 
+
+
+    sql """ RECOVER PARTITION p1 from ${table} as ${table_new}; """ 
+    sql """ RECOVER PARTITION p2 from ${table} as ${table_new}; """
+    sql """ RECOVER PARTITION p3 from ${table} as ${table_new}; """
```
dataroaring, line 62: If recover would drop some data in table_new , we should add a force keyword to make users sure what they do. If other recover commands works without force, we should add it.

PR #40800, opened 2024-09-13
[improvement](mem)Dereference for executor

wuwenchi:
## Proposed changes

When a cmd is executed, we can cancel the reference to the `executor` to release the related resources. Otherwise, these resources can only be released when the next cmd is executed.


wuwenchi:
run buildall

morningman (approved):
LGTM.
This is not a serious problem because the executor will always be released when executing next command in this session.
But release it actively after on command finish may be better because if this is no more command in this session, the executor will not be released and may occupy memory.

@morrySnow @yiguolei PTAL

morrySnow (approved):

yiguolei (approved):

PR #40810, opened 2024-09-13
[fix](mtmv) Fix partition trace wrong when partition name is same from both side of join (#40485)

seawinde:
## Proposed changes

pr: https://github.com/apache/doris/pull/40485
commitId: 18a374f6


seawinde:
run buildall

PR #40816, opened 2024-09-13
[enhance](mtmv)tasks tvf not getLock

zddr:
`MTMV.getJob` need `readMvLock()`, we We have added a queryTask method in MTMV to avoid obtaining lock


zddr:
run feut

morrySnow (approved):

CalvinKirs (approved):

PR #40353, opened 2024-09-04
[feature](restore) support atomic restore

w41ter:
Previously, during a restore job, if the OLAP table already existed, it would be marked with a RESTORE status, preventing any read or write operations.

This PR introduces an atomic restore feature. If enabled, data will first be loaded into a temporary table during the restore, and then atomically replace the original table, ensuring that the target table's read and write operations remain unaffected during the restore process.

PR #40817, opened 2024-09-13
[improve](restore) Disable alter ops during atomic restore

w41ter:
The atomic restore is introduced in #40353. It replaces tables instead of updating in place, but the origin table might be altered during the restoring, which might cause the restore job to fail.  This PR adds a property in_atomic_restore for those tables and forbids any update to those tables involved in restoring.


w41ter:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #40841, opened 2024-09-14
[fix](covar) Fix covar nullable on branch-2.1

zhiqiang-hhhh:
covar should not be always nullable.

This fix on branch-2.1 makes covar same with master on FE.


yiguolei:
run buildall

zhiqiang-hhhh:
run feut

PR #40854, opened 2024-09-14
[fix](orc-reader) fix  orc_reader reading the timestamp type in the Shanghai timezone.

BePPPower:
Fix the issue with orc_reader reading the timestamp type in the Shanghai timezone.


BePPPower:
run buildall

PR #40861, opened 2024-09-14
[fix](test) fix unstable case of partitions

Vallishp:
## Proposed changes

Issue Number: close #xxx
   row count skipped in common validation
   if table report not done for the all tablet 
    row count will be -1 if done can get actual data.
So row count can cause timing issue and make testcase unstable.
 we already have a checkpoint for rowcount with waiting . in common validation we have skiped rowcount.


Vallishp:
run buildall

dataroaring (approved):
LGTM

morningman (approved):

Yukang-Lian (approved):

PR #40736, opened 2024-09-12
[opt](partial update) Allow to only specify key columns in partial update

bobhan1:
branch-2.1-pick: https://github.com/apache/doris/pull/40863
branch-2.0-pick: https://github.com/apache/doris/pull/40864

PR #40863, opened 2024-09-14
[branch-2.1] Picks "[opt](partial update) Allow to only specify key columns in partial update #40736"

bobhan1:
picks https://github.com/apache/doris/pull/40736


bobhan1:
run buildall

PR #40876, opened 2024-09-14
[fix](restore) Fix view name in atomic restore

w41ter:
Only olap table need to be replaced in atomic.


w41ter:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #40635, opened 2024-09-10
[Fix](load) fix wrong replica number when loading to table with no partition

zclllyybb:
## Proposed changes

Issue Number: close #xxx

now in auto-partition situation it will occur. may effect the txn commit check.

PR #40881, opened 2024-09-14
[branch-2.1](load) fix wrong replica number when loading to table with no partition (#40635)

zclllyybb:
pick https://github.com/apache/doris/pull/40635


zclllyybb:
run buildall

PR #40891, opened 2024-09-17
[improvement](jdbc catalog) Optimize JdbcCatalog case mapping stability

zy-kkk:
This PR makes the following changes to the uppercase and lowercase mapping of JdbcCatalog
1. The identifierMapping is managed by JdbcExternalCatalog instead of JdbcClient to better control its lifecycle
2. The identifierMapping no longer loads remoteName alone, but Catalog controls the loading uniformly
3. The identifierMapping will be loaded when each FE performs makeSureInitialized() to ensure that each FE has a mapping
4. The initialization of mapping will only be performed once in makeSureInitialized(), which means that even if you use metaCache, if your source data is updated when identifierMapping is enabled, you must refresh the catalog to query normally.
5. The identifierMapping is only responsible for the properties of the Catalog and is no longer affected by the fe config, simplifying the processing logic
6. If lower_case_mete_names is false and meta_names_mapping is empty in the catalog properties, the identifierMapping will no longer take effect, further enhancing the stability of the default settings
7. The JdbcClient is no longer closed during onRefreshCache, reducing the repeated creation of resources, improving reuse, and reducing the leakage of some global shared threads


morningman (approved):
LGTM

CalvinKirs (approved):

PR #40897, opened 2024-09-18
[Improvement](rpc) Use  1-phase commit if only 1 BE exists

Gabriel39:
## Proposed changes

2-phase execution is used to ensure both of RPC receiver and sender are ready before the query starts. But if only 1 BE is used, we do not use RPC to shuffle data .


Gabriel39:
run buildall

Gabriel39:
run buildall

Gabriel39:
run buildall

yiguolei (approved):

HappenLee:
here must keep fragment ready order， should think we really need to do the work

PR #40913, opened 2024-09-18
[bugfix](test)Set the enable_fallback_to_original_planner behavior

wuwenchi:
## Proposed changes

The default value of `enable_fallback_to_original_planner` in master is `false`, but the default value in 2.1 is `true`.
So it needs to be unified to `false` for testing.


wuwenchi:
run buildall

morningman (approved):
LGTM

zy-kkk (approved):

PR #40918, opened 2024-09-18
[Improment]Add CustomStdAllocator

wangbo:
## Proposed changes
Add a custom allocator to instead of std::allocator, used for std::vector, so that we can add std::vector's memory to thread context's memory tracker.

Currently just add the CustomStdAllocator and using it in some  aggregation operator, later all block used by aggregation operator should use the CustomStdAllocator.


wangbo:
run buildall

PR #40919, opened 2024-09-18
[fix](config) enlarge batch size when insert cache hotspot table

freemandealer:
This adjustment is based on the analysis that each entry in the params map is roughly 1KB, thus the old limit resulted in a 50KB footprint. With no concurrency issues, expanding the batch size to 1000 should enhance processing efficiency with a reasonable 1MB memory footprint.


freemandealer:
run buildall

gavinchou (approved):

dataroaring (approved):
LGTM

PR #40907, opened 2024-09-18
[Chore][inverted index] remove duplicate null bitmap reader in function array index

airborne12:
## Proposed changes
duplicate code in array_contains inverted index evaluate function.

PR #40920, opened 2024-09-18
[Chore][inverted index] remove duplicate null bitmap reader in function array index #40907

airborne12:
cherry pick from #40907


airborne12:
run buildall

PR #40922, opened 2024-09-18
[enhance](auth)Optimize the authentication logic of Ranger Doris

zddr:
- Set the authentication type for row policy and datamask as `select` to avoid traversing all permission items within the ranger
- The permission items on the ranger page are consistent with the grant syntax permission items on Doris
- Add global permissions on the ranger side, and the ranger page needs to be specified in the input box*
- No longer using cache to cache datamask and row policy, as changing the log level and specifying permission items has already made the speed fast enough
- When using a ranger, there is no need to create a role with the same name within Doris

ranger ui: https://github.com/morningman/ranger/pull/1

performance testing:

Each time the ranger authentication method is called, it takes less than 1ms,

If the user has global/db/table permissions, querying a large wide table takes approximately 1ms.

If the user only has partial column query permission, it will take approximately 80ms to query 50 columns


zddr:
run buildall

PR #40923, opened 2024-09-18
[branch-2.1][fix](jdbc catalog) Fixed FE memory leak by enabling weak references in HikariCP

zy-kkk:
pick  (#40773)

When FE collects statistics, its ThreadLocal will hold the ThreadLocal of HikariCP in JdbcClient, making it difficult for HikariCP to be gc, so this problem can be alleviated by enabling weak references of HikariCP.


zy-kkk:
run buildall

PR #40940, opened 2024-09-18
[Fix](load) Restrict the import of VARCHAR(0) data to avoid coredump

liaoxin01:
The VARCHAR(0) created has an actual length of 0, and BE didn't restrict it, leading to a BE core dump. #[38427](https://github.com/apache/doris/pull/38427) changed VARCHAR(0) to have a length of 65533. This PR restricts data import when the length is 0 to avoid a core dump.

PR #40974, opened 2024-09-19
[Fix](load) Restrict the import of VARCHAR(0) data to avoid coredump #40940

liaoxin01:
cherry pick from #40940


liaoxin01:
run buildall

PR #41002, opened 2024-09-19
[fix](regression-test) Revome "UNIQUE KEY (k1)" case for test_dynamic_partition_mod_distribution_key

Hastyshell:
## Proposed changes

1. Remove "UNIQUE KEY (k1)" case, because for unique table hash column must be key column, but for that historical bugs, this case will fail if adding k2 unique key.
2. Seperate a p0 suite from docker suite because docker suite will not be triggered in community doris p0 CI.


TangSiyang2001:
run buildall

dataroaring (approved):
LGTM

luwei16 (approved):

PR #41004, opened 2024-09-19
[Bug](branch-20) fix udf return array_string type has error in branch-20

zhangstar333:
## Proposed changes

when return type is array_string, the nested column string offset is should be cumulative, start with offset[row-1]
instead of equal to 0
this bug only in branch-2.0


zhangstar333:
run buildall

Mryange (approved):

PR #41020, opened 2024-09-19
[bugfix](paimon)Get the file format by file name

wuwenchi:
## Proposed changes

When `file.format` is not specified for the paimon table, the default file format will be used. In 0.8, the default format was `orc`, and in 0.9, it was changed to `parquet`.
Therefore, when there was no `file.format` and the paimon table was in version 0.9, Doris mistakenly used the orc reader to read the parquet file.

Now modify it to directly determine the file format based on the file name, and then use the corresponding reader.


wuwenchi:
run buildall

wuwenchi:
run buildall

morningman (approved):
LGTM

suxiaogang223 (approved):

PR #40998, opened 2024-09-19
[Fix](test) fix flake case in both test_insert_strict_fail_url and test_insert_partition_fail_url

liaoxin01:
The error message may vary due to variations in fuzzy execution instance number or batch size. So just check  if it contains an error log url.

PR #41024, opened 2024-09-19
[Fix](test) fix flake case in both test_insert_strict_fail_url and test_insert_partition_fail_url #40998

liaoxin01:
cherry pick from #40998


liaoxin01:
run buildall

PR #41034, opened 2024-09-20
[Fix](scanner) lazy update block mem usage when status is not ok #40953

airborne12:
cherry pick from #40953


airborne12:
run buildall

PR #40831, opened 2024-09-13
[enhancement](cloud) support BE http action: /api/file_cache?op=hash

yagagagaga:
## Proposed changes

Add a http action which can calculate its hash value based on the path you input. It's useful when you debug.

### API
```http
GET /api/file_cache
```

### request parameter

|param|type|desc|require|
|:---|:---|:---|:---|
|op|string|the value must be `hash`, other value you can refer to #37484 |yes|
|value|string|the input you want to calc hash|yes|

### response

if success
|param|type|desc|
|:---|:---|:---|
|hash|string|the hash value of your input|

if fail
|param|type|desc|
|:---|:---|:---|
|status|string|error status|
|msg|string|error message|

#### case 1

```bash
curl  '172.100.0.4:8040/api/file_cache?op=hash&value=0200000000000001bf42c14374fff491ffb7c89a1a65c5bb_0.dat'
```

return
```json
{"hash":"c6a599f453f67f0949f80ad9990fa3dd"}
```

#### case 2

```bash
curl  '172.100.0.4:8040/api/file_cache?op=hash'
```

return
```json
{
    "status": "INVALID_ARGUMENT",
    "msg": "missing parameter: value is required"
}
```

PR #37484, opened 2024-07-08
(cloud-merge)  Supports online capacity expansion and contraction

Lchangliang:
1. Support to reset capacity online via BE HTTP requet
2. Support to analyze cache conf when only get path_name `file_cache_path=[{"path":"xxxx"}]`, we can only specify the path for a file cache instance from now on, if we don't care about the detailed capacity limit

```
# clear file cache
curl ip:http_port/api/file_cache?op=clear
# reset capacity of a certain instance of file cache
curl ip:http_port/api/file_cache?op=reset&capacity=${capacity}&path=${cache_path} 

curl 127.0.0.1:8060/api/file_cache?op=clear
curl 127.0.0.1:8060/api/file_cache?op=reset&capacity=102400000&path=/user/file_cache
```

PR #41037, opened 2024-09-20
[enhancement](cloud) support BE http action: list_cache and clear

yagagagaga:
## Proposed changes

Add a http action which is useful when you debug.

### API
```http
GET /api/file_cache
```

#### request parameter1

|param|type|desc|require|
|:---|:---|:---|:---|
|op|string|the value must be `list_cache`, other value you can refer to #40831 #37484 |yes|
|value|string|the segment file name |yes|

#### request parameter2

|param|type|desc|require|
|:---|:---|:---|:---|
|op|string|the value must be `clear`, other value you can refer to #40831 #37484 |yes|
|value|string|the segment file name |yes|
|sync|bool|clean local cache in sync |no|

#### response1

if success
|param|type|desc|
|:---|:---|:---|
||array|return the segment file cache in local path|

if fail
|param|type|desc|
|:---|:---|:---|
||array|empty array|

#### response2

if success
|param|type|desc|
|:---|:---|:---|
|status|string||
|msg|string||

#### case 1

```bash
curl  '172.100.0.4:8040/api/file_cache?op=list_cache&value=0200000000000001bf42c14374fff491ffb7c89a1a65c5bb_0.dat'
```

return
```json
["/opt/doris/be/file_cache/c6a/c6a599f453f67f0949f80ad9990fa3dd/0"]
```

#### case 2

```bash
curl '127.0.0.1:8040/api/file_cache?op=clear&sync=true&value=0200000000000001284b68fea3dcfe8a83e65cd88426b081_0.dat'
```

return
```json
{
    "status": "OK",
    "msg": "OK"
}
```


be/src/io/cache/block_file_cache.cpp
```diff
@@ -421,7 +425,7 @@ std::string BlockFileCache::clear_file_cache_async() {
                 for (auto& [_, cell] : offset_to_cell) {
                     ++num_cells_all;
                     if (cell.releasable()) {
-                        cell.is_deleted = true;
+                        cell.is_deleted = true;cell.releasable();
```
gavinchou, line 428: typo?

be/src/io/cache/block_file_cache.cpp
```diff
@@ -1807,6 +1811,61 @@ void BlockFileCache::update_ttl_atime(const UInt128Wrapper& hash) {
     };
 }
 
+std::string BlockFileCache::clear_file_cache_by_key(bool sync, const UInt128Wrapper& hash) {
```
gavinchou, line 1814: check  BlockFileCache::remove for more details

be/src/http/action/file_cache_action.cpp
```diff
@@ -39,10 +47,13 @@ constexpr static std::string_view SYNC = "sync";
 constexpr static std::string_view PATH = "path";
 constexpr static std::string_view CLEAR = "clear";
 constexpr static std::string_view RESET = "reset";
+constexpr static std::string_view HASH = "hash";
+constexpr static std::string_view HASH_PATH = "hash_path";
```
freemandealer, line 51: The API name is implicit, maybe list_cache? @gavinchou what do you think?

be/src/io/cache/block_file_cache_factory.h
```diff
@@ -68,6 +70,14 @@ class FileCacheFactory {
      */
     std::string clear_file_caches(bool sync);
 
+    /**
+     * Clears data of the file you special.
```
freemandealer, line 74: typo, it should be: Clear cache data of a specific segment file
plus, you may want to add another function that uses std::string segment_path as the second argument.

regression-test/suites/cloud_p0/cache/http/test_calc_cache_file_hash.groovy
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+
+    httpAction.call() {
+        respCode, body -> {
+            assertEquals(respCode, 200)
+            def map = parseJson(body)
+            assertEquals(map.get("hash"), "c6a599f453f67f0949f80ad9990fa3dd")
+        }
+    }
```
freemandealer, line 56: please do add
1. some bad hash calculation cases to prove it can handle errors as expected
2. clear by segment path cases

be/src/io/cache/block_file_cache.cpp
```diff
@@ -1807,6 +1811,61 @@ void BlockFileCache::update_ttl_atime(const UInt128Wrapper& hash) {
...
+    LOG(INFO) << "start clear_file_cache_by_key sync=" << sync;
+
+    if (!_files.contains(hash)) {
+        LOG(INFO) << "no file need to be deleted, hash=" << hash.to_string();
+    }
+
+    std::vector<Path> files;
+    for (auto& [offset, cell] : _files[hash]) {
+        std::string file =
+                cell.file_block->_mgr->_storage->get_local_file(cell.file_block->_key);
```
freemandealer, line 1826: We will have a couple of new Storage types that do not use the file system as the cache storage medium, e.g. In-memory storage that only uses system RAM. So make this interface more general, not only for 'local file'. IMHO, there should be a storage->delete(key) interface instead to keep FSFileCacheStorage's 'file' unexposed to upper class BlockFileCache.

freemandealer (changes requested):

be/src/http/action/file_cache_action.cpp
yagagagaga: nice idea

yagagagaga:
run buidlall

yagagagaga:
run buildall

gavinchou:
list what has been cleared, a file list with size

freemandealer (approved):
LGTM

gavinchou (approved):

PR #41038, opened 2024-09-20
[Fix]unified cancel method in coordinator

wangbo:
## Proposed changes
Coordinator has two cancel method, with args and without args.
They should has the same behavior, so query queue cancel should be moved to cancel method with args.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/qe/Coordinator.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/qe/Coordinator.java
@@ -1280,18 +1280,11 @@ public class Coordinator implements CoordInterface {
         }
     }
 
-    // Cancel execution of query. This includes the execution of the local plan
-    // fragment,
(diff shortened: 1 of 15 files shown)


wangbo:
run buildall

fe/fe-core/src/main/java/org/apache/doris/qe/Coordinator.java
```diff
@@ -1285,13 +1285,13 @@ public Status shouldCancel(List<Backend> currentBackends) {
     // if any, as well as all plan fragments on remote nodes.
     public void cancel() {
         cancel(new Status(TStatusCode.CANCELLED, "query is cancelled by user"));
```
yiguolei, line 1287: remove  this method. It is duplicated and ambitious

wangbo:
run buildall

wangbo:
run buildall

yiguolei (approved):

zclllyybb (approved):

PR #40944, opened 2024-09-18
[fix](table_options) fix potential NPE when quering table_options sys table

morningman:
Fix unstable issue like:

```
2024-09-17 12:41:04,755 WARN (thrift-server-pool-4|384) [FrontendServiceImpl.fetchSchemaTableData():2292] Failed to fetchSchemaTableData
java.lang.NullPointerException: Cannot invoke "org.apache.doris.catalog.MaterializedIndexMeta.getSchema()" because the return value of "java.util.Map.get(Object)" is null
        at org.apache.doris.catalog.OlapTable.getSchemaByIndexId(OlapTable.java:954) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.catalog.OlapTable.getSchemaByIndexId(OlapTable.java:947) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.catalog.OlapTable.getBaseSchema(OlapTable.java:2097) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.catalog.OlapTable.getKeyColAsString(OlapTable.java:2167) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.tablefunction.MetadataGenerator.tableOptionsForInternalCatalog(MetadataGenerator.java:1099) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.tablefunction.MetadataGenerator.tableOptionsMetadataResult(MetadataGenerator.java:1189) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.tablefunction.MetadataGenerator.getSchemaTableData(MetadataGenerator.java:273) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.service.FrontendServiceImpl.fetchSchemaTableData(FrontendServiceImpl.java:2289) ~[doris-fe.jar:1.2-SNAPSHOT]
        at jdk.internal.reflect.GeneratedMethodAccessor50.invoke(Unknown Source) ~[?:?]
        at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
        at java.lang.reflect.Method.invoke(Method.java:569) ~[?:?]
        at org.apache.doris.service.FeServer.lambda$start$0(FeServer.java:60) ~[doris-fe.jar:1.2-SNAPSHOT]
[...]

PR #41040, opened 2024-09-20
[fix](table_options) fix potential NPE when quering table_options systable (#40944)

morningman:
bp #40944


morningman:
run buildall

PR #41053, opened 2024-09-20
[fix](load) disable num segments check in compatibility mode

kaijchen:
## Proposed changes

When using mixed version of BEs, sink v2 on old BE won't report num segments to load streams on the new BE.
This will cause false positive segment num mismatch.

This PR addressed this issue by disabling num segments check when any tablets_to_commit proto has not set num_segments field.


kaijchen:
run buildall

liaoxin01 (approved):
LGTM

dataroaring (approved):
LGTM

PR #40384, opened 2024-09-04
[chore](thrift compatibility) Fix thrift id on master is not compatible with branch-2.1

zhiqiang-hhhh:
Ref https://github.com/apache/doris/pull/39852

PR #41057, opened 2024-09-20
[chore](thrift compatibility) Fix thrift id on master is not compatible with branch-2.1 #40384

zhiqiang-hhhh:
cherry pick from #40384


zhiqiang-hhhh:
run buildall

PR #40990, opened 2024-09-19
[fix](profile) Fix npe when updating profile for internal load task.

zhiqiang-hhhh:
Fix problem 
```
2024-09-19 15:39:08,025 WARN (mtmv-task-execute-1-thread-1|131) [StmtExecutor.updateProfile():1225] failed to update profile, ignore th
is error
java.lang.NullPointerException: Cannot read field "originStmt" because "this.originStmt" is null
        at org.apache.doris.qe.StmtExecutor.getSummaryInfo(StmtExecutor.java:402) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.qe.StmtExecutor.updateProfile(StmtExecutor.java:1219) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.nereids.trees.plans.commands.insert.AbstractInsertExecutor.executeSingleInsert(AbstractInsertExecutor.java:
193) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.nereids.trees.plans.commands.insert.InsertIntoTableCommand.runInternal(InsertIntoTableCommand.java:250) ~[d
oris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.nereids.trees.plans.commands.insert.InsertIntoTableCommand.run(InsertIntoTableCommand.java:117) ~[doris-fe.
jar:1.2-SNAPSHOT]
        at org.apache.doris.nereids.trees.plans.commands.insert.InsertOverwriteTableCommand.runInsertCommand(InsertOverwriteTableComman
d.java:204) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.nereids.trees.plans.commands.insert.InsertOverwriteTableCommand.insertInto(InsertOverwriteTableCommand.java
:267) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.nereids.trees.plans.commands.insert.InsertOverwriteTableCommand.run(InsertOverwriteTableCommand.java:175) ~
[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.job.extensions.mtmv.MTMVTask.exec(MTMVTask.java:231) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.job.extensions.mtmv.MTMVTask.run(MTMVTask.java:200) ~[doris-fe.jar:1.2-SNAPSHOT]
[...]

PR #41077, opened 2024-09-20
[fix](profile) Fix npe when updating profile for internal load task. #40990

zhiqiang-hhhh:
cherry pick from #40990


zhiqiang-hhhh:
run buildall

PR #41086, opened 2024-09-21
[opt](cache elimation) using step decending method to elimiate cache

yiguolei:
## Proposed changes

The cache elimation logic should not use a Linear continuity double value, should use a step by step value.
Or the cache limitation value will be changed often.


yiguolei:
run buildall

xinyiZzz (approved):
LGTM

wangbo (approved):
LGTM

PR #41093, opened 2024-09-22
[bugfix](allocatebytes) ignore null ptr column in Block

yiguolei:
## Proposed changes
Sometimes if expr failed, then there will be a nullptr column left in the block.
We should ignore the nullptr column, or exception will be thrown and some profile will not computed correctly.


yiguolei:
run buildall

jacktengg (approved):

BiteTheDDDDt (approved):

PR #40908, opened 2024-09-18
[fix](Variant) check enable light_schema_change when create table with variant type

eldenmoon:

PR #41102, opened 2024-09-23
[pick](Variant) check enable light_schema_change when create table with variant type

eldenmoon:
#40908


eldenmoon:
run buildall

PR #41120, opened 2024-09-23
[fix](restore) Fix view signature

w41ter:
1. reset with dbName instead of dbFullName, to be compatible with doris 2.0(full name has prefix `default_cluster:`)
2. since the referred db name of the restored view has changed, the next time to restore the view, the signature is not matched, so the db names should be reset and compared with the signature again.


w41ter:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #41123, opened 2024-09-23
[fix](memory) Fix refresh cgroup meminfo

xinyiZzz:
cgroup memory usage should be refreshed frequently.

### DIFF
--- a/be/src/util/mem_info.cpp
+++ b/be/src/util/mem_info.cpp
@@ -189,40 +189,37 @@ void MemInfo::refresh_proc_meminfo() {
     }
 
     // refresh cgroup memory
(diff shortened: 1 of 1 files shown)


wm1581066 (changes requested):
Any test cases ?

xinyiZzz:
> Any test cases ?

@wm1581066 `cgroup_util_test.cpp` in BE UT will test the correctness of cgroup memory refresh.

but the bug fixed by this PR is refresh interval is too long, which makes it difficult to add tests.

yiguolei (approved):

wangbo (approved):
LGTM

PR #40613, opened 2024-09-10
[fix](bytebuffer) fix allocate size improper in append_and_flush

sollhui:
fix allocate size improper in append_and_flush introduced by https://github.com/apache/doris/pull/38960

PR #38960, opened 2024-08-06
[opt](bytebuffer) allocate ByteBuffer memory by Allocator and make it exception safe

sollhui:
At present, the memory allocation of `ByteBuffer` is done through `new char[capacity_]`. Now, it is uniformly allocated by `Allocator` for the following purposes:
1. Better memory statistics
2. Better support for memory limit check

PR #41133, opened 2024-09-23
[fix](bytebuffer) fix allocate size improper in append_and_flush (#40613)

sollhui:
pick (#40613)

fix allocate size improper in append_and_flush introduced by https://github.com/apache/doris/pull/38960


sollhui:
run buildall

PR #40818, opened 2024-09-13
[opt](routine load) optimize routine load timeout logic

sollhui:
If IO/CPU resources are tight, routine load task is likely to timeout. The current method is self-adaption backoff https://github.com/apache/doris/pull/32227, but the problem is it will do some ineffective work to match proper timeout. For one routine load task, a better way to handle task is finishing executing instead of retry when  resources are tight. Therefore, this pr increase timeout to make "task always finish even if it is slow when resources are tight".

PR #32227, opened 2024-03-14
[opt](routine-load) self-adaption backoff timeout

sollhui:
## Proposed changes

Routine load is a long-term import task, assuming that the user has set a relatively small consumption interval, as the BE load changes, there may be a timeout for a certain period of time, leading to the problem of fake death.

Adaptive adjustment of timeout time to address the issue of possible changes in execution time caused by be load.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #41137, opened 2024-09-23
[opt](routine load) optimize routine load timeout logic (#40818)

sollhui:
pick (#40818)

If IO/CPU resources are tight, routine load task is likely to timeout. The current method is self-adaption backoff
https://github.com/apache/doris/pull/32227, but the problem is it will do some ineffective work to match proper timeout. For one routine load task, a better way to handle task is finishing executing instead of retry when resources are tight. Therefore, this pr increase timeout to make "task always finish even if it is slow when resources are tight".


sollhui:
run buildall

sollhui:
run buildall

sollhui:
run buildall

sollhui:
run buildall

PR #41140, opened 2024-09-23
[fix](build) Change the return type of `get_size` in `tablet_meta` for MacOS compatibility.

zy-kkk:
```
auto size = tablet->tablet_meta()->delete_bitmap().get_size();
root.AddMember("size", size, root.GetAllocator());
```

The compiler complains because there is an ambiguity in the call when passing size to the constructor of rapidjson::GenericValue. This is because the type of size (which may be size_t or unsigned long) is 64 bits on ARM64-based Macs (such as the M1 chip), and RapidJSON provides multiple possible matching constructors.

### DIFF
--- a/be/src/olap/tablet_meta.cpp
+++ b/be/src/olap/tablet_meta.cpp
@@ -1092,9 +1092,9 @@ uint64_t DeleteBitmap::cardinality() const {
     return res;
 }
 
-size_t DeleteBitmap::get_size() const {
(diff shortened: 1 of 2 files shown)


zy-kkk:
run buildall

hust-hhb (approved):
LGTM

zy-kkk:
run buildall

morningman (approved):

PR #41144, opened 2024-09-23
[Bug](materialized-view) only throw 'MV same with base table is useless' when addMVClause's Ke…

BiteTheDDDDt:
…ysType same with olapTable

## Proposed changes
only throw 'MV same with base table is useless' when addMVClause's KeysType same with olapTable


BiteTheDDDDt:
run buildall

QuakeWang (approved):
+1

HappenLee (approved):
LGTM

PR #41146, opened 2024-09-23
[fix](restore) only drop registered table when cancel restore

w41ter:
In atomic restore, a table registered to db must have a name with the prefix, otherwise, it has not been registered and can be ignored when canceling.


w41ter:
run buildall

dataroaring (approved):
LGTM

bobhan1 (approved):

Yukang-Lian (approved):

PR #41148, opened 2024-09-23
[fix](external-p2) ignore external p2 cases

morningman:
The env for external p2 cases has been changed.
So currently, all these p2 cases can not be run.
I comment them out temporarily, and will add them back after env is ready.


morningman:
run buildall

CalvinKirs (approved):

zy-kkk (approved):

PR #41155, opened 2024-09-23
[Fix](Nereids) fix fold const be return type

LiBinfeng-01:
When expression of be fold const can not get datatype, it should not throw an exception directly
intro by: https://github.com/apache/doris/pull/39723


LiBinfeng-01:
run buildall

PR #41168, opened 2024-09-23
[fix](migrate) Fix iterator returning too early

w41ter:
introduced in #41083.

Because the iterator returns too early, only one piece of data is returned each time a binlog is obtained for a certain version range. When the system triggers compaction and generates a rowset that spans versions, the binlog files copied by the migrate task may be missed.


gavinchou (approved):

bobhan1 (approved):

dataroaring (approved):
LGTM

PR #41176, opened 2024-09-23
[fix](ES Catalog)Only like on keyword can be applied to wildcard query

qidaye:
## Proposed changes

We map `text` and `keyword` both to `string` type in Doris. When enable `like_push_down`, we translate like to wildcard query in ES, which will lead unexpected result in `text` field. We should stick to `keyword` with wildcard query.
1. Add `column2typeMap` in `EsTable` to save the mapping of column_name to ES field data type.
2. Add new class `EsSchemaCacheValue` to get schema and column to type map
3. Init `column2typeMap` when cache init and build query process of ES external table
4. Support LIKE functionCallExpr for Nereids planner.
5. Add end to end like predicate test cases and UTs

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/datasource/es/EsUtil.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/datasource/es/EsUtil.java
@@ -301,16 +301,17 @@ public class EsUtil {
                     type = ScalarType.createStringType();
                     break;
                 case "nested":
-                case "object":
                     type = Type.JSONB;
                     break;
                 default:
                     type = Type.UNSUPPORTED;
             }
         } else {
-            type = ScalarType.createStringType();
-            column.setComment("Elasticsearch no type");
-            column2typeMap.put(fieldName, "no_type");
+            // When there is no explicit type in mapping, it indicates this type is an `object` in Elasticsearch.
+            // reference: https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html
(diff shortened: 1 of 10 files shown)


qidaye:
run buildall

xiaokang (commented):
Do you have a end to end testcase for `text_field LIKE xxx` ?

fe/fe-core/src/main/java/org/apache/doris/datasource/es/EsUtil.java
```diff
@@ -307,6 +310,7 @@ private static Column parseEsField(String fieldName, ObjectNode fieldValue, List
         } else {
             type = Type.JSONB;
```
xiaokang, line 311: STRING is better

qidaye:
run buildall

xiaokang (dismissed):
LGTM

fe/fe-core/src/main/java/org/apache/doris/datasource/es/EsUtil.java
qidaye: When there is no explicit type in mapping, it indicates this type is an `object` in Elasticsearch.

qidaye:
run buildall

qidaye:
run buildall

qidaye:
run buildall

morningman (approved):
LGTM

airborne12 (approved):
LGTM

PR #41177, opened 2024-09-23
[enhancement](start) ignore lowercase names in conf if it is invalid

dataroaring:
If lowercase_table_names in fe.conf and meta are not same, we can only use value in meta. So exit is helpless.

## Proposed changes

Issue Number: close #xxx


dataroaring:
run buildall

gavinchou (approved):

PR #41180, opened 2024-09-23
[Chore](tools) Add convenient building of be benchmark

zclllyybb:
## Proposed changes

Issue Number: close #xxx

now we can easily use google benchmark to test complex logic with Doris internal types and functions.


zclllyybb:
run buildall

PR #41181, opened 2024-09-23
[opt](cloud) Retry when gRPC to MS throws Status.Code.UNAVAILABLE (shutdown)

gavinchou:
It will throw StatusRuntimeException when we try to issue an RPC with a shutdown channel:
```
thread1 get client(not expired) -> thread2 get client (expired) -> thread2 shutdown the client -> thread1 issues rpc (complains shutdown)
```


w41ter (approved):

dataroaring (approved):
LGTM

PR #35644, opened 2024-05-30
[fix](mtmv)Fix that the storage medium specified for the mtmv is SSD, but the partition storage medium for the mtmv is still HDD

zddr:
fix when create mtmv use property `'storage_medium' = 'SSD'`,after refresh mtmv, partition's `storage_medium` is HDD

Not only is there a problem with mtmv, but essentially, when creating partitions, there is no inheritance of the table

PR #41192, opened 2024-09-24
[revert](storage) storage medium of partition should not inherit from table

zddr:
revert #35644

table if not set storage medium, will set default by `Config.default_storage_medium`, 
at present, we cannot distinguish whether the storage medium of the table is set by the user.
If it is not set by the user, it represents priority use rather than mandatory use. If we directly inherit the storage medium of the table, it may cause the partition to be forced to use 'Config. default_storage_madium', resulting in the partition being unable to be created


zddr:
run buildall

deardeng (approved):
LGTM

morrySnow (approved):

PR #41203, opened 2024-09-24
[improvement](statistics)Reduce partition column sample BE memory consumption.

Jibing-Li:
For string type columns, use xxhash_64 to transfer column value to an integer, and then calculate the NDV based on the integer hash value. In this case, we can reduce the memory cost of sample analyze and improve the performance.
For example, l_comment column of TPCH 100G lineitem table. The memory cost to calculate its NDV is reduced to 8GB from 22GB


Jibing-Li:
run buildall

Jibing-Li:
run buildall

morrySnow (approved):

zfr9527 (approved):
LGTM

PR #41207, opened 2024-09-24
[enhance](auth)Optimize the authentication logic of Ranger Doris

zddr:
- Set the authentication type for row policy and datamask as `select` to avoid traversing all permission items within the ranger
- The permission items on the ranger page are consistent with the grant syntax permission items on Doris
- Add global permissions on the ranger side, and the ranger page needs to be specified in the input box*
- No longer using cache to cache datamask and row policy, as changing the log level and specifying permission items has already made the speed fast enough
- When using a ranger, there is no need to create a role with the same name within Doris
- If you have sub level permissions, you can see the current level when showing. For example, if you have query permissions for table1 under db1, db1 will be displayed when showing databases

ranger ui: morningman/ranger#1

performance testing:

Each time the ranger authentication method is called, it takes less than 1ms,

If the user has global/db/table permissions, querying a large wide table takes approximately 1ms.

If the user only has partial column query permission, it will take approximately 80ms to query 50 columns


zddr:
run buildall

zddr:
run buildall

fe/fe-core/src/main/java/org/apache/doris/catalog/authorizer/ranger/doris/RangerDorisAccessController.java
```diff
@@ -19,32 +19,32 @@
...
-import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Date;
-import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 public class RangerDorisAccessController extends RangerAccessController {
     private static final Logger LOG = LogManager.getLogger(RangerDorisAccessController.class);
+    // ranger must set name, we agreed that this name must be used
```
morningman, line 45: What does mean "we agreed"?

fe/fe-core/src/main/java/org/apache/doris/catalog/authorizer/ranger/hive/RangerHiveAccessControllerFactory.java
```diff
@@ -31,6 +31,6 @@ public String factoryIdentifier() {
 
     @Override
     public CatalogAccessController createAccessController(Map<String, String> prop) {
-        return new RangerCacheHiveAccessController(prop);
```
morningman, line 34: RangerCacheHiveAccessController  can be removed?

fe/fe-core/src/main/java/org/apache/doris/catalog/authorizer/ranger/RangerAccessController.java
```diff
@@ -92,6 +93,7 @@ public List<? extends RowFilterPolicy> evalRowFilterPolicies(UserIdentity curren
             String tbl) {
         RangerAccessResourceImpl resource = createResource(ctl, db, tbl);
         RangerAccessRequestImpl request = createRequest(currentUser);
+        request.setAccessType(DorisAccessType.SELECT.name());
```
morningman, line 96: Add comment to explain why adding this?

fe/fe-core/src/main/java/org/apache/doris/catalog/authorizer/ranger/doris/RangerDorisAccessController.java
zddr: ranger ui上必须填一个name，但是全局的权限其实不需要name，因此需要约定一个name，然后校验全局权限的时候拿这个name校验

zddr:
run buildall

morningman (approved):
LGTM

morrySnow (approved):

PR #41215, opened 2024-09-24
[enhancement](Optimizer) optimize delete statement can not do partition prune when column is not lower case 

924060929:
## Proposed changes

optimize delete statement can not do partition prune when column is not lower case 

for example, this sql will send finishRealtimePush rpc for all partitions, but we only want to process one partition
```sql
delete from tbl where Dt = '2024-01-01'
```


924060929:
run buildall

PR #41220, opened 2024-09-24
[improvement](statistics)Remove analyze retry logic.

Jibing-Li:
Remove analyze retry logic when task failed. Because usually retry would fail again and retry would bring a long time of sleep, which cause the analyze job running too slow.
Master pr: https://github.com/apache/doris/pull/33703


Jibing-Li:
run buildall

PR #41228, opened 2024-09-24
[fix](block_rule)Fix SQL block rule not working after FE restart

zddr:
- Fix SQL block rule not working after FE restart
- Add fe configuration 'sql-block_rule_ignore_ admin' to control root and admin to bypass SQL block rules, avoiding any commands being unable to execute due to user set regularization


zddr:
run buildall

morrySnow (approved):

starocean999 (approved):

PR #40064, opened 2024-08-28
[fix](transaction) commit txn check txn status

mymeiyi:
1. In `checkBeforeUnprotectedCommitTransaction` of `DatabaseTransactionMgr`, if the txn status is not `PREPARE`, should skip calculate partition version, but the return logic does not work.
2. the `preCommitTransaction2PC` in `DatabaseTransactionMgr` does not hold `DatabaseTransactionMgr` write lock
3. ~~the `abortTxnWhenCoordinateBeRestart` and `abortTxnWhenCoordinateBeDown` in `GlobalTransactionMgr` does not hold table write lock~~
4. the `public void commitTransaction(long dbId, List<Table> tableList, long transactionId, List<TabletCommitInfo> tabletCommitInfos)` is only used for UT, it requires the caller hold table write lock, and may wrongly use , mark it as Deprecated.

PR #41229, opened 2024-09-24
[fix](transaction) commit txn check txn status (#40064)

mymeiyi:
pick https://github.com/apache/doris/pull/40064


mymeiyi:
run buildall

PR #41239, opened 2024-09-24
[Fix](LZ4 compression) Fix wrong LZ4 compression max input size limit 

Yukang-Lian:
## Proposed changes

LZ4 compression max supported value is LZ4_MAX_INPUT_SIZE, which is 0x7E000000(2,113,929,216 bytes). Doris use wrong max size INT_MAX, which is 2,147,483,647, to check. If input data size is between this two size, then it can pass the check but LZ4 compression will fail.

This PR fix it.

### DIFF
--- a/be/src/util/block_compression.cpp
+++ b/be/src/util/block_compression.cpp
@@ -127,11 +127,13 @@ public:
     }
 
     Status compress(const Slice& input, faststring* output) override {
-        if (input.size > INT_MAX) {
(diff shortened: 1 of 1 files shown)


Yukang-Lian:
run buildall

dataroaring (approved):
LGTM

bobhan1 (approved):

Hastyshell (approved):
LGTM

PR #41263, opened 2024-09-25
[branch-2.1] do not fallback to origin planner for call stmt

morningman:
`CALL xxx` statement is not supported by old planner.
And if fallback, the error message will be confused.

Only for 2.1, master does not have this issue


morningman:
run buildall

PR #41018, opened 2024-09-19
[improve](load) change memtable memory limiter policy

kaijchen:
## Proposed changes

#40912 has changed meaning of `write_mem` in memtable memory limiter.
This PR is a followup to change the active memtable flush policy accordingly.

It also changed:
1. The amount of memtable writers selected in one flush.
2. The memtable writers are selected in orders of its size.

PR #41278, opened 2024-09-25
[fix](load) fix priority queue order in memtable memory limiter

kaijchen:
## Proposed changes

#41018 used priority queue when selecting memtables to flush.
But the compare function is wrong and causing the order to be the opposite.

> Note that the Compare parameter is defined such that it returns true if its first argument comes before its second argument in a weak ordering. But because the priority queue outputs largest elements first, the elements that "come before" are actually output last. That is, the front of the queue contains the "last" element according to the weak ordering imposed by Compare.

This PR fixes the compare function to make larger memtables come front.


kaijchen:
run buildall

xinyiZzz (approved):
LGTM

liaoxin01 (approved):
LGTM

PR #41282, opened 2024-09-25
[spill](fix) fix reserve memory if enable_memory_overcommit

jacktengg:
## Proposed changes

Issue Number: close #xxx


be/src/runtime/memory/mem_tracker_limiter.h
```diff
@@ -171,6 +171,18 @@ class MemTrackerLimiter final {
...
+        if (_query_statistics) {
+            _query_statistics->set_max_peak_memory_bytes(peak_consumption());
+            _query_statistics->set_current_used_memory_bytes(consumption());
+        }
+        _reserved_counter.add(bytes);
+    }
+
     bool try_reserve(int64_t bytes) {
         if (UNLIKELY(bytes == 0)) {
             return true;
```
yiguolei, line 188: call reserve here

PR #41293, opened 2024-09-25
[Bug](tablet-shuffle) tablet shuffle sink data should use input block rather than convert_block

zhangstar333:
## Proposed changes
 the convert_block is empty firstly, after execute exprs will be filled.
 so convert_block maybe different with block.
 so when send data we still need use block.


HappenLee (approved):
LGTM

Mryange (approved):

PR #41316, opened 2024-09-25
[fix](jni-reader) fix hudi jni scanner

suxiaogang223:
## Proposed changes

Move the hadoop-aliyun and aliyun-sdk-oss dependency from fe-core to fe-common

### DIFF
--- a/fe/fe-common/pom.xml
+++ b/fe/fe-common/pom.xml
@@ -133,6 +133,15 @@ under the License.
             <artifactId>log4j-core</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.aliyun.oss</groupId>
+            <artifactId>aliyun-sdk-oss</artifactId>
+            <version>3.15.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-aliyun</artifactId>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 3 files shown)


morningman (dismissed):
LGTM

fe/fe-common/pom.xml
```diff
@@ -132,6 +132,15 @@ under the License.
...
         </dependency>
+        <dependency>
+            <groupId>com.aliyun.oss</groupId>
+            <artifactId>aliyun-sdk-oss</artifactId>
+            <version>3.15.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-aliyun</artifactId>
+        </dependency>
```
CalvinKirs, line 143: If the user does not use Alicloud OSS, can we remove this dependency without causing any exceptions? I'm unsure if there will be any issues.

suxiaogang223: Do you mean to package it together again, but to ask the user to add this package manually?

suxiaogang223:
run buildall

fe/fe-common/pom.xml
suxiaogang223: I think it should not affect functions that do not depend on oss

morningman (approved):

CalvinKirs (approved):

PR #41319, opened 2024-09-26
[fix](snapshot) Link binlog files according to consistent rowsets

w41ter:
In previous implementations, binlog files would be linked according to making snapshot requests. However, sometimes not all requests can be executed directly. For example, when a certain version in missing_version does not exist, it will fallback to the full snapshot. Therefore, it is correct to link binlog files according to consistent rowsets.


w41ter:
run buildall

dataroaring (approved):
LGTM

gavinchou (approved):

PR #41326, opened 2024-09-26
[case](mtmv)fix alter job case

zddr:
should allow task failed, because the previous insert overwrite may not have been canceled


zddr:
run buildall

regression-test/framework/src/main/groovy/org/apache/doris/regression/suite/Suite.groovy
```diff
@@ -1330,8 +1330,8 @@ class Suite implements GroovyInterceptable {
         } while (timeoutTimestamp > System.currentTimeMillis() && (status == 'PENDING' || status == 'RUNNING'  || status == 'NULL' || status == 'CANCELED'))
         if (status != "SUCCESS") {
             logger.info("status is not success")
+            assertTrue(result.toString().contains("same table"))
```
seawinde, line 1333: what does `same table` mean? what‘s the relation with method name `AllowCancel`

zddr: The error message for completion is' Not allowed running Insert Overwrite on same table ', I have taken two words

seawinde (approved):

hello-stephen (approved):

PR #41339, opened 2024-09-26
[Fix](Serde-2.1) fix potential mem leak in array serde write_one_cell_to_json

eldenmoon:
placement new may lead to mem leak in Field without calling it's desctructor


eldenmoon:
run buildall

PR #41357, opened 2024-09-26
[chore](file-cache) Enable file cache for cloud mode by force

Hastyshell:
## Proposed changes

Temp local rowset writer for external sorting replies on file cache.

### DIFF
--- a/be/src/cloud/cloud_rowset_writer.cpp
+++ b/be/src/cloud/cloud_rowset_writer.cpp
@@ -17,6 +17,7 @@
 
 #include "cloud/cloud_rowset_writer.h"
 
+#include "common/status.h"
 #include "io/cache/block_file_cache_factory.h"
 #include "io/fs/file_system.h"
 #include "olap/rowset/rowset_factory.h"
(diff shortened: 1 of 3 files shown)


TangSiyang2001:
run buildall

dataroaring (dismissed):
LGTM

TangSiyang2001:
run buildall

be/src/cloud/cloud_rowset_writer.cpp
```diff
@@ -34,7 +35,14 @@ Status CloudRowsetWriter::init(const RowsetWriterContext& rowset_writer_context)
     if (_context.is_local_rowset()) {
         // In cloud mode, this branch implies it is an intermediate rowset for external merge sort,
         // we use `global_local_filesystem` to write data to `tmp_file_dir`(see `local_segment_path`).
-        _context.tablet_path = io::FileCacheFactory::instance()->get_cache_path();
+        if (auto maybe_cache_path = io::FileCacheFactory::instance()->get_cache_path();
```
freemandealer, line 38: so cache path will be used even if enable_file_cache = false ? This will increase user's cognitive load somehow. A new path should be used in my opinion.

FYI, with more Cache Storage being added to Doris, disk file system is not the only place for cached data, e.g. system RAM for Memory-based Cache Storage. In this case, cache path will have value but not a directory.

TangSiyang2001:
run buildall

be/src/cloud/cloud_rowset_writer.cpp
TangSiyang2001: We enable file cache by false after duscussion. As for the availability of cache path, currently every cache has a working disk path. If mem cache is introduced, we should modify the implementation of FileCacheFactory then.

freemandealer (approved):
LGTM

dataroaring (dismissed):
LGTM

TangSiyang2001:
run buildall

TangSiyang2001:
run buildall

TangSiyang2001:
run buildall

TangSiyang2001:
run external

gavinchou (approved):

PR #41367, opened 2024-09-26
[test](mtmv)Check the lock with mtmv life and table

zfr9527:
## Proposed changes

[test](mtmv)Check the lock with mtmv life and table


zfr9527:
The execution time of the case is very long. Merge it into the "selectdb cases" project. Run it by myself.

PR #41371, opened 2024-09-26
[opt](index compaction)Use RAM dir to create tmp index_writer

qidaye:

### DIFF
--- a/be/src/olap/rowset/segment_v2/inverted_index_compaction.cpp
+++ b/be/src/olap/rowset/segment_v2/inverted_index_compaction.cpp
@@ -40,8 +40,9 @@ Status compact_column(int64_t index_id, std::vector<lucene::store::Directory*>& 
                     "debug point: index compaction error");
         }
     })
-    lucene::store::Directory* dir =
-            DorisFSDirectoryFactory::getDirectory(io::global_local_filesystem(), tmp_path.data());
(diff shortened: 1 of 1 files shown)


qidaye:
run buildall

airborne12 (approved):
LGTM

be/src/olap/rowset/segment_v2/inverted_index_compaction.cpp
```diff
@@ -40,8 +40,9 @@ Status compact_column(int64_t index_id, std::vector<lucene::store::Directory*>&
                     "debug point: index compaction error");
         }
     })
-    lucene::store::Directory* dir =
-            DorisFSDirectoryFactory::getDirectory(io::global_local_filesystem(), tmp_path.data());
+    bool can_use_ram_dir = true;
```
xiaokang, line 43: use be config

qidaye: This is the input arg to the getDirectory method.
If only `can_use_ram_dir` and `config::inverted_index_ram_dir_enable` both are true, we can actually use the RAM direcotry.

qidaye:
run buildall

zzzxl1993 (approved):
LGTM

Issue #37064, opened 2024-07-01
How does DESC display comment information


xqg1316:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

### Description

For example, DESC table name; The displayed information does not contain field remarks. Comment is expected

### Use case

DESC table_name;

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #41377, opened 2024-09-26
show column comment when execute desc table

ihadoop:
## Proposed changes

Issue Number: How does DESC display comment information #37064 

add new title comment and add value from column's comment


ihadoop:
run buildall

PR #40565, opened 2024-09-09
[fix](audit_loader) fix that old external audit loader plugin not work because of incompatibility with new audit plugin

caiconghui:
1、rename org.apache.doris.plugin.audit.AuditLoaderPlugin to  org.apache.doris.plugin.audit.AuditLoader to avoid potential conflict with old external audit plugin
2、rename org.apache.doris.plugin.audit.AuditEvent to org.apache.doris.plugin.AuditEvent to keep eventFilter in AuditPlugin  be compatible with old external audit plugin

PR #41401, opened 2024-09-27
[fix](audit_loader) fix that old external audit loader plugin not work because of incompatibility with new audit plugin (#40565)

morningman:
bp #40565


morningman:
run buidall

PR #41281, opened 2024-09-25
[chore](configuration) Change be jvm default max heap size.

kaka11chen:
## Proposed changes
[chore] (configuration) Change be jvm default max heap size.

PR #41403, opened 2024-09-27
[chore](configuration) Change be jvm default max heap size.

kaka11chen:
## Proposed changes

backport #41281


kaka11chen:
run buildall

PR #38552, opened 2024-07-31
[bugfix](paimon)Use the fs cache, otherwise it will cause OOM

wuwenchi:
## Proposed changes

When cache is not used, a new `com.aliyun.oss.ClientConfiguration` is created each time, causing OOM

PR #41411, opened 2024-09-27
[bugfix](paimon)Use the fs cache, otherwise it will cause OOM for 3.0 (#38552)

wuwenchi:
## Proposed changes

bp: #38552


wuwenchi:
run buildall

PR #41208, opened 2024-09-24
[fix](cluster key) non vertical compaction support cluster key

mymeiyi:

PR #41413, opened 2024-09-27
[fix](cluster key) Cluster key skip non vertical compact

mymeiyi:
now the cluster key does not support non vertical compaction.
https://github.com/apache/doris/pull/41208 try to implement it, but there are some problems, so we skip non vertical compaction for cluster key

### DIFF
--- a/be/src/olap/compaction.cpp
+++ b/be/src/olap/compaction.cpp
@@ -186,7 +186,7 @@ Status Compaction::merge_input_rowsets() {
     Status res;
     {
         SCOPED_TIMER(_merge_rowsets_latency_timer);
-        if (_is_vertical) {
+        if (_is_vertical || !_tablet->tablet_schema()->cluster_key_idxes().empty()) {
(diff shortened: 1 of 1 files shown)


mymeiyi:
run p0

mymeiyi:
run buildall

mymeiyi:
run buildall

be/src/olap/compaction.cpp
```diff
@@ -186,7 +186,7 @@ Status Compaction::merge_input_rowsets() {
     Status res;
     {
         SCOPED_TIMER(_merge_rowsets_latency_timer);
-        if (_is_vertical) {
+        if (_is_vertical || !_tablet->tablet_schema()->cluster_key_idxes().empty()) {
```
dataroaring, line 189: if non-vertical is configed and table has cluster key, just skip compaction by return error.

mymeiyi:
run buildall

mymeiyi:
run performance

dataroaring (approved):
LGTM

bobhan1 (approved):

PR #41414, opened 2024-09-27
[bugfix](hive)Handle exceptions when submitting tasks fails to prevent dead loop for 3.0 (#40708)

wuwenchi:
bp: #40708


wuwenchi:
run buildall

PR #41428, opened 2024-09-27
[opt](row store) call `_serialize_block_to_row_column` when the table has row store column

bobhan1:

PR #41431, opened 2024-09-27
[branch-2.1] Picks "[opt](row store) call _serialize_block_to_row_column when the table has row store column #41428"

bobhan1:
picks https://github.com/apache/doris/pull/41428


bobhan1:
run buildall

PR #41438, opened 2024-09-28
[pick](Serde-2.1) fix variant serde may lost num_rows when subcolumns empty

eldenmoon:
serialization object with empty subcolumns may lost num_rows, so need to record num_rows and set back num_rows in serdes

backport #38413


eldenmoon:
run buildall

PR #41439, opened 2024-09-28
[opt](merge-on-write) avoid to check delete bitmap while lookup rowkey in some situation to reduce CPU cost (#41480)

zhannngchen:
## Proposed changes

Issue Number: close #xxx

cherry-pick #41480


zhannngchen:
run buildall

PR #41450, opened 2024-09-29
[fix](mtmv) Generate mtmv cache should use ADMIN user, and rewritten plan should not check privilege (#40374)

seawinde:
## Proposed changes

pr: https://github.com/apache/doris/pull/40374
commitId: f3d92e48


seawinde:
run buildall

PR #41262, opened 2024-09-25
[fix](cloud) cloud group commit should skip repaly wal if label is already used and the txn state is committed or visible

mymeiyi:

PR #41461, opened 2024-09-29
[fix](cloud) cloud group commit should skip repaly wal if label is already used and the txn state is committed or visible (#41262)

mymeiyi:
pick https://github.com/apache/doris/pull/41262


mymeiyi:
run buildall

dataroaring:
run buildall

PR #40791, opened 2024-09-13
[fix](regression-test) Add varchar column as distribution key

Hastyshell:
## Proposed changes

As title.

PR #41471, opened 2024-09-29
[fix](regression-test) Add varchar column as distribution key (#40791)

Hastyshell:
## Proposed changes

pick: #40791


TangSiyang2001:
run buildall

PR #40440, opened 2024-09-05
[bugfix](external)Memory leak problem for external table with insert operation

wuwenchi:
## Proposed changes

Get the corresponding transaction through `txnId`, and then update the file list returned from be according to the transaction.

PR #41479, opened 2024-09-29
[bugfix](external)Memory leak problem for external table with insert operation for 2.1 (#40440)

wuwenchi:
bp: #40440


wuwenchi:
run buildall

PR #41480, opened 2024-09-29
[opt](merge-on-write) avoid to check delete bitmap while lookup rowkey in some situation to reduce CPU cost

zhannngchen:
## Proposed changes

Issue Number: close #xxx

MoW performs a lookup on the primary key index for each key during the data loading process, and when a key is hit in the index, it continues to check if the key has been marked for deletion. Generally this check is not very costly.
However, in some scenarios, users perform high-frequency real-time update operations on a larger table, and most of the writes are updating existing data. In this scenario, the version of the table grows very fast, and the delete bitmap is also dense because duplicate keys are continuously being written.
In this scenario, this check is very costly
1. because it means calling the contains method of the roaring bitmap for almost every version of the rowset hit by an imported key to check if it has been marked for deletion
2. due to the high frequency of imports, there are typically thousands of versions that are not merged to base compaction.
3. because of the high duplication rate, every key is basically hit in the index
4. so this means that for almost every imported key, a loop needs to be called up to thousands of times to check if it has been marked for deletion
5. This overhead becomes very exaggerated when we are doing load jobs of about 100,000+ rows per second for a table

Here's a flame diagram for this scenario
image

For tables that don't use seq columns, and for non-column update imports, this check can be skipped. Even if a key is already marked for deletion, it's not a problem to mark it for deletion again as if it existed.

### DIFF
--- a/be/src/olap/base_tablet.cpp
+++ b/be/src/olap/base_tablet.cpp
@@ -441,7 +441,8 @@ Status BaseTablet::lookup_row_key(const Slice& encoded_key, TabletSchema* latest
                                   const std::vector<RowsetSharedPtr>& specified_rowsets,
                                   RowLocation* row_location, uint32_t version,
                                   std::vector<std::unique_ptr<SegmentCacheHandle>>& segment_caches,
-                                  RowsetSharedPtr* rowset, bool with_rowid) {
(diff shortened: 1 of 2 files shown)


zhannngchen:
run buildall

zhannngchen:
run buildall

PR #41482, opened 2024-09-29
[cloud](Variant) limit column size in MetaService and support handle rowset schema in backend side

eldenmoon:
1. in local mode we limit column size in commit_txn phase, in cloud mode we limit it in `write_schema_dict` which in commit_rowset phase
2. add config to decide whether fill schema in BE side or MS side

### DIFF
--- a/be/src/cloud/pb_convert.cpp
+++ b/be/src/cloud/pb_convert.cpp
@@ -17,6 +17,7 @@
 
 #include "cloud/pb_convert.h"
 
+#include <common/logging.h>
 #include <gen_cpp/olap_file.pb.h>
 
 #include <utility>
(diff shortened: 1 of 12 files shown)


eldenmoon:
run buildall

eldenmoon:
run buildall

cloud/src/meta-service/meta_service_schema.cpp
```diff
@@ -297,6 +297,15 @@ void write_schema_dict(MetaServiceCode& code, std::string& msg, const std::strin
                << ", reached the limited size threshold of SchemaDictKeyList "
                << config::schema_dict_kv_size_limit;
             msg = ss.str();
+            return;
+        }
+        // Limit the count of dict keys
+        if (dict.column_dict_size() > config::schema_dict_key_count_limit) {
```
gavinchou, line 303: also print `dict.column_dict_size()`; 
code = MetaServiceCode::KV_TXN_COMMIT_ERR;  use "invalid argument" instead

and what will happen to the existing dict, which size is larger than the config, say 2048, if we change the config to a smaller value, say, 100.

cloud/src/common/config.h
```diff
@@ -135,8 +135,9 @@ CONF_mBool(snapshot_get_tablet_stats, "true");
 // Value codec version
 CONF_mInt16(meta_schema_value_version, "1");
 
-// Limit kv size of Schema SchemaDictKeyList, default 10MB
-CONF_mInt32(schema_dict_kv_size_limit, "10485760");
+// Limit kv size of Schema SchemaDictKeyList, default 5MB
+CONF_mInt32(schema_dict_kv_size_limit, "5242880");
+CONF_mInt32(schema_dict_key_count_limit, "2048");
```
gavinchou, line 140: also fix BE, the BE side does not work for now.

add more comment what these configs are used for

gavinchou (changes requested):

eldenmoon:
run buildall

cloud/src/common/config.h
eldenmoon: done

cloud/src/meta-service/meta_service_schema.cpp
eldenmoon: done

eldenmoon:
run buildall

be/src/cloud/pb_convert.cpp
```diff
@@ -185,6 +221,9 @@ void cloud_rowset_meta_to_doris(RowsetMetaPB* out, const RowsetMetaCloudPB& in)
     if (in.has_tablet_schema()) {
         cloud_tablet_schema_to_doris(out->mutable_tablet_schema(), in.tablet_schema());
     }
+    if (dict != nullptr) {
```
xiaokang, line 224: If dict is nullptr due to some exception and schema is not filled by meta service, the state is not valid. So you need to check the state if dict is nullptr.

cloud/src/common/config.h
```diff
@@ -135,8 +135,10 @@ CONF_mBool(snapshot_get_tablet_stats, "true");
 // Value codec version
 CONF_mInt16(meta_schema_value_version, "1");
 
-// Limit kv size of Schema SchemaDictKeyList, default 10MB
-CONF_mInt32(schema_dict_kv_size_limit, "10485760");
+// Limit kv size of Schema SchemaDictKeyList, default 5MB
+CONF_mInt32(schema_dict_kv_size_limit, "5242880");
```
xiaokang, line 139: Why change limit smaller?

cloud/src/meta-service/meta_service_schema.cpp
```diff
@@ -331,6 +340,12 @@ void read_schema_from_dict(MetaServiceCode& code, std::string& msg, const std::s
     LOG(INFO) << "Get schema_dict, column size=" << dict.column_dict_size()
               << ", index size=" << dict.index_dict_size();
 
+    // Return dict, let backend to fill schema with dict info
+    if (schema_op == GetRowsetRequest::RETURN_DICT) {
+        response->mutable_schema_dict()->Swap(&dict);
```
xiaokang, line 345: If the dict is very large, the response is memory consuming. 
Will it still be better than before, if only small ratio of dict is usefull for the response rowsets?

be/src/olap/rowset/segment_v2/vertical_segment_writer.cpp
```diff
@@ -665,8 +665,10 @@ Status VerticalSegmentWriter::_append_block_with_variant_subcolumns(RowsInBlock&
         _opts.rowset_ctx->merged_tablet_schema = _opts.rowset_ctx->tablet_schema;
     }
     TabletSchemaSPtr update_schema;
+    bool check_schema_size = true;
```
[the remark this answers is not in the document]
eldenmoon: used to check column size when flush

cloud/src/common/config.h
eldenmoon: @gavinchou   suggested

cloud/src/meta-service/meta_service_schema.cpp
eldenmoon: for most of cases the large  ratio of dict is usefull , so the optimization is worthy of

gavinchou (approved):

gavinchou:
LGTM

be/src/cloud/pb_convert.cpp
eldenmoon: if exception happend, then the response will fail, and retry.

xiaokang (approved):
LGTM

PR #41488, opened 2024-09-30
[fix](move-memtable) detect repeated stream open due to retry

kaijchen:
## Proposed changes

There may be retries when `LoadStreamStub::open` fails, causing the expected total streams count less than the actual open count.

### DIFF
--- a/be/src/runtime/load_stream.h
+++ b/be/src/runtime/load_stream.h
@@ -126,6 +126,8 @@ public:
         }
     }
 
+    void increase_use_count() { _total_streams++; }
+
     void close(int64_t src_id, const std::vector<PTabletID>& tablets_to_commit,
                std::vector<int64_t>* success_tablet_ids, FailedTablets* failed_tablet_ids);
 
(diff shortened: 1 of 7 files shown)


be/src/runtime/load_stream.h
```diff
@@ -126,6 +126,8 @@ class LoadStream : public brpc::StreamInputHandler {
         }
     }
 
+    void increase_use_count() { _total_streams++; }
```
kaijchen, line 129:
```suggestion
    void increase_total_streams() { _total_streams++; }
```

be/src/runtime/load_stream_mgr.cpp
```diff
@@ -62,6 +63,12 @@ Status LoadStreamMgr::open_load_stream(const POpenLoadStreamRequest* request,
             _load_streams_map[load_id] = std::move(p);
         }
         load_stream->add_source(request->src_id());
+        if (_known_connections.contains(conn_id)) {
+            // detected retry open, increase total streams
+            load_stream->increase_use_count();
```
kaijchen, line 68:
```suggestion
            load_stream->increase_total_streams();
```

PR #41419, opened 2024-09-27
[fix](mtmv)Fix show mtmv time should wrapped in double quotation marks

zddr:
Fix show mtmv time should wrapped in double quotation marks

PR #41501, opened 2024-09-30
[fix](mtmv)Fix show mtmv time should wrapped in double quotation mark…

zddr:
pick: https://github.com/apache/doris/pull/41419


zddr:
run buildall

PR #41512, opened 2024-09-30
(3.0)[fix](jdbc catalog) Fix type recognition error when using `query` tvf to query doris

zy-kkk:
pick  (#40122)


morningman:
already picked

PR #41515, opened 2024-10-01
[bugfix](brpc) Should use status to generate protobuf message, because it will encoding Backend Info

yiguolei:
## Proposed changes

Issue Number: close #xxx

PR #41522, opened 2024-10-04
[bugfix](brpc) Should use status to generate protobuf message, because it will encoding Backend Info (#41515)

yiguolei:
pick  #41515
Should use status to generate protobuf message, because it will encoding Backend Info

## Proposed changes

Issue Number: close #xxx


yiguolei:
run buildall

PR #41555, opened 2024-10-08
[local exchange](fix) Fix correctness caused by local exchange

Gabriel39:
## Proposed changes

For plan `local exchange (hash shuffle) -> union -> colocated agg`, we must ensure local exchange use the same hash algorithm as MPP shuffling.

This problem is covered by our test cases but only can be reproduced on multiple BEs so no case is added in this PR.


Gabriel39:
run buildall

Gabriel39:
run buildall

HappenLee (approved):
LGTM

BiteTheDDDDt (approved):

PR #41567, opened 2024-10-08
[Enhancement](inverted index) return OK instead of not supported in expr evaluate_inverted_index

airborne12:
## Proposed changes

Fix annoying error stack info


be/src/vec/exprs/vexpr.h
```diff
@@ -120,7 +120,7 @@ class VExpr {
 
     // execute current expr with inverted index to filter block. Given a roaring bitmap of match rows
     virtual Status evaluate_inverted_index(VExprContext* context, uint32_t segment_num_rows) {
-        return Status::NotSupported("Not supported execute_with_inverted_index");
+        return Status::OK();
```
xiaokang, line 123: add comment for semantics of return OK()

qidaye (approved):

csun5285 (approved):
LGTM

PR #41100, opened 2024-09-23
[Feat](Authentication)Supporting Authentication Plugins and Kernel Isolation

CalvinKirs:
## Proposed changes 
- Use a Child-First ClassLoader to isolate plugins from the kernel, giving priority to plugin classes to prevent conflicts between the plugin and kernel classes. 
- Allow users to place plugins in a specified directory, such as auth-lib, to avoid classpath conflicts by default. 
- Support developers in debugging plugins by allowing them to bring in plugins via Maven, making the debugging process more convenient.

PR #41574, opened 2024-10-09
[chore](plugin)It's allowed for the plugin directory to be empty when loading plugins

CalvinKirs:
## Proposed changes
 in #41100 An exception will be thrown if the user creates the plugin directory but it is empty.
When loading plugins, it's acceptable for the plugin directory to be empty. Users might create the directory without placing any plugins inside, but since some plugins are still located on the classpath, this doesn’t cause any issues. If users specify a particular plugin but don’t place it in the directory, the business logic should handle that situation. The general-purpose class shouldn’t be concerned with this.


morningman (approved):
LGTM

CalvinKirs:
run buildall

zy-kkk (approved):

PR #41576, opened 2024-10-09
[fix](nereids)adjust agg function nullability in PhysicalHashAggregate

starocean999:
## Proposed changes

`select sum(distinct c1) from t;`
assume c1 is not null, because there is no group by, sum(distinct c1)'s nullable is alwasNullable in rewritten phase. But in implementation phase, we may create 3 phase agg with group by key c1. And the sum(distinct c1)'s nullability should be changed depending on if there is any group by expressions. This pr update the agg function's nullability accordingly


starocean999:
run buildall

starocean999:
run buildall

starocean999:
run buildall

morrySnow (approved):

zhiqiang-hhhh (approved):
LGTM

PR #41565, opened 2024-10-08
[Fix](inverted index) add DATEV2 and DATETIMEV2 for inverted index reader

airborne12:
## Proposed changes

This PR adds support for DATEV2 and DATETIMEV2 types in the inverted index reader, aligning with the types already allowed in the inverted index writer.

PR #41579, opened 2024-10-09
[Fix](inverted index) add DATEV2 and DATETIMEV2 for inverted index reader #41565

airborne12:
cherry pick from #41565


airborne12:
run buildall

PR #41580, opened 2024-10-09
[Bug](materialized-view) Fixed the problem of using drop table force and create mv stmt at the…

BiteTheDDDDt:
… same time causing fe startup failure.

## Proposed changes
Fixed the problem of using drop table force and create mv stmt at the same time causing fe startup failure.

Consider this scenario：
create mv -> drop table force -> create mv stmt cancelled
At this time there is a checkpoint like this：
[---------checkpoint---------]   [-------- editlog--------]
create mv -> drop table force -> create mv stmt cancelled

In this case, when read meta and parse create mv stmt, it will fail because there is no corresponding table, further causing fe startup failure.


BiteTheDDDDt:
run buildall

BiteTheDDDDt:
run buildall

zhangstar333 (approved):

Mryange (approved):

PR #41564, opened 2024-10-08
[test](inverted index) fix test case for no need read data

airborne12:
## Proposed changes

fix regression case for no need read data opt.

PR #41582, opened 2024-10-09
[test](inverted index) fix test case for no need read data #41564

airborne12:
cherry pick from #41564


airborne12:
run buildall

PR #41596, opened 2024-10-09
[fix](outfile) fix quotes with case 

amorynan:
## Proposed changes
miss outfile to be update
Issue Number: close #xxx


amorynan:
run buildall

PR #41608, opened 2024-10-09
[bugfix](segmentload) should remove segment from segment cache if load segment failed

yiguolei:
## Proposed changes

1. If the segment load failed, it is not recoverable, should remove it from segment cache and the query will create new segment again.
2. Could not depend on remove segment from cache during load index failed, because the call once may throw exception and the error handle logic is not reached.

### DIFF
--- a/be/src/olap/rowset/segment_v2/segment.cpp
+++ b/be/src/olap/rowset/segment_v2/segment.cpp
@@ -464,7 +464,7 @@ Status Segment::load_pk_index_and_bf() {
     return Status::OK();
 }
 
-Status Segment::load_index()() {
+Status Segment::load_index() {
     return _load_index_once.call([this] {
         if (_tablet_schema->keys_type() == UNIQUE_KEYS && _pk_index_meta != nullptr) {
             _pk_index_reader = std::make_unique<PrimaryKeyIndexReader>();
--- a/be/src/olap/rowset/segment_v2/segment.cpp
+++ b/be/src/olap/rowset/segment_v2/segment.cpp
@@ -450,17 +450,12 @@ Status Segment::_load_pk_bloom_filter() {
     DCHECK(_tablet_schema->keys_type() == UNIQUE_KEYS);
     DCHECK(_pk_index_meta != nullptr);
     DCHECK(_pk_index_reader != nullptr);
-    auto status = [this]() {
-        return _load_pk_bf_once.call([this] {
-            RETURN_IF_ERROR(_pk_index_reader->parse_bf(_file_reader, *_pk_index_meta));
-            // _meta_mem_usage += _pk_index_reader->get_bf_memory_size();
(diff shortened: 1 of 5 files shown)


yiguolei:
run buildall

yiguolei:
run buildall

be/src/olap/rowset/segment_v2/segment.cpp
```diff
@@ -910,8 +916,14 @@ Status Segment::new_inverted_index_iterator(const TabletColumn& tablet_column,
     ColumnReader* reader = _get_column_reader(tablet_column);
     if (reader != nullptr && index_meta) {
         if (_inverted_index_file_reader == nullptr) {
-            RETURN_IF_ERROR(
-                    _inverted_index_file_reader_open.call([&] { return _open_inverted_index(); }));
+            auto status = [this]() {
+                return _inverted_index_file_reader_open.call(
+                        [&] { return _open_inverted_index(); });
+            }();
```
gavinchou, line 922: why not just 
```suggestion
            auto status = _inverted_index_file_reader_open.call([&] { return _open_inverted_index(); });
```

be/src/olap/rowset/segment_v2/segment.cpp
```diff
@@ -554,11 +554,17 @@ vectorized::DataTypePtr Segment::get_data_type_of(const ColumnIdentifier& identi
 }
 
 Status Segment::_create_column_readers_once() {
-    return _create_column_readers_once_call.call([&] {
-        DCHECK(_footer_pb);
-        Defer defer([&]() { _footer_pb.reset(); });
-        return _create_column_readers(*_footer_pb);
-    });
+    auto status = [this]() {
```
gavinchou, line 557: why not just
```
auto status = _create_column_readers_once_call.call(...);
```

yiguolei:
run buildall

yiguolei:
run buildall

eldenmoon (approved):
LGTM

be/src/olap/segment_loader.cpp
```diff
@@ -59,8 +59,14 @@ Status SegmentLoader::load_segments(const BetaRowsetSharedPtr& rowset,
     for (int64_t i = 0; i < rowset->num_segments(); i++) {
         SegmentCache::CacheKey cache_key(rowset->rowset_id(), i);
         if (_segment_cache->lookup(cache_key, cache_handle)) {
-            continue;
+            // Has to check the segment status here, because the segment in cache may has something wrong during
```
xiaokang, line 62: Why not check status in lookup and erase not healthy segment?

yiguolei: because the segment cache code is too complicated, it may produce memory leak issue during lookup.

xiaokang (approved):
LGTM

PR #41613, opened 2024-10-09
[fix](Nerieds) column prune should retain at least one column for union all 

feiniaofeiafei:
introduce by #31811 and #39450
```sql
select count(1) from(select 3, 6 union all select 1, 3) t
```
wrong LogicalUnion plan:
```sql
LogicalUnion( qualifier=ALL, outputs=[3#6], regularChildrenOutputs=[], constantExprsList=[[], []], hasPushedFilter=false
```
this sql will report error in explain, because the logical union outputs has a slot, but the logical union has no child and has a empty constantExprList, which is wrong set in column prune.
this pr fixes it by consider when require columns is empty and keep the min slot and min slot corresponding const expressions.


feiniaofeiafei:
run buildall

morrySnow (approved):

924060929 (approved):

PR #41386, opened 2024-09-27
[Enhancement](maxCompute)Implement compatibility with existing maxcompute catalogs from previous versions.

hubgeter:
before pr #40225
## Proposed changes

Issue Number: close #xxx

PR #40888, opened 2024-09-17
[fix](maxCompute)Fixed the bug that extra nulls are read when reading maxcompute.

hubgeter:
before pr #40225 
## Proposed changes

Fixed a bug where when reading maxcompute, if there are null values ​​in a batch, null values ​​will always be read out.

PR #41614, opened 2024-10-09
[Enhancement](MaxCompute)Refactoring maxCompute catalog using Storage API.(#40225 , #40888 ,#41386 )

hubgeter:
bp #40225 , #40888 ,#41386

## Proposed changes
Among them, #40225 is the new api of mc,
#40888 is used to fix the bug when reading null between the new and old apis,
#41386 is used for compatibility between the new and old versions


hubgeter:
run buildall

PR #41633, opened 2024-10-09
[Enhancement](multi-catalog) Set hdfs native client logger to glog and redirect jvm stdout/stderr logger to jni.log.

kaka11chen:
Backport #39540.


kaka11chen:
run buildall

PR #41398, opened 2024-09-27
[opt](load) print more detailed log when stream load finished

liaoxin01:

PR #41639, opened 2024-10-10
[opt](load) print more detailed log when stream load finished #41398

liaoxin01:
cherry pick from #41398


liaoxin01:
run buildall

PR #39210, opened 2024-08-12
[improve](nestedtype) support schema change for nested type

amorynan:
## Proposed changes

Issue Number: close #xxx

PR #41642, opened 2024-10-10
[fix](nestedtype) support change order in table for nested type #39210

amorynan:
## Proposed changes
backport: https://github.com/apache/doris/pull/39210
after this pr: 
support schema change for nested type , but only now can only support change order in table 
Issue Number: close #xxx


amorynan:
run buildall

PR #40697, opened 2024-09-11
[feature](array_agg) support array_agg with param is array/map/struct

amorynan:
## Proposed changes
this pr we support array_agg function support param with array map struct type
Issue Number: close #xxx

PR #41651, opened 2024-10-10
[feature](array_agg) support array_agg with param is array/map/struct…

amorynan:
… (#40697)

this pr we support array_agg function support param with array map struct type

## Proposed changes

Issue Number: close #xxx


amorynan:
run buildall

PR #41653, opened 2024-10-10
[fix](statistics)Fix clear stale mv row count bug.

Jibing-Li:
When drop an mv, the row count for that mv should be removed in memory as well. Before we use OlapTable.getIndexIds to get all index ids, but this function always returns null. This pr is to fix this bug, use getIndexIdList instead.


Jibing-Li:
run buildall

Jibing-Li:
run performance

Jibing-Li:
run buildall

morrySnow (approved):

Jibing-Li:
run buildall

zfr9527 (approved):
LGTM

PR #41660, opened 2024-10-10
[bugfix](segmentload) should remove segment from segment cache if load segment failed (#41608)

yiguolei:
## Proposed changes
cherrypick #41608

1. If the segment load failed, it is not recoverable, should remove it from segment cache and the query will create new segment again.
2. Could not depend on remove segment from cache during load index failed, because the call once may throw exception and the error handle logic is not reached.

---------

## Proposed changes

Issue Number: close #xxx


yiguolei:
run buildall

PR #41692, opened 2024-10-11
[Fix](compile) Fix control reaches end of non-void function

bobhan1:


dataroaring (approved):
LGTM

dataroaring:
run buildall

BiteTheDDDDt (approved):

be/src/olap/partial_update_info.h
```diff
@@ -66,6 +66,7 @@ struct PartialUpdateInfo {
         case UniqueKeyUpdateModePB::UPDATE_FLEXIBLE_COLUMNS:
             return "flexible partial update";
         }
+        return "";
```
zhannngchen, line 69:
```suggestion
switch () {
case xx:
    // do sth.
case yy:
    // do sth.
default:
        return "";
}
```

zhannngchen: see Google C++ style : https://google.github.io/styleguide/cppguide.html#Switch_Statements

bobhan1:
run cloud_p0

PR #41685, opened 2024-10-11
[bugfix](becore) use after free problem when the segment is pop

yiguolei:
## Proposed changes

Issue Number: close #xxx
introduced by #41608

PR #41697, opened 2024-10-11
[bugfix](becore) use after free problem when the segment is pop (#41685)

yiguolei:
## Proposed changes

pick #41685
Issue Number: close #xxx
introduced by #41608

## Proposed changes

Issue Number: close #xxx


yiguolei:
skip buildall

PR #41595, opened 2024-10-09
[fix](array/map) fix resize impl in array/map

amorynan:
## Proposed changes
this resize function called shrink function which defined in IColumn, and function set_num_rows in  Block.cpp call shrink to make column rows changed, maybe less or more.
In this situation if we would make less rows which like cut the rows, so we should also make nested data in array/map resize to the less number of rows , otherwise we will meet exception :
```
mysql> /*set ShuffleSendBytes=0|ShuffleSendRows=0|SqlHash=c359573c8c7c33238ec34129e8ef0131|peakMemoryBytes=203840|SqlDigest=|cloudClusterName=UNKNOWN|TraceId=|WorkloadGroup=normal|FuzzyVariables=batch_size=4064,broker_load_batch_size=16352,disable_streaming_preaggregations=false,enable_distinct_streaming_aggregation=true,parallel_fragment_exec_instance_num=3,parallel_pipeline_task_num=5,profile_level=1,enable_pipeline_engine=true,enable_parallel_scan=true,parallel_scan_max_scanners_count=48,parallel_scan_min_rows_per_scanner=16384,parallel_prepare_threshold=13,enable_fold_constant_by_be=true,enable_rewrite_element_at_to_slot=true,runtime_filter_type=12,enable_parallel_result_sink=true,sort_phase_num=0,rewrite_or_to_in_predicate_threshold=100000,enable_function_pushdown=false,enable_common_expr_pushdown=true,enable_local_exchange=false,partitioned_hash_join_rows_threshold=1048576,partitioned_hash_agg_rows_threshold=1048576,partition_pruning_expand_threshold=10,enable_share_hash_table_for_broadcast_join=false,enable_two_phase_read_opt=true,enable_common_expr_pushdown_for_inverted_index=true,enable_delete_sub_predicate_v2=false,min_revocable_mem=33554432,fetch_remote_schema_timeout_seconds=120,max_fetch_remote_schema_tablet_count=512,enable_join_spill=false,enable_sort_spill=false,enable_agg_spill=false,enable_force_spill=false,data_queue_max_blocks=1,spill_streaming_agg_mem_limit=268435456,external_agg_partition_bits=5*/ select col4 from table_7052055 limit 10;
ERROR 1105 (HY000): errCode = 2, detailMessage = [INTERNAL_ERROR][E6] nested_column's size 660, is not consistent with offsets_column's 100

[...]

PR #41699, opened 2024-10-11
[fix](array/map) fix resize impl in array/map (#41595)

amorynan:
backport: https://github.com/apache/doris/pull/41595
this resize function called shrink function which defined in IColumn, and function set_num_rows in Block.cpp call shrink to make column rows changed, maybe less or more.
In this situation if we would make less rows which like cut the rows, so we should also make nested data in array/map resize to the less number of rows , otherwise we will meet exception

## Proposed changes

Issue Number: close #xxx


amorynan:
run buildall

regression-test/suites/datatype_p0/nested_types/query/test_nested_type_with_resize.groovy
```diff
@@ -0,0 +1,73 @@
...
... """ /*set ShuffleSendBytes=0|ShuffleSendRows=0|FuzzyVariables=batch_size=4064,broker_load_batch_size=16352,disable_streaming_preaggregations=false,enable_distinct_streaming_aggregation=true,parallel_fragment_exec_instance_num=3,parallel_pipeline_task_num=5,profile_level=1,enable_pipeline_engine=true,enable_parallel_scan=true,parallel_scan_max_scanners_count=48,parallel_scan_min_rows_per_scanner=16384,parallel_prepare_threshold=13,enable_fold_constant_by_be=true,enable_rewrite_element_at_to_slot=true,runtime_filter_type=12,enable_parallel_result_sink=true,sort_phase_num=0,rewrite_or_to_in_predicate_threshold=100000,enable_function_pushdown=false,enable_common_expr_pushdown=true,enable_local_exchange=false,partitioned_hash_join_rows_threshold=1048576,partitioned_hash_agg_rows_threshold=1048576,partition_pruning_expand_threshold=10,enable_share_hash_table_for_broadcast_join=false,enable_two_phase_read_opt=true,enable_delete_sub_predicate_v2=false,enable_sort_spill=false,enable_agg_spill=false,enable_force_spill=false,data_queue_max_blocks=1,spill_streaming_agg_mem_limit=268435456,external_agg_partition_bits=5*/ select col22 from test_array_resize order by col1,col2,col3 limit 10; """
```
superdiaodiao, line 69: nit:
What about moving the common setting part outside the queries?

amorynan: I think It doesn't matter for this case

amorynan:
run p0

amorynan:
run buildall

eldenmoon (approved):
LGTM

PR #41668, opened 2024-10-10
[Fix](bug) Is null predicate get error query result

HappenLee:
## Proposed changes

Fix Before Error Query Result：
```
create table t(a int, b int) duplicate key(`a`, `b`) DISTRIBUTED BY RANDOM BUCKETS AUTO PROPERTIES ( "replication_allocation" = "tag.location.default: 1");
insert into t values(-2147483648, -2147483648);
mysql> select * from t where a is null;
+-------------+-------------+
| a           | b           |
+-------------+-------------+
| -2147483648 | -2147483648 |
+-------------+-------------+
1 row in set (0.03 sec)
```

PR #41702, opened 2024-10-11
[Fix](bug) Is null predicate get error query result

HappenLee:
## Proposed changes

cherry-pick from #41668


HappenLee:
run buildall

PR #41707, opened 2024-10-11
[opt](index compaction)Use RAM dir to create tmp index_writer (#41371)

qidaye:
## Proposed changes

bp #41371


qidaye:
run buildall

xiaokang:
This optimization is not very necessary for branch-2.0.

PR #41563, opened 2024-10-08
[fix](jdbc catalog) Disable oracle scan null operator pushdown

zy-kkk:
Because Oracle versions below Oracle21 do not support null as an operator, and considering that most users' Oracle versions are below Oracle21, we disable Oracle's null operator pushdown by default.

PR #41712, opened 2024-10-11
[fix](jdbc catalog) Disable oracle scan null operator pushdown 

zy-kkk:
Because Oracle versions below Oracle21 do not support null as an operator, and considering that most users' Oracle versions are below Oracle21, we disable Oracle's null operator pushdown by default.
pick (#41563)


zy-kkk:
run buildall

morrySnow (approved):

PR #41739, opened 2024-10-11
[refactor]remove the redundant nullable value calculation in function call

Yoruet:
## Proposed changes

Issue Number: close #xxx


be/src/vec/functions/function_conv.cpp
```diff
@@ -72,7 +72,7 @@ class FunctionConv : public IFunction {
         return get_variadic_argument_types_impl().size();
     }
 
-    bool use_default_implementation_for_nulls() const override { return false; }
+    bool use_default_implementation_for_nulls() const override { return true; }
```
zclllyybb, line 75: just delete it

be/src/vec/functions/function_timestamp.cpp
```diff
@@ -115,12 +115,9 @@ struct StrToDate {
         ColumnPtr argument_columns[2] = {
                 col_const[0] ? static_cast<const ColumnConst&>(*col0).convert_to_full_column()
                              : col0};
-        check_set_nullable(argument_columns[0], null_map, col_const[0]);
```
zclllyybb, line 118: for this struct, u can see `static constexpr auto name = "str_to_date";`
so make sure the functions have enough testcases too

Yoruet:
run buildall

Yoruet:
run buildall

be/src/vec/functions/function_bitmap.cpp
```diff
@@ -506,10 +504,10 @@ class FunctionBitmapCount : public IFunction {
         auto& null_map = data_null_map->get_data();
 
         auto column = block.get_by_position(arguments[0]).column;
-        if (auto* nullable = check_and_get_column<const ColumnNullable>(*column)) {
-            VectorizedUtils::update_null_map(null_map, nullable->get_null_map_data());
-            column = nullable->get_nested_column_ptr();
-        }
+        // if (auto* nullable = check_and_get_column<const ColumnNullable>(*column)) {
```
zclllyybb, line 507: don't comment it. just delete if need.

env.sh
```diff
@@ -19,6 +19,7 @@
 
 # check DORIS_HOME
 export LC_ALL=C
+export CUSTOM_NPM_REGISTRY="https://mirrors.huaweicloud.com/repository/npm/"
```
zclllyybb, line 22: dont upload this change

be/src/vec/functions/function_bitmap.cpp
Yoruet: done

env.sh
Yoruet: done

Yoruet:
run buildall

Yoruet:
run buildall

Yoruet:
run buildall

Yoruet:
run buildall

regression-test/suites/nereids_syntax_p0/test_regexp_replace.groovy
```diff
@@ -0,0 +1,58 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+suite("test_regexp_replace") {
+    sql "SET enable_nereids_planner=true"
```
zclllyybb, line 19: u don't need to set these to session variables now. in master branch now we only have nereids planner and no other

Yoruet:
run buildall

zclllyybb (approved):
LGTM

be/src/vec/functions/array/function_array_contains_all.cpp
```diff
@@ -37,8 +37,6 @@ class FunctionArrayContainsAll : public IFunction {
 
     String get_name() const override { return name; }
 
-    bool use_default_implementation_for_nulls() const override { return true; }
```
zhiqiang-hhhh, line 40: do not modify code like this if this pr just wants to do refactor on `check_set_nullable`

be/src/vec/functions/array/function_array_enumerate.cpp
```diff
@@ -56,7 +56,6 @@ class FunctionArrayEnumerate : public IFunction {
     static constexpr auto name = "array_enumerate";
     static FunctionPtr create() { return std::make_shared<FunctionArrayEnumerate>(); }
     String get_name() const override { return name; }
-    bool use_default_implementation_for_nulls() const override { return true; }
```
zhiqiang-hhhh, line 59: do not modify code like this if this pr just wants to do refactor on `check_set_nullable`

be/src/vec/functions/array/function_array_range.cpp
```diff
@@ -146,14 +144,6 @@ struct RangeImplUtil {
         for (int i = 0; i < 3; ++i) {
             argument_columns[i] =
                     block.get_by_position(arguments[i]).column->convert_to_full_column_if_const();
-            if (auto* nullable = check_and_get_column<ColumnNullable>(*argument_columns[i])) {
-                // Danger: Here must dispose the null map data first! Because
-                // argument_columns[i]=nullable->get_nested_column_ptr(); will release the mem
-                // of column nullable mem of null map
-                VectorizedUtils::update_null_map(args_null_map->get_data(),
```
zhiqiang-hhhh, line 153: `arg_null_map` was updated in this line, and was used be code after this line, then you removed this logic. Why does this function still work?

be/src/vec/functions/function_bitmap.cpp
```diff
@@ -349,8 +349,6 @@ class FunctionBitmapAlwaysNull : public IFunction {
 
     size_t get_number_of_arguments() const override { return 1; }
 
-    bool use_default_implementation_for_nulls() const override { return true; }
```
zhiqiang-hhhh, line 352: keep this, just focus on `check_set_nullable`

be/src/vec/functions/function_bitmap.cpp
```diff
@@ -1168,10 +1164,6 @@ class FunctionBitmapSubs : public IFunction {
 
         default_preprocess_parameter_columns(argument_columns, col_const, {1, 2}, block, arguments);
 
-        for (int i = 0; i < 3; i++) {
```
zhiqiang-hhhh, line 1171: same problem with `RangeImplUtil`, it seems you remove some update logic to `res_null_map`, but actually `res_null_map` is still being used in the code.

zhiqiang-hhhh (changes requested):
It seems that correctness can not be guaranteed.

be/src/vec/functions/array/function_array_range.cpp
Yoruet: because use_default_implementation_for_nulls() is true, when meet null it will work on default implementation

zhiqiang-hhhh: you should remove var `args_null_map` if this var not working.

be/src/vec/functions/array/function_array_contains_all.cpp
zhiqiang-hhhh: > the default value of this function is true

you can make another pr to refactor this instead of doing it in this pr.

be/src/vec/functions/array/function_array_range.cpp
Yoruet: I thought that `args_null_map` is working, and it's just dont need to update_null_map.

Yoruet: because in the following code, `args_null_map` is used as a function parameter. If it is deleted, unexpected errors may occur.

be/src/vec/functions/array/function_array_contains_all.cpp
Yoruet: actually the other reviewer suggest me to remove this code so i doing it here

be/src/vec/functions/array/function_array_range.cpp
zclllyybb: `arg_null_map` will be updated in two ways:
1. outside this function, in the default implementation of nulls. so these code could be safely deleted
2. in the functions Impl, that is, it's own logic. so we need this var.

be/src/vec/functions/array/function_array_contains_all.cpp
zclllyybb: maybe you can just change this pr title to `remove the redundant nullable value calculation in function call` to fit it

Yoruet: ok

BiteTheDDDDt (approved):

zclllyybb (approved):

zhangstar333 (approved):

PR #41746, opened 2024-10-12
[Fix](inverted index) fix match null for inverted index

airborne12:
## Proposed changes

str match null return null , not true event if str is null


airborne12:
run buildall

airborne12:
run buildall

eldenmoon (approved):
LGTM

csun5285 (approved):
LGTM

PR #40431, opened 2024-09-05
[fix](cloud-mow) Fix the issue of compaction holding mow table lock for a long time

hust-hhb:
Now compaction use the same lock id (i.e. -1) when getting mow table lock from ms, if there are a large number of continuous compaction tasks to be executed, compaction will hold the lock for a long time, which will make loading task hang on waiting mow table lock when commiting txn, my test show it may lead to commit txn fail.

PR #41759, opened 2024-10-12
[fix](cloud-mow) FE should release mow lock when calculate delete bitmap catch exception

hust-hhb:
Now mow table lock is released on ms when doing commit txn, however if calculate delete bitmap failed before commiting txn, this lock will not release which will lead to another loading task hang on geting mow lock until this lock is expired on last txn.


hust-hhb:
run buildall

hust-hhb:
run buildall

hust-hhb:
run buildall

dataroaring (dismissed):
LGTM

cloud/src/meta-service/meta_service.cpp
```diff
@@ -1682,6 +1682,29 @@ static bool check_delete_bitmap_lock(MetaServiceCode& code, std::string& msg, st
     return true;
 }
 
+static bool check_and_remove_delete_bitmap_update_lock(
```
mymeiyi, line 1685: why add this method, not do it in `remove_delete_bitmap_update_lock` directly?

hust-hhb:
run buildall

cloud/src/meta-service/meta_service.cpp
hust-hhb: already change it

cloud/src/meta-service/meta_service.cpp
```diff
@@ -1640,8 +1640,8 @@ void MetaServiceImpl::get_tablet_stats(::google::protobuf::RpcController* contro
 
 static bool check_delete_bitmap_lock(MetaServiceCode& code, std::string& msg, std::stringstream& ss,
                                      std::unique_ptr<Transaction>& txn, std::string& instance_id,
-                                     int64_t table_id, int64_t lock_id, int64_t lock_initiator) {
-    std::string lock_key = meta_delete_bitmap_update_lock_key({instance_id, table_id, -1});
```
mymeiyi, line 1644: why change this, the key seems same for all method

hust-hhb: now just using table_id as look_key, but in this pr https://github.com/apache/doris/pull/40431, it will use both table_id and tablet_id as lock_key, so this change will be work in next new future(i.e. https://github.com/apache/doris/pull/40431)

hust-hhb:
run buildall

fe/fe-core/src/main/java/org/apache/doris/cloud/transaction/CloudGlobalTransactionMgr.java
```diff
@@ -637,7 +639,13 @@ private void calcDeleteBitmapForMow(long dbId, List<OlapTable> tableList, long t
         Map<Long, List<TCalcDeleteBitmapPartitionInfo>> backendToPartitionInfos = getCalcDeleteBitmapInfo(
                 backendToPartitionTablets, partitionVersions, baseCompactionCnts, cumulativeCompactionCnts,
                         cumulativePoints);
-        sendCalcDeleteBitmaptask(dbId, transactionId, backendToPartitionInfos);
+        try {
+            sendCalcDeleteBitmaptask(dbId, transactionId, backendToPartitionInfos);
+        } catch (UserException e) {
+            LOG.warn("failed to sendCalcDeleteBitmaptask for txn=" + transactionId + ",exception=" + e);
```
zhannngchen, line 645: use `e.getMessage()`

hust-hhb:
run buildall

hust-hhb:
run buildall

fe/fe-core/src/main/java/org/apache/doris/cloud/transaction/CloudGlobalTransactionMgr.java
```diff
@@ -858,6 +866,33 @@ private void getDeleteBitmapUpdateLock(Map<Long, Set<Long>> tableToParttions, lo
...
+                response = MetaServiceProxy.getInstance().removeDeleteBitmapUpdateLock(request);
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("remove delete bitmap lock, transactionId={}, Request: {}, Response: {}",
+                            transactionId, request, response);
+                }
+            } catch (Exception e) {
+                LOG.warn("ignore get delete bitmap lock exception, transactionId={}, exception={}",
+                        transactionId, e);
+            }
+            Preconditions.checkNotNull(response);
```
mymeiyi, line 887: response is null if exception is thrown in line 883, the check must fail, so line 884 ignore the exception does not work? if you want to ignore the exception, maybe you should move these operations into the `try` logic?

hust-hhb:
run buildall

zhannngchen (approved):
LGTM

dataroaring (approved):
LGTM

PR #41762, opened 2024-10-12
[fix](mtmv)fix in the scenario of recreating a table, the materialized view may assume that the data has not changed

zddr:
snapshot of table,only compare tableName and version,when recerate table,and if the version remains unchanged 。

MV will assume that the data has not changed


zddr:
run buildall

zddr:
run cloud_p0

morrySnow (approved):

seawinde (approved):

PR #41672, opened 2024-10-10
[fix](delete) Fix potential delete job stuck util timeout if exception happend in FE DeleteJob execution

Hastyshell:
## Proposed changes

Fail task should also count down for the count down latch to prevent job stuck.

PR #41763, opened 2024-10-12
[fix](delete) Fix potential delete job stuck util timeout if exception happend in FE DeleteJob execution (#41672)

Hastyshell:
## Proposed changes

pick: #41672

Fail task should also count down for the count down latch to prevent job stuck.


TangSiyang2001:
run buildall

PR #41535, opened 2024-10-08
[enhancement](err-msg) Output column info when size invalid in block data convertor

Hastyshell:
## Proposed changes

As title.

PR #41764, opened 2024-10-12
[enhancement](err-msg) Output column info when size invalid in block data convertor (#41535)

Hastyshell:
## Proposed changes

pick: #41535

As title.


TangSiyang2001:
run buildall

PR #41735, opened 2024-10-11
[improvement](jdbc catalog) Print more errors when Jdbc Catalog fails to obtain a connection on FE

zy-kkk:
We need to print more detailed error information when Jdbc Catalog obtains a connection to facilitate troubleshooting

PR #41769, opened 2024-10-12
[branch-2.1][improvement](jdbc catalog) Print more errors when Jdbc Catalog fails to obtain a connection on FE

zy-kkk:
bp #41735


zy-kkk:
run buildall

PR #41799, opened 2024-10-14
[enhance](insertoverwrite)insert overwrite not fallback

zddr:
- Insert overwrite on NEREIDS can automatically clean up garbage temporary partitions after restart, which is not available on old optimizers
- When insert fails, no longer throw nereids exceptions


zddr:
run buildall

PR #39685, opened 2024-08-21
[improvement](statistics)Drop table stats in unprotectDropTable so that dropDB could call it as well.

Jibing-Li:
Drop table stats in unprotectDropTable so that dropDB could call it as well.

PR #41801, opened 2024-10-14
[improvement](statistics)Drop table stats in unprotectDropTable so that dropDB could call it as well. (#39685)

Jibing-Li:
backport: https://github.com/apache/doris/pull/39685


Jibing-Li:
run buildall

PR #40512, opened 2024-09-08
[Enhancement] Insert overwrite case need to keep old partition data in recycle-bin

Vallishp:
## Proposed changes

Issue Number: close #xxx

Insert overwrite command will trigger drop of partition and create partition and add the rows to table.
In this case need to move the old data to catalog recycle bin so that user get chance to recover the data if needed.

PR #41813, opened 2024-10-14
[fix](insertoverwrite)drop partition should erase partition

zddr:
cause by: https://github.com/apache/doris/pull/40512


zddr:
run buildall

PR #41675, opened 2024-10-10
[fix](auto bucket) fix auto buckets calc using the first k partition

yujun777:
If the first  k (at most 7) partition data size is ascending,  the result will be partion_size[k-1] + ema(first k partitons delta).

This is a bug,  should use the last k partitions,  but not the first k partitions to calculate.

PR #41820, opened 2024-10-14
[fix](auto bucket) fix auto buckets calc using the first k partition #41675

yujun777:
cherry pick from #41675


yujun777:
run buildall

PR #33496, opened 2024-04-10
[Feature](top-n) support topn filter on vscan node

BiteTheDDDDt:
## Proposed changes
support topn filter on vscan node

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #41823, opened 2024-10-14
[Feature](top-n) support topn filter on vscan node #33496

BiteTheDDDDt:
## Proposed changes
support topn filter on vscan node
pick from #33496


BiteTheDDDDt:
run buildall

PR #41691, opened 2024-10-11
[enhance](mtmv)Provide a method to determine whether incremental updates of multiple tables are allowed

zddr:
Currently, only range partitioning is supported, and it must be a dynamic partition or auto partition.

At the same time, it is required that the partitioning rules of different tables be consistent

Think about the list partition carefully before making any decisions

PR #41843, opened 2024-10-15
[enhance](mtmv)Provide a method to determine whether incremental updates of multiple tables are allowed (#41691)

zddr:
pick: https://github.com/apache/doris/pull/41691


zddr:
run buildall

PR #41850, opened 2024-10-15
[fix](catalog) should return error if try using a unknown database (#40479)

morningman:
bp #40479


morningman:
run buildall

morningman:
run buildall

PR #41609, opened 2024-10-09
[fix](heartbeat) fill default value for required field in TFrontendPingFrontendResult

morningman:
Otherwise, if ping failed, some of required fields are not set, causing some unexpected error log like:
img_v3_02ff_a120257d-1614-4e33-9b4a-1cbe33eb84dg

PR #41853, opened 2024-10-15
[fix](heartbeat) fill default value for required field in TFrontendPingFrontendResult (#41609)

morningman:
bp #41609


morningman:
run buildall

PR #41331, opened 2024-09-26
[feat](Nereids) support set var in hint when parse sql

LiBinfeng-01:
set var hint need to be enable to use before analyze, so it need to be set when parsing sql
now it would set twice when parse and begin of analyze

PR #41863, opened 2024-10-15
[feat](Nereids) support set var in hint when parse sql (#41331)

LiBinfeng-01:
cherry-pick: #41331 
set var hint need to be enable to use before analyze, so it need to be set when parsing sql
now it would set twice when parse and begin of analyze


LiBinfeng-01:
run buildall

PR #40912, opened 2024-09-18
[enhancement](memtable) make memtable memusage more accurate

yiguolei:
## Proposed changes

1. Add memtype to memtable, and save a weak ptr vector in memtable writer, so that we could get different memory usage by traverse the vector.
2. Using scoped memory usage to compute the mem usage of a memtable.
3. CHECK if the tracker is 0 when the memtable flush success.

PR #41866, opened 2024-10-15
[enhancement](memtable) make memtable memusage more accurate (#40912)

yiguolei:
#40912 
1. Add memtype to memtable, and save a weak ptr vector in memtable writer, so that we could get different memory usage by traverse the vector.
2. Using scoped memory usage to compute the mem usage of a memtable.
3. CHECK if the tracker is 0 when the memtable flush success.

---------

## Proposed changes

Issue Number: close #xxx


yiguolei:
run buildall

PR #41886, opened 2024-10-15
[fix](counters) fix MemoryUsage and PeakMemoryUsage counters of some operators (#41602)

jacktengg:
BP #41602


jacktengg:
run buildall

PR #41824, opened 2024-10-14
[Bug](predicate) Fixed the problem that the number of rows in inlist predicate was cou…

BiteTheDDDDt:
…nted twice

## Proposed changes
Fixed the problem that the number of rows in inlist predicate was counted twice

PR #41910, opened 2024-10-15
[Bug](predicate) Fixed the problem that the number of rows in inlist #41824

BiteTheDDDDt:
pick from #41824


yiguolei:
run buildall

PR #41915, opened 2024-10-15
[fix](index compaction) fix fd leak and mem leak while index compaction

cambyzju:
if `DORIS_TRY(inverted_index_file_readers[src_segment_id]->open(index_meta))` failed and return,
then `lucene::store::Directory*` stay in `src_index_dirs` will got fd leak and memory leak.

### DIFF
--- a/be/src/olap/rowset/segment_v2/inverted_index_compaction.cpp
+++ b/be/src/olap/rowset/segment_v2/inverted_index_compaction.cpp
@@ -24,7 +24,8 @@
 #include "util/debug_points.h"
 
 namespace doris::segment_v2 {
-Status compact_column(int64_t index_id, std::vector<lucene::store::Directory*>& src_index_dirs,
+Status compact_column(int64_t index_id,
+                      std::vector<std::unique_ptr<DorisCompoundReader>>& src_index_dirs,
                       std::vector<lucene::store::Directory*>& dest_index_dirs,
                       std::string_view tmp_path,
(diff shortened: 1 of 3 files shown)


cambyzju:
run buildall

be/src/olap/rowset/segment_v2/inverted_index_compaction.cpp
```diff
@@ -57,12 +62,6 @@ Status compact_column(int64_t index_id, std::vector<lucene::store::Directory*>&
     // when index_writer is destroyed, if closeDir is set, dir will be close
     // _CLDECDELETE(dir) will try to ref_cnt--, when it decreases to 1, dir will be destroyed.
     _CLDECDELETE(dir)
-    for (auto* d : src_index_dirs) {
-        if (d != nullptr) {
-            d->close();
```
qidaye, line 62: `DorisCompoundReader` should explicitly call the `close()` to release resource.

cambyzju: DorisCompoundReader::~DorisCompoundReader() {
    if (!_closed) {
        try {
            close();
        } catch (CLuceneError& err) {
            LOG(ERROR) << "DorisCompoundReader finalize error:" << err.what();
        }
    }
    _CLDELETE(entries)
}

qidaye (approved):
LGTM

lide-reed (approved):
LGTM

PR #41918, opened 2024-10-15
[imporve][load] support set stream-load data size limit by http header

jqcc:
## Proposed changes

Doris's stream load limits the amount of data that can be imported at one time. If the amount of data exceeds the limit, it will fail. If you want to increase the limit, you need to manually adjust the BE parameters(streaming_load_max_mb).

This adjustment adds a check rule for the amount of data during import. If the amount of imported data is set in the HTTP header, the size limit specified in HTTP can be dynamically loaded, avoiding the need to uniformly adjust the limit of the BE layer.

just like:
```
curl --location-trusted -u <doris_user>:<doris_password> \
    -H "data_size_limit:10240"
    -H "column_separator:," \
    -H "columns:user_id,name,age" \
    -T streamload_example.csv \
    -XPUT http://<fe_ip>:<fe_http_port>/api/testdb/test_streamload/_stream_load
```


jqcc:
run buildall

PR #41920, opened 2024-10-16
[fix](exchange) deadlock issue in exchange opeartor

mrhhsg:
## Proposed changes

It does not mean all channels are EOF when `_busy_channels` is equal to zero


mrhhsg:
run buildall

PR #41934, opened 2024-10-16
[fix](Nereids) fix leading tag can not work when leading is used (#40958)

LiBinfeng-01:
cherry-pick: (#40958)

isLeading in cascade context means do we have leading in this scope isLeadingDisableJoinReorder means does leading used we need to copy it after analyze outerside of cte

## Proposed changes

Issue Number: close #xxx


LiBinfeng-01:
run buildall

LiBinfeng-01:
run p0

PR #40404, opened 2024-09-05
[opt](udf-cache) cache more udf classes

morningman:
In this PR #34980, we cache the class load of java udf.
But when calling init() method, it is still very time consuming 
when calling `Class.forName()`.

This PR add a new class `UdfClassCache` to cache more things
so that the `init()` method will not cost too much time.

PR #41937, opened 2024-10-16
[improve](udaf)support class cache for java-udaf

zhangstar333:
## Proposed changes
in some user case, 

- the udf-jars package have some resource file, it's maybe about hundreds MBs,
   so if every instances load the jar, it's easy to cause the BE JVM OOM.

- or in some udf-jars  have some init operator cause many times, so wants all instance could only init it's once.

follow up https://github.com/apache/doris/pull/40404
support class cache for java-udaf


zhangstar333:
run buildall

zhangstar333:
run buildall

PR #41902, opened 2024-10-15
[fix](analytic) Fix data distribution after analytic operator

Gabriel39:
## Proposed changes

Fix data distribution after analytic operator

PR #41949, opened 2024-10-16
[fix](analytic) Fix data distribution after analytic operator (#41902)

Gabriel39:
Fix data distribution after analytic operator

## Proposed changes

pick #41902


Gabriel39:
run buildall

PR #41957, opened 2024-10-16
[opt](blacklist) Backend should not be added to blacklist easily #41170

zhiqiang-hhhh:
cherry pick from #41170


zhiqiang-hhhh:
run buildall

PR #41659, opened 2024-10-10
[Enhancement](ExternalTable)Optimize the performance of getCachedRowCount when reading ExternalTable

hubgeter:
## Proposed changes
Because ExternalTable will initialize the previously uninitialized table when `getCachedRowCount()`, which is unnecessary. So for the uninitialized table, we directly return -1.
This will increase the speed of our query `information_schema.tables`.

PR #41962, opened 2024-10-16
[Enhancement](ExternalTable)Optimize the performance of getCachedRowCount when reading ExternalTable (#41659)

hubgeter:
bp #41659 

## Proposed changes
Because ExternalTable will initialize the previously uninitialized table when `getCachedRowCount()`, which is unnecessary. So for the uninitialized table, we directly return -1.
This will increase the speed of our query `information_schema.tables`.


hubgeter:
run buildall

PR #41969, opened 2024-10-16
[fix](memory) Process Profile web page support copy to clipboard

xinyiZzz:
TODO, expect more information about process status, CPU, IO, etc.
image
image


yiguolei (approved):

wangbo (approved):
LGTM

PR #41894, opened 2024-10-15
[improvement](statistics)Use min row count of all replicas as tablet/table row count.

Jibing-Li:
Use min row count of all replicas with same version as tablet/table row count. Because replica with the least row count means it perform more compaction operation than the others. Use it as tablet row count is more accurate.
Meanwhile, use min row count as tablet row count while choosing tablets during sample analyze.

PR #41979, opened 2024-10-16
[improvement](statistics)Use min row count of all replicas as tablet/table row count. (#41894)

Jibing-Li:
backport: https://github.com/apache/doris/pull/41894


Jibing-Li:
run buildall

PR #41987, opened 2024-10-16
[enhancement](cloud) fix TTL cache size negative overflow

freemandealer:
reset_range enlarges the size if the download size is larger than the original size. TTL LRU queue missed that while still subtracting the larger size when entries were removed from the queue, causing the size metric negative overflow.


freemandealer:
run buildall

gavinchou (dismissed):

freemandealer:
run buildall

dataroaring (approved):
LGTM

gavinchou (approved):

PR #41993, opened 2024-10-16
[opt](scanner profile) Rename some filed name to keep consistent with audit log.

zhiqiang-hhhh:
`RawRowsRead` is moved from segment profile to scan operator profile.
`RawRowsRead` is renamed to `ScanRows` so that it has same name with audit log.
`ScanBytes` is added. It has same value with `CompressedBytesRead` in segment iterator profile

image

I am not sure whether we should modify the calculation of `CompressedBytesRead`.
Currently, `CompressedBytesRead` does not include size of page that is cached by doris page cache.


zhiqiang-hhhh:
run buildall

yiguolei (approved):

morningman (approved):

PR #41273, opened 2024-09-25
[fix](scanner) Fix incorrect _max_thread_num in scanner context when many queries are running.

zhiqiang-hhhh:
1. Minor refactor for scanner constructor, calculation of _max_thread_num is moved to init method
2. The expected value of _max_thread_num is changed. There is no need to submit too many scan task to scan scheduler, since thread num is limited.
3. Calculation of _max_bytes_in_queue is changed. _max_bytes_in_queue for each scan instance is limited to 100MB by default.

```
mysql [tpch]>select count(*) from supplier;
--------------
select count(*) from supplier
--------------

+----------+
| count(*) |
+----------+
|  1000000 |
+----------+
1 row in set (0.04 sec)

mysql [tpch]>select count(*) from revenue0;
--------------
select count(*) from revenue0
--------------

+----------+
| count(*) |
+----------+
|  1000000 |
+----------+
1 row in set (0.19 sec)
```
To illustrate the effect, we need to create much scanners, so 
```
set global experimental_parallel_scan_min_rows_per_scanner=29715
```
default value is `2097152`,  we can make scanner num almost equal to `experimental_parallel_scan_max_scanners_count` which is 48.

Lets use mysqlslap to do concurrent test.

Current master:
```text
[hezhiqiang@VM-10-8-centos be_1]$ mysqlslap -hxxxx -uroot -Pyyyy  --create-schema=tpch -c 20 -i 5 -q "select     s_suppkey,     s_name,     s_address,     s_phone,     total_revenue from     supplier,     revenue0 where     s_suppkey = supplier_no     and total_revenue = (         select             max(total_revenue)         from             revenue0     ) order by     s_suppkey;"
Benchmark
	Average number of seconds to run all queries: 12.480 seconds
	Minimum number of seconds to run all queries: 12.159 seconds
	Maximum number of seconds to run all queries: 12.843 seconds
	Number of clients running queries: 20
	Average number of queries per client: 1

[hezhiqiang@VM-10-8-centos be_1]$ mysqlslap -hyyyy -uroot -Pyyyy  --create-schema=tpch -c 25 -i 5 -q "select     s_suppkey,     s_name,     s_address,     s_phone,     total_revenue from     supplier,     revenue0 where     s_suppkey = supplier_no     and total_revenue = (         select             max(total_revenue)         from             revenue0     ) order by     s_suppkey;"
[...]

PR #42014, opened 2024-10-17
[fix](scanner) Fix incorrect _max_thread_num in scanner context when many queries are running. #41273

zhiqiang-hhhh:
cherry pick from #41273


zhiqiang-hhhh:
run buildall

PR #42019, opened 2024-10-17
[fix](mtmv) Fix compensate union all wrongly when query rewrite by materialized view #40803

seawinde:
## Proposed changes

pr: https://github.com/apache/doris/pull/40803
commitId: d7e5d461


seawinde:
run buildall

PR #41666, opened 2024-10-10
[fix](information-meta) fix information column_type and data_type for ip/variant

amorynan:
## Proposed changes
 fix information column_type and data_type for ip/variant
image

Issue Number: close #xxx

PR #42026, opened 2024-10-17
[fix](information-meta) fix information column_type and data_type missing for ip/variant 

amorynan:
pick #41666


amorynan:
run buildall

PR #41655, opened 2024-10-10
[enhancement](sequence col) add session variable to skip sequence column check while INSERT INTO

zhannngchen:
## Proposed changes

Issue Number: close #xxx

For unique key tables using a sequence column, the `INSERT INTO` operation requires a sequence value to be provided for each row. 
This PR add a new session variable `require_sequence_in_insert` to control this behavior.

PR #42030, opened 2024-10-17
[enhancement](sequence col) add session variable to skip sequence column check while INSERT INTO (#41655)

zhannngchen:
cherry-pick #41655

For unique key tables using a sequence column, the `INSERT INTO` operation requires a sequence value to be provided for each row. This PR add a new session variable `require_sequence_in_insert` to control this behavior.


zhannngchen:
run buildall

PR #42032, opened 2024-10-17
[fix](Nereids) fix `Is  Cached` is not Yes in Profile when enable_sql_cache=true

924060929:
## Proposed changes

fix `Is  Cached` is not Yes in Profile when enable_sql_cache=true, introduced by #33262


924060929:
run buildall

morrySnow (approved):

starocean999 (approved):

PR #42034, opened 2024-10-17
[fix](Nereids) fix Is Cached is not Yes in Profile when enable_sql_cache=true (#42032)

924060929:
## Proposed changes

cherry pick from #42032


924060929:
run buildall

PR #42039, opened 2024-10-17
[fix](move-memtable) abstract multi-streams to one logical stream

kaijchen:
## Proposed changes
Currently, an upstream BE (sink_v2) will open multiple streams to a downstream BE (load_stream).
If any of the streams fails, the use_cnt on the downstream BE will be messed up.
The load_stream will not report any success tablets to the sink_v2 since in its view there are still unfinished streams.

So fault tolerance when open streams is not meaningful in practical, and may cause data lost.
i.e. Upstream think there is still working streams to transfer data, but downstream does not report any commit info.

This PR removes fault tolerance when open multiple streams to the same backend.
If any of the open fails, the upstream sink_v2 should mark the downstream BE as failed replicas.

### DIFF
--- a/be/src/vec/sink/load_stream_map_pool.cpp
+++ b/be/src/vec/sink/load_stream_map_pool.cpp
@@ -35,22 +35,20 @@ LoadStreamMap::LoadStreamMap(UniqueId load_id, int64_t src_id, int num_streams, 
     DCHECK(num_use > 0) << "use num should be greater than 0";
 }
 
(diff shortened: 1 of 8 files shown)


kaijchen:
run buildall

kaijchen:
run buildall

kaijchen:
run buildall

liaoxin01 (approved):
LGTM

sollhui (approved):
LGTM

PR #16553, opened 2023-02-09
[BugFix](cooldown) push correct cooldownttl to be

ByteYue:
# Proposed changes

Issue Number: close #xxx
There were `cooldownttl` and `cooldownttlms` in StoragePolicy, it's so error-prone because they served nearly the same.
For example, the `init` function would only assign the ttl timestamp to `cooldownttl`, which would end up pushing cooldownttl 0 to be.
image

image

BTW, the former logic would record microsecond but would push seconds to be, so I changed the time unit to second.

## Problem summary

Describe your changes.

## Checklist(Required)

1. Does it affect the original behavior: 
    - [ ] Yes
    - [x] No
    - [ ] I don't know
2. Has unit tests been added:
    - [ ] Yes
    - [x] No
    - [ ] No Need
3. Has document been added or modified:
    - [ ] Yes
    - [x] No
    - [ ] No Need
4. Does it need to update dependencies:
    - [ ] Yes
    - [x] No
5. Are there any changes that cannot be rolled back:
    - [ ] Yes (If Yes, please explain WHY)
    - [x] No

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #42083, opened 2024-10-18
[fix](cooldown) fix cooldown datetime

justfortaste:
## Proposed changes

Issue Number: close #xxx

#16553 cooldown_ttl and cooldown_datetime cann't set at the same time.  If  cooldown_datetime is set, cooldown_ttl will be set to -1.   
Tablet::_get_newest_cooldown_time  will always get storage_policy->cooldown_datetime , then all new rowsets will be  incorrectly cooldowned.


dataroaring:
if cooldown_ttl is -1, then cooldown_datetime is used, original code is right. Could you show an bad case?

lide-reed (approved):
LGTM

PR #42084, opened 2024-10-18
[fix](AuditLog) add `cloudClusterName` field for `audit_log` system table

BePPPower:
## Proposed changes

Issue Number: close #xxx


fe/fe-core/src/main/java/org/apache/doris/catalog/InternalSchema.java
```diff
@@ -152,6 +152,8 @@ public class InternalSchema {
         AUDIT_SCHEMA.add(
                 new ColumnDef("workload_group", TypeDef.create(PrimitiveType.STRING), ColumnNullableType.NULLABLE));
         AUDIT_SCHEMA.add(new ColumnDef("stmt", TypeDef.create(PrimitiveType.STRING), ColumnNullableType.NULLABLE));
+        AUDIT_SCHEMA.add(new ColumnDef("cloudClusterName", TypeDef.create(PrimitiveType.STRING),
```
dataroaring, line 155: It should be computeGroupName

dataroaring (changes requested):

PR #42102, opened 2024-10-18
[improvement](jdbc catalog) Disallow non-constant type conversion pushdown and implicit conversion pushdown

zy-kkk:
Add a variable `enable_jdbc_cast_predicate_push_down`, the default value is false, which prohibits the pushdown of non-constant predicates with type conversion and all predicates with implicit conversion. This change can prevent the wrong predicates from being pushed down to the Jdbc data source, resulting in query data errors, because the predicates with cast were not correctly pushed down to the data source before.
If you find that the data is read correctly and the performance is better before this change, you can manually set this variable to true

```
| Expression                                          | Can Push Down |
|-----------------------------------------------------|---------------|
| column type equals const type                       | Yes           |
| column type equals cast const type                  | Yes           |
| cast column type equals const type                  | No            |
| cast column type equals cast const type             | No            |
| column type not equals column type                  | No            |
| column type not equals cast const type              | No            |
| cast column type not equals const type              | No            |
| cast column type not equals cast const type         | No            |

```


zy-kkk:
run buildall

fe/fe-core/src/main/java/org/apache/doris/datasource/jdbc/source/JdbcScanNode.java
```diff
@@ -138,6 +139,11 @@ private void createJdbcFilters() {
     private List<Expr> collectConjunctsToPushDown(List<Expr> conjunctsList, List<String> errors) {
         List<Expr> pushDownConjuncts = new ArrayList<>();
         for (Expr p : conjunctsList) {
+            if (ConnectContext.get() != null && !ConnectContext.get().getSessionVariable().jdbcCastExprPushdown) {
```
morningman, line 142: If `ConnectContext.get()` is null, I think we should also disallow to push down cast expr

morningman: And why not write this logic in `shouldPushDownConjunct()`?

fe/fe-core/src/main/java/org/apache/doris/qe/SessionVariable.java
```diff
@@ -504,6 +504,8 @@ public class SessionVariable implements Serializable, Writable {
 
     public static final String JDBC_ORACLE_NULL_PREDICATE_PUSHDOWN = "jdbc_oracle_null_predicate_pushdown";
 
+    public static final String JDBC_CAST_EXPR_PUSHDOWN = "jdbc_cast_expr_pushdown";
```
morningman, line 507:
```suggestion
    public static final String JDBC_CAST_EXPR_PUSHDOWN = "enable_jdbc_push_down_cast_expr";
```

So that we can unify the variable name like `enable_jdbc_push_down_xxx`

morningman (dismissed):
LGTM

morningman:
run buildall

zy-kkk:
run buildall

morningman (approved):
LGTM

CalvinKirs (approved):

PR #41939, opened 2024-10-16
[fix](catalog) fix filtered database when use_meta_cache=true

morningman:
Followup #38244
When setting `use_meta_cache=true`, the `include_database_list` and `exclude_database_list`
will have no effect. This PR fix it.

Also fix a bug, related to #40479.
if `include_database_list` and `exclude_database_list` is set, and `use` a database which is excluded,
Doris should return "database not found" error

PR #42114, opened 2024-10-18
[fix](catalog) fix filtered database when use_meta_cache=true #41939

morningman:
cherry pick from #41939


morningman:
run buildall

PR #42115, opened 2024-10-18
[bugfix](handshake) brpc handshake should not use light pool

yiguolei:
## Proposed changes

The light pool may be full. Handshake is used to check the connection state of brpc. Should not be interfered by the thread pool logic.


mrhhsg (approved):
lgtm

yiguolei:
run buildall

liaoxin01 (approved):
LGTM

PR #41892, opened 2024-10-15
[opt](log) avoid print too much log when get partitions failed

morningman:
If get partitions failed, there may be too much logs to print partition values.

PR #42116, opened 2024-10-18
[opt](log) avoid print too much log when get partitions failed #41892

morningman:
cherry pick from #41892


morningman:
run buildall

PR #41662, opened 2024-10-10
[Fix](Export) `show export` statement supports specify the catalog name

BePPPower:
## Proposed changes

Issue Number: close #xxx

PR #42117, opened 2024-10-18
[Fix](Export) `show export` statement supports specify the catalog name #41662

morningman:
cherry pick from #41662


morningman:
run buildall

PR #41606, opened 2024-10-09
[Test](tvf) add regression tests for testing orc reader

BePPPower:
## Proposed changes

Issue Number: close #xxx

PR #42120, opened 2024-10-18
[Test](tvf) add regression tests for testing orc reader #41606 #42188

morningman:
cherry pick from #41606 #42188


morningman:
run buildall

PR #42127, opened 2024-10-19
[bugfix](handshake) brpc handshake should not use light pool (#42115)

yiguolei:
The light pool may be full. Handshake is used to check the connection state of brpc. Should not be interfered by the thread pool logic.

---------
pick #42115

## Proposed changes

Issue Number: close #xxx


yiguolei:
run buildall

PR #42130, opened 2024-10-19
[Enhancement][recycle-bin] unpartition table need to keep in recycle bin in insert overwrite case

Vallishp:
follow up for #40512 

Issue Number: close #xxx

if a insert overwrite happens to un-partitioned table, partition info needs to be kept inside the recycle bin. (without this tablets may leak)
once the data kept in recycle bin, it can be recovered using below step.

recover the partition to table as another partition in the table.
create another temp user table , and select the recover partition info and insert to it.
replace original user table from  temp user table.


Vallishp:
run buildall

dataroaring (approved):
LGTM

yujun777 (approved):
LGTM

PR #42001, opened 2024-10-17
[fix](case) fix txn_insert_inject_case

mymeiyi:

PR #42141, opened 2024-10-21
[fix](case) fix txn_insert_inject_case (#42001)

mymeiyi:
pick https://github.com/apache/doris/pull/42001


mymeiyi:
run buildall

PR #42145, opened 2024-10-21
[opt][file cache] Do not return error if file cache can not be created if ignore_broken_disk is set

TszKitLo40:
## Proposed changes

Issue Number: close #xxx

If `ignore_broken_disk` is set, do not return error if file cache can not be created


morningman (approved):
LGTM

yiguolei (approved):

yiguolei:
run buildall

PR #36543, opened 2024-06-19
[enhance](mtmv)mtmv query sql expand star

zddr:
before: SELECT * from user

now: SELECT `internal`.`zd`.`user`.`k1`, `internal`.`zd`.`user`.`k2` from `internal`.`zd`.`user`

Therefore, there is no need for EnvInfo to store catalog and db information when creating materialized views

PR #42152, opened 2024-10-21
[fix](mtmv)Fix the old version of the materialized view error 'Current database is not set'`

zddr:
cause by: https://github.com/apache/doris/pull/36543

after 36543,not need set db&catalog in `ConnectContext`, because query sql Because the query statement was rewritten to the full path (ctl.db. table. col) during creation,

but query sql of old materialized view not was rewritten, so need set db&catalog in `ConnectContext`


zddr:
run buildall

zddr:
run buildall

morrySnow (approved):

seawinde (approved):
LGTM

PR #41875, opened 2024-10-15
[enhance](parquet) support reading brotli compressed parquet file

suxiaogang223:
## Proposed changes

Impl BrotliBlockCompression to uncompressed brotli parquet data.
fix parquet case: group0/large_string_map.brotli.parquet

PR #42177, opened 2024-10-21
[enhance](parquet) support reading brotli compressed parquet file #41875

morningman:
cherry pick from #41875


morningman:
run buildall

PR #42179, opened 2024-10-21
[fix](jdbc catalog) Disable oracle scan null operator pushdown #41563

morningman:
cherry pick from #41563


morningman:
run buildall

PR #42181, opened 2024-10-21
[improvement](jdbc catalog) Print more errors when Jdbc Catalog fails to obtain a connection on FE #41735

morningman:
cherry pick from #41735


morningman:
run buildall

PR #42188, opened 2024-10-21
[fix](regression-test) fix unstable case `test_hdfs_orc_group0_orc_files`

BePPPower:
We should add `order by col1` becuase there are multiple repeated col0 rows.


BePPPower:
run buildall

morningman (approved):

CalvinKirs (approved):

PR #42190, opened 2024-10-21
[Bug](exchange)fix the exchange not skip offset rows

zhangstar333:
## Proposed changes
before the fix, the exchange operator will return some error result.

the `set_num_rows` function will save some rows[0, length), remove others;
but here need skip offset rows, so need skip[0,offset) rows, should use `skip_num_rows`.


zhangstar333:
run buildall

HappenLee (approved):
LGTM

jacktengg (approved):

PR #42203, opened 2024-10-21
[fix](case)Fix multiple case tables with the same name causing case failure

zddr:
Fix multiple case tables with the same name causing case failure


zddr:
run buildall

morrySnow (approved):

starocean999 (approved):

PR #42206, opened 2024-10-21
[fix](nereids) Fix not check column name when create or alter view

seawinde:
## Proposed changes
This is brought by https://github.com/apache/doris/pull/32743

set enable_unicode_name_support = true;
If run create view sql should fail beausel_shipdate column name  contains invalid char '(' and ')', but now success
this pr fix this and throw exception
`ERROR 1105 (HY000): errCode = 2, detailMessage = Incorrect column name '(日期)'. Column regex is '^[_a-zA-Z@0-9\s/][.a-zA-Z0-9_+-/?@#$%^&*"\s,:]{0,255}$'`

```sql
CREATE VIEW view1
AS
SELECT "零售公司", l_shipdate as '(日期)', l_receiptdate as k2
FROM lineitem;
```
and if run create view sql as following, should success:
```sql
CREATE VIEW view2
AS
SELECT "零售公司", l_shipdate as '日期', l_receiptdate as k2
FROM lineitem;
```

and the schema of view2 should be
```sql
mysql> desc view2;
+-------------+-------------+------+-------+---------+-------+
| Field       | Type        | Null | Key   | Default | Extra |
+-------------+-------------+------+-------+---------+-------+
| __literal_0 | varchar(16) | No   | false | NULL    |       |
| 日期        | date        | No   | false | NULL    |       |
| k2          | date        | No   | false | NULL    |       |
+-------------+-------------+------+-------+---------+-------+
3 rows in set (0.01 sec)
```

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/nereids/trees/plans/commands/info/AlterViewInfo.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/nereids/trees/plans/commands/info/AlterViewInfo.java
@@ -20,7 +20,6 @@ package org.apache.doris.nereids.trees.plans.commands.info;
 import org.apache.doris.analysis.AlterViewStmt;
 import org.apache.doris.analysis.ColWithComment;
 import org.apache.doris.analysis.TableName;
-import org.apache.doris.catalog.Column;
 import org.apache.doris.catalog.DatabaseIf;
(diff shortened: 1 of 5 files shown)


seawinde:
run buildall

feiniaofeiafei (approved):

morrySnow (approved):

PR #42211, opened 2024-10-21
[fix](regression-test) Make `test_decimal256_outfile_csv` export the data to S3 rather than local file system

BePPPower:
Because it is difficult to check out the outfile files when export data to local file system, it's better to export data to S3.


BePPPower:
run buildall

morningman (approved):

zddr (approved):

PR #37403, opened 2024-07-07
[Fix] Error: missing field initializer

wyxxxcat:
## Proposed changes

Issue Number: close #xxx

Compiling with -Werror turns a warning into an error , clang-18 now compiles correctly
[](https://img.picui.cn/free/2024/07/07/668a98dcd88ea.png)

PR #42215, opened 2024-10-21
[Fix] (compile) Missing field initializer and unused const variable

wyxxxcat:
## Proposed changes

Issue Number: close #xxx

Earlier, we turned on the -Wno-missing-field-initializers and -Wno-unused-const-variable switches, now choose to disable them

same as #37403


zclllyybb (approved):
LGTM

HappenLee (approved):
LGTM

PR #42217, opened 2024-10-21
[regression](case)fix mc regression test p2 case.

hubgeter:
## Proposed changes
In previous versions, use a db that does not exist in show databases will not cause an error, so this test can be run before. 

Since the test account does not have the list all projects permission, there is no db want to be used when show databases， which causes the case to fail.

Therefore, creat a new catalog and  specify  `mc.default.project`  to fix this case.


hubgeter:
run buildall

morningman (dismissed):
LGTM

hubgeter:
run buildall

morningman (approved):

CalvinKirs (approved):

PR #41795, opened 2024-10-14
[fix](OrcReader) fix the issue that orc_reader can not read DECIMAL(0,0) type of orc file

BePPPower:
Because HIVE 0.11 & 0.12 does not support precision and scale for decimal
The decimal type of orc file produced by HIVE 0.11 & 0.12 are DECIMAL(0,0)
We should set a default precision and scale for these orc files.

PR #42220, opened 2024-10-21
[fix](OrcReader) fix the issue that orc_reader can not read DECIMAL(0,0) type of orc file #41795

morningman:
cherry pick from #41795


morningman:
run buildall

PR #42113, opened 2024-10-18
[fix](multi-catalog) Disable string dictionary filtering when predicate express is not slot

suxiaogang223:
## Proposed changes
follow up https://github.com/apache/doris/pull/35335/
When the `"case when ... then ... when ... then ... else"` occurs, function_expr may not exist in the pushed down predicate, but the handling of null values ​​is still problematic.

table data:
```text
mysql> select o_orderpriority from test_string_dict_filter_orc;
+-----------------+
| o_orderpriority |
+-----------------+
| 5-LOW           |
| 1-URGENT        |
| 5-LOW           |
| NULL            |
| 5-LOW           |
+-----------------+
```

before:
```text
mysql> select count(o_orderpriority) from ( select (case when o_orderpriority = 'x' then '1' when o_orderpriority = 'y' then '2' else '0' end) as o_orderpriority from test_string_dict_filter_orc ) as A where o_orderpriority = '0';
+------------------------+
| count(o_orderpriority) |
+------------------------+
|                      4 |
+------------------------+
```

after:
```text
mysql> select count(o_orderpriority) from ( select (case when o_orderpriority = 'x' then '1' when o_orderpriority = 'y' then '2' else '0' end) as o_orderpriority from test_string_dict_filter_orc ) as A where o_orderpriority = '0';
+------------------------+
| count(o_orderpriority) |
+------------------------+
|                      5 |
+------------------------+
```

PR #42223, opened 2024-10-21
[fix](multi-catalog) Disable string dictionary filtering when predicate express is not slot #42113

morningman:
cherry pick from #42113


morningman:
run buildall

PR #41772, opened 2024-10-12
[Fix](count on index) fix count on index opt when count project expr

airborne12:
## Proposed changes

This PR addresses an issue where counting on an index with a projected expression was suboptimal.

PR #42228, opened 2024-10-22
[Fix](count on index) fix count on index opt when count project expr #41772

airborne12:
cherry pick from #41772


airborne12:
run buildall

PR #42244, opened 2024-10-22
[improvement](statistics)Drop expired external stats only when the catalog is dropped.

Jibing-Li:
Drop expired external stats only when the catalog is dropped, to reduce meta store access.
Before, we go through all external catalogs and their DBs and tables to check expired stats, which may bring lots of meta store access.


Jibing-Li:
run buildall

Jibing-Li:
run buildall

Jibing-Li:
run feut

Jibing-Li:
run feut

morrySnow (approved):

zfr9527 (approved):
LGTM

PR #42246, opened 2024-10-22
[fix](agenttask) failed tasks and remove them from queue if an error

dataroaring:
happens

## Proposed changes

Issue Number: close #xxx


dataroaring:
run buildall

fe/fe-core/src/main/java/org/apache/doris/task/AgentBatchTask.java
```diff
@@ -172,7 +173,6 @@ public void run() {
                     try {
                         agentTaskRequests.add(toAgentTaskRequest(task));
                     } catch (Exception e) {
-                        task.failed();
```
yujun777, line 175: delete  "try catch" at line 172.  it only throw, no really  handle the exception

PR #42247, opened 2024-10-22
[fix](inverted index) Fix errors caused by enable_need_read_data_opt #42064

airborne12:
cherry pick from #42064


airborne12:
run buildall

PR #42253, opened 2024-10-22
[Chore](job)Provides configuration of job execution queue size

CalvinKirs:
## Proposed changes

When dealing with a large number of tasks, the default execution queue size is 1024. This can lead to tasks being dropped if the queue becomes full. 
eg

`dispatch instant task failed, job id is xxx`

To address this, you can add the parameters `insert_task_queue_size` and `mtmv_task_queue_size` in the `fe.conf` configuration file. These parameters must be set to a power of 2. 

**Keep in mind, increasing this value is recommended only when thread resources are limited; otherwise, you should consider increasing the number of task execution threads.**

### DIFF
--- a/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
@@ -1759,7 +1759,7 @@ public class Config extends ConfigBase {
      * corresponding type of job
      * The value should be greater than 0, if it is 0 or <=0, set it to 5
      */
-    @ConfField(description = {"用于分发定时任务的线程数",
+    @ConfField(masterOnly = true, description = {"用于分发定时任务的线程数",
             "The number of threads used to dispatch timer job."})
     public static int job_dispatch_timer_job_thread_num = 2;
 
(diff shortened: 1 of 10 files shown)


CalvinKirs:
run buildall

fe/fe-common/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -1776,6 +1776,12 @@ public class Config extends ConfigBase {
             "Maximum number of persistence allowed per task in a job,exceeding which old tasks will be discarded，"
                    + "If the value is less than 1, it will not be persisted." })
     public static int max_persistence_task_count = 100;
+    @ConfField(description = {"MV task 的等待队列大小，值必须是 2 的幂，否则将使用默认值 1024", "The size of the MV task's"
```
morningman, line 1779: use `setter` to check if value is power of 2.
And you can also check if the value is too large or too small.
And need to test unit test

fe/fe-core/src/main/java/org/apache/doris/job/manager/TaskDisruptorGroupManager.java
```diff
@@ -65,8 +65,10 @@ public class TaskDisruptorGroupManager<T extends AbstractTask> {
     private static final int DISPATCH_MTMV_THREAD_NUM = Config.job_mtmv_task_consumer_thread_num > 0
             ? Config.job_mtmv_task_consumer_thread_num : DEFAULT_CONSUMER_THREAD_NUM;
 
-    private static final int DISPATCH_INSERT_TASK_QUEUE_SIZE = DEFAULT_RING_BUFFER_SIZE;
-    private static final int DISPATCH_MTMV_TASK_QUEUE_SIZE = DEFAULT_RING_BUFFER_SIZE;
+    private static final int DISPATCH_INSERT_TASK_QUEUE_SIZE = isPowerOfTwo(Config.insert_task_queue_size)
```
morningman, line 68: Do not check `isPowerOfTwo` here

CalvinKirs:
run buildall

morningman (approved):
LGTM

CalvinKirs:
run buildall

CalvinKirs:
run cloudp0

CalvinKirs:
run cloud_p0

morrySnow (approved):

PR #41992, opened 2024-10-16
[improvement](jdbc catalog) Add catalog property to enable jdbc connection pool

zy-kkk:
We initially introduced jdbc connection pool to improve the connection performance of jdbc catalog, but we always found that connection pool would bring some unexpected errors, so we chose to add a catalog property: `enable_connection_pool` to choose whether to enable the jdbc connection pool of jdbc catalog, and the default false.However, the created catalog will still open the connection pool when it is upgraded, and only the newly created catalog will be false

And we conducted performance tests on this, the performance loss is within the expected range.

- Enable connection pool:
mysqlslap -uroot -h127.0.0.1 -P9030 --concurrency=1 --iterations=100 --query='SELECT * FROM mysql.test.test limit 1;' --create-schema=mysql --delimiter=";" --verbose 
Benchmark
        Average number of seconds to run all queries: 0.008 seconds
        Minimum number of seconds to run all queries: 0.004 seconds
        Maximum number of seconds to run all queries: 0.133 seconds
        Number of clients running queries: 1
        Average number of queries per client: 1

- Disable connection pool:
mysqlslap -uroot -h127.0.0.1 -P9030 --concurrency=1 --iterations=100 --query='SELECT * FROM mysql_no_pool.test.test limit 1;' --create-schema=mysql --delimiter=";" --verbose
Benchmark
        Average number of seconds to run all queries: 0.054 seconds
        Minimum number of seconds to run all queries: 0.047 seconds
        Maximum number of seconds to run all queries: 0.184 seconds
        Number of clients running queries: 1
        Average number of queries per client: 1

PR #42255, opened 2024-10-22
[2.1][improvement](jdbc catalog) Add catalog property to enable jdbc connection pool

zy-kkk:
pick (#41992)

We initially introduced jdbc connection pool to improve the connection performance of jdbc catalog, but we always found that connection pool would bring some unexpected errors, so we chose to add a catalog property: `enable_connection_pool` to choose whether to enable the jdbc connection pool of jdbc catalog, and the default false.However, the created catalog will still open the connection pool when it is upgraded, and only the newly created catalog will be false

And we conducted performance tests on this, the performance loss is within the expected range.

- Enable connection pool: mysqlslap -uroot -h127.0.0.1 -P9030 --concurrency=1 --iterations=100 --query='SELECT * FROM mysql.test.test limit 1;' --create-schema=mysql --delimiter=";" --verbose
Benchmark
        Average number of seconds to run all queries: 0.008 seconds
        Minimum number of seconds to run all queries: 0.004 seconds
        Maximum number of seconds to run all queries: 0.133 seconds
        Number of clients running queries: 1
        Average number of queries per client: 1

- Disable connection pool: mysqlslap -uroot -h127.0.0.1 -P9030 --concurrency=1 --iterations=100 --query='SELECT * FROM mysql_no_pool.test.test limit 1;' --create-schema=mysql --delimiter=";" --verbose
Benchmark
        Average number of seconds to run all queries: 0.054 seconds
        Minimum number of seconds to run all queries: 0.047 seconds
        Maximum number of seconds to run all queries: 0.184 seconds
        Number of clients running queries: 1
        Average number of queries per client: 1


zy-kkk:
run buildall

zy-kkk:
run buildall

PR #42042, opened 2024-10-17
[fix](routine load) make the timeout of load channel consistent with routine load task

sollhui:
Routine load task timeout is max_batch_interval * 10, but load channel timeout is max_batch_interval * 2.

related pr: https://github.com/apache/doris/pull/40818

PR #42267, opened 2024-10-22
[fix](routine load) make the timeout of load channel consistent with routine load task (#42042)

sollhui:
pick (#42042)

Routine load task timeout is max_batch_interval * 10, but load channel timeout is max_batch_interval * 2.


sollhui:
run buildall

PR #42271, opened 2024-10-22
[fix](regression-test) Make `test_decimal256_outfile_csv` export the data to S3 rather than local file system #42211

morningman:
cherry pick from #42211


morningman:
run buildall

PR #42200, opened 2024-10-21
[enhance](hive) Add regression-test cases for hive text ddl and hive text insert and fix reading null string bug

suxiaogang223:
## Proposed changes
Add regression-test cases for hive text table properties:
| **Property**                 | **Description**                                                     | **Example Value** | **Supported in Doris** |
|------------------------------|---------------------------------------------------------------------|-------------------|----------------------|
| `field.delim`                | Defines the delimiter between columns in each row.                  | `\1`              | Yes                  |
| `collection.delim`           | Defines the delimiter for items in an array (collection type).      | `\2`              | Yes                  |
| `mapkey.delim`               | Defines the delimiter between keys and values in a map.             | `\3`              | Yes                  |
| `serialization.null.format`  | Defines how `NULL` values are represented in the text file.         | `\\N`             | Yes                  |
| `escape.delim`               | Specifies the escape character used for escaping special characters.| `\\`              | Yes                  |
| `line.delim`                 | Defines the delimiter between rows or lines in the file.            | `\n`              | Yes                  |

### Explanation:
- **`field.delim`**: This property is used to specify how columns are separated in a row. For example, `\1` indicates the delimiter is a non-printable character.
- **`collection.delim`**: Used to define how elements in an array or collection are separated. In this case, `\2` is used as a separator.
- **`mapkey.delim`**: This property defines how keys and values in a map are separated. For instance, `\3` is used to separate map keys and values.
- **`serialization.null.format`**: This setting specifies the format for `NULL` values in the data. `\\N` is commonly used to represent `NULL`.
- **`escape.delim`**: Defines the escape character used in the text file for escaping special characters, such as the delimiter itself. Here, `\\` is used as the escape character.
- **`line.delim`**: This property is used to specify the delimiter between lines or rows. Typically, `\n` (newline) is used as the line delimiter.
**Note**: Unlike the other delimiters, `line.delim` is not escaped. If the content in the table contains the same character as the line delimiter, it may cause query errors. However, the other delimiters (`field.delim`, `collection.delim`, `mapkey.delim`) are escaped, so they will not cause issues.

PR #42272, opened 2024-10-22
[enhance](hive) Add regression-test cases for hive text ddl and hive text insert and fix reading null string bug #42200

morningman:
cherry pick from #42200


morningman:
run buildall

PR #42279, opened 2024-10-22
[test](mtmv) Auto calc the value of sync_cbo_rewrite variable to makesure the same test code in different version

seawinde:
## Proposed changes

Auto calc the value of sync_cbo_rewrite variable to make sure the same test code in different version.
Because the query rewrite by mv code is a little different from 2.1 to 3.0 and master.
But should keep same regression test code in different version. 
So add the methoed `enable_sync_mv_cost_based_rewrite` to  get `enable_sync_mv_cost_based_rewrite`  session variable, then decide to regression test logic by the  variable auto.


seawinde:
run buildall

seawinde:
run buildall

seawinde:
run buildall

zfr9527 (approved):
LGTM

morrySnow (approved):

PR #42280, opened 2024-10-22
[opt](index compaction)Add dual write inverted index file switch

qidaye:
## Proposed changes

To check correctness of index file produced by index compaction process, we add a switch `dual_write_inverted_index_enable` (default is `false`). 
When both `inverted_index_compaction_enable` and `dual_write_inverted_index_enable` are `true`, Doris will produce index file through both normal compaction process and index compaction process, and compares both index files, which are theoretically identical. Doris will log FATAL and crash after the check failed.

We add this feature only for test, *DO NOT* use it in production.

### DIFF
--- a/be/src/common/config.cpp
+++ b/be/src/common/config.cpp
@@ -1042,6 +1042,8 @@ DEFINE_Int32(max_depth_in_bkd_tree, "32");
 DEFINE_mBool(inverted_index_compaction_enable, "false");
 // Only for debug, do not use in production
 DEFINE_mBool(debug_inverted_index_compaction, "false");
+// Only for debug, do not use in production
+DEFINE_mBool(dual_write_inverted_index_enable, "false");
 // index by RAM directory
 DEFINE_mBool(inverted_index_ram_dir_enable, "true");
 // use num_broadcast_buffer blocks as buffer to do broadcast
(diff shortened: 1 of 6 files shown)


qidaye:
run buildall

be/src/olap/rowset/segment_v2/inverted_index_file_writer.cpp
```diff
@@ -282,6 +282,10 @@ int64_t InvertedIndexFileWriter::write_v1() {
             ram_dir.deleteFile(idx_name.c_str());
             ram_dir.close();
 
+            if (is_index_compaction && config::dual_write_inverted_index_enable) {
```
airborne12, line 285: DO NOT couple debug code with the main process logic.

qidaye: Done.

qidaye:
run buildall

airborne12 (approved):
LGTM

csun5285 (commented):
Modify the configuration within the pipeline to ensure that checks are effective in the pipeline.
1. regression-test/pipeline/p0/conf/be.conf
2. regression-test/pipeline/p1/conf/be.conf
3. regression-test/pipeline/cloud_p0/conf/be.conf
4. regression-test/pipeline/cloud_p1/conf/be.conf

qidaye:
> Modify the configuration within the pipeline to ensure that checks are effective in the pipeline.
We do not want to this correctness check logic affect the normal index compaction flow, so I don't recommend add this switch to pipeline.  
We can use it in independent test cases.

csun5285 (approved):
LGTM

PR #42051, opened 2024-10-17
[opt](index compaction)Optimize logic of picking columns for index compaction

qidaye:
## Proposed changes

1. Rename `skip_inverted_index` to `columns_to_do_index_compaction` to make it more clear.
2. Remove redundant rowid_conversion check in index compaction
3. Rename function `construct_skip_inverted_index` to `construct_index_compaction_columns`
4. Check `index.col_unique_ids` empty before using to avoid crash and log tablet schema for debug.
5. Avoid unnecessary bkd and variant index checks by advancing column type checking before `has_inverted_index`

PR #42287, opened 2024-10-22
[opt](index compaction)Optimize logic of picking columns for index compaction(#42051)

qidaye:
## Proposed changes

bp #42051


airborne12 (approved):
LGTM

PR #41529, opened 2024-10-08
[test](copy into) add copy into regression test case

sollhui:
Add copy into regression test case.

PR #42289, opened 2024-10-22
[test](copy into) add copy into regression test case (#41529)

sollhui:
pick (#41529)

Add copy into regression test case.


sollhui:
run buildall

PR #42297, opened 2024-10-22
[fix](load) Fix the issue of high-concurrency single-replica load getting stuck

liaoxin01:
In high-concurrency single-replica load, the tablet_writer_add_block RPC may occupy the _heavy_work_pool completely, causing the response_slave_tablet_pull_rowset RPC to have no available threads for processing. As a result, tablet_writer_add_block waits indefinitely for a response from the slave tablet, leading to the import getting stuck until it times out.

response_slave_tablet_pull_rowset is relatively lightweight, so it can be handled by the _light_work_pool.


liaoxin01:
run buildall

dataroaring (approved):
LGTM

sollhui (approved):
LGTM

PR #42314, opened 2024-10-23
[enhancement](blacklist) should check the real error reason when sendfragment timeout

yiguolei:
## Proposed changes

1. send fragment in BE may run for long time.
2. timeout is not related with BE down.
3. cancel logic should not handle blacklist. because if a BE is down, send fragment will find it.


yiguolei:
run buildall

PR #42210, opened 2024-10-21
[Fix](fe) Avoid interrupt daemon thread and use proper polling interval

zclllyybb:
## Proposed changes

Issue Number: close #xxx

use a proper sleep interval to balance both:
a. no interrupt so no bdbje writing break which leads to crash
b. the daemon threads don't need much time to know its interval change.

PR #42322, opened 2024-10-23
[branch-3.0](fe) Avoid interrupt daemon thread and use proper polling interval (#42210)

zclllyybb:
pick https://github.com/apache/doris/pull/42210


zclllyybb:
run buildall

PR #42323, opened 2024-10-23
[fix](nereids) Fix not check column name when create or alter view (#42206)

seawinde:
pr: https://github.com/apache/doris/pull/42206
commitId: 2bcaa5b4


seawinde:
run buildall

PR #42327, opened 2024-10-23
[fix](txn insert) fix txn_insert case

mymeiyi:
when connect to observer, if the first load sql throw exception before execute(such as unpected label), the txn is not began


mymeiyi:
run buildall

dataroaring (approved):
LGTM

bobhan1 (approved):

zhannngchen (approved):
LGTM

PR #42339, opened 2024-10-23
[fix](mtmv)fix in the scenario of recreating a table, the materialize…

zddr:
…d view may assume that the data has not changed (#41762)

pick: https://github.com/apache/doris/pull/41762


zddr:
run buildall

PR #42349, opened 2024-10-23
[fix](estimate) func call with not filter will estimate some statistics #41989

amorynan:
## Proposed changes
bp: #41989


amorynan:
run buildall

PR #42060, opened 2024-10-18
[Improvement](segment iterator) Optimize column row reservation to reduce overhead

airborne12:
## Proposed changes

This PR improves the segment iterator by reducing the reserved rows for a column when the row bitmap is smaller than the block row. This optimization aims to enhance memory efficiency and improve performance by reducing unnecessary row allocations.

PR #42372, opened 2024-10-24
[Improvement](segment iterator) Optimize column row reservation to reduce overhead #42060

airborne12:
cherry pick from #42060


airborne12:
run buildall

PR #42374, opened 2024-10-24
[fix](index compaction) fix fd leak and mem leak while index compaction #41915

airborne12:
cherry pick from #41915


airborne12:
run buildall

PR #42380, opened 2024-10-24
[fix](Nereids) fix bind using join for compatible with MySQL behavior

924060929:
## Proposed changes

fix bind slot in using join for compatible with MySQL behavior

for example
```sql
select *
from t1
left join t2 using(id)
left join t3 using(id)
```

we should generate the join condition:
1. `t1.id = t2.id`
2. `t1.id = t3.id`,  previously, it was `t2.id = t3.id`


924060929:
run buildall

924060929:
run buildall

924060929:
run buildall

starocean999 (approved):

englefly (approved):

PR #41012, opened 2024-09-19
[fix](create table) Fix `test_partition_create_tablet_rr` docker regression case

deardeng:

PR #42388, opened 2024-10-24
[fix](create table) Fix `test_partition_create_tablet_rr` docker regression case #41012

deardeng:
cherry pick from #41012


deardeng:
run buildall

PR #42399, opened 2024-10-24
[improvement](statistics)Support auto analyze columns that haven't been analyzed for a long time.

Jibing-Li:
Support auto analyze columns that haven't been analyzed for a long time. Add a very low priority job queue for auto analyze to process this kind of columns.

The purpose of this change is to make sure all tables could be auto analyzed within a certain time. In the earlier Doris versions, users often encounter this kind of issues:
User load some new data to a large table everyday, but the change rate (percentage of new data) is very low, because there is a large size of old data. In this case, auto analyze for this table will not be triggered for a very long time, because the default trigger threshold of auto analyze is 40% (more than 40% of the data in a table is changed since last analyze). This will probably cause a bad plan because min/max/ndv statistics are outdated.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/analysis/ShowAutoAnalyzeJobsStmt.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/analysis/ShowAutoAnalyzeJobsStmt.java
@@ -40,7 +40,7 @@ import com.google.common.collect.ImmutableList;
  *        [TABLE]
  *        [
  *            WHERE
- *            [PRIORITY = ["HIGH"|"MID"|"LOW"]]
+ *            [PRIORITY = ["HIGH"|"MID"|"LOW"|"VERY_LOW"]]
  *        ]
  */
 public class ShowAutoAnalyzeJobsStmt extends ShowStmt implements NotFallbackInParser {
@@ -175,7 +175,7 @@ public class ShowAutoAnalyzeJobsStmt extends ShowStmt implements NotFallbackInPa
 
         if (!valid) {
             throw new AnalysisException("Where clause should looks like: "
-                    + "PRIORITY = \"HIGH|MID|LOW\"");
+                    + "PRIORITY = \"HIGH|MID|LOW|VERY_LOW\"");
         }
     }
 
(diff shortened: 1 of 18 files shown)


Jibing-Li:
run buildall

Jibing-Li:
run buildall

Jibing-Li:
run buildall

Jibing-Li:
run cloud_p0

Jibing-Li:
run buildall

fe/fe-core/src/main/java/org/apache/doris/analysis/ShowAutoAnalyzeJobsStmt.java
```diff
@@ -40,7 +40,7 @@
  *        [TABLE]
  *        [
  *            WHERE
- *            [PRIORITY = ["HIGH"|"MID"|"LOW"]]
+ *            [PRIORITY = ["HIGH"|"MID"|"LOW"|"LONG_TIME"]]
```
morningman, line 43: How about "VERY_LOW"?
LONG_TIME is not a "priority"

fe/fe-core/src/main/java/org/apache/doris/statistics/LongTimeJobAppender.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+/**
+ * This is a daemon thread to append long time not analyzed columns to long time job queue in AnalysisManager.
+ * Using this separate thread instead of using the existing StatisticsJobAppender is to avoid blocking other
+ * higher priority jobs.
+ */
+public class LongTimeJobAppender extends MasterDaemon {
```
morningman, line 44:
```suggestion
public class LongTimeStatisticJobAppender extends MasterDaemon {
```

Jibing-Li:
run buildall

Jibing-Li:
run p0

morningman (approved):

morrySnow (approved):

PR #42400, opened 2024-10-24
[fix](nereids) Fix not check when duplicated column names are  in cte select list

seawinde:
## Proposed changes

This is brought by https://github.com/apache/doris/pull/21727
If you run query as following:
Should throw analyze exception, because the duplicated column name in `tmp1`,but now not
this pr fix this
```sql
    with tmp1 as (
    select unique_id, unique_id  from test_table
    )
    select * from tmp1;
```


seawinde:
run buildall

seawinde:
run buildall

starocean999 (approved):

morrySnow (approved):

PR #42403, opened 2024-10-24
[fix](log) fix group commit warn log

mymeiyi:
if stream load in group commit mode connect to be directly, the fe does not select backend for it. the map used to record <table, be, dataSize> in fe is empty, so there are too many warn log:
```
[GroupCommitManager.updateLoadDataInternal():399] can not find backend id: 12229430
2024-10-21 08:58:28,975 WARN (thrift-server-pool-7|132)
```


mymeiyi:
run buildall

dataroaring (approved):
LGTM

CalvinKirs (approved):

PR #42404, opened 2024-10-24
[backport][branch-3.0] unpartition table need to keep in recycle bin in insert overwrite case

Vallishp:
bp #42130 

follow up for https://github.com/apache/doris/pull/40512

Issue Number: close #xxx

if a insert overwrite happens to un-partitioned table, partition info needs to be kept inside the recycle bin. (without this tablets may leak)
once the data kept in recycle bin, it can be recovered using below step.

recover the partition to table as another partition in the table.
create another temp user table , and select the recover partition info and insert to it.
replace original user table from temp user table.


Vallishp:
run buildall

PR #42412, opened 2024-10-24
[improvement](statistics)Drop expired external stats only when the catalog is dropped. (#42244)

Jibing-Li:
backport: https://github.com/apache/doris/pull/42244


Jibing-Li:
run buildall

PR #42419, opened 2024-10-24
[fix](auth)Fix concurrency issue during role manager upgrade 

zddr:
Change the map to a concurrency safe map，because there is a possibility of concurrent read and write operations

For example, when a node performs the rectifyPrivs operation, it needs to traverse 'roles' and may be adding or deleting roles at the same time


zddr:
run buildall

morrySnow (approved):

seawinde (approved):

PR #41064, opened 2024-09-20
[fix](schema change) fix delete predicate incorrect comparisons result with empty strings during schema change

luwei16:
### problem 

```
insert into t1 (k1,  k2,  v1) value(1, '', 2);
delete form t1 where k1 = 1 and k2 = '';
alter table modify column v1 string
select * from t1 //  expect 0 rows in return, but get 1 row.
```

PR #42427, opened 2024-10-24
[fix](schema change) fix delete predicate incorrect comparisons result with empty strings during schema change (#41064)

luwei16:
```
insert into t1 (k1,  k2,  v1) value(1, '', 2);
delete form t1 where k1 = 1 and k2 = '';
alter table modify column v1 string
select * from t1 //  expect 0 rows in return, but get 1 row.
```


luwei16:
run buildall

PR #42453, opened 2024-10-25
[improve](common) Add synchronized to avoid concurrent modification #42384

w41ter:
cherry pick from #42384


w41ter:
run buildall

PR #42470, opened 2024-10-25
[fix](cluster key) improve cluster key lookup rowkey

mymeiyi:
the min max key in mow with cluster key tables is also the min and max of primay keys, so when lookup rowkey, can use it to filter some segments.


mymeiyi:
run buildall

dataroaring (approved):
LGTM

zhannngchen (approved):
LGTM

zhannngchen:
the min max key in mow with cluster key tables is also the min and max of primay keys, so when lookup rowkey, can use it to filter some segments.

PR #42481, opened 2024-10-25
Revert "[2.1][improvement](jdbc catalog) Add catalog property to enable jdbc connection pool"

zy-kkk:
Reverts apache/doris#42255

We have found that after closing the connection pool, there will be class loading problems and connection release problems for some data sources. We will remove this function first and re-add it after solving and testing it completely.


zy-kkk:
run buildall

PR #42483, opened 2024-10-25
[refine](exchange buffer) Share the exchange sink buffer at the Fragment level part 1

Mryange:
## Proposed changes

 This is the first part of sharing the exchange sink buffer at the Fragment level.
 Some unused code has been removed.
 The buffer does not store send_id  be_number.


Mryange:
run buildall

PR #42489, opened 2024-10-25
[Fix](expr) fix wrong debug string of cast expr and remove useless variable

zclllyybb:
## Proposed changes

Issue Number: close #xxx

before we print target type as the argument type.

and for target type, we only need type itself. don't need dummy column.


zclllyybb:
run buildall

Mryange (approved):

Gabriel39 (approved):

PR #42494, opened 2024-10-25
[Improment](branch-2.1)optimize help module check

cjj2010:
## Proposed changes

Issue Number: close #xxx


zclllyybb:
why we need this? need elaborate

cjj2010:
> why we need this? need elaborate

When help.dip is missing, when fe starts, port 8030 starts normally, while port 9030 cannot start, and there are no relevant prompts in the log. fe is in a semi boot state. It has been confirmed that the lack of help.dip should not affect the normal start of the fe process, so some optimization is needed here

dataroaring:
run buildall

PR #42507, opened 2024-10-26
[enhancement](jni)append exception check for jni code.

hubgeter:
## Proposed changes
Because in jni, we need to manually check for exceptions thrown, and if we do not handle the exception, the program will continue to execute, which is not like the try catch logic we often use. Therefore, in order to reduce the potential risks caused by not handling exceptions, I checked most of the jni code and added exception detection for it.

### DIFF
--- a/be/src/util/jvm_metrics.cpp
+++ b/be/src/util/jvm_metrics.cpp
@@ -22,7 +22,9 @@
 #include <functional>
 
 #include "common/config.h"
+#include "util/defer_op.h"
 #include "util/metrics.h"
+
 namespace doris {
 
 #define DEFINE_JVM_SIZE_BYTES_METRIC(name, type)                                     \
@@ -78,7 +80,7 @@ DEFINE_COUNTER_METRIC_PROTOTYPE_5ARG(jvm_gc_g1_old_generation_time_ms, MetricUni
 
 const char* JvmMetrics::_s_hook_name = "jvm_metrics";
 
-JvmMetrics::JvmMetrics(MetricRegistry* registry, JNIEnv* env) {
+JvmMetrics::JvmMetrics(MetricRegistry* registry) {
     DCHECK(registry != nullptr);
     _registry = registry;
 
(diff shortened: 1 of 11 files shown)


hubgeter:
run buildall

be/src/util/jvm_metrics.cpp
```diff
@@ -90,9 +92,12 @@ JvmMetrics::JvmMetrics(MetricRegistry* registry, JNIEnv* env) {
             break;
         }
         try {
-            _jvm_stats.init(env);
+            Status st = _jvm_stats.init();
+            if (!st) {
+                LOG(WARNING) << "jvm Stats Init Fail. " << st.to_string();
```
morningman, line 97: Should we call `break` if `!st`?

be/src/util/jvm_metrics.cpp
```diff
@@ -182,193 +188,260 @@ void JvmMetrics::update() {
...
-    env = ENV;
-    _managementFactoryClass = env->FindClass("java/lang/management/ManagementFactory");
-    if (_managementFactoryClass == nullptr) {
-        LOG(WARNING)
-                << "Class java/lang/management/ManagementFactory Not Find.JVM monitoring fails.";
-        return;
-    }
+Status JvmStats::init() {
+    JNIEnv* env = nullptr;
+    RETURN_IF_ERROR(JniUtil::GetJNIEnv(&env));
```
morningman, line 193: Why do we need to get `env` again here?

hubgeter:
run buildall

be/src/util/jvm_metrics.cpp
hubgeter: There is really no need

be/src/util/jvm_metrics.cpp
hubgeter: It doesn't matter if it is needed or not, because I will check `_jvm_stats.init_complete()`.

morningman (approved):
LGTM

CalvinKirs (approved):

PR #42067, opened 2024-10-18
[fix](agg) Aggregating string types with null values may result in incorrect result

liaoxin01:
Aggregating string types with null values may result in incorrect result because using the replace_column_data function can cause incorrect offsets in the column. 

A reproducible case：
```
CREATE TABLE `test_scan_keys_with_bool_type` (
    `col1` tinyint NOT NULL,
    `col2` int NOT NULL,
    `col3` tinyint NOT NULL,
    `col5` boolean REPLACE NOT NULL,
    `col4` datetime(2) REPLACE NOT NULL,
    `col6` double REPLACE_IF_NOT_NULL NULL,
    `col7` varchar(100) REPLACE_IF_NOT_NULL NULL
) ENGINE=OLAP
AGGREGATE KEY(`col1`, `col2`, `col3`)
DISTRIBUTED BY HASH(`col1`, `col2`, `col3`) BUCKETS 1
PROPERTIES (
    "replication_allocation" = "tag.location.default: 1",
    "disable_auto_compaction" = "true"
);

insert into test_scan_keys_with_bool_type values
( -100 ,    1 ,  -82 ,    1 , '2024-02-16 04:37:37.00' , -1299962421.904282 , NULL ),
( -100 ,    0 ,  -82 ,    1 , '2024-02-16 04:37:37.00' , -1299962421.904282 , "hi" ),
( -100 ,    1 ,   92 ,    1 , '2024-02-16 04:37:37.00' ,   23423423.0324234 , NULL );

 insert into test_scan_keys_with_bool_type values
( -100 ,    1 ,  1 ,    1 , '2024-02-16 04:37:37.00' , -1299962421.904282 , "doris" );

MySQL [test]> select * from test_scan_keys_with_bool_type;
+------+------+------+------+------------------------+---------------------+-------+
| col1 | col2 | col3 | col5 | col4                   | col6                | col7  |
+------+------+------+------+------------------------+---------------------+-------+
| -100 |    0 |  -82 |    1 | 2024-02-16 04:37:37.00 | -1299962421.9042821 | hi    |
| -100 |    1 |  -82 |    1 | 2024-02-16 04:37:37.00 | -1299962421.9042821 | NULL  |
| -100 |    1 |    1 |    1 | 2024-02-16 04:37:37.00 | -1299962421.9042821 | hidor |
| -100 |    1 |   92 |    1 | 2024-02-16 04:37:37.00 |    23423423.0324234 | NULL  |
+------+------+------+------+------------------------+---------------------+-------+
[...]

PR #42512, opened 2024-10-27
[fix](agg) Aggregating string types with null values may result in incorrect result when compaction

liaoxin01:
Follow-up for #42067.


liaoxin01:
run buildall

PR #42299, opened 2024-10-22
[opt](cloud) Remove patchelf in start.sh for meta-service

gavinchou:
`patchelf` is not always available in all runtime environments, remove it from start script for ms.

PR #42513, opened 2024-10-27
[opt](cloud) Remove patchelf in start.sh for meta-service (#42299)

gavinchou:
`patchelf` is not always available in all runtime environments, remove it from start script for ms.


gavinchou:
run buildall

PR #42543, opened 2024-10-28
[fix](cases)fix ip case2.1 with sync after streamload

amorynan:
## Proposed changes
fix ip case2.1 with sync after streamload
Issue Number: close #xxx


amorynan:
run buildall

PR #42585, opened 2024-10-28
[Fix](PaimonCatalog) fix the problem that paimon catalog can not access to OSS-HDFS

BePPPower:
Fix the problem that paimon catalog can not access to OSS-HDFS.

There are 2 problems in paimon catalog:
1. Doris FE can not list paimon tables.
    This is because we pass these three properties -- `fs.oss.endpoint / fs.oss.accessKeyId / fs.oss.accessKeySecret` -- to the PaimonCatalog. When PaimonCatalog get these three properties, it will use `OSSLoader` rather than `HadoopFileIOLoader`.

2. Doris BE does not use libhdfs to access OSS-HDFS
    This is because the `tmpLocation` in `LocationPath` does not contain `oss-dls.aliyuncs`. We should use `endpoint` to judge if user wants to access OSS-HDFS

What's more, if you want to access OSS-HDFS with PaimonCatalog, you should:
1. Download Jindo SDK: https://github.com/aliyun/alibabacloud-jindodata/blob/latest/docs/user/zh/jindosdk/jindosdk_download.md
2. copy `jindo-core.jar、jindo-sdk.jar` to `${DORIS_HOME}/fe/lib` and `${DORIS_HOME}/be/lib/java_extensions/preload-extensions` directory.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/common/util/LocationPath.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/util/LocationPath.java
@@ -131,7 +131,19 @@ public class LocationPath {
                 tmpLocation = convertPath ? convertToS3(tmpLocation) : tmpLocation;
                 break;
             case FeConstants.FS_PREFIX_OSS:
-                if (isHdfsOnOssEndpoint(tmpLocation)) {
+                String endpoint = "";
+                if (props.containsKey(OssProperties.ENDPOINT)) {
+                    endpoint = props.get(OssProperties.ENDPOINT);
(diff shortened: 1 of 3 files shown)


BePPPower:
run buildall

morningman:
run buildall

morningman (dismissed):
LGTM

wuwenchi (approved):

CalvinKirs (dismissed):

BePPPower:
run buildall

morningman (approved):

PR #42649, opened 2024-10-28
[improve](load) do not block delta writer if memtable memory is low

kaijchen:
## Proposed changes

Currently when the process memory usage is high, even if load memory usage is low,
the load could be blocked by process memory reach hard limit.
This may cause query memory cannot be released in `INSERT INTO SELECT` statements.

This PR changes this behavior. If the total load memory usage is lower than 5% (current default),
the load will no longer be blocked by hard limit.


kaijchen:
run buildall

liaoxin01 (approved):
LGTM

sollhui (approved):
LGTM

PR #42670, opened 2024-10-29
[fix](schema-change) Make sc job state updated after the table state is changed

Hastyshell:
## Proposed changes

Similar to #39327

Note that schema change job state and table state should be atomic and persistent, but neither is ensured by current implementation. We should optimize this point in the future.


TangSiyang2001:
run buildall

dataroaring (approved):
LGTM

luwei16 (approved):

PR #42609, opened 2024-10-28
[Test](bloom filter) add retry and retry timeout for bloom filter test case

airborne12:
## Proposed changes
fix case error as below
```
Exception in bloom_filter_p0/test_bloom_filter_hit_with_renamed_column.groovy(line 154):

            } else {
                attempt++
                if (attempt < maxRetries) {
                    log.info("profileId is null, retrying after ${waitSeconds} second(s)... (Attempt ${attempt + 1}/${maxRetries})")
                    sleep(waitSeconds * 1000)
                }
            }
        }

        assertTrue(profileId != null, "Failed to retrieve profileId after ${maxRetries} attempts")
^^^^^^^^^^^^^^^^^^^^^^^^^^ERROR LINE^^^^^^^^^^^^^^^^^^^^^^^^^^
        return profileId
    }

    def query = """select C_COMMENT_NEW from ${tableName} where C_COMMENT_NEW='OK'"""
    def profileId = getProfileIdWithRetry(query, 3, 1)
    log.info("profileId:{}", profileId)
    def profileDetail = httpGet("/rest/v1/query_profile/" + profileId)
    log.info("profileDetail:{}", profileDetail)
    assertTrue(profileDetail.contains("BloomFilterFiltered:&nbsp;&nbsp;15.0K&nbsp;&nbsp;(15000)"))


Exception:
org.opentest4j.AssertionFailedError: Failed to retrieve profileId after 3 attempts ==> expected: <true> but was: <false>
```

PR #42672, opened 2024-10-29
[Test](bloom filter) add retry and retry timeout for bloom filter test case #42609

airborne12:
cherry pick from #42609


airborne12:
run buildall

PR #42681, opened 2024-10-29
[test](p0) fix load stream leak in injection cases

kaijchen:
## Proposed changes

Some injection cases (e.g. `VTabletWriterV2._open_streams_to_backend.node_info_null`) will cause LoadStream leak.
Because the injection was disabled too early.

When some sink operators returns "failed" status to the user (regression framework), other sinks are still executing.
Causing partially opened LoadStreams (some instances are not fault-injected, and successfully open the stream).
We should wait a small amount of time to safely assume all instances are done, before disabling the injection.


kaijchen:
run buildall

dataroaring (approved):
LGTM

liaoxin01 (approved):

PR #42686, opened 2024-10-29
[fix](cooldown) use partition's storage policy instead of table's for creating tablets in schema change

DarvenDuan:
## Proposed changes

Issue Number: close #xxx

#### Problem Summary:
In schema change job, Doris creates new tablets and sets the tablets' storage policy id to table's storage policy id, but if we just set the partition's storage policy id which the tablets belong to, the new tablets will be set a wrong storage policy id. So the data write to the new tablets can't be cooldown any more.

#### resolve:
Set new tablets' storage policy id to the partition's storage policy id.


DarvenDuan:
run buildall

PR #42676, opened 2024-10-29
[fix](Cases) fix alias conflict case problem of driver url error

LiBinfeng-01:
change driver url to doris internal one to get driver process more robust
## Proposed changes

Issue Number: close #xxx

PR #42691, opened 2024-10-29
[fix](cases) fix alias conflict case problem of driver url error (#42676)

LiBinfeng-01:
pick: #42676

change driver url to doris internal one to get driver process more robust

## Proposed changes

Issue Number: close #xxx


LiBinfeng-01:
run buildall

PR #42881, opened 2024-10-29
[feature](stream-load) support stream load endpoint redirect policy

liaoxin01:
## Overview
Supports flexible endpoint configuration for Stream Load operations, allowing you to control how load requests are redirected. This is particularly useful in cloud environments or when working with different network segments.

## Redirect Policies
When sending Stream Load requests, you can specify a redirect policy using the `redirect-policy` header. Available policies are:

- `public`: Redirects to public endpoint
- `private`: Redirects to private endpoint
- `direct`: Redirects to a random backend's host
- `""` (empty/default): Uses private endpoint if configured, falls back to random BE

### 1. Single IP Scenario
- BE node has only one IP
- All communications use this IP
- Configuration: No special configuration needed

### 2. VIP Scenario
- BE node keeps its original internal IP for cluster internal communication
- Provides Stream Load service through VIPs:
  - Public VIP: for external access
  - Private VIP: for internal access

### 3. Multi-NIC Scenario
- BE node configured with multiple network cards:
  - Internal NIC: for cluster internal communication and internal access
  - External NIC: for external access

### 1. FE-level Configuration
Set global endpoint configurations through FE configs:

```sql
-- Set public endpoint for all BEs
ADMIN SET FRONTEND CONFIG ('be_public_endpoint' = '1.2.3.4:8011');

-- Set private endpoint for all BEs
ADMIN SET FRONTEND CONFIG ('be_private_endpoint' = '1.2.3.5:8012');
```

### 2. Backend-specific Configuration
Configure endpoints for specific backend nodes:

```sql
-- Add/Modify BE with custom endpoints
ALTER SYSTEM ADD BACKEND 'be_host:heartbeat_port' properties(
    'tag.public_endpoint' = '11.10.10.10:8010',
    'tag.private_endpoint' = '10.10.10.9:8020'
);
```

### 3. Compute Group Configuration
In cloud mode, you can configure endpoints at the compute group level:

```sql
-- Configure compute group endpoints
ALTER SYSTEM ADD BACKEND '' properties(
    'tag.compute_group_name' = 'group_name',
    'tag.compute_group_public_endpoint' = '1.2.3.4:8011',
    'tag.compute_group_private_endpoint' = '1.2.3.5:8012'
);
```

## Usage Example
Here's how to use the redirect policy when sending a Stream Load request:

```bash
curl --location-trusted -u username:password \
    -H "redirect-policy:public" \
    -T data.csv \
    http://fe_host:fe_port/api/db_name/table_name/_stream_load
```

## Priority Order
The endpoint selection follows this priority:
1. Backend-specific configuration
2. Compute group configuration (in cloud mode)
3. FE-level configuration
4. Default BE Host

### DIFF
--- a/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
@@ -3100,9 +3100,15 @@ public class Config extends ConfigBase {
     @ConfField(mutable = true)
     public static int audit_event_log_queue_size = 250000;
 
-    @ConfField(description = {"存算分离模式下streamload导入使用的转发策略, 可选值为public-private或者空",
-            "streamload route policy in cloud mode, availale options are public-private and empty string"})
+    @ConfField(mutable = true, description = {"streamload导入使用的转发策略, 可选值为public-private/public/private/random_be/空",
+            "streamload route policy, availale options are public-private/public/private/random_be and empty string"})
     public static String streamload_redirect_policy = "";
+
+    @ConfField(mutable = true, description = {"be的公网endpoint", "be public endpoint"})
+    public static String be_public_endpoint = "";
+
+    @ConfField(mutable = true, description = {"be的私网endpoint", "be private endpoint"})
+    public static String be_private_endpoint = "";
 
     @ConfField(description = {"存算分离模式下建表是否检查残留recycler key, 默认true",
         "create table in cloud mode, check recycler key remained, default true"})
--- a/fe/fe-core/src/main/java/org/apache/doris/resource/Tag.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/resource/Tag.java
@@ -69,9 +69,12 @@ public class Tag implements Writable {
     public static final String CLOUD_CLUSTER_NAME = "cloud_cluster_name";
     public static final String CLOUD_CLUSTER_ID = "cloud_cluster_id";
     public static final String CLOUD_UNIQUE_ID = "cloud_unique_id";
-    public static final String CLOUD_CLUSTER_PUBLIC_ENDPOINT = "cloud_cluster_public_endpoint";
-    public static final String CLOUD_CLUSTER_PRIVATE_ENDPOINT = "cloud_cluster_private_endpoint";
     public static final String CLOUD_CLUSTER_STATUS = "cloud_cluster_status";
+
+    public static final String COMPUTE_GROUP_PUBLIC_ENDPOINT = "compute_group_public_endpoint";
+    public static final String COMPUTE_GROUP_PRIVATE_ENDPOINT = "compute_group_private_endpoint";
+    public static final String PUBLIC_ENDPOINT = "public_endpoint";
+    public static final String PRIVATE_ENDPOINT = "private_endpoint";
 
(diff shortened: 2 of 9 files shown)


liaoxin01:
run buildall

fe/fe-core/src/main/java/org/apache/doris/resource/Tag.java
```diff
@@ -69,10 +69,13 @@ public class Tag implements Writable {
     public static final String CLOUD_CLUSTER_NAME = "cloud_cluster_name";
     public static final String CLOUD_CLUSTER_ID = "cloud_cluster_id";
     public static final String CLOUD_UNIQUE_ID = "cloud_unique_id";
-    public static final String CLOUD_CLUSTER_PUBLIC_ENDPOINT = "cloud_cluster_public_endpoint";
-    public static final String CLOUD_CLUSTER_PRIVATE_ENDPOINT = "cloud_cluster_private_endpoint";
     public static final String CLOUD_CLUSTER_STATUS = "cloud_cluster_status";
 
+    public static final String COMPUTE_GROUP_PUBLIC_ENDPOINT = "compute_group_public_endpoint";
+    public static final String COMPUTE_GROUP_PRIVATE_ENDPOINT = "compute_group_private_endpoint";
```
gavinchou, line 75: we should do some investigation before changing it
# there are some existed tags named "cloud_cluster_public_endpoint"
# there may be some managements procedure relies on the name "cloud_cluster_public_endpoint"

fe/fe-common/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -3100,10 +3100,16 @@ public static int metaServiceRpcRetryTimes() {
...
+    @ConfField(mutable = true, description = {"streamload导入使用的转发策略, 可选值为public-private/public/private/random_be/空",
+            "streamload route policy, availale options are public-private/public/private/random_be and empty string"})
     public static String streamload_redirect_policy = "";
 
+    @ConfField(mutable = true, description = {"be的公网endpoint", "be public endpoint"})
+    public static String be_public_endpoint = "";
+
+    @ConfField(mutable = true, description = {"be的私网endpoint", "be private endpoint"})
+    public static String be_private_endpoint = "";
+
```
gavinchou, line 3112: Seems confusing. We should alter backend properties("tag.be_public_endpoint" = "xxx" ... ) for each backend to set the endpoints instead of sharing the same endpoint for default behaviors

fe/fe-core/src/main/java/org/apache/doris/resource/Tag.java
liaoxin01: The tag name is not updated here, it only changed a variable name.

fe/fe-common/src/main/java/org/apache/doris/common/Config.java
liaoxin01: In compute-storage integrated scenarios, it's convenient to configure a global VIP to avoid having to modify each time when adding a BE , which could lead to omissions.

liaoxin01:
run buildall

liaoxin01:
run buildall

PR #42699, opened 2024-10-29
[Test](count on index) enhance count on index case retrying for statistics

airborne12:
## Proposed changes

Fix case error as below
```
Exception in inverted_index_p0/test_count_on_index.groovy(line 151):


        sql "sync"
        sql """ set enable_common_expr_pushdown = true """
        sql """set experimental_enable_nereids_planner=true;"""
        sql """set enable_fallback_to_original_planner=false;"""
        sql """analyze table ${testTable_dup} with sync""";
        // wait BE report every partition's row count
        sleep(10000)
        // case1: test duplicate table
        explain {
^^^^^^^^^^^^^^^^^^^^^^^^^^ERROR LINE^^^^^^^^^^^^^^^^^^^^^^^^^^
            sql("select COUNT() from ${testTable_dup} where request match 'GET'")
            contains "pushAggOp=COUNT_ON_INDEX"
        }
        qt_sql(" select COUNT(*) from ${testTable_dup} where request match 'images' ")
        sql " set enable_count_on_index_pushdown=false; "
        qt_sql(" select COUNT(*) from ${testTable_dup} where request match 'images' ")
        sql " set enable_count_on_index_pushdown=true; "

        // case1.1: test duplicate table with null values.
        sql " insert into ${testTable_dup} values(1683964756,null,'GET /images/hm_bg.jpg HTTP/1.0 ',null,null); "

Exception:
java.lang.IllegalStateException: Explain and check failed, expect contains 'pushAggOp=COUNT_ON_INDEX', but actual explain string is:
```

PR #42887, opened 2024-10-30
[Test](count on index) enhance count on index case retrying for statistics #42699

airborne12:
cherry pick from #42699


airborne12:
run buildall

PR #42639, opened 2024-10-28
[test](move memtable) fix move memtable injection case fail

sollhui:
Fix move memtable injection case fail.

PR #42891, opened 2024-10-30
[test](move memtable) fix move memtable injection case fail (#42639)

sollhui:
Fix move memtable injection case fail.

pick (#42639)


sollhui:
run buildall

PR #42897, opened 2024-10-30
[fix](Nerieds) fix FoldConstantRuleOnFE too slow when CaseWhen clause nesting multiple layers

924060929:
## Proposed changes

fix FoldConstantRuleOnFE too slow when CaseWhen clause nesting multiple layers.

### DIFF
--- a/regression-test/suites/nereids_syntax_p0/slow_fold_constant_case_when.groovy
+++ b/regression-test/suites/nereids_syntax_p0/slow_fold_constant_case_when.groovy
@@ -176,8 +176,7 @@ suite("slow_fold_constant_case_when") {
         """
 
 
-    test {
-        sql """
+    def sqlStr = """
                 SELECT `_____` ,
          `c24` ,
          `c28` ,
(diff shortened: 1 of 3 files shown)


924060929:
run buildall

morrySnow (dismissed):

924060929:
run buildall

regression-test/suites/nereids_syntax_p0/slow_fold_constant_case_when.groovy
```diff
@@ -0,0 +1,824 @@
...
+                    AND _____ != '______' ) t
+            GROUP BY  grouping sets((_____) ,())
+            ORDER BY  _____ nulls last limit 3000 ) t
+        ORDER BY  _____ nulls last
+            """
+
+    sql "set enable_fold_constant_by_be=false"
+    test {
+        sql sqlStr
+        time 60000
```
zclllyybb, line 816: is it acceptable in P0 rather than P1?

924060929: It should be fast, so we don't need to move to p1

morrySnow (approved):

starocean999 (approved):

PR #42043, opened 2024-10-17
[fix](cloud) Fix tablet_id is zero when txn lazy commit

swjtu-zhanglei:
* when recycler and ms convert tmp rowsets at the same times, and finish convert task with different partitions severally, may cause tablet_id variable is zero

PR #42905, opened 2024-10-30
[fix](cloud) Fix tablet_id is zero when txn lazy commit (#42043)

swjtu-zhanglei:
* when recycler and ms convert tmp rowsets at the same times, and finish convert task with different partitions severally, may cause tablet_id variable is zero

## Proposed changes

Issue Number: close #xxx


SWJTU-ZhangLei:
pick #42043

SWJTU-ZhangLei:
run buildall

PR #42906, opened 2024-10-30
[fix](hive)fix hive catalog miss partition that have special characters.

hubgeter:
## Proposed changes
Previously, when processing partition values, Hive catalog parsed them in the URL format. However, this is different from the encoding method of Hive, which results in missing some partitions with special characters when reading the partition table. 
This PR is to fix this problem.
Ref: common/src/java/org/apache/hadoop/hive/common/FileUtils.java:`makePartName(List<String> partCols, List<String> vals,String defaultStr)`


hubgeter:
run buildall

hubgeter:
run buildall

hubgeter:
run buildall

hubgeter:
run buildall

morningman (approved):
LGTM

suxiaogang223 (approved):

PR #42907, opened 2024-10-30
[fix](Nerieds) fix FoldConstantRuleOnFE too slow when CaseWhen clause nesting multiple layers (#42897)

924060929:
## Proposed changes
cherry pick from #42897


924060929:
run buildall

PR #20762, opened 2023-06-13
[Enhancement](merge-on-write) optimize contains_agg when calculate delete bitmap

liaoxin01:
## Proposed changes

Issue Number: close #xxx

image

The CPU overhead of `contains_agg` function is relatively high. 

Tpch 7G data, single tablet，56 segments.
After optimization, the time for calculating delete bitmap: 32s -> 16s

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #42919, opened 2024-10-30
[opt](merge-on-write) optimize delete bitmap contains cost while lookup rowkey

zhannngchen:
## Proposed changes

Issue Number: close #xxx

#20762 change `contains_agg` to `contains_agg_without_cache` to avoid AggCache's lock conflict, which would cost lots of CPU on loading large amount of data.

But `contains_agg_without_cache` would cost lost of cpu when base compaction is not so instantly and there accumulates lots of versions delete bitmap, this PR reintroduces the use of `get_agg` to reduce the cost of calls to RoaringBitmap.contains.

Note that in the flame graph quoted in the PR description of #20762, the CPU is mainly consumed in CacheHandle requests and releases, and since the LRUCache needs to count the reference counts of the CacheHandle, write locks are used for each `lookup` and `release` request to the cache, so when loading hundreds of millions of pieces of data, hundreds of millions of `lookup_row_key` and corresponding `lookup` and `release` accesses to `AggCache` will be generated in a short period of time, in which case the lock of `AggCache` will become a bottleneck.

By reducing the frequency of accessing `AggCache`, the issue #20762 can be resolved. This PR avoids accessing the cache every time a rowkey is looked up by calling `get_agg` in advance.

This not only addresses the high `RoaringBitmap::contains` CPU cost caused by traversing all versions of the delete bitmap when many versions accumulate, but also prevents the `AggCache` lock from becoming a bottleneck during large data loading.

### DIFF
--- a/be/src/olap/base_tablet.cpp
+++ b/be/src/olap/base_tablet.cpp
@@ -442,6 +442,7 @@ Status BaseTablet::lookup_row_key(const Slice& encoded_key, TabletSchema* latest
                                   bool with_seq_col,
                                   const std::vector<RowsetSharedPtr>& specified_rowsets,
                                   RowLocation* row_location, uint32_t version,
+                                  DeleteBitmapPtr cur_version_delete_bitmap,
                                   std::vector<std::unique_ptr<SegmentCacheHandle>>& segment_caches,
                                   RowsetSharedPtr* rowset, bool with_rowid,
                                   std::string* encoded_seq_value, OlapReaderStatistics* stats) {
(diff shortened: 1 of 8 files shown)


zhannngchen:
run buildall

zhannngchen:
run buildall

zhannngchen:
run buildall

zhannngchen:
run buildall

zhannngchen:
run buildall

dataroaring:
run buildall

PR #42948, opened 2024-10-30
[fix](jdbc catalog) fix be crash when create jdbc catalog without driver jar

zy-kkk:
When creating a JDBC Catalog, a connection test is performed. However, if the JDBC Driver path does not exist, we should report an error that the file cannot be opened and not clean up the resources, because the resources have not been created. Cleaning up non-existent resources will cause the program to crash. This PR is changed to clean up only when the resources are created.


zy-kkk:
run buildall

morningman (approved):
LGTM

hubgeter (approved):

PR #42971, opened 2024-10-31
[fix](jdbc catalog) fix be crash when create jdbc catalog without driver jar #42948

morningman:
cherry pick from #42948


morningman:
run buildall

PR #42937, opened 2024-10-30
[Test](bloom filter) add retry query for bloom filter test case

airborne12:
## Proposed changes

add query into retry logic for bloom filter test case

PR #42974, opened 2024-10-31
[Test](bloom filter) add retry query for bloom filter test case #42937

airborne12:
cherry pick from #42937


airborne12:
run buildall

PR #42984, opened 2024-10-31
[chore](restore) Save restore stmt to the resp if RestoreSnapshot failed #42933

w41ter:
cherry pick from #42933


w41ter:
run buildall

PR #43009, opened 2024-10-31
[improvement](statistics)External table getRowCount return -1 when row count is not available or row count is 0.

Jibing-Li:
External table getRowCount return -1 when row count is not available or row count is 0. So the behavior of external table could match with internal olap table.


Jibing-Li:
run external

Jibing-Li:
run external

Jibing-Li:
run external

morrySnow (approved):

xzj7019 (approved):

PR #43010, opened 2024-10-31
[test](mv) Insert into more data when first insert into to make sure using sync mv

seawinde:
## Proposed changes

**Root Cause Analysis:**
Currently, the statistics reported by BE (Backend) nodes have higher priority than those from ANALYZE statements. During the first INSERT INTO operation, the system waits for row count reports from all tablets before updating the table statistics.
Subsequent INSERT INTO operations cannot obtain the status of all tablets, so the system continues to use the statistical information from the first INSERT INTO operation. This leads to a lower estimated cost for the original table's query plan, resulting in the selection of the original table's query plan instead of the materialized view.

**Conclusion:**
The test case should be modified to include a larger dataset in the first INSERT INTO operation, which will increase the likelihood of utilizing the materialized view. This is because the cost estimation will better reflect the actual data distribution and size, leading to more accurate plan selection.


seawinde:
run buildall

morrySnow (approved):

englefly (approved):

PR #43017, opened 2024-10-31
[extension](kettle)Adding checks

qsy840035156:
Proposed changes
add kettle plugins to write doris
Implementation principle:

1.adding checks


extension/kettle/impl/pom.xml
```diff
@@ -110,6 +110,11 @@ under the License.
       <artifactId>log4j</artifactId>
       <version>1.2.17</version>
     </dependency>
+      <dependency>
+          <groupId>org.json</groupId>
+          <artifactId>json</artifactId>
+          <version>20240303</version>
+      </dependency>
```
JNSimba, line 117: Is it possible to use Jackson directly, so that there is no need to introduce a new pom?

extension/kettle/impl/src/main/java/org/pentaho/di/trans/steps/dorisstreamloader/DorisStreamLoaderMeta.java
```diff
@@ -235,10 +244,101 @@ public void getFields( RowMetaInterface rowMeta, String origin, RowMetaInterface
...
       String[] input, String[] output, RowMetaInterface info, VariableSpace space, Repository repository,
       IMetaStore metaStore ) {
     //todo: check parameters
+
+    //check fenodes  user password
+    if (null == fenodes || fenodes.trim().isEmpty()){
+        remarks.add(new CheckResult(CheckResult.TYPE_RESULT_NONE, BaseMessages.getString( PKG,
+            "DorisStreamLoaderMeta.CheckResult.FenodesIsNUll" ), stepMeta));
+    }else {
+        if (checkLogin()){
```
JNSimba, line 258: why we need to checkLogin？

PR #43037, opened 2024-10-31
[fix](S3FileWriter) Fix boundary issue when multipart upload

gavinchou:
When the file data size is a multiple of config::s3_write_buffer_size, number of parts may exceed the actual number of parts that need to be uploaded. This is because it is incremented by 1 in advance within the S3FileWriter::appendv method.


gavinchou:
run buildall

dataroaring (approved):
LGTM

Yukang-Lian (approved):

PR #40113, opened 2024-08-29
[feat](Authenticator) Pluginization of Authenticator

CalvinKirs:
## abstract
introduces the pluginization of the Authenticator component, enabling greater flexibility and modularity within the authentication system. By refactoring the Authenticator into a pluggable SPI (Service Provider Interface), we allow for custom authentication mechanisms to be seamlessly integrated and managed.

## Key Changes
AuthenticatorFactory Interface:

This interface defines the create method, which is used to create an Authenticator instance based on initialization properties, enabling support for various authentication mechanisms.
The factoryIdentifier method provides an identifier for the factory, allowing differentiation between various AuthenticatorFactory implementations (e.g., LDAP, default).
Implemented Specific Authentication Factories:

We have implemented factories for different authentication mechanisms (such as LDAP), providing the necessary configuration support for each.

## eg

In your Maven pom.xml, add the fe-core dependency:

```
        <dependency>
            <groupId>org.apache.doris</groupId>
            <artifactId>fe-core</artifactId>
            <version>1.2-SNAPSHOT</version>
            <scope>provided</scope>
        </dependency>

```

```
import org.apache.doris.common.ConfigBase;

public class LocalAuthenticatorFactory implements AuthenticatorFactory {


    @Override
    public LocalAuthenticator create(Properties initProps) {
        return new LocalAuthenticator();
    }

    @Override
    public String factoryIdentifier() {
        return "local";
    }
}
import java.util.HashMap;
import java.util.Map;

public class LocalAuthenticator implements Authenticator {
    private Map<String, String> userStore = new HashMap<>();

    public LocalAuthenticator() {
        // Hardcoded usernames and passwords
        userStore.put("user1", "password1");
        userStore.put("user2", "password2");
        userStore.put("admin", "adminpass");
    }

    @Override
    public boolean authenticate(String username, String password) {
        // Authenticate by checking the hardcoded user store
        String storedPassword = userStore.get(username);
        return storedPassword != null && storedPassword.equals(password);
    }
}




```
To integrate with the SPI (Service Provider Interface) mechanism, we created a file at **src/main/resources/META-INF/services/org.apache.doris.mysql.authenticate.AuthenticatorFactory** containing **org.example.LocalAuthenticatorFactory,** which allows the system to dynamically load and utilize the LocalAuthenticatorFactory implementation.

[...]

PR #43057, opened 2024-11-01
[Fix](Authenticator)defaultAuthenticator not init

CalvinKirs:
The `defaultAuthenticator` is not initialized. When a configured authentication plugin implements the `canDea`l method and returns `false`, the `defaultAuthenticator` will be used, resulting in a login failure. The client exception message is `RROR 2013 (HY000): Lost connection to MySQL server at 'reading authorization packet', system error: 2`. This issue only affects users who use non-built-in authentication plugins that implement `canDeal` and return false under certain circumstances.

https://github.com/apache/doris/pull/40113 introduced this issue


CalvinKirs:
run buildall

morningman (approved):

wuwenchi (approved):

PR #43059, opened 2024-11-01
[fix](jdbc catalog) Optimize JDBC Connection Closing to Ensure ProperResource Release

zy-kkk:
This pull request optimizes the JDBC connection and resource management in `JdbcClient` and related methods. The main changes are as follows:

- **Manual Resource Closing**: Explicitly place resource acquisition and closing within a `try-catch-finally` structure. In the `finally` block, manually close resources like `Connection`, `Statement`, and `ResultSet` to ensure that resources are properly released even if an exception occurs.

- **Remove try-with-resources**: All usages of `try-with-resources` have been removed, switching to manual resource management instead.

- **Improve Exception Handling**: When closing resources, if an exception occurs, log a warning but do not throw a new exception. This avoids masking the original exception during the resource release phase.

- **Unified Resource Management**: Ensure consistent resource acquisition and closing methods across all functions, improving code readability and maintainability.

Functions modified include but are not limited to:

- `executeStmt`
- `getColumnsFromQuery`
- `getDatabaseNameList`
- `getJdbcColumnsInfo`
- `processTable`
- `getColumnsDataTypeUseQuery`

These changes enhance the robustness of the code, prevent potential resource leaks, and ensure that all JDBC connections and resources are properly closed without using `try-with-resources`.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/datasource/jdbc/client/JdbcMySQLClient.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/datasource/jdbc/client/JdbcMySQLClient.java
@@ -76,10 +76,11 @@ public class JdbcMySQLClient extends JdbcClient {
 
     @Override
     public List<String> getDatabaseNameList() {
-        Connection conn = getConnection();
+        Connection conn = null;
         ResultSet rs = null;
         List<String> remoteDatabaseNames = Lists.newArrayList();
         try {
+            conn = getConnection();
             if (isOnlySpecifiedDatabase && includeDatabaseMap.isEmpty() && excludeDatabaseMap.isEmpty()) {
                 String currentDatabase = conn.getCatalog();
                 remoteDatabaseNames.add(currentDatabase);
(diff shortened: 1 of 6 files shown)


zy-kkk:
run buildall

fe/fe-core/src/main/java/org/apache/doris/datasource/jdbc/client/JdbcMySQLClient.java
```diff
@@ -294,30 +296,33 @@ private boolean isConvertDatetimeToNull(JdbcClientConfig jdbcClientConfig) {
...
             }
         } catch (SQLException e) {
             throw new JdbcClientException("failed to get jdbc columns info for remote table `%s.%s`: %s",
                     remoteDbName, remoteTableName, Util.getRootCauseMessage(e));
         } finally {
-            close(resultSet, conn);
+            close(resultSet, stmt, conn);
         }
-        return fieldtoType;
+        return fieldToType;
```
CalvinKirs, line 325: Using try-with-resources is preferable; otherwise, you have to be extra careful about potential null pointers. In a finally block, you’d need to check if each item is null before closing it,

```
finally {
    // Null check before closing each resource
    if (resultSet != null) {
        try {
            resultSet.close();
        } catch (SQLException e) {
            LOG.warn("Failed to close ResultSet", e);
        }
    }
    if (stmt != null) {
        try {
            stmt.close();
        } catch (SQLException e) {
            LOG.warn("Failed to close Statement", e);
        }
    }
    if (conn != null) {
        try {
            conn.close();
        } catch (SQLException e) {
            LOG.warn("Failed to close Connection", e);
        }
    }
}
```

 which makes the code messy.

```
try (Connection conn = getConnection();
     Statement stmt = conn.createStatement();
     ResultSet resultSet = stmt.executeQuery(queryBuf.toString())) {

    while (resultSet.next()) {

        String fieldName = resultSet.getString("Field");
        String typeName = resultSet.getString("Type");
        fieldtoType.put(fieldName, typeName);
    }
} catch (SQLException e) {
    throw new JdbcClientException(String.format("failed to get jdbc columns info for remote table `%s.%s`: %s",
            remoteDbName, remoteTableName, Util.getRootCauseMessage(e)), e);
}
```

zy-kkk: I checked whether the object is null in the close() method, and because some drivers do not implement AutoCloseable, we need to close it manually.

zy-kkk:
run buildall

morningman (approved):
LGTM

wuwenchi (approved):

PR #41667, opened 2024-10-10
[Bug](runtime-filter) send ignored rf when hash join build closed early

BiteTheDDDDt:
## Proposed changes
send ignored rf when hash join build closed early to avoid runtime filter sync/merge error
Follow-up : https://github.com/apache/doris/pull/41292

PR #43078, opened 2024-11-01
[Bug](runtime-filter) avoid wrong partitial ignore minmax filter

BiteTheDDDDt:
In #https://github.com/apache/doris/pull/41667 we support ignore filter partitial, however, when sync_filter_size is turned off, the final filter may only contain part of the data because some filters are judged to be 'in'.
For example:
there are rf001(2 instances, in_or_bloom),rf000(2 instances, min_max),
instance_1 has 1e8 row and rf001 change to bloom, the rf000 will not ignored
instance_2 has 1 row and rf001 change to in, the rf000 will ignored
finally, rf000 applied and make wrong result

### DIFF
--- a/be/src/exprs/runtime_filter_slots.h
+++ b/be/src/exprs/runtime_filter_slots.h
@@ -77,6 +77,10 @@ public:
             if (filter->get_real_type() != RuntimeFilterType::IN_FILTER) {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


BiteTheDDDDt:
run buildall

zhangstar333 (approved):

Gabriel39 (approved):

PR #43080, opened 2024-11-01
[fix](mtmv)Fix the problem where the job does not exist, which prevents the deletion of MTMV

zddr:
Fix the problem where the job does not exist, which prevents the deletion of MTMV

The fundamental reason is that deleting materialized views is not guaranteed by an independent editlog. Here, let's first ensure that materialized views can be deleted normally


zddr:
run buildall

morrySnow (approved):

CalvinKirs (approved):

PR #43172, opened 2024-11-04
[fix](restore) Add synchronized to avoid concurrent modification

w41ter:
Without synchronized protection, when the user manually executes cancel, the backupHandler may also trigger cancel simultaneously. Modifying the snapshotInfo by both at the same time will cause a concurrent modification exception. 

The pending state will synchronously wait for the task of creating a replica to end. Therefore, a creating state needs to be added to convert the synchronous wait into asynchronous polling to avoid blocking the user's cancel/list job info requests.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/backup/RestoreJob.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/backup/RestoreJob.java
@@ -445,7 +445,7 @@ public class RestoreJob extends AbstractJob implements GsonPostProcessable {
         checkIfNeedCancel();
 
         if (status.ok()) {
-            if ((state != RestoreJobState.PENDING || state != RestoreJobState.CREATING)
+            if (state != RestoreJobState.PENDING && state != RestoreJobState.CREATING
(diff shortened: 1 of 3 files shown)


w41ter:
run buildall

w41ter:
run buildall

fe/fe-core/src/main/java/org/apache/doris/backup/RestoreJob.java
```diff
@@ -132,7 +131,8 @@ public class RestoreJob extends AbstractJob implements GsonPostProcessable {
     public enum RestoreJobState {
         PENDING, // Job is newly created. Check and prepare meta in catalog. Create replica if necessary.
                  // Waiting for replica creation finished synchronously, then sending snapshot tasks.
-                 // then transfer to SNAPSHOTING
+                 // then transfer to PENDING.
```
dataroaring, line 134:
```suggestion
                 // then transfer to CREATING.
```

dataroaring (approved):
LGTM

dataroaring (dismissed):
LGTM

w41ter:
run buildall

Yukang-Lian (approved):

dataroaring (approved):
LGTM

PR #43185, opened 2024-11-04
[fix](orc) fix predicate filter failed when use hive 1.x version

fantasy12345zsq:
Proposed changes
 1. original issue:
     we found presto scan byte much more bigger then Doris, such as query:
     select l_orderkey from dev.lineitem_orc_100g_backup1 where l_orderkey = 7192579;
    this query just have 1 rows result, scan bytes presto vs Doris:
    presto:
image
    Doris:
image
image
  2. root cause:
     we use hive 1.x version,query predicate column name not match orc column name, in this case: 
     schame is l_orderkey name, in orc field name is _col1。
3. how to fix:
    in _init_search_argument change l_orderkey -> _col1。
4. profile after fix:
    query latency:
image
    scan bytes:
image

### DIFF
--- a/be/src/vec/exec/format/orc/vorc_reader.cpp
+++ b/be/src/vec/exec/format/orc/vorc_reader.cpp
@@ -734,6 +734,11 @@ bool OrcReader::_init_search_argument(
     if (predicates.empty()) {
         return false;
     }
+
+    if (_is_hive1_orc_or_use_idx) {
+        for (OrcPredicate& it : predicates) it.col_name = _col_name_to_file_col_name[it.col_name];
+    }
+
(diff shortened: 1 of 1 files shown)


suxiaogang223:
Hello, thank you for your reply. We have reconstructed the logic of orc's pushdown predicate construction in this pr #43255 , and used _col_name_to_file_col_name_low_case to find the corresponding column name in the file when constructing OrcLiteral. You can use the latest code of the master branch to check whether the problem exists :)

suxiaogang223:
@fantasy12345zsq  This PR will be picked to the branch-2.1 because the branch-2.1 has not yet used the new orc predicate pushdown code. Thanks again for the contribution :)

PR #42245, opened 2024-10-22
[fix](compaction) Filter out all not running tablets if cumu compaction for shadow tablets is disabled

Hastyshell:
## Proposed changes

As title. If cumu compaction for shadow tablets is disabled, shadow tablets may be picked to do cumu compaction and result in error status.

PR #43209, opened 2024-11-04
[fix](compaction) Filter out all not running tablets if cumu compaction for shadow tablets is disabled (#42245)

Hastyshell:
## Proposed changes

pick: #42245

As title. If cumu compaction for shadow tablets is disabled, shadow tablets may be picked to do cumu compaction and result in error status.


TangSiyang2001:
run buildall

PR #43210, opened 2024-11-04
[fix](cloud) fix read cache block file return stat NOT_FOUND

freemandealer:
Previously, cache blocks could be downloaded as different types than the target type we intended to read, leading to false cache misses. E.g., a block might be downloaded as an 'idx' type when the current context expected a 'ttl' type.

The root cause of this problem is the original design encoding meta info such as type and expiration time into cache block file path and readers of this cache block file have inconsistent view of the type so they use different name to locate file and run into NOT_FOUND error in the end.

This commit tries other type if the initial type failed to locate the file. Be ware that this is only a nasty quick fix. We will elimite the metadate encoded in the file path in the near future to get rid of all the path related problems.

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Check List (For Committer)

- Test 

    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:

    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- Release note

    
    
    None

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/be/src/common/config.cpp
+++ b/be/src/common/config.cpp
@@ -1035,8 +1035,6 @@ DEFINE_mInt64(file_cache_ttl_valid_check_interval_second, "0"); // zero for not 
 // If true, evict the ttl cache using LRU when full.
 // Otherwise, only expiration can evict ttl and new data won't add to cache when full.
 DEFINE_Bool(enable_ttl_cache_evict_using_lru, "true");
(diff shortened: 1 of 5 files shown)


freemandealer:
duplicated

PR #42930, opened 2024-10-30
[Refactor](BE) Uniform Expr, Block, Columns size-relative interface (Part I)

zclllyybb:
## Proposed changes

Issue Number: close #xxx

In this pr we refactored some base interface to avoid narrow cast with data consistent kept.

We uniformed these type of numbers:
```
columns in Block        - U32 (just use no more than U16)
Rows in Block           - U64 (others U32, ColumnString64 U64)
column_id in expr arg   - I32 (because -1 is meaningful)
child of Expr           - U16 (FE use int32, have a restriction of 10000. adjustable but no guarantee for result)
```

So, for functions' argument column id, we use `uint32` to replace `size_t`.
For rows, we use `size_t` to replace some `int`.
For child of expr, we cast it as `uint_16` to use, rather than `size_t` as `vector::size()`

Refactor of table functions: https://github.com/apache/doris/pull/43218

PR #43218, opened 2024-11-04
[Refactor](BE) Uniform Expr, Block, Columns size-relative interface (Part II)

zclllyybb:
part 2 of https://github.com/apache/doris/pull/42930

two modifications:
1. when parsing data of json array, cast to target type when parse int data no matter what source type is.
2. in explode_numbers, the argument column type is certain Int32. so reduce some virtual function call.


zclllyybb:
run buildall

Mryange (approved):

HappenLee (approved):
LGTM

PR #43220, opened 2024-11-04
[fix](cloud) fix read cache block file return stat NOT_FOUND

freemandealer:
Previously, cache blocks could be downloaded as different types than the target type we intended to read, leading to false cache misses. E.g., a block might be downloaded as an 'idx' type when the current context expected a 'ttl' type.

The root cause of this problem is the original design encoding meta info such as type and expiration time into cache block file path and readers of this cache block file have inconsistent view of the type so they use different name to locate file and run into error in the end.

This commit tries other type if the initial type failed to locate the file (return NOT_FOUND). Be ware that this is a nasty quick fix. We will elimite the metadate encoded in the file path in the near future to get rid of all the path related problems.

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Check List (For Committer)

- Test 

    - [ ] Regression test
    - [ ] Unit Test
        - [x] This is a refactor/code format and no logic has been changed.
        - [ ] Other reason 

- Behavior changed:

    - [x] No.
    - [ ] Yes. 

    - [x] No.
    - [ ] Yes. 

- Release note

    
    
    None

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label


be/src/io/cache/fs_file_cache_storage.cpp
```diff
@@ -270,6 +276,18 @@ std::string FSFileCacheStorage::get_path_in_local_cache_old_ttl_format(const std
...
 
+std::vector<std::string> FSFileCacheStorage::get_path_in_local_cache_all_candidates(
+        const std::string& dir, size_t offset) {
+    std::vector<std::string> candidates;
+    std::string base = get_path_in_local_cache(dir, offset, FileCacheType::NORMAL);
+    candidates.push_back(base);
+    candidates.push_back(base + "_idx");
+    candidates.push_back(base + "_ttl");
+    candidates.push_back(base + "_disposable");
+    candidates.push_back(base + "_tmp");
```
gavinchou, line 287: _tmp cannot be a candidate, it is may be not fully downloaded

freemandealer:
run buildall

freemandealer:
run external

gavinchou (approved):

dataroaring (approved):
LGTM

PR #43244, opened 2024-11-05
[fix](statistics)Change column statistics table varchar length to 1024.

Jibing-Li:
Change column statistics table varchar length to 1024. Including db name, table name, column name and partition name. Because the maximum length of the names above is 256 characters. (For 256 Chinese character, we need varchar 1024 to store it)


Jibing-Li:
run buildall

morrySnow (approved):

zfr9527 (approved):
LGTM

PR #43261, opened 2024-11-05
[fix](cloud-mow) Should clear GetDeleteBitmapResponse when geting delete bitmap fail and retry

hust-hhb:
When geting  delete bitmap fail, it will retry, howerve the GetDeleteBitmapResponse doesn't clear now, which will lead to the delete bitmap data  of last request residual in GetDeleteBitmapResponse and return to be, it a wrong delete bitmap data and may cause be core.


hust-hhb:
run buildall

dataroaring (approved):
LGTM

zhannngchen (approved):
LGTM

PR #39597, opened 2024-08-20
[enhance](privilege)Grant check name

zddr:
grant stmt check ctl/db/table if exist

PR #43275, opened 2024-11-05
[fix](auth)Fix after fe restarting, external permissions are lost

zddr:
cause by: https://github.com/apache/doris/pull/39597

This PR checks whether the authorized tableName exists. After restarting, when replaying the log, the metadata of the external cannot be obtained, so the replay will report an error, but it does not affect the startup, only printing the log.

Solution: No longer verify during playback


zddr:
run buildall

morrySnow (approved):

zddr:
run buildall

zddr:
run p0

seawinde (approved):

PR #43279, opened 2024-11-05
[improvement](statistics)Remove useless stats validation check.

Jibing-Li:
Remove useless stats validation check.
Before, we will not load the column stats when table rowCount > 0 and ndv == 0 and nullCount != rowCount. This is to avoid using invalid stats.
Now, we remove this validation because the planner side added validation to the column stats, (see https://github.com/apache/doris/pull/41790). Besides, after remove of the validation, it is easier to add regression test using stats injection.


Jibing-Li:
run p0

morrySnow (approved):

englefly (approved):

PR #43250, opened 2024-11-05
[fix](export) make sure the export task being executed after writing "create export job" edit log

morningman:
### What problem does this PR solve?

Problem Summary:

Fix bug like:
```
java.lang.NullPointerException: Cannot invoke "org.apache.doris.load.ExportJob.replayExportJobState(org.apache.doris.load.ExportJobState)" because "job" is null
        at org.apache.doris.load.ExportMgr.replayUpdateJobState(ExportMgr.java:475) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.persist.EditLog.loadJournal(EditLog.java:390) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.catalog.Env.replayJournal(Env.java:2965) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.master.Checkpoint.doCheckpoint(Checkpoint.java:135) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.master.Checkpoint.runAfterCatalogReady(Checkpoint.java:80) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.common.util.MasterDaemon.runOneCycle(MasterDaemon.java:58) ~[doris-fe.jar:1.2-SNAPSHOT]
        at org.apache.doris.common.util.Daemon.run(Daemon.java:119) ~[doris-fe.jar:1.2-SNAPSHOT]
```

This is because the export job is finished before adding it.
The "finish" log is saved before "creating" log, so when replaying the "finish" log,
the export job can not be found.

This PR change the order of "writing create export job log" and `adding export task` event,
to avoid this error

### Check List (For Committer)

- Test 

    - [ ] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [x] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No colde files have been changed.
        - [ ] Other reason 

- Behavior changed:

    - [x] No.
    - [ ] Yes. 

- Does this need documentation?

    - [x] No.
    - [ ] Yes. 

- Release note

    
    
    None

### Check List (For Reviewer who merge this PR)

- [x] Confirm the release note
- [x] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #43291, opened 2024-11-05
[fix](export) make sure the export task being executed after writing "create export job" edit log #43250

morningman:
cherry-pick #43250


morningman:
run buildall

PR #43288, opened 2024-11-05
[fix](regression) test_primary_key_partial_update fail due to the U.S. change from DST to standard time

zhannngchen:
### What problem does this PR solve?

The case used time zone `America/New_York`, and the  result of `TIMESTAMPDIFF` will change when U.S. switch between DST(daylight saving time) and standard time, we should use a time zone not sensitive to this.

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Check List (For Committer)

- Test 

    - [x] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No colde files have been changed.
        - [ ] Other reason 

- Behavior changed:

    - [x] No.
    - [ ] Yes. 

- Does this need documentation?

    - [x] No.
    - [ ] Yes. 

- Release note

    
    
    None

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [x] Add branch pick label

PR #43299, opened 2024-11-06
[fix](regression) test_primary_key_partial_update fail due to the U.S. change from DST to standard time (#43288)

zhannngchen:
cherry-pick #43288


zhannngchen:
run buildall

PR #42262, opened 2024-10-22
[fix](auditlog) add missing audit log fields and duplicate audit log error

morningman:
### What problem does this PR solve?
Problem Summary:

#### Issue 1
There are some fields that is missing in audit log table.
This PR add them all:
- shuffle_send_rows
- shuffle_send_bytes
- scan_bytes_from_local_storage
- scan_bytes_from_remote_storage
- is_nereids
- compute_group

Notice that `compute_group` is previously name `cloudClusterName` in fe.audit.log,
which is incorrect, so I change it to the right name.

After this PR, all these fields will be saved in both audit log table and fe.audit.log

#### Issue 2
The `AuditEventBuilder` need to be reset at each run, the there will be duplicate audit log.

#### Issue 3
Add a new statement `call flush_audit_log()`. It will flush the audit log immediately to audit_log table.
This is useful in test case, so that we don't need to wait 1min to flush the audit log data.

### Check List (For Committer)

- Test 

    - [x] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No colde files have been changed.
        - [ ] Other reason 

- Behavior changed:

    - [ ] No.
    - [x] Yes. `cloudClusterName` in fe.audit.log change to `computeGroup`. And some fields are added to the audit log table.

- Does this need documentation?

    - [ ] No.
    - [x] Yes. I will add this later

- Release note

    
    
    [fix](auditlog) add missing audit log fields and duplicate audit log error

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #43303, opened 2024-11-06
[fix](auditlog) add missing audit log fields and duplicate audit log error #42262

morningman:
cherry pick from #42262


morningman:
run buildall

PR #43309, opened 2024-11-06
[Fix](PaimonCatalog) fix the problem that paimon catalog can not access to OSS-HDFS #42585

morningman:
cherry pick from #42585


morningman:
run buildall

PR #43313, opened 2024-11-06
[improvement](jdbc catalog) Disallow non-constant type conversion pushdown and implicit conversion pushdown #42102

morningman:
cherry pick from #42102


morningman:
run buildall

PR #43196, opened 2024-11-04
[fix](test) fix clickhouse jdbc catalog func push down case

zy-kkk:
### What problem does this PR solve?

Issue Number: close #xxx

Related PR: #42102

Problem Summary:

Because we previously disabled the pushdown of cast, a test case error occurred when checking the pushdown of the clickhosue function. Because the column type is largeint, a cast will be generated when using from_unixtime, causing the function to fail to be pushed down. So we changed the column type to bigint.

### Check List (For Committer)

- Test 

    - [x] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No colde files have been changed.
        - [ ] Other reason 

- Behavior changed:

    - [x] No.
    - [ ] Yes. 

- Does this need documentation?

    - [x] No.
    - [ ] Yes. 

- Release note

    
    
    None

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [x] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #43315, opened 2024-11-06
[fix](test) fix clickhouse jdbc catalog func push down case #43196

morningman:
cherry pick from #43196


morningman:
run buildall

PR #43318, opened 2024-11-06
[fix](hive)fix hive catalog miss partition that have special characters. #42906

morningman:
cherry pick from #42906


morningman:
run buildall

PR #43287, opened 2024-11-05
[Test](inverted index) add test data for inverted index p2 case

airborne12:
### What problem does this PR solve?

Problem Summary:
For run p2 case of inverted index

### Check List (For Committer)

- Test 

    - [x] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No colde files have been changed.
        - [ ] Other reason 

- Behavior changed:

    - [ ] No.
    - [ ] Yes. 

- Does this need documentation?

    - [ ] No.
    - [ ] Yes. 

- Release note

    
    
    None

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [x] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #43387, opened 2024-11-07
[Test](inverted index) add test data for inverted index p2 case #43287

airborne12:
cherry pick from #43287


airborne12:
run buildall

PR #43469, opened 2024-11-07
[feature](hive)support hive catalog read json table. 

hubgeter:

Problem Summary:
Support reading json format hive table like:
```mysql
mysql> show create table basic_json_table;
CREATE TABLE `basic_json_table`(
  `id` int,
  `name` string,
  `age` tinyint,
  `salary` float,
  `is_active` boolean,
  `join_date` date,
  `last_login` timestamp,
  `height` double,
  `profile` binary,
  `rating` decimal(10,2))
ROW FORMAT SERDE
  'org.apache.hive.hcatalog.data.JsonSerDe'
```

Behavior changed:
To implement this feature, this pr modifies `new_json_reader`. Previously, `new_json_reader` could only insert data into columnString. In order to support inserting data into columns of other types, `DataTypeSerDe` is introduced to insert data into columns. To maintain compatibility with previous versions, changes to this pr are triggered only when reading hive json tables.

Limitation of Use:
1. Currently, only query is supported, and writing is not supported.
2. Currently, only the `ROW FORMAT SERDE 'org.apache.hive.hcatalog.data.JsonSerDe';` scenario is supported. For some properties specified in `with serdeproperties`, Doris does not take effect.
3. Since Hive does not allow columns with the same name but different case when creating a table in Json format (including inside a Struct), we convert the field names in the Json data to lowercase when reading the Json data file, and then match according to the lowercase field names. For field names that are duplicated after being converted to lowercase in the data, the value of the last field is used (consistent with Hive behavior).
example:
```
create table json_table(
    column int 
)ROW FORMAT SERDE 'org.apache.hive.hcatalog.data.JsonSerDe';

a.json:
{"column":1,"COLumn",2,"COLUMN":3}
{"column":10,"COLumn",20}
{"column":100}
in Hive : load a.json to table json_table

in Doris query:
---
3
20
100
---
```

Todo(in next pr):
Merge `serde` and `json_reader` ,because they have logical conflicts.

### Release note
Hive catalog support read json format table.

### Check List (For Author)

- Test 
    - [x] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [x] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label


hubgeter:
run buildall

be/src/vec/data_types/serde/data_type_serde.h
```diff
@@ -332,6 +336,11 @@ class DataTypeSerDe {
                                           Arena& mem_pool, int row_num) const;
     virtual Status read_one_cell_from_json(IColumn& column, const rapidjson::Value& result) const;
 
+    virtual DataTypeSerDeSPtrs get_nested_serdes() const {
```
hubgeter, line 339: maybe we can support 
`get_key_serde()` `get_value_serde()` for map     
`get_data_serde()` for array  nullcolumn
`get_serde(int)` for struct 
not  `get_nested_serdes()` for any complex type

hubgeter:
run buildall

be/src/vec/exec/format/json/new_json_reader.cpp
```diff
@@ -756,7 +770,23 @@ Status NewJsonReader::_set_column_value(rapidjson::Value& objectValue, Block& bl
         _append_empty_skip_bitmap_value(block, cur_row_count);
     }
 
-    for (auto* slot_desc : slot_descs) {
+    if (_is_hive_table) {
+        //don't like _fuzzy_parse,each line read in must modify name_map once.
+        for (auto* v : slot_descs) {
```
morningman, line 775: I think it will seriously impact the performance.
What if we set a session variable to control it? And by default, in most cases, there is no duplicate column name in json value.

be/src/vec/exec/format/json/new_json_reader.cpp
```diff
@@ -840,79 +871,145 @@ Status NewJsonReader::_set_column_value(rapidjson::Value& objectValue, Block& bl
...
-    const char* str_value = nullptr;
-    char tmp_buf[128] = {0};
-    int32_t wbytes = 0;
-    std::string json_str;
-
     ColumnNullable* nullable_column = nullptr;
-    if (slot_desc->is_nullable()) {
+    vectorized::IColumn* data_column_ptr = column_ptr;
+    DataTypeSerDeSPtr data_serde = serde;
+    vectorized::DataTypeSerDe::FormatOptions _options;
```
morningman, line 881: Looks like this `_options` can be created only once at reuse it in each call.

be/src/vec/exec/format/json/new_json_reader.cpp
```diff
@@ -840,79 +871,145 @@ Status NewJsonReader::_set_column_value(rapidjson::Value& objectValue, Block& bl
...
-        wbytes = 1;
-        str_value = (char*)"1";
-        break;
-    case rapidjson::Type::kNullType:
-        if (!slot_desc->is_nullable()) {
-            RETURN_IF_ERROR(_append_error_msg(
-                    *value, "Json value is null, but the column `{}` is not nullable.",
-                    slot_desc->col_name(), valid));
-            return Status::OK();
+    } else if (type_desc.type == TYPE_STRUCT) {
```
morningman, line 924: Can we use `switch ... case` for following logic?

be/src/vec/exec/format/json/new_json_reader.cpp
```diff
@@ -840,79 +871,145 @@ Status NewJsonReader::_set_column_value(rapidjson::Value& objectValue, Block& bl
...
-            wbytes = end - tmp_buf;
+    if (_is_load || !type_desc.is_complex_type()) {
+        if (value->IsString()) {
+            Slice slice {value->GetString(), value->GetStringLength()};
+            RETURN_IF_ERROR(
+                    data_serde->deserialize_one_cell_from_json(*data_column_ptr, slice, _options));
+
         } else {
-            return Status::InternalError<false>("It should not here.");
+            // Maybe we can `switch (value->GetType()) case: kNumberType`.
```
morningman, line 917: Why not using `switch (value->GetType())`?

be/src/vec/exec/format/json/new_json_reader.cpp
hubgeter: Here we not only need to deal with the same column names, but also the order of the table column names and the json data. I think I can simplify this double for loop by using a layer of for loop plus map.

be/src/vec/exec/format/json/new_json_reader.cpp
hubgeter: yeah , i will fix it.

be/src/vec/exec/format/json/new_json_reader.cpp
hubgeter: I'm not sure if using `switch (value->GetType())` to get the string behavior would improve performance, but I might give it a try.

be/src/vec/exec/format/json/new_json_reader.cpp
hubgeter: i need check `_is_load` and  `type_desc` , It seems that there is no way to use `switch`.

hubgeter:
run buildall

hubgeter:
run buildall

morningman (dismissed):
LGTM

eldenmoon (dismissed):
LGTM

hubgeter:
run buildall

hubgeter:
run buildall

fe/fe-core/src/main/java/org/apache/doris/datasource/hive/source/HiveScanNode.java
```diff
@@ -364,14 +364,20 @@ public TableIf getTargetTable() {
...
+        String inputFormatName = table.getSd().getInputFormat();
         String hiveFormat = HiveMetaStoreClientHelper.HiveFileFormat.getFormat(inputFormatName);
         if (hiveFormat.equals(HiveMetaStoreClientHelper.HiveFileFormat.PARQUET.getDesc())) {
             type = TFileFormatType.FORMAT_PARQUET;
         } else if (hiveFormat.equals(HiveMetaStoreClientHelper.HiveFileFormat.ORC.getDesc())) {
             type = TFileFormatType.FORMAT_ORC;
         } else if (hiveFormat.equals(HiveMetaStoreClientHelper.HiveFileFormat.TEXT_FILE.getDesc())) {
-            type = TFileFormatType.FORMAT_CSV_PLAIN;
+            String serDeLib = table.getSd().getSerdeInfo().getSerializationLib();
+            if (serDeLib.equals("org.apache.hive.hcatalog.data.JsonSerDe")) {
```
msridhar78, line 376: Will it be better to check if serDeLib.contains("JsonSerDe") as there can be other imports for JsonSerDe lib  like org.apache.hadoop.hive.serde2.JsonSerDe

morningman:
run buildall

fe/fe-core/src/main/java/org/apache/doris/datasource/hive/source/HiveScanNode.java
morningman: Added, thank you

kaka11chen (approved):
LGTM

morningman (approved):

CalvinKirs (approved):

PR #42379, opened 2024-10-24
[fix](cloud-mow) MS should delete the existing keys before rewriting it when processing old version delete bitmap on cu compaction

hust-hhb:
pr https://github.com/apache/doris/pull/40204 support deleting  old version delete bitmap when doing cu compaction, it will update delete bitmap with agg result, then deleting  old version delete bitmap. Updating delete bitmap means overwriting existing keys, however delete bitmap may split into multiple kvs to store in fdb, so we should delete the existing keys and rewrite it to fdb instead of overwriting directly.

PR #40204, opened 2024-08-30
[improve](cloud-mow) merge and remove old version of delete bitmap when cumulative compaction is done

hust-hhb:
When the count of delete bitmap is big, it may lead to high cpu use rate,  we can merge old version delete bitmap to reduce the count of delete bitmap when cumulative compaction is done to  reduce the use rate of cpu.
Here is an example:
before cu compaciton ,rowset layout is

    "cumulative point": 10,
    "rowsets": [
        "[0-1] 0 DATA NONOVERLAPPING 0200000000000000ffffffffffffffecffffffffffffffff 0",
        "[2-9] 1 DATA NONOVERLAPPING 0200000000002cc1a8447a4e0ba5850f773803ae2d534284 1.21 KB",
        "[10-10] 1 DATA NONOVERLAPPING 0200000000002d07a8447a4e0ba5850f773803ae2d534284 1.09 KB",
        "[11-11] 1 DATA NONOVERLAPPING 0200000000002d14a8447a4e0ba5850f773803ae2d534284 1.10 KB",
        "[12-12] 1 DATA NONOVERLAPPING 0200000000002d23a8447a4e0ba5850f773803ae2d534284 1.10 KB",
        "[13-13] 1 DATA NONOVERLAPPING 0200000000002d33a8447a4e0ba5850f773803ae2d534284 1.10 KB",
        "[14-14] 1 DATA NONOVERLAPPING 0200000000002d40a8447a4e0ba5850f773803ae2d534284 1.10 KB"
    ],
base rowset is [2-9], now cu compaciton input rowset range is [10-14], after compaciton, rowset layout is

    "cumulative point": 10,
    "rowsets": [
        "[0-1] 0 DATA NONOVERLAPPING 0200000000000000ffffffffffffffecffffffffffffffff 0",
        "[2-9] 1 DATA NONOVERLAPPING 0200000000002cc1a8447a4e0ba5850f773803ae2d534284 1.21 KB",
        "[10-14] 1 DATA NONOVERLAPPING 0200000000002d63a8447a4e0ba5850f773803ae2d534284 1.21 KB"
    ],
1. the delete bitmap of rowset [10-10] -> [11-11] -> [12-12] -> [13-13] -> [14-14] will be deleted when delete_expired_stale_rowsets
2. the delete bitmap of version 10~14, will do agg first, then update the agg result on verison 14 delete bitmap
3. if update sucess, add the delete bitmap of version 10~13 to the queue which name is to_remove_vec
[...]

PR #43479, opened 2024-11-08
[fix](cloud-mow) MS should delete the existing keys before rewriting it when processing old version delete bitmap on cu compaction

hust-hhb:
pr https://github.com/apache/doris/pull/40204 support deleting old version delete bitmap when doing cu compaction, it will update delete bitmap with agg result, then deleting old version delete bitmap. Updating delete bitmap means overwriting existing keys, however delete bitmap may split into multiple kvs to store in fdb, so we should delete the existing keys and rewrite it to fdb instead of overwriting directly.
pick pr https://github.com/apache/doris/pull/42379


hust-hhb:
run buildall

PR #43337, opened 2024-11-06
[fix](cloud-mow)Full compaction no need to sync delete bitmap

hust-hhb:
sync delete bitmap may cost too much time,which may lead to full compaction fail, but full compaction no need to sync it.

PR #43613, opened 2024-11-11
[fix](cloud-mow)Full compaction no need to sync delete bitmap

hust-hhb:
sync delete bitmap may cost too much time,which may lead to full compaction fail, but full compaction no need to sync it.
pick pr:https://github.com/apache/doris/pull/43337


hust-hhb:
run buildall

PR #42471, opened 2024-10-25
[fix](cloud-mow) Add enable config for procesing old version delete bitmap when doing cu compaction

hust-hhb:
pr https://github.com/apache/doris/pull/40204 support deleting  old version delete bitmap when doing cu compaction, it a new function and better to add a config to enable it or not.

PR #43615, opened 2024-11-11
[fix](cloud-mow) Add enable config for procesing old version delete bitmap when doing cu compaction

hust-hhb:
pr https://github.com/apache/doris/pull/40204 support deleting old version delete bitmap when doing cu compaction, it a new function and better to add a config to enable it or not.
pick pr:https://github.com/apache/doris/pull/42471


hust-hhb:
run buildall

gavinchou (approved):

PR #43480, opened 2024-11-08
[fix](ngram bloomfilter) fix narrow conversion for ngram bf_size

airborne12:
### What problem does this PR solve?

Problem Summary:
Fix ngram bloomfilter index coredump as below
```
*** SIGFPE integer divide by zero (@0x56294f026472) received by PID 4016941 (TID 4019213 OR 0x7f294eb4d640) from PID 1325556850; stack trace: ***
 0# doris::signal::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*) at be/src/common/signal_handler.h:421
 1# PosixSignals::chained_handler(int, siginfo*, void*) [clone .part.0] in bin/jdk-17.0.2/lib/server/libjvm.so
 2# JVM_handle_linux_signal in bin/jdk-17.0.2/lib/server/libjvm.so
 3# 0x00007F3071042520 in /lib/x86_64-linux-gnu/libc.so.6
 4# doris::segment_v2::NGramBloomFilter::add_bytes(char const*, unsigned int) at be/src/olap/rowset/segment_v2/ngram_bloom_filter.cpp:61
 5# doris::ITokenExtractorHelper<doris::NgramTokenExtractor>::string_to_bloom_filter(char const*, unsigned long, doris::segment_v2::BloomFilter&) const at be/src/olap/itoken_extractor.h:61
 6# doris::segment_v2::NGramBloomFilterIndexWriterImpl::add_values(void const*, unsigned long) at be/src/olap/rowset/segment_v2/bloom_filter_index_writer.cpp:250
```

### Release note

None

### Check List (For Author)

- Test 
    - [x] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

- Does this need documentation?
    - [ ] No.
    - [ ] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [x] Confirm test cases
- [ ] Confirm document
- [x] Add branch pick label

PR #43654, opened 2024-11-12
[fix](ngram bloomfilter) fix narrow conversion for ngram bf_size #43480

airborne12:
cherry pick from #43480


airborne12:
run buildall

PR #43788, opened 2024-11-12
[Fix](Nereids) fix floor/round/ceil/truncate functions type compute precision problem (#43422)

LiBinfeng-01:
pick: #43422 

- Problem function like ```select floor(300.343, 2)``` precision should be 5 and scale should be 2, but now is (6, 2) after compute precision, but after folding const on fe, it changed to (5, 2) but upper level of plan still expect the output of child to be (6, 2). So it would rise an exception when executing.

- How it was fixed  fix folding constant precision of floor/round/ceil/truncate functions from (5, 2) to (6, 2) in upper case

- Notion when second value is negative and it absolute value >= precision - value, it can not be expressed in fe which result is zero with decimal type (3, 0). like 000. So just let it go back and no using folding constant by fe.

- Related PR: #40744

- Release note Fix floor/round/ceil functions precision problem in folding constant

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/nereids/trees/expressions/functions/executable/NumericArithmetic.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/nereids/trees/expressions/functions/executable/NumericArithmetic.java
@@ -694,12 +694,17 @@ public class NumericArithmetic {
         return input;
     }
 
+    private static Expression castDecimalV3Literal(DecimalV3Literal literal, int precision) {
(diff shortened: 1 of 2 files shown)


LiBinfeng-01:
run buildall

PR #43879, opened 2024-11-13
[fix](cloud-mow) Fix sending commiting rpc to FE twice problem

hust-hhb:
this pr check transaction status before sending delete bitmap task to be, if transaction status is committed or visible, it no need to recalculate delete bitmap again, just retrun rpc success to be
pick pr:https://github.com/apache/doris/pull/41395


hust-hhb:
run buildall

PR #43533, opened 2024-11-08
[feature](security) Support block specific query with AST names

Hastyshell:
### What problem does this PR solve?

Support block specific query with AST names when necessary for security reasons, configure the name list in fe.conf, for example:

```
block_sql_ast_names="CreateFileStmt, CreateFunctionStmt"
```

### Release note

Support to block some SQL statements, e.g. CreateFileStmt, when necessary for security reasons.

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [x] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

- Does this need documentation?
    - [ ] No.
    - [x] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #43887, opened 2024-11-13
[feature](security) Support block specific query with AST names (#43533)

Hastyshell:
pick #43533

Support block specific query with AST names when necessary for security reasons, configure the name list in fe.conf, for example:
```
block_sql_ast_names=CreateFileStmt, CreateFunctionStmt
```


TangSiyang2001:
run buildall

PR #43904, opened 2024-11-13
[enhancement] Support TCP Keep-Alive on MySQL connections

lide-reed:
The default behavior of Options.KEEP_ALIVE is typically disabled, meaning that TCP keep-alive is not enabled by default. This means that if you do not explicitly enable it, the connection will not send keep-alive probes when idle.


liutang123 (approved):
LGTM

xy720 (approved):
LGTM

lide-reed:
run buildall

morrySnow (approved):

PR #43913, opened 2024-11-13
[Opt](multi-catalog)Improve performance by introducing cache of list directory files when getting split for each query.

kaka11chen:

Refer to trino to implement the cache mechanism of multiple hive tables at the query level to obtain the file split list of each partition.
Because files within a query should have the same visibility, the split list of partitions that see the same table should be consistent across the query scope. So this cache is reasonable and should be enabled by default.
The mechanism in Trino is transactional level. A transaction can see the same table, so the command is `TransactionScopeCachingDirectoryLister`. This name is retained for Doris to expand to the transaction concept in the future.
In addition, for this scenario, because the caffeine cache currently used by doris has an elimination phase strategy, the existing cache items in the window area may be eliminated immediately after the weight is updated. Therefore, `EvictableCache` which based on guava was introduced and eliminated based on segment LRU.

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [x] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
+++ b/fe/fe-common/src/main/java/org/apache/doris/common/Config.java
@@ -2214,6 +2214,10 @@ public class Config extends ConfigBase {
         "Max cache number of external table row count"})
     public static long max_external_table_row_count_cache_num = 100000;
 
+    @ConfField(description = {"每个查询的外表文件元数据缓存的最大文件数量。",
+            "Max cache file number of external table split file meta cache at query level."})
+    public static long max_external_table_split_file_meta_cache_num = 100000;
+
     /**
      * Max cache loader thread-pool size.
      * Max thread pool size for loading external meta cache
--- a/fe/fe-core/src/main/java/org/apache/doris/fs/TransactionScopeCachingDirectoryListerFactory.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/fs/TransactionScopeCachingDirectoryListerFactory.java
@@ -1,0 +1,59 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+// This file is copied from
+// https://github.com/trinodb/trino/blob/438/plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/TransactionScopeCachingDirectoryListerFactory.java
+// and modified by Doris
+
+package org.apache.doris.fs;
+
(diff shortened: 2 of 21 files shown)


fe/fe-common/src/main/java/org/apache/doris/common/Config.java
```diff
@@ -2134,6 +2134,10 @@ public class Config extends ConfigBase {
         "Max cache number of external table row count"})
     public static long max_external_table_row_count_cache_num = 100000;
 
+    @ConfField(description = {"每个查询 list 文件目录数。",
```
morningman, line 2137: Add more detail explain.
And I think this is only for external table query? Better rename it.

fe/fe-core/src/main/java/org/apache/doris/fs/TransactionScopeCachingDirectoryListerFactory.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.Optional;
+import java.util.OptionalLong;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class TransactionScopeCachingDirectoryListerFactory {
+    //TODO use a cache key based on Path & SchemaTableName and iterate over the cache keys
+    // to deal more efficiently with cache invalidation scenarios for partitioned tables.
+    private final Optional<Cache<TransactionDirectoryListingCacheKey, FetchingValueHolder>> cache;
+    private final AtomicLong nextTransactionId = new AtomicLong();
```
morningman, line 36: Why not using query id?

fe/fe-core/src/main/java/org/apache/doris/nereids/glue/translator/PhysicalPlanTranslator.java
```diff
@@ -252,6 +257,8 @@ public PhysicalPlanTranslator(PlanTranslatorContext context) {
     public PhysicalPlanTranslator(PlanTranslatorContext context, StatsErrorEstimator statsErrorEstimator) {
         this.context = context;
         this.statsErrorEstimator = statsErrorEstimator;
+        this.directoryLister = new TransactionScopeCachingDirectoryListerFactory(100000).get(
```
morningman, line 260: `100000` should be a config

kaka11chen:
run buildall

kaka11chen:
run buildall

kaka11chen:
run buildall

fe/fe-core/src/main/java/org/apache/doris/fs/TransactionScopeCachingDirectoryListerFactory.java
kaka11chen: The mechanism in Trino is transactional level. A transaction can see the same table, so the command is TransactionScopeCachingDirectoryLister. This name is retained for Doris to expand to the transaction concept in the future.

fe/fe-common/src/main/java/org/apache/doris/common/Config.java
kaka11chen: ok

kaka11chen:
run buildall

kaka11chen:
run buildall

kaka11chen:
run buildall

kaka11chen:
run buildall

starocean999 (dismissed):

morningman (dismissed):

morningman:
run buildall

kaka11chen:
run buildall

morningman:
run buildall

morningman (approved):
LGTM

CalvinKirs (approved):

PR #43767, opened 2024-11-12
[Fix](multi-catalog)Fixed incorrect parameter transmission when obtaining files by hms partition and incorrect parameter transmission concurrently.

kaka11chen:
### What problem does this PR solve?

Problem Summary:

### Release note

Fixed incorrect parameter transmission when obtaining files by hms partition and incorrect parameter transmission concurrently.
This is probably a clerical error, and the current test can cover it.

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [x] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [x] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [] Other reason: 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

- Does this need documentation?
    - [x] No.
    - [ ] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #43914, opened 2024-11-14
[Fix](multi-catalog)Fixed incorrect parameter transmission when obtaining files by hms partition and incorrect parameter transmission concurrently. #43767

morningman:
cherry pick from #43767


morningman:
run buildall

PR #43853, opened 2024-11-13
[fix](regression-test) add checks for existence and successful upload of data files in hive-metastore.sh

suxiaogang223:
…hive-metastore.sh

### What problem does this PR solve?

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

- Does this need documentation?
    - [ ] No.
    - [ ] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #43924, opened 2024-11-14
[fix](regression-test) add checks for existence and successful upload of data files in hive-metastore.sh #43853

morningman:
cherry pick from #43853


morningman:
run buildall

PR #43968, opened 2024-11-14
[fix](cloud-mow) tablet_id should use int64_t to avoid truncation problem when removing old delete bitmap

hust-hhb:
tablet_id is int64_t, use int32_t may have truncation problem, which will lead to send wrong tablet_id to ms, and do nothing on remove old delete bitmap
pick:https://github.com/apache/doris/pull/43956


hust-hhb:
run buildall

zhannngchen (approved):
LGTM

dataroaring (approved):
LGTM

PR #42328, opened 2024-10-23
[Refactor](inverted index) refactor inverted index file writer for v1/v2 index write

airborne12:
## Proposed changes

The original write_v1 and write_v2 methods in the inverted index file writer were overly long and difficult to maintain. This PR refactors these methods to improve code maintainability and readability.

PR #43993, opened 2024-11-15
[Refactor](inverted index) refactor inverted index file writer for v1/v2 index write #42328

airborne12:
cherry pick from #42328


airborne12:
run buildall

PR #43224, opened 2024-11-04
[improve](load) show 99.99% instead of 100.00% in progess when load is not finished

kaijchen:
### What problem does this PR solve?

Currently when all instances finished, but the load is not finished,
the progress will display `100.00%`, which is misleading to users.

This PR detected changes it to `99.99%`.

### Check List (For Committer)

- Test 

    - [ ] Regression test
    - [ ] Unit Test
    - [x] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No colde files have been changed.
        - [ ] Other reason 

- Behavior changed:

    - [x] No.
    - [ ] Yes. 

- Does this need documentation?

    - [x] No.
    - [ ] Yes. 

- Release note

    
    
    None

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [x] Add branch pick label

PR #44034, opened 2024-11-15
[improve](load) show 99.99% instead of 100.00% in progess when load is not finished (#43224)

kaijchen:
backport #43224


kaijchen:
run buildall

kaijchen:
run buildall

PR #44044, opened 2024-11-15
[Fix]Reset Connection's resource tag in OlapScanNode

wangbo:

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:
Currently Tag is set to ConnectionContext when mysql client is connected, but when user's tag property changes, tag in ConnectionContext not changed, this may cause query in the connection using the stale tag, user has to reconnect mysql client then user's new property will take effect.

So we can still store Tag in ConnectionContext, but reset it when local select table's replica.

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [x] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/planner/OlapScanNode.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/planner/OlapScanNode.java
@@ -765,6 +765,7 @@ public class OlapScanNode extends ScanNode {
         boolean skipMissingVersion = false;
         ConnectContext context = ConnectContext.get();
         if (context != null) {
+            context.resetResourceTags();
             allowedTags = context.getResourceTags();
             needCheckTags = context.isResourceTagsSet();
             useFixReplica = context.getSessionVariable().useFixReplica;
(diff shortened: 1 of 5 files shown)


wangbo:
run buildall

fe/fe-core/src/main/java/org/apache/doris/planner/OlapScanNode.java
```diff
@@ -765,6 +765,7 @@ private void addScanRangeLocations(Partition partition,
         boolean skipMissingVersion = false;
         ConnectContext context = ConnectContext.get();
         if (context != null) {
+            context.resetResourceTags();
```
morningman, line 768: You can not only reset resource tags in olap scan node.

wangbo:
This is not a good solution. It may cause too frequent calls when if a query has many tables, there is no particularly good solution in the short term.

PR #44057, opened 2024-11-15
[chore](autopick)When the PR is merged, adding a label afterward should also trigger the workflow.

CalvinKirs:
## Important Note:
There’s a potential issue here: if the label was added before the merge and later removed and then readd, the workflow may still be triggered again. Please avoid performing these operations repeatedly, as CI resources are valuable.

### DIFF
--- a/.github/workflows/auto-cherry-pick.yml
+++ b/.github/workflows/auto-cherry-pick.yml
@@ -21,6 +21,7 @@ on:
   pull_request_target:
     types:
       - closed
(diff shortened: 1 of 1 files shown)


morningman (approved):

dataroaring (approved):
LGTM

dataroaring (approved):
LGTM

airborne12 (approved):
lgtm

PR #43509, opened 2024-11-08
[fix](cloud-mow) MS should create new Transaction to continue geting delete bitmap when encounter TXN_TOO_OLD

hust-hhb:
When delete bitmap count is big,  geting delete bitmap may encounter TXN_TOO_OLD, ms should create a new transaction to reading the remaining data instead of returning TXN_TOO_OLD code.

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [x] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

- Does this need documentation?
    - [x] No.
    - [ ] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #44074, opened 2024-11-16
[fix](cloud-mow) MS should create new Transaction to continue geting delete bitmap when encounter TXN_TOO_OLD

hust-hhb:
When delete bitmap count is big, geting delete bitmap may encounter TXN_TOO_OLD, ms should create a new transaction to reading the remaining data instead of returning TXN_TOO_OLD code.
pick pr:https://github.com/apache/doris/pull/43509


hust-hhb:
run buildall

PR #44083, opened 2024-11-17
[enhancement]Add comments to block.h and add unit tests

Yoruet:

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label


zclllyybb:
run buildall

be/src/vec/core/block.h
```diff
@@ -95,8 +95,9 @@ class Block {
     Block(Block&& block) = default;
     Block& operator=(Block&& other) = default;
 
+    /// Reserve memory for internal containers
     void reserve(size_t count);
-    // Make sure the nammes is useless when use block
+    /// Clear all column names and name index mappings in the block
```
zclllyybb, line 100: like these, we'd better make people know `index_by_name` and all `data`'s name will be clear.

zclllyybb (commented):
We can by the way unify the format, and the function is not related to the large paragraph description with /**/, and then the function of the comments are unified with //, if some kind of function categorisation of the comments can be used /// (for example, "the following is the query interface", etc.)

zclllyybb (commented):
for all test, we should cover the corner case, at least cover:
1. empty block
2. Column's type is const+nullable

be/test/vec/core/block_test.cpp
```diff
@@ -693,6 +693,1801 @@ void serialize_and_deserialize_test_array() {
...
+        EXPECT_NO_THROW(block.reserve(100));
+        block.reserve(3);
+        block.insert({col1->get_ptr(), type, "col1"});
+        block.insert({col2->get_ptr(), type, "col2"});
+        block.insert({col3->get_ptr(), type, "col3"});
+        EXPECT_EQ(3, block.columns());
+
+        // Test clear_names
+        block.clear_names();
+        EXPECT_EQ("", block.get_by_position(0).name);
```
zclllyybb, line 805: I think all these functions should have exception test. such as `EXPECT_DEATH(empty_block.get_by_potition(1))`

Yoruet:
run buildall

PR #41245, opened 2024-09-25
[fix](load) fix memtable memory limiter total mem usage

kaijchen:
## Proposed changes

Previously, `mem_usage = write_mem + flush_mem`, because `active_mem` is included in `write_mem`.

After #40912, `write_mem` becomes `queue_mem`, which no longer includes `active_mem`.
This PR fixes this problem, by setting `mem_usage = active_mem + queue_mem + flush_mem`

PR #44135, opened 2024-11-18
[improve](load) more accurate memtable memory usage and new flush policy

kaijchen:
backport #40912 #41018 #41245 #41278


kaijchen:
run buildall

PR #44061, opened 2024-11-15
[fix](compile) fix macos compile failed

924060929:
### What problem does this PR solve?

fix macos compile failed, introduced by #40813, #42930, #43218, #43289

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [x] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

- Does this need documentation?
    - [x] No.
    - [ ] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #44201, opened 2024-11-18
[fix](compile) fix macos compile failed (#44061)

924060929:

cherry pick from #44061


924060929:
run buildall

PR #38716, opened 2024-08-01
[pick](nestedtype)support nested type with agg replace_if_not_null

amorynan:
## Proposed changes
backport: https://github.com/apache/doris/pull/38304
Issue Number: close #xxx

PR #44210, opened 2024-11-18
(fix)(compaction) Fix mor/agg table compaction problem introduced by #38716

xiaokang:
introduced by #38716


xiaokang:
run buildall

PR #42004, opened 2024-10-17
[Opt](orc)Optimize the merge io when orc reader read multiple tiny stripes.

hubgeter:
### What problem does this PR solve?

When reading orc files, we may encounter a scenario where the stripe byte size is very small but the number of stripes is very large. 

This pr introduces three session variables `orc_tiny_stripe_threshold_bytes`, `orc_once_max_read_bytes`, and `orc_max_merge_distance_bytes` to optimize io reading for the above scenarios. 

If a stripe byte size is less than `orc_tiny_stripe_threshold_bytes`, we will consider it as a tiny stripe. For multiple tiny stripes, we will perform IO merge reading according to the `orc_once_max_read_bytes` and `orc_max_merge_distance_bytes` parameters. Among them, `orc_once_max_read_bytes` indicates the maximum size of the merged IO. You should not set `orc_once_max_read_bytes` less than `orc_tiny_stripe_threshold_bytes`, although we will not force an error. When using tiny stripe reading optimization, since tiny stripes are not necessarily continuous, when the distance between two tiny stripes is greater than `orc_max_merge_distance_bytes`, we will not merge them into one IO.

If you don't want to use this optimization, you can `set orc_tiny_stripe_threshold_bytes = 0`.

Default parameters:
```mysql
orc_tiny_stripe_threshold_bytes = 8388608 (8M)
orc_once_max_read_bytes = 8388608 (8M)
orc_max_merge_distance_bytes = 1048576 (1M)
```

We also add relevant profiles for this purpose so that parameters can be adjusted to optimize reading.
`RangeCacheFileReader`:
1. `CacheRefreshCount`: how many IOs are merged
2. `ReadToCacheBytes`: how much data is actually read after merging
3. `ReadToCacheTime`: how long it takes to read data after merging
4. `RequestBytes`: how many bytes does the apache-orc library actually need to read the orc file
5. `RequestIO`: how many times the apache-orc library calls this read interface
6. `RequestTime`: how long it takes the apache-orc library to call this read interface

It should be noted that `RangeCacheFileReader` is a wrapper of the reader that actually reads data, such as the hdfs reader, so strictly speaking, `CacheRefreshCount` is not equal to how many IOs are initiated to hdfs, because each time the hdfs reader is requested, the hdfs reader may not be able to read all the data at once.

This pr also involves changes to the apache-orc third-party library: https://github.com/apache/doris-thirdparty/pull/244.
[...]

PR #44239, opened 2024-11-19
[Opt](orc)Optimize the merge io when orc reader read multiple tiny stripes. (#42004)

hubgeter:
bp #42004

When reading orc files, we may encounter a scenario where the stripe byte size is very small but the number of stripes is very large.

This pr introduces three session variables
`orc_tiny_stripe_threshold_bytes`, `orc_once_max_read_bytes`, and `orc_max_merge_distance_bytes` to optimize io reading for the above scenarios.

If a stripe byte size is less than `orc_tiny_stripe_threshold_bytes`, we will consider it as a tiny stripe. For multiple tiny stripes, we will perform IO merge reading according to the `orc_once_max_read_bytes` and `orc_max_merge_distance_bytes` parameters. Among them, `orc_once_max_read_bytes` indicates the maximum size of the merged IO. You should not set `orc_once_max_read_bytes` less than `orc_tiny_stripe_threshold_bytes`, although we will not force an error. When using tiny stripe reading optimization, since tiny stripes are not necessarily continuous, when the distance between two tiny stripes is greater than `orc_max_merge_distance_bytes`, we will not merge them into one IO.

If you don't want to use this optimization, you can `set orc_tiny_stripe_threshold_bytes = 0`.

Default parameters:
```mysql
orc_tiny_stripe_threshold_bytes = 8388608 (8M)
orc_once_max_read_bytes = 8388608 (8M)
orc_max_merge_distance_bytes = 1048576 (1M)
```

We also add relevant profiles for this purpose so that parameters can be adjusted to optimize reading.
`RangeCacheFileReader`:
1. `CacheRefreshCount`: how many IOs are merged
2. `ReadToCacheBytes`: how much data is actually read after merging
3. `ReadToCacheTime`: how long it takes to read data after merging
4. `RequestBytes`: how many bytes does the apache-orc library actually need to read the orc file
5. `RequestIO`: how many times the apache-orc library calls this read interface
6. `RequestTime`: how long it takes the apache-orc library to call this read interface

It should be noted that `RangeCacheFileReader` is a wrapper of the reader that actually reads data, such as the hdfs reader, so strictly speaking, `CacheRefreshCount` is not equal to how many IOs are initiated to hdfs, because each time the hdfs reader is requested, the hdfs reader may not be able to read all the data at once.

This pr also involves changes to the apache-orc third-party library: https://github.com/apache/doris-thirdparty/pull/244. Reference implementation:
https://github.com/trinodb/trino/blob/master/lib/trino-orc/src/main/java/io/trino/orc/OrcDataSourceUtils.java#L36

#### Summary:
```mysql
set orc_tiny_stripe_threshold_bytes = xxx;
set orc_once_max_read_bytes = xxx;
set orc_max_merge_distance_bytes = xxx;

# xxx is the size in bytes
```

### Release note
Introduces three session variables `orc_tiny_stripe_threshold_bytes`, `orc_once_max_read_bytes`, and `orc_max_merge_distance_bytes` to optimize io reading of scenarios where the orc stripe byte size is very small but the number of stripes is very large.

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Release note

None

### Check List (For Author)

- Test 
    - [x] Regression test
    - [x] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [x] Yes. 

    - [ ] No.
    - [x] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label


hubgeter:
run buildall

hubgeter:
run buildall

hubgeter:
run buildall

hubgeter:
run buildall

PR #44338, opened 2024-11-20
[fix](planner) query should be cancelled if limit reached

morningman:

Problem Summary:
When there is a `limit` cluse in SQL, if FE has obtained data with more than the `limit` number of rows,
it should send a cancel command to BE to cancel the query to prevent BE from reading more data.
However, this function has problems in the current code and does not work.
Especially in external table query, this may result in lots of unnecessary network io read.

1. `isBlockQuery`

    In the old optimizer, if a query statement contains a `sort` or `agg` node,
    `isBlockQuery` will be marked as true, otherwise it will be false.
    In the new optimizer, this value is always true.
  
    Regardless of the old or new optimizer, this logic is wrong.
    But only when `isBlockQuery = false` will the reach limit logic be triggered.

2. Calling problem of reach limit logic

    The reach limit logic judgment will only be performed when `eos = true` in the rowBatch returned by BE.
    This is wrong.
    Because for `limit N` queries, each BE's own `limit` is N. But for FE, as long as the total number of rows
    returned by all BEs exceeds N, the reach limit logic can be triggered.
    So it should not be processed only when `eos = true`.

The PR mainly changes:

1. Remove `isBlockQuery`

    `isBlockQuery` is only used in the reach limit logic. And it is not needed. Remove it completely.

2. Modify the judgment position of reach limit.

    When the number of rows obtained by FE is greater than the limit,
    it will check the reach limit logic.

3. fix wrong `limitRows` in `QueryProcessor`

    the limitRows should be got from the first fragment, not last.

4. In scanner scheduler on BE side, if scanner has limit, ignore the scan bytes threshold per round.

### Release note

[fix](planner) query should be cancelled if limit reached

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [x] Unit Test
    - [x] Manual test
        test single and multi-be nodes with limit sql.
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/be/src/vec/exec/scan/scanner_scheduler.cpp
+++ b/be/src/vec/exec/scan/scanner_scheduler.cpp
@@ -268,8 +268,9 @@ void ScannerScheduler::_scanner_scan(std::shared_ptr<ScannerContext> ctx,
             }
 
             size_t raw_bytes_threshold = config::doris_scanner_row_bytes;
-            size_t raw_bytes_read = 0; bool first_read = true;
-            bool has_limit = scanner->limit() > 0;
+            size_t raw_bytes_read = 0;
+            bool first_read = true;
+            int64_t limit = scanner->limit();
             while (!eos && raw_bytes_read < raw_bytes_threshold) {
                 if (UNLIKELY(ctx->done())) {
                     eos = true;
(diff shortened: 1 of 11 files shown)


morningman:
run buildall

924060929 (dismissed):

morningman:
run buildall

morningman:
run buildall

morningman:
run buildall

924060929 (dismissed):

kaka11chen (dismissed):
LGTM

morningman:
run buildall

morrySnow (dismissed):

924060929 (dismissed):

HappenLee (approved):
LGTM

kaka11chen (approved):
LGTM

PR #42252, opened 2024-10-22
[test](routine load) fix some unstable routine load cases

sollhui:
Fix some unstable routine load case.

PR #44376, opened 2024-11-21
[test](routine load) fix some unstable routine load cases (#42252)

sollhui:
pick #42252

Fix some unstable routine load case.


sollhui:
run buildall

PR #43230, opened 2024-11-04
[test](routine load) add routine load doc case

sollhui:
Add routine load doc case.

PR #44382, opened 2024-11-21
[test](routine load) add routine load doc case (#43230)

sollhui:
pick #43230

Add routine load doc case.


sollhui:
run buildall

PR #37590, opened 2024-07-10
[opt](regression-test) move some cases to p2 to accelerate the testing speed of p0

sollhui:
move some cases to p2 to accelerate the testing speed of p0

PR #44389, opened 2024-11-21
[opt](regression-test) move some cases to p2 to accelerate the testing speed of p0 (#37590)

sollhui:
pick #37590

move some cases to p2 to accelerate the testing speed of p0


sollhui:
run buildall

PR #44017, opened 2024-11-15
[refine](column) Remove some unnecessary get_int functions.

Mryange:
### What problem does this PR solve?
get_int is a virtual function, but in some cases, the type can be explicitly determined, 
 so this function is not needed.

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

- Does this need documentation?
    - [ ] No.
    - [ ] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #44466, opened 2024-11-22
Revert "[refine](column) Remove some unnecessary get_int functions."

zclllyybb:
Reverts apache/doris#44017

In this pr, for some code path that `col2`'s type is not `FromType2` we assert it's wrong and throw error. but in fact because of some misuse in our code, **many date function didn't specify the `FromType2`**. so the "wrong" path is useful now. Only when we fixed all the calling with wrong `FromType2`, this pr could be applied.


zclllyybb:
run buildall

Mryange (approved):

BiteTheDDDDt (approved):

PR #44496, opened 2024-11-23
[opt](file_cache) Add config to enable base compaction output to file cache (#44497)

gavinchou:
Previous implementation does not allow the output of base compaction write into file cache, which may have some performance penalty.

This commit add a config to make that policy configurable. be.conf `enable_file_cache_keep_base_compaction_output` it is false by default.

If your file cache is ample enough to accommodate all the data in your database, enable this option; otherwise, it is recommended to leave it disabled.

master branch PR #44497


gavinchou:
run buildall

dataroaring (approved):
LGTM

PR #44547, opened 2024-11-25
[fix](hyperscan) Fix hyper scan fall back to re2

zhiqiang-hhhh:

* Core modification
When hyper scan failed, we should not set_error in FunctionContext.
Since set_error will try cancel query, but actually we want to fall back to re2 in this case.

* Some refactor
Rename FunctionRegexp so that we can distinguish regexp match with regexp_extract.

* Reproduce
```cpp
SELECT * FROM regexp_test_chinese WHERE city REGEXP "^上海｜^北京" ORDER BY id;
```
Note, the `|` in above sql is a Chinese character.

### Release note

None

### Check List (For Author)

- Test 
    - [x] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label


zhiqiang-hhhh:
run buildall

xiaokang (dismissed):
LGTM

be/src/vec/functions/like.cpp
```diff
@@ -499,9 +496,6 @@ Status FunctionLikeBase::hs_prepare(FunctionContext* context, const char* expres
         hs_free_database(*database);
         *database = nullptr;
         *scratch = nullptr;
-        if (context) {
```
cambyzju, line 502: please leave a comment here, avoid somebody add these codes back again next time.

zhiqiang-hhhh: OK，I will add some test and comment.

xiaokang (approved):
LGTM

cambyzju (approved):

PR #44635, opened 2024-11-26
[Opt] (multi-catalog) opt max scanner thread number in batch split mode.

kaka11chen:

Problem Summary:

There's only one scan range for each backend in batch split mode. Each backend only starts up one ScanNode instance. However, when calculating the concurrency of scanners in the scan operator in batch split mode, it is not divided by 1, but by `query_parallel_instance_num`, resulting in poor performance of batch split mode.

### Release note
There's only one scan range for each backend in batch split mode. Each backend only starts up one ScanNode instance. However, when calculating the concurrency of scanners in the scan operator in batch split mode, it is not divided by 1, but by `query_parallel_instance_num`, resulting in poor performance of batch split mode.

Fix calculation of maximum number of scanner threads in batch split mode.

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
    1. Create a hive partition table which contains many hive partitions(> 1024);
    2. Execute `select * from par_table` in doris.
    3. Open the profile to view the number of file scan operators and the number of `MaxScannerThreadNum`.
    You should be able to see that each be node of the scan operator is 1, and there will be more `MaxScannerThreadNum` in the scan operator.

        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/be/src/pipeline/exec/file_scan_operator.cpp
+++ b/be/src/pipeline/exec/file_scan_operator.cpp
@@ -40,8 +40,7 @@ Status FileScanLocalState::_init_scanners(std::list<vectorized::VScannerSPtr>* s
     auto& p = _parent->cast<FileScanOperatorX>();
     // There's only one scan range for each backend in batch split mode. Each backend only starts up one ScanNode instance.
     uint32_t shard_num =
-            std::min(config::doris_scanner_thread_pool_thread_num /
-                             (_batch_split_mode ? 1 : state()->query_parallel_instance_num()),
+            std::min(config::doris_scanner_thread_pool_thread_num / p.query_parallel_instance_num(),
                      _max_scanners);
     shard_num = std::max(shard_num, 1U);
     _kv_cache.reset(new vectorized::ShardedKVCache(shard_num));
@@ -55,17 +54,6 @@ Status FileScanLocalState::_init_scanners(std::list<vectorized::VScannerSPtr>* s
     return Status::OK();
 }
 
-Status FileScanLocalState::start_scanners(
-        const std::list<std::shared_ptr<vectorized::ScannerDelegate>>& scanners) {
-    auto& p = _parent->cast<FileScanOperatorX>();
-    // There's only one scan range for each backend in batch split mode. Each backend only starts up one ScanNode instance.
-    _scanner_ctx = vectorized::ScannerContext::create_shared(
-            state(), this, p._output_tuple_desc, p.output_row_descriptor(), scanners, p.limit(),
-            _scan_dependency, p.is_serial_operator(), true,
-            _batch_split_mode ? 1 : state()->query_parallel_instance_num());
-    return Status::OK();
-}
-
 std::string FileScanLocalState::name_suffix() const {
     return fmt::format(" (id={}. nereids_id={}. table name = {})",
                        std::to_string(_parent->node_id()), std::to_string(_parent->nereids_id()),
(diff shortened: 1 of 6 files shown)


kaka11chen:
run buildall

kaka11chen:
run buildall

be/src/pipeline/exec/file_scan_operator.cpp
```diff
@@ -37,9 +38,11 @@ Status FileScanLocalState::_init_scanners(std::list<vectorized::VScannerSPtr>* s
     }
 
     auto& p = _parent->cast<FileScanOperatorX>();
-    uint32_t shard_num = std::min(
-            config::doris_scanner_thread_pool_thread_num / state()->query_parallel_instance_num(),
-            _max_scanners);
+    // There's only one scan range for each backend in batch split mode. Each backend only starts up one ScanNode instance.
+    uint32_t shard_num =
+            std::min(config::doris_scanner_thread_pool_thread_num /
+                             (_batch_split_mode ? 1 : state()->query_parallel_instance_num()),
```
morningman, line 44: I suggest to extract this `_batch_split_mode ? 1 : state()->query_parallel_instance_num()` to a method in file scan operator.

be/src/pipeline/exec/scan_operator.cpp
```diff
@@ -979,18 +979,19 @@ Status ScanLocalState<Derived>::_prepare_scanners() {
...
         COUNTER_SET(_num_scanners, static_cast<int64_t>(scanners.size()));
-        RETURN_IF_ERROR(_start_scanners(_scanners));
+        RETURN_IF_ERROR(start_scanners(_scanners));
     }
     return Status::OK();
 }
 
 template <typename Derived>
-Status ScanLocalState<Derived>::_start_scanners(
+Status ScanLocalState<Derived>::start_scanners(
```
morningman, line 988: The only difference between this and `file_scan_operator` is that `file_scan_operator` pass the `query_parallel_instance_num` separately.
So I suggest to create a new virtual method `get_query_parallel_instance_num()`,
and for `scan_operator`, it returns `state()->query_parallel_instance_num()`.
For `file_scan_operator`, it returns `_batch_split_mode ? 1 : state()->query_parallel_instance_num()`

be/src/pipeline/exec/file_scan_operator.cpp
kaka11chen: If extract this function to scan operator, It seems that the runtime state cannot be obtained through `state()` except through the parameters of this function. But this feels a bit strange.

be/src/pipeline/exec/scan_operator.cpp
kaka11chen: If extract this function to scan operator, It seems that the runtime state cannot be obtained through state() except through the parameters of this function. But this feels a bit strange.

kaka11chen:
run buildall

kaka11chen:
run buildall

be/src/pipeline/exec/scan_operator.cpp
kaka11chen: Done.

morningman (approved):
LGTM

wuwenchi (approved):

PR #44782, opened 2024-11-29
[fix](cache) fix same sql return wrong result when switch database with `use db` and enable sql cache

htyoung:

It will return wrong result when running same sql in different db with `enable_sql_cache=true`

for example, the `db1` and `db2` has the same table `tbl` but the data are not equals, 
if execute the below sql in `db1` and cache the result, then execute it in `db2`, it will return the wrong result
```sql
select count(*) from tbl
```

### Check List (For Author)

- Test 
    - [x] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/common/NereidsSqlCacheManager.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/NereidsSqlCacheManager.java
@@ -74,9 +74,11 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 
-/** NereidsSqlCacheManager */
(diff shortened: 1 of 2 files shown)


htyoung:
run build all

924060929 (approved):

924060929:
run buildall

morningman (approved):
LGTM

PR #44783, opened 2024-11-29
[fix](nereids) fix merge_percentile_to_array when has same agg function

feiniaofeiafei:

Related PR: #34313

Problem Summary
The original PR did not handle the following scenario:
```sql
SELECT SUM(a), PERCENTILE(pk, 0.1) AS c1, PERCENTILE(pk, 0.1) AS c2, PERCENTILE(pk, 0.4) AS c3 FROM test_merge_percentile;
```
In this case, the aggregate outputs include two identical functions (PERCENTILE(pk, 0.1)). When constructing the LogicalProject, a map was used where the key is the child of an Alias and the value is the Alias itself. However, this approach loses information when two Aliases share the same child.
This PR modifies the map structure to use the child of an Alias as the key and a list of Alias objects as the value. This ensures that all Alias instances with the same child are preserved, resolving the issue of lost information in such cases.
### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label


feiniaofeiafei:
run buildall

morrySnow (approved):

starocean999 (approved):

PR #44273, opened 2024-11-19
[improvement](compaction) reduce tablet skip compaction time

luwei16:
The time for tablet skip compaction is 120 seconds, which is too long. In the scenario of high-frequency import (mow), it leads to a high compaction score. Therefore, reducing the skip time to 10 seconds is necessary.

PR #44794, opened 2024-11-29
[improvement](compaction) reduce tablet skip compaction time (#44273)

luwei16:
pick master #44273 

The time for tablet skip compaction is 120 seconds, which is too long. In the scenario of high-frequency import (mow), it leads to a high compaction score. Therefore, reducing the skip time to 10 seconds is necessary.


luwei16:
run buildall

luwei16:
run buildall

PR #44810, opened 2024-11-30
[fix](sql_cache) fix sql cache result wrong of from_unixtime(col, 'yyyy-MM-dd HH:mm:ss') (#44631)

924060929:
cherry pick from #44631


924060929:
run buildall

PR #44845, opened 2024-12-02
[bugfix](hive)Fix split assignment leak. releated to issue (#40683).

biohazard4321:
Also fix stop split too early in select join statement when enable_profile is true. Releated to issue (#40683).

Case appearing condition:  use batchmode and set enable_profile=true，hive table has enough split （like more than 10）

Under enable_profile is true, in query with join like "select  dt.d_year,dt.d_moy from  date_dim dt ,store_sales where dt.d_date_sk = store_sales.ss_sold_date_sk group by dt.d_year, dt.d_moy order by  dt.d_year, dt.d_moy limit 100",
the old fix in commit 83797e36aeb stops SplitAssignment in FileScanNode.getNodeExplainString, that would be called in normal query (in StmtExecutor.updateProfile)，then BE would get empty split and make wrong result。
So i changed the place where SplitAssignment need stop in explain only statement.

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/datasource/FileScanNode.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/datasource/FileScanNode.java
@@ -126,7 +126,6 @@ public abstract class FileScanNode extends ExternalScanNode {
         output.append(prefix);
         if (isBatchMode()) {
             output.append("(approximate)");
-            splitAssignment.stop();
         }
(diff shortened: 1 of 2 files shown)


biohazard4321:
run buildall

wuwenchi (approved):
LGTM

biohazard4321:
run buildall

morningman (approved):
LGTM

PR #44847, opened 2024-12-02
[fix](cache) fix same sql return wrong result when switch database with `use db` and enable sql cache (44782)

htyoung:
cherry pick from #44782


htyoung:
run build all

PR #44859, opened 2024-12-02
[Bug](runtime-filter) fix publish not inited rf when broadcast join meet wake up by downsteam #44823

BiteTheDDDDt:
pick from #44823


BiteTheDDDDt:
run buildall

PR #44959, opened 2024-12-03
[fix](auth)fix be enable http auth, some request link never return.

hubgeter:
Problem Summary:
if you `enable_all_http_auth = true` in be.conf, then restart be, and keep using `curl   -u  "xxxx:xxxx" http://127.0.0.1:8040/api/health` while be is starting. You may encounter a situation where the link does not return. 
Reason:
When be is still starting, there is no information about fe master. When you make an api request to be http port, be needs to request authentication information from fe, which will cause it to request a machine with empty ip and port 0. This rpc call will definitely fail (this is not equivalent to a password error). After receiving this failure, be does not `send_reply` to the api requester, so this api request cannot be returned.
 

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/be/src/http/http_handler_with_auth.cpp
+++ b/be/src/http/http_handler_with_auth.cpp
@@ -85,7 +85,8 @@ int HttpHandlerWithAuth::on_header(HttpRequest* req) {
 #ifndef BE_TEST
     TNetworkAddress master_addr = _exec_env->cluster_info()->master_fe_addr;
     if (master_addr.hostname.empty() || master_addr.port == 0) {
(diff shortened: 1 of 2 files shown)


hubgeter:
run buildall

gavinchou:
LGTM

morningman (approved):

gavinchou (approved):

PR #44821, opened 2024-11-30
[Refactor](query) refactor lock in fragment mgr and change std::unorder_map to phmap

HappenLee:
…

### What problem does this PR solve?

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

- Does this need documentation?
    - [ ] No.
    - [ ] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #45069, opened 2024-12-05
[Refactor](query) refactor lock in fragment mgr and change std::unorder_map to phmap

HappenLee:

Related PR: #44821


HappenLee:
run buildall

PR #44528, opened 2024-11-25
[opt](auto-inc) Allow to miss auto-increment column and other value columns in partial update

bobhan1:
### What problem does this PR solve?

Previously, if a partial update misses auto-increment key columns and other value columns, the load will fail with error `distributed column not found, column=xxx`. Considering the uniqueness of the generated value of auto-increment column, this PR converts the partial update load to upsert in this situation to make it llegal.

### Release note

Allow to miss auto-increment key column and other value columns in partial update load

### Check List (For Author)

- Test 
    - [x] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [x] Yes. 

- Does this need documentation?
    - [x] No.
    - [ ] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #45073, opened 2024-12-06
branch-2.1: [opt](auto-inc) Allow to miss auto-increment column and other value columns in partial update #44528

bobhan1:
pick https://github.com/apache/doris/pull/44528


bobhan1:
run buildall

PR #45126, opened 2024-12-06
[fix](mysql-buffer) fix special buffer size with nested type

amorynan:
in special table with sepcial data insert into 
we could make sql with mysql return Malformed packet
```
mysql> select * from use1r;
ERROR 2027 (HY000): Malformed packet
```
this pr can resolve this problem
Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Release note

None

### Check List (For Author)

- Test 
    - [x] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label


amorynan:
run buildall

be/src/util/mysql_row_buffer.cpp
```diff
@@ -107,7 +107,11 @@ MysqlRowBuffer<is_binary_format>::~MysqlRowBuffer() {
 template <bool is_binary_format>
 void MysqlRowBuffer<is_binary_format>::open_dynamic_mode() {
     if (!_dynamic_mode) {
-        *_pos++ = NEXT_EIGHT_BYTE;
+        // if _pos now exactly at the end of _buf memory,
+        // we should reserve 1 byte for _dynamic_mode flag byte to avoid *pos = 254
+        // cause _dynamic_mode flag byte be overwritten
+        reserve(1);
+        *_pos++ = NEXT_EIGHT_BYTE; // *_pos = 254 ; _pos++
```
xiaokang, line 114: Can you check that is there any more place should reserve() be called before *_pos++ ?

amorynan: I checked it, in be/src/util/mysql_row_buffer.cpp we have only one place to make *_pos++ 
and I also checked ```push_xxx``` func with reserve func or not , and anything else have I missed ？

eldenmoon (approved):
LGTM

yiguolei (approved):

PR #42416, opened 2024-10-24
[fix](bloomFilter)fix data type supported behavior same in FE and BE

amorynan:
## Proposed changes
fix create table stmt for properties: bloom_filter_columns,  which should same with BE behavior 
Issue Number: close #xxx

PR #45180, opened 2024-12-09
[fix](bloomFilter)fix data type supported behavior same in FE and BE (#42416)

airborne12:
pick from #42416


airborne12:
run buildall

PR #45224, opened 2024-12-10
[fix](catalog) opt the count pushdown rule for iceberg/paimon/hive scan node (#44038)

morningman:
bp #44038


morningman:
run buildall

morningman:
run buildall

PR #45232, opened 2024-12-10
[Refactor](query) refactor lock in fragment mgr and change std::unorder_map to phmap #44821

zhiqiang-hhhh:
cherry pick from #44821


zhiqiang-hhhh:
run buildall

PR #45416, opened 2024-12-13
[fix](arrow-flight-sql) Fix arrow flight sql query profile and correct close

xinyiZzz:

1. Fix the bug that caused `finalizeQuery` to fail, the query status in the profile was always `RUNNING`, not `OK`.
2. Arrow Flight is a stateless protocol, FE cannot know when the Client has completed fetching data. therefore, when the query profile returns `isDone=true`, it is considered that the query has ended and executes `finalizeArrowFlightSqlRequest`.
3. After the above fix we can get an accurate query profile, so I did some performance testing again and got the following conclusion (I had the same conclusion on Arrow 15.0.2)
```
// Note: The time cost may be much longer than the actual query execution time.
// You may get `cost: 700ms` here, but the query time in Doris FE's query profile is only `20ms`.
// It may be that the process of Java generating QueryResult is too slow,
// Does Arrow Flight do row-column conversion when generating QueryResult?
// Because the time taken to generate ResultSet is similar to that of Jdbc DriverManager,
// is Arrow cheating us? :) which is worth further research.
// In addition, the speed of AdbcStatement/QueryResult generated by JdbcDriver seems to be
// faster than FlightSqlDriver, this seems like a joke.
```

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [x] Other reason 

- Behavior changed:
    - [ ] No.
    - [x] Yes. 

    - [ ] No.
    - [x] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label


Thearas:

xinyiZzz:
run buildall

xinyiZzz:
run buildall

PR #45575, opened 2024-12-18
[fix](suites) Enable common expr pushdown in the restore inverted idx case #45547

w41ter:
cherry pick from #45547


w41ter:
run buildall

PR #46205, opened 2024-12-31
[fix](catalog) Fix mark handling for failed tasks to maintain getLeftMarks

w41ter:

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

Many pieces of code use `getLeftMarks` to check whether a task is finished. Therefore, when a task fails, its mark should not be deleted directly. For example, when creating a tablet, the left marks are used to track how many replicas were successfully created. If the number of successful replicas exceeds the quorum, the tablet is considered successfully created. However, if the mark is deleted due to failure, it could lead to a situation where the majority of replicas fail to be created, but the tablet is still mistakenly considered as successfully created.

This PR addresses the issue by saving the mark of the failed task in a separate map while maintaining the method's idempotency.

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/common/MarkedCountDownLatch.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/common/MarkedCountDownLatch.java
@@ -28,11 +28,13 @@ import java.util.concurrent.CountDownLatch;
 public class MarkedCountDownLatch<K, V> extends CountDownLatch {
 
     private Multimap<K, V> marks;
+    private Multimap<K, V> failedMarks;
     private Status st = Status.OK;
 
     public MarkedCountDownLatch(int count) {
         super(count);
         marks = HashMultimap.create();
+        failedMarks = HashMultimap.create();
     }
 
     public synchronized void addMark(K key, V value) {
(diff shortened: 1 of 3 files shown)


hello-stephen:

w41ter:
run buildall

fe/fe-core/src/main/java/org/apache/doris/common/MarkedCountDownLatch.java
```diff
@@ -28,11 +28,13 @@
 public class MarkedCountDownLatch<K, V> extends CountDownLatch {
 
     private Multimap<K, V> marks;
+    private Multimap<K, V> failedMarks;
```
yujun777, line 31: marks and failedMarks's had overlap effect  where they maybe both respent a failure;

use  com.google.common.collect.Table<K, V, Options<Boolean>> marks ?

for its Options<Boolean> value:
1.  equals Options.empty()  =>  no response;
2. Options.of(true) => reponse and succ;
3. Options.of(false) => response and failed;

then for function:
1. markedCountDown => Options.empty() => Options.of(true)
2. markedCountDownStatus => Options.empty() =>  Options.of(status.ok ? true : of)
3. getLeftMarks => those Options.empty() || Options.of(false);

w41ter: Currently, there is no specific requirement to differentiate between failed and unfinished cases, and countDownToZero may not be able to distinguish between them fully. Perhaps it would be worth considering your proposed solution if the need arises in the future.

w41ter: This PR needs to be picked to earlier versions, so keeping the changes as simple as possible will facilitate code review during the pick process. I think the optimizations can be addressed in a subsequent PR.

deardeng (approved):
LGTM

gavinchou (approved):

PR #46629, opened 2025-01-08
[fix](spark-connector) fix selected_columns in the response of open_scanner result is inconsistent with the data order.

gnehil:
… order

Issue Number: close #xxx

Related PR: #39627 

Problem Summary:

When the Spark connector reads Doris, it first requests `/_query_plan` to obtain the execution plan. Since the previous pr updated the planner to the Nereids planner, the tuple descriptor in the plan fragment has two parts, one is the tuple output by the `scan node`, the order of which is consistent with the column order in the table, and the other is the tuple output by the `result sink`, the order of which is consistent with the select order.
The difference is that there is only one tuple descriptor in the query plan generated by the old planner, so When executing a query and the column order in the select is inconsistent with that defined in the table, the connector requests be to call the `open_scanner` interface to execute the `exec_external_plan_fragment` method, and by default goes back to obtain the column order in the first tuple descriptor and fills in the selected_columns attribute, which causes the spark connector to have a type mismatch when obtaining data in the arrow vector according to selected_columns.

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/be/src/runtime/fragment_mgr.cpp
+++ b/be/src/runtime/fragment_mgr.cpp
@@ -1208,7 +1208,8 @@ Status FragmentMgr::exec_external_plan_fragment(const TScanOpenParams& params,
                "processed";
         return Status::InvalidArgument(msg.str());
     }
-    TupleDescriptor* tuple_desc = desc_tbl->get_tuple_descriptor(0);
+    auto output_tuple_id = t_query_plan_info.plan_fragment.plan.nodes.front().output_tuple_id;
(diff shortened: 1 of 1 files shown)


hello-stephen:

gnehil:
run buildall

PR #47144, opened 2025-01-17
[Improve](audit) retry when load failed

qzsee:

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

When the audit log load failed due to be downtime or other reasons, we need to try again instead of discard it directly

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [x] This is a refactor/code format and no logic has been changed.
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/plugin/audit/AuditLoader.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/plugin/audit/AuditLoader.java
@@ -20,6 +20,7 @@ package org.apache.doris.plugin.audit;
 import org.apache.doris.catalog.Env;
 import org.apache.doris.common.util.DigitalVersion;
 import org.apache.doris.common.util.TimeUtils;
+import org.apache.doris.metric.MetricRepo;
 import org.apache.doris.plugin.AuditEvent;
 import org.apache.doris.plugin.AuditPlugin;
 import org.apache.doris.plugin.Plugin;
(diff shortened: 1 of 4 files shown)


Thearas:

fe/fe-core/src/main/java/org/apache/doris/plugin/audit/AuditLoader.java
```diff
@@ -184,24 +185,29 @@ public synchronized void loadIfNecessary(boolean force) {
...
-                    return;
-                }
-                AuditStreamLoader.LoadResponse response = streamLoader.loadBatch(auditLogBuffer, token);
-                if (LOG.isDebugEnabled()) {
-                    LOG.debug("audit loader response: {}", response);
-                }
-            } catch (Exception e) {
-                if (LOG.isDebugEnabled()) {
-                    LOG.debug("encounter exception when putting current audit batch, discard current batch", e);
+                while (true) {
```
morningman, line 188: There are some issues:
1. `while(true)` is dangerous. If there is no more audit log coming and load fail due to bug, this will run forever.
2. The cost of `exceedAuditDataSize()` is too high, it calls `fillLogBuffer()` duplicately for each audit log, and there may be tens of thousands of audit log in queue because you sleep 10 seconds after failure.

morningman: I suggest:
1. At least add a `max_retry_time`
2. think twice about calling `fillLogBuffer()` in `exceedAuditDataSize ()`, can we avoid that?

By the way, there is one more important thing about audit log,
which is how to make sure that the audit log in `fe.audit.log` file is complete? Because despite of the audit table, we even can't guarantee the integrity of `fe.audit.log`.

morningman: Also, we need add some UT to make sure this retry logic works well

PR #47375, opened 2025-01-23
[improve](cloud-mow) reduce ms update_delete_bitmap kv confict

mymeiyi:

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

1. `update_delete_bitmap` may split to several transactions to avoid delete bitmap size is larger than the fdb transaction limit
2. multi compaction jobs will change the initiators of the lock_info, which will cause txn_conflict of `update_delete_bitmap`. 
3. for update with multi transactions, the txn_confict error is more easily to happen, even after some retries, the `update_delete_bitmap` will fail
4. the root cause is multi compactions should not conflict, pr 48024 solve it
5. but branch-3.0 does not contain pr 48024, so modify the check lock_id to snapshot read to avoid txn_conflict. if lock_id is changed, the final `commit_txn` or `commit_job` can handle it

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/cloud/src/meta-service/meta_service.cpp
+++ b/cloud/src/meta-service/meta_service.cpp
@@ -1749,10 +1749,12 @@ void MetaServiceImpl::get_tablet_stats(::google::protobuf::RpcController* contro
 static bool check_delete_bitmap_lock(MetaServiceCode& code, std::string& msg, std::stringstream& ss,
                                      std::unique_ptr<Transaction>& txn, int64_t table_id,
                                      int64_t lock_id, int64_t lock_initiator, std::string& lock_key,
-                                     DeleteBitmapUpdateLockPB& lock_info) {
+                                     DeleteBitmapUpdateLockPB& lock_info,
+                                     bool snapshot_read = false) {
     std::string lock_val;
-    LOG(INFO) << "check_delete_bitmap_lock, table_id=" << table_id << " key=" << hex(lock_key);
-    auto err = txn->get(lock_key, &lock_val);
+    LOG(INFO) << "check_delete_bitmap_lock, table_id=" << table_id << " key=" << hex(lock_key)
+              << ", lock_id=" << lock_id << ", snapshot_read=" << snapshot_read;
+    auto err = txn->get(lock_key, &lock_val, snapshot_read);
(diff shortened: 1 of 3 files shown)


Thearas:

cloud/src/meta-service/meta_service.cpp
```diff
@@ -1685,7 +1685,7 @@ static bool check_delete_bitmap_lock(MetaServiceCode& code, std::string& msg, st
     std::string lock_key = meta_delete_bitmap_update_lock_key({instance_id, table_id, -1});
     std::string lock_val;
     DeleteBitmapUpdateLockPB lock_info;
-    auto err = txn->get(lock_key, &lock_val);
```
zhannngchen, line 1688: add a constexpr for compaction lock id, rather than magic number

zhannngchen: like this：
```
static constexpr int COMPACTION_DELETE_BITMAP_LOCK_ID = -1;
```

cloud/src/meta-service/meta_service.cpp
```diff
@@ -1685,7 +1685,7 @@ static bool check_delete_bitmap_lock(MetaServiceCode& code, std::string& msg, st
...
+    auto err = txn->get(lock_key, &lock_val, lock_id == -1 /*snapshot*/);
```
cosven, line 1688: We should add some unit/p0 tests to verify this branch.

At least, the following test case should be added
image

cosven: Besides, the following cases should also be added
* compaction-load conflict
* compaction-compaction conflict
* ??? load-compaction conflict

mymeiyi:
run buildall

dataroaring (approved):
LGTM

zhannngchen (approved):

PR #48017, opened 2025-02-18
Revert "[fix](java) should use JAVA_OPTS_FOR_JDK_17 instead of JAVA_OPTS"

morningman:
Reverts apache/doris#47913
Before this PR, the BE's JVM heap size does not follow the JAVA_OPTS_FOR_JDK17 in be.conf,
it use default value, which is 1/4 of physical memory.
After this PR, the BE's JVM heap size is set to 2GB, and it cause some test pipeline OOM.
So I revert this PR temporarily to try investigate the usage of BE's JVM first.

### DIFF
--- a/be/src/util/jni-util.cpp
+++ b/be/src/util/jni-util.cpp
@@ -107,8 +107,7 @@ const std::string GetKerb5ConfPath() {
     setenv("CLASSPATH", classpath.c_str(), 0);
 
     // LIBHDFS_OPTS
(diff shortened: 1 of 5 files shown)


morningman:
run buildall

CalvinKirs (approved):

wuwenchi (approved):

kaka11chen (approved):
LGTM

PR #48022, opened 2025-02-18
[fix](Export) change the export logical

BePPPower:
Problem Summary:

Previously, Export would split the task into multiple threads, and each thread would further split the task into multiple outfiles, which were executed sequentially.
Each thread will decide how many outfiles to split into according to `maximum_tablets_of_outfile_in_export` and the actual number of partitions/buckets of the data.
Each thread executes multiple outfiles sequentially, with each Outfile task waiting for the previous one to complete before starting. This not only reduces export performance but also worsens the user experience, as users cannot clearly determine the actual concurrency performance of an Export Job.

Now, this logic has been simplified, and each thread will only handle one outfile without splitting it into multiple Outfile tasks. The parallelism value specified by the user in the `EXPORT` statement corresponds to the number of threads, with each thread handling one Outfile statement. This not only improves the user experience but also increases CPU utilization.

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [x] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [x] Yes. 

    - [ ] No.
    - [x] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/load/ExportTaskExecutor.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/load/ExportTaskExecutor.java
@@ -54,7 +54,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 public class ExportTaskExecutor implements TransientTaskExecutor {
     private static final Logger LOG = LogManager.getLogger(ExportTaskExecutor.class);
 
-    List<StatementBase> selectStmtLists;
+    Optional<StatementBase> selectStmt;
 
     ExportJob exportJob;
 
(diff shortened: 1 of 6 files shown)


BePPPower:
run buildall

fe/fe-core/src/main/java/org/apache/doris/load/ExportTaskExecutor.java
```diff
@@ -89,16 +89,14 @@ public void execute() throws JobException {
         LOG.debug("[Export Task] taskId: {} updating state to EXPORTING", taskId);
         exportJob.updateExportJobState(ExportJobState.EXPORTING, taskId, null, null, null);
         List<OutfileInfo> outfileInfoList = Lists.newArrayList();
-        for (int idx = 0; idx < selectStmtLists.size(); ++idx) {
-            LOG.debug("[Export Task] taskId: {} processing statement {}/{}",
-                    taskId, idx + 1, selectStmtLists.size());
+        if (selectStmt.isPresent()) {
```
morningman, line 92: In what situation the selectStmt can be null?

BePPPower: If we export table without data, we would construct an `ExportTaskExecuto`r with empty `selectStmt`

morningman (approved):

wuwenchi (approved):

CalvinKirs (approved):

zy-kkk (approved):

PR #48093, opened 2025-02-19
Revert "[fix](fold) fixed an issue with be computing constants"

zhangstar333:
Reverts apache/doris#43410 @felixwluo 
this PR not solve the issue
it's return the right result as it's thrown exception, so the fold failed, then execute the whole plan by BE, and execute SQL.
`2025-02-19 17:45:49,859 WARN (mysql-nio-pool-0|236) [FoldConstantRuleOnBE.evalOnBE():346] query b37a233e96e64721-81d9ea4d2ec4899e failed to get const expr value from be: Index: 0`.

 and now all return string fold function will be failed. as resultContent can't call getBytesValue function
 because BE put the result into string value, it's should call getStringValue function
and in another PR will solve the problem https://github.com/apache/doris/pull/48096


Thearas:

zhangstar333:
run buildall

zhangstar333:
run buildall

zhangstar333:
run buildall

regression-test/suites/nereids_p0/expression/fold_constant/fold_constant_by_be.groovy
```diff
@@ -22,9 +22,6 @@ suite("fold_constant_by_be") {
     sql 'set enable_fallback_to_original_planner=false'
     sql 'set enable_fold_constant_by_be=true'
 
-    qt_sql """ select hex(from_base64('wr2JEDVXzL9+2XtRhgIloA==')); """
```
LiBinfeng-01, line 25: would this case behaviour reverted? I think we should keep this case

PR #46648, opened 2025-01-08
[fix](inverted index) Ensure that col_unique_ids in TabletIndex are not empty.

zzzxl1993:
### What problem does this PR solve?

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:
1. Ensure that col_unique_ids in TabletIndex are not empty when TabletIndex exists after a schema change.
2. Use unique_id to search for and complete index schema information.

### Release note

None

### Check List (For Author)

- Test 
    - [x] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

- Does this need documentation?
    - [x] No.
    - [ ] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #42882, opened 2024-10-29
[fix](build index)Fix build index failed on renamed column

qidaye:
## Proposed changes

Add `column_unique_ids` in `TOlapTableIndex` thrift struct, to get col_unique_id when building index.

PR #48988, opened 2025-03-12
[fix](Index)Make column unique ids in index dynamically computed

### LINKED ISSUE #48989: [Bug] dynamic partition is not work
### Search before asking

### Version

3.0.4

### What's Wrong?

```
mysql> alter table ods_kafka_ai_chat_prompt_rt add partition p20250316 values [('2025-03-16 00:00:00'),('2025-03-17 00:00:00'));
ERROR 1105 (HY000): Unexpected exception: Cannot invoke "java.util.List.iterator()" because "this.columnUniqueIds" is null
```

### What You Expected?

Add partition normally.

### How to Reproduce?

Create table in 3.0.2.
```
CREATE TABLE `ods_kafka_ai_chat_prompt_rt` (
  `msg_id` largeint NULL COMMENT "消息id",
  `api_key` varchar(2048) NULL COMMENT "apiKey",
  `api_key_alias` varchar(4096) NULL COMMENT "apiKey 别名",
  `model` varchar(2048) NULL COMMENT "模型",
  `messages` text NULL COMMENT "消息内容",
  `qpm` int NULL COMMENT "每分钟请求数限制",
  `temperature` decimal(10,2) NULL,
  `max_tokens` int NULL COMMENT "每分钟请求数限制",
  `token_per_min` bigint NULL COMMENT "每分钟token限额",
  `token_per_msg` bigint NULL COMMENT "单条token限额",
  `create_time` datetime NULL COMMENT "创建时间",
  `update_time` datetime NULL COMMENT "修改时间",
  INDEX idx_msg_id (`msg_id`) USING INVERTED,
  INDEX idx_api_key (`api_key`) USING INVERTED,
  INDEX idx_api_key_alias (`api_key_alias`) USING INVERTED,
  INDEX idx_model (`model`) USING INVERTED
) ENGINE=OLAP
DUPLICATE KEY(`msg_id`, `api_key`, `api_key_alias`, `model`)
PARTITION BY RANGE(`create_time`)()
DISTRIBUTED BY RANDOM BUCKETS AUTO
PROPERTIES (
"file_cache_ttl_seconds" = "0",
"bloom_filter_columns" = "api_key, model, msg_id, api_key_alias",
"is_being_synced" = "false",
"dynamic_partition.enable" = "true",
"dynamic_partition.time_unit" = "DAY",
"dynamic_partition.time_zone" = "Asia/Shanghai",
"dynamic_partition.start" = "-2147483648",
"dynamic_partition.end" = "3",
"dynamic_partition.prefix" = "p",
"dynamic_partition.buckets" = "10",
"dynamic_partition.create_history_partition" = "false",
"dynamic_partition.history_partition_num" = "-1",
"dynamic_partition.hot_partition_num" = "0",
"dynamic_partition.reserved_history_periods" = "NULL",
"storage_medium" = "hdd",
"storage_format" = "V2",
"inverted_index_storage_format" = "V2",
"light_schema_change" = "true",
"disable_auto_compaction" = "false",
"enable_single_replica_compaction" = "false",
"group_commit_interval_ms" = "10000",
"group_commit_data_bytes" = "134217728"
); 
```

Upgrade to 3.0.4

Then dynamic partition will not work.

```

ALTER TABLE ods_kafka_ai_chat_prompt_rt SET ("dynamic_partition.enable" = "false");
alter table ods_kafka_ai_chat_prompt_rt add partition p20250316 values [('2025-03-16 00:00:00'),('2025-03-17 00:00:00'));
```

### Anything Else?

_No response_

### Code of Conduct

qidaye:

Issue Number: close #48989

Related PR: #46648 #42882

Problem Summary:

We introduced `Index.columnUniqueIds` in 3.0.3, and make sure `Index.columnUniqueIds` is not empty in 3.0.4.
But when we upgrade Doris from 3.0.2 and before to 3.0.4, the `Index.columnUniqueIds` will be `null`.

Versions 302 and earlier do not have this variable, and gson deserialization will assign it to null.
We need to initialize the column unique ids associated with the index correctly.

Key Changes

Removed stored `columnUniqueIds` from `Index` class:

Previously, column unique IDs were stored as a field in the `Index` class
Now they're dynamically computed when needed rather than stored redundantly
Added dynamic computation method:

New method `getColumnUniqueIds(List<Column> schema)` computes IDs at runtime
Column IDs are looked up by matching index column names with schema columns

Updated method signatures:

Changed `toThrift()` to `toThrift(List<Integer> indexColumnUniqueIds)`
Changed `toPb()` to `toPb(Map<Integer, Column> columnMap, List<Integer> indexColumnUniqueIds)`
Updated all callers to pass column IDs as parameters

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [x] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label


qidaye:
run buildall

airborne12 (commented):
need ut

fe/fe-core/src/main/java/org/apache/doris/catalog/Index.java
```diff
@@ -50,7 +51,7 @@
  * Internal representation of index, including index type, name, columns and comments.
  * This class will be used in olap table
  */
-public class Index implements Writable {
+public class Index implements Writable, GsonPostProcessable {
```
airborne12, line 54: what's the meaning of GsonPostProcessable?

qidaye: GsonPostProcessable can process more actions after read by GSON, so we can initialize uniqueIds if it is null.

fe/fe-core/src/main/java/org/apache/doris/datasource/InternalCatalog.java
```diff
@@ -1766,6 +1766,8 @@ public PartitionPersistInfo addPartition(Database db, String tableName, AddParti
             if (!isCreateTable) {
                 beforeCreatePartitions(db.getId(), olapTable.getId(), partitionIds, indexIds, isCreateTable);
             }
+            // init column unique id for indexes
+            olapTable.initIndexColumnUniqueId();
```
yujun777, line 1770: why recreate column uniqud id list for every time creating a partition ?  if every time recreating, it should not use a  variable to store it,  instead should use a Getter function from OlapTable.

qidaye: It's not recreating. We set the column unique ids for index to make sure the index is associated with column unique id.
Because the older version index dose not have the columnUniqueIds filed, when Doris upgrades to new version which have the columnUniqueIds, columnUniqueIds will be initialized to empty list.

qidaye: I'll add a check to prevent duplicated adding.

fe/fe-core/src/main/java/org/apache/doris/catalog/MaterializedIndexMeta.java
```diff
@@ -401,6 +401,18 @@ public void initSchemaColumnUniqueId() {
         });
     }
 
+    public void initIndexColumnUniqueId() {
```
airborne12, line 404: Need to check if columnUniqueIds already exists

fe/fe-core/src/main/java/org/apache/doris/datasource/InternalCatalog.java
yujun777: OlapTable.gsonPostProcess can handle the oldVersions ?

fe/fe-core/src/main/java/org/apache/doris/catalog/Index.java
```diff
@@ -260,18 +268,22 @@ public TOlapTableIndex toThrift() {
...
 
     public OlapFile.TabletIndexPB toPb(Map<Integer, Column> columnMap) {
         OlapFile.TabletIndexPB.Builder builder = OlapFile.TabletIndexPB.newBuilder();
         builder.setIndexId(indexId);
         builder.setIndexName(indexName);
-        for (Integer columnUniqueId : columnUniqueIds) {
-            Column column = columnMap.get(columnUniqueId);
-            if (column != null) {
-                builder.addColUniqueId(column.getUniqueId());
+        if (columnUniqueIds != null) {
```
airborne12, line 281: Need to throw excepation here

qidaye:
run buildall

yujun777 (approved):
LGTM

airborne12 (approved):
LGTM

dataroaring (approved):
LGTM

PR #49050, opened 2025-03-13
[fix](cloud) fix corner case when warm up data larger than cache capacity

freemandealer:
when warm up target table's first partition is larger cache capacity of the cluster/compute group, the deleted code will break and leave warmUpTotalFileCache zero, resulting no error&warning but 0 batch size.

this commit will warn user this situation and allow user use FORCE warm up to load partial data into the file cache.

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label


Thearas:

freemandealer:
run buildall

fe/fe-core/src/main/java/org/apache/doris/cloud/CacheHotspotManager.java
```diff
@@ -542,9 +542,6 @@ private Map<Long, List<Tablet>> warmUpNewClusterByTable(long jobId, String dstCl
             List<Partition> warmUpPartitions = new ArrayList<>();
             for (Partition partition : partitions) {
                 Long partitionSize = partition.getDataSize(true);
-                if ((warmUpTotalFileCache + partitionSize) > totalFileCache) {
```
gavinchou, line 545: we should move this check after line 549 instead removing them
because it may iterate over all partitions of the table which may exceed too much of the totalFileCache
it will do useless work just flushing the file cache on BE over and over again.

and also add efficient logs for this statement , say what table is to warmup, what is the job content  and whether it is a force warmup.

gavinchou (changes requested):

fe/fe-core/src/main/java/org/apache/doris/cloud/CacheHotspotManager.java
freemandealer: that is why the problem occurs.

simply move it to 549 will skip all necessary jobs below and make this job not only useless but also risky.

a better way, if we want to improve this, is to skip large one and cache little one, since the original problem is cache space is not utilized fully when doing warm up

freemandealer:
run buildall

gavinchou (approved):

dataroaring (approved):
LGTM

PR #22769, opened 2023-08-09
[fix](cache) Fix enable sql cache lead to FE Full GC or OOM

xinyiZzz:
## Proposed changes

Enable sql cache will keep the complete query results in FE during the query process. After the query is completed, the size of the results will be judged and inserted into the cache. 
This will cause FE Full GC or OOM when the query results are too large.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #22645, opened 2023-08-06
[fix](cache) add sql cache conf `cache_result_max_data_size`

xinyiZzz:
## Proposed changes

Only the maximum number of rows in sql cache `cache_result_max_row_count` is not enough. If a row of data is too large, FE may OOM.

## Further comments

If this is a relatively large or complex change, kick off the discussion at [dev@doris.apache.org](mailto:dev@doris.apache.org) by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #49424, opened 2025-03-24
[fix](sql cache) Fix enable sql cache lead to FE Full GC or OOM

cambyzju:
pick https://github.com/apache/doris/pull/22645 and https://github.com/apache/doris/pull/22769 to branch-1.2


lide-reed (approved):
LGTM

PR #49771, opened 2025-04-02
[Bug](Compaction): fix compact version miss when too many invisible version

Userwhite:

when too many invisible version, compact may choose some versions before `visible version`, and choose some versions after `visible_version + keep_invisible_version_limit`, it will lead that check version miss fail

image

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Release note

None

### Check List (For Author)

- Test 
    - [x] Regression test
    - [x] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

    - [x] No.
    - [ ] Yes. 

- [x] Confirm the release note
- [x] Confirm test cases
- [x] Confirm document
- [  ] Add branch pick label

### DIFF
--- a/be/src/olap/tablet.cpp
+++ b/be/src/olap/tablet.cpp
@@ -1247,7 +1247,8 @@ std::vector<RowsetSharedPtr> Tablet::_pick_visible_rowsets_to_compaction(
             update_time_long ? config::compaction_keep_invisible_version_min_count
                              : config::compaction_keep_invisible_version_max_count;
 
-    std::vector<RowsetSharedPtr> candidate_rowsets;
+    std::vector<RowsetSharedPtr> candidate_rowsets_visible;
+    std::vector<RowsetSharedPtr> candidate_rowsets_invisible;
     {
         std::shared_lock rlock(_meta_lock);
         for (const auto& [version, rs] : _rs_version_map) {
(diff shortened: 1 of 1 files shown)


hello-stephen:

dataroaring:
run buildall

be/src/olap/tablet.cpp
```diff
@@ -1296,8 +1296,8 @@ std::vector<RowsetSharedPtr> Tablet::_pick_visible_rowsets_to_compaction(
             // 1. had been visible;
             // 2. exceeds the limit of keep invisible versions.
             int64_t version_end = version.second;
-            if (version_end <= visible_version ||
-                version_end > visible_version + keep_invisible_version_limit) {
+            if (version_end <= visible_version || ((candidate_rowsets.empty() || candidate_rowsets.front()->version().second > visible_version)
```
yujun777: suppose max version = 1000,  visible version = 100,   keep_invisible_version_limit = 20,  then all the versions can split into 3 parts:
A. 1 ~ 100  (1 to visible version);
B. 101 ~ 120 (visible version  + 1 to visible_version + keep_visible_version_limit);
C.  121 ~ 1000;

will chose part A,  part C to compact,  but not chose part B to compact (if compact cross visible version,  sql query will throw error E-230 version had been compact ) ;

so we need to compact part A + part C,   but the problem is,  for base compaction, need the candidate rowsets continous (but not require from version 1 ),  so if the candidate rowset is all part A or all part C is ok,  but if include them both will fail.

here is the fix:
use two candidate_rowset  vectors:
one is to hold  part A:  version_end <= visible version
the other is to hold part C:  version_end > visible_version + keep_invisible_version_limit

if vector size of part A >1, then return part A;
else return part C;

yujun777: there's a docker regression case 'compaction/test_compaction_with_visible_version.groovy', need to pass the docker case, but notice github pr pipeline don't run the docker case,  need run docker case in locale machine.   

how to run a docker suite can see the annotations in case 'demo_p0/docker_action.groovy'

yujun777 (changes requested):
need change

be/src/olap/tablet.cpp
Userwhite: ok, i'll fix it

Userwhite:
run buildall

Userwhite:
run buildall

PR #50160, opened 2025-04-18
[opt](nereids) Optimize type conversion strategy for ComparisonPredicate with literals

### LINKED ISSUE #50159: [Enhancement] Optimize type conversion strategy for ComparisonPredicate with literals
### Search before asking

### Description

In the production environment, the following issues were encountered:
- There exists an external Hive table `app_day` with partition fields of string type.
- When users execute the following SQL query, it takes a long time and consumes a large amount of network bandwidth on the BE:
  ```sql
  select count(*) from hive_ods.app_day where day_id = 20250411 and latn_id = 20250411
  ```
- The root cause is that the SQL is optimized to:
  ```sql
  select count(*) from hive_ods.app_day where cast(day_id as double) = 20250411 and cast(latn_id as double) = 20250411
  ```
- Due to the query conditions not being pushed down, it results in scanning all partition data.

### Solution

In a comparison operation (ComparisonPredicate), the literal on the right side is preferentially cast to the type of the expression on the left side. For example, the above SQL will be automatically optimized to:
```sql
SELECT COUNT(*) FROM hive_ods.app_day WHERE day_id = '20250411' AND latn_id = '20250411';
```
This optimization allows the query conditions to be pushed down to the underlying layer of Hive, thereby reducing the amount of data scanned and improving query efficiency.

### Code of Conduct

XLPE:

Issue Number: close #50159 

Problem Summary:
In a comparison operation (ComparisonPredicate), the literal on the right side is preferentially cast to the type of the expression on the left side.  

This optimization allows the query conditions to be pushed down to the external storage layer , thereby reducing the amount of data scanned and improving query efficiency.

### Check List (For Author)

- Test 
    - [x] Regression test
    - [x] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [x] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/nereids/util/TypeCoercionUtils.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/nereids/util/TypeCoercionUtils.java
@@ -986,7 +986,7 @@ public class TypeCoercionUtils {
             }
             return comparisonPredicate.withChildren(left, right);
         }
-            
+
         // try to convert the right literal to left type.
         if (isRightLiteralConvertible(left, right)) {
             if (!supportCompare(left.getDataType())) {
@@ -1143,6 +1143,7 @@ public class TypeCoercionUtils {
 
     /**
      * check should downgrade from commonTypeClazz to targetTypeClazz.
+     *
      * @param commonTypeClazz before downgrade type
      * @param targetTypeClazz try to downgrade to type
      * @param commonType original common type
@@ -1150,7 +1151,6 @@ public class TypeCoercionUtils {
      * @param commonTypePredicate constraint for original type
      * @param otherPredicate constraint for other expressions aka literals
      * @param others literals
-     *
      * @return true for should downgrade
      */
     private static boolean shouldDowngrade(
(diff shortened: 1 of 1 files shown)


hello-stephen:

starocean999:
string and integer type's common type is double in Doris. So if we compare the two types, doris will cast them to double. It's better to modify sql to avoid the auto-cast behavior in doris. like you mentioned `SELECT COUNT(*) FROM hive_ods.app_day WHERE day_id = '20250411' AND latn_id = '20250411';` And because the implicit cast logical is a unified behavior in doris, we can not change the current behavior as this PR.

XLPE:
Under the scenario described in this issue, we found that the current implicit conversion strategy poses significant risks in production environments and is not the optimal implementation. 
Rather than having users modify SQL after a failure occurs, it would be better to replace implicit conversions with explicit type checking and directly return type error messages. This approach can help identify issues earlier and prevent potential production failures.
This PR primarily optimizes the potential side effects caused by implicit conversions, ensuring that the generated SQL statements better align with expected behavior.
I manually tested numerous type conversions and found no other issues in the final generated SQL.

starocean999:
The common type of string and integer has been double since very early release of Doris, and many users may rely on this for a long time. It will be risky if we change the behavior now. But the implicit conversions of Doris may not be optimal as you mentioned and this whole implicit conversion system will be redesigned in near future.

PR #50301, opened 2025-04-22
[Fix](predefine) add many cases port from original cases

eldenmoon:
1. If the path is typed, it means the path is not a sparse column, so we can't read the sparse column even if the sparse column size is reached limit,so should not use SparseColumnExtractReader for typed column
2. handle Int128 in visitor correctly
3. ScalarType should not be modified


hello-stephen:

PR #50547, opened 2025-04-29
[fix](lexer) fix unclosed comment handling

Baymine:

The previous lexer rule for `BRACKETED_COMMENT` in `DorisLexer.g4` incorrectly handled unclosed comments (comments starting with `/*` but not ending with `*/` before the end of the input).

When an unclosed comment occurred, the rule's alternative ending condition (`| {markUnclosedComment();} EOF`) allowed the rule to **successfully match** by consuming all subsequent characters up to the end of the file (EOF) into a single `BRACKETED_COMMENT` token. This effectively hid valid SQL code following the unclosed comment start from the parser.

For example, consider the following SQL:
```sql
select * from t1 /*this is /*a comment*/ limit 1;
```
Under the old rule:
1.  The lexer started matching `BRACKETED_COMMENT` at the first `/*`.
2.  It recursively handled the inner `/*a comment*/`.
3.  It continued trying to find the closing `*/` for the *outer* comment.
4.  Upon reaching EOF, it matched the `| {markUnclosedComment();} EOF` condition.
5.  The *entire* string `/*this is /*a comment*/ limit 1;` was consumed as **one** `BRACKETED_COMMENT` token.
6.  The parser only saw `SELECT * FROM t1`, leading to the incorrect result of returning all rows instead of just one.

Problem Summary:
The primary goal of this change is to correctly support **unclosed nested bracketed comments** (e.g., `/* outer /* inner */`) within SQL statements parsed by the Doris Nereids lexer. This aligns with standard SQL practices, including MySQL, which allows such nesting.

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

```sql
-- setting
CREATE TABLE IF NOT EXISTS t1 (id INT) DISTRIBUTED BY HASH(id) BUCKETS 1 PROPERTIES ('replication_num' = '1');
INSERT INTO t1 VALUES (1), (2), (3);

-- Test Case : Unclosed Comment (Original Bug Case - Nested)
select * from t1 /* outer /* inner */ limit 1;
-- EXPECTED: return 1 row
-- OBSERVED: return 3 rows

-- Test Case : Valid query using table after nested comment
select * from t1 /* outer /* inner */ limit 1;
-- EXPECTED: Returns 1 row.
-- OBSERVED: Returns 1 row.

```

- Behavior changed:
    - [ ] No.
    - [] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/fe/fe-core/src/main/antlr4/org/apache/doris/nereids/DorisLexer.g4
+++ b/fe/fe-core/src/main/antlr4/org/apache/doris/nereids/DorisLexer.g4
@@ -20,10 +20,6 @@
 lexer grammar DorisLexer;
 
 @members {
-  /**
-   * When true, parser should throw ParseExcetion for unclosed bracketed comment.
-   */
-  public boolean has_unclosed_bracketed_comment = false;
 
   /**
    * Verify whether current token is a valid decimal token (which contains dot).
@@ -47,15 +43,6 @@ lexer grammar DorisLexer;
     }
   }
 
-  /**
-   * This method will be called when the character stream ends and try to find out the
-   * unclosed bracketed comment.
-   * If the method be called, it means the end of the entire character stream match,
-   * and we set the flag and fail later.
-   */
-  public void markUnclosedComment() {
-    has_unclosed_bracketed_comment = true;
-  }
 }
 
 SEMICOLON: ';';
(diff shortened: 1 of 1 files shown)


hello-stephen:

PR #49430, opened 2025-03-25
[Fix](function) Fix unexpected result of unix_timestamp when input is out of bound

zclllyybb:
### What problem does this PR solve?

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

before:
```sql
mysql> SELECT UNIX_TIMESTAMP('9999-12-30 23:59:59.999');
+-------------------------------------------+
| UNIX_TIMESTAMP('9999-12-30 23:59:59.999') |
+-------------------------------------------+
|                                     0.999 |
+-------------------------------------------+
```
now:
```sql
mysql> SELECT UNIX_TIMESTAMP('9999-12-30 23:59:59.999');
+-------------------------------------------+
| UNIX_TIMESTAMP('9999-12-30 23:59:59.999') |
+-------------------------------------------+
|                                     0.000 |
+-------------------------------------------+
```

### Release note

Fix unexpected result of `unix_timestamp` when input is out of bound

### Check List (For Author)

- Test 
    - [x] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

- Does this need documentation?
    - [x] No.
    - [ ] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #50610, opened 2025-05-06
[branch-3.0](function) Fix unexpected result of unix_timestamp when input is out of bound (#49430)

zclllyybb:
pick https://github.com/apache/doris/pull/49430 but removed be-ut because we can't.


zclllyybb:
run buildall

dataroaring (approved):
LGTM

PR #51213, opened 2025-05-24
[vector search] Step forward on stability and functionality

zhiqiang-hhhh:
A huge step forward on stability and functionality.

### Functionality
1. Search parameters like `ef_search`, can be passed to index as session variables. This behavior is same with pg-vector and duckdb vector search plug-in.
2. Correct processing for order by desc. Fallback to brute force search when it is necessary.
3. Support using inner product as index metric and order by inner_product.
4. When metrics of sql dismatches with index, fallback to brute force.

### Stability
1. More unit test
2. Virtual column iterator.
3. According to custom script, result of range search, topn search & compound search is almost same with native faiss. The overlap rate of result is more than 90%. The 10% difference is introduced by batch insert mode of native faiss.


hello-stephen:

PR #54226, opened 2025-08-01
[fix](replica) handle replica status drop as not alive

dataroaring:
Drop is set by user manually, so query should not be scheduled to it.

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/catalog/Replica.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/catalog/Replica.java
@@ -858,7 +858,7 @@ public class Replica {
     public boolean isAlive() {
         return getState() != ReplicaState.CLONE
                 && getState() != ReplicaState.DECOMMISSION
-                && !isBad();
+                && !isBad() && !isUserDrop();
     }
 
     public boolean isScheduleAvailable() {


hello-stephen:

dataroaring:
run buildall

gavinchou (approved):

deardeng (approved):
LGTM

fe/fe-core/src/main/java/org/apache/doris/catalog/Replica.java
```diff
@@ -858,7 +858,7 @@ public boolean isScaleInDrop() {
     public boolean isAlive() {
         return getState() != ReplicaState.CLONE
                 && getState() != ReplicaState.DECOMMISSION
-                && !isBad();
+                && !isBad() && !isUserDrop();
```
yujun777, line 861: isUserDrop means the replica going to be relocate,  just like decommission a be will relocate all its replicas.
notice when decommission a be,  all its normal replicas are still alive,   for the same logic,  when set a replica drop,  if this replica state is  still normal,  it must be  alive.

cannot change a userDrop replica with normal state to non-alive,   for example,   for replication_num = 1,  when we set the only one replica as userDrop,  if we think this replica is non-alive,  then  the tablet will have 0 alive replica,  and this tablet's status will be unrecoverable,   and the tabletScheduler wouldn't sched it since it's unrecoverable

yujun777 (changes requested):
don't change,  need close this pr

fe/fe-core/src/main/java/org/apache/doris/catalog/Replica.java
yujun777: set a replica drop just like decommission a be,   there replicas still queryable,   it want reading prefer skip them,  can modify function Tablet.getQueryableReplicas() to prefer  chose other queryable replicas

PR #54363, opened 2025-08-06
[fix](filecache) warm up download data less than actual

freemandealer:
the offset calculation here is buggy. we should use task-level offset to calculate download size other than file-level offset. otherwise, when offset is not zero, file download size is incorrect, causing data download is less than actual.

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [x] Previous test can cover this change.
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/be/src/io/cache/block_file_cache_downloader.cpp
+++ b/be/src/io/cache/block_file_cache_downloader.cpp
@@ -265,10 +265,12 @@ void FileCacheBlockDownloader::download_segment_file(const DownloadFileMeta& met
 
     std::unique_ptr<char[]> buffer(new char[one_single_task_size]);
 
+    size_t task_offset = 0;
     for (size_t i = 0; i < task_num; i++) {
-        size_t offset = meta.offset + i * one_single_task_size;
-        size_t size =
-                std::min(one_single_task_size, static_cast<size_t>(meta.download_size - offset));
+        size_t offset = meta.offset + task_offset;
+
+        size_t size = std::min(one_single_task_size,
+                               static_cast<size_t>(meta.download_size - task_offset));
         size_t bytes_read;
         VLOG_DEBUG << "download_segment_file, path=" << meta.path << ", read_at offset=" << offset
                    << ", size=" << size;
@@ -285,6 +287,7 @@ void FileCacheBlockDownloader::download_segment_file(const DownloadFileMeta& met
             g_file_cache_download_failed_num << 1;
             return;
         }
+        task_offset += size;
         g_file_cache_download_finished_size << size;
     }
 


freemandealer:
run buildall

hello-stephen:
# BE UT Coverage Report
Increment line coverage `0.00% (0/5)` :tada:

[Increment coverage report](http://coverage.selectdb-in.cc/coverage/6f10a2ed621a1dc77459b3201bba5fc4e3192586_6f10a2ed621a1dc77459b3201bba5fc4e3192586/increment_report/index.html)
[Complete coverage report](http://coverage.selectdb-in.cc/coverage/6f10a2ed621a1dc77459b3201bba5fc4e3192586_6f10a2ed621a1dc77459b3201bba5fc4e3192586/report/index.html)
| Category          | Coverage           |
|-------------------|--------------------|
| Function Coverage | 58.23% (16357/28091) |
| Line Coverage     | 47.14% (147796/313553) |
| Region Coverage   | 36.08% (110529/306382) |
| Branch Coverage   | 38.97% (49075/125926) |

dataroaring (approved):
LGTM

be/src/io/cache/block_file_cache_downloader.cpp
```diff
@@ -265,10 +265,12 @@ void FileCacheBlockDownloader::download_segment_file(const DownloadFileMeta& met
 
     std::unique_ptr<char[]> buffer(new char[one_single_task_size]);
 
+    size_t task_offset = 0;
     for (size_t i = 0; i < task_num; i++) {
-        size_t offset = meta.offset + i * one_single_task_size;
-        size_t size =
-                std::min(one_single_task_size, static_cast<size_t>(meta.download_size - offset));
+        size_t offset = meta.offset + task_offset;
```
gavinchou, line 270: add unit test for this function using syncpoint to construct  meta and reader (use local FS & filereader instead)

be/src/io/cache/block_file_cache_downloader.cpp
```diff
@@ -285,6 +287,7 @@ void FileCacheBlockDownloader::download_segment_file(const DownloadFileMeta& met
             g_file_cache_download_failed_num << 1;
             return;
         }
+        task_offset += size;
```
gavinchou, line 290: it actually should be `bytes_read`? make a `DCHECK_EQ(bytes_read, size)` and return if it is not true.

be/src/io/cache/block_file_cache_downloader.cpp
```diff
@@ -265,10 +265,12 @@ void FileCacheBlockDownloader::download_segment_file(const DownloadFileMeta& met
...
 
+    size_t task_offset = 0;
     for (size_t i = 0; i < task_num; i++) {
-        size_t offset = meta.offset + i * one_single_task_size;
-        size_t size =
-                std::min(one_single_task_size, static_cast<size_t>(meta.download_size - offset));
+        size_t offset = meta.offset + task_offset;
+
+        size_t size = std::min(one_single_task_size,
+                               static_cast<size_t>(meta.download_size - task_offset));
```
gavinchou, line 273: seems cannot use `meta.download_size` here, should use `download_size` initiated at line 263?

gavinchou (changes requested):

be/src/io/cache/block_file_cache_downloader.cpp
freemandealer: yes, I also noticed that and I found they are actually the same value by any means. you can see the size is used to update all the stats. bytes_read here is only used for the read at API so I left it alone

be/src/io/cache/block_file_cache_downloader.cpp
freemandealer: sure, I'd love to.

be/src/io/cache/block_file_cache_downloader.cpp
freemandealer: download_size should not exist ... 这里 download_size 如果是 0 就应该不下载。如果 0 不符合预期，应该去找上游的问题，而不是按全部文件大小去全部下载。特别是 offset != 0 时还按照文件大小下载逻辑上更不对了

be/src/io/cache/block_file_cache_downloader.cpp
freemandealer: i.e. bytes_read here is nothing like its name... kinda misleading somehow. anyway, I will update all the statistics with bytes_read to make it looks right.

gavinchou (approved):

Issue #42045, opened 2024-10-17
[Feature] Support an open api to view tablet meta for doris cloud mode


TangSiyang2001:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/doris/issues?q=is%3Aissue) and found no similar issues.

### Description

Since version 3.0, apache doris have supported Compute-Separated architecture (known as doris-cloud) based on the original Compute-Integrated architecture(known as doris-local), and both working on the same code base in this doris repo, distinct by compile options and config options. Refer to [docs](https://doris.apache.org/docs/3.0/compute-storage-decoupled/overview/). We intend to extend some workaround functionality on doris-local to doris-cloud currently.

This issue requests a BE open API to view tablet meta for doris-cloud to gain visibility as doris-local. For doris-local, we have had such an [API](https://doris.apache.org/docs/admin-manual/open-api/be-http/meta/), we need to extend it to cloud mode.

Here is an [pull request](https://github.com/apache/doris/pull/38489) as an example to support BE http API both for local mode and cloud mode.

If you are interested in this feature, you could refer to the example and start your contribution to apache doris. Welcome to contact with me to exchange ideas.

### Use case

Refer to https://doris.apache.org/docs/admin-manual/open-api/be-http/meta/

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

TangSiyang2001:
@sparklxb Assigned.

PR #55061, opened 2025-08-20
update meta_action to support both cloud and local mode

MACCCTK:

Issue Number: open #42045

Related PR: ###

Problem Summary:

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label


be/src/http/action/meta_action.cpp
```diff
@@ -67,18 +67,36 @@ Status MetaAction::_handle_header(HttpRequest* req, std::string* json_meta) {
...
+            }
+            json2pb::Pb2JsonOptions json_options;
+            json_options.pretty_json = true;
+            json_options.bytes_to_base64 = enable_byte_to_base64;
+            tablet_meta.to_json(json_meta, json_options);
+            return Status::OK();
+
+            // option 2: get meta data using ExecEnv::get_tablet
+            // same code to the local mode
+        }
```
dataroaring, line 97: Why is DATA_SIZE not handled?

MACCCTK: I have considered this point before. Currently, I have chosen not to implement this method since I am not entirely sure if my understanding of the compute-storage decoupled architecture is accurate. In local mode, data_size seems to be used to reflect the distribution of the tablet across the current BE node and other BE nodes. However, in cloud mode, BE nodes are no longer responsible for storage functions. So, I wonder is there still a need to maintain the distribution info of the tablet among BE nodes？

PR #55425, opened 2025-08-28
[fix](constant fold)Do not do BE constant fold when float/double is NaN

Jibing-Li:
### What problem does this PR solve?
Do not do BE constant fold when float/double is NaN
Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Release note

None

### Check List (For Author)

- Test 
    - [x] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

- Does this need documentation?
    - [x] No.
    - [ ] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #55874, opened 2025-09-10
branch-3.1: [fix](constant fold)Do not do BE constant fold when float/double is NaN #55425

Jibing-Li:
backport: #55425


Jibing-Li:
run buildall

morrySnow (approved):

PR #55922, opened 2025-09-11
[fix](jdbc catalog) Use SQL to get column without using ` enclosed

zy-kkk:
Sometimes the table name or database name may be a reserved keyword, so you need to surround it with `


CalvinKirs (dismissed):

zy-kkk:
run buildall

morningman (approved):

CalvinKirs (approved):

PR #55951, opened 2025-09-12
[fix](compaction) use rowset meta FS in segcompaction and add RPC client ready check

luwei16:
1 Segcompaction uses the FS from rowset meta, leveraging the functionality provided by the FS abstraction layer.
2 Add an interface in thrift_rpc_helper to check whether the RPC client is ready, avoiding core dumps when BE has just started but the RPC client is not yet initialized.


dataroaring (approved):
LGTM

gavinchou (approved):

PR #55969, opened 2025-09-12
[fix](compaction) use rowset meta FS in segcompaction and add RPC client ready check #55951

luwei16:
pick master #55951


dataroaring (approved):
LGTM

PR #54611, opened 2025-08-12
[feat](warm up) introduce immediate warmup on read cluster

zhannngchen:
### What problem does this PR solve?

Problem Summary:

introduce a config `enable_warmup_immediately_on_new_rowset`, if user set it to `true`
warm up will be triggerd automatically and immediately on syncing new rowsets

NOTE：
the method `get_rowset_warmup_state()` is used for #53540

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

- Does this need documentation?
    - [ ] No.
    - [ ] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #56737, opened 2025-10-09
[Fix](warmup) one-time and periodic warm up job should be triggered regardless whether the rowset has been warmed up before

bobhan1:

Related PR: https://github.com/apache/doris/pull/54611

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/be/src/cloud/cloud_tablet.cpp
+++ b/be/src/cloud/cloud_tablet.cpp
@@ -1662,24 +1662,25 @@ bool CloudTablet::add_rowset_warmup_state(const RowsetMeta& rowset, WarmUpTrigge
     return add_rowset_warmup_state_unlocked(rowset, source, start_tp);
 }
 
-void CloudTablet::update_rowset_warmup_state_inverted_idx_num(WarmUpTriggerSource source,
+bool CloudTablet::update_rowset_warmup_state_inverted_idx_num(WarmUpTriggerSource source,
                                                               RowsetId rowset_id, int64_t delta) {
     std::lock_guard wlock(_meta_lock);
-    update_rowset_warmup_state_inverted_idx_num_unlocked(source, rowset_id, delta);
+    return update_rowset_warmup_state_inverted_idx_num_unlocked(source, rowset_id, delta);
 }
 
-void CloudTablet::update_rowset_warmup_state_inverted_idx_num_unlocked(WarmUpTriggerSource source,
+bool CloudTablet::update_rowset_warmup_state_inverted_idx_num_unlocked(WarmUpTriggerSource source,
                                                                        RowsetId rowset_id,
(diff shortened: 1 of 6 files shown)


dataroaring (dismissed):
LGTM

be/src/cloud/cloud_tablet.cpp
```diff
@@ -1647,51 +1649,101 @@ Status CloudTablet::check_delete_bitmap_cache(int64_t txn_id,
...
-    if (_rowset_warm_up_states.contains(rowset.rowset_id())) {
-        return false;
+    auto rowset_id = rowset.rowset_id();
+
+    // Check if rowset already has warmup state
+    if (_rowset_warm_up_states.contains(rowset_id)) {
+        auto existing_state = _rowset_warm_up_states[rowset_id].state;
+
+        // For job-triggered warmup (one-time and periodic warmup), allow it to proceed
+        // except when there's already another job-triggered warmup in progress
```
freemandealer, line 1696: what if the job-triigered warmup is done? still return false?

be/src/cloud/cloud_tablet.cpp
```diff
@@ -1647,51 +1649,101 @@ Status CloudTablet::check_delete_bitmap_cache(int64_t txn_id,
...
         g_file_cache_warm_up_rowset_triggered_by_job_num << 1;
-    } else if (state == WarmUpState::TRIGGERED_BY_SYNC_ROWSET) {
+    } else if (source == WarmUpTriggerSource::SYNC_ROWSET) {
         g_file_cache_warm_up_rowset_triggered_by_sync_rowset_num << 1;
+    } else if (source == WarmUpTriggerSource::EVENT_DRIVEN) {
+        g_file_cache_warm_up_rowset_triggered_by_event_driven_num << 1;
     }
-    _rowset_warm_up_states[rowset.rowset_id()] = {
-            .state = state, .num_segments = rowset.num_segments(), .start_tp = start_tp};
+    _rowset_warm_up_states[rowset_id] = {
```
freemandealer, line 1716: Since two different trigger types will share the same states (and the same counter in it), will this replacement of elements in the map interfere with other warmup types with the same rowsetid?

bobhan1: only one trigger source can update a rowset's state in any time

freemandealer (approved):
LGTM

dataroaring (approved):
LGTM

PR #56805, opened 2025-10-10
[fix](csv) fix wrong result when escape and enclose are same

morningman:


morningman:
fixed by https://github.com/apache/doris/pull/57632

PR #56854, opened 2025-10-11
[fix](ubsan) fix ub error of uninitialized variable

morningman:
Problem Summary:
The `bool` value should be initialized, otherwise it will cause crash when enabling UBSAN

```
be/src/vec/sink/writer/vfile_result_writer.cpp:139:71: runtime error: load of value 190, which is not a valid value for type 'bool'
```

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [x] Previous test can cover this change.
        - [ ] Other reason 

- Behavior changed:
    - [x] No.
    - [ ] Yes. 

    - [x] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/be/src/pipeline/exec/result_sink_operator.h
+++ b/be/src/pipeline/exec/result_sink_operator.h
@@ -47,8 +47,8 @@ struct ResultFileOptions {
     std::vector<TParquetSchema> parquet_schemas;
     TParquetCompressionType::type parquet_commpression_type;
     TParquetVersion::type parquet_version;
-    bool parquert_disable_dictionary;
-    bool enable_int96_timestamps;
+    bool parquert_disable_dictionary = false;
+    bool enable_int96_timestamps = false;
     //note: use outfile with parquet format, have deprecated 9:schema and 10:file_properties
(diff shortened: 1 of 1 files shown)


be/src/pipeline/exec/result_sink_operator.h
```diff
@@ -47,8 +47,8 @@ struct ResultFileOptions {
     std::vector<TParquetSchema> parquet_schemas;
     TParquetCompressionType::type parquet_commpression_type;
     TParquetVersion::type parquet_version;
-    bool parquert_disable_dictionary;
```
yiguolei, line 50: why not init all the members of ResultFileOptions?

morningman: only primitive type like int and bool need init.
Other types have default constructor.

zclllyybb (approved):

hubgeter (approved):

CalvinKirs (approved):

PR #57063, opened 2025-10-16
[fix](catalog)Disable Hadoop FileSystem cache for multi-configurationobject storage catalogs

CalvinKirs:
… 

Iceberg and Paimon catalogs access object storage (OSS, OBS, etc.) using Hadoop FileSystem libraries. In environments with multiple storage configurations, the FileSystem cache may cause instances to be shared across configurations, leading to potential conflicts or incorrect access.

This PR disables the Hadoop FileSystem cache (`fs.oss.impl.disable.cache=true`) for each OSS to ensure configuration isolation and prevent cache overwrites.

If necessary, the cache can also be explicitly enabled for specific catalogs 
by setting `fs.oss.impl.disable.cache=false`.


morningman (dismissed):
LGTM

hubgeter (approved):

zy-kkk (approved):

PR #57514, opened 2025-10-30
[chore](dep)Upgrade ranger to 2.7.0

CalvinKirs:


zgxme (approved):

morningman (approved):

fe/fe-core/pom.xml
```diff
@@ -698,6 +698,10 @@ under the License.
             <groupId>com.zaxxer</groupId>
             <artifactId>HikariCP</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-clients</artifactId>
+        </dependency>
```
CalvinKirs, line 704: In the previous version, range brought in this client as a transitive dependency, but in the new version, we need to add it explicitly.
That said, we shouldn’t include ↳kafka-client at all — even for testing — since the FE module doesn’t interact with Kafka in any way.

PR #57521, opened 2025-10-30
[fix](iceberg) Support client.region and iceberg.rest.* properties invended credentials

zy-kkk:
When using Iceberg REST catalog with vended credentials enabled:
  - **Polaris** returns `s3.region`, `s3.access-key-id`, etc.
  - **Gravitino** returns `client.region`, `s3.access-key-id`, etc. (following Iceberg official naming)

  Current implementation only filters properties with `s3.`, `oss.`, etc. prefixes, causing `client.region` to be filtered out, which leads to `No storage
  properties found for schema: s3` error.


zy-kkk:
run buildall

morningman (approved):
LGTM

CalvinKirs (approved):

PR #58322, opened 2025-11-25
[regression](filecache) Update eviction size assertion in TTL LRU test

freemandealer:
because the data loaded into doris are compressed, so the origin assertion of amount of data evict by overfill the cache would be to large for the compressed data.


freemandealer:
run buildall

gavinchou (approved):

Yukang-Lian (approved):

koarz (approved):

PR #58105, opened 2025-11-18
[fix](nereids)allow sync mv has same columns as base table when it has where clause

starocean999:
```
create table d_table(
    k1 int null,
    k2 int not null,
    k3 bigint null,
    k4 varchar(100) null
)
duplicate key (k1,k2,k3)
distributed BY hash(k1) buckets 3
properties("replication_num" = "1");

create materialized view kx1234 as select k1 as y1, k2 as y2, k3 as y3, k4 as y4 from d_table where k1 = 1;

before this pr, the create mv command will fail because the columns in the select list is same as base table. 
But the mv is useful, because it has where clause to filter some data. This pr correct it
```
### What problem does this PR solve?

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

- Does this need documentation?
    - [ ] No.
    - [ ] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #58342, opened 2025-11-25
branch-3.1: [fix](nereids)allow sync mv has same columns as base table when it has where clause #58105

starocean999:
picked from #58105


starocean999:
run buildall

morrySnow (approved):

PR #48462, opened 2025-02-28
[fix](information_schema) The time type in tables under the information_schema should change according to the time_zone.

BePPPower:
Problem Summary:
Fix the issue where the time type in tables under the information_schema does not change with the `time_zone` session variable.

### Release note

None

### Check List (For Author)

- Test 
    - [x] Regression test
    - [ ] Unit Test
    - [ ] Manual test (add detailed scripts or steps below)
    - [ ] No need to test or manual test. Explain why:
        - [ ] This is a refactor/code format and no logic has been changed.
        - [ ] Previous test can cover this change.
        - [ ] No code files have been changed.
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [x] Yes. 

- Does this need documentation?
    - [x] No.
    - [ ] Yes. 

### Check List (For Reviewer who merge this PR)

- [ ] Confirm the release note
- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

PR #58412, opened 2025-11-26
[fix](information_schema) Fix timezone format incompatibility when using offset timezone

zy-kkk:
Related PR: #48462

  When `time_zone` is set to offset format like `+08:00`, `cctz::time_zone::name()`
  returns internal format `Fixed/UTC+08:00:00` which Java `ZoneId.of()` cannot parse.

  Use the original timezone string instead of `_timezone_obj.name()` when passing
  timezone to FE via thrift RPC.


morningman (dismissed):
LGTM

zy-kkk:
run buildall

morningman (approved):
LGTM

CalvinKirs (approved):

PR #58467, opened 2025-11-27
[[fix](rest-s3)Set AWS Request Checksum Calculation only if not set void issues on non-S3 storage

CalvinKirs:
AWS SDK for Java v2 uses the system property AwsSystemSetting.AWS_REQUEST_CHECKSUM_CALCULATION to control request payload checksum calculation.

According to the official discussion AWS SDK v2 Discussion https://github.com/aws/aws-sdk-java-v2/discussions/5802 

By default, the SDK may automatically calculate checksums when required.

For AWS S3, setting the default value "WHEN_REQUIRED" ensures checksums are calculated only when necessary, providing both compatibility and performance benefits.

For non-S3 object storage (such as COS, OSS, OBS), using this default value may cause errors, for example:

`aws-chunked encoding is not supported with the specified x-amz-content-sha256 value`

because these services do not support AWS-specific chunked encoding or SHA256 checksum.


CalvinKirs:
run buildall

CalvinKirs:
run buildall

morningman (approved):
LGTM

zy-kkk (approved):

PR #58516, opened 2025-11-28
[fix](unique function) fix push unique function through join

yujun777:

For those expression above join,  if it contains unique function,  then cann't push through or inside join.  For example,  for sql `t1 join t2 where random() < 0.5`,   suppose t1 contains m rows,  t2 contains n rows,  then this sql expect to contains about  m * n * 0.5 rows.   If we push this random below join's child,  it will rewrite as `(t1 where random() < 0.5)  join (t2 where random() < 0.5)`,  then it will expect to contains about (m*0.5) * (n*0.5) = m * n * 0.25 rows,   which is less than the origin sql.

For join reorder rule,  if a  filter's child is join,  this rule may extract conjuncts from the filter and push them into the join,   then it will have the problem too,  so we need to disable this rule when the filter contains unique function.

### Release note

None

### Check List (For Author)

- Test 
    - [x] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label


morrySnow (dismissed):

924060929 (dismissed):

yujun777:
run buildall

924060929 (dismissed):

morrySnow (dismissed):

924060929 (approved):

morrySnow (approved):

PR #58566, opened 2025-12-01
[fix](S3)Downgrade the AWS SDK version to avoid impacting other S3-compatible services.

CalvinKirs:
…

This PR addresses an issue related to AWS SDK for Java v2 S3 client behavior described in: https://github.com/aws/aws-sdk-java-v2/issues/5805

Starting from newer versions of the SDK, the S3 client introduces unexpected behavior that causes our application to fail during normal S3 operations. Due to this problem, we were forced to downgrade aws-s3 from the latest version to 2.29.x, which restored correct functionality. The regression affects stability in our production environment and prevents us from upgrading the SDK.

#57226


hubgeter (approved):

morningman (approved):
LGTM

PR #58574, opened 2025-12-01
 [fix](jdbc) Fix JNI global reference leak in JdbcConnector close

zy-kkk:
## Problem
  JNI Global References may not be deleted when `JdbcConnector::close()` fails
  midway due to early returns from `RETURN_IF_ERROR` macros. This prevents Java
  GC from collecting `BaseJdbcExecutor` and its associated `ResultSet` objects,
  causing old generation heap to grow to 99%+.

  ## Solution
  - C++: Always delete all three Global References regardless of Java close() failure
  - Java: Always clear member variable references (`resultSet`, `stmt`, `conn`) in finally block


zhangstar333 (approved):

morningman (approved):
LGTM

PR #58603, opened 2025-12-02
[fix](iceberg) Fix partition value conversion for Iceberg data file generation

suxiaogang223:
## Problem
INSERT operations fail for Iceberg tables partitioned by TIMESTAMP type. The root cause is that `DataFiles.Builder.withPartitionValues()` does not support TIMESTAMP partition types, which prevents proper data file generation for timestamp-partitioned tables.

## Root Cause
The code was using `DataFiles.Builder.withPartitionValues()` which accepts a `List<String>` of human-readable partition values. However, this method has limitations and does not properly support TIMESTAMP partition types, causing INSERT operations to fail for tables partitioned by timestamp columns.

## Solution
This PR fixes the issue by switching from `withPartitionValues()` to `withPartition()` method, which accepts a `PartitionData` object. The solution includes:

1. **Added `parsePartitionValueFromString()` method in `IcebergUtils.java`**:
   - Converts human-readable partition value strings to appropriate Java types based on Iceberg type
   - Supports all partition types: STRING, INTEGER, LONG, FLOAT, DOUBLE, BOOLEAN, DATE, TIMESTAMP, DECIMAL
   - For DATE: converts "yyyy-MM-dd" format to epoch days
   - For TIMESTAMP: uses Doris's built-in datetime parser to handle various timestamp formats and converts to microseconds

2. **Added `parseTimestampToMicros()` helper method**:
   - Parses timestamp strings using Doris's `DateLiteral.parseDateTime()` for format flexibility
   - Converts parsed timestamps to microseconds since epoch
   - Handles timezone-aware and timezone-naive timestamps

3. **Refactored `IcebergWriterHelper.java`**:
   - Changed from using `List<String>` partition values to `PartitionData` 
   - Added `convertToPartitionData()` method to convert human-readable partition values from Backend to Iceberg's `PartitionData` format
   - Updated `genDataFile()` to use `PartitionData` and call `DataFiles.Builder.withPartition()` instead of `withPartitionValues()`

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label


morningman (approved):
LGTM

CalvinKirs (approved):

PR #58644, opened 2025-12-02
branch-3.1: [fix](mtmv) Fix sync materialized view use wrong check method when in the same regression test database #58245

seawinde:
picked from #58245


seawinde:
run buildall

morrySnow (approved):

PR #58656, opened 2025-12-03
[Fix](jsonb) tolerate invalid jsonb document with size 0 #58523

### LINKED ISSUE #58523: [Bug] Invalid JSONB document: too small size(0) or null pointer
### Search before asking

### Version

V3.1.2

### What's Wrong?

Hi, after updating to version 3.1.2 from 2.1.11 I noticed that compaction fails constantly leading to versions accumulating. 

By inspecting the tablet: 
```
{
    "cumulative policy type": "size_based",
    "cumulative point": 490516,
    "last cumulative failure time": "2025-11-28 09:09:52.925",
    "last base failure time": "1970-01-01 01:00:00.000",
    "last full failure time": "1970-01-01 01:00:00.000",
    "last cumulative success time": "1970-01-01 01:00:00.000",
    "last base success time": "1970-01-01 01:00:00.000",
    "last full success time": "1970-01-01 01:00:00.000",
    "last cumulative schedule time": "2025-11-28 09:10:04.419",
    "last base schedule time": "1970-01-01 01:00:00.000",
    "last full schedule time": "1970-01-01 01:00:00.000",
    "last cumulative status": "[INVALID_ARGUMENT]Invalid JSONB document: too small size(0) or null pointer",
    "last base status": "",
    "last full status": "",
}
```
In be.INFO multiple entries with:
`W20251128 15:19:49.485358 3219940 tablet.cpp:1858] failed to do cumulative compaction, tablet=27034405 : [INVALID_ARGUMENT]Invalid JSONB document: too small size(0) or null pointer`

### What You Expected?

Same behavior as in V2.1.11

### How to Reproduce?

_No response_

### Anything Else?

_No response_

### Code of Conduct

Hongyang66666:
## Proposed changes

Issue Number: close #58523

### Problem
When upgrading Doris from older versions to newer versions, legacy data might contain empty jsonb values (where `size` is 0).

The current strict validation in `checkAndCreateDocument` throws an `InvalidArgument` error ("Invalid JSONB document: too small size(0)") when encountering such data. This causes critical failures during the compaction process.

### Fix
Modify `be/src/util/jsonb_document.h` to add a compatibility check.
If `size == 0`, the function now returns `Status::OK()` (treating it as a valid empty/null object) instead of throwing an error. This ensures compaction and queries can proceed normally on legacy data.

## Check List
- [x] The code has been compiled and tested locally.
- [x] This PR fixes a bug.
- [ ] This PR refactors code.
- [ ] This PR adds a new feature.

### DIFF
--- a/be/src/util/jsonb_document.h
+++ b/be/src/util/jsonb_document.h
@@ -1042,6 +1042,19 @@ struct ArrayVal : public ContainerVal {
 inline Status JsonbDocument::checkAndCreateDocument(const char* pb, size_t size,
                                                     JsonbDocument** doc) {
     *doc = nullptr;
+    // Fix Issue #58523: Tolerate empty data from legacy versions.
+    // If size is 0, we return a static valid "Null" document.
+    if (size == 0) {
+        // 手动构造一个合法的 Null 文档二进制数据
+        // 第1个字节：Version = 1 (JSONB_VER)
+        // 第2个字节：Type    = 0 (T_Null)
+        static char s_null_buf[] = {1, 0};
+
+        // 将 doc 指向这个静态的内存区域
+        *doc = reinterpret_cast<JsonbDocument*>(s_null_buf);
+        return Status::OK();
+    }
+
     if (!pb || size < sizeof(JsonbHeader) + sizeof(JsonbValue)) {
         return Status::InvalidArgument("Invalid JSONB document: too small size({}) or null pointer",
                                        size);


hello-stephen:

be/src/util/jsonb_document.h
```diff
@@ -1042,6 +1042,11 @@ struct ArrayVal : public ContainerVal {
 inline Status JsonbDocument::checkAndCreateDocument(const char* pb, size_t size,
                                                     JsonbDocument** doc) {
     *doc = nullptr;
+    // Fix Issue #58523: Tolerate empty data from legacy versions, treat as NULL to avoid errors.
```
yiguolei: If return OK here, the other logic will be ok???
For example, 
JsonbDocument* doc = nullptr;
            THROW_IF_ERROR(JsonbDocument::checkAndCreateDocument(
                    result.writer->getOutput()->getBuffer(), result.writer->getOutput()->getSize(),
                    &doc));
            result.value = doc->getValue();
I think it will core here.

be/src/util/jsonb_document.h
```diff
@@ -1042,6 +1042,11 @@ struct ArrayVal : public ContainerVal {
...
+    if (size == 0) {
+        return Status::OK();
```
mrhhsg, line 1055: You can’t simply return OK here, `doc` needs to be set, otherwise it may lead to a null pointer error.

mrhhsg (changes requested):

be/src/util/jsonb_document.h
Hongyang66666: @yiguolei Thanks for the review! You are absolutely right.
Returning OK with a nullptr doc will indeed cause a crash in the caller.
I have updated the patch. Now, if size == 0, I initialize a static valid Null JsonbDocument (using JsonbWriter to construct it once) and assign it to *doc. This ensures the caller gets a safe, valid object representing a JSON null.
Please verify the latest commit.

be/src/util/jsonb_document.h
Hongyang66666: @mrhhsg Thanks for the review! You are absolutely right.
Returning OK with a nullptr doc will indeed cause a crash in the caller.
I have updated the patch. Now, if size == 0, I initialize a static valid Null JsonbDocument (using JsonbWriter to construct it once) and assign it to *doc. This ensures the caller gets a safe, valid object representing a JSON null.
Please verify the latest commit.

yiguolei: If two threads are calling checkAndCreateDocument and try to update s_null_buf, what will happen?

yiguolei (dismissed):

Hongyang66666 (commented):
Hi @yiguolei @mrhhsg, thanks for the review.

I have addressed the potential crash issue. 
Instead of returning `nullptr`, I now check if `size == 0` and return a static valid Null JsonbDocument (constructed with `{1, 0}`). 
This ensures `*doc` is always valid and safe to use by the caller.

Please review again, thanks!

be/src/util/jsonb_document.h
```diff
@@ -1042,6 +1042,19 @@ struct ArrayVal : public ContainerVal {
 inline Status JsonbDocument::checkAndCreateDocument(const char* pb, size_t size,
                                                     JsonbDocument** doc) {
     *doc = nullptr;
+    // Fix Issue #58523: Tolerate empty data from legacy versions.
+    // If size is 0, we return a static valid "Null" document.
+    if (size == 0) {
+        // 手动构造一个合法的 Null 文档二进制数据
+        // 第1个字节：Version = 1 (JSONB_VER)
+        // 第2个字节：Type    = 0 (T_Null)
+        static char s_null_buf[] = {1, 0};
```
yiguolei, line 1051: 1. 在beut 里增加一个case，验证一下 这种构造方式构造出的jsonbdocument 的version = 1 and type = T NULL。防止未来有人重构代码，把jsonb的bytes 结构改乱了。
2. 在beut 中还要验证一下把这个null jsonb document 转成一个bytes 数组，然后可以checkAndCreateDocument 再创建出这个null jsonb document。

be/src/util/jsonb_document.h
```diff
@@ -1042,6 +1042,19 @@ struct ArrayVal : public ContainerVal {
 inline Status JsonbDocument::checkAndCreateDocument(const char* pb, size_t size,
                                                     JsonbDocument** doc) {
     *doc = nullptr;
+    // Fix Issue #58523: Tolerate empty data from legacy versions.
+    // If size is 0, we return a static valid "Null" document.
+    if (size == 0) {
+        // 手动构造一个合法的 Null 文档二进制数据
```
yiguolei, line 1048: 所有的注释都得使用英文。

PR #58728, opened 2025-12-04
[performance](exec) fix zone map in effective in column preidate

HappenLee:

When a page contains some NULL values, we set both the min value and its associated flag to NULL for filtering purposes in IN and comparison operations. In such cases, we can directly compare the min/max values to determine whether the zone map is effective, without needing to handle the scenario where the min value is NULL. However, if all values in the page are NULL, the entire page can be directly filtered out.

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/be/src/olap/comparison_predicate.h
+++ b/be/src/olap/comparison_predicate.h
@@ -119,8 +119,8 @@ public:
     }
 
     bool evaluate_and(const std::pair<WrapperField*, WrapperField*>& statistic) const override {
-        if (statistic.first->is_null()) {
-            return true;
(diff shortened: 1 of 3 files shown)


BiteTheDDDDt (approved):
LGTM

zhangstar333 (approved):

PR #58828, opened 2025-12-08
[Opt](tbl_priv) Optimize PrivTable search and insert with ordered lookup

linrrzqqq:

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

Previous collection of user table permission entries would perform something similar to the following:
```java
bool hasEntry = false;
for (auto entry : entries) {
    if (entry == newEntry)
        hasEntry = true;
}

if (!hasEntry){
	entries.add(newEntry);
	Collections.sort(entries);
} else {
    mergeTwoEntries();
}
```
now we locate existing entries by binary search and insert at the proper position to keep ordering, merging privilege bits when a matching key exists.

Before:
```text
Doris> SELECT COUNT(*) FROM ( SELECT * FROM information_schema.table_privileges ) t;
+----------+
| COUNT(*) |
+----------+
|    50000 |
+----------+
1 row in set (29.59 sec)
```
After
```text
Doris> SELECT COUNT(*) FROM ( SELECT * FROM information_schema.table_privileges ) t;
+----------+
| COUNT(*) |
+----------+
|    50000 |
+----------+
1 row in set (0.46 sec)
```

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/PrivTable.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/PrivTable.java
@@ -81,7 +81,6 @@ public abstract class PrivTable {
             }
             int insertPos = -idx - 1;
             entries.add(insertPos, newEntry);
-            LOG.info("add priv entry: {}", newEntry);
             return newEntry;
         }
 
@@ -91,9 +90,6 @@ public abstract class PrivTable {
         }
 
         mergePriv(existingEntry, newEntry);
-        if (LOG.isDebugEnabled()) {
-            LOG.debug("merge priv entry: {}", existingEntry);
-        }
         return existingEntry;
     }
 
--- a/fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/PrivTable.java
+++ b/fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/PrivTable.java
@@ -74,24 +74,25 @@ public abstract class PrivTable {
             }
         }
 
-        PrivEntry existingEntry = getExistingEntry(newEntry);
-        if (existingEntry == null) {
+        int idx = Collections.binarySearch(entries, newEntry);
+        if (idx < 0) {
             if (errOnNonExist) {
                 throw new DdlException("entry does not exist");
             }
-            entries.add(newEntry);
-            Collections.sort(entries);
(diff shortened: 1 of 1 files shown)


fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/PrivTable.java
```diff
@@ -74,24 +74,25 @@ public PrivEntry addEntry(PrivEntry entry, boolean errOnExist, boolean errOnNonE
...
+        int idx = Collections.binarySearch(entries, newEntry);
+        if (idx < 0) {
             if (errOnNonExist) {
                 throw new DdlException("entry does not exist");
             }
-            entries.add(newEntry);
-            Collections.sort(entries);
+            int insertPos = -idx - 1;
+            entries.add(insertPos, newEntry);
             LOG.info("add priv entry: {}", newEntry);
```
HappenLee: why here each entry need log the info ？

linrrzqqq: I think there's no need for log every info, and it's easy to cause fe.log flooded.

fe/fe-core/src/main/java/org/apache/doris/mysql/privilege/PrivTable.java
```diff
@@ -74,25 +74,22 @@ public PrivEntry addEntry(PrivEntry entry, boolean errOnExist, boolean errOnNonE
...
-            } else {
-                mergePriv(existingEntry, newEntry);
-                if (LOG.isDebugEnabled()) {
-                    LOG.debug("merge priv entry: {}", existingEntry);
-                }
-            }
         }
+
+        PrivEntry existingEntry = entries.get(idx);
+        if (errOnExist) {
```
HappenLee, line 87: move the check up the line 87

HappenLee (approved):
LGTM

CalvinKirs (approved):

PR #58965, opened 2025-12-11
[perf](filecache) support memory storage cache sharding to improve concurrency

HaolinGuan7211:


be/src/io/cache/shard_mem_cache.cpp
freemandealer, line 1: it's strange to have an empty source file here

HaolinGuan7211: I forget to put the implementation here ....

be/src/io/cache/file_cache_storage.h
```diff
@@ -43,6 +43,10 @@ struct FileWriterMapKeyHash {
     }
 };
 
+struct MemBlock {
```
freemandealer, line 46: this structure is used only in mem storage. I see no reason to move it to the base class header. so keep it as it was?

HaolinGuan7211: You are right, I put it here only for correctly compiling at that time for simply verified the performance of sharding. Let me make it elegant tomorrow.

PR #58993, opened 2025-12-12
[Feat](opt) Optimize Python UDF/UDAF/UDTF for Doris

sjyango:

Issue Number: close #xxx

Related PR: #xxx

Problem Summary:

### Release note

None

### Check List (For Author)

- Test 
    - [ ] Regression test
    - [ ] Unit Test
        - [ ] Other reason 

- Behavior changed:
    - [ ] No.
    - [ ] Yes. 

    - [ ] No.
    - [ ] Yes. 

- [ ] Confirm test cases
- [ ] Confirm document
- [ ] Add branch pick label


Thearas:

be/src/vec/aggregate_functions/aggregate_function_python_udaf.cpp
```diff
@@ -200,18 +199,17 @@ void AggregatePythonUDAF::create(AggregateDataPtr __restrict place) const {
     new (place) Data();
     DCHECK(reinterpret_cast<Data*>(place)) << "Place must not be null";
 
-    // Get or initialize shared client (thread-safe)
-    PythonUDAFClient* client = _get_shared_client();
+    // Get or create thread-local client (thread-safe)
```
HappenLee: what happen if thread change in agg operator

be/src/vec/aggregate_functions/aggregate_function_python_udaf.cpp
```diff
@@ -247,15 +252,53 @@ void AggregatePythonUDAF::add(AggregateDataPtr __restrict place, const IColumn**
...
+
+    std::shared_ptr<arrow::RecordBatch> batch;
+    Status batch_status = convert_to_arrow_batch(input_block, schema, arrow::default_memory_pool(),
+                                                 &batch, timezone_obj);
+    if (UNLIKELY(!batch_status.ok())) {
+        throw doris::Exception(ErrorCode::INTERNAL_ERROR, batch_status.to_string());
+    }
+
+    // Get client from first place (all places in same thread share the same client)
+    auto client = this->data(places[0] + place_offset).client;
```
HappenLee: what happen if thread change in agg operator different place may have different client

sjyango: revert the add_batch's optimized code, because it has bug when thread changing

be/src/udf/python/python_udf_runtime.h
```diff
@@ -81,7 +73,23 @@ class PythonUDFProcess {
 
     bool is_alive() const {
```
HappenLee: need more comment for the two api, why we need the two api?

be/src/udf/python/python_server.cpp
```diff
@@ -33,51 +34,130 @@
...
-        if (_pools.find(version) == _pools.end()) {
-            PythonUDFProcessPoolPtr new_pool = std::make_unique<PythonUDFProcessPool>(
-                    version, config::max_python_process_nums, config::min_python_process_nums);
-            RETURN_IF_ERROR(new_pool->init());
-            _pools[version] = std::move(new_pool);
+Status PythonServerManager::get_process(const PythonVersion& version, ProcessPtr* process) {
+    std::lock_guard<std::mutex> lock(_pools_mutex);
+    std::vector<ProcessPtr>& pool = _process_pools[version];
+
+    if (pool.empty()) {
```
HappenLee: unlikely

be/src/vec/aggregate_functions/aggregate_function_python_udaf.cpp
```diff
@@ -59,40 +61,83 @@ Status AggregatePythonUDAFData::add(int64_t place_id, const IColumn** columns,
...
+        input_block.insert(
+                ColumnWithTypeAndName(columns[i]->get_ptr(), argument_types[i], std::to_string(i)));
+    }
+
+    auto places_col = ColumnInt64::create(num_rows);
+    auto& places_data = places_col->get_data();
 
+    // Fill places column with place IDs for the slice [start, end)
+    for (size_t i = start; i < end; ++i) {
+        places_data[i] = reinterpret_cast<int64_t>(places[i] + place_offset);
```
HappenLee, line 98: if each udaf have different place id，maybe we have too big hash table?

be/src/udf/python/python_udaf_client.h
```diff
@@ -27,6 +27,20 @@ class PythonUDAFClient;
...
+// Fixed-size binary metadata structure for UDAF operations (Request)
+// Total size: 34 bytes (removed client_id - place_id is globally unique)
+struct __attribute__((packed)) UDAFMetadata {
+    uint8_t operation; // 1 byte: UDAFOperation enum
+    int64_t place_id;  // 8 bytes: aggregate state identifier (globally unique)
+    // ACCUMULATE-specific metadata (25 bytes)
+    uint8_t is_single_place; // 1 byte: boolean (0 or 1)
+    int64_t row_start;       // 8 bytes: start row index
+    int64_t row_end;         // 8 bytes: end row index (exclusive)
+    int64_t place_offset;    // 8 bytes: offset within each place
```
HappenLee: the place_offset seems always zero it's really useful?

be/src/udf/python/python_udaf_client.cpp
```diff
@@ -217,329 +110,419 @@ Status PythonUDAFClient::accumulate(int64_t place_id, bool is_single_place,
...
+            binary_builder.Append(binary_data->data(), static_cast<int32_t>(binary_data->size())));
+    RETURN_DORIS_STATUS_IF_ERROR(binary_builder.Finish(&binary_array));
+    columns.push_back(binary_array);
+
+    *out = arrow::RecordBatch::Make(_schema, 1, columns);
+    return Status::OK();
+}
+
+Status PythonUDAFClient::_create_empty_request_batch(std::shared_ptr<arrow::RecordBatch>* out) {
+    std::vector<std::shared_ptr<arrow::Array>> columns;
```
HappenLee, line 497: maybe only need create one time, after create, all op maybe use the same batch. because the batch seems unless?

be/src/udf/python/python_udf_runtime.h
sjyango: I've merged these two APIs into one, using a `force_check` parameter to distinguish between them. The reason for this distinction is that invoking `running` on a `boost::process::child` to check the process status triggers a system call. Frequent calls would result in excessive context switches between user space and kernel space, negatively impacting performance. Therefore, I've implemented caching here—`boost::process::child::running()` is actually called only once every 5,000 invocations.

be/src/udf/python/python_udaf_client.h
sjyango: That's right—place_offset is indeed redundant here, so I've removed it.

be/src/udf/python/python_udaf_client.cpp
sjyango: That's a good optimization point! already fixed.

be/src/udf/python/python_udf_runtime.h
sjyango: these two APIs have removed! There is no need to check process status before call GRPC APIs

